PR #10941, opened 2021-07-05
[squeezebox] Cleanup / code simplification

maniac103:
Improve code readability a bit by
- using lambdas for anonymous listener instantiations
- decoding URL encoded responses earlier so the code can check against non-encoded responses

### DIFF
--- a/bundles/org.openhab.binding.squeezebox/src/main/java/org/openhab/binding/squeezebox/internal/handler/SqueezeBoxServerHandler.java
+++ b/bundles/org.openhab.binding.squeezebox/src/main/java/org/openhab/binding/squeezebox/internal/handler/SqueezeBoxServerHandler.java
@@ -584,12 +584,7 @@ public class SqueezeBoxServerHandler extends BaseBridgeHandler {
                 // Save player if we haven't seen it yet
                 if (!players.containsKey(macAddress)) {
                     players.put(macAddress, player);
-                    updatePlayer(new PlayerUpdateEvent() {
-                        @Override
-                        public void updateListener(SqueezeBoxPlayerEventListener listener) {
-                            listener.playerAdded(player);
-                        }
-                    });
+                    updatePlayer(listener -> listener.playerAdded(player));
                     // tell the server we want to subscribe to player updates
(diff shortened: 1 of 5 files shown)


bundles/org.openhab.binding.squeezebox/src/main/java/org/openhab/binding/squeezebox/internal/handler/SqueezeBoxServerHandler.java
```diff
@@ -997,25 +908,32 @@ private void handleFavorites(String message) {
...
             boolean isTypePlaylist = false;
             for (String part : messageParts) {
+                final String decoded = decode(part);
+                int colonPos = decoded.indexOf(":");
+                if (colonPos < 0) {
+                    continue;
+                }
+
+                final String key = decoded.substring(0, colonPos);
+                final String value = decoded.substring(colonPos);
```
mhilbush, line 918: Shouldn't this be `colonPos + 1`?

mhilbush (changes requested):
I left one comment for you to check.

Otherwise, LGTM

bundles/org.openhab.binding.squeezebox/src/main/java/org/openhab/binding/squeezebox/internal/handler/SqueezeBoxServerHandler.java
maniac103: Oops, of course. Thanks, fixed.

maniac103: NB: I thought about factoring out this parsing to a separate method, but due the lack of a destructuring binding in Java (e.g. Kotlin: `val (key, value) = ...` this didn't really improve readability.

maniac103: Ended up going a step back and parsed the full CLI response into a map.

kaikreuzer (approved):
Thanks!

Issue #10839, opened 2021-06-09
[mqtt.homie] Missing support for "Duration" and "DateTime"


ccoors:
## Expected Behavior
`org.openhab.binding.mqtt.homie` should support [Duration](https://homieiot.github.io/specification/spec-core-v3_0_0/#duration) and [DateTime](https://homieiot.github.io/specification/spec-core-v3_0_0/#datetime) as specified in the Homie 3.0.0 specification.

## Current Behavior
Adding a Homie-Thing in openHAB leads to multiple exceptions looking like this on the console:

```
java.lang.IllegalArgumentException: No enum constant org.openhab.binding.mqtt.homie.internal.homie300.PropertyAttributes.DataTypeEnum.duration_
        at java.base/java.lang.Enum.valueOf(Enum.java:240)
        at org.openhab.binding.mqtt.generic.mapping.SubscribeFieldToMQTTtopic.numberConvert(SubscribeFieldToMQTTtopic.java:114)
        at org.openhab.binding.mqtt.generic.mapping.SubscribeFieldToMQTTtopic.processMessage(SubscribeFieldToMQTTtopic.java:152)
        at org.openhab.core.io.transport.mqtt.internal.Subscription.processMessage(Subscription.java:85)
```
<details>
<summary>Click to see more of the exception...</summary>

```
        at org.openhab.core.io.transport.mqtt.internal.Subscription.lambda$1(Subscription.java:81)
        at java.base/java.util.stream.ForEachOps$ForEachOp$OfRef.accept(ForEachOps.java:183)
        at java.base/java.util.concurrent.ConcurrentHashMap$KeySpliterator.forEachRemaining(ConcurrentHashMap.java:3566)
        at java.base/java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:484)
        at java.base/java.util.stream.ForEachOps$ForEachTask.compute(ForEachOps.java:290)
        at java.base/java.util.concurrent.CountedCompleter.exec(CountedCompleter.java:746)
        at java.base/java.util.concurrent.ForkJoinTask.doExec(ForkJoinTask.java:290)
        at java.base/java.util.concurrent.ForkJoinTask.doInvoke(ForkJoinTask.java:408)
        at java.base/java.util.concurrent.ForkJoinTask.invoke(ForkJoinTask.java:736)
        at java.base/java.util.stream.ForEachOps$ForEachOp.evaluateParallel(ForEachOps.java:159)
[...]

PR #10946, opened 2021-07-05
[mqtt.homie] Workaround for #10839. DateTime and Duration

THKDev:
This commit adds support for DateTime and Duration payload. Duration is a workaround only. Duration will be displayed as it is like "P0DT0H10M9S". Maybe a DurationType and DurationItem needs be added to "openhab.core".

### DIFF
--- a/bundles/org.openhab.binding.mqtt.homie/src/main/java/org/openhab/binding/mqtt/homie/internal/homie300/Property.java
+++ b/bundles/org.openhab.binding.mqtt.homie/src/main/java/org/openhab/binding/mqtt/homie/internal/homie300/Property.java
@@ -26,16 +26,17 @@ import org.eclipse.jdt.annotation.NonNullByDefault;
 import org.eclipse.jdt.annotation.Nullable;
 import org.openhab.binding.mqtt.generic.ChannelConfigBuilder;
 import org.openhab.binding.mqtt.generic.ChannelState;
-import org.openhab.binding.mqtt.generic.mapping.AbstractMqttAttributeClass;
 import org.openhab.binding.mqtt.generic.mapping.AbstractMqttAttributeClass.AttributeChanged;
 import org.openhab.binding.mqtt.generic.mapping.ColorMode;
 import org.openhab.binding.mqtt.generic.values.ColorValue;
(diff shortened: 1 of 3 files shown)


fwolter:
You could use `QuantityType` (`Number:Time`) for the duration.

THKDev:
Not sure how to use Number:Time. Here is my implementation.

```java
public class DurationValue extends Value {
    public DurationValue() {
        super(CoreItemFactory.NUMBER, Stream.of(QuantityType.class, StringType.class).collect(Collectors.toList()));
    }

    @Override
    public void update(final Command command) throws IllegalArgumentException {
        if (command instanceof QuantityType) {
            state = ((QuantityType) command);
        } else {
            final Duration duration = Duration.parse(command.toString());
            state = new QuantityType<>(duration.toSeconds(), Units.SECOND);
        }
    }

    @Override
    public String getMQTTpublishValue(@Nullable String pattern) {
        if (state == UnDefType.UNDEF) {
            return "";
        }
        return state.toFullString();
    }
}
```

The result is "1970-01-01T00:05:06Z" . I am missing the place where the formatting of the output is done. I think the result should look like this: 0 Year, 10 Day, 5 Hours and 10 Minutes .

fwolter:
You can set the format in the state pattern of the channel. See here for example: https://github.com/openhab/openhab-addons/blob/012254a0acbb7fe790f0893b23fe21dce00b7674/bundles/org.openhab.binding.luxtronikheatpump/src/main/resources/OH-INF/thing/channels.xml#L299

THKDev:
I tried this with the following.

```xml
	<channel-type id="duration">
		<item-type>Number:Time</item-type>
		<label>Duration Value</label>
		<description>A duration of time</description>
		<state pattern="%d %unit%" readOnly="true"/>
	</channel-type>
```

But this does not work. The "homie" binding is dynamic and i see no connection between DurationValue and the channel-type  definition.

lolodomo:
Marked as "stale" as there is no more discussion since 8 months.

lolodomo:
Closing that PR as there is no activity/feedback since more than one year.
@THKDev : feel free to reopen it in case you would like to finish.

PR #10949, opened 2021-07-06
[enocean] Adds support for the D2_06_50 EEP type

DrRSatzteil:
Signed-off-by: Thomas Lauterbach <lauterbachthomas@gmail.com>

[enocean] Adds support for the D2_06_50 EEP type (window sash and handle sensors)

This PR adds a new device type windowSashHandleSensor (channels: handle state, sash state, breach alarm trigger, battery level, battery low) with the EEP profile D2_60_50 as used by Siegenia Senso Secure devices. I'm aware that these devices could be additionally configured with a proprietary teach-in cmd, however since I do not own their gateway and this does not seem to be documented somewhere just a regular tech-in cmd is used.


DrRSatzteil:
I still found two minor issues that I will have to look into. Maybe you can give me some guidance:

1. When adding the sensor through the auto discovery it is automatically assigned the EEP type D2_06_50_ (note the trailing underscore) as shown in this screenshot:
Screenshot_2021-07-08-18-20-08-580_org openhab habdroid
Consequently the EEP selection in the UI is not ticked:
Screenshot_2021-07-08-18-16-44-336_org openhab habdroid
Everything works fine however until I tick the selection: then I see an error that the message type D2_06_50 is not known.
2. When I select the trigger channel in a rule via Main UI it does not show the label of the channel but this: 
Screenshot_2021-07-08-18-16-09-062_org openhab habdroid


As I said I will try to find the reason behind this by myself but any help is of course appreciated 😊

DrRSatzteil:
Just checked with other bindings and they behave the same. This seems to be either "as designed" or a more general bug. So I won't investigate any further

fruggy83:
Hi @DrRSatzteil 

I would also guess that this behaviour ist correct. If you create a trigger rule by dsl, you also have to set the channel Id (this is what you see here). If you do not want to use the channel Id, you still have to use the thing name and channel name  to clearly define the trigger channel. So I guess they just use channel Id to make things easy :)

By the way, you have requested a review but you are still pushing commits to this pr. Could you revoke the review request or give me sign when you are finished and I can start with the review? Thanks a lot.

DrRSatzteil:
Hi @fruggy83 

Sorry, yes I shouldn't have requested the review already. I converted this to a draft again and will review my changes once more. I realised that my last commit requires some Readme changes. Once I'm done with that I'll give you a sign. Thank you!

DrRSatzteil:
I udpated the README and I did not discover any more problems. This should be ready for review.

bundles/org.openhab.binding.enocean/src/main/resources/OH-INF/thing/WindowSashHandleSensor.xml
```diff
@@ -0,0 +1,37 @@
...
+	xmlns:thing="https://openhab.org/schemas/thing-description/v1.0.0"
+	xsi:schemaLocation="https://openhab.org/schemas/thing-description/v1.0.0 https://openhab.org/schemas/thing-description-1.0.0.xsd">
+
+	<!-- WindowSashHandleSensor -->
+	<thing-type id="windowSashHandleSensor">
+		<supported-bridge-type-refs>
+			<bridge-type-ref id="bridge"/>
+		</supported-bridge-type-refs>
+
+		<label>Window Sash and Handle Sensor</label>
```
fwolter, line 13: Labels are expected to be as short as possible. Guideline is 2-3 words with max 20-25 chars. See https://www.openhab.org/docs/developer/bindings/thing-xml.html#formatting-labels-and-descriptions

fwolter (changes requested):

bundles/org.openhab.binding.enocean/src/main/resources/OH-INF/thing/WindowSashHandleSensor.xml
DrRSatzteil: Thanks for pointing me to the documentation. I shortened the label in my last commit.

fwolter (approved):
LGTM

PR #10951, opened 2021-07-06
[miio] change deviceID to Xiaomi used string

marcelrv:
Change the deviceId from the current hexadecimal to the string used by
Xiaomi. 

This is needed as we have some newer devicestypes coming that have deviceIds that are
non-numeric, hence breaking the current logic.

Note: separately removing the upnp discovery as this has become irrelevant with cloud discovery and devices supporting the udp regular discovery.


marcelrv:
note: jenkins builderror seems unrelated to the PR

bundles/org.openhab.binding.miio/src/main/java/org/openhab/binding/miio/internal/handler/MiIoAbstractHandler.java
```diff
@@ -351,16 +351,21 @@ protected void disconnected(@Nullable String message) {
         }
         @Nullable
         String deviceId = configuration.deviceId;
+        if (deviceId.length() == 8 && deviceId.matches("^.*[a-zA-Z]+.*$")) {
+            logger.warn(
+                    "As per openHab version 3.2 the deviceId is no longer a string with hexadecimals, instead it is a string with the numeric respresentation of the deviceId. If you continue seeing this message, update deviceId in your thing configuration");
```
fwolter, line 356:
```suggestion
                    "As per openHAB version 3.2 the deviceId is no longer a string with hexadecimals, instead it is a string with the numeric respresentation of the deviceId. If you continue seeing this message, update deviceId in your thing configuration");
```

fwolter (changes requested):

fwolter (approved):

marcelrv:
Thanks!

Issue #10954, opened 2021-07-06
[irobot] Test unstable


fwolter:
The test org.openhab.binding.irobot.internal.handler.RoombaHandlerTest.testDock is unstable:
```
org.mockito.exceptions.misusing.UnnecessaryStubbingException: 

Unnecessary stubbings detected.
Clean & maintainable test code requires zero unnecessary code.
Following stubbings are unnecessary (click to navigate to relevant line of code):
  1. -> at org.openhab.binding.irobot.internal.handler.RoombaHandlerTest.setUp(RoombaHandlerTest.java:78)
Please remove unnecessary stubbings or use 'lenient' strictness. More info: javadoc for UnnecessaryStubbingException class.
```

@rimago can you take a look?

Does the test make any sense as it contains no assertions nor `verify()`s?

rimago:
It is used to test the protocol against a local iRobot instance. 
We did not add any assertions or verify's that it also runs if there is no local iRobot instance.

So in this way it is not a real Unity test, yet it is still useful to share the code. I am open to any suggestions if there is a better way to handle this.

To fix the Exception we could simply add a lenient() in line 78. If this is fine for you I could create a pull request for it.

fwolter:
Sure, go ahead.

Issue #10753, opened 2021-05-23
[homematic] Unit tests unstable


fwolter:
See https://ci.openhab.org/job/PR-openHAB-Addons/org.openhab.addons.bundles$org.openhab.binding.homematic/4245/testReport/org.openhab.binding.homematic.internal.communicator.virtual/ButtonDatapointTest/testDoublePress/

```
java.lang.AssertionError: 

Expected: is "DOUBLE_PRESSED"
     but: was null
	at org.openhab.binding.homematic.internal.communicator.virtual.ButtonDatapointTest.testDoublePress(ButtonDatapointTest.java:106)
```

@FStolte @gerrieg @mdicke2s Can you take a look?

fwolter:
You could use waitForAssertion(() -> ...) provided from the JavaTest class instead of the sleeps, which are error prone and depend on the load of the build machine.

MHerbst:
Thanks for the hint. I will have a closed look at it next weekend.

MHerbst:
@fwolter I had a look into the implementation of waitForAssertion(), but I am not sure that it can be used to replace the sleep calls completely.

In the first part of the test, a double press is simulated and the sleep() is used to have a small interval between two short press events. If I replaced the sleep() with waitForAssert there would be no sleep before the first test. 

The second sleep() is necessary because a short press should be detected as a single press and not a double press. Therefore it needs a longer pause, and this won't work with waitForAssert().

But since the problem occurs in the first part of the test, we could try to see if the problem can be solved this way.

fwolter:
After looking into this test for a few minutes, I don't understand what happens there really. In any case there's something fishy about the sleep in the "mock". Seems to me some state is reset after some time. This could be done manually in the test as it seems the sleep in the mock is the only concurrent operation in the unit test. So, if you replace the sleep in the mock by an external manual reset, the sleep in the tests might be removed, too.

fwolter:
I'd consider waiting for assertions in unit tests a hack, tough. Often that means you also test some scheduler along with your code, which shouldn't be your intention when writing unit tests.

MHerbst:
I agree with you and I have the same problem to understand why it is implemented this way. In my opinion the whole test does not make sense.

lsiepel:
The double press event was removed, so the test is also removed.
For reference: 
https://github.com/openhab/openhab-addons/pull/11186

PR #10956, opened 2021-07-06
Disable unstable tests

fwolter:
See #10753 and #10954.


kaikreuzer:
Despite being marked as ignored, the iRobot test is still failing in this PR build. Could it be that mockito already complains even without executing the test? In that case, we probably have to put it in comments instead.

fwolter:
Interesting, that the setup method is also executed when the test is set to ignored. I removed the changes in the irobot binding as @rimago is around. https://github.com/openhab/openhab-addons/issues/10954#issuecomment-875371218

kaikreuzer (approved):
Thanks!

PR #10957, opened 2021-07-06
[miio] Add new custom refresh methods and improve response to channel matching

marcelrv:
* Improve matching the command responses to the sending channel
* Use newer method to get device list using `device_list_page`
* Adding the ability to send custom requests to cloud
* Adding the ability to send custom commands with additional elements in the json


kaikreuzer (approved):
lgtm, thanks!

PR #10959, opened 2021-07-07
[tr064] Add Wifi Signal Strength & Wifi Speed channels to LAN Subdevice

tobiloeb:
This PR adds three addional channels to the tr064 binding. 
First: The IP Address of all given MAC Address (entered in MacOnline list configuration)
Second: The Wifi Signal Strength of all given MAC Address  (entered in MacOnline list configuration)
Third: The Wifi Speed of all given MAC Address (3ntered in MacOnline list configuration)

The main goal of this additional channels is to determine the position of a device inside the house. The Wifi Signal Strength or Wifi Speed can be used to check where a device is currently located.

The WLAN Configuration of tr064 api offers multiple configurations ( 1,2,3 ) depending of the access points 2.4Ghz 5Ghz. So there are two channels for Signal Strength and Wifi Speed. One for the 2.4Ghz and one for the 5Ghz access point. See documentation here: [https://avm.de/fileadmin/user_upload/Global/Service/Schnittstellen/wlanconfigSCPD.pdf](https://avm.de/fileadmin/user_upload/Global/Service/Schnittstellen/wlanconfigSCPD.pdf)

I can't upload the .jar file with my changes. Where can I upload it? Otherwise just build the tr064 bundle and copy it to the addons folder in openhab. After restart the changes should appear.

If you have any further questions, let me know. 😃


Skinah:
Can you use system.signal-strength channel? See link below and other bindings which are now implementing this to abstract the strength.

https://www.openhab.org/docs/developer/bindings/thing-xml.html#system-state-channel-types

tobiloeb:
@Skinah  Thanks for your suggestion. If I got it right, the system.signal-strength channel describes the signal-strength of its thing. In my case the signal strength is not the signal strength of the subdevice (LANDevice) itself. The LANDevice (subdevice) includes all Lan & Wifi information and configuration from the AccessPoint (bridge). And it can give you information (e.g. signal strength and speed) for a given MAC address as well.
My MR adds the possibility to get the information about speed and strength for each MAC address added to the list in the configuration fo the LANDevice. Currently you only have the MacOnline Channel for each entered MAC address.

I think the system.signal-strength wouldn't fit in this scenario, right?

Skinah:
Yes I believe `system.signal-strength` should be used.

fwolter:
I think `system.signal-strength` can be used here. But I won't insist on changing it. @tobiloeb WDYT?

tobiloeb:
> I think `system.signal-strength` can be used here. But I won't insist on changing it. @Skinah @tobiloeb WDYT?

I have read the documentation for the system channels, but didn't get everything from the concept and how to implement it in my case. For my usecase (and hopefully for some more people) it would be enough to let it as is. 👍 :)

fwolter:
If there's a comprehension problem, don't hesitate to ask! It feels wrong to make a decision if there's some uncertainty.

tobiloeb:
> If there's a comprehension problem, don't hesitate to ask! It feels wrong to make a decision if there's some uncertainty.

Yeah I got your point. For me there are two things:
When I use the sytem.signal-strength with vaules from 0 - 4 I will lose information cause now I have a range between 0 - 100. Nevertheless it should work in my case. But do I have to implement a mapper between the different ranges and where? And is it right to just change the type from "number" to "system.signal-strength"? Or is it required to add the channels to the thing-type definition? This confuses me, cause the signal strength of my added channels are not the signal strength of the tr064 LANDevice itself.

Thanks :)

fwolter:
In most cases you don't need high resolution for signal strengths. And in most cases you don't have the accuracy, too. That's why most wifi icons have five states or less. Yes, it's hard to map the original signal strength to the 0-4 scale, but if you don't do it, you leave it up to the user. I think you have better insights to do that, than the normal user.

To change it to `system.signal-strength`, simply set the channel type to `system.signal-strength` and keep it `Number`.

Skinah:
If you need an example to follow, look at the opensprinkler binding as I made the change a week or two ago and it is merged.
I don't care strongly either way, but would lean towards changing it. Currently the wifi icon does not change, but I'm guessing that if/when that is fixed then the user gets an animated icon showing the signal strength with the standard wifi icon. Also there may be a way to click on the semantic model to show all wifi strengths and view them all on the page with a common look. I have noticed that for battery levels and battery low, there are two system channels, shame there is not two for wifi strength, but really in 99% of cases the strength is not important and having persistance storing a tiny change in db over and over is not a smart thing IMHO. Actually that is another difference, you will not have persistance storing the values when you change the channel over to system.signal-strength.

tobiloeb:
Thanks for the explanation, I will take a look at your changes and adjust the channel type. Your points sounds good to me and make sense. I will set this PR to WIP and go on when I am ready.

tobiloeb:
@fwolter Here is my update:
I have added the typeId to the macSignalStrength1 and macSignalStrength2 channels. I have decided to add the "typeId" attributes to the channel definition in channel.xml, cause all channels are configured there. It seems that this is a special case for the tr064 binding. All other bindings (as I can see) configure the channels in the thing-types.xml. For the tr064 binding there is a SOAPValueConverter and a concept of postProcessing. I used this concept to map the signalStrength values from range 0-100 to 0-4. 

I have tested my changes and its woring fine. Is it right that the type is still "Number:Dimensionless"? Openhab is not showing the signalStrength Icon automatically?

tobiloeb:
I have just testet a bit and made some code cleanings. I am using it in my openhab instance and its working well.

@fwolter, @Skinah From my perspective everythings is done. It would be great if you can take a look again. 😄 Thanks!

bundles/org.openhab.binding.tr064/README.md
```diff
@@ -119,6 +119,11 @@ The call-types are the same as provided by the FritzBox, i.e. `1` (inbound), `2`
 | `wifi5GHzEnable`           | `Switch`                  |          | Enable/Disable the 5.0 GHz WiFi device.                        |
 | `wifiGuestEnable`          | `Switch`                  |          | Enable/Disable the guest WiFi.                                 |
 | `macOnline`                | `Switch`                  |     x    | Online status of the device with the given MAC                 |
+| `macIP`                    | `String`                  |     x    | IP of the device with the given MAC                            |
+| `macSignalStrength1` | `Number:Dimensionless`    |     x    | Wifi Signal Strength of the device with the given MAC. This is set in case the Device is connected to 2.4Ghz  |
```
fwolter, line 123: As the system channel type is of type `Number`, this needs to be `Number`, too.

fwolter (changes requested):
Thanks!

bundles/org.openhab.binding.tr064/README.md
tobiloeb: Yes you are right, I missed to update the readme documentation. The type for the channels is already of type Number.

fwolter (approved):
LGTM

PR #10970, opened 2021-07-10
[sonyprojector] Switch to the default system channel type for power c…

lolodomo:
…hannel

Set semantics to the powerstate channel type


kaikreuzer (approved):
lgtm, thanks!

PR #10972, opened 2021-07-11
[openweathermap] Semantics added to channel types

lolodomo:
Also fix few categories to use only official categories

### DIFF
--- a/bundles/org.openhab.binding.openweathermap/src/main/resources/OH-INF/thing/channel-types.xml
+++ b/bundles/org.openhab.binding.openweathermap/src/main/resources/OH-INF/thing/channel-types.xml
@@ -455,6 +455,10 @@
 		<label>Apparent Temperature</label>
 		<description>Current apparent temperature.</description>
 		<category>Temperature</category>
+		<tags>
+			<tag>Measurement</tag>
+			<tag>Temperature</tag>
+		</tags>
 		<state readOnly="true" pattern="%.1f %unit%"/>
 	</channel-type>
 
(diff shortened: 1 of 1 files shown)


bundles/org.openhab.binding.openweathermap/src/main/resources/OH-INF/thing/channel-types.xml
```diff
@@ -399,6 +440,10 @@
 		<label>Forecasted Temperature</label>
 		<description>Forecasted outdoor temperature.</description>
 		<category>Temperature</category>
+		<tags>
+			<tag>Measurement</tag>
```
kaikreuzer, line 444: I am wondering here: A forecast is not really a measurement. I'd therefore rather refrain from adding this tag.
We might want to think of something like a "forecast" tag for the tag library instead. Wdyt?

lolodomo: You're right, a forecast is not exactly a measurement.
I will add a new Point "Forecast" in the semantics tag.

lolodomo: I am going to replace "Measurement" by "Status" for forecast channels.

kaikreuzer:
I'd suggest we park this PR until we finish the discussion on https://github.com/openhab/openhab-core/issues/1791. Putting "Status" on everything where we don't have dedicated semantic tags for does not feel right to me and I wouldn't want to see this to be done for 300+ bindings now...

lolodomo:
@kaikreuzer : to make progress, I propose to only keep tags that are obvious (in particular measurements) and suppress "Status" where we don't have dedicated semantic tags. Can we at least do that ?

kaikreuzer:
Sorry for dropping the ball here. 😩 

>  to make progress, I propose to only keep tags that are obvious (in particular measurements like temperature, humidity, pressure, ...) and suppress "Status" where we don't have dedicated semantic tags. Can we at least do that ?

Yes, let's do that, please!

lolodomo:
@kaikreuzer : PR is updated.

lolodomo:
As there was a long pause in the discussion, I do not remember if this is appropriate to add measurement) temperatures on min temperature or dew point temperature? 
Some of the bindings have already done.

kaikreuzer (approved):

PR #10977, opened 2021-07-11
[hue] Semantics added to channel types

lolodomo:

### DIFF
--- a/bundles/org.openhab.binding.hue/src/main/resources/OH-INF/thing/channels.xml
+++ b/bundles/org.openhab.binding.hue/src/main/resources/OH-INF/thing/channels.xml
@@ -26,10 +26,6 @@
 		<item-type>Number</item-type>
 		<label>Light Level</label>
 		<description>Current light level.</description>
-		<tags>
-			<tag>Measurement</tag>
-			<tag>Light</tag>
-		</tags>
 		<state readOnly="true" pattern="%d"/>
 	</channel-type>
 
--- a/bundles/org.openhab.binding.hue/src/main/resources/OH-INF/thing/channels.xml
+++ b/bundles/org.openhab.binding.hue/src/main/resources/OH-INF/thing/channels.xml
@@ -15,6 +15,10 @@
 		<item-type>Number:Illuminance</item-type>
 		<label>Illuminance</label>
 		<description>Current illuminance.</description>
+		<tags>
+			<tag>Measurement</tag>
+			<tag>Light</tag>
+		</tags>
 		<state readOnly="true" pattern="%.1f %unit%"/>
 	</channel-type>
 
@@ -22,6 +26,10 @@
 		<item-type>Number</item-type>
 		<label>Light Level</label>
 		<description>Current light level.</description>
+		<tags>
+			<tag>Measurement</tag>
+			<tag>Light</tag>
+		</tags>
 		<state readOnly="true" pattern="%d"/>
 	</channel-type>
 
@@ -125,6 +133,10 @@
 		<label>Temperature</label>
 		<description>Current temperature.</description>
 		<category>Temperature</category>
+		<tags>
+			<tag>Measurement</tag>
+			<tag>Temperature</tag>
+		</tags>
 		<state readOnly="true" pattern="%.1f %unit%"/>
 	</channel-type>
 
@@ -133,6 +145,10 @@
 		<item-type>String</item-type>
 		<label>Alert</label>
 		<description>The alert channel allows a temporary change to the bulb’s state.</description>
+		<tags>
+			<tag>Control</tag>
+			<tag>Alarm</tag>
+		</tags>
 		<state>
 			<options>
 				<option value="NONE">None</option>
--- a/bundles/org.openhab.binding.hue/src/main/resources/OH-INF/thing/channels.xml
+++ b/bundles/org.openhab.binding.hue/src/main/resources/OH-INF/thing/channels.xml
@@ -142,7 +142,6 @@
 		<label>Alert</label>
 		<description>The alert channel allows a temporary change to the bulb’s state.</description>
 		<tags>
-			<tag>Control</tag>
 			<tag>Alarm</tag>
 		</tags>
 		<state>


Skinah (approved):
EDIT: @lolodomo  I had a second look after playing with tags the last few days....

 'Control' and 'Light' tags give the locations tabs a number of how many lights are turned ON in that location.. etc.
See this for example and note the bulb icon with a number of how many lights are on in that location...
http://www.pcmus.com/openhab/locations.png

Some of the channels are tagged with these and since I don't own any HUE I am not certain but some I would question like the effect channel. Can the effect be turned ON when the light is actually OFF? I also don't believe it will make any difference tagging the String channels with this. Perhaps err on the side of caution and only tag the ones that control the light actually being ON and OFF, otherwise the light icon may read the wrong number of lights turned on in the locations tab.

jlaur:
@lolodomo - is this ready for being merged?

lolodomo:
> @lolodomo - is this ready for being merged?

No because we have a veto of @kubawolanin regarding this approach.

bundles/org.openhab.binding.hue/src/main/resources/OH-INF/thing/channels.xml
```diff
@@ -15,27 +15,43 @@
...
+		</tags>
 		<state readOnly="true" pattern="%.1f %unit%"/>
 	</channel-type>
 
 	<channel-type id="light_level" advanced="true">
 		<item-type>Number</item-type>
 		<label>Light Level</label>
 		<description>Current light level.</description>
+		<tags>
+			<tag>Measurement</tag>
```
kaikreuzer, line 30: I'd prefer "Measurement" tags to be only used for Numbers with dimension, so that it is clear, what exactly is measured.

bundles/org.openhab.binding.hue/src/main/resources/OH-INF/thing/channels.xml
```diff
@@ -15,27 +15,43 @@
...
 		<state readOnly="true" pattern="%d"/>
 	</channel-type>
 
 	<channel-type id="dark">
 		<item-type>Switch</item-type>
 		<label>Dark</label>
 		<description>Light level is below the darkness threshold.</description>
+		<tags>
+			<tag>Status</tag>
+			<tag>Light</tag>
```
kaikreuzer, line 42: Property tags are supposed to be used with "Control" and "Measurement" points, but not with "Status" (I know, this isn't clearly documented anywhere, sorry...).

bundles/org.openhab.binding.hue/src/main/resources/OH-INF/thing/channels.xml
```diff
@@ -15,27 +15,43 @@
...
 		<state readOnly="true"/>
 	</channel-type>
 
 	<channel-type id="daylight">
 		<item-type>Switch</item-type>
 		<label>Daylight</label>
 		<description>Light level is above the daylight threshold.</description>
+		<tags>
+			<tag>Status</tag>
+			<tag>Light</tag>
```
kaikreuzer, line 53: see above

bundles/org.openhab.binding.hue/src/main/resources/OH-INF/thing/channels.xml
```diff
@@ -133,6 +153,10 @@
 		<item-type>String</item-type>
 		<label>Alert</label>
 		<description>The alert channel allows a temporary change to the bulb’s state.</description>
+		<tags>
+			<tag>Control</tag>
+			<tag>Light</tag>
```
kaikreuzer, line 158: Actually, I think the semantic of this one should be "Alarm", not "Light". One uses it in case one want to signal an alarm and not in order to have light in the room.

bundles/org.openhab.binding.hue/src/main/resources/OH-INF/thing/channels.xml
```diff
@@ -148,6 +172,10 @@
 		<label>Color Loop</label>
 		<description>The effect channel allows putting the bulb in a color looping mode.</description>
 		<category>ColorLight</category>
+		<tags>
+			<tag>Control</tag>
+			<tag>Light</tag>
```
kaikreuzer, line 177: Sending "ON" to a "Control-Light" semantic item, I would expect to turn on some light (to make the room lit) and not to start some fancy effect.

bundles/org.openhab.binding.hue/src/main/resources/OH-INF/thing/channels.xml
```diff
@@ -168,5 +196,9 @@
 		<item-type>String</item-type>
 		<label>Scene</label>
 		<description>The scene channel allows recalling a scene to all lights that belong to the scene.</description>
+		<tags>
+			<tag>Control</tag>
+			<tag>Light</tag>
```
kaikreuzer, line 201: From a machine-readability point, I also think those tags don't have much meaning, since a machine can't make any sense from a String item state in terms of what this means for a light.

kaikreuzer (changes requested):
> No because we have a kind of veto from @kubawolanin regarding this approach.

You probably wanted to mention me instead here.
As commented on the other PRs, I would prefer to reduce this to the "obvious" fits.
As @Skinah mentioned above, the tags in this PR might have some unexpected effects to users.
I left some comments below on channels, which I consider problematic.

lolodomo:
I am going to consider review comments.

bundles/org.openhab.binding.hue/src/main/resources/OH-INF/thing/channels.xml
lolodomo: But what else to use in this case ?

bundles/org.openhab.binding.hue/src/main/resources/OH-INF/thing/channels.xml
lolodomo: Tags removed

bundles/org.openhab.binding.hue/src/main/resources/OH-INF/thing/channels.xml
lolodomo: Tags removed

bundles/org.openhab.binding.hue/src/main/resources/OH-INF/thing/channels.xml
lolodomo: Updated to use Alarm

bundles/org.openhab.binding.hue/src/main/resources/OH-INF/thing/channels.xml
lolodomo: Tags removed

bundles/org.openhab.binding.hue/src/main/resources/OH-INF/thing/channels.xml
lolodomo: Tags removed

lolodomo:
@kaikreuzer : I updated the PR, remains only one open point to discuss

bundles/org.openhab.binding.hue/src/main/resources/OH-INF/thing/channels.xml
kaikreuzer: Following our agreement to only tag what is obvious, I'd vote to not tag it at all.

lolodomo: Ok, this is now removed.

lolodomo:
@kaikreuzer : last change done

bundles/org.openhab.binding.hue/src/main/resources/OH-INF/thing/channels.xml
```diff
@@ -133,6 +141,10 @@
 		<item-type>String</item-type>
 		<label>Alert</label>
 		<description>The alert channel allows a temporary change to the bulb’s state.</description>
+		<tags>
+			<tag>Control</tag>
+			<tag>Alarm</tag>
```
jlaur, line 146: Is this combination valid?

lolodomo: You're right, I was just checking the semantics, they are both "Point".
I am going to suppress "Control".

lolodomo: Fixed

lolodomo:
@kaikreuzer : I am counting 27 bindings defining tags on some of their channels. Maybe you could check that they all define tags properly.

1. airquality
2. danfossairunit
3. deconz
4. dmx
5. feican
6. homeconnect
7. ipobserver
8. lifx
9. max
10. mihome
11. milight
12. millheat
13. mqtt.espmilighthub
14. nikohomecontrol
15. openweathermap
16. openwebnet
17. qbus
18. rotel
19. sagercaster
20. sensibo
21. shelly
22. sonyprojector
23. souliss
24. touchwand
25. vigicrues
26. wled
27. yeelight

jlaur:
@lolodomo - would you like code owner (@cweitkamp), or @kaikreuzer to review before merging? It looks good to me now after removing the tags which were not straight-forward.

jlaur (approved):
LGTM.

lolodomo:
> @lolodomo - would you like code owner (@cweitkamp), or @kaikreuzer to review before merging? It looks good to me now after removing the tags which were not straight-forward.

No, I don't think this is necessary, there is now almost nothing changed and Kai already made his comments.

PR #2457, opened 2017-07-13
[WIP] [audio] Added support for new AudioSink API

cweitkamp:
Follow-Up PR for [https://github.com/eclipse/smarthome/pull/3764](https://github.com/eclipse/smarthome/pull/3764)

* Added missing `getSupportedStreams()` methods
* Added `UnsupportedAudioStreamException` to method `process()`
* Minor changes for an unified implementation

Signed-off-by: Christoph Weitkamp <github@christophweitkamp.de>

PR #10979, opened 2021-07-12
[sonos] Add new channel codec for several models

### LINKED ISSUE #9997: [Sonos] Support for audio formats on HDMI input
At least for the soundbars with HDMI or optical input there is the possibility to read out the audio format (possibly also on the Sonos Amp).
The audio format can be found as a code in the HTAudioIn parameter:
`` 2021-01-30 12:04:51.255 [TRACE] [s.internal.handler.ZonePlayerHandler] - Received pair 'HTAudioIn':'33554434' (service 'DeviceProperties') for thing 'sonos:zoneplayer:Wohnzimmer'
``

So far, I have been able to find the following codes related to an audio format.
| audio format           | code     |
|------------------------|----------|
| Silence                | 22       |
| Silence 2.0            | 33554454 |
| Stereo PCM 2.0         | 33554434 |
| Dolby Digital 2.0      | 33554488 |
| Dolby Digital 5.1      | 84934713 |
| Dolby Digital Plus 5.1 | 84934714 |
| DTS (unsupported)      | 32       |

I am sure there are other codes as well. For example for Dolby Atmos.
I determined the codes with a Sonos Arc and a Sonos Beam. Both were connected to a TV via HDMI Arc.
This property also appears on devices without audio input. Here the value is always 0.

The current audio format can be found under "About My System" in the Sonos app. The corresponding code can be read out in the trace or under the corresponding IP address on the support page (HTAudioInCode under "Zone Player Info"). For example: http://192.168.1.218:1400/support/review

It is possible to simply output the code as an item and then translate it with a transformation.

morph166955:
Fix #9997 

Signed-off-by: Ben Rosenblum <rosenblumb@gmail.com>


bundles/org.openhab.binding.sonos/README.md
```diff
@@ -58,6 +58,7 @@ The devices support the following channels:
 | batterycharging     | Switch    | R           | Indicator set to ON when the battery is charging                                                                                                          | Move                                 |
 | batterylevel        | Number    | R           | Current battery level                                                                                                                                     | Move                                 |
 | clearqueue          | Switch    | W           | Suppress all songs from the current queue                                                                                                                 | all                                  |
+| codec		      | String    | R           | Name of codec currently being decoded                                                                                                                     | Arc, Arc SL
```
lolodomo, line 61: Please fix the alignment

lolodomo: Why only Arc and Arc SL ?

bundles/org.openhab.binding.sonos/src/main/java/org/openhab/binding/sonos/internal/handler/ZonePlayerHandler.java
```diff
@@ -1468,6 +1477,48 @@ public boolean isOutputLevelFixed() {
         return stateMap.get("MusicSurroundLevel");
     }
 
+    public @Nullable String getCodec() {
+        String codec = stateMap.get("HTAudioIn");
+        switch (codec) {
```
lolodomo, line 1482: The proper approach is to keep the numeric value as state and define state options on the channel to propose clear labels to the user. It would even allow translating in different languages.

bundles/org.openhab.binding.sonos/src/main/resources/OH-INF/thing/channels.xml
```diff
@@ -45,6 +45,13 @@
 		<description>Suppress all songs from the current queue</description>
 	</channel-type>
 
+	<channel-type id="codec" advanced="true">
+		<item-type>String</item-type>
+		<label>Codec</label>
+		<description>Name of codec currently being decoded</description>
+		<state readOnly="true"/>
```
lolodomo, line 52: List of options to be added here.
Add the tag telling that the state is not limited to the list of options (because your list of known codecs is probably partial).

lolodomo (changes requested):
Two main requests:
1 - use state options
2 - why adding this channelm only on Arc model ?

morph166955:
@lolodomo I'll go through these tonight if I can. With respect to it being limited to Arc, my understanding was that this was only for devices with HDMI inputs. Is that not the case?

lolodomo:
> With respect to it being limited to Arc, my understanding was that this was only for devices with HDMI inputs. Is that not the case?

This is very probably available to all devices having a digital input. If I am not wrong, it means Playbar, Playbase, Beam, Port and Amp in addition to Arc.

bundles/org.openhab.binding.sonos/src/main/resources/OH-INF/thing/Arc.xml
```diff
@@ -15,6 +15,7 @@
 			<channel id="alarmproperties" typeId="alarmproperties"/>
 			<channel id="alarmrunning" typeId="alarmrunning"/>
 			<channel id="bass" typeId="bass"/>
+			<channel id="codec" typeId="codec"/>
```
lolodomo, line 18: Please move the channel with the model specific channels.

bundles/org.openhab.binding.sonos/src/main/resources/OH-INF/thing/ArcSL.xml
```diff
@@ -15,6 +15,7 @@
 			<channel id="alarmproperties" typeId="alarmproperties"/>
 			<channel id="alarmrunning" typeId="alarmrunning"/>
 			<channel id="bass" typeId="bass"/>
+			<channel id="codec" typeId="codec"/>
```
lolodomo, line 18: Please move the channel with the model specific channels.

morph166955:
> > With respect to it being limited to Arc, my understanding was that this was only for devices with HDMI inputs. Is that not the case?
> 
> This is very probably available in all devices having a digital input. If I am not wrong, it means Playbar, Playbase, Beam and Amp in addition to Arc.
> 
> Edit Sonos Port has no digital input so is not concerned.

Works for me. I'll go add those to the list. I'm intrigued about the SPDIF input. I always assumed that it received effectively a PCM 5.1 feed from the source. I didn't realize that it actually did anything codec specific like the Arc does with Atmos.

lolodomo:
Those without HDMI input will support only a subset of codec like Dolby Digital 5.1 for example but of course not Dolby Atmos. But you don't have to care about that.

bundles/org.openhab.binding.sonos/src/main/java/org/openhab/binding/sonos/internal/handler/ZonePlayerHandler.java
morph166955: Correct me if I'm wrong, but state options only apply to UIs correct (I don't use them for anything currently so I'm not very familiar beyond what I've read on the forum)?  Things like rules would still have to parse the numeric values?  Here's my use case and concern.  As part of my testing (and intention for long term use), I have the state of this channel displayed on my Neeo remote.  I like to see that I'm actually decoding in Atmos for those videos (because I have issues sometimes with it not doing what it should).  If I understand this correctly, I would see the numeric value on the Neeo and not the name.  It could also get confusing for people who would use this in a rule.  They should be able to specify "Dolby Atmos" instead of "63".  

Alternative, instead of setting the output to an English output, we could do something like "dolbyAtmos" or "dolbyPCM51" that could be then converted.

lolodomo: > Alternative, instead of setting the output to an English output, we could do something like "dolbyAtmos" or "dolbyPCM51" that could be then converted.

Yes, you could keep a conversion to values like NoSignal, DolbyAtmos, ... and then have the state options for better display in UI (and with ability of translations). In documentation, you will have to mention these values (DolbyAtmos, ...).

morph166955:
I believe I pushed everything from above.  I can't find what's wrong with the alignment of the README file, it looks clean on my screen.

lolodomo:
>  I can't find what's wrong with the alignment of the README file, it looks clean on my screen.

image

bundles/org.openhab.binding.sonos/src/main/java/org/openhab/binding/sonos/internal/handler/ZonePlayerHandler.java
```diff
@@ -1468,6 +1477,51 @@ public boolean isOutputLevelFixed() {
...
+            case "0":
+                codec = "noSignal";
+                break;
+            case "21":
+                codec = "noSignal";
+                break;
```
lolodomo, line 1488: I rather suggest:
```
            case "0":
            case "21":
                codec = "noSignal";
                break;
```

bundles/org.openhab.binding.sonos/src/main/java/org/openhab/binding/sonos/internal/handler/ZonePlayerHandler.java
```diff
@@ -1468,6 +1477,51 @@ public boolean isOutputLevelFixed() {
...
+            case "32":
+                codec = "DTS";
+                break;
+            case "59":
+                codec = "dolbyAtmos";
+                break;
+            case "63":
+                codec = "dolbyAtmos";
+                break;
```
lolodomo, line 1500: Same here

bundles/org.openhab.binding.sonos/src/main/java/org/openhab/binding/sonos/internal/handler/ZonePlayerHandler.java
```diff
@@ -1468,6 +1477,51 @@ public boolean isOutputLevelFixed() {
...
+            case "33554454":
+                codec = "silence20";
+                break;
```
lolodomo, line 1503: Why not just using `silence` ? We don't really care about the number of channels in this case.

bundles/org.openhab.binding.sonos/src/main/resources/OH-INF/thing/ArcSL.xml
lolodomo: You forgot to remove this line.

bundles/org.openhab.binding.sonos/src/main/resources/OH-INF/thing/channels.xml
```diff
@@ -45,6 +45,26 @@
...
 
+	<channel-type id="codec" advanced="true">
+		<item-type>String</item-type>
+		<label>Codec</label>
+		<description>Name of codec currently being decoded</description>
+		<state readOnly="true">
+			<options>
+				<option value="noSignal">No Signal</option>
+				<option value="silence">Silence</option>
+				<option value="silence20">Silence 2.0</option>
```
lolodomo, line 56: Just one entry

bundles/org.openhab.binding.sonos/src/main/resources/OH-INF/thing/channels.xml
```diff
@@ -45,6 +45,26 @@
 		<description>Suppress all songs from the current queue</description>
 	</channel-type>
 
+	<channel-type id="codec" advanced="true">
+		<item-type>String</item-type>
+		<label>Codec</label>
+		<description>Name of codec currently being decoded</description>
+		<state readOnly="true">
+			<options>
```
lolodomo, line 53: Please add the tag notifying that the state is not limited to the provided list of options. Sorry, I don't remember its exact name but probably something like `limitedToOptions` (to be checked)

lolodomo (changes requested):
Just few last remarks

morph166955:
> > I can't find what's wrong with the alignment of the README file, it looks clean on my screen.
> 
> image

Ok that's weird.  I 1000% see it on github as off, but my editor shows it as perfectly lined up.

readme

morph166955:
The editor on github also shows the alignment as correct.  I'm not sure why this is looking weird when viewed...

readme2

morph166955:
I believe the last commit covers all comments minus whatever is going on with README.

morph166955:
Fixed the readme file.  No idea what was causing that.  I deleted the line on the github editor and just put it back in.  Now it looks clean.

lolodomo (approved):
LGTM now

morph166955:
Thank you!

bundles/org.openhab.binding.sonos/src/main/java/org/openhab/binding/sonos/internal/handler/ZonePlayerHandler.java
```diff
@@ -1468,6 +1477,45 @@ public boolean isOutputLevelFixed() {
         return stateMap.get("MusicSurroundLevel");
     }
 
+    public @Nullable String getCodec() {
+        String codec = stateMap.get("HTAudioIn");
+        switch (codec) {
```
lolodomo, line 1482: I think you should first check that the value is not null, just before the switch.

lolodomo (commented):
I think a null check is missing.

bundles/org.openhab.binding.sonos/src/main/java/org/openhab/binding/sonos/internal/handler/ZonePlayerHandler.java
morph166955: Done.

lolodomo (approved):
Ok

kaikreuzer (approved):
Thanks!

morph166955:
@lolodomo - Not entirely sure why, but this isn't working in the recent snapshot #2457 which it was compiled in.  Just getting null.  I'm getting a few channels (volume and mute seem to be ok), but others seem to be dead in the air too.

lolodomo:
You introduced an error in the channels.xml file, with your tag `limitToOptions`. Either a bad name or at the wrong place.
I will fix that !
You did not see the validation error in your editor ?

morph166955:
My editor didn't balk at all, and it compiled cleanly as well.  When compared to others in the repository it looks the same to me.  Not sure what I missed and where.  

https://github.com/openhab/openhab-core/search?q=limitToOptions
https://www.openhab.org/docs/developer/bindings/config-xml.html

lolodomo:
`limitToOptions` is only for configuration parameter, not for state description.
That is a little my fault, I should not have requested to add this tag. On your side, you could have tested that the binding was still working.

morph166955:
I just removed it from the xml and recompiled and you are correct, that was the offending issue.  I am a little confused as I did compile and test the last version of the code and it seemed to work.  Ran it all night and it was clean.  Not sure if I grabbed an old jar or something.  My apologies there.

lolodomo:
If you use a "correct" editor, it should check your XML file against the XSD and in this case it would have shown the invalid syntax due to unexpected tag.

PR #10983, opened 2021-07-12
[amplipi] Initial contribution of AmpliPi binding

kaikreuzer:
This binding supports the AmpliPi home audio system - see https://www.micro-nova.com/amplipi.


sumnerboy12:
Nice, following with interest! Once I get mine I might be able to help out also.

kaikreuzer:
FTR, some discussion on the right modelling is taking place at https://github.com/micro-nova/AmpliPi/issues/54.

kaikreuzer:
@openhab/add-ons-maintainers This is now ready for review. The implementation contains the main features that make sense to have. Future PRs will add further functionality.

bundles/org.openhab.binding.amplipi/README.md
```diff
@@ -0,0 +1,91 @@
...
+Both the `zone` and `group` Things only require a single configuration parameter `id`, which corresponds to their id on the AmpliPi.
+
+## Channels
+
+These are the channels of the `controller` Bridge:
+
+| Channel  | Type   | Description                                                                                          |
+|----------|--------|------------------------------------------------------------------------------------------------------|
+| preset   | Number | Allows setting a pre-configured preset. The available options are dynamically read from the AmpliPi. |
+| input1   | String | The selected input of source 1                                                                       |
```
sumnerboy12, line 35: Is the type correct for `input1-4`? Should be `Number` based on your example below?

bundles/org.openhab.binding.amplipi/README.md
```diff
@@ -0,0 +1,91 @@
...
+Bridge amplipi:controller:1 "My AmpliPi" [ hostname="amplipi.local" ] {
+    zone zone2 "Living Room" [ id=1 ]
+}
+```
+
+amplipi.items:
+
+```
+Number      Preset      "Preset"                { channel="amplipi:controller:1:preset" }
+Number      Input1      "Input 1"               { channel="amplipi:controller:1:input1" }
```
sumnerboy12, line 63: Are these input item types correct? Should be strings based on the channel definitions above?

bundles/org.openhab.binding.amplipi/src/main/java/org/openhab/binding/amplipi/internal/AmpliPiGroupHandler.java
```diff
@@ -0,0 +1,148 @@
...
+    private final HttpClient httpClient;
+    private final Gson gson;
+
+    private @Nullable AmpliPiHandler bridgeHandler;
+
+    public AmpliPiGroupHandler(Thing thing, HttpClient httpClient) {
+        super(thing);
+        this.httpClient = httpClient;
+        this.gson = new Gson();
+        id = Integer.valueOf(thing.getConfiguration().get(AmpliPiBindingConstants.CFG_PARAM_ID).toString());
```
fwolter, line 67: Shouldn't this be set in initialize() to be updated on configuration updates?

bundles/org.openhab.binding.amplipi/src/main/java/org/openhab/binding/amplipi/internal/AmpliPiGroupHandler.java
```diff
@@ -0,0 +1,148 @@
...
+                    logger.error("AmpliPi API returned HTTP status {}.", response.getStatus());
+                    logger.debug("Content: {}", response.getContentAsString());
+                } else {
+                    updateStatus(ThingStatus.ONLINE);
+                }
+            } catch (InterruptedException | TimeoutException e) {
+                updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.COMMUNICATION_ERROR,
+                        "AmpliPi request failed: " + e.getMessage());
+            } catch (ExecutionException e) {
+                logger.error("HTTP request to AmpliPi failed: {}.", e.getMessage());
```
fwolter, line 130: It doesn't seem resonable to log to error when the network fails.

bundles/org.openhab.binding.amplipi/src/main/java/org/openhab/binding/amplipi/internal/AmpliPiHandler.java
```diff
@@ -0,0 +1,215 @@
...
+            } catch (Exception e) {
+                logger.error("Unexpected error occurred: {}", e.getMessage());
+            }
+        }, 0, config.refreshInterval, TimeUnit.SECONDS);
+    }
+
+    private void setProperties(Status currentStatus) {
+        String version = currentStatus.getInfo().getVersion();
+        Map<String, String> props = editProperties();
+        props.put("firmwareVersion", version);
```
fwolter, line 172:
```suggestion
        props.put(Thing.PROPERTY_FIRMWARE_VERSION, version);
```

bundles/org.openhab.binding.amplipi/src/main/java/org/openhab/binding/amplipi/internal/AmpliPiZoneHandler.java
```diff
@@ -0,0 +1,148 @@
...
+    private final HttpClient httpClient;
+    private final Gson gson;
+
+    private @Nullable AmpliPiHandler bridgeHandler;
+
+    public AmpliPiZoneHandler(Thing thing, HttpClient httpClient) {
+        super(thing);
+        this.httpClient = httpClient;
+        this.gson = new Gson();
+        id = Integer.valueOf(thing.getConfiguration().get(AmpliPiBindingConstants.CFG_PARAM_ID).toString());
```
fwolter, line 67: See above. Move to initialize(). Please check all.

fwolter (changes requested):
Did you add the generated files by intention?

There are some checkstyle warnings left.

bundles/org.openhab.binding.amplipi/README.md
kaikreuzer: Indeed, those need to be Strings. Fixed it, thanks!

bundles/org.openhab.binding.amplipi/src/main/java/org/openhab/binding/amplipi/internal/AmpliPiGroupHandler.java
kaikreuzer: Good point. I have now removed the field completely and use a method instead. This way, we can be sure that it's always the latest value that is used.

bundles/org.openhab.binding.amplipi/src/main/java/org/openhab/binding/amplipi/internal/AmpliPiGroupHandler.java
kaikreuzer: If there's a network issue, we run into the `TimeoutException` which is caught above.
I am not exactly sure in which situations an `ExecutionException` exactly occurs - that's why I rather considered it a bug situation as long as it isn't better handled. Do you happen to know when we can expect the `ExecutionException` from the httpClient?

bundles/org.openhab.binding.amplipi/src/main/java/org/openhab/binding/amplipi/internal/AmpliPiHandler.java
kaikreuzer: Fixed.

kaikreuzer:
> Did you add the generated files by intention?

Yes, absolutely. The code has actually be manually adapted after generation and the generator is configured to not overwrite existing files. I thought I leave the maven plugin in the pom, so that it is easier in future to generate additional model classes when the Swagger file is extended by new features.

bundles/org.openhab.binding.amplipi/src/main/java/org/openhab/binding/amplipi/internal/AmpliPiGroupHandler.java
fwolter: Interesting question. I always assumed that a `ExecutionException` can be caused by a network failure. I did a quick search and couldn't find any official documentation, but I found some stack traces showing this (in an OH context):
```
Caused by: java.util.concurrent.ExecutionException: java.net.ConnectException: Connection refused
at org.eclipse.jetty.client.util.FutureResponseListener.getResult(FutureResponseListener.java:118)
...
```
I would handle this as a network failure.

kaikreuzer: I am not sure. A "Connection refused" means that the network is available, but that the contacted device did not accept the request at all - meaning that the request was invalid (e.g. by using a wrong port), so it might be due to a bug in the code.
But it's ok for me to treat it like a network issue - users will complain, if their things go OFFLINE for an ununderstood reason anyhow. I've changed the code accordingly.

fwolter: Or the host is available, but the service is down. I think you can't clearly differentiate between temprorary or fatal errors here.

fwolter (approved):
LGTM

fwolter:
The build leaves an untracked file: `bundles/org.openhab.binding.amplipi/.openapi-generator/amplipi-api.yml.sha256` Should we add it to gitignore?

kaikreuzer:
Yes. I have created https://github.com/openhab/openhab-addons/pull/11081 for it.

PR #10988, opened 2021-07-12
[epsonprojector] Add configurable volume channel range

mlobstein:
Previously the volume channel only supported projectors with a volume range of 0-20. At least one user in the community forum reported that their projector used a volume range of 0-40 and found that the binding would not work to control their volume. This PR adds the ability to configure the volume range for the user's projector (currently 0-20 and 0-40). The volume channel was changed to a dimmer type and the configured volume range of the projector will be mapped across the dimmer's 0-100% range.


fwolter (approved):
LGTM

PR #10989, opened 2021-07-13
[ecotouch] Fix decimal values being truncated

### LINKED ISSUE #10964: [ecotouch] Decimal digits are ignored when setting values
## Expected Behavior
I want to change the value of e.g. "nviSollKuehlen" with a stepper widget in the Main UI with steps of "0.1".
When I change the temperature e.g. from "20.0" to "20.1", the value in the Waterkotte heatpump should be changed accordingly.

## Current Behavior
Instead, the value is just set to "20.0" with my decimal digit ignored. I.e. the value in the heatpump does not change and the value in the stepper is set to "20.0" again after the next refresh cycle, when the real value is fetched from the heatpump.

I checked (in Firefox Developer Tools) the payload of the POST request to (in my case) `/rest/items/WaterkotteEcoTouchGeo_nviSollKuehlen` after increasing the stepper, it is correctly set to "20.1".

But on my Raspberry PI, where openHAB is installed and the heatpump is connected to, I monitored the traffic and I see this line in the tcpdump:

`HTTP: GET /cgi/writeTags?returnValue=true&n=1&t1=A109&v1=200 HTTP/1.1`

The raw value of the parameter "v1" is "200", which corresponds to the real value "20.0" of my stepper.
The raw value in the URL parameter should be "201" (corresponding to "20.1").

So, somewhere between my frontend with the correct POST payload and the wrong URL parameter sent to my heatpump, the decimal digit is lost.
Maybe in some transformation / calculation, like this?
https://github.com/openhab/openhab-addons/blob/ab816b08c6914fb01464a2ed66c24d60efd7b7de/bundles/org.openhab.binding.ecotouch/src/main/java/org/openhab/binding/ecotouch/internal/EcoTouchHandler.java#L134

I first thought of some reason around regional setting, i.e. decimal point vs. comma (in English, it is "20.1", but in German it is "20,1"), but my system is all english/US and the whole POST route is probably a core element in openhab and not specific to the Ecotouch binding, so I don't think that there's a bug here, but I don't know.

End-to-end only possible with a Waterkotte heatpump.
But maybe there is a test scenario for the data flow from the UI POST request to the Ecotouch binding, where the URL is constructed?

Maybe first starting point is to check if the command is properly handled, i.e. if the correct payload "20.1" is transformed properly to the raw value "201", in the handleCommand method https://github.com/openhab/openhab-addons/blob/ab816b08c6914fb01464a2ed66c24d60efd7b7de/bundles/org.openhab.binding.ecotouch/src/main/java/org/openhab/binding/ecotouch/internal/EcoTouchHandler.java#L89

## Your Environment

* Version used: openHAB 3.1.0 and ecotouch 3.1.0
* Tested in Desktop with Firefox 89.0.2 and with the Android App 2.18.0
* openHAB is running on the latest openHABian 1.6.5

@sibbi77, since you're the author, I'll add a mention for you. Maybe you can help?

sibbi77:
fixes #10964

### DIFF
--- a/bundles/org.openhab.binding.ecotouch/src/main/java/org/openhab/binding/ecotouch/internal/EcoTouchHandler.java
+++ b/bundles/org.openhab.binding.ecotouch/src/main/java/org/openhab/binding/ecotouch/internal/EcoTouchHandler.java
@@ -119,8 +119,7 @@ public class EcoTouchHandler extends BaseThingHandler {
                             QuantityType<?> value = (QuantityType<?>) command;
                             QuantityType<?> rawUnit = value.toUnit(ecoTouchTag.getUnit());
                             if (rawUnit != null) {
-                                int raw = rawUnit.intValue();
-                                raw *= ecoTouchTag.getDivisor();
+                                int raw = (int) (rawUnit.doubleValue() * ecoTouchTag.getDivisor());
                                 localConnector.setValue(ecoTouchTag.getTagName(), raw);
                             }
                         } else {


kaikreuzer (approved):
lgtm, thanks!
Merging as small patch exception without sign-off.

PR #10990, opened 2021-07-13
[homeconnect] Fix update of channels linked to program options

### LINKED ISSUE #10955: [homeconnect] Channels keeping old value when changing program
Most of washer programs are defined with a washing temperature and a spin speed.
But some of them are not using a washing temperature, for example the program used to only rinse your laundry.
When you switch to such a program, the channel for the temperature remains with its old value (from the previous program) rather than being set to UNDEF.

This is only an example but many channels are in fact concerned because not all channels are used by all programs.

The problem is certainly in this method:
https://github.com/openhab/openhab-addons/blob/main/bundles/org.openhab.binding.homeconnect/src/main/java/org/openhab/binding/homeconnect/internal/handler/AbstractHomeConnectThingHandler.java#L1331
In this method, if no option value is provided, we should set the corresponding channel to UNDEF rather than doing nothing.

lolodomo:
Fix #10955

### DIFF
--- a/bundles/org.openhab.binding.homeconnect/src/main/java/org/openhab/binding/homeconnect/internal/handler/AbstractHomeConnectThingHandler.java
+++ b/bundles/org.openhab.binding.homeconnect/src/main/java/org/openhab/binding/homeconnect/internal/handler/AbstractHomeConnectThingHandler.java
@@ -75,6 +75,7 @@ import org.openhab.core.thing.binding.BaseThingHandler;
 import org.openhab.core.thing.binding.BridgeHandler;
 import org.openhab.core.types.Command;
 import org.openhab.core.types.RefreshType;
+import org.openhab.core.types.State;
 import org.openhab.core.types.StateOption;
 import org.openhab.core.types.UnDefType;
 import org.slf4j.Logger;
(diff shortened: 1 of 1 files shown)


lolodomo:
I tested it seriously but not when a washing program is running. That is why I marked as "additional testing preferred". I will do that the next time I will run a program.

bundles/org.openhab.binding.homeconnect/src/main/java/org/openhab/binding/homeconnect/internal/handler/AbstractHomeConnectThingHandler.java
```diff
@@ -469,6 +470,21 @@ protected boolean isThingOffline() {
         }
     }
 
+    /**
+     * Get thing linked channel by given channel id.
+     *
+     * @param channelId channel id
+     * @return channel if linked
+     */
+    protected Optional<Channel> getThingLinkedChannel(String channelId) {
```
bruestel, line 479: I would prefer `getLinkedChannel` but up to you.

lolodomo: I agree with you.

lolodomo:
@bruestel : for your information, once merged, I will submit another PR replacing calls to getChannel by getLinkedChannel when it is appropriate.

bruestel:
> I tested it seriously but not when a washing program is running. That is why I marked as "additional testing preferred". I will do that the next time I will run a program.

Where you able to test it?

lolodomo:
> Where you able to test it?

No, I have not yet done the additional tests I would like to do.

lolodomo:
Doing more tests, I discovered that the option "vario perfect" Is disappearing from the API response when the program is running, even if it was set and available for this program. I should think about what to do.

This is ok for all other channels I am able to test.

lolodomo:
@bruestel : that is now ready for a review.

bruestel (approved):

Issue #8388, opened 2020-09-03
[unifi] Add support for Dream Machine Pro


regnets:
<a alt="bounty" href="https://www.bountysource.com/issues/92746043-unifi-add-support-for-dream-machine-pro"><img src="https://api.bountysource.com/badge/issue?issue_id=92746043"></a>

As far as i know currently the unifi Binding doesn't support the dream machine pro.

I got an 404 if i configure the unifi binding on my openhab installation. @ericsante on the openhab community forum experienced the same behaviour (https://community.openhab.org/t/ubiquiti-unifi-binding-feature-discussion/14520/1072?u=regnets).

regnets:
I created a small bounty for this feature request:
https://www.bountysource.com/issues/92746043-unifi-add-support-for-dream-machine-pro

cweitkamp:
@regnets Thanks for posting a bounty on this issue :+1:.

martinvw:
@mgbowman FYI

martinvw:
@regnets could you elaborate on the issue, could you show/share the logs such that it might be easier to fix/reproduce without access to a UDM Pro

HoLiuS85:
@martinvw I do have the same Issue as described in this thread. I guess part of the problem is that interface are published via different ports that previously (now port 80, before 8443)...if that is all, I do not know. I tried configuring port 80, but it did not work.

I set the log level for the Unifi binding to `log:set TRACE org.openhab.binding.unifi` and restarted it, the following are the only entries I see:

`2020-09-09 10:31:17.918 [DEBUG] [.handler.UniFiControllerThingHandler] - Initializing the UniFi Controller Handler with config = UniFiControllerConfig{host = 192.168.1.1, port = 80, username = openhab, password = *****, refresh = 10}`
`2020-09-09 10:31:17.919 [TRACE] [nal.api.model.UniFiControllerRequest] - >> POST https://192.168.1.1:80/api/login`

If I can provide any further infromation to help fix this issue, let me know.

martinvw:
That seems like the main problem is that port 80 is most likely not https, does it not have an https available? Maybe on port 443, which is the default. Could you validate this?

So the fact that the unifi binding expects https could be the only problem.

martinvw:
@HoLiuS85 I replaced the hardcoded https by HTTP (still hardcoded) as a test, could you try:

https://download.martinvw.nl/org.openhab.binding.unifi-2.5.9-SNAPSHOT.jar

[...]

Issue #10935, opened 2021-07-04
[unifi] Thing UNINITIALIZED (HANDLER_CONFIGURATION_PENDING) after upgrade to 3.1


jlaur:
## Expected Behavior
When thing is configured correctly, it should switch to state ONLINE.

## Current Behavior
In these scenarios the thing will not work:
- After upgrading to openHAB 3.1 (without changing configuration)
- When adding a new UniFi thing from UI without changing the "UniFi OS" option.
- When adding a new UniFi thing from text file without explicitly setting: unifios="false".

## Possible Solution
When option "UniFi OS" is not set (in UI or text file) it should default to "false".

Current work-arounds:
- Text: Add unifios="false"
- UI: Enable "UniFi OS" and save. Then disable and save.

## Steps to Reproduce (for Bugs)
Create new thing without enabling/touching "UniFi OS".

## Environment
* Version used: openHAB 3.1 with bundled binding.
* Distribution: openHABian.

PR #11004, opened 2021-07-16
[unifi] Fixed 404 error with reconnect and no default value set for UniFiOS Checkbox

WatcherWhale:
This should fix: #8388 and #10935

This PR adds the following:
- Fixes the 404 error
- Sets a default value for the UniFi OS checkbox (false)
- Adds a bit of documentation


fwolter (approved):
LGTM

@mgbowman Please raise your hand if you think this shouldn't have been merged for some reason.

PR #11023, opened 2021-07-18
[jdbc] Select the proper schema when retrieving the item tables

nimric:
Signed-off-by: Riccardo Nimser-Joseph <github@nimric.de>

[jdbc] Select the proper schema when retrieving the item tables from PostgreSQL

JDBC persistence does not specify a (configurable) schema when the tables are created. Therefore, a schema other than 'public' may be used if the 'search_path' is configured in PostgreSQL. In this case the correct schema must be determined when the item tables are retrieved.


kaikreuzer (approved):
Thank you!

PR #11028, opened 2021-07-19
[haywardomnilogic] Remove erroneous chlorinator code from filterhandler

matchews:
Remove erroneous chlorinator code from filterhandler


kaikreuzer (approved):
Thanks!

PR #11034, opened 2021-07-20
[pulseaudio] Remove useless config file

### LINKED ISSUE #11020: [pulseaudio] Unused config file
@peuter : I think the file config.xml is useless. It defines a config with URI "binding:pulseaudio" which is then never used.
Can you please confirm that I am not missing something ?

lolodomo:
Fix #11020


kaikreuzer (changes requested):
It seems this config file is actively used, see https://github.com/openhab/openhab-addons/blob/main/bundles/org.openhab.binding.pulseaudio/src/main/java/org/openhab/binding/pulseaudio/internal/PulseaudioHandlerFactory.java#L135-L141.

lolodomo:
> It seems this config file is actively used

Oups, you're right. Not in a common way. I close my PR. We could just rename the config URI.

kaikreuzer:
> Not in a common way. 

I actually think that this is the common way for adding a "binding configuration" description, isn't it?

Nonetheless it seems that the Main UI actually does not support binding configurations at all (the Paper UI did).
We might have to create a bug report for it?

lolodomo:
> I actually think that this is the common way for adding a "binding configuration" description, isn't it?

Yes for the way to declare it, of course.

> Nonetheless it seems that the Main UI actually does not support binding configurations at all (the Paper UI did).
> We might have to create a bug report for it?

In fact, MainUI allows to edit the binding configuration but this is very "hidden". You have to select your binding then expand and you see a link "Configure".
This was already discussed with @ghys.

lolodomo:
> We could just rename the config URI.

In fact, we should just remove the URI. This is the same case as other bindings like the Sonos binding for example.

lolodomo:
The difference with other bindings is that the configuration is not defined in the binding.xml file but in a config file. That is the reason why there is an URI defined.

My suggestion would be to move the content of the config.xml file in the binding.xml file (and remove the URI).

kaikreuzer:
You're right, for Sonos I see the entries, but for Pulseaudio, I don't:

Screenshot 2021-07-25 at 10 36 45

So indeed, the Main UI correctly supports it already as can be seen with the Sonos binding.

> In fact, we should just remove the URI.

That would mean to move the config description to the binding.xml file as it is the case for the Sonos binding, right? Yes, that would probably be the cleanest way.

kaikreuzer:
> My suggestion would be to move the content of the config.xml file in the binding.xml file (and remove the URI).

Go for it, thanks!

lolodomo:
New PR submitted.

PR #11043, opened 2021-07-21
[airthings] Enhance logging

dw-8:
Occasionally, the Airthings thing will stop updating without any warnings, errors, or explanations. Often, these problems are due to connect, disconnect, read, and resolving issues that may occur between the Airthings device and the bluetooth controller, requiring some form of 'reset'.

This PR makes logging enhancements so that the log files can be monitored for INFO, WARN, and ERROR messages.

The new error messages are all prefixed with "ERROR: Controller reset needed." so that a log monitoring script can be made to reset the bluetooth controller (eg. hciconfig hci0 reset in linux) and/or disable and re-enable the bluetooth controller thing and airthings thing within OpenHab.

Four variables are used as counters to keep track of the number of unsuccessful attempts.

- **errorConnectCounter** - _keeps track of Connect failures_
- **errorReadCounter** - _keeps track of Read failures_
- **errorDisconnectCounter** - _keeps track of Disconnect failures_
- **errorResolvingCounter** - _keeps track of stuck in Resolving state failures_

Each respective error counter will increment by one for each unsuccessful attempt, and will reset back to zero after a successful attempt. The first five unsuccessful attempts are logged as soft warnings, while each subsequent unsuccessful attempts after five are logged as hard errors.

A hard error will also change the thing status to OFFLINE, so that users can be notified of issues.

**INFO level messages**
Initialize Device
Dispose Device {}
Parsed data []

**WARN level messages**
Connecting to device {} failed {} times
Disconnect from device {} failed
Read data from device {} failed {} times
Unhandled serviceState {} on device {}

**ERROR level messages**
ERROR: Controller reset needed. Connecting to device {} failed {} times
ERROR: Controller reset needed. Disconnect from device {} failed {} times
ERROR: Controller reset needed. Read data from device {} failed {} times
ERROR: Controller reset needed. Unhandled serviceState {} on device {}

Signed-off-by: Davy Wong <davy.wong.on+github@gmail.com>


kaikreuzer (changes requested):
Thanks for looking into this!
I also noticed that my Airthings often loose their connection and I haven't really figured out yet, how to resolve this. Does a controller reset always help for you? We might actually want to reach out to Airthings support and see if there's any issue with their firmware that could be fixed on their end...

Your suggested code changes unfortunately go against our logging guidelines, which essentially say that everything should be logged on debug level, unless there is an unexpected situation that users should report as a bug, so that it gets fixed in the code.

I'd hence suggest to change all logging to debug and keep the status update to OFFLINE.

dw-8:
> 
> 
> Thanks for looking into this!
> I also noticed that my Airthings often loose their connection and I haven't really figured out yet, how to resolve this. Does a controller reset always help for you? We might actually want to reach out to Airthings support and see if there's any issue with their firmware that could be fixed on their end...
> 
> Your suggested code changes unfortunately go against our logging guidelines, which essentially say that everything should be logged on debug level, unless there is an unexpected situation that users should report as a bug, so that it gets fixed in the code.
> 
> I'd hence suggest to change all logging to debug and keep the status update to OFFLINE.

I made the edits to changes as suggested, but somehow have everyone else's commits merged into my PR now.  
I am new to using github, so I don't know how to fix this now....help?

dw-8:
I have now reverted the 40 extra commits, so that only one file has any changes, but all the other reviewers of those 40 commits are still in this PR, which are not really required.  Is there a way to remove all those extra reviewers?

kaikreuzer (approved):
lgtm, thanks.
Yeah, it seems you need to get some experience with git. Try using "rebase" and "squash" a lot and NEVER "merge". Always create your branch from the latest "main".

dw-8:
> I also noticed that my Airthings often loose their connection and I haven't really figured out yet, how to resolve this. Does a controller reset always help for you? We might actually want to reach out to Airthings support and see if there's any issue with their firmware that could be fixed on their end...

I have been accessing my Airthings Wave Gen 1 from OH for 11 months now, since September 2020.  First, I was using a modified version of Pali's OH2.5 code to just support the Wave Gen1 instead of the Wave Plus.  When you extended Pali's code to include the Wave Mini, in April 2021, it enabled me to add the Wave Gen 1 code, but I couldn't get it to compile in Eclipse anymore because of some missing dependencies that cropped up in mid-April with OH3.0.  Once OH3.1 came out at end of June 2021, I was able to finalize my code and send in this PR.

Since July 1, the following errors appeared in my _org.openhab.binding.bluetooth.airthings_ logs.
- "ERROR:  Controller reset needed.  Read data from device" --> occurred **1x** in last 24 days
- "ERROR:  Controller reset needed.  Connecting to device" --> occurred **10x** in last 24 days

I also have the following error from the _org.openhab.binding.bluetooth.bluez_ logs.  
- "Characteristic .* is missing" --> occurred **13x** in last 24 days, where the * is a wildcard in my grep regex.

For the "ERROR:  Controller reset needed.  Connecting to device", my bash script just issues a "hciconfg hci0 reset".

For both the "ERROR:  Controller reset needed.  Read data from device" and "Characteristic .* is missing", my bash script does the following:
1. disables the Airthings thing
2. disables the BlueZ thing
3. sleep 5
4. "hciconfig hci0 reset"
5. sleet 5
6. re-enable the BlueZ thing
7. re-enable the Airthings thing

My Airthings Wave Gen 1 has been publishing its updates into OH reliably for a while now, using this bash script to automatically reset itself as needed.  At the beginning, I was only issuing the "hciconfig hci0 reset" for everything and then added the disable/re-enable things to the two above errors, as the "hciconfig hci0 reset" wasn't sufficient.  I recall trying just the disable/enable things without an "hciconfig hci0 reset" at one point, and it didn't work consistently.

The connection, read, and missing characteristic errors appear to occur randomly without any identifiable pattern, so I don't think it is a bug in Airthings firmware.  My guess is it is a corrupted bluetooth packet being received and is not being handled correctly/gracefully within either the bluez driver and/or bluez binding, which causes the Airthings to go offline.

Once I get the "Characteristic .* is missing" error from org.openhab.binding.bluetooth.bluez.internal.BlueZBluetoothDevice, it will continue to occur every time it connects to the Airthings, until I disable the Airthings and BlueZ things, reset the bluetooth controller, and then re-enable the Airthings and BlueZ thing.

The read and connect errors may occur a few times in a row and self-correct itself, which is why I used a soft-warning for the first 5 consecutive occurrences; but, if it doesn't self-correct within 5 tries, it stays stuck until I using my bash script.

kaikreuzer:
Thanks for your insights!
Just adding two observations from my side, which made me belief that it could all be due to a buggy Airthings firmware:
- I am not using BlueZ, but have a Bluegiga dongle on a Mac mini and I am seeing similar effects. So it can't be due to anything within the BlueZ stack as it isn't involved in my case.
- My Airthings app also stops being able to connect after a while (a few weeks) and I need to re-pair the devices (although I haven't really figured out a reliable way to do so yet). My assumption so far was that maybe the app interferes with the binding as both require a connection to the sensor, which means that the other connection needs to be dropped for that time. Do you happen to use the app additionally to the binding? If so, does that work reliably for you?

dw-8:
Since I got mine running reliably, I don't keep the Airthings app running in the background anymore, and only enable it periodically to update the Airthings cloud.  At the beginning, when I was running OH2.5, I did have it syncing all the time, and did not notice any issues, and never had to re-pair the device.

Do the BlueZ/Bluegiga bindings rely on a common bluetooth binding and/or reuse a common bluetooth framework, which may contain a bug?

Can you also check your debug logs and see if your _execute_ subroutine (and read subroutine) is being called twice per refreshinterval.  I didn't notice this with OH2.5, but with OH3.0 and OH3.1, I found onServicesDiscovered is getting triggered (which then calls the execute and read subroutines and) when the disconnect routine is run, without any prior call to discoverServices being made during that loop since it was in a RESOLVED state.  So it seems that BlueZ and/or Airthings is forcing a service discovery when disconnecting, if a service discovery was not performed during that connection session.
Normally, the discoverServices is only run once after first initializing the device, and then sets the serviceState to RESOLVED, so that it is not run on all subsequent connections/reads.  To fix this extra read, I removed the execute(); line from onServicesDiscovered.  I wondering if this issue is unique to BlueZ and/or the Airthings Wave Gen 1, or if it occurs with Bluegiga and/or the Airthings Wave Mini as well.

paulianttila (commented):
Some late comments even this is already merged.

kaikreuzer:
@dw-8 Could you please do a follow-up PR to address @paulianttila comments? Thanks!

PR #11045, opened 2021-07-22
[fsinternetradio] Add further supported devices

siglp:
Add binding for Hama IR100, Hama DIR3100 and Auna Connect CD


kaikreuzer (approved):
Thanks!

PR #11047, opened 2021-07-22
[espmilight] Add color channels ability to trigger white LEDs

Skinah:
Before these changes with RGB and dual white globes you could not use the color gradients from a color element to switch back to using white LEDs. This change allows a user threshold to be specified where low SATURATION values will change to using the real white LEDs instead of a pastel colour. This is a big improvement as the openHAB colour indicator would show white when the globe was not white.

PR also contains minor refactoring to fix some compiler warnings.


fwolter (approved):
LGTM

Issue #5982, opened 2019-08-26
[pushbullet] Backward 1.x compatibility issue with broadcast notifications


jsetton:
## Expected Behavior
In the Pushbullet 1.x binding, notifications could be broadcast to all the devices of a given account by calling the [action](https://github.com/openhab/openhab1-addons/blob/master/bundles/action/org.openhab.action.pushbullet/src/main/java/org/openhab/action/pushbullet/internal/PushbulletAPIConnector.java#L105) method with a title and message.

## Current Behavior
While testing the new binding introduced in #5668, this functionality isn't working anymore. It seems that now a recipient is always needed. So, when calling the action with two parameters, the recipient would be set with the title causing the notification being sent out to miss that information.

### Rule
```php
val actions = getActions("pushbullet", "pushbullet:bot:default")
actions.sendPushbulletNote("Test Notification", "Test 2.x")
```

### Logs
```plain
2019-08-22 17:54:12.365 [DEBUG] [t.internal.handler.PushbulletHandler] - sendPush is called for
2019-08-22 17:54:12.365 [DEBUG] [t.internal.handler.PushbulletHandler] - Thing org.eclipse.smarthome.core.thing.internal.ThingImpl@6f24a793
2019-08-22 17:54:12.366 [DEBUG] [t.internal.handler.PushbulletHandler] - Thing Label: 'Pushbullet Default Bot'
2019-08-22 17:54:12.366 [DEBUG] [t.internal.handler.PushbulletHandler] - CFG org.openhab.binding.pushbullet.internal.PushbulletConfiguration@176fd872
2019-08-22 17:54:12.366 [DEBUG] [t.internal.handler.PushbulletHandler] - Headers: {Content-Type=application/json, User-Agent=openHAB / Pushbullet binding 2.5.0.201908182253, Access-Token=xxxxxxxxxx}
2019-08-22 17:54:12.366 [DEBUG] [t.internal.handler.PushbulletHandler] - Recipient is 'Test Notification'
2019-08-22 17:54:12.367 [DEBUG] [t.internal.handler.PushbulletHandler] - Title is     ''
2019-08-22 17:54:12.367 [DEBUG] [t.internal.handler.PushbulletHandler] - Message is   'Test 2.x'
2019-08-22 17:54:12.367 [WARN ] [t.internal.handler.PushbulletHandler] - Invalid recipient: Test Notification
[...]

PR #11057, opened 2021-07-25
[pushbullet] Update README.md with recipient broadcast information

jsetton:
Documentation on when a note will be broadcast based on recipient value.

Related to #5982


kaikreuzer (approved):
Thanks!

PR #11063, opened 2021-07-28
[yamahareceiver] Add possiblity to switch HDMI outputs on and off on Yamaha receivers (openhab#6327)

fragger42:
[yamahareceiver] Add possiblity to switch HDMI outputs on and off on Yamaha receivers

This is an improvement to add channels that allow switching the HDMI1 Output and HDMI2 Output on and off on some Yamaha receivers. The readme lists the two new channels.


hmerk:
@fragger42 Thanks for your contribution. It seems to add your changes to the tests, therefor the PR build is failing.
Could you please update the tests as well, so the build is ok.

fragger42:
@hmerk: The problem seems to be that I did separate requests to get the current state instead of using the Main_Zone XML document where the information is also provided. These requests then resulted in a Null pointer exception because they don't work on the build server. I just committed the change and on my local eclipse the tests work again. Hopefully, now everything is ok.

hmerk:
@fragger42 Great, let's wait and see if Jenkins is happy again ...

fragger42:
@hmerk: Seems like it worked fine ;)

kaikreuzer (approved):
lgtm, thanks!

PR #11067, opened 2021-07-31
[ipcamera] Improve FFmpeg motion detection to support wider FPS range

### LINKED ISSUE #10683: [IPCamera] No FFMPEG Motion Detection
Motion detection via ffmpeg is not working reliably in low time resolution environments. 

## Expected Behavior
Motion detection triggered upon scene change.

## Current Behavior
No motion detected regardless of scene changes and configured motion threshold.

## Possible Solution
Looking at the code, I suspect that counting motion indicators (scene changes "lavfi.scene") and status reports ("speed=") is not working correctly. ffmpeg, by default, appears to be reporting "speed=" every second. Therefore any detected scene changes will immediately be cancelled out. 

Alternatives
* Trigger motion alarm per each (except the first) detected scene. Then, it all depends on the  threshold configuration
* Keeping time and counting the number of changed scenes per time interval. ==> less intuitive for the user because there are two tuning parameters (sensitivity and time frame)
* Increasing the reference period without explicit time keeping. I.e., counting the number of scene changes per, say, 10 status ("speed=") reports. 

1) Set up an mjpeg stream with a time resolution of 1Hz or slower.
2) Ingest in ipcamera and configure motion detection (threshold != 0)
3) Enabled debug logs for ipcamera binding
4) Watch the logs

## Context
Detect motion in a stream of 1 picture per second coming off an ESP32-cam.

## Your Environment
* On docker (latest), x64
* ffmpeg version 4.1.6-1~deb10u1 Copyright (c) 2000-2020 the FFmpeg developers
built with gcc 8 (Debian 8.3.0-6)

Skinah:
This PR improves the FFmpeg motion detection to support a wider range of framerates. More in depth details are in the Issue that this PR Closes #10683

The changes make the whole range of the motionControl channel give better useable results.

Refactoring also done to remove the last of the compiler warnings.


fwolter (approved):
LGTM

PR #11079, opened 2021-08-01
[avmfritz] Invert opening level of HAN-FUN blinds, when a command switch is set to ON

quidam:
While openHAB generally interprets a blind opening level of 0% as fully closed and 100% as fully open, the HAN-FUN blinds interpret the values vice versa. This leads to the need to convert the values in rules and to wrong representation of the opening levels in widgets.
So I have implemented a command switch 'invertLevel' for the HAN-FUN Blind device type. If this switch is set to ON, the opening level of the blind is inverted (i.e. newValue is 100 - oldValue) both when reading the state and when sending commands.

### DIFF
--- a/bundles/org.openhab.binding.avmfritz/README.md
+++ b/bundles/org.openhab.binding.avmfritz/README.md
@@ -156,6 +156,11 @@ If the FRITZ!Powerline 546E is added via auto-discovery it determines its own `a
 
 - `ain` (mandatory), no default (AIN number of the device)
 
+### HAN-FUN Blinds
+
+- `ain` (mandatory), no default (AIN number of the device)
+- `invertLevel` (mandatory), default false (Set this to true to invert the shutter level to match the openHAB standard, where 0 = closed, 100 = open)
+
 ### Finding The AIN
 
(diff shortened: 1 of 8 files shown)


quidam:
It seems that I was mislead by a community widget (universal_shutter_v4) showing a value of 0 as completely closed and the Alexa bindig, which does the same (because Alexa does). So my feature is not needed and I will close my pull request.

BTW: I decided to use a switch because I could not find a specification for the level in the interface specification, so other rollershutter types, than the two that I own, may function differently. Besides, people may have created rules based on the present functionality, which I did not want to break.

 

Thanks for your attentiveness and sorry for the unnecessary work I created.

 

Von: Kai Kreuzer ***@***.***> 
Gesendet: Montag, 2. August 2021 12:48
An: openhab/openhab-addons ***@***.***>
Cc: quidam ***@***.***>; Author ***@***.***>
Betreff: Re: [openhab/openhab-addons] [avmfritz] Invert opening level of HAN-FUN blinds, when a command switch is set to ON (#11079)

 

@kaikreuzer requested changes on this pull request.

openHAB generally interprets a blind opening level of 0% as fully closed and 100% as fully open

Err, no. It is the other way round: For openHAB 0% is up (=open), 100% is down (=closed).

So how do HANFUN devices interpret it? If there is only one way to correctly wire them, I do not see any need for an additional configuration. The binding should instead simply do it right and no rules would be required for anyone. What do I miss?

PR #11089, opened 2021-08-07
[ipcamera] Fix bugs, IndexOutOfBoundsException and remove sleep.

Skinah:
Changes are to:
1. Fix a reported bug (via the forum) where a IndexOutOfBoundsException is thrown when the binding sends a PTZ request before ONVIF is fully connected.
2. Remove the use of sleep and to change to using threadpool.
3. Fixes a bug that can happen when a camera keeps sending ONVIF events even after the handler is destroyed.


kaikreuzer (approved):
lgtm, thanks

PR #11094, opened 2021-08-09
[homematic] Added inverted info for rollershutters

bredmich:
Hi there,

Currently openHAB and the CCU are using different values for the same state of a rollershutter.
This is not part of the docs and could confuse new users.
There are already posts in the community about this with workarounds like proxy items.

Added information about this under Troubleshooting and listed some examples.

Signed-off-by: Michael Bredehorn michael@bredehorn.nrw (github: bredmich)


kaikreuzer (approved):
lgtm, thanks

Issue #10325, opened 2021-03-14
[influxdb] "analyze" and charts not showing any data when using alias


SNival:
## Expected Behavior
"analyze" item data (and charts) should display values whether or not they are persisted using an alias in influxDB.

## Current Behavior
"Analyze" draws the chart of item data that are persisted without alias, but is empty for aliased ones.

## Steps to Reproduce (for Bugs)
Persistence settings
```
influxdb.persist
...
gpTemperature* : strategy = everyUpdate
gpPower* -> "power" : strategy = everyChange
```

Items settings
```
test.items
Number:Power Pwr_Shly1     "Cons. instant. [%.1f %unit%]" <energy>
  (geLight, gpPower) ["Measurement", "Power"]
  { channel="shelly:shellyplugs:6a5cc9:meter#currentWatts" }

Number:Temperature Tmp_EnOcean1 "Temp. EnOcean [%.1f %unit%]" <temperature>
  (geThermoEO, gpTemperature) ["Measurement", "Temperature"]
  { channel="enocean:temperatureSensor:FTXMIIG5:demoTemp:temperature" }
```

Data stored in the database:
```
> select * from Tmp_EnOcean1 order by time desc limit 5
name: Tmp_EnOcean1
time                     item         value
----                     ----         -----
2021-01-31T08:10:45.946Z Tmp_EnOcean1 20.862745098039216
2021-01-31T07:53:15.855Z Tmp_EnOcean1 20.862745098039216
2021-01-31T07:35:45.441Z Tmp_EnOcean1 21.333333333333332
2021-01-31T05:07:12.139Z Tmp_EnOcean1 20.54901960784314
2021-01-31T04:51:27.372Z Tmp_EnOcean1 20.07843137254902

> select * from power order by time desc limit 5
name: power
time                     item      value
----                     ----      -----
2021-01-31T08:07:32.359Z Pwr_Shly1 0
2021-01-31T08:06:32.191Z Pwr_Shly1 21.5
2021-01-31T08:05:32.01Z  Pwr_Shly1 21.6
2021-01-31T08:04:31.832Z Pwr_Shly1 21.5
2021-01-31T08:02:31.472Z Pwr_Shly1 21.6
```

## Possible Solution
It seems that the data are queried by item name in both cases: here are the queries logged by influxdb:
For the non aliased item:
```
[...]

PR #11099, opened 2021-08-12
[influxdb] Use item instead of _measurement for influxdbV2 (#10325) 

martin67:
# Use item instead of _measurement for influxdbV2 (#10325)

## DESCRIPTION
There is a problem when using aliases in influxdb.persist and the naming of measurements. See #10325.
 
This PR removes the dependency on the _measurement and instead uses the item tag for selecting values/series from influx. As the item names are unique, this should (hopefully :-)) work as before.

I can't really see the requirement of defining a _measurement manually. Normally each series has the same _measurement name as the item name, unless an alias is used (in influxdb.persist) - then the alias will be used as _measurement instead. The item name/tag will be the same in both cases. So why not use the item tag directly when selecting from influx?

The documentation will have to be updated as well - but the PR should probably be discussed before that.


lujop (changes requested):
Using only item tag name will have regressions on migrated old data.
See #10937

martin67:
I understand the setup now and the regression requirements. I'll see if I can come up with a new PR that handles that.

PR #11098, opened 2021-08-10
[mqtt.espmilighthub] Sematic Tags added.

Skinah:
Added sematic tags.

Signed-off-by: Matthew Skinner <matt@pcmus.com>

PR #11100, opened 2021-08-12
[OmniLink] Add semantic tags to channels

ecdye:
Signed-off-by: Ethan Dye <mrtops03@gmail.com>


digitaldan:
Hi @ecdye , have not yet played with adding semantic tags to the thing definition xml yet, definitely peaked my interest!   After looking up the documentation, it seems to be limited right now to a small number of tags,  not sure if the documentation is up to date, but curious where the tags you are using come from?  The docs at  https://next.openhab.org/docs/developer/bindings/thing-xml.html#default-tags  states:

"Please note that only tags from a pre-defined tag library should be used. This library is still in development., and only a very small set of tags are defined so far:"

Tag | Item Types | Description
-- | -- | --
Lighting | Switch, Dimmer, Color | A light source, either switchable, dimmable or color
Switchable | Switch, Dimmer, Color | An accessory that can be turned off and on.
CurrentTemperature | Number, Number:Temperature | An accessory that provides a single read-only temperature value.
TargetTemperature | Number, Number:Temperature | A target temperature that should engage a thermostats heating and cooling actions.
CurrentHumidity | Number | An accessory that provides a single read-only value indicating the relative humidity.

<h3 id="state-description" style="font-weight: 600; line-height: 1.25; font-size: 1.35rem; margin-top: -3.1rem; padding-top: 4.6rem; margin-bottom: 0px; color: rgb(221, 221, 221); font-family: -apple-system, system-ui, &quot;Segoe UI&quot;, Roboto, Oxygen, Ubuntu, Cantarell, &quot;Fira Sans&quot;, &quot;Droid Sans&quot;, &quot;Helvetica Neue&quot;, sans-serif; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; background-color: rgb(17, 17, 17); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;"><a href="https://next.openhab.org/docs/developer/bindings/thing-xml.html#state-description" class="header-anchor" style="font-weight: 500; color: rgb(255, 102, 0); text-decoration: none; font-size: 0.85em; float: left; margin-left: -0.87em; padding-right: 0.23em; margin-top: 0.125em; opacity: 1;">#</a></h3><br class="Apple-interchange-newline">

ecdye:
Hmm, interestingly a number of other PRs with similar tags have been accepted, including ones not listed in that list. See https://github.com/openhab/openhab-addons/pull/10969
https://github.com/openhab/openhab-addons/pull/11098
https://github.com/openhab/openhab-addons/pull/10977/files

digitaldan:
Yeah, i figured the docs might be out of date and some precedence was being set elsewhere, thanks for pointing out those PRs.

ecdye:
> Yeah, i figured the docs might be out of date and some precedence was being set elsewhere, thanks for pointing out those PRs.

No problem, a user pointed it out to me on the forums and asked if I was willing to implement this because many of us have large numbers of lights and other devices that can be a pain to manually tag one at a time.

lolodomo:
This Is controversial. Please read the PRs I submitted, @kaikreuzer does not like the current approach consisting in adding tags on channels. 
I don't know if he agrees to add at least "obvious" tags?

Skinah:
@digitaldan I don't know if they are documented, but here is how I find them...
When you edit an item you can get a drop down list of possible options. `Semantic Class` and `Semantic Property`.
[www.pcmus.com/openhab/sematic.png](url)

Why add them? See this picture of my locations and note the Exclaimation mark, this is what appears when an item is marked as an Alarm and it's state turns ON. The temperature shows up when 'Measurement' and 'Temperature' is tagged and these are POINTS directly in the location, must not be under an equipment. 'Control' and 'Light' gives the locations a number of how many lights are turned ON in that location.. etc.
[www.pcmus.com/openhab/locations.png](url)

Note that this only takes effect if you create the items AFTER the binding is already updated and the user is offered the ability to change the tags when the items are created.

Personally I believe the framework could be made to not need the tags added for some use cases, unless it increases the complexity of the core too much then it is probably easier to leave it to tags as a binding maintainer will have better grasp on what is a good set of tags. One example is a channel that uses Number:Temperature and also marked as read only, if there are no binding tags should the core treat it as 'Measurement' and 'Temperature' ? My opinion in this case is yes.

Another example is a color channel in the WLED binding. These led light strips can have a primary color and second color selected and then the light will alternate between these two colors. The fact that the color item has an ON state is wrong as the light can be turned OFF, so you really don't want the sematic labels showing that a room has a light left on then the light is OFF. Marking every color item as a 'Control' for a 'Light' in this case is wrong.

I have had to work this out myself, I have not seen any guidance on the topic anywhere.

bundles/org.openhab.binding.omnilink/src/main/resources/OH-INF/thing/thermostat.xml
```diff
@@ -85,6 +89,10 @@
 		<label>Outdoor Temperature</label>
 		<description>The current outdoor temperature detected by this thermostat.</description>
 		<category>Temperature</category>
+		<tags>
+			<tag>Measurement</tag>
+			<tag>Temperature</tag>
+		</tags>
 		<state readOnly="true" pattern="%.1f %unit%"/>
 	</channel-type>
```
Skinah, line 97: I would question this one as I believe a better way forward would be to move to using system.outdoor-temperature See https://www.openhab.org/docs/developer/bindings/thing-xml.html#system-state-channel-types and then you allow the framework to handle the tags for this.

bundles/org.openhab.binding.omnilink/src/main/resources/OH-INF/thing/unit.xml
```diff
@@ -198,6 +198,10 @@
 		<label>Unit Level</label>
 		<description>Increase/Decrease the level of this unit.</description>
 		<category>Slider</category>
+		<tags>
+			<tag>Control</tag>
+			<tag>Light</tag>
+		</tags>
 		<state min="0" max="100" pattern="%d %%"/>
 	</channel-type>
```
Skinah, line 206: Is this a light globe or a screen brightness control? See my long comment that shows how the lights appear on the location tabs, is that a desired result you want? I Don't have any of the devices from this binding to understand from reading the bindings documentation.

bundles/org.openhab.binding.omnilink/src/main/resources/OH-INF/thing/unit.xml
```diff
@@ -206,6 +210,10 @@
 		<label>Switch</label>
 		<description>Turn this unit on/off.</description>
 		<category>Switch</category>
+		<tags>
+			<tag>Control</tag>
+			<tag>Light</tag>
+		</tags>
```
Skinah, line 216: As above, does this only switch lights on and off as it will place a count of how many lights are turned on in the location tabs as per my long comment.

bundles/org.openhab.binding.omnilink/src/main/resources/OH-INF/thing/unit.xml
```diff
@@ -273,6 +281,10 @@
 		<label>Switch</label>
 		<description>Turn this room on/off.</description>
 		<category>Switch</category>
+		<tags>
+			<tag>Control</tag>
+			<tag>Light</tag>
+		</tags>
```
Skinah, line 287: As above comments, as I am confused why the category is not `lightbulb` if this channel is used for lights. Docs do not give enough info on what this channel does at a quick glance.

bundles/org.openhab.binding.omnilink/src/main/resources/OH-INF/thing/unit.xml
ecdye: Yes, an omnilink unit is traditionally a light switch that turns on or of a light, with dimmer functionality. There were a few non light switch units available but even when they were made they were rare.

bundles/org.openhab.binding.omnilink/src/main/resources/OH-INF/thing/unit.xml
ecdye: Yes same as above, but this is only for simple on/off with no dimmer functions for switches that don't support dimming

bundles/org.openhab.binding.omnilink/src/main/resources/OH-INF/thing/thermostat.xml
ecdye: I have it as that already but as of now the UI does not take advantage of that in displaying it to the user in the locations page.

bundles/org.openhab.binding.omnilink/src/main/resources/OH-INF/thing/unit.xml
ecdye: The reasoning is that physically, the thing is a switch but the switch controls the light. The smart functionality is not in the bulb as it is in some cases but rather in the switch that controls the light. As such by following the strict definition of an openHAB Thing it made the most sense to make it a switch categorically.

bundles/org.openhab.binding.omnilink/src/main/resources/OH-INF/thing/thermostat.xml
Skinah: You must be using 3.1 Stable or older, as of 3.2 Milestone 1 I can confirm that the tags are added by the core as I tested it just now and I saw the PR get merged.
Just a tip it needs to be a POINT added directly in the semantic model under the LOCATION, it can not be inside an EQUIPMENT group for it to display.

bundles/org.openhab.binding.omnilink/src/main/resources/OH-INF/thing/unit.xml
Skinah: If you think it needs to be tagged then feel free to resolve this. The tags are only suggestions and can be changed at any time by the user, so if you think more people will benefit VS the number of people that see it as a bug and complain if the light counter is not accurate, I'm happy to agree with your assessment.

bundles/org.openhab.binding.omnilink/src/main/resources/OH-INF/thing/unit.xml
Skinah: As above comment. If you think it needs to be tagged then feel free to resolve this.

bundles/org.openhab.binding.omnilink/src/main/resources/OH-INF/thing/unit.xml
Skinah: If <category>lightbulb</category> is used you get a light bulb icon that dynamically changes to reflect the state. But this is not important, all that needs to be considered is can it get used to switch a FAN on and off? If its use is only for lights, then I agree with the tags getting added. If the channel can be used for switching multiple different devices like a fan, then I suggest it should not be tagged in this way. As per the above comments, feel free to resolve this yourself if my concerns are wrong.

swamiller: Adding my two cents, I currently have ~53 UPB units controlled by the omnilink binding.  All are lights except for 2 bathroom fans.  The OmniPro board (and binding) treats them the same.  They are all UPB units and all show up as lights as default. I would prefer to go back and correct the 2 fans after discovery versus manually tagging all 53 for an odd or different device.

bundles/org.openhab.binding.omnilink/src/main/resources/OH-INF/thing/thermostat.xml
ecdye: > Just a tip it needs to be a POINT added directly in the semantic model under the LOCATION, it can not be inside an EQUIPMENT group for it to display.

That would probably be why mine doesn't display

bundles/org.openhab.binding.omnilink/src/main/resources/OH-INF/thing/unit.xml
ecdye: That is my opinion as well, @swamiller.

Skinah: Happy with that so closing this.

ecdye:
Any updates on this PR? What needs to be done before it can be merged?

ecdye:
Bump

ecdye:
@lolodomo Has there been any update on if this is going to be the accepted approach or not? If it is not, then can you point me to what is so I can update the PR.

lolodomo:
Unfortunately, this approach was more or less rejected without any concrete alternative. 
Adding just a tag "Control" makes no sense, I agree with that. 
Discussion here: https://github.com/openhab/openhab-core/issues/1791

ecdye:
@lolodomo A follow up question, if by all technicality this works as of right now's there any problem with implementing it as is until a better solution has been created/decided. 

@kaikreuzer Your input would be welcome as well.

Personally I believe that for the moment we should at the very least allow this sort of workaround to make the user experience better until a better solution is made. Especially because even advanced users like myself don't take the time to create very custom pages in the webUI. Just imagine the beginner user who is trying to add these items and then never finds them because the were not tagged properly in the model.

I personally see the overview tabs / semantics tabs as very powerful if used correctly for all users which could make openHAB easier to get into for the average user as they have understandable and readily available access to their devices directly from the home page without having to go manually create a sitemap which IMHO is actually a very daunting task for a brand new home automation user.

jlaur:
> One example is a channel that uses Number:Temperature and also marked as read only, if there are no binding tags should the core treat it as 'Measurement' and 'Temperature' ? My opinion in this case is yes.

Let me add an example that might make this choice less obvious, or at least add something more to think about. In the Miele@home binding there are temperature channels for an oven - a **target** temperature and a **measured** temperature. Both are read-only since the API doesn't allow setting the target temperature, unfortunately. So in this case it would be wrong to treat it as a measurement.

bundles/org.openhab.binding.omnilink/src/main/resources/OH-INF/thing/bridge.xml
```diff
@@ -27,11 +27,11 @@
...
 		<properties>
-			<property name="model number"/>
-			<property name="major version"/>
-			<property name="minor version"/>
+			<property name="modelNumber"/>
+			<property name="majorVersion"/>
+			<property name="minorVersion"/>
 			<property name="revision"/>
-			<property name="phone number"/>
+			<property name="phoneNumber"/>
```
jlaur, line 34: For dynamically set properties you don't need to define them here. This will result in them being shown as empty properties until they will be set by code.

jlaur: Are they set in code? Since renamed, question is if code should be adapted as well, or you renamed them here to align with names in current code? Besides that, if these are firmware properties, modelNumber should be renamed to **modelId** and majorVersion/minorVersion/revision could be combined in **firmwareVersion**.

bundles/org.openhab.binding.omnilink/src/main/java/org/openhab/binding/omnilink/internal/OmnilinkBindingConstants.java
```diff
@@ -104,7 +104,7 @@
     public static final String CHANNEL_ROOM_SCENE_D = "scene_d";
     public static final String CHANNEL_ROOM_STATE = "state";
 
-    public static final String CHANNEL_SYSTEMDATE = "sysdate";
+    public static final String CHANNEL_SYSTEMDATE = "system_date";
```
jlaur, line 107: This is a breaking change for anyone having already linked items to the old channel name. Is it needed?

bundles/org.openhab.binding.omnilink/src/main/resources/OH-INF/thing/audio-zone.xml
```diff
@@ -13,10 +13,10 @@
 		<description>An audio zone configured in the controller.</description>
 		<channels>
 			<channel id="zone_power" typeId="audio_zone_power"/>
-			<channel id="zone_mute" typeId="audio_zone_mute"/>
-			<channel id="zone_volume" typeId="audio_zone_volume"/>
+			<channel id="zone_mute" typeId="system.mute"/>
```
jlaur, line 16: It's a good idea to use system channel types when possible, but you will need to add own label and description now since they are no longer taken from the now deleted custom channel type.

jlaur (changes requested):
A few generic comments added as I don't know this binding.

lolodomo:
@jlaur : for your information,, @kaikreuzer is generally against this approach consisting in adding semantics on channels. So before merging, maybe you should have his agreement. I imagine he agrees only when the semantic is obvious for the channel.

kaikreuzer (approved):
Sorry that I never commented here.

I just had to recap myself on the semantic tags. I think my main concerns where tags, which weren't 100% obvious that they are correct like in [this](https://github.com/openhab/openhab-addons/pull/10973#pullrequestreview-703854388) or [that](https://github.com/openhab/openhab-addons/pull/10972#pullrequestreview-703822544) case.
Seeing the issues that users have with manually tagging items through the Main UI right now, I would agree that it is absolutely fine and welcome to add tags in the obvious situations - like I consider those in that PR.
So from my side it can be merged, once @jlaur's comments are addressed.

lolodomo:
This is a breaking change for users Using UI to setup things/channels. Replacing the channel type of a channel is a breaking change. It just means that a new entry is expected in the file update,ps1 in the district repo after the merge.

jlaur:
> Replacing the channel type of a channel is a breaking change.

Please see https://github.com/openhab/openhab-addons/pull/11668#discussion_r764135912

lolodomo:
> > Replacing the channel type of a channel is a breaking change.
> 
> Please see [#11668 (comment)](https://github.com/openhab/openhab-addons/pull/11668#discussion_r764135912)

My feeling here is that we are EXACTLY in the case of the hue binding where existing channel types were replaced by system channel types. This lead to many complaints in the community forum (do a search) because users did not see anymore their channels (in the thing UI) until they remove and create again the thing (meaning all channel links have to be redone).

I am not discouraging this change, I am even in favour of this change. I just want that we alert the users.

One time again, this change is fully transparent for any user like me who relies on config files.

jlaur:
> My feeling here is that we are EXACTLY in the case of the hue binding where existing channel types were replaced by system channel types.

I don't know the details of that case. But I tested (while working in 3.2) that changing **channel type id** while preserving **channel id** and **item type** does not cause any problems after openHAB has been restarted. If this scenario is the same as you have bad feelings about, I can only encourage you to try it out also for certainty.

lolodomo:
> I don't know the details of that case. But I tested (while working in 3.2) that changing **channel type id** while preserving **channel id** and **item type** does not cause any problems after openHAB has been restarted. If this scenario is the same as you have bad feelings about, I can only encourage you to try it out also for certainty.

Yes, I should probably just to convince myself. Did your test match exactly this case ?
I can't imagine so many users could have mentioned an issue which is not one. They discovered it after migrating to a new OH version (from 3.0 to 3.1). You can even find at least two issues created in GitHub.

bundles/org.openhab.binding.omnilink/src/main/java/org/openhab/binding/omnilink/internal/OmnilinkBindingConstants.java
ecdye: Not necessarily needed, but it would bring the naming of this channel in line with the style used on the other channels in the binding. I can omit the change if you think it would be better.

bundles/org.openhab.binding.omnilink/src/main/resources/OH-INF/thing/audio-zone.xml
ecdye: How would I add just a description and nothing else? I may being missing something but I don't see an example in the docs to reference.

bundles/org.openhab.binding.omnilink/src/main/resources/OH-INF/thing/bridge.xml
ecdye: They are set in code, and it was just to align them with the code. I like your idea of using the builtin openhanded properties though so I switch to that.

bundles/org.openhab.binding.omnilink/src/main/resources/OH-INF/thing/audio-zone.xml
jlaur: See #11668 for inspiration as this was a similar case.

ecdye:
Force pushed to bring branch up to date with head.

bundles/org.openhab.binding.omnilink/src/main/resources/OH-INF/thing/audio-zone.xml
ecdye: Fixed

bundles/org.openhab.binding.omnilink/src/main/resources/OH-INF/thing/bridge.xml
ecdye: Fixed, removed the properties from XML files as they are all dynamically set in code.

bundles/org.openhab.binding.omnilink/src/main/java/org/openhab/binding/omnilink/internal/OmnilinkBindingConstants.java
ecdye: @jlaur Would you prefer to revert this change or move forward. It is not necessary just brings the style in line with the rest of the code.

jlaur: @ecdye - three options as I see it. Up to you how much effort you want to put into it, I'm fine with any of them:
1. Leave it as is to not introduce a breaking change. It may not be pretty, but binding doesn't seem to conform to channel naming convention anyway.
2. Rename it and introduce a breaking change. This must be mentioned in release notes, so another PR for add a line to https://github.com/openhab/openhab-distro/blob/main/distributions/openhab/src/main/resources/bin/update.lst would be needed.
3. Rename it and provide backwards compatibility in code. See for example #11668.

ecdye: @jlaur I decided to go with option one for them moment, I would rather get this merged and save that change for when I have the time to do it properly.

jlaur:
@ecdye - please update resources/OH-INF/i18n/omnilink.properties according to new keys in XML files.

jlaur:
> Yes, I should probably just to convince myself. Did your test match exactly this case ?

My test was exactly the case of changing **channel type id** while preserving **channel id** and **item type**. This worked in 3.2 (didn't test earlier versions) after restarting openHAB. I still don't know the exact Hue case, so can't compare. But the concern raised for this PR seems to match what I previously verified not being a problem.

ecdye:
@jlaur Fixed, I think we are ready to merge.

bundles/org.openhab.binding.omnilink/src/main/resources/OH-INF/thing/lock.xml
```diff
@@ -33,6 +29,9 @@
 		<label>Lock/Unlock</label>
 		<description>Lock or unlock this lock.</description>
 		<category>Switch</category>
+		<tags>
+			<tag>Control</tag>
+		</tags>
```
lolodomo, line 34: It was decided that a simple "Control" tag must be avoided.

lolodomo (changes requested):

lolodomo:
@jlaur : what is your status ?
Regarding properties and semantic tags, for me, it is OK.
I have not checked the other changes about channel types and i18n.

jlaur (approved):
Thanks, looks good!

jlaur:
> what is your status ?
> Regarding properties and semantic tags, for me, it is OK.

Looks good to me also.

> I have not checked the other changes about channel types and i18n.

I didn't verify everything, but it seems correct. Since migrated to using system channel types, the labels/descriptions had to be moved from custom channel types to channels directly, which meant also changed i18n keys.

lolodomo:
I let you the lead to merge this PR.

lolodomo:
@ecdye : please keep an eye on the community forum, in case users complain about channels having disappeared.

PR #11103, opened 2021-08-13
[AirQuality] Binding enhancements

clinique:
Adding alert level.
Added semantic tags
Code refinement.
Meaningless channels (location, stationId, city name) moved to properties.
Bug corrected on wrong color returned for a given alert level.
Only keep channels that are provided by the given station.


kaikreuzer:
This PR seems to do a lot more than its title ("Adding semantic tags") suggests.
Would you be able to draw this apart into 2 PRs?

clinique:
> 
> 
> This PR seems to do a lot more than its title ("Adding semantic tags") suggests.
> Would you be able to draw this apart into 2 PRs?

You're right, I came with a very small purpose (adressing an open issue that wasn't one btw) and I found many things that could be enhanced or refined. Instead of making two PR should I rename the PR ?

kaikreuzer:
"Binding enhancements" is not really telling anything, and this is what ends up in the release notes. Also, we cannot have "bug" and "enhancement" labels at the same time on a PR. Therefore it is much appreciated, if a single PR deals with a single bug or enhancement that then clearly turns up in the release notes. If your code changes are too entangled, I would at least suggest to extract the bug fix into a separate PR and keep the rest as various enhancements.

clinique:
Hello @kaikreuzer : closing this attempt and will provide a new one.

Issue #7722, opened 2020-05-21
Removal of dependency on 'org.apache.commons.*'


cweitkamp:
In https://github.com/openhab/openhab-core/pull/1433 and https://github.com/openhab/openhab-core/pull/1441 (and more) we removed dependency on `org.apache.commons.*` libraries. They will be removed from OH3 Core API and thus not available anymore for OH add-ons too. Most of the methods which are used by several bindings are convenience helper methods and can be replaced easily by basic Java methods. Other methods are not yet available in Java 8 but in Java 11 and have to be replace later. It would be a good start to remove as much as we can before.

lolodomo:
I already created 12 PRs to eliminate calls to `IOUtils.closeQuietly`.
As a first feedback in one of these PRs, @J-N-K said we should not close streams silently. So in additional to suppress the call to `IOUtils.closeQuietly`, we need to add a log.
I will update the already existing PRs.

lolodomo:
> I already created 12 PRs to eliminate calls to `IOUtils.closeQuietly`.
> As a first feedback in one of these PRs, @J-N-K said we should not close streams silently. So in additional to suppress the call to `IOUtils.closeQuietly`, we need to add a log.
> I will update the already existing PRs.

Is it oj for all mainteners that we should not close stuff (streams) silently ?
3 of my PRs include a log in case of error and are ready for a review/merge.
9 of my PRs are now tagged with "WIP" waiting to know if this is the way to go or not.

cweitkamp:
Imo it should be okay to log the exception even if I do not really expect it to happen too often. But we should check two other things in advance:

1. Do we really need to close it? Or does the source of the stream takes care of it on its own (e.g. it is not necessary to close an `InputStream` from the [Socket#getInputStream()](https://docs.oracle.com/javase/8/docs/api/java/net/Socket.html#getInputStream--) method)
1. Can it maybe handled even better by using a try-with-resources statement

lolodomo:
> https://docs.oracle.com/javase/8/docs/api/java/net/Socket.html#close--
> Closing this socket will also close the socket's InputStream and OutputStream.

Ok, you're right, this is useless for socket.
We have to check if it is required or not for serial port.

lolodomo:
For serial port, it could depends on the framework implementation.
gnu.io.SerialPort does not seem to close automatically the in/out streams.
https://www.docjava.com/book/cgij/jdoc/gnu/io/CommPort.html#close()

[...]

PR #11109, opened 2021-08-14
[modbus] [WIP] Add support for Sungrow Hybrid Inverters

ferdisdot:
This PR adds support for Sungrow Hybrid Inverters connected via Modbus TCP.
I tested the binding with a Sungrow SH10RT device, but, acccording to  the register definition it should work for all devices of SH*RT series and for devices of SH*RS and SH*K* series.

As this is my first contribution to  openHAB I'm happy on any feedback.


bundles/org.openhab.binding.modbus.sungrow/src/main/java/org/openhab/binding/modbus/sungrow/internal/dto/InverterModelBlock13k.java
```diff
@@ -0,0 +1,208 @@
...
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package org.openhab.binding.modbus.sungrow.internal.dto;
+
+import java.util.Optional;
+
+import org.apache.commons.lang3.builder.ReflectionToStringBuilder;
+import org.apache.commons.lang3.builder.ToStringStyle;
```
ssalonen, line 18: I believe you should rewrite the toString without apache commons. See https://github.com/openhab/openhab-addons/issues/7722. I am suggesting auto-generated methods with eclipse IDE.

bundles/org.openhab.binding.modbus.sungrow/src/main/java/org/openhab/binding/modbus/sungrow/internal/discovery/SungrowDiscoveryProcess.java
```diff
@@ -0,0 +1,386 @@
...
+        if (block.moduleID == FINAL_BLOCK) {
+            parsingFinished();
+        } else {
+            baseAddress += block.length;
+            if (block.moduleID == COMMON_BLOCK) {
+                readCommonBlock(block); // This is an asynchronous task
+                return;
+            } else {
+                // createDiscoveryResult(block);
+                // lookForSunSpecBlock();
```
ssalonen, line 233: Commented out code -- either remove or uncomment.

fwolter:
@ferdisdot Did you address @ssalonen comments?

ferdisdot:
Hi @fwolter, I started to work on it, but last month were quite busy. So the last few month I couldn't work on it. I hope I may get back to this until end of January. If you would like to contribute in the meantime feel free to do so.
As this is my first work in java I'm not sure about the type questions that @ssalonen raised (int vs. Integer). I would appreciate any help here.

fwolter:
If you use `Integer` instead of `int` you're consuming a bit more resources and `Integer` can also be `null`, which needs to be handled correctly. So, using `int` makes things simpler and consumes less resources.

You made some changes to the code. Please raise your hand if you're read for another review round.

lsiepel:
@ferdisdot what is the current state of this PR? Are there still things to address before another (hopefully last)  review round can take place?

lsiepel:
Unfortunately it looks like we cannot bring this PR to the finish line. After two years of inactivity i close this PR. If there is a need to restart, feel free to re-open or start a new PR.

PR #11114, opened 2021-08-15
[chamberlainmyq] Added new Close Error and Open Error Switch Items

ofirster:
MyQ Garage Door may have a problem that prevents the door from being opened or closed. This is described in Chamberlain's website: [How to Resolve a Close Error in the myQ App
](https://support.chamberlaingroup.com/s/article/How-to-resolve-a-Close-error-in-the-myQ-app)
When the problem occurs, we would like to inform the user that there is a problem with the door.
The purpose of this Pull Request is to expose two new OpenHAB Switch Items:

1. closeError
2. openError

Each of the items will be ON when the error occurs in the door, and OFF when it doesn't occur.

The new Items can be used to inform the user that he has a problem with the Garage door.


ofirster:
closing because of failed tests

Issue #8264, opened 2020-08-07
[zway] Allow smaller polling intervalls


DataWorm:
Apparently the pollingIntervall in zway binding is the one setting that affects how fast I can receive state updates but it doesn't allow values below one minute. Obviously this reaction time is quite useless for several use cases. For example I don't like to wait a minute for the (not via zway controlled) light to be turned on when movement is detected inside the room. Therefore I would like to ask why I can't use smaller values (like 1-3 seconds) and request to change that.

Ryz3D:
@kaikreuzer Thanks for merging #11122, thus this issue can be closed.

PR #11122, opened 2021-08-16
[zway] Change minimum polling interval to 1 second.

Ryz3D:
Hi,

when using the Z-Way binding there is a significant delay in receiving a new value as there are no events in the Z-Way API, but instead there is constant polling to detect change. For example the Z-Wave Frontend shows constant requests to ``/ZAutomation/api/v1/devices?since=[timestamp]`` as seen here:
nBK8goS
As you can see by the timestamp in the request URL, they are being sent every 1-2 seconds, while the openHAB binding allows a polling interval from 60 seconds to the default 1 hour. Since the Z-Wave interface does such frequent polling I believe it is appropiate for openHAB to at least support the same. I am unsure whether the default 1 hour interval is appropiate, so I left it unchanged.
I have found #8264 mentioning the same thing, but there is no activity there.

Thank you.

### DIFF
--- a/bundles/org.openhab.binding.zway/src/main/resources/OH-INF/config/bridge-config.xml
+++ b/bundles/org.openhab.binding.zway/src/main/resources/OH-INF/config/bridge-config.xml
@@ -52,7 +52,7 @@
 			<description>Password to access the Z-Way server.</description>
 		</parameter>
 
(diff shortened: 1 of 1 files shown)


kaikreuzer (approved):
Thanks, should be fine to use smaller values indeed.

Issue #9792, opened 2021-01-12
[telegram] sendTelegramPhoto is returning 401 when using basic auth


RobertHerter:
Using sendTelegramPhoto with:
val telegramAction = getActions("telegram","telegram:telegramBot:6e13b54c0d")
telegramAction.sendTelegramPhoto(1234L, "http://192.168.0.17/mjpeg/snap.cgi?chn=1", "Kinderzimmer", "user", "password")

always getting error:
[WARN ] [gram.internal.action.TelegramActions] - Download from http://192.168.0.17/mjpeg/snap.cgi?chn=1 failed with status: 401

same sendTelegramPhoto() was working on 2.5 without any problems.

Using Openhab 3.0.0 on ubuntu amd64 20.04

czahorski:
+1 

Same problem after upgrade to OH3 

val telegramAction = getActions("telegram","telegram:telegramBot:9755518a")
telegramAction.sendTelegramPhoto("http://192.168.4.110/cgi-bin/snapshot.cgi", null, "user", "password")

Error :
[gram.internal.action.TelegramActions] - Download from http://192.168.4.110/cgi-bin/snapshot.cgi failed with status: 401

Works if no auth is needed.
val telegramAction = getActions("telegram","telegram:telegramBot:9755518a")
telegramAction.sendTelegramPhoto("http://192.168.4.110/cgi-bin/snapshot.cgi", null)

Skinah:
Does your camera have BASIC or DIGEST auth? The url @czahorski gave is for Dahua and in newer firmwares they only support DIGEST and the telegram binding only supports BASIC. If you use the ipCamera binding that will allow it to work as it supports both methods and then can serve the file with no password needed.

czahorski:
I'm using Basic, I'm still on the old firmware. Did a curl on the URL I
provided and checked it is BASIC (actually my firmware doesn't support
Digest on that specific camera).

Wikibear:
Same issue here with 3.1.M2 and before. http://username:passwort@192.168.1.5/mjpeg/snap.cgi?chn=0 show my webcam picture and telegramPhoto gives me a 401.
telegramPhoto("http://username:passwort@192.168.1.5/mjpeg/snap.cgi?chn=0",null,"username","password")

Wikibear:
Are there no solutions? Month ago and no feedback...

Wikibear:
Build a php proxy, that running on my apache server as workaround.
`<?php
[...]

Issue #7279, opened 2020-04-03
[telegram] Support ReplyKeyboardMarkup


misery:
I like to send a ReplyKeyboardMarkup in Telegram. Please add support for that with a possible parameter.

Currently InlineKeyboardMarkup is supported.

eric1905:
Do you think about something like that one: https://www.schimmer-media.de/forum/index.php?thread/1692-how-to-telegram-sub-men%C3%BCs-erstellen/

I just also came here to open an issue because I'd like to have this enhancement too. I came from iobroker and liked the possibility to show me a preconfigured keyboard and then send some commands to my telegram bot. So "thumbs up" for your issue.

misery:
Yeah, that's what I mean. ReplyKeyboardMarkup is the underlying API for that.

I switch from ioBroker, too. So my bot misses this feature now. ;-)

Necromancer1982:
Hi, 
I'm also looking for the custom Keyboard.
Used it in several µC-Projects and like it very much :-)

eric1905:
@Skinah  I saw that you updated the library in https://github.com/openhab/openhab-addons/pull/11125 and mentioned this issue. But this does not help or is there a solution yet?

Skinah:
It’s on a long list of things I would like to look at. But no code has been written. The library update was needed to get support for the feature so it now just needs to be implemented by someone.

PR #11125, opened 2021-08-18
[telegram] Send message if a file download fails and bug fixes.

### LINKED ISSUE #10807: [telegram] sendTelegramPhoto should be more resilient
I'm using telegram actions to send a photo when someone is ringing the doorbell. Sometimes it happens, that the event is triggered and the photo URL is available, but the photo isn't ready to download.

This error occurs in that case: "Download from https://... failed with status: 404".

That is fatal, because it causes that no telegram message is sent. It would be good when at least the message/caption from sendTelegramPhoto is sent without the photo. Or better: When the telegram action trys to sent it again 1 or 2 seconds later.

OpenHAB 3.0.2 is used.

This is my example rule:

rule "Doorbell ring"
when
    Channel "netatmo:NAWelcomeHome:xy:welcomeHomeEvent" triggered INCOMING_CALL
then
    val telegramAction = getActions("telegram","telegram:telegramBot:xy")
    telegramAction.sendTelegramPhoto(WelcomeHome_SnapshotURL.state.toString, "Doorbell ring")
end

Skinah:
Contains bug fixes and also updates the library that is needed to implement a requested feature in #7279
Possible fix for #9792 where basic auth is not working for some (?) people.
closes #10807 Sends message when downloading of a file occurs.

### DIFF
--- a/bundles/org.openhab.binding.telegram/pom.xml
+++ b/bundles/org.openhab.binding.telegram/pom.xml
@@ -22,7 +22,7 @@
     <dependency>
       <groupId>com.github.pengrad</groupId>
       <artifactId>java-telegram-bot-api</artifactId>
-      <version>4.9.0</version>
+      <version>5.2.0</version>
       <scope>compile</scope>
     </dependency>
     <dependency>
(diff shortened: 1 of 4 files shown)


kaikreuzer (approved):
lgtm, thanks!

PR #11127, opened 2021-08-18
[semsportal] Fix lastUpdate parse format to accept the format provided by the inverter

jgesser:
The format used by Semsportal in the field `last_refresh_date` varies according to the locale of the inverter.
It was not working here in my installation (Brazil) due to formatting exception to parse the date of the last_refresh_date (lastUpdate) field.

However, Semsportal api provides in its response a field with the date format used in the last_refresh_date field (Only the date part of the format, not the hour part, that's odd).
So, I changed the code to parse the lastModified using the date format provided by the inverter and removed the hard coded date format.
I also changed one test to be parametrized to run using two different inputs as the Semsportal response (with different date format).

Signed-off-by: Julio Gesser <jgesser@gmail.com>


bundles/org.openhab.binding.semsportal/src/main/java/org/openhab/binding/semsportal/internal/dto/StationInfo.java
```diff
@@ -0,0 +1,30 @@
...
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package org.openhab.binding.semsportal.internal.dto;
+
+import com.google.gson.annotations.SerializedName;
+
+/**
+ * POJO containing details about the inverter. Only a very small subset of the available properties is mapped.
+ * 
+ * @author Julio Gesser
```
kaikreuzer, line 20:
```suggestion
 * @author Julio Gesser - Initial contribution
```

kaikreuzer (approved):
lgtm, many thanks!

PR #11121, opened 2021-08-16
[danfossairunit] Change Humidity channel type from Number to Number:Dimensionless

jlaur:
Fixes #9755

Change channel humidity#humidity to Number:Dimensionless to correctly express and format relative humidity as percent.

As effect, items tracking this channel can be seen in logs like:
[INFO ] [openhab.event.ItemStateChangedEvent ] - Item 'DanfossHRV_Humidity' changed from 59.2 % to 59.6 %

As opposed to previous behaviour:
[INFO ] [openhab.event.ItemStateChangedEvent ] - Item 'DanfossHRV_Humidity' changed from 59.2 to 59.6

With a properly defined item without any custom formatting:
`Number:Dimensionless DanfossHRV_Humidity "Relative humidity" <humidity> ["Measurement", "Humidity"] { channel = "danfossairunit:airunit:a2:humidity#humidity" }`

Sitemaps can format value correctly:
`Text item=DanfossHRV_Humidity`

Previously something like this was needed for Number item:
`Text item=DanfossHRV_Humidity label="Relative humidity [%s %%]"`

Signed-off-by: Jacob Laursen <jacob-github@vindvejr.dk>

PR #11129, opened 2021-08-18
[danfossairunit] Add channel descriptions and categories

jlaur:

Changes:
- Use system default channel type for battery level.
- Add channel categories where applicable.
- Add missing channel descriptions.
- Fix two channel descriptions (also in README) as they are not 10-step.

Semantic tags postponed until further investigated and to avoid conflict with #11121.


kaikreuzer (approved):
lgtm, thanks!

PR #11130, opened 2021-08-19
[telegram] [influxdb] Update okhttp3 lib

Skinah:
This makes it easier to change the lib version in the future and brings two bindings into using the same version.

Link for the change log of the library:
https://square.github.io/okhttp/changelog_3x/


lujop:
Looks good, although in influxdb case it's an inherited dependency from influxdb driver and not completely controlled
And don't know if in the future can be some problems because with a middle or major version upgrade.

lujop (approved):

kaikreuzer (approved):
Thanks!

PR #11135, opened 2021-08-21
[LuxtronikHeatpump] Fix/update channels for setting hot water target temperature

sgiehl:
Based on some feedback on the forum (https://community.openhab.org/t/migration-of-novelan-luxtronic-binding-to-openhab2/70743/118) we were able to figure out, that the socket parameter used to set the hot water target temperature actually was not fully correct.

The former implementation was based on the binding for OpenHAB2, so seems it had been "wrong" there already.

It seems the socket parameter `2` used to set the hot water target temperature actually sets a value called coverage heat pump ("Deckung WP") in the heat pump control, which seems to be a value related to the hot water target temperature.
The hot water target temperature seems to be set by parameter `105`, which automatically also updates parameter `2` to the same value. (While setting `2`, doesn't update `105`)


kaikreuzer (approved):
lgtm, thanks.

PR #11137, opened 2021-08-21
[robonect] Channel for distance from charging station on remote start

t2000:
New channel "status-distance" that shows the distance of the robot from
its charging station while it is searching for the remote starting point.


kaikreuzer (approved):
lgtm, thanks!

PR #11139, opened 2021-08-21
[influxdb] Fix for influxdbv1 retention and table names containing keywords or special chars

### LINKED ISSUE #9790: openhab 3.0.0 influxdb persistence defect with retentionpolicy
Influxdb 1.8.2 with retentionpolicy is used:
`create retention policy "2year" on openhab duration 120w replication 1 default`

Since openhab 3 the behavior is not as before with openhab2 in influxql, ,so the queries are not correct:

writing to database works correctly (result 204):
`[httpd] 192.168.2.244 - openhab [10/Jan/2021:16:12:19 +0000] "POST /write?db=openhab&rp=2year&precision=n&consistency=one HTTP/1.1" 204 0 "-" "okhttp/3.14.4" 9d2bb4d6-535e-11eb-9ee3-0242ac1b0002 2908`

reading from database is not correctly (result 400):
`[httpd] 192.168.2.244 - openhab [10/Jan/2021:15:48:54 +0000] "GET /query?db=openhab&epoch=ms&q=SELECT+value+FROM+2year.LichtEG_Gang+ORDER+BY+time+DESC+LIMIT+1%3B HTTP/1.1" 400 85 "-" "okhttp/3.14.4" 57a442d8-535b-11eb-9d3d-0242ac1b0002 262`

Manual Try on database:

> > SELECT value FROM 2year.LichtEG_Gang limit 1
> ERR: error parsing query: found 2year, expected identifier at line 1, char 19
> > SELECT value FROM "2year".LichtEG_Gang limit 1
> name: LichtEG_Gang
> time                value
> ----                -----
> 1588262006927000000 0

So it seems on reading from database the double quotes were not send urlcoded before and after the retentionpolicyname

### LINKED ISSUE #10398: [influxdb] Error on retrieval of data with non-standard retention policy name (escape issue)
Ever since I migrated to OH3 (now on OH3.1M2), I am no longer able to retrieve historical data from influxdb (1.8). OH3 does, however, persist all the data into the database.

To limit the influxdb size, I am using it as a round-robin-database, and created new retention policies. The default one is named "1m_for_24h", and while this is a valid influxdb name (and works for inserting data into the database from the binding), the binding fails to correctly escape the underscores when retrieving data.

A minimal example from a rule to retrieve data from 10 Minutes ago would be:
`item.historicState(now.minusMinutes(10),"influxdb")`

The query generated by this is:
`SELECT value FROM 1m_for_24h.item WHERE time <= '2021-03-26T01:41:27.811737Z' ORDER BY time DESC LIMIT 1;`
which gives the following error:
`19:19:18.284 [ERROR] [.internal.handler.ScriptActionHandler] - Script execution of rule with UID 'influxtest-1' failed: error parsing query: found 1m, expected identifier at line 1, char 19 in influxtest`

Modifying the query to the correct escape sequence like this:
`SELECT value FROM "1m_for_24h".item WHERE time <= '2021-03-26T01:41:27.811737Z' ORDER BY time DESC LIMIT 1;`
results in correctly retrieving data from the database.

fremel75:
Bugfix for using itemnames beginning with numbers or having retention namnes beginning with numbers
Also solves if using old db where retention was named default wich is now a influxQL keyword

Resolves #9790
Resolves #10398

Added doubleqoutes for retention and table name as documentation for influxQL states
https://docs.influxdata.com/influxdb/v1.8/query_language/spec/#identifiers

### DIFF
--- a/bundles/org.openhab.persistence.influxdb/src/main/java/org/openhab/persistence/influxdb/internal/influx1/Influx1FilterCriteriaQueryCreatorImpl.java
+++ b/bundles/org.openhab.persistence.influxdb/src/main/java/org/openhab/persistence/influxdb/internal/influx1/Influx1FilterCriteriaQueryCreatorImpl.java
@@ -122,7 +122,7 @@ public class Influx1FilterCriteriaQueryCreatorImpl implements FilterCriteriaQuer
         if (escapeTableName) {
             Appender.appendName("\"" + tableName + "\"", sb);
         } else {
-            sb.append("\"" + tableName + "\"");
+            sb.append(tableName);
         }
         return sb.toString();
     }
--- a/bundles/org.openhab.persistence.influxdb/src/main/java/org/openhab/persistence/influxdb/internal/influx1/Influx1FilterCriteriaQueryCreatorImpl.java
+++ b/bundles/org.openhab.persistence.influxdb/src/main/java/org/openhab/persistence/influxdb/internal/influx1/Influx1FilterCriteriaQueryCreatorImpl.java
@@ -117,12 +117,12 @@ public class Influx1FilterCriteriaQueryCreatorImpl implements FilterCriteriaQuer
 
     private String fullQualifiedTableName(String retentionPolicy, String tableName, boolean escapeTableName) {
         StringBuilder sb = new StringBuilder();
-        Appender.appendName(retentionPolicy, sb);
+        Appender.appendName("\"" + retentionPolicy + "\"", sb);
         sb.append(".");
         if (escapeTableName) {
-            Appender.appendName(tableName, sb);
+            Appender.appendName("\"" + tableName + "\"", sb);
         } else {
-            sb.append(tableName);
+            sb.append("\"" + tableName + "\"");
         }
         return sb.toString();
     }
--- a/bundles/org.openhab.persistence.influxdb/src/main/java/org/openhab/persistence/influxdb/internal/influx1/Influx1FilterCriteriaQueryCreatorImpl.java
+++ b/bundles/org.openhab.persistence.influxdb/src/main/java/org/openhab/persistence/influxdb/internal/influx1/Influx1FilterCriteriaQueryCreatorImpl.java
@@ -117,10 +117,10 @@ public class Influx1FilterCriteriaQueryCreatorImpl implements FilterCriteriaQuer
 
     private String fullQualifiedTableName(String retentionPolicy, String tableName, boolean escapeTableName) {
         StringBuilder sb = new StringBuilder();
-        Appender.appendName("\"" + retentionPolicy + "\"", sb);
+        sb.append('"').append(retentionPolicy).append('"');
         sb.append(".");
         if (escapeTableName) {
-            Appender.appendName("\"" + tableName + "\"", sb);
+            sb.append('"').append(tableName).append('"');
         } else {
             sb.append(tableName);
         }
--- a/bundles/org.openhab.persistence.influxdb/src/main/java/org/openhab/persistence/influxdb/internal/influx1/Influx1FilterCriteriaQueryCreatorImpl.java
+++ b/bundles/org.openhab.persistence.influxdb/src/main/java/org/openhab/persistence/influxdb/internal/influx1/Influx1FilterCriteriaQueryCreatorImpl.java
@@ -19,7 +19,6 @@ import static org.openhab.persistence.influxdb.internal.InfluxDBStateConvertUtil
 import org.eclipse.jdt.annotation.NonNullByDefault;
 import org.eclipse.jdt.annotation.Nullable;
 import org.influxdb.dto.Query;
-import org.influxdb.querybuilder.Appender;
 import org.influxdb.querybuilder.BuiltQuery;
 import org.influxdb.querybuilder.Select;
 import org.influxdb.querybuilder.Where;
--- a/bundles/org.openhab.persistence.influxdb/src/test/java/org/openhab/persistence/influxdb/internal/InfluxFilterCriteriaQueryCreatorImplTest.java
+++ b/bundles/org.openhab.persistence.influxdb/src/test/java/org/openhab/persistence/influxdb/internal/InfluxFilterCriteriaQueryCreatorImplTest.java
@@ -76,7 +76,7 @@ public class InfluxFilterCriteriaQueryCreatorImplTest {
         FilterCriteria criteria = createBaseCriteria();
 
         String queryV1 = instanceV1.createQuery(criteria, RETENTION_POLICY);
-        assertThat(queryV1, equalTo("SELECT \"value\"::field,\"item\"::tag FROM origin.sampleItem;"));
+        assertThat(queryV1, equalTo("SELECT \"value\"::field,\"item\"::tag FROM \"origin\".\"sampleItem\";"));
 
         String queryV2 = instanceV2.createQuery(criteria, RETENTION_POLICY);
         assertThat(queryV2,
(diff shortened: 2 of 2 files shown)


J-N-K:
IIRC item names are not allowed to start with a number.

lujop (changes requested):
Thanks for the PR.
I think that the retentionPolicy doesn't need to be escaped because it's covered by the `appendName` function logic.
About the tablename it's ok to escape for the default keyword that isn't covered by the `appendName` but only when `escapeTableName` is true.
When it's false it's because tableName it's `/.*/` and it must not be escaped.

fremel75:
Thanks for the feedback
I have reverted the code for escapeTableName function

But the appendName function doesent seem to work for this properly i guess since its not escaped?
Since it does not work without this change

From influx code
`  public static StringBuilder appendName(final String name, final StringBuilder stringBuilder) {
    String trimmedName = name.trim();
    if (trimmedName.startsWith("\"") || COLUMN_NAME_PATTERN.matcher(trimmedName).matches()) {
      stringBuilder.append(trimmedName);
    } else {
      stringBuilder.append('"').append(trimmedName).append('"');
    }
    return stringBuilder;
  }`

lujop:
> Thanks for the feedback
> I have reverted the code for escapeTableName function
> 
> But the appendName function doesent seem to work for this properly i guess since its not escaped?
> Since it does not work without this change
> 
> From influx code
> ` public static StringBuilder appendName(final String name, final StringBuilder stringBuilder) { String trimmedName = name.trim(); if (trimmedName.startsWith("\"") || COLUMN_NAME_PATTERN.matcher(trimmedName).matches()) { stringBuilder.append(trimmedName); } else { stringBuilder.append('"').append(trimmedName).append('"'); } return stringBuilder; }`

The way it is now looks good to me, except that I think that the escaping inside the `escapeTablename` if it's redundant.

Can you provide a case where it isn't working with the inputs?

fremel75:
I have tested with "default" as retention name since this is a keyword now for the default retention in influx.
14:52:45.806 [DEBUG] [e.influxdb.InfluxDBPersistenceService] - Got a query for historic points!
14:52:45.829 [ERROR] [.internal.JSONResponseExceptionMapper] - Unexpected exception occurred while processing REST request.
org.influxdb.InfluxDBException: error parsing query: found DEFAULT, expected identifier at line 1, char 40


And with my item named "1WireOutsideTempSensor_Temperature"
```
23:47:15.841 [DEBUG] [e.influxdb.InfluxDBPersistenceService] - Got a query for historic points!
23:47:15.844 [ERROR] [.internal.JSONResponseExceptionMapper] - Unexpected exception occurred while processing REST request.
org.influxdb.InfluxDBException: error parsing query: found .1, expected ; at line 1, char 47
```
or by rest api
`"{\"error\":{\"message\":\"error parsing query: found .1, expected ; at line 1, char 47\",\"http-code\":500,\"exception\":{\"class\":\"org.influxdb.InfluxDBException\",\"message\":\"error parsing query: found .1, expected ; at line 1, char 47\",\"localized-message\":\"error parsing query: found .1, expected ; at line 1, char 47\"}}}"`

I see now that Documentation for Openhab states "Names must not begin with number" i did create this in Webui so i thought it was ok and in rrd4j it was working
https://www.openhab.org/docs/configuration/items.html#name

lujop:
> I have tested with "default" as retention name since this is a keyword now for the default retention in influx.
> 
> ```
> 14:52:45.806 [DEBUG] [e.influxdb.InfluxDBPersistenceService] - Got a query for historic points!
> 14:52:45.829 [ERROR] [.internal.JSONResponseExceptionMapper] - Unexpected exception occurred while processing REST request.
> org.influxdb.InfluxDBException: error parsing query: found DEFAULT, expected identifier at line 1, char 40
> ```
> 
> And with my item named "1WireOutsideTempSensor_Temperature"
> 
> ```
> 23:47:15.841 [DEBUG] [e.influxdb.InfluxDBPersistenceService] - Got a query for historic points!
> 23:47:15.844 [ERROR] [.internal.JSONResponseExceptionMapper] - Unexpected exception occurred while processing REST request.
> org.influxdb.InfluxDBException: error parsing query: found .1, expected ; at line 1, char 47
> ```
> 
> or by rest api
> `"{\"error\":{\"message\":\"error parsing query: found .1, expected ; at line 1, char 47\",\"http-code\":500,\"exception\":{\"class\":\"org.influxdb.InfluxDBException\",\"message\":\"error parsing query: found .1, expected ; at line 1, char 47\",\"localized-message\":\"error parsing query: found .1, expected ; at line 1, char 47\"}}}"`
> 
> I see now that Documentation for Openhab states "Names must not begin with number" i did create this in Webui so i thought it was ok and in rrd4j it was working
> https://www.openhab.org/docs/configuration/items.html#name

But with current PR code the retentionPolicy isn't escaped, isn't it?
I can't see why the default isn't escaped with current PR code.

fremel75:
> 
> But with current PR code the retentionPolicy isn't escaped, isn't it?
> I can't see why the default isn't escaped with current PR code.
This PR adds ecaped doubleqoutes on retentionpolicy and item name
Since influx function appendName only adds doubleqoutes and that is not good enough
As you could see the appendName function does deal with if it called with "\"" so there is no problem doing it this way

lujop:
> > But with current PR code the retentionPolicy isn't escaped, isn't it?
> > I can't see why the default isn't escaped with current PR code.
> > This PR adds ecaped doubleqoutes on retentionpolicy and item name
> > Since influx function appendName only adds doubleqoutes and that is not good enough
> > As you could see the appendName function does deal with if it called with """ so there is no problem doing it this way

Excuseme, my last comment was incorrect, I wanted to say, the retentionPolicy **is** escaped, isn't it?

I think that we have some confusion in latest comments. For my the actual PR is ok, I only think that the doublequotes inside the `escapetablename` are redundant but not problematic.

When you say that you have problems, do you mean with this PR code or with actual code?

fremel75:
First i experience no problems with the PR
I understand that you wonder about `escapetablename` but the appender function used there is not working for this plugin.
I dont understand either but it needs to be escaped doubleqoute for the plugin to work and the appenderfunction does not do that as it only doubleqoutes.

lujop:
I don't understand what do you mean with:
`the appenderfunction does not do that as it only doubleqoutes.`
The PR is doing exactly that, doublequoting the tablename. Why do you say that the appenderfunction does not work?

That's the code of the appended function:
```
private static final Pattern COLUMN_NAME_PATTERN = Pattern.compile("\\w+(?:\\[.+\\])?");
public static StringBuilder appendName(String name, StringBuilder stringBuilder) {
        String trimmedName = name.trim();
        if (!trimmedName.startsWith("\"") && !COLUMN_NAME_PATTERN.matcher(trimmedName).matches()) {
            stringBuilder.append('"').append(trimmedName).append('"');
        } else {
            stringBuilder.append(trimmedName);
        }

        return stringBuilder;
    }
```

Can you debug or check logs and say for which input it doesn't work?
For words, it's not escaped, but in theory, it isn't needed.
I'm fine with the escaping, but want to know in which case it's needed and not done by the `appendName`

fwolter:
@lujop Your proposed changes are now part of this PR. Are you confident that these will work? Can you test them?

lujop:
Hi fwolter, I'm pretty sure, but I hadn't time to test.
I will try next week, and also check if I can finish the SQL part on DBQuery addon.

lujop:
@fwolter I've done some little tests and it seems to work and escape it's done:
```2022-01-07 23:51:31.025 [TRACE] [.influxdb.InfluxDBPersistenceService] - Query SELECT "value"::field,"item"::tag FROM "autogen"."LocalMoon_StartTime" ORDER BY time DESC LIMIT 1;```

lujop (approved):

fwolter:
There are some formatting issues. You can fix them with `mvn spotless:apply`.

lsiepel:
@lujop could you fix the formatting to get this build going?

jlaur:
@fwolter, @lsiepel - applied spotless.

fwolter (approved):

PR #11145, opened 2021-08-22
Make spacing consistent in iCloud binding rule example

OneYellowLemon:
Fix an extra space in the iCloud binding docs to make spacing consistent


hmerk (approved):
LGTM

PR #11153, opened 2021-08-24
[openhabcloud] reconnect on connection errors

ssalonen:
Requires https://github.com/openhab/openhab-osgiify/pull/28

Aiming to resolve issues like openhab/openhab-cloud#134 . Common workaround seems to be to restart the bundle, see https://community.openhab.org/t/myopenhab-service-not-working/121711/5 and https://community.openhab.org/t/connection-to-myopenhab-broken/73808/80.

According to documentation (albeit for 2.x Socket IO version) [1], reconnection is responsibility of the user on connect_error events. Indeed, default implementation ([1.0.0](https://github.com/socketio/socket.io-client-java/blob/45fd4fbdd8401ea63c30f6bbda266afd99df751f/src/main/java/io/socket/client/Manager.java#L248-L253) and [1.0.1](https://github.com/socketio/socket.io-client-java/blob/89ef9d09cee799ffb85e0252bcaf1993f9d49af9/src/main/java/io/socket/client/Manager.java#L252-L257)) only tries to reconnect if it is the first time we are connecting... **EDIT: ** please see https://github.com/openhab/openhab-addons/pull/11153#issuecomment-905567791 for more discussion how this behaves in socket io 1.x java client.

[1] Lifecycle diagram in https://socketio.github.io/socket.io-client-java/socket_instance.html

Signed-off-by: Sami Salonen <ssalonen@gmail.com>

### DIFF
--- a/bundles/org.openhab.io.openhabcloud/src/main/java/org/openhab/io/openhabcloud/internal/CloudClient.java
+++ b/bundles/org.openhab.io.openhabcloud/src/main/java/org/openhab/io/openhabcloud/internal/CloudClient.java
@@ -231,10 +231,10 @@ public class CloudClient {
             public void call(Object... args) {
                 if (args[0] instanceof Exception) {
                     Exception e = (Exception) args[0];
-                    logger.warn("Socket.IO re-connect attempt error: {} {}", e.getClass().getSimpleName(),
+                    logger.debug("Socket.IO re-connect attempt error: {} {}", e.getClass().getSimpleName(),
                             e.getMessage());
                 } else {
-                    logger.warn("Socket.IO re-connect attempt error: {}", args[0]);
+                    logger.debug("Socket.IO re-connect attempt error: {}", args[0]);
                 }
             }
         }).on(Socket.EVENT_RECONNECT_FAILED, new Emitter.Listener() {
--- a/bundles/org.openhab.io.openhabcloud/src/main/java/org/openhab/io/openhabcloud/internal/CloudClient.java
+++ b/bundles/org.openhab.io.openhabcloud/src/main/java/org/openhab/io/openhabcloud/internal/CloudClient.java
@@ -125,7 +125,7 @@ public class CloudClient {
     private Set<String> exposedItems;
 
     /**
-     * Backoff strategy for reconnecting on connect_error events
+     * Back-off strategy for reconnecting when manual reconnection is needed
      */
     private final Backoff reconnectBackoff = new Backoff();
 
@@ -263,9 +263,13 @@ public class CloudClient {
                         logger.error("Error during communication");
                     }
                 } else {
-                    // Socket.IO 1.x: 'error' event is emitted from Socket on connection errors that are not retried
+                    // We are not connected currently, manual reconnection is needed to keep trying to (re-)establish
+                    // connection.
                     //
-                    // In Socket.IO 2.x, Socket emits 'connect_error' event.
+                    // Socket.IO 1.x java client: 'error' event is emitted from Socket on connection errors that are not
+                    // retried, but also with error that are automatically retried. If we
+                    //
+                    // Note how this is different in Socket.IO 2.x java client, Socket emits 'connect_error' event.
                     // OBS: Don't get confused with Socket IO 2.x docs online, in 1.x connect_error is emitted also on
                     // errors that are retried by the library automatically!
                     long delay = reconnectBackoff.duration();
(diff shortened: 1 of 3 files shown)


bundles/org.openhab.io.openhabcloud/src/main/java/org/openhab/io/openhabcloud/internal/CloudClient.java
```diff
@@ -174,6 +174,17 @@ public void call(Object... args) {
...
+        }).on(Manager.EVENT_CONNECT_ERROR, new Emitter.Listener() {
+            @Override
+            public void call(Object... args) {
+                // Try reconnecting on connection errors
+                if (logger.isDebugEnabled()) {
+                    logger.error("Error connecting to the openHAB Cloud instance: {}. Reconnecting.", args[0]);
+                } else {
+                    logger.error("Error connecting to the openHAB Cloud instance. Reconnecting.");
+                }
+                socket.close().connect();
```
ssalonen, line 186: To be checked: throttling/delay of retries

digitaldan: I agree, it would be nice to both delay (maybe randomize it) and throttle this call.  One issue we have with the cloud service is the "Thundering Hurd" problem where when we disconnect a large number of users (loose a load balancer, or more likely linode has a network issue) , those  users constantly DDOS us and try to reconnect at the same time.  In the case of a total network outage, this can be 20k-30k clients.

digitaldan:
Thanks for pointing out the lifecycle documentation and addressing this issue!

ssalonen:
On closer look, `Socket.EVENT_CONNECT_ERROR` was named `Socket.EVENT_ERROR` in 1.x, see https://socketio.github.io/socket.io-client-java/migrating_from_1_x.html. Reading the 1.x codebase carefully [1], one can see that with the default retry-parameters, the Socket/Manager `connect_error` events are in fact retried automatically by socket.io library. I have tested to the extent possibly by simulating tcp delays and packet loss.

However, error event in 1.0.1 can be also transmitted even though connection is well established, see e.g. https://github.com/socketio/socket.io-client-java/blob/socket.io-client-1.0.1/src/main/java/io/socket/client/Socket.java#L318-L320

We can see that in 2.0.1, `Socket.EVENT_CONNECT_ERROR` is emitted when Manager emits `error` event and Socket is unconnected:

https://github.com/socketio/socket.io-client-java/blob/socket.io-client-2.0.1/src/main/java/io/socket/client/Socket.java#L91-L98

One way to confirm changed behaviour is to introduce invalid UUID, which is rejected by the server ("not authorized" error). With the new code, client would automatically retry. Old implementation stopped alltogether without retriying. 


[1] `Manager.open()` is first called with `null` `OpenCallback`. Error  events are captured during connection phase by subscribing to engine socket events. Note how this error handler is removed when connection has been successfully subscribed (`onopen` calls `cleanup()`).

We can see how `connect_error` event is emitted in the error handler. Since  `OpenCallback == null`, `maybeReconnectOnOpen` will be called. The conditional passes and `reconnect` method is called. 

In `reconnect` method, `Manager.open` is called again, but now with non-null `OpenCallback` (!), that is, there is a custom error handler for errors occurring during connect. The custom error handler simply re-connects (with no upper bound by default), and thus side-steps the guards set in `maybeReconnectOnOpen`.

One can easily test this by blocking traffic to myopenhab.org, and observing how connection is retried, even though `connect_error` events are emitted.

digitaldan:
Hi @ssalonen is this ready to review and try out ?

ssalonen:
@digitaldan yes, please. No further changes planned at this stage.

Nils-github:
Maybe the review is picked up faster without WIP in the title?

ssalonen:
This should not be merged yet since it requires the other PR. Therefore I am keeping the WIP.

Nils-github:
@ssalonen:
Ok, sorry for my misunderstanding.
Maybe also a "awaiting other PR" label then?

lolodomo:
The awaiting PR is now merged, isn't it?

digitaldan:
thanks @lolodomo  for the reminder, i'll take a look.

digitaldan (approved):
LGTM!  I have been running this for the last week without issues, thanks for taking the time to dig into this, i also think the additional debug logging will be helpful for troubleshooting.

lolodomo (approved):
LGTM

ssalonen:
> i also think the additional debug logging will be helpful for troubleshooting.

Exactly. Not sure if this PR will fix all the issues but hopefully points us to right direction with any open issues.

Btw, I noticed that the socket io client is somewhat old version... I am unsure of these are maintained anymore and if update would be preferable. Unfortunately, this might also imply updating server side...

PR #11160, opened 2021-08-26
[ipcamera] Move to using port 8080 servlet not Netty.

Skinah:
+ Main change has the aim to make Alexa integration easier if all cameras are served from the same port. Setup is now easier as a result. *BREAKING CHANGE*
+ Binding now has the new ability to create snapshots.mjpeg rotating feeds from a camera group that skip cameras that have no motion, or display all cameras when there is no motion on any.
+ Changed how cameras check for online status so they don't poll snapshots using up bandwidth just to check if they are online.
+ Multiple bug fixes found during testing that have been in the binding for a while.

The files will now get served on these URLs and the string channels are updated.
```
http://openhab:8080/ipcamera/<cameraUID>/ipcamera.jpg
http://openhab:8080/ipcamera/<cameraUID>/ipcamera.mjpeg
http://openhab:8080/ipcamera/<cameraUID>/ipcamera.m3u8
http://openhab:8080/ipcamera/<cameraUID>/snapshots.mjpeg
http://openhab:8080/ipcamera/<cameraUID>/autofps.mjpeg
```

Precompiled jar can be downloaded here:
http://pcmus.com/openhab/IpCameraBinding/ipcamera-2021-09-05.zip
NOTE: Installing the `tellstick` binding will take care of any Netty dependency missing warnings.


digitaldan:
FYI i have been using this PR and its working great, my existing items and video elements in the OH3 UI all worked out of the box, so there was no migration work needed after upgrading to this version.

Skinah:
Thanks @digitaldan good to hear it is working well for you and thanks for testing. I have tested in depth and the binding is working better than the merged code on my system here and I have hit the point where I need others to test in ways that I don't think of, or with cameras that I don't have.

This will cause issues for people that have hard coded URLS into rules and widgets. Not a big deal to fix and people can prepare for the change by using the String channels to provide the URLs, then any changes are automatic.

How to get the framework to warn about a breaking change? I have seen other bindings do this when I `apt upgrade` on the command. Also this information seems to be missing from the change logs that get published on the forum. If the build has breaking changes flagged, can it be added to the list of changes automatically so user can get a heads up on these before upgrading?

bundles/org.openhab.binding.ipcamera/src/main/java/org/openhab/binding/ipcamera/internal/servlet/CameraServlet.java
```diff
@@ -0,0 +1,311 @@
...
+                } else {
+                    localFfmpeg.setKeepAlive(8);
+                    sendFile(resp, pathInfo, "application/x-mpegURL");
+                    return;
+                }
+                // Allow files to be created, or you get old m3u8 from the last time this ran.
+                try {
+                    Thread.sleep(HLS_STARTUP_DELAY_MS);
+                } catch (InterruptedException e) {
+                }
```
fwolter, line 135: `doGet()` should be ended as quickly as possible when an `InterruptedException` is thrown.

bundles/org.openhab.binding.ipcamera/src/main/java/org/openhab/binding/ipcamera/internal/servlet/CameraServlet.java
```diff
@@ -0,0 +1,311 @@
...
+                        // Never stop streaming until IOException. Occurs when browser stops the stream.
+                        snapshotStreamsOpen--;
+                        if (snapshotStreamsOpen == 0) {
+                            handler.streamingSnapshotMjpeg = false;
+                            handler.stopSnapshotPolling();
+                            logger.debug("All snapshots.mjpeg streams have stopped.");
+                        }
+                        return;
+                    }
+                } while (true);
```
fwolter, line 167: If I understand correctly, this runs as long as the user has opened the browser. So, this will occupy a thread of the HTTP server. Are you aware from which thread pool you get the thread from? Same for the others.

bundles/org.openhab.binding.ipcamera/src/main/java/org/openhab/binding/ipcamera/internal/servlet/CameraServlet.java
```diff
@@ -0,0 +1,311 @@
...
+                } else if (pathInfo.contains(".m4s") || pathInfo.contains(".mp4")) {
+                    sendFile(resp, pathInfo, "video/mp4");
+                }
+                return;
+        }
+    }
+
+    private void sendFile(HttpServletResponse response, @Nullable String fileUri, String contentType)
+            throws IOException {
+        File file = new File(handler.cameraConfig.getFfmpegOutput() + fileUri);
```
fwolter, line 253: What happens if the user enters some `../../` in the path. Can he/she read the whole filesystem?

bundles/org.openhab.binding.ipcamera/src/main/java/org/openhab/binding/ipcamera/internal/servlet/GroupServlet.java
```diff
@@ -0,0 +1,210 @@
...
+            case "/ipcamera.m3u8":
+                if (!groupHandler.hlsTurnedOn) {
+                    logger.debug(
+                            "HLS requires the groups startStream channel to be turned on first. Just starting it now.");
+                    String channelPrefix = "ipcamera:" + groupHandler.getThing().getThingTypeUID() + ":"
+                            + groupHandler.getThing().getUID().getId() + ":";
+                    groupHandler.handleCommand(new ChannelUID(channelPrefix + CHANNEL_START_STREAM), OnOffType.ON);
+                    try {
+                        TimeUnit.MILLISECONDS.sleep(HLS_STARTUP_DELAY_MS);
+                    } catch (InterruptedException e) {
```
fwolter, line 94: See above. Throw `InterruptedException`.

bundles/org.openhab.binding.ipcamera/src/main/java/org/openhab/binding/ipcamera/internal/servlet/GroupServlet.java
```diff
@@ -0,0 +1,210 @@
...
+import org.slf4j.LoggerFactory;
+
+/**
+ * The {@link GroupServlet} is responsible for serving files for a rotating feed of multiple cameras back to the Jetty
+ * server normally found on port 8080
+ *
+ * @author Matthew Skinner - Initial contribution
+ */
+@NonNullByDefault
+public class GroupServlet extends HttpServlet {
```
fwolter, line 47: There seems to be a lot of redundant code in CameraServlet and GroupServlet. Did you think about making an abstract class with the common code?

fwolter (changes requested):

fwolter:
> How to get the framework to warn about a breaking change? I have seen other bindings do this when I `apt upgrade` on the command. Also this information seems to be missing from the change logs that get published on the forum. If the build has breaking changes flagged, can it be added to the list of changes automatically so user can get a heads up on these before upgrading?

The breaking changes printed when running `apt upgrade` are located here https://github.com/openhab/openhab-distro/blob/main/distributions/openhab/src/main/resources/bin/update.lst. I'm sure it would be beneficial if this could be automated further.

bundles/org.openhab.binding.ipcamera/src/main/java/org/openhab/binding/ipcamera/internal/servlet/CameraServlet.java
Skinah: It will keep the thread until the stream is not accepted by a browser. You can press stop on a browser to halt it or close the tab. Also if someone has lazy loading on a widget setup it can open and stay open for a long time even when navigated away from the page that has the widget. To stop that it requires the refresh to be pressed when not on the widgets page.

I do not know where the threads are coming from only that each request will get its own dedicated thread hence why I feel it is ok to use thread sleep. Thank you for looking at this as it is not my strong point.

bundles/org.openhab.binding.ipcamera/src/main/java/org/openhab/binding/ipcamera/internal/servlet/CameraServlet.java
Skinah: No only the contents of 1 folder and it will not go into sub folders. Even then it only serves a few file extensions like *.ts and *.mp4 which are necessary for HLS and any recordings done by the binding. The rest must be exact matches. So only media files and a call to fetch password.txt in that folder would fail.

bundles/org.openhab.binding.ipcamera/src/main/java/org/openhab/binding/ipcamera/internal/servlet/CameraServlet.java
Skinah: Done

bundles/org.openhab.binding.ipcamera/src/main/java/org/openhab/binding/ipcamera/internal/servlet/GroupServlet.java
Skinah: Done

bundles/org.openhab.binding.ipcamera/src/main/java/org/openhab/binding/ipcamera/internal/servlet/GroupServlet.java
Skinah: DONE: I made the change and it saved 15 lines of code.

bundles/org.openhab.binding.ipcamera/src/main/java/org/openhab/binding/ipcamera/internal/servlet/CameraServlet.java
fwolter: If I see correctly, it will serve any file in the whole filesystem which name contains one of the extensions in the code. What would prevent the code to go into subfolders or parent directories via `../`?

Skinah: Thank you for raising this, as I was not aware you could mix ABSOLUTE and RELATIVE stuff like that. It appears to not be an issue when tested in 3 browsers. Safari, Edge and Chrome were resolved by the browser before the GET request is made, could another browser handle it differently? I also tried wget and curl on the linux terminal and they both also failed to act differently.

In all 5 cases I tested, entering `http://openhab:8080/ipcamera/mycam/../ipcamera51.ts` into the address bar simply resolves to a get request being sent to `http://openhab:8080/ipcamera/ipcamera51.ts`
The servlet is not listening to requests at that url, so it gets a generic openhab `Requested content not found`.
I guess it will return a file IF someone had a file named as that in the static folder, but in that case you can already request the file in the html/static folder as that is its intention.

The issue I see is that with low latency HLS you can be making file requests every 250ms or even faster depending on how far you wish to push things in the search for low latency video feeds. So I would be happy to insert more code on the other file formats, but I am reluctant to bloat the calls to the .ts format for this reason unless you advise it should be done?

One thought would be to..

String truncated = pathInfo.substring(pathInfo.lastIndexOf("/"));

This way it throws away anything before the last "/" in the requested URL, and will not go anywhere other than a single folder which is fine with the binding. Is the overhead cost worth doing this? If yes then I am happy to do it I just need someone with more knowledge than me to make the call.

a) Do it to all but .ts files.
b) Do to all files.
c) its not an issue, leave as is.

bundles/org.openhab.binding.ipcamera/src/main/java/org/openhab/binding/ipcamera/internal/servlet/CameraServlet.java
fwolter: It will probably consume a thread from the Jetty thread pool, which defaults to 50. https://github.com/openhab/openhab-distro/blob/9bdedaa6f537b8b31bfd4082ee77f5a9401f19b2/distributions/openhab/src/main/resources/userdata/etc/org.ops4j.pax.web.cfg#L9

I think that's ok.

bundles/org.openhab.binding.ipcamera/src/main/java/org/openhab/binding/ipcamera/internal/servlet/CameraServlet.java
fwolter: Better not rely on the client resolving the URL to a safe URL. I don't know if Jetty has some internal mechanism to resolve the URL before it is handed over to your servlet. I'd go for the substring call to limit the path to a file instead of an entire path. I don't see any performance issues invoking this operation. I'd apply it in `sendFile()` to be sure future enhancements don't open some door.

Skinah: DONE. It is changed so it should not be possible to reach parent or child folders.

fwolter (approved):
LGTM

Issue #11152, opened 2021-08-23
[miio] Miio binding shuts down shared executor service


janvyb:
Miio binding sometimes shuts down shared "thingHandler" scheduler when deleting a thing. All other plugins that use it will then get `java.util.concurrent.RejectedExecutionException` when they try to schedule new tasks, which will make the entire openHAB installation unusable until it is restarted.

## Expected Behavior

Shared executors must not be disposed by binding.

## Current Behavior

Binding sometimes disposes shared executor.

## Possible Solution

The problem is in [`org.openhab.binding.miio.internal.handler.MiIoAbstractHandler`](https://github.com/openhab/openhab-addons/blob/main/bundles/org.openhab.binding.miio/src/main/java/org/openhab/binding/miio/internal/handler/MiIoAbstractHandler.java), specifically in initialization of miioScheduler. It is initialized as:
```java
protected ScheduledExecutorService miIoScheduler = scheduler;
```
where scheduler is property of `BaseThingHandler`:
```java
protected final ScheduledExecutorService scheduler = ThreadPoolManager.getScheduledPool("thingHandler");
```
This scheduler is then overwritten during `initialize()`:
```java
ScheduledThreadPoolExecutor miIoScheduler = new ScheduledThreadPoolExecutor(3,
                new NamedThreadFactory(getThing().getUID().getAsString(), true));
        miIoScheduler.setExecuteExistingDelayedTasksAfterShutdownPolicy(false);
        miIoScheduler.setRemoveOnCancelPolicy(true);
        this.miIoScheduler = miIoScheduler;
```
and shut down in `dispose()` method:
```java
logger.debug("Disposing Xiaomi Mi IO handler '{}'", getThing().getUID());
miIoScheduler.shutdown();
```

The problem is that if thing was disabled when openHAB started, `MiIoAbstractHandler` is created but `initialize()` method is never called. Then if you delete the thing from UI, openHAB calls `dispose()` method and `MiIoAbstractHandler` will call shutdown on `miIoScheduler`, but it points to shared pool instead of the one created by binding.

## Steps to Reproduce (for Bugs)

1. Discover any miio thing in UI
2. Disable that thing
3. Restart openHAB
4. Delete the thing from UI

PR #8981, opened 2020-11-07
[miio]  Cloud Communication for devices

marcelrv:
Allows to define if communication to devices is direct or send via the
Xiaomi cloud. (#7276)
Introduce additional channel to execute commands via cloud.

Other small improvements
* Use common method from abstract handler to send commands
* Common way to handle custom commands
* Introduce small delay before refreshing robot properties after sending
commands (similar to the basic handler) so devices have time to update
their properties

Signed-off-by: Marcel Verpaalen <marcel@verpaalen.com>

PR #11166, opened 2021-08-27
[miio] Avoid shutting down shared executor service

marcelrv:
* Closing #11152
* Better identify communication threads


kaikreuzer:
Could you just briefly explain me, why keeping the standard scheduler and simply not shutting it down is not an option here?

marcelrv:
This was introduced as part of the review feedback here

https://github.com/openhab/openhab-addons/pull/8981#pullrequestreview-526679368

kaikreuzer:
Thanks for your quick answer. I don't fully understand the linked discussion and I am not in the details of what async jobs the miio binding launches and what it needs to correctly interrupt/cancel them. So if the current solution works for you, let's go for it (even when we should avoid adding further threadpools for different bindings, because the number of threads of a running system is already terrifyingly large...).

kaikreuzer (approved):

PR #11171, opened 2021-08-29
[pulseaudio] Fix playing time with pulseaudio sink (#11170)

### LINKED ISSUE #11170: [pulseaudio] Restoring volume after the "play" action is not timely accurate
## Expected Behavior
When using the sound volume parameter in the play sound action, with the pulseaudio binding as a sink, the sound volume should return to its previous state at the right time (i.e. when the play action is finished, not before or after)

## Current Behavior
The pulseaudio binding return to openhab (thus implying the end of the play action) when the raw audio data has been transmitted to the pulseaudio server. But the pulseaudio server may accept data faster than it plays it.

## Possible Solution
Estimate the playing time by parsing the input stream, if possible, then introduce a intelligent wait (Thread.sleep) before ending the play method in the pulseaudio binding.
I have done the change described here, and will make a pull request soon.

1. Set a pulseaudio audio sink in OpenHAB
2. Play a 10 seconds sound, with the play action and the volume property set. For properly spotting the bug, the sound must NOT fade at the end.

## Context
Bugs reported [from the forum](https://community.openhab.org/t/pulse-audio-binding/74265/46).

dalgwen:
Fixes #11170 by introducing an intelligent Thread.sleep (getting the duration of the sound if possible, then wait the appropriate time for letting the sound play).

Also fix a minor issue with the last volume not propertly saved.

And fix some minor warnings by using final local variable.

Signed-off-by: Gwendal ROULLEAU <gwendal.roulleau@gmail.com>

[Link to jar](https://github.com/dalgwen/openhab-addons/releases/tag/3.2.0-SNAPSHOT_pulseaudio_11170)

### DIFF
--- a/bundles/org.openhab.binding.pulseaudio/src/main/java/org/openhab/binding/pulseaudio/internal/PulseAudioAudioSink.java
+++ b/bundles/org.openhab.binding.pulseaudio/src/main/java/org/openhab/binding/pulseaudio/internal/PulseAudioAudioSink.java
@@ -15,15 +15,20 @@ package org.openhab.binding.pulseaudio.internal;
 import java.io.IOException;
 import java.io.InputStream;
 import java.net.Socket;
+import java.time.Duration;
+import java.time.Instant;
 import java.util.HashSet;
 import java.util.Locale;
+import java.util.Map;
 import java.util.Set;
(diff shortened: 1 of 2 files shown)


kaikreuzer (approved):
lgtm, thanks!

PR #11178, opened 2021-08-31
[icalendar] Fix internal calculation for retrieving events for command tags

### LINKED ISSUE #11084: [iCalendar] Single change to repeating calendar entry generates 2 events.
Situation:
There is a daily event generated from the google calendar (eg: good_morning_routine at 7.00).
One day you want to sleep in and move the item in google calendar to 8.00.

## Expected Behavior
The good_morning_routine event gets triggered that day at 8.00 only.

## Current Behavior
The good_morning_routine event gets triggered that day at 7.01 and at 8.00 only.

I reproduced this a few times like this:

In google calendar create repeating item in the past (to make it possible to do the test now):
```
Title: TEST_REPEATING_EVENT
Freq: Daily 14:10 – 14:15
Comment: BEGIN:E_Test_Cal:ON

```
Change the repeating event in google Calendar of today to 14:20 – 14:25  and "Save only this occurrence"
The repeating event of today is a little later than the other days:
image

After a few minutes (refresh is 3 min) the icaldar items are updated as expected: (Today the event should start at 14:20)
image

In the log the event is only listed at 14:20 as expected:
```
2021-08-04 14:20:00.583 [INFO ] [openhab.event.ItemStateChangedEvent ] - Item 'C_Current_Event_Name' changed from K_C to TEST_REPEATING_EVENT
2021-08-04 14:25:00.354 [INFO ] [openhab.event.ItemStateChangedEvent ] - Item 'C_Current_Event_Name' changed from TEST_REPEATING_EVENT to K_C
```
Un the rules logging however you can see that the rule is triggerd twice:

```
2021-08-04 14:11:05.002 [INFO ] [RULES] - R_Test_Cal
2021-08-04 14:20:00.672 [INFO ] [RULES] - R_Test_Cal

```
Once at the general repeating time of the event and once at the corrected time of the event.
(I did this test a few times and I noticed the repeating trigger is always about a minute after the repeating start time)

## Your Environment
* Version used: openHAB 3.1.0 release build / iCalendar Binding 3.1.0
* Environment name and version: Windows 10 - Azul Java 11.0.11

daMihe:
Had also to update the dependencies as the binding was not resolvable in previous state. Reduced dependencies in that case, but jackson was pinned to a specific version by my external dependency.

This PR fixes #11084.


kaikreuzer:
> was not resolvable in previous state.

Could you elaborate on this? I just tried installing the icalendar binding with the latest 3.2 snapshot and it correctly resolved and started:
```
312 │ Active │  80 │ 3.2.0.202109080355    │ openHAB Add-ons :: Bundles :: iCalendar Binding
```
If you definitely need other versions that those that are already available on our Artifactory (since other core components depend on them), you might rather want to embed them in your bundle directly.

daMihe:
I found this out while building the binding using `mvn clean install -pl :org.openhab.binding.icalendar`:
```
[ERROR] Failed to execute goal org.apache.karaf.tooling:karaf-maven-plugin:4.3.2:verify (karaf-feature-verification) on project org.openhab.binding.icalendar: Feature resolution failed for [openhab-binding-icalendar/3.2.0.SNAPSHOT]
[ERROR] Message: Unable to resolve root: missing requirement [root] osgi.identity; osgi.identity=openhab-binding-icalendar; type=karaf.feature; version=3.2.0.SNAPSHOT; filter:="(&(osgi.identity=openhab-binding-icalendar)(type=karaf.feature)(version>=3.2.0.SNAPSHOT))" [caused by: Unable to resolve openhab-binding-icalendar/3.2.0.SNAPSHOT: missing requirement [openhab-binding-icalendar/3.2.0.SNAPSHOT] osgi.identity; osgi.identity=org.openhab.binding.icalendar; type=osgi.bundle; version="[3.2.0.202109082159,3.2.0.202109082159]"; resolution:=mandatory [caused by: Unable to resolve org.openhab.binding.icalendar/3.2.0.202109082159: missing requirement [org.openhab.binding.icalendar/3.2.0.202109082159] osgi.wiring.package; filter:="(&(osgi.wiring.package=biweekly)(version>=0.6.0)(!(version>=1.0.0)))"]]
```

If i interpret this correctly, the `biweekly`-package can't be resolved using `${jackson.version}`. Changing the version of jackson exactly to the requirement of biweekly, the build works. I assume the POM of biweekly specifies the version too exact for a successful resolution:

```
<dependency>
	<groupId>com.fasterxml.jackson.core</groupId>
	<artifactId>jackson-core</artifactId>
	<version>2.12.1</version>
</dependency>
```

Embedding the components is done using `<scope>compile</scope>` as done currently, right?

kaikreuzer:
> If i interpret this correctly

I guess you don't.
My assumption is that the karaf feature verification does not always correctly work, if a single project is built (since the Karaf feature is at the end assembled to a single "addons" feature file).

If you run the project from its root, you will see that the build succeeds, so there does not seem to be any issue with the version of the dependencies. Also, as mentioned above, the binding correctly installs on the snapshot build, so that proves that there is no issue with the resolution.

daMihe:
> If you run the project from its root, you will see that the build succeeds, so there does not seem to be any issue with the version of the dependencies. Also, as mentioned above, the binding correctly installs on the snapshot build, so that proves that there is no issue with the resolution.

Yes it does, so i changed back to the version variable. Is there a good way to build just one addon for verifying if the build will work? Maybe my invocation is just wrong?

kaikreuzer:
> Is there a good way to build just one addon for verifying if the build will work? Maybe my invocation is just wrong?

I am not aware of any, if there add-on has dependencies like this one, sorry. So a full build is the only way to now whether it's fine.

kaikreuzer (approved):
lgtm, thanks!

daMihe:
> I am not aware of any, if there add-on has dependencies like this one, sorry. So a full build is the only way to now whether it's fine.

Okay, thank you for clarification and review.

PR #11179, opened 2021-08-31
[bluetooth.bluegiga] Fix CompareObjectsWithEquals findings regression causing all devices to be beacons

jmelhus:


wborn (approved):
Sorry for having introduced this regression and thanks for the fix!

PR #11181, opened 2021-08-31
[kodi] Add channel for KODI JSON-RPC call Input.ButtonEvent

thumm:
Extend KODI addon to support ButtonEvent RPC introduced with KODI v12 Matrix (see [KODI JSON-RPC API](https://kodi.wiki/view/JSON-RPC_API/v12#Input.ButtonEvent)).

With KODI v12 Matrix a new RPC call Input.ButtonEvent has been introduced. With this call an arbitrary button press event can be send to KODI.

The pull requests implements an additional channel in the KODI addon to support this new RPC call and make it available in Openhab. Any kind of button from a remote control can be sent via Openhab to KODI.


kaikreuzer (approved):
lgtm, thanks!

Issue #4756, opened 2019-01-29
[homematic] Change Channel Event Implementation BUTTON DOUBLE_PRESSED


lukicsl:
## Expected Behavior
Pressing a button twice (or multiple) within a threshold  period (e.g. 0.3s) should generate an BUTTON DOUBLE_PRESSED event, no other events should be triggered.
It is either an SHORT or DOUBLE event not both.

## Current Behavior
First press of button generates a BUTTON SHORT_PRESSED event, if within 2s a second short press happens a BUTTON DOUBLE_PRESSED event is generated additionally.

## Possible Solution
Create a global  HM parameter (Bridge?) for double click threshold.
Suppress BUTTON SHORT_PRESSED for this threshold if within this time another short press event happens, create a BUTTON DOUBLE_PRESSED otherwise issue the held back event.

## Context
Double press event are rathe fast events and should be used for special (overridden) function of a button. In the current implementation this event is useless as the single click event comes first. 
Implementing this filtering threshold in a rule would be possible but rather complicated.

## Your Environment
OH 2.4.0 docker

rancor187:
True, I experienced the same issue.

hmerk:
closed due to inactivity

PR #11186, opened 2021-09-01
[homematic] Remove double press events and improve long press events for button trigger

maniac103:
HM devices provide not only 'long press' events, but also 'long
press continued' (sent in configured long press interval) and 'long
press released' events. So far, those events were swallowed in the
button datapoint handler.
Improve the situation by forwarding those events to the button trigger
channel, making them usable in e.g. rules that react on button long
presses.

### DIFF
--- a/bundles/org.openhab.binding.homematic/src/main/java/org/openhab/binding/homematic/internal/communicator/virtual/ButtonVirtualDatapointHandler.java
+++ b/bundles/org.openhab.binding.homematic/src/main/java/org/openhab/binding/homematic/internal/communicator/virtual/ButtonVirtualDatapointHandler.java
@@ -33,6 +33,9 @@ import org.slf4j.LoggerFactory;
 public class ButtonVirtualDatapointHandler extends AbstractVirtualDatapointHandler {
     private final Logger logger = LoggerFactory.getLogger(ButtonVirtualDatapointHandler.class);
 
+    private static final String LONG_REPEATED_EVENT = "LONG_REPEATED";
+    private static final String LONG_RELEASED_EVENT = "LONG_RELEASED";
+
     @Override
     public String getName() {
         return VIRTUAL_DATAPOINT_NAME_BUTTON;
@@ -45,7 +48,7 @@ public class ButtonVirtualDatapointHandler extends AbstractVirtualDatapointHandl
                 HmDatapoint dp = addDatapoint(device, channel.getNumber(), getName(), HmValueType.STRING, null, false);
                 dp.setTrigger(true);
                 dp.setOptions(new String[] { CommonTriggerEvents.SHORT_PRESSED, CommonTriggerEvents.LONG_PRESSED,
-                        CommonTriggerEvents.DOUBLE_PRESSED });
+                        LONG_REPEATED_EVENT, LONG_RELEASED_EVENT, CommonTriggerEvents.DOUBLE_PRESSED });
             }
         }
     }
@@ -57,10 +60,13 @@ public class ButtonVirtualDatapointHandler extends AbstractVirtualDatapointHandl
 
     @Override
     public void handleEvent(VirtualGateway gateway, HmDatapoint dp) {
-        HmDatapoint vdp = getVirtualDatapoint(dp.getChannel());
+        HmChannel channel = dp.getChannel();
+        HmDatapoint vdp = getVirtualDatapoint(channel);
+        int usPos = dp.getName().indexOf("_");
+        String pressType = usPos == -1 ? dp.getName() : dp.getName().substring(usPos + 1);
+        boolean isLongPressActive = CommonTriggerEvents.LONG_PRESSED.equals(vdp.getValue())
+                || LONG_REPEATED_EVENT.equals(vdp.getValue());
         if (MiscUtils.isTrueValue(dp.getValue())) {
-            int usPos = dp.getName().indexOf("_");
-            String pressType = usPos == -1 ? dp.getName() : dp.getName().substring(usPos + 1);
             switch (pressType) {
                 case "SHORT":
                     if (dp.getValue() == null || !dp.getValue().equals(dp.getPreviousValue())) {
@@ -72,18 +78,45 @@ public class ButtonVirtualDatapointHandler extends AbstractVirtualDatapointHandl
                     }
                     break;
                 case "LONG":
-                    vdp.setValue(CommonTriggerEvents.LONG_PRESSED);
+                    if (LONG_REPEATED_EVENT.equals(vdp.getValue())) {
+                        // Suppress long press events during an ongoing long press
+                        vdp.setValue(LONG_REPEATED_EVENT);
+                    } else {
+                        vdp.setValue(CommonTriggerEvents.LONG_PRESSED);
+                    }
                     break;
                 case "LONG_RELEASE":
+                    // Only send release events if we sent a pressed event before
+                    vdp.setValue(isLongPressActive ? LONG_RELEASED_EVENT : null);
+                    break;
                 case "CONT":
+                    // Clear previous value to force re-triggering of repetition
                     vdp.setValue(null);
+                    // Only send repetitions if there was a pressed event before
+                    // (a CONT might arrive simultaneously with the initial LONG event)
+                    if (isLongPressActive) {
+                        vdp.setValue(LONG_REPEATED_EVENT);
+                    }
                     break;
                 default:
                     vdp.setValue(null);
                     logger.warn("Unexpected vaule '{}' for PRESS virtual datapoint", pressType);
             }
         } else {
-            vdp.setValue(null);
+            if ("LONG".equals(pressType) && LONG_REPEATED_EVENT.equals(vdp.getValue())) {
+                // If we're currently processing a repeated long-press event, don't let the initial LONG
+                // event time out the repetitions, the CONT delay handler will take care of it
+                vdp.setValue(LONG_REPEATED_EVENT);
+            } else if (isLongPressActive) {
+                // We seemingly missed an event (either a CONT or the final LONG_RELEASE),
+                // so end the long press cycle now
+                vdp.setValue(LONG_RELEASED_EVENT);
+            } else {
+                vdp.setValue(null);
+            }
         }
+        logger.debug("Handled virtual button event on {}:{}: press type {}, value {}, button state {} -> {}",
+                channel.getDevice().getAddress(), channel.getNumber(), pressType, dp.getValue(), vdp.getPreviousValue(),
+                vdp.getValue());
     }
 }
--- a/bundles/org.openhab.binding.homematic/src/main/java/org/openhab/binding/homematic/internal/communicator/virtual/ButtonVirtualDatapointHandler.java
+++ b/bundles/org.openhab.binding.homematic/src/main/java/org/openhab/binding/homematic/internal/communicator/virtual/ButtonVirtualDatapointHandler.java
@@ -48,7 +48,7 @@ public class ButtonVirtualDatapointHandler extends AbstractVirtualDatapointHandl
                 HmDatapoint dp = addDatapoint(device, channel.getNumber(), getName(), HmValueType.STRING, null, false);
                 dp.setTrigger(true);
                 dp.setOptions(new String[] { CommonTriggerEvents.SHORT_PRESSED, CommonTriggerEvents.LONG_PRESSED,
-                        LONG_REPEATED_EVENT, LONG_RELEASED_EVENT, CommonTriggerEvents.DOUBLE_PRESSED });
+                        LONG_REPEATED_EVENT, LONG_RELEASED_EVENT });
             }
         }
     }
@@ -68,15 +68,11 @@ public class ButtonVirtualDatapointHandler extends AbstractVirtualDatapointHandl
                 || LONG_REPEATED_EVENT.equals(vdp.getValue());
         if (MiscUtils.isTrueValue(dp.getValue())) {
             switch (pressType) {
-                case "SHORT":
-                    if (dp.getValue() == null || !dp.getValue().equals(dp.getPreviousValue())) {
-                        vdp.setValue(CommonTriggerEvents.SHORT_PRESSED);
-                    } else {
-                        // two (or more) PRESS_SHORT events were received
-                        // within AbstractHomematicGateway#DEFAULT_DISABLE_DELAY seconds
-                        vdp.setValue(CommonTriggerEvents.DOUBLE_PRESSED);
-                    }
+                case "SHORT": {
+                    vdp.setValue(null); // Force sending new event
+                    vdp.setValue(CommonTriggerEvents.SHORT_PRESSED);
                     break;
+                }
                 case "LONG":
                     if (LONG_REPEATED_EVENT.equals(vdp.getValue())) {
                         // Suppress long press events during an ongoing long press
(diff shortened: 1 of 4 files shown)


maniac103:
Added another commit that fixes up the double press handling and re-enables the double press unit test. There still is a slight potential for failure (if processing two events takes more than 400ms), but that seems rather unlikely. Fixing that would require static mocking (of either `Instant.now()` or `Duration.between()`), but the used mocker doesn't seem to support that. Suggestions welcome.

maniac103:
@MHerbst Mind giving this a review? :-)

bundles/org.openhab.binding.homematic/src/main/java/org/openhab/binding/homematic/internal/communicator/virtual/ButtonVirtualDatapointHandler.java
```diff
@@ -57,33 +61,64 @@ public boolean canHandleEvent(HmDatapoint dp) {
 
     @Override
     public void handleEvent(VirtualGateway gateway, HmDatapoint dp) {
```
MHerbst, line 63: You have made major changes to the logic here. Is the behavior regarding the triggered events really like before? It would be bad if rules behaved differently than before, because the recognition of a double click or long press works differently.

I don't have any buttons in use right now, otherwise I could do some regression tests.

bundles/org.openhab.binding.homematic/src/main/java/org/openhab/binding/homematic/internal/handler/HomematicThingHandler.java
```diff
@@ -384,8 +384,10 @@ protected void updateDatapointState(HmDatapoint dp) {
     private void updateChannelState(final HmDatapoint dp, Channel channel)
             throws IOException, GatewayNotAvailableException, ConverterException {
         if (dp.isTrigger()) {
-            if (dp.getValue() != null) {
-                triggerChannel(channel.getUID(), dp.getValue() == null ? "" : dp.getValue().toString());
+            final Object value = dp.getValue();
+            final Object prevValue = dp.getPreviousValue();
```
MHerbst, line 388: A separate varaible vor prevValue is not necessary, because you only use in the next statement.

bundles/org.openhab.binding.homematic/src/test/java/org/openhab/binding/homematic/internal/communicator/virtual/ButtonDatapointTest.java
```diff
@@ -16,9 +16,11 @@
 import static org.hamcrest.MatcherAssert.assertThat;
```
MHerbst, line 16: This code should better be reviewed by someone who knows it better. I'm not 100% sure if the double click can really be tested cleanly automatically, because the timing is critical.

bundles/org.openhab.binding.homematic/src/main/java/org/openhab/binding/homematic/internal/communicator/virtual/ButtonVirtualDatapointHandler.java
maniac103: I hope so ;-) The goal is SINGLE_PRESS being triggered as before (just with shorter spacing due to the changed DOUBLE_PRESS handling), LONG_PRESS being triggered as before and just the new 2 events being triggered in addition to that.

maniac103:
After thinking about it a bit more, I think that the `DOUBLE_PRESSED` handling should be removed. Rationale for that would be
- Since the double press timeout effectively is 2 seconds, the `DOUBLE_PRESSED` event prevents processing `SINGLE_PRESSED` events in quick succession. This may be subjective, but I think presses more than 1 second apart should be handled as 2 single presses, not a single double press. A quick googling for `openhab homematic double_pressed' suggests I'm not the only one thinking so: there are [confused](https://homematic-forum.de/forum/viewtopic.php?t=51242) [people](https://community.openhab.org/t/solved-disable-button-double-pressed-channel-event-homematic/65979), and some [work around the situation](https://forum.mods.de/bb/thread.php?TID=216051&PID=1248807098#reply_1248807098). This point is also what made me wanting to change that behavior.
- The double press handling is not a real double press handling anyway, since the first press event is sent as SINGLE_PRESS. One can not meaningfully distinguish 'single press' and 'double press' because the former is always part of the latter.
- As we have noticed, the double press handling can not meaningfully be tested.
- If needed, the double press handling can easily be restored in a rule by using a timer.

@mdicke2s Since you were the one who introduced the double press logic in https://github.com/eclipse-archived/smarthome/pull/6587 ... could you please comment on whether I am missing something here or whether you would be OK with removing the double press handling?

bundles/org.openhab.binding.homematic/src/test/java/org/openhab/binding/homematic/internal/communicator/virtual/ButtonDatapointTest.java
maniac103: Removed the double press handling (and respective unit test) for the reasons outlined [below](https://github.com/openhab/openhab-addons/pull/11186#issuecomment-921548347)

bundles/org.openhab.binding.homematic/src/main/java/org/openhab/binding/homematic/internal/communicator/virtual/ButtonVirtualDatapointHandler.java
maniac103: With the removal of DOUBLE_PRESSED events, the goal is now
- triggering SHORT_PRESSED events on *every* short button press (before it was either SHORT_PRESSED or DOUBLE_PRESSED, depending on timing)
- triggering LONG_PRESSED events on every long press start (as before)
- triggering LONG_REPEATED events on long press repetition (CONT being sent from device; new, didn't exist before)
- triggering LONG_RELEASED events on long press end (new, didn't exist at all before)

MHerbst:
I doubt anyone has actually used `DOUBLE_PRESS`, but wouldn't bet on it.

> * Since the double press timeout effectively is 2 seconds, the `DOUBLE_PRESSED` event prevents processing `SINGLE_PRESSED` events in quick succession. This may be subjective, but I think presses more than 1 second apart should be handled as 2 single presses, not a single double press. A quick googling for 'openhab homematic double_pressed' suggests I'm not the only one thinking so: there are [confused](https://homematic-forum.de/forum/viewtopic.php?t=51242) [people](https://community.openhab.org/t/solved-disable-button-double-pressed-channel-event-homematic/65979), and some [work around the situation](https://forum.mods.de/bb/thread.php?TID=216051&PID=1248807098#reply_1248807098). This point is also what made me want to change that behavior.

I absolutely agree, 2s is far too long. In other environments, I think the timespan between two clicks must be less than 500ms to detect it as double clcik.

> * The double press handling is not a real double press handling anyway, since the first press event is sent as SINGLE_PRESS. One can not meaningfully distinguish 'single press' and 'double press' because the former is always part of the latter. -> [[homematic] Change Channel Event Implementation BUTTON DOUBLE_PRESSED #4756](https://github.com/openhab/openhab-addons/issues/4756)

I am not sure whether the SINGLE_PRESS is really sent. If a SHORT_PRESS event is received the binding waits 2s for another press.
By removing the DOUBLE_PRESS detection this delayed processing can also be removed.

> * If needed, the double press handling can easily be restored in a rule by using a timer.

The DOUBLE_PRESS is documented, so you would also have to change the README.md and a hint should be added to the OH release notes to inform the user about this (potentially) breaking change.

maniac103:
> I am not sure whether the SINGLE_PRESS is really sent. If a SHORT_PRESS event is received the binding waits 2s for another press.
> By removing the DOUBLE_PRESS detection this delayed processing can also be removed.

Nope ;-) Just double pressed one of my buttons:
openhab.log
```
2021-09-17 12:45:37.453 [DEBUG] [ommunicator.AbstractHomematicGateway] - Received new (Integer) value '-49' for 'DBPBI00002:0#RSSI_DEVICE' from gateway with id '5cb6fc84'
2021-09-17 12:45:37.511 [DEBUG] [ommunicator.AbstractHomematicGateway] - Received new (Boolean) value 'true' for 'DBPBI00002:6#INSTALL_TEST' from gateway with id '5cb6fc84'
2021-09-17 12:45:37.513 [DEBUG] [ommunicator.AbstractHomematicGateway] - Received new (Boolean) value 'true' for 'DBPBI00002:6#PRESS_SHORT' from gateway with id '5cb6fc84'
2021-09-17 12:45:37.514 [DEBUG] [ematic.internal.misc.DelayedExecuter] - Delaying event for 2.0 seconds: 'DBPBI00002:6#PRESS_SHORT'
2021-09-17 12:45:38.823 [DEBUG] [ommunicator.AbstractHomematicGateway] - Received new (Boolean) value 'true' for 'DBPBI00002:6#INSTALL_TEST' from gateway with id '5cb6fc84'
2021-09-17 12:45:38.824 [DEBUG] [ommunicator.AbstractHomematicGateway] - Received new (Boolean) value 'true' for 'DBPBI00002:6#PRESS_SHORT' from gateway with id '5cb6fc84'
2021-09-17 12:45:38.825 [DEBUG] [ematic.internal.misc.DelayedExecuter] - Delaying event for 2.0 seconds: 'DBPBI00002:6#PRESS_SHORT'
2021-09-17 12:45:40.827 [DEBUG] [ematic.internal.misc.DelayedExecuter] - Executing delayed event for 'DBPBI00002:6#PRESS_SHORT'
```

events.log
```
2021-09-17 12:45:37.515 [INFO ] [openhab.event.ChannelTriggeredEvent ] - homematic:HG-HB-PBI-8-FM:5cb6fc84:DBPBI00002:6#BUTTON triggered SHORT_PRESSED
2021-09-17 12:45:38.826 [INFO ] [openhab.event.ChannelTriggeredEvent ] - homematic:HG-HB-PBI-8-FM:5cb6fc84:DBPBI00002:6#BUTTON triggered DOUBLE_PRESSED
```

You see that the first SHORT_PRESSED is sent *immediately* after receiving the PRESS_SHORT event.
You may have confused that due to the (very confusing) log entry 'Executing delayed event...' (here at 12:45:40.827) ... that handling sets the PRESS datapoint back to false, so is unrelated to short/double press  processing.

> The DOUBLE_PRESS is documented, so you would also have to change the README.md and a hint should be added to the OH release notes to inform the user about this (potentially) breaking change.

I already updated README.md. I also agree to the release notes hint, but am unsure how to ensure that happens?

mdicke2s:
@maniac103 I am OK with removing the DOUBLE_PRESS detection. When I implemented this, I must have had a wrong understanding of the actual device's behaviour.

MHerbst:
> You may have confused that due to the (very confusing) log entry 'Executing delayed event...' (here at 12:45:40.827) ... that handling sets the PRESS datapoint back to false, so is unrelated to short/double press processing.

Yes, I only saw this message but never had a closer look into the code :-).

> I also agree to the release notes hint, but am unsure how to ensure that happens?

As far as I know you have to create a PR in openhab/openhab-distro for the file:distributions/openhab/src/main/resources/bin/update.lst

maniac103:
> As far as I know you have to create a PR in openhab/openhab-distro for the file:distributions/openhab/src/main/resources/bin/update.lst

Hmm, that file doesn't look like it's supposed to be read by humans, but rather be parsed by a machine?

MHerbst:
Yes, it will be interpreted and displayed if you execute the update for an OH installation. I am not sure but maybe also parts of the release notes (https://github.com/openhab/openhab-distro/wiki/openHAB-3.1-Release-Notes#breaking-changes-that-require-manual-interaction-after-the-upgrade) are maybe generated from this file

maniac103:
> Yes, it will be interpreted and displayed if you execute the update for an OH installation.

Thanks, I think I understand the file format now, I only halfway looked through it before.
-> https://github.com/openhab/openhab-distro/pull/1314

kaikreuzer (approved):
Sounds all reasonable to me despite being a breaking change - let's hope that indeed nobody uses the double press event...

maniac103:
Thanks, @kaikreuzer! I undrafted the distro PR for the release notes update that I mentioned above.

Elle4u:
Hi. I used double_press in my rules. It did not work very save, so maybe it will be better now.
But I have to change all the rules, now, to use a timer... I don't think I am the only one?

maniac103:
> I don't think I am the only one?

Probably not, but I'd be really (as in: **really**) suprised if more than a few people are using this. I guess it would help understanding your specific use case, in particular
- What's the specific use case you use this for?
- How do you distinguish between single and double button presses in your rules (given the former inevitable precede the latter)?

Elle4u:
I used it for my rollershutters.

`rule "BüroMarkus -> Rolladen Doppelklick"
  when
    Channel 'homematic:HmIPW-DRI32:CCU3:00171BE99324D8:2#BUTTON' triggered DOUBLE_PRESSED or
    Channel 'homematic:HmIPW-DRI32:CCU3:00171BE99324D8:3#BUTTON' triggered DOUBLE_PRESSED
  then
    Rolladen_BueroM_Proxy.sendCommand(Rolladen_BueroM_Autom_ZielSun.state as Number)
  end
`

So if I want the rollershutter to be placed at a given point (manual sunprotection) I pressed the buttons twice.
The single_pressed events I did not use in openhab. I have direct connections between the HomematicIP-Wired buttons and the rollershutters (also HomeMaticIP Wired). So this also works without openhub running.

But I will rewrite the rules with a timer. Should be no big thing (I home ;)).

maniac103:
> The single_pressed events I did not use in openhab

That's the explanation why it works for you ;-)
I originally wanted to use both single and double press events, but noticed it's impossible to meaningfully support both (that is: send double and **only** double on double press, send single on single press) in the current code structure. Since I figured vastly more people will use single over double press I removed the latter. The rule shouldn't be too hard though, it's basically (pseudocode):

```
var doublePressTimer

If thing triggered single press {
    If doublePressTimer != null {
        cancelTimer(doublePressTimer)
        doublePressTimer = null
        Execute double press code
    } Else {
        doublePressTimer = createTimer(500ms, {
            doublePressTimer = null
        }
    }
}
```

Elle4u:
> Wasn't there a `WORKING` datapoint for that purpose? That should help with the other questions as well ;-)

Thanks. Good hint. Maybe it is... I will have a look to it.

> That's a good question I honestly can't answer, but the answer likely is 'probably not'. But this problem should have existed before as well, shouldn't it?

I think so. But I do not understand that behavior.

I am friend of "It works out of the box (direct connections) as far as possible" and I will add functions with openhab.
So noone (my wife :) ) must be scary of "nothing is working if you are not there"...

maniac103:
> I think so. But I do not understand that behavior.

If the genuine HM and HMIP devices behave like AskSinPP (see [here](https://github.com/pa-pa/AskSinPP/blob/master/Remote.h#L66-L75)), they're sending out button release events only to peers (direct connections) while it will broadcast long press events (so the central will also get them). Do you get LONG_RELEASED events while having the direct connection active?
 (The reason for that behavioral difference is completely unclear to me, BTW)

> I am friend of "It works out of the box (direct connections) as far as possible" and I will add functions with openhab.
> So noone (my wife :) ) must be scary of "nothing is working if you are not there"...

Heh, I know that feeling ;-) I'm just a bit more 'lucky' to be forced to route everything through OH because I mainly control lighting stuff via the hue binding, so can't use direct connections there. Just mission critical things like underfloor heating control is linked directly.

PR #11189, opened 2021-09-01
[elroconnects] New binding for Elro Connects

mherwege:

This PR introduces a binding for the Elro Connects smarthome system.

The binding code is complete. I have had it running for a while without issues. I am aware of a few people using it as well, and have not received issues.


bundles/org.openhab.binding.elroconnects/src/main/resources/OH-INF/thing/thing-types.xml
```diff
@@ -0,0 +1,268 @@
...
+		</config-description>
+	</thing-type>
+	<thing-type id="motionsensor">
+		<supported-bridge-type-refs>
+			<bridge-type-ref id="connector"/>
+		</supported-bridge-type-refs>
+		<label>Motion sensor</label>
+		<description>ELRO Connects motion sensor</description>
+		<channels>
+			<channel id="motionalarm" typeId="motionalarm"/>
```
Skinah, line 141: Channels should be lowerCamelCase

https://next.openhab.org/docs/developer/guidelines.html#java-coding-style

Please apply to all channels.

bundles/org.openhab.binding.elroconnects/src/main/resources/OH-INF/thing/thing-types.xml
```diff
@@ -0,0 +1,268 @@
...
+				<label>Device ID</label>
+				<description>ID of the ELRO Connects Device.</description>
+			</parameter>
+		</config-description>
+	</thing-type>
+	<thing-type id="motionsensor">
+		<supported-bridge-type-refs>
+			<bridge-type-ref id="connector"/>
+		</supported-bridge-type-refs>
+		<label>Motion sensor</label>
```
Skinah, line 138:
```suggestion
		<label>Motion Sensor</label>
```
Labels should have all words start with a capital. Please change all that need this done.

bundles/org.openhab.binding.elroconnects/src/main/resources/OH-INF/thing/thing-types.xml
```diff
@@ -0,0 +1,268 @@
...
+	<channel-type id="temperature">
+		<item-type>Number:Temperature</item-type>
+		<label>Temperature</label>
+		<description>Current temperature</description>
+		<category>Temperature</category>
+		<tags>
+			<tag>Measurement</tag>
+			<tag>Temperature</tag>
+		</tags>
+		<state pattern="%.0f %%" readOnly="true"/>
```
Skinah, line 266:
```suggestion
		<state readOnly="true" pattern="%.1f %unit%" />
```
You should leave the unit here so it auto adds degrees C or F for the user. Also did you want no decimal places or is 1 useful?

See https://next.openhab.org/docs/concepts/units-of-measurement.html#defining-channeltypes

bundles/org.openhab.binding.elroconnects/src/main/resources/OH-INF/thing/thing-types.xml
```diff
@@ -0,0 +1,268 @@
...
+				<label>Device ID</label>
+				<description>ID of the ELRO Connects Device.</description>
+			</parameter>
+		</config-description>
+	</thing-type>
+
+	<channel-type id="scene">
+		<item-type>String</item-type>
+		<label>Scene</label>
+		<description>Scene</description>
```
Skinah, line 195: Could you add a little more to the description to give a more in depth detail beyond what the label is?

bundles/org.openhab.binding.elroconnects/src/main/java/org/openhab/binding/elroconnects/internal/handler/ElroConnectsDeviceHandler.java
```diff
@@ -0,0 +1,175 @@
...
+    @Override
+    public void initialize() {
+        ElroConnectsDeviceConfiguration config = getConfigAs(ElroConnectsDeviceConfiguration.class);
+        Integer id = config.deviceId;
+        deviceId = id;
+
+        if (id == null) {
+            updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.CONFIGURATION_ERROR, "Device ID not set");
+            return;
+        }
```
Skinah, line 55: Just as a comment no need to change anything, you could have marked the config.deviceId as a required config in the xml. Then the core will not create your handler unless all the configs that are required have a value. This means not having them marked as Nullable and you don't need to check for it at least 6 times that I have noticed so far.

bundles/org.openhab.binding.elroconnects/src/main/java/org/openhab/binding/elroconnects/internal/handler/ElroConnectsDeviceConfiguration.java
```diff
@@ -0,0 +1,27 @@
...
+
+/**
+ * The {@link ElroConnectsDeviceConfiguration} class contains fields mapping thing configuration parameters.
+ *
+ * @author Mark Herwege - Initial contribution
+ */
+@NonNullByDefault
+public class ElroConnectsDeviceConfiguration {
+
+    public @Nullable Integer deviceId;
```
Skinah, line 26:
```suggestion
    public Integer deviceId=0;
```

If its done like this and marked as required in the XML it saves it from needing to be marked as Nullable. Less code.

bundles/org.openhab.binding.elroconnects/src/main/java/org/openhab/binding/elroconnects/internal/handler/ElroConnectsBridgeHandler.java
```diff
@@ -0,0 +1,880 @@
...
+            return;
+        }
+
+        InetAddress addr = null;
+        try {
+            addr = getAddr();
+        } catch (IOException e) {
+            logger.warn("Communication error when finding IP address.");
+            updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.COMMUNICATION_ERROR,
+                    "Error trying to find IP address for connector ID " + connectorId + ".");
```
Skinah, line 179: See https://next.openhab.org/docs/developer/guidelines.html#f-logging

Suggest you remove the logger.warn as any status update already gets logged to the event.log

bundles/org.openhab.binding.elroconnects/src/main/java/org/openhab/binding/elroconnects/internal/handler/ElroConnectsBridgeHandler.java
```diff
@@ -0,0 +1,880 @@
...
+            stopCommunication();
+            return;
+        }
+        if (addr == null) {
+            logger.warn("No IP address for connector found.");
+            updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.COMMUNICATION_ERROR,
+                    "Error trying to find IP address for connector ID " + connectorId + ".");
```
Skinah, line 186: and here

bundles/org.openhab.binding.elroconnects/src/main/java/org/openhab/binding/elroconnects/internal/handler/ElroConnectsBridgeHandler.java
```diff
@@ -0,0 +1,880 @@
...
+            stopCommunication();
+            return;
+        }
+
+        String ctrlKey = this.ctrlKey;
+        if (ctrlKey.isEmpty()) {
+            logger.warn("Error initializing communication, no key received.");
+            updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.COMMUNICATION_ERROR,
+                    "Communication data error while starting communication.");
```
Skinah, line 195: here

bundles/org.openhab.binding.elroconnects/src/main/java/org/openhab/binding/elroconnects/internal/handler/ElroConnectsBridgeHandler.java
```diff
@@ -0,0 +1,880 @@
...
+        }
+
+        DatagramSocket socket;
+        try {
+            socket = createSocket(false);
+            this.socket = socket;
+        } catch (IOException e) {
+            logger.warn("Error initializing communication, could not create socket.");
+            updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.COMMUNICATION_ERROR,
+                    "Socket error while starting communication.");
```
Skinah, line 207: and here, please go through all uses of the logger and if your changing the status of the thing, remove the logger line.

bundles/org.openhab.binding.elroconnects/src/main/java/org/openhab/binding/elroconnects/internal/handler/ElroConnectsBridgeHandler.java
```diff
@@ -0,0 +1,880 @@
...
+        DatagramPacket packet = new DatagramPacket(buffer, buffer.length);
+        return receive(socket, buffer, packet);
+    }
+
+    private void send(DatagramSocket socket, String query, boolean broadcast) throws IOException {
+        final InetAddress address = broadcast
+                ? InetAddress.getByName(networkAddressService.getConfiguredBroadcastAddress())
+                : addr;
+        if (address == null) {
+            logger.debug("No address specified, cannot send");
```
Skinah, line 599: Should this be a warn with a slightly longer message to guide users to select a network in network settings?

bundles/org.openhab.binding.elroconnects/src/main/java/org/openhab/binding/elroconnects/internal/handler/ElroConnectsBridgeHandler.java
```diff
@@ -0,0 +1,880 @@
...
+        String connectorId = this.connectorId;
+        String ctrlKey = this.ctrlKey;
+        DatagramSocket socket = this.socket;
+        if (!connectorId.isEmpty() && !ctrlKey.isEmpty() && (socket != null)) {
+            logger.debug("Get device names");
+            ElroConnectsMessage elroMessage = new ElroConnectsMessage(msgIdIncrement(), connectorId, ctrlKey,
+                    ELRO_GET_DEVICE_NAME).withDeviceId(0);
+            sendElroMessage(socket, elroMessage, true);
+        } else {
+            // This should not happen if properly initialized
```
Skinah, line 670: Should the thing go OFFLINE with a config error message? Another way is to make the config fields REQUIRED so that the handler is not created by the framework until the user enters in values.

bundles/org.openhab.binding.elroconnects/src/main/java/org/openhab/binding/elroconnects/internal/handler/ElroConnectsBridgeHandler.java
```diff
@@ -0,0 +1,880 @@
...
+        String connectorId = this.connectorId;
+        String ctrlKey = this.ctrlKey;
+        DatagramSocket socket = this.socket;
+        if (!connectorId.isEmpty() && !ctrlKey.isEmpty() && (socket != null)) {
+            logger.debug("Get all equipment status");
+            ElroConnectsMessage elroMessage = new ElroConnectsMessage(msgIdIncrement(), connectorId, ctrlKey,
+                    ELRO_GET_DEVICE_STATUSES);
+            sendElroMessage(socket, elroMessage, true);
+        } else {
+            // This should not happen if properly initialized
```
Skinah, line 690: As above, since this is occurring multiple times, it may be worth looking at the suggestion to make the code shorter.

bundles/org.openhab.binding.elroconnects/src/main/java/org/openhab/binding/elroconnects/internal/devices/ElroConnectsDeviceTemperatureSensor.java
```diff
@@ -0,0 +1,87 @@
...
+                break;
+            default:
+                handler.updateState(BATTERY_LEVEL, new DecimalType(batteryLevel));
+                handler.updateState(LOW_BATTERY, (batteryLevel < 15) ? OnOffType.ON : OnOffType.OFF);
+                handler.updateState(TEMPERATURE, new QuantityType<>(temperature, CELSIUS));
+                handler.updateState(HUMIDITY, new QuantityType<>(humidity, Units.PERCENT));
+                handler.updateStatus(ThingStatus.ONLINE);
+        }
+
+        logger.debug("Device ID {}, battery level {}, temperature {}, humidity {}", deviceId, batteryLevel, temperature,
```
Skinah, line 84: This is mostly a double up and should be considered to be removed to improve performance, a check is done to see what level the logger is set to. Any channels that get updated will have this already logged to the event.log

bundles/org.openhab.binding.elroconnects/src/main/java/org/openhab/binding/elroconnects/internal/devices/ElroConnectsDeviceGenericAlarm.java
```diff
@@ -0,0 +1,131 @@
...
+                handler.updateState(LOW_BATTERY, UnDefType.UNDEF);
+                handler.updateStatus(ThingStatus.ONLINE, ThingStatusDetail.NONE, "Device " + deviceId + " has a fault");
+                break;
+            default:
+                handler.updateState(BATTERY_LEVEL, new DecimalType(batteryLevel));
+                handler.updateState(LOW_BATTERY, (batteryLevel < 15) ? OnOffType.ON : OnOffType.OFF);
+                handler.updateStatus(ThingStatus.ONLINE);
+        }
+
+        logger.debug("Device ID {}, battery level {}, status {}", deviceId, batteryLevel, elroStatus);
```
Skinah, line 129: As per previous comment. Already going into event.log

bundles/org.openhab.binding.elroconnects/src/main/java/org/openhab/binding/elroconnects/internal/devices/ElroConnectsDeviceEntrySensor.java
```diff
@@ -0,0 +1,106 @@
...
+@NonNullByDefault
+public class ElroConnectsDeviceEntrySensor extends ElroConnectsDevice {
+
+    private final Logger logger = LoggerFactory.getLogger(ElroConnectsDeviceEntrySensor.class);
+
+    // device states
+    private static final String STAT_OPEN = "55";
+    private static final String STAT_STILL_OPEN = "66";
+    private static final String STAT_FAULT = "11";
+    private static final String STAT_NORMAL = "AA";
```
Skinah, line 48: You are using the same constants in multiple classes. Consider moving them into a single static class so they can be re-used in all classes. When you make something static it will use RAM/memory even if you have no instances of the class, so by doing it multiple times it will use more ram then a single class that you re-use.

bundles/org.openhab.binding.elroconnects/src/main/java/org/openhab/binding/elroconnects/internal/ElroConnectsBindingConstants.java
```diff
@@ -0,0 +1,169 @@
...
+import java.util.Map;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+import org.eclipse.jdt.annotation.NonNullByDefault;
+import org.openhab.core.thing.ThingTypeUID;
+
+/**
+ * The {@link ElroConnectsBindingConstants} class defines common constants, which are
+ * used across the whole binding.
```
Skinah, line 24: Place your constants used in multiple classes into this one.

Skinah:
Binding looks pretty good, thanks for creating it... When you ready, just remove the WIP tag and post you want someone to take a look/review.

bundles/org.openhab.binding.elroconnects/src/main/resources/OH-INF/thing/thing-types.xml
mherwege: I will put in 1, but I am even not sure the api returns decimals.

bundles/org.openhab.binding.elroconnects/src/main/resources/OH-INF/thing/thing-types.xml
mherwege: Does this also apply for thing type id and channel type id?

bundles/org.openhab.binding.elroconnects/src/main/java/org/openhab/binding/elroconnects/internal/devices/ElroConnectsDeviceEntrySensor.java
mherwege: The problem is that each of the devices may have different codes for these. Many overlap, but not all. That's why they have been separated out. That way, the codes for one device type are kept together.

bundles/org.openhab.binding.elroconnects/src/main/java/org/openhab/binding/elroconnects/internal/ElroConnectsBindingConstants.java
mherwege: See above.

bundles/org.openhab.binding.elroconnects/src/main/resources/OH-INF/thing/thing-types.xml
Skinah: Usually bindings use a single word that is all lowercase for thing type ID.   Yes it is preferred that all channel id is this format as it helps textual config users if the names are predictable so that errors stand out.

bundles/org.openhab.binding.elroconnects/src/main/resources/OH-INF/thing/thing-types.xml
Skinah: Consider what happens when the api uses Fahrenheit and the core converts it to celcius, the extra decimal gives less rounding errors. You can change the default display to whatever you wish using metadata on the item. I think it is called pattern.

bundles/org.openhab.binding.elroconnects/src/main/java/org/openhab/binding/elroconnects/internal/devices/ElroConnectsDeviceEntrySensor.java
Skinah: No problem leave as is and I will close this.

bundles/org.openhab.binding.elroconnects/src/main/java/org/openhab/binding/elroconnects/internal/handler/ElroConnectsDeviceHandler.java
mherwege: I did some refactoring and avoided many of these checks.

mherwege:
@Skinah Many thanks for your review. I did the requested changes except for the constants, which I kept in the separate classes. Unfortunately, these constants can be different by device type.
I would still prefer more testing by other people, but I have not seen anyone picking up on it. I tested with the devices I have. I don't have all the devices the binding should support.
From my perspective, I removed the WIP label. I am OK to wait a bit before merging in the hope someone steps up. The alternative is to merge and react on feedback we may get.

bundles/org.openhab.binding.elroconnects/README.md
```diff
@@ -0,0 +1,190 @@
+# ElroConnects Binding
+
+The ElroConnects binding provides integration with the [Elro Connects](https://www.elro.eu/en/smart-home) smart home system.
```
Skinah, line 3: I believe from looking at their website that the correct way to write their name is `ELRO Connects` note the capitals. Please change any reference to their name to respect how they wish their brand to be presented in the readme. This matters to some companies.

bundles/org.openhab.binding.elroconnects/README.md
```diff
@@ -0,0 +1,190 @@
...
+
+The system uses a Wifi Hub (K1 Connector) to enable communication with various smart home devices.
```
Skinah, line 5:
```suggestion
The system uses a Wi-Fi Hub (K1 Connector) to enable communication with various smart home devices.
```

bundles/org.openhab.binding.elroconnects/README.md
```diff
@@ -0,0 +1,190 @@
...
+* Windows/door contact: `entrysensor`
+* Motion detector: `motionsensor`
+* Temperature and humidity monitor: `temperaturesensor`
+* Plug-in switch: `powersocket`
+
+The `connector` is the bridge thing.
+All other things are connected to the bridge.
+
+Testing was only done with smoke and water detectors connected to a K1 connector.
+The firmware version of the K1 connector was 2.0.3.30 at time of test.
```
Skinah, line 33:
```suggestion
The firmware version of the K1 connector was 2.0.3.30 at the time of testing.
```

bundles/org.openhab.binding.elroconnects/README.md
```diff
@@ -0,0 +1,190 @@
...
+Older versions of the firmware are known to have differences in the communication protocol.
+
+## Discovery
+
+The K1 connector `connector` cannot be auto-discovered.
+Once the bridge thing representing the K1 connector is correctly set up and online, discovery will allow discovering all devices connected to the K1 connector.
+
+The communication between the hub and devices may be disturbed.
```
Skinah, line 41: What can disturb them? perhaps re-word this if the intension is to point out that discovery may not work if the device is out of reliable RF range.

bundles/org.openhab.binding.elroconnects/src/main/resources/OH-INF/thing/thing-types.xml
```diff
@@ -0,0 +1,286 @@
...
+		</config-description>
+	</thing-type>
+
+	<channel-type id="scene">
+		<item-type>String</item-type>
+		<label>Scene</label>
+		<description>Scene selection from scenes configured in the Elro Connects app, enables configuring alarm modes</description>
+	</channel-type>
+
+	<channel-type id="smokealarm">
```
Skinah, line 198:
```suggestion
	<channel-type id="smokeAlarm">
```
I'm pretty sure you changed to using lowerCamelCase in the readme, but it appears they are not changed here. Please look at all channels as I wont raise a comment on each one.

mherwege: @Skinah I changed the channel id's to lowerCamelCase, but not the channel-type id's. I believe it is the channel id's that are visible to the user when configuring. I also use system channel type id's and they have yet another convention (system.xxx-xxx). There is a mistake in the README by the way I will correct. It should be Channel ID, not Channel Type ID.
Do you want me to change the channel type ID's to lowerCamelCase as well? And what about the thing type ID's?

mherwege:
@Skinah Skinah Thank you for the second review. Can you give me your view on my question regarding channel type ID's?

bundles/org.openhab.binding.elroconnects/src/main/resources/OH-INF/thing/thing-types.xml
Skinah: So long as the channels that the user sees and has to do textual config with, that is all that matters.

Skinah:
I usually make them the same, leave as is. The binding looks good you will need to wait now for someone to review that can merge it.

Skinah (approved):
LGTM

fwolter (changes requested):

bundles/org.openhab.binding.elroconnects/src/main/resources/OH-INF/thing/thing-types.xml
```diff
@@ -0,0 +1,286 @@
...
+	</channel-type>
+	<channel-type id="temperature">
+		<item-type>Number:Temperature</item-type>
+		<label>Temperature</label>
+		<description>Current temperature</description>
+		<category>Temperature</category>
+		<tags>
+			<tag>Measurement</tag>
+			<tag>Temperature</tag>
+		</tags>
```
fwolter, line 283: @lolodomo I'm not familiar with the sematic model. Does this make sense?

bundles/org.openhab.binding.elroconnects/README.md
```diff
@@ -0,0 +1,189 @@
...
+                    
+There is one optional advanced parameter, `refreshInterval`.
+This parameter controls the connection refresh heartbeat interval.
+The default is 60s.
+
+The syntax for configuration in a `.things` file is:
+
+```
+Thing elroconnects:connector:<connectorId> [connectorId="ST_<MAC address of connector>", refreshInterval=<heartbeat interval>]
+```
```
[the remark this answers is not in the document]
mherwege: There are bindings with the syntax in the configuration section (I looked at the Hue binding, although it is an example in that case and not the full syntax). This is the full syntax, not an example, and there are examples in the Full Example section. So I prefer keeping it hear.

bundles/org.openhab.binding.elroconnects/src/main/java/org/openhab/binding/elroconnects/internal/devices/ElroConnectsDevice.java
```diff
@@ -0,0 +1,127 @@
...
+    protected @Nullable ElroConnectsDeviceHandler getHandler() {
+        return bridge.getDeviceHandler(deviceId);
+    }
+
+    /**
+     * Update all {@link ElroConnectsDeviceHandler} channel states with information received from the device. This
+     * method needs to be implemented in the concrete subclass when any state updates are received from the device.
+     */
+    public void updateState() {
+    }
```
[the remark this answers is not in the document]
mherwege: Not all of the device types support all of the methods. updateState could be made abstract, but the other 3 methods cannot.

bundles/org.openhab.binding.elroconnects/src/main/java/org/openhab/binding/elroconnects/internal/devices/ElroConnectsDeviceCxsmAlarm.java
```diff
@@ -0,0 +1,136 @@
...
+            case UNDEF:
+                handler.updateState(BATTERY_LEVEL, UnDefType.UNDEF);
+                handler.updateState(LOW_BATTERY, UnDefType.UNDEF);
+                handler.updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.COMMUNICATION_ERROR,
+                        "Device " + deviceId + " is not syncing with K1 hub");
+                break;
+            case FAULT:
+                handler.updateState(BATTERY_LEVEL, UnDefType.UNDEF);
+                handler.updateState(LOW_BATTERY, UnDefType.UNDEF);
+                handler.updateStatus(ThingStatus.ONLINE, ThingStatusDetail.NONE, "Device " + deviceId + " has a fault");
```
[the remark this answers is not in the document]
mherwege: It shouldn't. It is the device explicitely sending a FAULT message, but the communication is still alive and we can send commands.

bundles/org.openhab.binding.elroconnects/src/main/java/org/openhab/binding/elroconnects/internal/handler/ElroConnectsBridgeHandler.java
```diff
@@ -0,0 +1,855 @@
...
+        }
+
+        ackPacket = new DatagramPacket(ACK, ACK.length, addr, PORT);
+
+        logger.debug("Connected to connector {} at {}:{}", connectorId, addr, PORT);
+
+        try {
+            // Start ELRO Connects listener. This listener will act on all messages coming from
+            // ELRO K1 Connector.
+            (new Thread(this::runElroEvents, THREAD_NAME_PREFIX + thing.getUID().getAsString())).start();
```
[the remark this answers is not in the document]
mherwege: NamedThreadFactory creates a name like OH-id-counter. I thought the convention for binding threads was OH-binding-id. I will change it, but see limited added value.

bundles/org.openhab.binding.elroconnects/src/main/java/org/openhab/binding/elroconnects/internal/handler/ElroConnectsBridgeHandler.java
```diff
@@ -0,0 +1,855 @@
...
+                break;
+            case TH_SENSOR:
+                device = new ElroConnectsDeviceTemperatureSensor(deviceId, this);
+                break;
+            default:
+                logger.debug("Device type {} not supported", deviceType);
+                return null;
+        }
+        device.setDeviceType(deviceType);
+        devices.put(deviceId, device);
```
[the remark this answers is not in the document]
mherwege: Discovery can use this for devices that have not been created as things.

bundles/org.openhab.binding.elroconnects/src/main/java/org/openhab/binding/elroconnects/internal/handler/ElroConnectsDeviceHandler.java
```diff
@@ -0,0 +1,150 @@
...
+    public void updateStatus(ThingStatus thingStatus) {
+        super.updateStatus(thingStatus);
+    }
+
+    /**
+     * Method to be called when an alarm event is received from the K1 hub. This should trigger a trigger channel. The
+     * method needs to be implemented in subclasses for the appropriate trigger channel.
+     */
+    public void triggerAlarm() {
+    }
```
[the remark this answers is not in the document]
mherwege: Not all devices implement a trigger alarm.

bundles/org.openhab.binding.elroconnects/src/main/java/org/openhab/binding/elroconnects/internal/handler/ElroConnectsBridgeHandler.java
```diff
@@ -0,0 +1,855 @@
...
+        DatagramSocket socket = this.socket;
+
+        if (socket != null) {
+            logger.debug("Listening for messages");
+
+            try {
+                byte[] buffer = new byte[4096];
+                DatagramPacket packet = new DatagramPacket(buffer, buffer.length);
+                keepRunning = true;
+                while (keepRunning) {
```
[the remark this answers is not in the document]
mherwege: Thanks for this.

bundles/org.openhab.binding.elroconnects/src/main/java/org/openhab/binding/elroconnects/internal/handler/ElroConnectsBridgeHandler.java
```diff
@@ -0,0 +1,855 @@
...
+
+    /**
+     * Send keep alive message.
+     *
+     * @throws IOException
+     */
+    private void keepAlive() throws IOException {
+        DatagramSocket socket = this.socket;
+        if (socket != null) {
+            logger.debug("Keep alive");
```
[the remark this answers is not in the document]
mherwege: Done.

bundles/org.openhab.binding.elroconnects/src/main/java/org/openhab/binding/elroconnects/internal/ElroConnectsBindingConstants.java
```diff
@@ -0,0 +1,172 @@
+/**
+ * Copyright (c) 2010-2021 Contributors to the openHAB project
```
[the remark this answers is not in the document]
mherwege: It didn't update them for me. I did a find and replace.

bundles/org.openhab.binding.elroconnects/src/main/resources/OH-INF/thing/thing-types.xml
```diff
@@ -0,0 +1,286 @@
...
+			<tag>OpenState</tag>
+		</tags>
+		<state readOnly="true"></state>
+	</channel-type>
+
+	<channel-type id="mutealarm">
+		<item-type>Switch</item-type>
+		<label>Mute Alarm</label>
+		<description>Mute</description>
+		<category>SoundVolume</category>
```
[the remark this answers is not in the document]
mherwege: used soundvolume_mute instead

mherwege:
@fwolter Happy New Year and many thanks for your review.
I have made many of the changes you requested, but also commented on a few. You may want to have another look at my comments and let me know what you think.

I have now also published the binding in the marketplace. I hope that makes it easier for people to test.

mherwege:
I discovered a case where removing the binding still had the communication running (while the handlers were disposed). The last commit should fix that.

bundles/org.openhab.binding.elroconnects/README.md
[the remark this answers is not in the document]
fwolter: The syntax is described in the general openHAB documentation. The binding documentation should only contain binding specific points to make it maintainable more easily, which is key with hundereds of bindings.

bundles/org.openhab.binding.elroconnects/src/main/java/org/openhab/binding/elroconnects/internal/devices/ElroConnectsDevice.java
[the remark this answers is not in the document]
fwolter: It's good practice to make methods abstract, which are not implemented in in the super class. If a subclass doesn't implement it, you can leave the method body empty. If you are keen you can add a comment like "// nothing" in the body to show future code editors, that it is kept empty intentionally.

bundles/org.openhab.binding.elroconnects/src/main/java/org/openhab/binding/elroconnects/internal/handler/ElroConnectsBridgeHandler.java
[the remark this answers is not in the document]
fwolter: OK, it would be better to keep discovery related code in the discovery class.

bundles/org.openhab.binding.elroconnects/src/main/resources/OH-INF/thing/thing-types.xml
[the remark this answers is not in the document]
fwolter: Are you sure this category exists? It isn't listed here https://www.openhab.org/docs/concepts/categories.html#thing-categories

bundles/org.openhab.binding.elroconnects/README.md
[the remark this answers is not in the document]
mherwege: OK, will remove as there are already examples.

bundles/org.openhab.binding.elroconnects/src/main/java/org/openhab/binding/elroconnects/internal/ElroConnectsBindingConstants.java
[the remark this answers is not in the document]
mherwege: Update now works.

bundles/org.openhab.binding.elroconnects/src/main/java/org/openhab/binding/elroconnects/internal/devices/ElroConnectsDevice.java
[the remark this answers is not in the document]
mherwege: OK, done

bundles/org.openhab.binding.elroconnects/src/main/resources/OH-INF/thing/thing-types.xml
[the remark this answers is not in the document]
mherwege: My fault, I looked at the icons, not categories.

bundles/org.openhab.binding.elroconnects/src/main/java/org/openhab/binding/elroconnects/internal/handler/ElroConnectsBridgeHandler.java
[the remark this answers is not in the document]
mherwege: This is not straightforward. The whole communication loop is inside the bridge, and it is not easy to do the discovery requests independently. While it might be possible to separate it out, I think it adds complexity. So I would prefer keeping it as is.

mherwege:
@fwolter Many thanks again for the review. I think I have addresses your comments. Can you check?

fwolter:
There are some formatting issues. You can fix them with `mvn spotless:apply`.

mherwege:
@fwolter it was a bit of a struggle to find what went wrong, but I hope this fixes the formatting issue. The issue was with the pom files in the bom and main bundles folder, and was not visible (and fixed) when running mvn spotless:apply only on the binding.

mherwege:
@jlaur @fwolter I believe I did all requested changes and provided feedback so far. Is there anything left on this for me?
In the mean time, I have been working on further enhancements to this binding, beyond what has been submitted for review here. In this PR, I only added minimal things requested by users to avoid creating a lot of extra review work. If you feel it would be better to include that directly in this PR, and not as a follow-up PR, let me know and I will do so. Otherwise, what is left for this PR, so I can build upon it?

bundles/org.openhab.binding.elroconnects/src/main/resources/OH-INF/thing/thing-types.xml
fwolter: @lolodomo ping

bundles/org.openhab.binding.elroconnects/src/main/java/org/openhab/binding/elroconnects/internal/handler/ElroConnectsBridgeHandler.java
```diff
@@ -0,0 +1,862 @@
...
+            updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.CONFIGURATION_ERROR,
+                    "Device ID not of format ST_xxxxxxxxxxxx with xxxxxxxxxxxx the lowercase MAC address of the connector");
+            return;
+        }
+
+        this.addr = null;
+        String ipAddress = config.ipAddress;
+        if (!ipAddress.isEmpty()) {
+            try {
+                InetAddress addr = InetAddress.getByName(ipAddress);
```
fwolter, line 171: `InetAddress.getByName` is a network operation and can block `intialize()`, which should return fast. You could resolve it in `startCommunication()`.

fwolter (changes requested):

fwolter:
> In this PR, I only added minimal things requested by users to avoid creating a lot of extra review work.

This is exactly the right approach! After the last issue is fixed, this should be ready to merge.

mherwege:
@fwolter Thank you for your review. I think the new commit resolves your finding.

bundles/org.openhab.binding.elroconnects/src/main/java/org/openhab/binding/elroconnects/internal/handler/ElroConnectsBridgeHandler.java
mherwege: Updated

fwolter (approved):
LGTM

PR #11192, opened 2021-09-02
[AirQuality] Tagging channels, prepared for Crowdin, dynamic channels

clinique:
With this upgrade you'll find : 

- Tagged channels
- A new pictogram channel based on the AQI value
- Pollutant channels are now dynamic, depending upon capabilities of the station
- Dynamic channels can be defined to set a warning based on specific sensitive groups toward the dominent pollutant
- Per pollutant, the binding calculates the actual QuantityType associated to a given AQI value (e.g. AQI 13 => 18,4 ppm)
- The binding has been adapted to be able to take benefits of Crowdin translations
- Some additional weather channels are available (dew-point, wind-speed).


clinique:
> What is the reason for configuring the API key at binding level? Did you consider using a Bridge?

It's a personal consideration, I dislike having empty bridges that only holds an ApiKey and no channel at all.

fwolter:
It's a common pattern to have a bridge for the credentials. The advantage over holding the credentials at binding level is that you can have multiple accounts configured. It's also beneficial for the usability when the configuration is consistent across bindings.

You already encapsulated the common API communication code in a separate class, which is good design. So, I'd like to encourage you to pull that design through and do this for the configuration, too.

clinique:
> It's a common pattern to have a bridge for the credentials. The advantage over holding the credentials at binding level is that you can have multiple accounts configured. It's also beneficial for the usability when the configuration is consistent across bindings.
> 
> You already encapsulated the common API communication code in a separate class, which is good design. So, I'd like to encourage you to pull that design through and do this for the configuration, too.

Ok, done.

bundles/org.openhab.binding.airquality/src/main/resources/OH-INF/thing/thing-types.xml
```diff
@@ -4,211 +4,231 @@
...
+		<state readOnly="true"/>
 	</channel-type>
 
-	<channel-type id="aqiColor" advanced="true">
-		<item-type>Color</item-type>
-		<label>AQI Color</label>
-		<description>Color associated to given AQI Index.</description>
+	<channel-type id="sensitive-group">
+		<item-type>Switch</item-type>
+		<label>Sensitive group</label>
```
fwolter, line 215: Words in labels should be capitalized (except prepositions and so on). See https://www.openhab.org/docs/developer/bindings/thing-xml.html#formatting-labels-and-descriptions

fwolter (changes requested):

bundles/org.openhab.binding.airquality/src/main/java/org/openhab/binding/airquality/internal/AirQualityHandlerFactory.java
```diff
@@ -27,23 +32,25 @@
...
- * The {@link AirQualityHandlerFactory} is responsible for creating things and thing
- * handlers.
+ * The {@link AirQualityHandlerFactory} is responsible for creating thing and thing
+ * handler.
  *
- * @author Kuba Wolanin - Initial contribution
+ * @author Gaël L'hopital - Initial contribution
  */
-@Component(service = ThingHandlerFactory.class, configurationPid = "binding.airquality")
+@Component(service = ThingHandlerFactory.class)
```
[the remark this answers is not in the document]
clinique: Yes, I thought it was pointless as the binding now has no configuration. In the doubt, I reintroduced it.

bundles/org.openhab.binding.airquality/src/main/resources/OH-INF/thing/thing-types.xml
clinique: Done

lolodomo:
The French properties file has to be loaded in Crowdin.

Issue #11202, opened 2021-09-04
[openthermgateway] Add support for Ventilation/Heat Recovery and Solar Boiler units


ArjenKorevaar:
With OpenTherm protocol v3.0 support was added for Ventilation/Heat Recovery and Solar Boiler units, but the OpenThermGateway binding currently only supports Central Heating Boiler and Thermost units.

The general idea is to create a Bridge for the connection to the OpenTherm Gateway device and then add separate Things for the Central Heating Boiler, Ventilation/Heat Recovery and Solar Boiler units.

I am already working on the Bridge part and support for Ventilation/Heat recovery.

The request for support for Ventilation/Heat Recovery units started with a post on the community forum by [Tomek Krzyzak](https://community.openhab.org/t/opentherm-gateway-binding/39160/231) on March 11. Since then several users have joined in on testing new versions. I am currently finishing up coding and testing and will then submit a PR.

PR #11203, opened 2021-09-04
[openthermgateway] Add support for Ventilation/Heat Recovery units

ArjenKorevaar:
Added support for Ventilation/Heat Recovery units to the OpenTherm Gateway binding.

More info at #11202

Solar Boiler units are currently only partially supported by the OpenTherm Gateway device itself, so this may be added to the binding at a later time.


ArjenKorevaar:
Not sure why the build failed, it seems unrelated to the changes made to the binding.

Stacktrace
```
org.opentest4j.AssertionFailedError: Error cleaning up test (deleting table)
Caused by: java.util.concurrent.ExecutionException: software.amazon.awssdk.services.dynamodb.model.ResourceNotFoundException: Cannot do operations on a non-existent table (Service: DynamoDb, Status Code: 400, Request ID: 5c0b87d4-911e-4f17-a872-69038f24a1dd, Extended Request ID: null)
Caused by: software.amazon.awssdk.services.dynamodb.model.ResourceNotFoundException: Cannot do operations on a non-existent table (Service: DynamoDb, Status Code: 400, Request ID: 5c0b87d4-911e-4f17-a872-69038f24a1dd, Extended Request ID: null)
```

ArjenKorevaar:
Rebased onto main, now PR build succeeds

ArjenKorevaar:
Rebased onto main, build fails but seems unrelated to PR: `The JAR/ZIP file (/home/runner/.m2/repository/org/openhab/core/bundles/org.openhab.core.addon.marketplace/3.2.0-SNAPSHOT/org.openhab.core.addon.marketplace-3.2.0-SNAPSHOT.jar) seems corrupted`

bundles/org.openhab.binding.openthermgateway/README.md
```diff
@@ -1,195 +1,432 @@
...
-Number:Dimensionless DomesticHotWaterBurnerStarts "Domestic hot water burner starts" { channel="openthermgateway:otgw:1:dhwburnerstarts" }
-Number:Time BurnerHours "Burner hours" { channel="openthermgateway:otgw:1:burnerhours" }
-Number:Time CentralHeatingPumpHours "Central heating pump hours" { channel="openthermgateway:otgw:1:chpumphours" }
-Number:Time DomesticHotWaterPumpValveHours "Domestic hot water pump/valve hours" { channel="openthermgateway:otgw:1:dhwpvhours" }
-Number:Time DomesticHotWaterBurnerHours "Domestic hot water burner hours" { channel="openthermgateway:otgw:1:dhwburnerhours" }
-Text SendCommand "Send command channel" { channel="openthermgateway:otgw:1:sendcommand" }
-```
-
-### demo.sitemap
+#### Example
```
fwolter, line 45: To be consistent with the other binding documentations, the examples should reside at the end of the readme.

bundles/org.openhab.binding.openthermgateway/README.md
```diff
@@ -1,195 +1,432 @@
...
+
+The `ventilationheatrecovery` thing does not require any configuration settings.
+
+#### Channels
+
+The `ventilationheatrecovery` thing supports the following channels:
+
+| Channel ID | Item Type | Description | Access |
+|------------|-----------|-------------|--------|
+| vh_ventilationenable | Switch | Ventilation enabled | R |
```
fwolter, line 260: What's the reason for adding a suffix to the channel names?

bundles/org.openhab.binding.openthermgateway/README.md
```diff
@@ -1,195 +1,432 @@
...
+```
+
+## Transparent Slave Parameters and Fault History Buffer channels
+
+The transparent slave parameters (TSP) and fault history buffer (FHB) use a variable number of entries.
+The number of entries is determined by a TSP or FHB size message.
+Channels for TSP and FHB entries are automatically created when a TSP or FHB size message is received.
+An index number is added to the base channel name to create a unique channel name for each entry.
+
+For example, if a TSP size message is received for a boiler unit (OpenTherm DATA-ID 10) with value 60, then channels `tspentry_0` through `tspentry_59` will be automatically created and linked to the corresponding TSP entry (OpenTherm DATA-ID 11).
```
fwolter, line 415: How are these Channels used? Do you need a sufficient number of Items configured to display the values in the UI?

bundles/org.openhab.binding.openthermgateway/README.md
```diff
@@ -1,195 +1,432 @@
...
+        Text item="VentilationSetpoint" label="Ventilation Setpoint" minValue="0" maxValue="100" step="1"
+        Text item="TransparentSlaveParameterNumber" label="Transparent Slave Parameter Number"
+        Text item="TransparentSlaveParameterEntry" label="Transparent Slave Parameter Entry"
+        Text item="FaultHistoryBufferNumber" label="Fault History Buffer Number"
+        Text item="FaultHistoryBufferEntry" label="Fault History Buffer Entry"
+    }
+}
+```
+
+## Transparent Slave Parameters and Fault History Buffer channels
```
fwolter, line 408: These chapters should reside above the "Full Example" section.

bundles/org.openhab.binding.openthermgateway/src/main/java/org/openhab/binding/openthermgateway/handler/BaseDeviceHandler.java
```diff
@@ -0,0 +1,168 @@
...
+    }
+
+    @Override
+    public void bridgeStatusChanged(ThingStatusInfo bridgeStatusInfo) {
+        if (bridgeStatusInfo.getStatus() == ThingStatus.ONLINE) {
+            updateStatus(ThingStatus.ONLINE);
+        } else {
+            updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.BRIDGE_OFFLINE);
+        }
+    }
```
fwolter, line 70: Actually, this is handled by the framework. It do works when the state of the Thing is UNKNOWN, ONLINE, OFFLINE and so on, but not when it's INITIALIZING.

bundles/org.openhab.binding.openthermgateway/src/main/java/org/openhab/binding/openthermgateway/handler/BaseDeviceHandler.java
```diff
@@ -0,0 +1,168 @@
...
+
+                State state = dataItem.createState(message);
+
+                logger.debug("Received update for channel {}: {}", channelId, state);
+                updateState(channelId, state);
+            }
+        }
+    }
+
+    private void VerifyTspFhbChannels(int id, int size) {
```
fwolter, line 115:
```suggestion
    private void verifyTspFhbChannels(int id, int size) {
```

bundles/org.openhab.binding.openthermgateway/src/main/java/org/openhab/binding/openthermgateway/handler/OpenThermGatewayHandler.java
```diff
@@ -52,38 +42,35 @@
...
     @Override
     public void initialize() {
-        logger.debug("Initializing OpenTherm Gateway handler for uid '{}'", getThing().getUID());
+        logger.debug("Initializing OpenThermGateway handler for uid {}", getThing().getUID());
 
-        updateStatus(ThingStatus.UNKNOWN, ThingStatusDetail.NONE, "Initializing");
-
-        config = getConfigAs(OpenThermGatewayConfiguration.class);
+        configuration = getConfigAs(OpenThermGatewayConfiguration.class);
+        logger.info("Using configuration: {}", configuration.toString());
```
fwolter, line 65: Can this message be replaced by using the debugger? See point 4 https://www.openhab.org/docs/developer/guidelines.html#f-logging

bundles/org.openhab.binding.openthermgateway/src/main/java/org/openhab/binding/openthermgateway/handler/OpenThermGatewayHandler.java
```diff
@@ -108,184 +95,177 @@ public void handleCommand(ChannelUID channelUID, Command command) {
...
     public void handleRemoval() {
-        logger.debug("Removing OpenTherm Gateway handler");
+        logger.debug("Removing OpenThermGateway handler");
         disconnect();
         super.handleRemoval();
     }
 
     @Override
     public void dispose() {
+        logger.debug("Disposing OpenThermGateway handler");
```
fwolter, line 177: Can this message be replaced by using the debugger? See point 4 https://www.openhab.org/docs/developer/guidelines.html#f-logging
```
void myFun() {
    logger.trace("Enter myfun"); // DONT, DONT, really DONT do that
    doSomething();
    logger.trace("Leave myfun"); // DONT, DONT, really DONT do that
}
```

bundles/org.openhab.binding.openthermgateway/src/main/resources/OH-INF/thing/channels.xml
```diff
@@ -322,38 +342,378 @@
 	</channel-type>
 
 	<channel-type id="burnerhours">
-		<item-type>Number:Time</item-type>
+		<item-type>Number:Dimensionless</item-type>
```
fwolter, line 345: What's the reason for this change?

bundles/org.openhab.binding.openthermgateway/src/main/resources/OH-INF/thing/channels.xml
```diff
@@ -322,38 +342,378 @@
...
+		<label>Fault History Buffer Entry</label>
+		<description>Fault history buffer entry</description>
+		<state readOnly="true"/>
+	</channel-type>
+
+	<!-- ventilationheatrecovery -->
+
+	<channel-type id="vh_ventilationenable">
+		<item-type>Switch</item-type>
+		<label>Ventilation Enabled</label>
```
fwolter, line 404:
```suggestion
		<label>Ventilation</label>
```

bundles/org.openhab.binding.openthermgateway/src/main/resources/OH-INF/thing/channels.xml
```diff
@@ -322,38 +342,378 @@
...
+		<description>OpenTherm version</description>
+		<state readOnly="true"/>
+	</channel-type>
+
+	<channel-type id="vh_versiontype">
+		<item-type>Number:Dimensionless</item-type>
+		<label>Version Type</label>
+		<description>Version type</description>
+		<state readOnly="true"/>
+	</channel-type>
```
fwolter, line 586: These seem rather porperties than channels.

bundles/org.openhab.binding.openthermgateway/src/main/resources/OH-INF/thing/channels.xml
```diff
@@ -322,38 +342,378 @@
...
+	<channel-type id="vh_actualinletfanspeed">
+		<item-type>Number:Dimensionless</item-type>
+		<label>Actual Inlet Fan Speed</label>
+		<description>Actual inlet fan speed</description>
+		<state readOnly="true" pattern="%d rpm"/>
+	</channel-type>
+
+	<channel-type id="vh_nominalventenable">
+		<item-type>Switch</item-type>
+		<label>Nominal Ventilation Value Transfer</label>
```
fwolter, line 653: Labels are expected to be as short as possible. Guideline is 2-3 words with max 20-25 chars. See https://www.openhab.org/docs/developer/bindings/thing-xml.html#formatting-labels-and-descriptions

bundles/org.openhab.binding.openthermgateway/src/main/resources/OH-INF/thing/channels.xml
```diff
@@ -322,38 +342,378 @@
...
+			</options>
+		</state>
+	</channel-type>
+
+	<channel-type id="vh_nominalventilationvalue">
+		<item-type>Number:Dimensionless</item-type>
+		<label>Nominal Ventilation Value</label>
+		<description>Nominal ventilation value</description>
+		<state readOnly="true" pattern="%d %%"/>
+	</channel-type>
```
fwolter, line 675: Do these have the same label by intention?

fwolter (changes requested):
There are some checkstyle warnings left. You could take a look at `target/code-analysis/report.html` after building it with `mvn clean install`.

Are there any breaking changes like renamed channels or renamed Thing/Bridge UIDs?

bundles/org.openhab.binding.openthermgateway/README.md
ArjenKorevaar: The number of TSP and FHB channels is different for each device. The gateway first requests the number of TSP and FHB values and then proceeds to fetch each value. You would indeed need corresponding Items if you would like to display the values in the UI. Based on documentation of the device, a user should be able to tell how many values there are and what each value means. They may choose to create Items only for those channels that are of interest.

bundles/org.openhab.binding.openthermgateway/src/main/resources/OH-INF/thing/channels.xml
ArjenKorevaar: It's not a Time data type, it's just a number of hours (integer value).

ArjenKorevaar:
@fwolter thank you for the review! I will look into each comment but this may take some time. I'll let you know once it's ready to be reviewed again.

bundles/org.openhab.binding.openthermgateway/README.md
ArjenKorevaar: The names of the channels are mostly a copy/paste from the OpenTherm specifications. And those names do not always follow openHAB coding guidlines when it comes to the number of words, length in characters or things like using a suffix. So I find myself abbreviating or leaving words out.. just to comply with coding guidlines really but not making anything better in terms of readability imho or matching with the OT specs. So.. I don't know.. should I change them?

bundles/org.openhab.binding.openthermgateway/README.md
ArjenKorevaar: I have struggled a bit with this. There are multiple things  and I could either document each thing directly followed by an example, or split it up and first document all configs of all things, then all channels of all things, followed by all examples.  Which to me makes it scattered around and hard to read.

Regarding consistency... I haven't really found any :) especially when it comes to bindings with bridges and multiple things.

bundles/org.openhab.binding.openthermgateway/README.md
ArjenKorevaar: Please see my previous comment about readme structure.

bundles/org.openhab.binding.openthermgateway/src/main/resources/OH-INF/thing/channels.xml
ArjenKorevaar: I think this is in the same line as your comment about using a prefix? In that case, please read my reply on that.

bundles/org.openhab.binding.openthermgateway/src/main/resources/OH-INF/thing/channels.xml
ArjenKorevaar: Please see my comment on names derived form OT specification

bundles/org.openhab.binding.openthermgateway/src/main/resources/OH-INF/thing/channels.xml
ArjenKorevaar: Interesting thought, haven't looked at it like that. The values are read from the incoming stream from the OTGW just as all other values, and then based on the message code the corresponding channel is updated. This is generic logic for all channels. Apart from the value being read at some point and not known upfront.. changing the generic logic for handling incoming messages to support some of them being mapped to properties instead of channels would have quite an impact. Perhaps a nice to have for future versions?

bundles/org.openhab.binding.openthermgateway/src/main/java/org/openhab/binding/openthermgateway/handler/BaseDeviceHandler.java
ArjenKorevaar: Thanks, changed it to lc

bundles/org.openhab.binding.openthermgateway/src/main/java/org/openhab/binding/openthermgateway/handler/OpenThermGatewayHandler.java
ArjenKorevaar: I use logger.debug mainly so I can ask people using the binding to send me their logs to analyze what happened. I personally don't have a ventilation heat recovery unit for example and my boiler only supports certain commands. Also I use a different kind of gateway which makes it hard to figure out what happens without some logging. Being able to trace whats going on, when functions were called, when a connection was dropped or when the background connection thread ended has proven to be very valuable. Afaik this is the main purpose of debug logging?

bundles/org.openhab.binding.openthermgateway/src/main/resources/OH-INF/thing/channels.xml
ArjenKorevaar: Not by my intention :) .. again taken from the OT specs (or rather from the OTGW source code) which I just checked and indeed has two different message codes with the same description. I don't know what they mean exactly, the percentage value seems clear but the r/w value isn't to me. Mapping every message code to a channel, I simply aim to support the full set.

bundles/org.openhab.binding.openthermgateway/src/main/java/org/openhab/binding/openthermgateway/handler/BaseDeviceHandler.java
ArjenKorevaar: So, this can be removed alltoghether and thing status is automatically updated when bridge status updates ?

bundles/org.openhab.binding.openthermgateway/src/main/resources/OH-INF/thing/channels.xml
fwolter: Numer:Time represents a timespan (e.g. in hours). You might have mixed it up with with the Item type DateTime, which represents a date.

By using Units of Measure (Number:Time), the user can choose the desired unit and the framework takes care for the correct conversion.

bundles/org.openhab.binding.openthermgateway/README.md
fwolter: I'd remove the prefix as it seems redundant for all Channels.

bundles/org.openhab.binding.openthermgateway/README.md
fwolter: Ok, if the Channel count is constant it should be fine.

bundles/org.openhab.binding.openthermgateway/src/main/java/org/openhab/binding/openthermgateway/handler/BaseDeviceHandler.java
fwolter: Yes

bundles/org.openhab.binding.openthermgateway/src/main/resources/OH-INF/thing/channels.xml
fwolter: My comment has nothing to do with the prefix in the Channel UID.

"Enabled" is redundant as the switch shows if Ventilation is ON or OFF. However, I won't insist on changing it.

bundles/org.openhab.binding.openthermgateway/src/main/resources/OH-INF/thing/channels.xml
fwolter: I won't insist on changing it either, but keep in mind that changing them to properties in a later release breaks backward compatibility. So, If you change them, you should do it now.

bundles/org.openhab.binding.openthermgateway/src/main/resources/OH-INF/thing/channels.xml
fwolter: If they can't be shortend without loosing necessary information, I think it's ok to leave them as-is.

bundles/org.openhab.binding.openthermgateway/src/main/resources/OH-INF/thing/channels.xml
fwolter: OK

fwolter:
> Fixed most of them.. just the import of .* I haven't found a more elegant solution to

You can simply use the Organize Imports function of eclipse to resolve these.

I added the portentially not backward compatible label. The changes should be mentioned in the update.lst to show a warning when the user updates the system. See https://github.com/openhab/openhab-distro/blob/main/distributions/openhab/src/main/resources/bin/update.lst

bundles/org.openhab.binding.openthermgateway/src/main/java/org/openhab/binding/openthermgateway/internal/OpenThermGatewaySocketConnector.java
```diff
@@ -96,51 +90,49 @@ public void run() {
...
     }
 
     @Override
     public boolean isConnected() {
-        return connected;
+        return (socket != null && socket.isConnected());
     }
 
     @Override
     public void sendCommand(GatewayCommand command) {
```
andrewfg, line 120: The method sendCommand() can be called both from outside in the handler, and from inside the `run()` method, which are on different threads, so there is a risk of clashes; therefore it needs to be `synchronized`..

```
@Override
    public synchronized void sendCommand(GatewayCommand command) {```

File: bundles/org.openhab.binding.openthermgateway/src/main/java/org/openhab/binding/openthermgateway/handler/OpenThermGatewayHandler.java [hunk 15]
```diff
@@ -108,184 +94,177 @@ public void handleCommand(ChannelUID channelUID, Command command) {
...
+            }
+        }
+    }
 
-                if (state != null) {
-                    logger.debug("Received update for channel '{}': {}", channelId, state);
-                    updateState(channelId, state);
-                }
+    @Override
+    public void connectionStateChanged(ConnectionState state) {
```
andrewfg, line 144: The socket connecter is running in its own thread, with is a different thread than the OH framework, and in my tests I found that sometimes if the OH framework would block or is delayed, that causes the socket connector's `run()` method to be delayed (which can lead to e.g. connection errors or timeouts. Instead I suggest to fully decouple the two threads by interposing a separate scheduler task as follows...

```
 @Override
    public void connectionStateChanged(ConnectionState state) {
        scheduler.submit(() -> connectionStateChangedTask());
    }
    
    private void connectionStateChangedTask(ConnectionState state) {
```

bundles/org.openhab.binding.openthermgateway/src/main/java/org/openhab/binding/openthermgateway/handler/OpenThermGatewayHandler.java
```diff
@@ -108,184 +94,177 @@ public void handleCommand(ChannelUID channelUID, Command command) {
...
-                    updateState(OpenThermGatewayBindingConstants.CHANNEL_OVERRIDE_CENTRAL_HEATING2_ENABLED,
-                            OnOffType.from(!gatewayCommand.getMessage().equals("0.0")));
+            sendCommand(gatewayCommand);
+
+            if (code.equals(GatewayCommandCode.CONTROLSETPOINT)) {
+                if (gatewayCommand.getMessage().equals("0.0")) {
+                    updateState(CHANNEL_OVERRIDE_CENTRAL_HEATING_WATER_SETPOINT, UnDefType.UNDEF);
                 }
-            } else {
-                connect();
```
andrewfg, line 130: Just so you know: I fully agree with your deletion of the call to 'connect()` .. it was completely wrong to call that at this place :)

andrewfg (changes requested):
Review step 1 on the `OpenThermGatewaySocketConnector` and `OpenThermGatewayHandler` classes..

andrewfg:
> 2. Code review of all the other files in this PR.

@ArjenKorevaar before I start work on the general code review, I would like to ask your opinion about how you prefer to handle it. There are two possible ways..
1. I could make review comments and proposed changes here in this PR; the advantage is that all proposals are discussed and formally tracked in this; and the disadvantage is that you have to make the requested changes in your IDE yourself and make respective commits, or 
2. A alternatively I could derive a git branch on my repo from your own repo/branch, and make the changes directly in my IDE, and then make a 'sub-' PR from my repo/branch to your repo/branch; which if you would adopt my code changes, will then automatically flow into this PR here; the discussions and changes would then be tracked on your repo/branch, rather than here, but the advantage for you is that it saves you making the requested changes in your IDE yourself.

So it depends a bit on your workload and what you prefer. => Please let me know..

bundles/org.openhab.binding.openthermgateway/src/main/java/org/openhab/binding/openthermgateway/internal/OpenThermGatewaySocketConnector.java
```diff
@@ -63,31 +61,27 @@ public OpenThermGatewaySocketConnector(OpenThermGatewayCallback callback, String
 
     @Override
     public void run() {
```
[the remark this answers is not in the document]
andrewfg: Note: after making the post above, I made a slight edit to the above sample code, so please look at the last version.

ArjenKorevaar:
@andrewfg just a heads-up: I won't be able to spend time to look into this any time soon.

bundles/org.openhab.binding.openthermgateway/src/main/java/org/openhab/binding/openthermgateway/internal/OpenThermGatewaySocketConnector.java
```diff
@@ -41,17 +38,18 @@
 public class OpenThermGatewaySocketConnector implements OpenThermGatewayConnector {
     private static final int COMMAND_RESPONSE_TIME_MILLISECONDS = 100;
     private static final int COMMAND_TIMEOUT_MILLISECONDS = 5000;
+    private static final int READ_TIMEOUT_MILLISECONDS = 60000;
```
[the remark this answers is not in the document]
ArjenKorevaar: this was a rollback from a previous attempt by another user to fix the issue. how do you manage this irl ?

bundles/org.openhab.binding.openthermgateway/src/main/resources/OH-INF/thing/channels.xml
andrewfg: IMHO `ENABLED` is quite different than `ON` .. the former means that it is allowed to be either on or off; whereas the latter means that it is definitely off..

andrewfg:
> won't be able to spend time to look into this any time soon

@ArjenKorevaar no stress. But just so you know it, I am working on some changes that will do the following..
- Resolve some of the open issues of @fwolter 
- Resolve the merge conflict between this and my PR #12185 
- Resolve any other issues that I myself have found
- Update the ReadMe to the style requested by @fwolter 

When I am done, I will make a PR from my repo/branch to your repo/branch, so that if you merge my changes, they will cascade into this PR.

andrewfg:
> When I am done, I will make a PR from my repo/branch to your repo/branch, so that if you merge my changes, they will cascade into this PR

**_.. well that did not go quite as I expected .._** I really hate Git! Instead of pushing my proposed changes to a PR on @ArjenKorevaar 's repo (so he could optionally choose whether to adopt them or not, and then cascade into this PR), it seems that all of my commits landed directly in this PR automatically! (I don't really know how that happened..) But anyway, @ArjenKorevaar if you are NOT happy with my suggestions, then you will need to unwind the 15 commits made by me today February 3rd, 2022.

Please accept my humble apologies for the unintended usurpation of your PR. The good news is that all of my code review suggestions, (and some of @fwolter 's still open suggestions), are now included in the PR, so I can now give you an LGTM on the final result!

bundles/org.openhab.binding.openthermgateway/README.md
andrewfg: Done

bundles/org.openhab.binding.openthermgateway/README.md
andrewfg: Done

bundles/org.openhab.binding.openthermgateway/src/main/java/org/openhab/binding/openthermgateway/handler/BaseDeviceHandler.java
andrewfg: Done

bundles/org.openhab.binding.openthermgateway/src/main/resources/OH-INF/thing/channels.xml
andrewfg: Done

bundles/org.openhab.binding.openthermgateway/src/main/java/org/openhab/binding/openthermgateway/internal/OpenThermGatewaySocketConnector.java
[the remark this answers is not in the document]
andrewfg: Please see my new version of the `run()` method. I think the problem was that in the original code, SocketTimeOut exceptions were not causing the `callback.disconnect()` to be called. With the result that the connector had terminated, but the handler was not aware of the error, and so was nor auto restarting.

Note: I also converted the constant timeout values in the code to configuration parameters of the binding, so they can be tweaked by users in future.

andrewfg (approved):

andrewfg:
PS I note the following..
- the CI build failed, but I think it is not the fault of this code ??
- it also reports potential merge conflicts on two files, but I am not clever enough to resolve them..

andrewfg:
> potential merge conflicts on two files

The two files `OpenThermGatewayBindingConstants.java` and `DataType.java` are ones that have been moved (to another package in the project) or re-moved entirely from the project. I have no idea why the CI build system chokes on that, but I can assure you that it all builds fine locally in Eclipse or MVN.

andrewfg:
^
I don't know if this helps, but regarding the merge conflicts (see screenshot below)..
- `OpenThermGatewayBindingConstants.java` is tagged in Git as being a) deleted, and then b) added
- `DataType.java` is tagged in Git as being renamed to `ConnectionSate.java`

image

bundles/org.openhab.binding.openthermgateway/README.md
```diff
@@ -1,195 +1,404 @@
...
-| fault                     | Switch               | Fault indication                                         | R      |
-| servicerequest            | Switch               | Service required                                         | R      |
-| lockout-reset             | Switch               | Lockout-reset enabled                                    | R      |
-| lowwaterpress             | Switch               | Low water pressure fault                                 | R      |
-| gasflamefault             | Switch               | Gas or flame fault                                       | R      |
-| airpressfault             | Switch               | Air pressure fault                                       | R      |
-| waterovtemp               | Switch               | Water over-temperature fault                             | R      |
-| oemfaultcode              | Switch               | OEM fault code                                           | R      |
-| diag                      | Switch               | Diagnostics indication                                   | R      |
```
andrewfg, line 74: @ArjenKorevaar one more question: last year you introduced this `diag` channel which (I think) would indicate that the OTGW is talking to some kind of master or slave device (??). It is not that well documented. But I am surprised that you deleted it. If my assumption above is true then it would probably be a channel of the `openthermgateway` bridge thing, rather than a channel on a `boiler` or `ventilationheatrecovery` thing. ??

andrewfg (changes requested):

bundles/org.openhab.binding.openthermgateway/README.md
ArjenKorevaar: I think removing the diag channel was done unintentionally. Since the diag code is sent by the actual unit rather than the OTGW device (its part of the OpenTherm specifications), the channel should be on the boiler thing. I don't know if ventilation / heatrecovery units in general make use of this message id as well. And even then, some types of units may support it, others may not. I would usually add them on request.

andrewfg: Apparently `diag` wasn't deleted from DataItemGroup, so presumably it is still supported in reality? In that case it just needs to be restored to the ReadMe. Or??

andrewfg: .. and it also needs to be restored the `boiler.xml` file ..

andrewfg: ..see this https://github.com/openhab/openhab-addons/pull/11203/commits/47d4696e62a9a98d0cad53b477b5bab3f3286e76

andrewfg: so resolved

andrewfg (approved):

bundles/org.openhab.binding.openthermgateway/README.md
ArjenKorevaar: Indeed, the DataItemGroup is basically generated from the list of all possible messages id's of the OpenTherm specifications. The only thing that is needed then is a channel to map the incoming value to.

andrewfg:
One of my prior commits had broken the Thing initialization, so I pushed a fix just now.
Also I have installed the current JAR on my operative system, and will let it run fir a day or two.

bundles/org.openhab.binding.openthermgateway/src/main/java/org/openhab/binding/openthermgateway/handler/BaseDeviceHandler.java
```diff
@@ -0,0 +1,164 @@
...
+    public BaseDeviceHandler(Thing thing) {
+        super(thing);
+    }
+
+    @Override
+    public void initialize() {
+        Bridge bridge = getBridge();
+
+        if (bridge == null) {
+            updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.NONE, "Bridge is missing");
```
jlaur, line 60: Maybe `ThingStatusDetail.BRIDGE_UNINITIALIZED` and skip custom text?

bundles/org.openhab.binding.openthermgateway/src/main/java/org/openhab/binding/openthermgateway/handler/BaseDeviceHandler.java
```diff
@@ -0,0 +1,164 @@
...
+        } else if (bridge.getStatus() != ThingStatus.ONLINE) {
+            updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.BRIDGE_OFFLINE);
```
jlaur, line 62: If you leave initialize() with at least status `UNKNOWN`, this is not needed - will be done automatically by the framework.

bundles/org.openhab.binding.openthermgateway/src/main/java/org/openhab/binding/openthermgateway/handler/BaseDeviceHandler.java
```diff
@@ -0,0 +1,164 @@
...
+        } else {
+            updateStatus(ThingStatus.ONLINE, ThingStatusDetail.NONE);
```
jlaur, line 64: How do you know that device is `ONLINE` at this point?

bundles/org.openhab.binding.openthermgateway/src/main/java/org/openhab/binding/openthermgateway/handler/BaseDeviceHandler.java
andrewfg: Yes.

bundles/org.openhab.binding.openthermgateway/src/main/java/org/openhab/binding/openthermgateway/handler/BaseDeviceHandler.java
andrewfg: If the bridge is online and the thing is enabled then the thing is also online. Otherwise not.

bundles/org.openhab.binding.openthermgateway/src/main/java/org/openhab/binding/openthermgateway/handler/BaseDeviceHandler.java
andrewfg: Is it documented somewhere what is done automatically by the framework?

andrewfg: > If you leave initialize() with at least status UNKNOWN, this is not needed

@jlaur I think your statement is not true; if you examine the code of `bridgeStatusChanged` (below) you can see that the status is only updated if the thing has ThingStatusDetail.BRIDGE_OFFLINE

```
            updateStatus(ThingStatus.ONLINE, ThingStatusDetail.NONE);
        }
    }
```

jlaur: @fwolter - maybe you can shed a bit more light on this? Was remembering this comment: https://github.com/openhab/openhab-addons/pull/12061#discussion_r791120507. Now not sure if I introduced a bug in hdpowerview, and if more framework code is involved and will handle this. Was not able to find documentation.

fwolter: @jlaur is right. @andrewfg the status is also changed when the bridge becomes offline.

andrewfg: @fwolter & @jlaur can you please explain this more? I still think your statements are wrong. The code that your comments are pointing to, is in the thing handler’s initialize() method. Yes of course the bridgeStatusChanged() in the base thing handler does change the thing state to match the bridge state, BUT IT ONLY DOES SO IF THE THING HAS ALREADY REACHED THE STATE ONLINE, OR OFFLINE/BRIDGE-OFFLINE, otherwise the thing status is NOT UPDATED, and the thing never goes online.

To be more specific, I tried removing the code that you are asking me to, AND THE BINDING DOES NOT WORK. So I will DEFINITELY NOT make the change you are asking for.

fwolter: The framework updates the ThingStatuses of the child Things only if they are ONLINE, OFFLINE or *UNKNOWN*. If you leave `initialize()` with either of these, it should work.

andrewfg: > framework updates the ThingStatuses of the child Things only if they are .. UNKNOWN

@fwolter please point me to the code in the framework where it updates the state if the thing state is UNKNOWN. I have tested setting it at UNKNOWN, and the thing does NOT get updated.

> framework updates the ThingStatuses of the child Things only if they are ONLINE, OFFLINE

Exactly. It is for that exact reason that I am setting it so in the initialize() method. So why do you want me to NOT do that?

fwolter: This calls `initialize()` on the Things when the bridge is initialized: https://github.com/openhab/openhab-core/blob/87ecfade7473b9c5c05c80a521679fea2c08ad6b/bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/internal/ThingManagerImpl.java#L235 The actual online/offline/unknown check is here: https://github.com/openhab/openhab-core/blob/cabb3f73150d13f44d647f46af8362e396684b92/bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/util/ThingHandlerHelper.java#L37

I just tested it with the LCN binding and it works.

andrewfg: @fwolter many thanks for the insights. But can you please be specific: what specifically do you want me to change in the thing’s initialize() method? It looks fine to me. And it works fine. And I don’t understand what you don’t like about it. Please help me!

fwolter: These lines can be safely removed:
```
} else if (bridge.getStatus() != ThingStatus.ONLINE) {
```
It's just an optimization. I won't insist on changing it.

fwolter:
Yes, how can I help?

andrewfg:
> how can I help?

@fwolter the merge conflict is due to two files..
- one is a renamed class where it seems that the changes were so big that GitHub cannot recognise

fwolter:
There were other commits in the meantime, modifying the two files. That's why they are conflicting in this PR. @ArjenKorevaar you need to rebase your branch with `git rebase upstream/main` and resolve the conflicts by hand when executing the command.

fwolter:
Sounds good.

ArjenKorevaar:
Sounds good to me too. I am currently abroad and won't be able to work on this for atleast another couple of weeks.

bundles/org.openhab.binding.openthermgateway/src/main/java/org/openhab/binding/openthermgateway/handler/BaseDeviceHandler.java
fwolter: For the posterity, this works in #12330.

fwolter:
Closed in favor of #12367.

Issue #11102, opened 2021-08-13
[Telegram] Rule execution aborted after strange SocketTimeoutException after switch to OH3


foobar26:
I use the `telegramAction.sendTelegram()` in several of my rules. When I was running on OH2 everything worked fine. After I switched to OH3, I rarely get scripts not being fully executed and being aborted when the telegramAction is fired. I suspect the issue within the telegram binding, because I added loggings before and after the telegramAction and can see that the logging after it is not executed.

This is how the rule looks like:
```

   logInfo("Shadowing", "Start shadowing")
   val telegramAction = getActions("telegram","telegram:telegramBot:Telegram_Bot")
   telegramAction.sendTelegram(XXX, "Start shadowing")
   logInfo("Shadowing", "Telegram sent")


```
This produces the following log (as said in rare cases):

```
2021-08-11 08:37:25.807 [INFO ] [.openhab.core.model.script.Shadowing] Start shadowing
2021-08-11 08:38:40.367 [ERROR] [internal.handler.ScriptActionHandler] - Script execution of rule with UID 'shadowing-7' failed: java.net.SocketTimeoutException: timeout in shadowing

```

I don't see any stack trace or anything more than that in the logs.
Internet connectivity at that time was existing. OH is running on a server directly connected to the internet, so also no WIFI or something in between which might cause network issues.

## Expected Behavior
rule not to be aborted, detailed error report

## Current Behavior
rule being aborted after the error occurs. Telegram message not sent out.

## Possible Solution
In some older posts I found there was someone providing a beta version of the binding with a changed connect timeout. I didn't find a way to configure connect timeouts for recent versions of the binding. It might help to increase the timeout maybe.

## Steps to Reproduce (for Bugs)
Hard to reproduce even for me. This happens one times out of one hundred calls of the rule I would guess.

## Context
In the concrete case above I just want to send Telegram messages when I start shadowing my rooms and pulling down shutters. I have a light sensor running which is monitored by OH and this triggers the shadowing after some threshold is reached (brightness over some time).
But I have a bunch of other rules running with different intent and the problem occurs randomly in all rules in very rare cases. But this makes rule execution unreliable.

## Your Environment
openHAB 3.1.0 (stable)
Telegram Binding 3.1.0
OpenJDK 64-Bit Server VM (build 11.0.11+9-Ubuntu-0ubuntu2.18.04, mixed mode, sharing)

Skinah:
[...]

PR #11215, opened 2021-09-07
[telegram] Fixes exceptions that stop rules/actions from finishing

### LINKED ISSUE #11048: [telegram] Update chat id before message text and other channels
## Expected Behavior
In a "LastMessageText was updated to /something" rule, the script should be able to run something like `sendTelegram(Long::parseLong(TelegramBot_ChatId.state.toString), …)` and that should send the message to the chat where the command came from.

## Current Behavior

ChatId is updated *after* LastMessageText:
https://github.com/openhab/openhab-addons/blob/487dc0e49fb83282dc77d823869945d706ca1d19/bundles/org.openhab.binding.telegram/src/main/java/org/openhab/binding/telegram/internal/TelegramHandler.java#L327-L340

So when the last chat changes, one message goes to the wrong chat.

## Possible Solution

Reorder these lines of code?

Skinah:
+ Fixes a NPE causing action/rule to not complete, reported on forum here https://community.openhab.org/t/solved-telegram-sendtelegramanswer-works-but-error-in-log-file-null-in-xxx/126039/14

+ Possible fix for #11102 which is a SocketTimeoutException that causes an action to not complete.

+ Closes #11048 Where the order that channels are updated can causes rule complexity and prone to errors.

### DIFF
--- a/bundles/org.openhab.binding.telegram/src/main/java/org/openhab/binding/telegram/internal/TelegramHandler.java
+++ b/bundles/org.openhab.binding.telegram/src/main/java/org/openhab/binding/telegram/internal/TelegramHandler.java
@@ -324,20 +324,20 @@ public class TelegramHandler extends BaseThingHandler {
                             update.callbackQuery().data());
                 }
             }
-            updateChannel(LASTMESSAGETEXT, lastMessageText != null ? new StringType(lastMessageText) : UnDefType.NULL);
+            updateChannel(CHATID, chatId != null ? new StringType(chatId.toString()) : UnDefType.NULL);
+            updateChannel(REPLYID, replyId != null ? new StringType(replyId) : UnDefType.NULL);
             updateChannel(LASTMESSAGEURL, lastMessageURL != null ? new StringType(lastMessageURL) : UnDefType.NULL);
-            updateChannel(LASTMESSAGEDATE, lastMessageDate != null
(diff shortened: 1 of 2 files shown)


fwolter:
The code doesn't compile, because it generates a code analysis error.

@ZzetT Are you ok with these changes?

Skinah:
code analysis Report in the target folder is empty so appears all good here, suspect it will pass if I trigger a rebuild in a few days time as all the latest PR are failing.

fwolter:
I get this error:
```
[ERROR] org.openhab.binding.telegram.internal.action.TelegramActions.java:[138]
Count of placeholder(2) is not equal to count of parameter(3)
```

Skinah:
Thanks I have fixed the issue, but for some reason I get this missing warning that must be stopping the code analysis from working.

```
[INFO] Fork Value is true
     [java] The following classes needed for analysis were missing:
     [java]   android.os.Build$VERSION
     [java]   android.util.Log
     [java]   org.conscrypt.Conscrypt
     [java]   org.conscrypt.Conscrypt$ProviderBuilder
[INFO] Done SpotBugs Analysis....
```

fwolter (approved):
Strange I never saw this error message.

However, @zzett please raise your hand if you feel this shouldn't have been merged for whatever reason.

PR #8902, opened 2020-10-29
[Miele] Fixes #8877

kgoderis:
Signed-off-by: Karel Goderis <karel.goderis@me.com>

PR #11218, opened 2021-09-07
[miele] Fix broken things file support.

### LINKED ISSUE #8877: [miele] Things via .things broken for ZigBee protocol devices
https://github.com/openhab/openhab2-addons/commit/4c6ad1efef157735fec858f85e549464e36b9727 breaks the Miele binding for Zigbee devices defined via a .things file. @MartinLepsy  are you still around?

jlaur:
Fixes #8877

This is a MVP bugfix for broken support of things files.

This is an alternative solution to #8902 where protocol information is kept as a property instead of being added to configuration. This has the advantage of being transparent for users, i.e. no need to configure anything.

From code perspective protocol is kept as either **ZigBee** or **LAN** without introducing further protocol confusion by adding a third option (**Wi-Fi**) being mapped to LAN.

PR can be merged as-is for fixing the root cause, in which case I plan to provide another PR for simplifying and abstracting the protocol (as separate issue). In case of impatience from my side, I might also consider adding more commits into this PR. Let me know what you think.

Manual testing performed for:
- ZigBee device configured through UI.
- ZigBee device configured through file.
- Wi-Fi device configured through UI.
- Wi-Fi device configured through file.

### DIFF
--- a/bundles/org.openhab.binding.miele/src/main/java/org/openhab/binding/miele/internal/handler/MieleApplianceHandler.java
+++ b/bundles/org.openhab.binding.miele/src/main/java/org/openhab/binding/miele/internal/handler/MieleApplianceHandler.java
@@ -143,8 +143,7 @@ public abstract class MieleApplianceHandler<E extends Enum<E> & ApplianceChannel
 
     @Override
     public void onApplianceStateChanged(String UID, DeviceClassObject dco) {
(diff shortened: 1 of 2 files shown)


kaikreuzer (approved):
This looks like a pretty straight forward fix. As you say that you have tested it with all possible constellations, I'd say this is good to merge. Many thanks for your work!

PR #11226, opened 2021-09-08
[bsblan] Several configuration parameters are not visible in Main UI

hypetsch:
The main UI in OH3 seems to not show configuration parameters of Things when `groupName` is specified but the related `parameter-group` node is missing in the definition XML.
Could possibly be a change in behavior compared to the OH 2.5 (?) UI?

A similar issue has already been discussed here: [V2 vs V3 - Option Parameter not showing](https://community.openhab.org/t/solved-v2-vs-v3-option-parameter-not-showing/115685/3)

Signed-off-by: Schraffl Peter <p.schraffl@gmx.at>


lolodomo (approved):
LGTM

lolodomo:
We should check if the same fix is not required for other bindings.

PR #11227, opened 2021-09-08
[amazonechocontrol] Fix announcment with TTS volume

mgeramb:
This PR fix an issue happens through a change from amazon. It uses a workaround which is commented in the code.


mgeramb:
This PR beaks Echo Dot Gen 1 -> closed

Issue #6658, opened 2019-12-22
[enocean] Support EEP-Profile D2-06-01 (Multisensor Window Handle / Soda window handle)


blumentarzan:
It would be great to support the Soda window handle in openHAB by default. Thanks! Adrian

DrRSatzteil:
Maybe I can have a go at this. Just finished D2_06_50 so this should be similar

blumentarzan:
This would be great! Still have no solution for that... If I could help with testing just tell me.

DrRSatzteil:
Good if you're still interested. Since I do not own one of their devices you would need to do the testing.  The specifications for D2-02-01 are rather special since it theoretically supports a large number of different sensors (may differ from device to device) so depending on your devices abilities we might not be able to test all features of this EEP. I'll look some more info that in the next couple of days and will let you know when you can help me out with something.

DrRSatzteil:
Hi @blumentarzan 

I finished a very first version of this EEP that you might have a look at.

Please do not expect anything to work since I cannot test this myself. Hence it would be great if you could try to discover a device, attach the thing channels to items and play around a bit with it. Please watch the logs and paste any errors here. It might also help to adjust the log level (ssh to your OH on port 8101 e.g. "ssh openhab@localhost -p 8101" with password habopen and type: log:set debug org.openhab.binding.enocean)

A couple of notes:
* This implementation is currently read-only! That means that you won't be able to adjust any settings of your handle but only read sensor data from it.
* According to the specifications these devices potentially support a large amount of sensors (multiple buttons, motion sensor, humidity, temperature, etc.) however these may not all be present in every device. However you will see all channels not matter if the corresponding feature is supoorted by your device or not. If a channel is not supported it should never return any values. Right now I don't think this binding is designed to dynamically remove non-existent channels but this would be something for a later stage anyway.  

For installation you should follow the following steps:
1. Remove the binding via UI (Note that all your devices will be temporarily offline if you use your production instance so make sure that this does not cause any trouble)
2. Wait until your enOcean bridge and devices complain about a missing handler (watch things in UI)
3. Move this file to your addons folder:
https://drive.google.com/file/d/1LosRGjvgjCgA83lLueWGRqN35McxdBVF/view?usp=sharing
4. Wait a couple of seconds
5. Move this file to your addons folder  
https://drive.google.com/file/d/1vveOtN7u2yq7gLMSu9XY4Yvone2osSfW/view?usp=sharing
[...]

PR #11230, opened 2021-09-08
Add support for Soda Handles (EEP D2-06-01)

DrRSatzteil:
This PR adds support for the D2-06-01 EEP used by the window handles of Soda GmbH.

Fixes issue #6658 (please see discussion there for more details on this PR)


DrRSatzteil:
Uh oh I just realised that there are some commits included that should not be there. I'll convert this to a draft and fix these issues before a review makes sense.

DrRSatzteil:
These commits didn't do any harm but I reverted them anyway...

bundles/org.openhab.binding.enocean/README.md
```diff
@@ -99,6 +100,8 @@ Hence if your device supports one of the following EEPs the chances are good tha
 
 ² These are just examples of supported devices
 
+³ Note that the soda handles potentially contain a wide range of different sensors and buttons. However the amount of built-in sensors and buttons may vary between different models. In case your particular device does not contain one of the potentially supported features the corresponding channel will never trigger an update. Please see the manual of your particular model to check which channels should be supported before opening an issue.  
+
```
lolodomo, line 104: Please return after each sentence.

bundles/org.openhab.binding.enocean/README.md
```diff
@@ -255,11 +258,12 @@ The channels of a thing are determined automatically based on the chosen EEP.
...
 | pushButton          | Trigger            | Channel type system:rawbutton, emits PRESSED and RELEASED events |
+| pushButton2         | Trigger            | Channel type system:rawbutton, emits PRESSED and RELEASED events |
 | doublePress         | Trigger            | Channel type system:rawbutton, emits PRESSED |
 | longPress           | Trigger            | Channel type system:rawbutton, emits PRESSED and RELEASED events |
 | rockerswitchA/B     | Trigger            | Channel type system:rawrocker, emits DIR1_PRESSED, DIR1_RELEASED, DIR2_PRESSED, DIR2_RELEASED events |
-| windowHandleState   | String             | Textual representation of handle position (OPEN, CLOSED, TILTED) |
-| windowSashState     | String             | Textual representation of sash position (OPEN, CLOSED, TILTED) |
+| windowHandleState   | String             | Textual representation of handle position (OPEN, CLOSED, TILTED or UP, DOWN, LEFT, RIGHT for the D2_06_01 EEP) |
+| windowSashState     | String             | Textual representation of sash position (OPEN, CLOSED, TILTED or TILTED, NOT TILTED for the D2_06_01 EEP) |
```
lolodomo, line 266: In the description, maybe reorder it with something like: ths, and this for the D2_06_01 EEP and this, this this for all others

bundles/org.openhab.binding.enocean/src/main/java/org/openhab/binding/enocean/internal/eep/D2_06/D2_06_01.java
```diff
@@ -0,0 +1,194 @@
...
+import org.openhab.core.library.types.QuantityType;
+import org.openhab.core.library.types.StringType;
+import org.openhab.core.library.unit.SIUnits;
+import org.openhab.core.library.unit.Units;
+import org.openhab.core.thing.CommonTriggerEvents;
+import org.openhab.core.types.State;
+import org.openhab.core.types.UnDefType;
+
+/**
+ *
```
lolodomo, line 33: Please add a description of your class

bundles/org.openhab.binding.enocean/src/main/java/org/openhab/binding/enocean/internal/eep/D2_06/D2_06_01.java
```diff
@@ -0,0 +1,194 @@
...
+ * @author Thomas Lauterbach - Initial contribution
+ */
+public class D2_06_01 extends _VLDMessage {
```
lolodomo, line 36: Please add @NonNullByDefault

bundles/org.openhab.binding.enocean/src/main/resources/OH-INF/thing/channels.xml
```diff
@@ -216,6 +221,29 @@
 		</event>
 	</channel-type>
 
+	<channel-type id="protectionPlusEvent">
+		<kind>trigger</kind>
+		<label>Protection Plus Event</label>
+		<description>Triggered when a Protection Plus capable device detects a break-in attempt.</description>
+		<event>
+			<options>
+				<option value="ALARM">alarm</option>
```
lolodomo, line 230: A detzil but you could put the first letter of your label in uppercase like for the other triggers.

bundles/org.openhab.binding.enocean/src/main/resources/OH-INF/thing/channels.xml
```diff
@@ -216,6 +221,29 @@
...
+	</channel-type>
+
+	<channel-type id="vacationModeToggleEvent">
+		<kind>trigger</kind>
+		<label>Vacation Mode Toggle Event</label>
+		<description>Triggered when the vacation mode has been toggled on the device.</description>
+		<event>
+			<options>
+				<option value="ACTIVATED">activated</option>
+				<option value="DEACTIVATED">deactivated</option>
```
lolodomo, line 242: A detzil but you could put the first letter of your label in uppercase like for the other triggers.

lolodomo (changes requested):

bundles/org.openhab.binding.enocean/src/main/java/org/openhab/binding/enocean/internal/eep/D2_06/D2_06_01.java
DrRSatzteil: Hi @lolodomo 

I added the annotation but I'm struggling with the annotation of the getCurrentStateFunc parameter of the convertToStateImpl method. If I annotate it with Nullable it will still give me an Exception due to a mismatched annotation. I think this might be due to the fact that State is annotated as NonNull but I can't figure out how to fix that... Any help would be highly appreciated, thank you!

lolodomo: Maybe the problem is due to parent classes not being null annotated? 
Did you try to do annotate them too? 
If this is the problem, maybe we can release my comment as my intention was not to tell you to annotate many classes.

DrRSatzteil: This could be the case but it would go way beyond the scope of this PR to annotate large parts of the binding (no null annotations are used so far).

The correct way to do this for my single class (as the start of an iterative approach) would be as I understand it from the documentation to annotate the parameters with Nullable. This works fine for all parameter except for the Function<String, State> parameter with its generic types. What's special is that the State type is an openHAB core class that is properly annotated while the others are plain basic Java classes that are not annotated at all. I tried a lot of different things without really knowing what I'm doing and finally gave up yesterday night...

lolodomo: Ok so please forget this in the context of this PR and push your updates.

lolodomo (approved):
LGTM

lolodomo:
@fruggy83 : do you like having a look on this PR before merging it?

fruggy83:
@lolodomo Yes, I have found some issues. I will do the review later.

bundles/org.openhab.binding.enocean/src/main/java/org/openhab/binding/enocean/internal/eep/D2_06/D2_06_01.java
```diff
@@ -0,0 +1,198 @@
...
+        }
+        return UnDefType.UNDEF;
+    }
+
+    @Override
+    protected String convertToEventImpl(String channelId, String channelTypeId, String lastEvent,
+            Configuration config) {
+
+        // Sensor values
+        if (bytes[0] == 0x00) {
```
fruggy83, line 128: To support readability I would add an enum (or at least a const) for the different message types with their id. You could replace all the other magic numbers (at least the different states like sash, handle, motion) with a const, too.

bundles/org.openhab.binding.enocean/src/main/java/org/openhab/binding/enocean/internal/eep/EEPType.java
```diff
@@ -195,6 +196,10 @@
             CHANNEL_WINDOWHANDLESTATE, CHANNEL_CONTACT, CHANNEL_BATTERY_VOLTAGE),
     MechanicalHandle03(RORG._4BS, 0x14, 0x0A, false, A5_14_0A.class, THING_TYPE_MECHANICALHANDLE,
             CHANNEL_WINDOWHANDLESTATE, CHANNEL_CONTACT, CHANNEL_VIBRATION, CHANNEL_BATTERY_VOLTAGE),
+    MechanicalHandle04(RORG.VLD, 0x06, 0x01, false, "Soda", 0x0043, D2_06_01.class, THING_TYPE_MECHANICALHANDLE,
```
fruggy83, line 199: If it is not neccessary I would not narrow the EEP to manufacturer Soda. This feature is just needed for manufactures who implement an EEP in a different way as normal. This is in most cases just true for elt... (the blue one ;)

bundles/org.openhab.binding.enocean/src/main/resources/OH-INF/thing/channels.xml
```diff
@@ -211,7 +216,30 @@
...
+		<description>Triggered when a Protection Plus capable device detects a break-in attempt.</description>
+		<event>
+			<options>
+				<option value="ALARM">Alarm</option>
+			</options>
+		</event>
+	</channel-type>
+
+	<channel-type id="vacationModeToggleEvent">
+		<kind>trigger</kind>
```
fruggy83, line 236: Is this really an event or can it be implemented as a state? Does the handle really send a "no change" (0x0) message for the vacation mode or just 0x1 and 0x2?

fruggy83 (changes requested):
Hi @DrRSatzteil 

thanks a lot for your work. Do you own such a handle? Would it be possible to implement sending control and settings messages (type 0x80) to the handle too? Especially enabling the vacation mode remotely would be a nice feature.

DrRSatzteil:
> Hi @DrRSatzteil
> 
> thanks a lot for your work. Do you own such a handle? Would it be possible to implement sending control and settings messages (type 0x80) to the handle too? Especially enabling the vacation mode remotely would be a nice feature.

No I do not own one of these devices. @blumentarzan did the testing so far. I think implementing the other messages wouldn't be a huge problem however the timing for sending these messages seems to require some additional logic: The specification states that the handle is in receive mode for only 500ms after a message was sent so the binding would need to queue messages and get notified when a message was received to send out the queued messages. I don't think that this functionality is already there. It's probably not a huge effort but would likely require some fundamental design decisions being made. So I'm not sure if this should be part of the same PR.

bundles/org.openhab.binding.enocean/src/main/resources/OH-INF/thing/channels.xml
DrRSatzteil: According to the specification this can only be implemented as an event: This event will only be emitted when the physical vacation mode button is pressed on the handle, it will not be sent when the vacation mode is set via a configuration message. Also (again according to the specs) the event is not repeated but only sent once.

On this screenshot from #6658 you can see that the handle indeed sends "No change" if the button was not pressed: https://user-images.githubusercontent.com/17730574/127765547-6209aa03-1b4e-48b5-b37d-14c60cf685d8.png

bundles/org.openhab.binding.enocean/src/main/java/org/openhab/binding/enocean/internal/eep/EEPType.java
DrRSatzteil: Since Soda created this EEP specification it should be fine to remove the manufacturer specific code. However since I don't own the device I cannot test this myself. Do you think it is it safe to remove these parameters?

bundles/org.openhab.binding.enocean/src/main/java/org/openhab/binding/enocean/internal/eep/D2_06/D2_06_01.java
DrRSatzteil: I added some enums for the message type and the sash, handle and motion states.

lolodomo:
@fragger42 : let us know when this PR is ok for you.

fruggy83:
Hi @DrRSatzteil this message receive and answer logic (request/response) is already implemented in the binding. However without owning such a device it is hard to implement this feature. So I am ok with your changes. Thanks a lot.

DrRSatzteil:
Ok that's great I didn't know that. We might add this functionality later if @blumentarzan or someone else can do some more testing.

Issue #10360, opened 2021-03-21
[linky] Authentication failing (step 3)


lolodomo:
We have again a problem to connect to the service.
It is failign at step 3 and the thing is OFFLINE (COMMUNICATION_ERROR ) with this detail: 
```
Connection failed step 3 - auth1 : <html><head><title>Votre requete n'a pas pu aboutir.</title></head><body>Votre requete n'a pas pu aboutir.</body></html>
```
I checked my "internalAuthId" is correct.

@clinique: Gaël: do you have the same problem ?

I suspect that the URL built by the binding has to be updated.

clinique:
Hello @lolodomo , yes there may be another change on Enedis website :-( Sorry but I'm currently moving and will not have time to investigate this in the coming weeks.

lolodomo:
The URL used at step 3 (auth1) to get the auth data now returns a 500 status code.

lolodomo:
@clinique : do you remember where you find your inspiration when you did the last refactoring ? I am curious if the authentification process was fixed in that place.

If no solution is found, I think we should suppress this binding from next 3.2 distributions. WDYT ?

clinique:
@lolodomo : IIRW I think I did observe behaviour of the browser using debug console. Sure we can not rely on such a fragile approach and unless a more robust is provided, it could be removed as still it does not work at all.

lolodomo:
What's a pitty, we spend time on this binding.

clinique:
You're right, but it may help sometimes in the future.

lolodomo:
> I did observe behaviour of the browser using debug console

Maybe we can try that again.

lolodomo:
After a deeper analysis, I have the feeling the problem is the cookie we are building.

PR #11233, opened 2021-09-09
[linky] Removed from the distribution until it is fixed

lolodomo:

### DIFF
--- a/bom/openhab-addons/pom.xml
+++ b/bom/openhab-addons/pom.xml
@@ -741,11 +741,13 @@
       <artifactId>org.openhab.binding.lifx</artifactId>
       <version>${project.version}</version>
     </dependency>
+    <!-- linky binding suppressed from the distribution until it is fixed
     <dependency>
       <groupId>org.openhab.addons.bundles</groupId>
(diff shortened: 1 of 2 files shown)


fwolter:
Can you add a reference what is broken?

lolodomo:
The authentication code is broken. 
Look at issue #10360 for more details,

clinique (approved):

fwolter:
Are you sure @clinique and you won't be able to fix it until December, when OH 3.2 is released?

lolodomo:
I seriously doubt it will be fixed soon as noone is working on it and/or has the beginning of a solution.
In case a solution is found, we could then remove the comments I added in this PR.

fwolter:
Ok, I think we should take down the binding documentation, too, then.

fwolter:
Can you run spotless?

fwolter (approved):

kaikreuzer:
This PR broke our build, see https://ci.openhab.org/job/sandbox-openhab3-release/363/execution/node/354/log/.
I guess the feature.xml needs to be removed/renamed as well, so that it isn't picked up by the feature aggregation.

PR #11238, opened 2021-09-11
[magentatv] Fix: Start UPNP listener to catch power off button, thing id in log

markus7017:
This PR fixes a small bug, which didn't started the UPnP listener. This is used to catch the situation where the receiver is switch off with the physical power button


lolodomo (approved):

lolodomo:
Build was aborted due to timeout.

PR #11240, opened 2021-09-12
[loxone] Support for SHA-256

ppieczul:
Added support for SHA-256, without this the recent Miniserver update will reject to assign authentication tokens.

Signed-off-by: Pawel Pieczul <pieczul@gmail.com>


lolodomo:
One question: why using Boolean rather than boolean?

ppieczul:
No reason. It can be boolean (except for the class parsed from JSON). Should I change it?

lolodomo:
I think it would be better if you can.

lolodomo (approved):

PR #11247, opened 2021-09-13
[xmppclient][WIP] Add send image throught HTTP

gkfabs:
When I have an alert on my phone, I would like to be able to see an image which comes from the corresponding camera. For example when I feed the fishs or when I am not at home, the security camera corresponding to the PIR.

At first I wanted to integrate the jingle file transfert. But I think it will take time to see it in Smack https://github.com/igniterealtime/Smack/pull/489

This is why I did a small improvement using https://github.com/igniterealtime/Smack/tree/master/smack-experimental/src/main/java/org/jivesoftware/smackx/httpfileupload

But I am getting some problems to add a new library: smack-experimental.

Thanks


pavel-gololobov (approved):
Looks good. Thanks.

lolodomo:
I let another maintainer merge this PR as I am not sure about the dependencies stuff (bcprov).

lolodomo:
@wborn : can you have a look on this one?

fwolter:
You should be able to set the scope to `provided` in the pom.xml, if you add the dependency via features. Otherwise the dependency will be added to the binding's JAR *and* downloaded separately when installing the binding, AFAIK.

gkfabs:
@fwolter Do you want me to set all the smack libs dependencies to provided or only the new one smack-experimental?

fwolter:
Each dependency, which exists in features.xml.

PR #11258, opened 2021-09-18
[kaleidescape] Improve discovery service

mlobstein:
Implement listening for UDP broadcast packets from Kaleidescape components for discovery instead of scanning all IP addresses in the subnet. This also allows for the removal of Apache SubnetUtils.


kaikreuzer (approved):
lgtm, thanks!

PR #11260, opened 2021-09-18
[hdpowerview] Add Hub configuration option hardRefreshBatteryLevel

### LINKED ISSUE #11259: [hdpowerview] Create option for refreshing battery status more frequently
Battery level for each shade is updated automatically by the Hub once a week. However, it is possible to force refresh values by a special API call.

Quote from documentation:
_Since Shades are the central resource in the PowerView eco-system, it is useful to understand a few
key state relationships between Hubs and Shades. Hubs maintain transient shade state such as
position and battery level. Any API call that returns the shade position and/or the battery level is
delivering the last Hub saved value of these attributes. In general, shades are moved via API calls to
Shades, Scenes, and Scene Collections. Since the Hub is always involved in these motion events, it
tracks the final shade position and saves it; however, shades can be moved without the Hub’s
knowledge. An individual can manually move a shade simply by pressing the motion button on the
side of the shade. In addition, shades can be moved using a PowerView Motorization handheld
remote control device. In both of these cases, the Hub is not told of the shade’s new position.
Consequently, to ensure an accurate Hub view of shade position, an application may choose to call
the GET /shades/{id} API with the refresh query string set to true. This forces the Hub to query that
shade for its current position and save it. Similarly, the battery level is transient and is only updated
automatically by the Hub once a week. If an application wants to synchronize the battery level state
sooner, then the application may choose to call the GET /shades/{id} API with the
updateBatteryLevel query string set to true._

A new option for refreshing more often would be nice to be able to track for example voltage more closely.

## See also
- #11154 
- #11200 
- #11205

jlaur:
Add Hub configuration option **hardRefreshBatteryLevel** for refreshing battery status more frequently.

Also, when battery data is missing or invalid, explicitly update channels to UNDEF. Previously, **lowBattery** and **batteryLevel** would be skipped, and **batteryVoltage** would be updated to 0 V.

Fixes #11259

### DIFF
--- a/bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/handler/HDPowerViewHubHandler.java
+++ b/bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/handler/HDPowerViewHubHandler.java
@@ -165,7 +165,7 @@ public class HDPowerViewHubHandler extends BaseBridgeHandler {
             future.cancel(false);
         }
         if (hardRefreshBatteryLevelInterval > 0) {
-            logger.debug("Scheduling hard battery level refresh every {}hours", hardRefreshBatteryLevelInterval);
+            logger.debug("Scheduling hard battery level refresh every {} hours", hardRefreshBatteryLevelInterval);
             this.hardRefreshBatteryLevelFuture = scheduler.scheduleWithFixedDelay(
                     this::requestRefreshShadeBatteryLevels, 1, hardRefreshBatteryLevelInterval, TimeUnit.HOURS);
         }
--- a/bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/handler/HDPowerViewHubHandler.java
+++ b/bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/handler/HDPowerViewHubHandler.java
@@ -67,11 +67,13 @@ public class HDPowerViewHubHandler extends BaseBridgeHandler {
     private final HttpClient httpClient;
 
     private long refreshInterval;
-    private long hardRefreshInterval;
+    private long hardRefreshPositionInterval;
+    private long hardRefreshBatteryLevelInterval;
 
     private @Nullable HDPowerViewWebTargets webTargets;
     private @Nullable ScheduledFuture<?> pollFuture;
-    private @Nullable ScheduledFuture<?> hardRefreshFuture;
+    private @Nullable ScheduledFuture<?> hardRefreshPositionFuture;
+    private @Nullable ScheduledFuture<?> hardRefreshBatteryLevelFuture;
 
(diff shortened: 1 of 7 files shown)


bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/handler/HDPowerViewHubHandler.java
```diff
@@ -147,14 +150,24 @@ private void schedulePoll() {
...
+        future = this.hardRefreshPositionFuture;
         if (future != null) {
             future.cancel(false);
         }
-        if (hardRefreshInterval > 0) {
-            logger.debug("Scheduling hard refresh every {}minutes", hardRefreshInterval);
-            this.hardRefreshFuture = scheduler.scheduleWithFixedDelay(this::requestRefreshShades, 1,
-                    hardRefreshInterval, TimeUnit.MINUTES);
+        if (hardRefreshPositionInterval > 0) {
+            logger.debug("Scheduling hard position refresh every {}minutes", hardRefreshPositionInterval);
```
kaikreuzer, line 158:
```suggestion
            logger.debug("Scheduling hard position refresh every {} minutes", hardRefreshPositionInterval);
```

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/handler/HDPowerViewHubHandler.java
```diff
@@ -147,14 +150,24 @@ private void schedulePoll() {
...
+            this.hardRefreshPositionFuture = scheduler.scheduleWithFixedDelay(this::requestRefreshShadePositions, 1,
+                    hardRefreshPositionInterval, TimeUnit.MINUTES);
+        }
+
+        future = this.hardRefreshBatteryLevelFuture;
+        if (future != null) {
+            future.cancel(false);
+        }
+        if (hardRefreshBatteryLevelInterval > 0) {
+            logger.debug("Scheduling hard battery level refresh every {}hours", hardRefreshBatteryLevelInterval);
```
kaikreuzer, line 168:
```suggestion
            logger.debug("Scheduling hard battery level refresh every {} hours", hardRefreshBatteryLevelInterval);
```

kaikreuzer (approved):
lgtm, thanks!

jlaur:
@kaikreuzer - funny, locally I made exactly the same formatting changes as you, but reverted as it seemed somehow intended. You missed the first one though:

`logger.debug("Scheduling poll for 5000ms out, then every {}ms", refreshInterval);`

I'll create a new PR for this for consistency.

kaikreuzer:
Why do you think it was intended? For "ms", yes, I could imagine as people sometimes write the unit directly after the number, but for minutes and hours it simply looks wrong.

> I'll create a new PR for this for consistency.

Ok, thanks!

jlaur:
@kaikreuzer - only reason why it might have been intended was that it was consistent for "ms" and "minutes". To me it also looked wrong. You are right that some prefer unit after number without space. We can also decide to leave it, what do you think?

kaikreuzer:
I do not care too much - feel free to decide yourself :-)

PR #11261, opened 2021-09-18
Upgrade to Karaf 4.3.3

wborn:
* Syncs the karaf.version so the new Maven plugin is used
* Resolves itest runbundles for the new runtime dependencies

---

Related to openhab/openhab-distro#1315


kaikreuzer (approved):
Thanks!

PR #11262, opened 2021-09-18
[lifx] Support HEV clean cycle

wborn:
This PR adds support to the binding for starting/stopping the [LIFX Clean](https://www.lifx.com/products/lifx-clean-antibacterial-smart-light) HEV cycle:

* Implement HEV packets
* Add colorhevlight thing type with a hevcycle channel
* Update documentation


bundles/org.openhab.binding.lifx/src/main/java/org/openhab/binding/lifx/internal/handler/LifxLightHandler.java
```diff
@@ -486,121 +513,129 @@ private void sendPacket(Packet packet) {
...
+                if (command instanceof PercentType) {
+                    return () -> handleTemperatureCommand((PercentType) command);
+                } else if (command instanceof IncreaseDecreaseType) {
+                    return () -> handleIncreaseDecreaseTemperatureCommand((IncreaseDecreaseType) command);
+                } else {
+                    return null;
+                }
+            default:
+                try {
+                    if (channelUID.getId().startsWith(CHANNEL_COLOR_ZONE)) {
```
martinvw, line 607: It make sense to split up this part or other parts of this method, given it became pretty big.

martinvw (approved):

bundles/org.openhab.binding.lifx/src/main/java/org/openhab/binding/lifx/internal/handler/LifxLightHandler.java
wborn: Yes this whole class is actually getting too big so it means it is doing too much. Creating some subclasses for each thing type might result in nicer code and it would also group related code. :slightly_smiling_face:

PR #11267, opened 2021-09-19
[melcloud] Fix thing type id in documentation

DirkSander:
correct channel is "melcloud:heatpumpdevice:"


kaikreuzer (approved):
Thanks, merging without DCO as small patch exception.

PR #11275, opened 2021-09-20
[powermax] Add null annotations

lolodomo:


lolodomo:
This version is now tested and I see no particular issue.

lolodomo:
Oups, I just found a difference (bug) which I don't reproduce in the current version.
I don't understand yet how null annotations could introduce that but I have to analyze the problem.

PR #11289, opened 2021-09-22
[hue] Make sure bridge status is set when returning from initialize

kaikreuzer:
I just experienced my bridge being in INITIALIZING state, although the handler successfully returned from the `initialize()` method.

### DIFF
--- a/bundles/org.openhab.binding.hue/src/main/java/org/openhab/binding/hue/internal/handler/HueBridgeHandler.java
+++ b/bundles/org.openhab.binding.hue/src/main/java/org/openhab/binding/hue/internal/handler/HueBridgeHandler.java
@@ -685,6 +685,8 @@ public class HueBridgeHandler extends ConfigStatusBridgeHandler implements HueCl
                 hueBridge = new HueBridge(ip, hueBridgeConfig.getPort(), hueBridgeConfig.getProtocol(), scheduler);
                 hueBridge.setTimeout(5000);
 
+                updateStatus(ThingStatus.UNKNOWN);
+
                 // Try a first connection that will fail, then try to authenticate,
                 // and finally change the bridge status to ONLINE
                 initJob = scheduler.submit(new PollingRunnable() {


lolodomo:
@kaikreuzer : my feeling is there is probably one case leading the init job not setting any thing status. I think it would be better to fix this case (the init job should always set a proper status).
I see your fix more as a workaround, but of course it would avoid to have the thing in INITIALIZING state.

kaikreuzer:
You're clearly right, @lolodomo, but even if it is not "hanging" in this state, the Thing status needs to be set by the handler before returning from the method - so this fix is required, even if it is just about a few milliseconds.

The cause that led to this state isn't actually fully clear to me yet - the thing handler thread pool was exhausted, because various bindings were waiting for the HTTP calls through the common httpClient to return - but none of them did. For some reason, the Jetty client was hanging and the HTTP timeouts weren't considered. I have no clue how that can happen, but it clearly wasn't a bug of the hue binding.

lolodomo:
> You're clearly right, @lolodomo, but even if it is not "hanging" in this state, the Thing status needs to be set by the handler before returning from the method - so this fix is required, even if it is just about a few milliseconds.

I have in mind that he recommendation (and the comment in the binding template) is to have the method returning fast, not necessarly to have the method setting a thing status (fast). Maybe the comment should be updated, if required. I don't find the template anymore to check now.
I guess many bindings use the same pattern as the hue binding did, that is creating a thread to return fast (without setting any status) and letting the thread setting the thing status.
This approach with setting first the thing status to UNKNOWN is also used by bindings (I myself used it for several bindings I believe).
Maybe it should be clarified and if setting the thing status before returning from initialize() is important, we should check that all bindings respect this rule. I have a serious doubt.

lolodomo (approved):

kaikreuzer:
@lolodomo Here's the section and I think it is already correctly documented the way described above:

https://github.com/openhab/openhab-core/blob/main/tools/archetype/binding/src/main/resources/archetype-resources/src/main/java/internal/__bindingIdCamelCase__Handler.java#L69-L74

lolodomo:
Ok, perfect

jlaur:
@lolodomo, @kaikreuzer - should this be added to milestone 3.2?

ghys:
I have problems with my Hue lights since I upgraded my distro (I didn't have problems for years), and since this is the only commit that could be relevant I was wondering if it could be the culprit.
Everything works ok at first; but the next day, none of the lights work - when I operate them I notice this log entry:
```
[DEBUG] [.hue.internal.handler.HueLightHandler] - hue light not known on bridge. Cannot handle command.
```
Also, the bridge is in the UNKNOWN state (which is apparently related to this PR), and as a result all the Hue devices are in error as well.

I tried:
1) disabling/reenabling the bridge - doesn't work, the bridge stays in UNKNOWN state, and nothing of note in the logs
2) restarting the Hue binding - doesn't work either
3) restarting openHAB - then it works again, until it doesn't

I can control the lights with the Hue app or even another openHAB instance so it's not a problem with the network or the brige.

Would this PR be the cause, and how to debug this further?

Issue #11288, opened 2021-09-22
[influxdb] no datapoints found if metadata contains `item`


maglnet:
## Expected Behavior

Given the following item configuration:
```
Number:Temperature someStupidName (gTempSensors) { influxdb="temperature" [item="tempOutdoor"] }
```

I expect to read and write data to and from influxdb with a tag `item=tempOutdoor`.

## Current Behavior

- data is written to influxdb with tag `item=tempOutdoor`
- data is queried with `WHERE item = 'someStupidName';` for v1 and `filter(fn: (r) => r["item"] == "someStupidName")` for v2

## Possible Solution

Within the `createQuery` methods it may be possible to get the correct item name from metadata if available.
This would be managed the same way, like getting the measurementName from metadata if available here:

I'd create a PR with a possible solution soon.

https://github.com/maglnet/openhab-addons/blob/1f18dda30134066cb62e772b04353dc77f964929/bundles/org.openhab.persistence.influxdb/src/main/java/org/openhab/persistence/influxdb/internal/influx1/Influx1FilterCriteriaQueryCreatorImpl.java#L52
https://github.com/maglnet/openhab-addons/blob/1f18dda30134066cb62e772b04353dc77f964929/bundles/org.openhab.persistence.influxdb/src/main/java/org/openhab/persistence/influxdb/internal/influx2/Influx2FilterCriteriaQueryCreatorImpl.java#L52

## Steps to Reproduce (for Bugs)

see #11290 

## Context

the reason why I set an item name via the metadata is, that my OH-ItemNames are named by the hardware (not sure if that's good practice) but my influxdb data should have some more clear names to be able to create queries via the web interface easily.

## Your Environment

- OpenHAB 3.1.0
- Raspberry PI
- InfluxDB v2

lujop:
Not sure to want to accept that, because the current item and tag logic is quite complex to support migrated data from old versions and other edge cases. And I don't want to complicate it even more if it's not absolutely necessary and benefits common uses cases.

With current options, you can set the measurement name by using metadata, and you can also set arbitrary tag names. I think that with that you can have your data with good structure and meaning, despite not controlling the names of the items.

Why do you don't want to use custom tags instead of hacking with item tagname?

maglnet:
I really don't see that much complexity in the tag and item name logic during the query time, please check the linked PR that fixes the described issue with very low effort.

> Trying to override the item tag name it's not a documented neither expected behavior in my opinion.

[...]

PR #11290, opened 2021-09-22
[influxdb] - use item tag from metadata if available

maglnet:
Use item tag from metadata for queries when provided

If an item has metadata containing the tag `item` this should be used when querying data since the data has been written to influxdb with that tag.

for details see #11288


maglnet:
WIP status removed since it's working as expected.

lujop (changes requested):
I'm sorry, you have done a good job, PR it's very well done, clean, and tested.
But I don't agree with adding that behavior.
See my comments in the issue #11288

maglnet:
It's fine for me.
I'll create a new PR later to prevent others from making the same mistake as I did by adding a hint to the `README`.

PR #11295, opened 2021-09-23
[SagerCaster] Added semantic tags

clinique:
Make it ready for online translation of texts.
Small code improvements.
Used system channels when possible.


lolodomo:
I just opened the PR, I think your PR title is clearly not describing its content. The "semantic tag" is 1% of your changes ;)

And I do not understand why you used `@text` for channel state options, rather than using the usual way ?

clinique:
You're correct, it was my initial intent, but I also found some enhancements to proceed with and switched to the easiest translation mechanism. I did it this way for channel state options as it is easier to translate this way. I'm going to change the PR title.

lolodomo:
@clinique : can you please first resolve the conflicting files, I will then have a look.

clinique:
@lolodomo : you should now be ready to go.

clinique:
@lolodomo : build fails for errors out of this PR, you can proceed if/when you want.

PR #11298, opened 2021-09-23
[miele] Add new power and water consumption channels for dishwashers and washing machines

### LINKED ISSUE #11297: [miele] Channels for readings of power and water consumption
Some Miele appliances supports readings of power and water consumption, for example certain models of washing machines and dishwashers. The information is embedded into a binary string exposed as **ExtendedDeviceState** in the Rest API as well as the JSON-RPC API implemented in the openHAB binding.

I reversed engineered some years ago for an app I was working on. Originally for the WMV 960 WPS (in my case connected through ZigBee), then later also for G 6895 SCVi XXL K2O (Wi-Fi). Results for the dishwasher was posted here: https://techblog.vindvejr.dk/?p=198

I would like to provide these values for the openHAB binding also, as they could be quite useful for logging and statistical purposes.

jlaur:
Fixes #11297


kaikreuzer (approved):
lgtm, thanks!

PR #11300, opened 2021-09-25
[ipcamera] Fix ipcamera.jpg can return an old picture

Skinah:
Fixes an issue now that the binding is not polling the snapshot anymore where the jpg could return an old image from cache.

Precompiled jar can be downloaded here:
http://pcmus.com/openhab/IpCameraBinding/


wborn:
> I thought that the string building only occurred if the logger level was true, that was why we used the {} placeholders so the string was only built if it was actually needed?

The only magic going on here is that slf4j conditionally calls `.toString()` on the arguments and substitutes them in the log message depending on the configured log level. Any code used for creating the arguments passed to the logger is still executed.

When slf4j 2.0 is available, you can use also use lamdba expressions which will then only be executed depending on the log level, see: https://jira.qos.ch/browse/SLF4J-371

J-N-K:
What you can do is using `Logger.isTraceEnabled()` if you want to keep the logging and only build the log message if necessary.

PR #11302, opened 2021-09-25
[miele] Fix coffee machine auto-discovery.

### LINKED ISSUE #8779: [miele] Coffee Machine not showing up in autodiscovery
## Expected Behavior
I'm using a Miele XGW3000 gateway and have five devices connected over zigbee. 4 of 5 of this devices get discovered by the autodiscover feature of the binding. However the coffeemachine isn't showing up. Since the binding has a Thing type "coffemachine" the coffeemachine should be also discoverable.

## Current Behavior
Currently I can add the machine manually when I add it in Paper UI -> Inbox -> + -> Miele Binding -> Add manually.
For this I have to input the UID of the device. I fount the UID and have confirmed that it was the right UID. But after adding the Thing all the string values stay empty. However I can use the switch to turn the device ON and OFF.

coffee

Could it be that there is another type of coffemachine that isn't supported yet? What I can see from the homebus XML is that my coffemachine should have multiple switchable items. My coffemachine has a lightswitch that the current coffemachine Thing doesn't support.

homebus

homebus2

I'm using openhab 2.5.9 and tried it with the miele binding 2.5.9 and 2.5.10.Snapshot.

Since the coffemachine is the most important kitchen appliance for me, supporting my coffemachine would be much appreciated.

jlaur:
Fixes #8779

### DIFF
--- a/bundles/org.openhab.binding.miele/src/main/java/org/openhab/binding/miele/internal/MieleBindingConstants.java
+++ b/bundles/org.openhab.binding.miele/src/main/java/org/openhab/binding/miele/internal/MieleBindingConstants.java
@@ -44,6 +44,9 @@ public class MieleBindingConstants {
     public static final ThingTypeUID THING_TYPE_WASHINGMACHINE = new ThingTypeUID(BINDING_ID, "washingmachine");
     public static final ThingTypeUID THING_TYPE_COFFEEMACHINE = new ThingTypeUID(BINDING_ID, "coffeemachine");
 
+    // Miele devices classes
+    public static final String MIELE_DEVICE_CLASS_COFFEE_SYSTEM = "CoffeeSystem";
+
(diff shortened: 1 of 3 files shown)


jlaur:
The fix has been verified by the user who created the issue, and his coffee machine is now confirmed to work for:
- Thing created from auto-discovery.
- Manually configured thing from UI.
- Thing configured from file.

Please see thread in linked issue for details.

lolodomo (approved):
LGTM

jlaur:
@lolodomo, @kaikreuzer - should this be added to milestone 3.2?

kaikreuzer:
Yes, thanks for noticing.

@lolodomo As you're new, just fyi: Please always make sure to have set "bug" or "enhancement" label and the correct milestone whenever merging a PR. Thanks!

PR #11303, opened 2021-09-25
Allow multi release JARs to resolve

Skinah:
This allows bnd to resolve multi release jar files.

A number of people have posted they have troubles adding dependancies on the forum with this error.
`Classes found in the wrong directory: {META-INF/versions/9/module-info.class=module-info}`

Issue about it is here:
https://github.com/bndtools/bnd/issues/2227

An alternative fix is to add the following into each bindings pom.xml

```
  <properties>
    <bnd.fixupmessages>"Classes found in the wrong directory"; is:=warning</bnd.fixupmessages>
  </properties>
```

Not sure what the correct way to handle this is but these are two ways I have tested. The suggested way will mean any future people wont hit the error and can spend more time coding.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -81,7 +81,7 @@
 
     <bnd.importpackage/>
     <bnd.exportpackage/>
-    <bnd.fixupmessages/>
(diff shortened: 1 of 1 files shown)


kaikreuzer (approved):
lgtm, thank you!

Issue #11283, opened 2021-09-22
[mielecloud] AbstractMieleThingHandlerTest unstable


wborn:
Often the `AbstractMieleThingHandlerTest` fails.

See for instance this build: https://ci.openhab.org/job/openHAB-Addons/478

```
TEST org.openhab.binding.mielecloud.internal.handler.WineStorageDeviceThingHandlerTest#testHandleCommandProgramStopToStartStopChannel() <<< ERROR: 
Wanted but not invoked:
mieleWebservice.putProcessAction(
    "000124430017",
    STOP
);
-> at org.openhab.binding.mielecloud.internal.handler.AbstractMieleThingHandlerTest.lambda$4(AbstractMieleThingHandlerTest.java:437)

However, there was exactly 1 interaction with this mock:
mieleWebservice.dispatchDeviceState(
    "000124430017"
);
-> at org.openhab.binding.mielecloud.internal.handler.AbstractMieleThingHandler.initialize(AbstractMieleThingHandler.java:100)


Wanted but not invoked:
mieleWebservice.putProcessAction(
    "000124430017",
    STOP
);
-> at org.openhab.binding.mielecloud.internal.handler.AbstractMieleThingHandlerTest.lambda$4(AbstractMieleThingHandlerTest.java:437)

However, there was exactly 1 interaction with this mock:
mieleWebservice.dispatchDeviceState(
    "000124430017"
);
-> at org.openhab.binding.mielecloud.internal.handler.AbstractMieleThingHandler.initialize(AbstractMieleThingHandler.java:100)


	at org.openhab.binding.mielecloud.internal.handler.AbstractMieleThingHandlerTest.lambda$4(AbstractMieleThingHandlerTest.java:437)
	at org.openhab.core.test.java.JavaTest.waitForAssert(JavaTest.java:152)
	at org.openhab.core.test.java.JavaTest.waitForAssert(JavaTest.java:120)
	at org.openhab.core.test.java.JavaTest.waitForAssert(JavaTest.java:73)
	at org.openhab.binding.mielecloud.internal.handler.AbstractMieleThingHandlerTest.testHandleCommandProgramStopToStartStopChannel(AbstractMieleThingHandlerTest.java:436)
```

Can you have a look at this @BjoernLange?

BjoernLange:
Of course.

BjoernLange:
[...]

PR #11306, opened 2021-09-26
[mielecloud] Increase timeout for integration tests

BjoernLange:
Increases the integration test timeout to avoid failures in the CI build, see #11283.


wborn (approved):
Thanks for having a look! Let's test to see if this helps.

PR #11307, opened 2021-09-26
[mail] Add support for e-mail headers

andrewfg:
**Background**
Previously e-mails sent by this binding could not be grouped into a "threaded view" in the user's email client, because the binding could not provide an e-mail "Reference" header (which is the key for message threading).

**Solution**
This PR allows one to (optionally) add e-mail headers such as, but not limited to, the "Reference" header.

**Example Rule**
```
rule "Send Mail with a Reference Header"
when
    ...
then
    val mailer = getActions("mail","mail:smtp:xyz")
    mailer.addHeader("Reference", "<unique-thread-identifier>")
    mailer.sendMail("joe@abc.com", "message-subject", "message-text")
end
```

### DIFF
--- a/bundles/org.openhab.binding.mail/README.md
+++ b/bundles/org.openhab.binding.mail/README.md
@@ -127,3 +127,23 @@ val List<String> attachmentUrlList = newArrayList(
 val mailActions = getActions("mail","mail:smtp:sampleserver")
 mailActions.sendHtmlMail("mail@example.com", "Test subject", "<h1>Header</h1>This is the mail content.", attachmentUrlList)
 ```
+
+## Mail Headers
+
+The binding allows one to add custom e-mail headers to messages that it sends.
+For example if you want e-mails sent by this binding to be grouped into a "threaded view" in your email client, you must provide an e-mail "Reference" header, which acts as the key for grouping messages together.
+Headers can be added inside a rule by calling the `mailActions.addHeader()` method before calling the respective `mailActions.sendMail()` method.
+See the example below.
+
+```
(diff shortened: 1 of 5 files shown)


martinvw (approved):

J-N-K:
@openhab/add-ons-maintainers I would like to revert this change and replace it with the thread-safe implementation in https://github.com/smarthomej/addons/pull/199/files. Functionally this is equivalent (besides being thread-safe), but it is a breaking change regarding the usage. WDYT?

jlaur:
@J-N-K - I agree. We cannot preserve the current API/usage since it stores state across calls as you mentioned [here](https://github.com/openhab/openhab-addons/pull/11307#issuecomment-939428676). It shouldn't be a big deal since there is a new solution - of course needs to be mentioned in distro notes.

@openhab/add-ons-maintainers - other opinions?

andrewfg:
@J-N-K would your proposed change still support my original use-case (below)? (If so, then I would have no objections..)

```
rule "Send Mail with a Reference Header"
when
    ...
then
    val mailer = getActions("mail","mail:smtp:xyz")
    mailer.addHeader("Reference", "<unique-thread-identifier>")
    mailer.sendMail("joe@abc.com", "message-subject", "message-text")
end
```

J-N-K:
You would have to create a `Map<String,String>` and then use `mail.sendMail("joe@abc.com", "message-subject", "message-text", headers)` but I think that also works for you.

andrewfg:
^
LGTM :)

PR #11309, opened 2021-09-26
[lifx] Add channels for controlling absolute color temperature in Kelvin

wborn:
Adds an 'abstemperature' channel to all Thing Types for controlling the absolute color temperature in Kelvin.
MultiZone lights also have 'abstemperaturezone' channels that allow for controlling the color temperature of a zone in Kelvin.
These channels make it easier to use the same color temperature with lights that have a different color temperature range.
Furthermore the channel type tags have been updated which simplifies using the generated items with the semantic model.

### DIFF
--- a/bundles/org.openhab.binding.lifx/src/main/resources/OH-INF/thing/channel.xml
+++ b/bundles/org.openhab.binding.lifx/src/main/resources/OH-INF/thing/channel.xml
@@ -4,13 +4,38 @@
 	xmlns:thing="https://openhab.org/schemas/thing-description/v1.0.0"
 	xsi:schemaLocation="https://openhab.org/schemas/thing-description/v1.0.0 https://openhab.org/schemas/thing-description-1.0.0.xsd">
 
+	<channel-type id="abstemperature" advanced="true">
+		<item-type>Number</item-type>
+		<label>Temperature</label>
+		<description>Sets the color temperature of the light in Kelvin</description>
+		<category>ColorLight</category>
+		<tags>
+			<tag>Control</tag>
+			<tag>ColorTemperature</tag>
+		</tags>
+		<state min="1500" max="9000" step="1" pattern="%.0f K"/>
+	</channel-type>
+
+	<channel-type id="abstemperaturezone" advanced="true">
+		<item-type>Number</item-type>
+		<label>Temperature Zone</label>
+		<description>Sets the zone color temperature of the light in Kelvin</description>
+		<category>ColorLight</category>
+		<tags>
+			<tag>Control</tag>
+			<tag>ColorTemperature</tag>
+		</tags>
+		<state min="1500" max="9000" step="1" pattern="%.0f K"/>
+	</channel-type>
+
 	<channel-type id="brightness">
 		<item-type>Dimmer</item-type>
 		<label>Brightness</label>
 		<description>Sets the brightness of the light</description>
 		<category>DimmableLight</category>
 		<tags>
-			<tag>Lighting</tag>
+			<tag>Control</tag>
+			<tag>Light</tag>
 		</tags>
 		<config-description-ref uri="channel-type:lifx:brightness"/>
 	</channel-type>
@@ -21,7 +46,8 @@
 		<description>Selects the color of the light</description>
 		<category>ColorLight</category>
 		<tags>
-			<tag>Lighting</tag>
+			<tag>Control</tag>
+			<tag>Light</tag>
 		</tags>
 		<config-description-ref uri="channel-type:lifx:color"/>
 	</channel-type>
@@ -32,7 +58,8 @@
 		<description>Selects the zone color of the light</description>
 		<category>ColorLight</category>
 		<tags>
-			<tag>Lighting</tag>
+			<tag>Control</tag>
+			<tag>Light</tag>
 		</tags>
 	</channel-type>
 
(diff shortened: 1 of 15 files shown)


bundles/org.openhab.binding.lifx/src/main/resources/OH-INF/thing/channel.xml
```diff
@@ -4,13 +4,38 @@
 	xmlns:thing="https://openhab.org/schemas/thing-description/v1.0.0"
 	xsi:schemaLocation="https://openhab.org/schemas/thing-description/v1.0.0 https://openhab.org/schemas/thing-description-1.0.0.xsd">
 
+	<channel-type id="abstemperature" advanced="true">
```
kaikreuzer, line 7: I assume you are aware that we have a [system channel type](https://github.com/openhab/openhab-core/blob/e5e62d15a388ea188eccb4250257ec4a44479d39/bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/DefaultSystemChannelTypeProvider.java#L228-L237) in place as well. Did you not refer to it because of the range 1500-9000K for LIFX? This then probably renders our system channel rather useless as we'll likely have a similar situation for most bindings...

bundles/org.openhab.binding.lifx/src/main/resources/OH-INF/thing/channel.xml
```diff
@@ -4,13 +4,38 @@
...
+		<description>Sets the color temperature of the light in Kelvin</description>
+		<category>ColorLight</category>
+		<tags>
+			<tag>Control</tag>
+			<tag>ColorTemperature</tag>
+		</tags>
+		<state min="1500" max="9000" step="1" pattern="%.0f K"/>
+	</channel-type>
+
+	<channel-type id="abstemperaturezone" advanced="true">
```
kaikreuzer, line 19: Now this is really identical to the one above and should thus be removed.
If all you need to do is to define a different label/desc, this can easily be overwritten for the channel (which you already do in the code).

kaikreuzer (changes requested):

bundles/org.openhab.binding.lifx/src/main/resources/OH-INF/thing/channel.xml
wborn: Yes I still wanted to look into if it is possible to use more system channel types with the binding. Would it also be possible to just override certain defaults in the Thing Type defs? Like adding a channel config and maybe limit the range that way. Though the issue with that is you can probably not reuse such customizations. It would be nice to be able to extend your own channel type from a system channel type. 🙂

Originally I also wanted to add some channel config like default power on brightness on these new channels. But then I realized the item type does not allow for ON/OFF commands. Which explains why this is an advanced channel and the other channels are preferred because they allow for more commands.

wborn: I gave overriding system default channels a test but that does not work:

override

So whenever you want to use a channel configuration with system channel type you end up duplicating it. :cry: 

That's why I initially added the channel types because I wanted the to add a channel config. But now that I no longer use those, I can change them to system channels. I'll also copy/paste the label/description etc. from the system color/brightness channels to the lifx channels types.

bundles/org.openhab.binding.lifx/src/main/resources/OH-INF/thing/channel.xml
wborn: In my latest commit I've reused the system default channel types where possible @kaikreuzer.

bundles/org.openhab.binding.lifx/src/main/resources/OH-INF/thing/channel.xml
kaikreuzer: > Would it also be possible to just override certain defaults in the Thing Type defs? 

No, then it would not be the same type anymore and hence the re-use would not be given.

> But then I realized the item type does not allow for ON/OFF commands. Which explains why this is an advanced channel and the other channels are preferred because they allow for more commands.

Well, an item that is a number with unit Kelvin wouldn't know what to do with ON/OFF commands, right? But anyhow also for the "normal" color temperature channel, ON only means 100%, i.e. "full heat" in the temperature. It is not meant to change the brightness of a light, that's what the brightness/color channels are for.

>  I can change them to system channels.

How do you deal with the non-matching ranges? As LIFX only supports 1500-9000K, what will happen if you send 1000-1499K or 9001-10000K?

bundles/org.openhab.binding.lifx/src/main/resources/OH-INF/thing/channel.xml
kaikreuzer: 👍

bundles/org.openhab.binding.lifx/src/main/resources/OH-INF/thing/channel.xml
wborn: Currently it will still use the command but it applies the [light temperature range as upper/lower bound](https://github.com/openhab/openhab-addons/pull/11309/files#diff-3a50ab8c9d889eda73de943840c9d29a8ba262823100e9f64af8b92725f41f65R111-R112). That is probably more user friendly as refusing the command and logging a warning, whenever a user sends a command to a group of lights, where the command is outside the range of some of the lights. :wink:

kaikreuzer: Ok, makes sense. And it is definitely much much better than throwing some random exception. 😉

kaikreuzer (approved):
Thanks!

PR #11313, opened 2021-09-26
Fix/remove outdated "Default Tags"

wborn:
There used to be the following "Default Tags" which are no longer used in the current semantic model:

* Lighting
* Switchable
* CurrentTemperature
* TargetTemperature
* CurrentHumidity

See:

* https://www.openhab.org/docs/developer/bindings/thing-xml.html#default-tags
* https://www.openhab.org/docs/tutorial/model.html#semantic-model

---

I also fixed this for the LIFX Binding in https://github.com/openhab/openhab-addons/pull/11309.

### DIFF
--- a/bundles/org.openhab.binding.danfossairunit/src/main/resources/OH-INF/thing/thing-types.xml
+++ b/bundles/org.openhab.binding.danfossairunit/src/main/resources/OH-INF/thing/thing-types.xml
@@ -201,7 +201,8 @@
 		<label>Humidity</label>
 		<category>Humidity</category>
 		<tags>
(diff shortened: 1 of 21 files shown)


Skinah (approved):
Lgtm

J-N-K:
I'm not sure if HomeKit relies on CurrentTemperature and TargetTemperature.

wborn:
I think HomeKit uses its own metadata nowadays, see:

https://www.openhab.org/addons/integrations/homekit/#ui-based-configuration

t-8ch (approved):

kaikreuzer (approved):

PR #11316, opened 2021-09-27
[influxdb] Add info that it's no good idea to override the item tag via metadata

### LINKED ISSUE #11288: [influxdb] no datapoints found if metadata contains `item`

maglnet:
closes #11288

:-)

### DIFF
--- a/bundles/org.openhab.persistence.influxdb/README.md
+++ b/bundles/org.openhab.persistence.influxdb/README.md
@@ -93,7 +93,7 @@ temperature,item=tempBath,floor=firstfloor
 
 You can now easily select all temperatures of the firstfloor or the average temperature of the groundfloor.
 
-*Warning: Do **not** override the tag `item` within the metadata. This tag is used internally by OpenHAB and changing it will lead to problems querying the persisted datapoints.*
+*Warning: Do **not** override the tag `item` within the metadata. This tag is used internally by openHAB and changing it will lead to problems querying the persisted datapoints.*
 
 #### Extended automatic tagging
 
--- a/bundles/org.openhab.persistence.influxdb/README.md
+++ b/bundles/org.openhab.persistence.influxdb/README.md
@@ -93,6 +93,8 @@ temperature,item=tempBath,floor=firstfloor
 
 You can now easily select all temperatures of the firstfloor or the average temperature of the groundfloor.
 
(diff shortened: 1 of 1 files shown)


bundles/org.openhab.persistence.influxdb/README.md
```diff
@@ -93,6 +93,8 @@ temperature,item=tempBath,floor=firstfloor
 
 You can now easily select all temperatures of the firstfloor or the average temperature of the groundfloor.
 
+*Warning: Do **not** override the tag `item` within the metadata. This tag is used internally by OpenHAB and changing it will lead to problems querying the persisted datapoints.*
```
wborn, line 96: I have a very small suggestion :slightly_smiling_face: :

```suggestion
*Warning: Do **not** override the tag `item` within the metadata. This tag is used internally by openHAB and changing it will lead to problems querying the persisted datapoints.*
```

wborn (approved):
Thanks!

Issue #10635, opened 2021-05-02
Many bindings not available in Crowdin


lolodomo:
Translations for the Netatmo binding is not available in Crowdin.

@cweitkamp : any idea why ? Could it be because there is no netatmo.propertioes file in the binding ?

PS: I volunteer to approve French translations in the openhab-addons repo

lolodomo:
In fact, it looks like a lot of bindings are missing in Crowdin.
In French translations, there are 42 bindings.
Countining the projects having i18n with properties: 105 I believe
As another example, my sonyprojector binding is missing too.

cweitkamp:
@lolodomo Yes, exactly. To be recognized and uploaded to Crowdin the bundle needs one or more `*.properties` files in the `OH-INF/i18n/` folder. Once it has been placed there it will be synchronized.

Currently I am working on a XSLT transformation to automatically generate properties files from our XML definitions. I will commit a draft later.

Thanks for volunteering. I granted Proofread permissions to you on Crowdin for French.

lolodomo:
> Currently I am working on a XSLT transformation to automatically generate properties files from our XML definitions. I will commit a draft later.

Just for my understanding: why now do we have certain bindings but not others ? They were added manually in Crowdin ?

cweitkamp:
Crowdin integration looks for an existing `*.properties` file to synchronize it. Property files having a location suffix are excluded. That why only part of them are already available for translation.

cweitkamp:
See https://github.com/openhab/openhab-addons/pull/10636 for a draft of the XSLT transformation.

clinique:
I second @lolodomo on this. My dev time is a bit restricted currently, so I can help working on translations, and I was also looking for Netatmo binding in Crowdin :-).

cweitkamp:
@clinique Thanks you for offering to help translating add-ons. If you like to add translations for Netatmo someone has to create the properties fuel containing all English source strings and commit it to this repository. After that they will be uploaded automatically and available on Crowdin.

cweitkamp:
Closed by https://github.com/openhab/openhab-addons/pull/11760 (and others).

PR #11319, opened 2021-09-28
[nikohomecontrol] Prepare for translation

mherwege:
- Externalize strings to properties file to allow translation as a next step.
- Give listener thread a name.

### DIFF
--- a/bundles/org.openhab.binding.nikohomecontrol/src/main/java/org/openhab/binding/nikohomecontrol/internal/NikoHomeControlBindingConstants.java
+++ b/bundles/org.openhab.binding.nikohomecontrol/src/main/java/org/openhab/binding/nikohomecontrol/internal/NikoHomeControlBindingConstants.java
@@ -31,6 +31,9 @@ public class NikoHomeControlBindingConstants {
 
     public static final String BINDING_ID = "nikohomecontrol";
 
+    // Listener threadname prefix
+    public static final String THREAD_NAME_PREFIX = "OH-" + BINDING_ID + "-";
+
     // List of all Thing Type UIDs
 
     // bridge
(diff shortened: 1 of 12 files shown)


lolodomo:
Please look at other bindings how they defined translations for thing type label/description or config parameter label/description. There is no need to add `@text` in you thing files. You have predefined properties automatically defined for that.

lolodomo:
And please note that you did not use strictly the recommended thread name convention. The threads name should start by "OH-binding-" followied by the thing UID and optionally followed by an additional specific part in case you create several threads for the same thing.

mherwege:
> Please look at other bindings how they defined translations for thing type label/description or config parameter label/description. There is no need to add `@text` in you thing files. You have predefined properties automatically defined for that.
> Take a look at the NTP binding as an example.

@lolodomo Thank you for looking into this. I am well aware I do not strictly need the @text in my thing files. But if I understood well from [#10635](https://github.com/openhab/openhab-addons/issues/10635), I would then need to replicate all strings anyway for them to be picked up by Crowdin. I consider that prone to error when one makes a change and forgets to keep them in sync. Therefore I opted for the @text construct.

mherwege:
> And please note that you did not use strictly the recommended thread name convention. The threads name should start by "OH-binding-" followied by the thing UID and optionally followed by an additional specific part in case you create several threads for the same thing.

@lolodomo, thank you for noticing, I didn’t look far enough in my research and will correct.

lolodomo:
https://www.openhab.org/docs/concepts/things.html#thing-status
`
The statuses UNINITIALIZED, INITIALIZING and REMOVING are set by the framework, whereas the statuses UNKNOWN, ONLINE and OFFLINE are assigned from a binding.`

lolodomo:
> According to our documentation this is a valid way to define translatable phrases. There are other bindings which are already designed in this way. There are no strict guidelines for the future way of localization and thus it is up to the developers to choose their favorite path. From my POV @mherwege arguments for his decision are strong and I personally agree on his opinion.

I don't really like this choice being dictated by a tool (Crowdin) and the fact that we have 2 different ways of doing it, but if this is compliant with our coding rules, I will review (and approve) this part too. Thank you for the answer.

mherwege:
@lolodomo Thank you for your review. I believe I have fixed all thing lifecycle issues. Can you review again?

PR #11323, opened 2021-09-29
[automower] Support for GPS coordinates provided by API

bern77:
This PR adds support for the GPS coordinates provided by the Husqvarna Automower Connect API.
When requested the latest 'Mower Data' via the API the last 50 GPS positions are also provided.
They can be useful to visualize the mower's path or identify his last position in case of an error.

Considering the large number of additional channels, this PR introduces also some channel groups to structure the channels and separate the GPS channels from the rest.

### DIFF
--- a/bundles/org.openhab.binding.automower/README.md
+++ b/bundles/org.openhab.binding.automower/README.md
@@ -40,87 +40,91 @@ With the default value of 10min this would mean ~4300 requests per month per sin
 
 ## Channels
 
-### Channel Group 'Mower Status'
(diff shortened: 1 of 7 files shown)


bern77 (commented):
I've applied the requested changes, however I've now removed all channel groups just as the binding was implemented before in order to not break any user configurations.
Apparently it's not possible to mix channel with and without channel groups.

fwolter (approved):
LGTM Thanks for also fixing the readme!

PR #11324, opened 2021-09-29
[heliosventilation] reference modbus.helioseasycontrols binding

bern77:
There are two binding supporting different types of Helios ventilation systems. Cross referencing the bindings in their descriptions should help users identify the right binding for their device.


lolodomo (approved):
LGTM

kaikreuzer (approved):
Thanks!

PR #11325, opened 2021-09-29
[helioseasycontrols] reference heliosventilation binding

bern77:
There are two binding supporting different types of Helios ventilation systems. Cross referencing the bindings in their descriptions should help users identify the right binding for their device.


lolodomo (approved):
LGTM

lolodomo:
@kaikreuzer : do we need to set a milestone for documentation PR?

kaikreuzer (changes requested):
I prefer urls not to be shown as text unless they carry any relevant information.

kaikreuzer:
> @kaikreuzer : do we need to set a milestone for documentation PR?

Yes, this should be set for any PR.

kaikreuzer (approved):
Thank you!

PR #11327, opened 2021-09-30
[Astro] Removing org.apache.commons.*

clinique:
Some code cleansing

### DIFF
--- a/bundles/org.openhab.binding.astro/src/main/java/org/openhab/binding/astro/internal/action/AstroActions.java
+++ b/bundles/org.openhab.binding.astro/src/main/java/org/openhab/binding/astro/internal/action/AstroActions.java
@@ -14,8 +14,6 @@ package org.openhab.binding.astro.internal.action;
 
 import java.time.ZonedDateTime;
 
-import javax.measure.quantity.Angle;
-
 import org.eclipse.jdt.annotation.NonNullByDefault;
 import org.eclipse.jdt.annotation.Nullable;
 import org.openhab.binding.astro.internal.AstroBindingConstants;
@@ -42,7 +40,7 @@ import org.slf4j.LoggerFactory;
 public class AstroActions implements ThingActions {
 
     private final Logger logger = LoggerFactory.getLogger(AstroActions.class);
-    protected @Nullable AstroThingHandler handler;
+    private @Nullable AstroThingHandler handler;
 
     public AstroActions() {
         logger.debug("Astro actions service instanciated");
@@ -61,7 +59,7 @@ public class AstroActions implements ThingActions {
     }
 
     @RuleAction(label = "get the azimuth", description = "Get the azimuth for a given time.")
-    public @Nullable @ActionOutput(name = "getAzimuth", label = "Azimuth", type = "org.openhab.core.library.types.QuantityType<javax.measure.quantity.Angle>") QuantityType<Angle> getAzimuth(
+    public @Nullable @ActionOutput(name = "getAzimuth", label = "Azimuth", type = "org.openhab.core.library.types.QuantityType<javax.measure.quantity.Angle>") QuantityType<?> getAzimuth(
             @ActionInput(name = "date", label = "Date", required = false, description = "Considered date") @Nullable ZonedDateTime date) {
         logger.debug("Astro action 'getAzimuth' called");
         AstroThingHandler theHandler = this.handler;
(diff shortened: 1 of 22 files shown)


bundles/org.openhab.binding.astro/src/main/java/org/openhab/binding/astro/internal/model/Radiation.java
```diff
@@ -61,21 +60,21 @@ public void setTotal(double total) {
     /**
      * Returns the total radiation.
      */
-    public QuantityType<Intensity> getTotal() {
+    public QuantityType<?> getTotal() {
```
lolodomo, line 63: Same question for Inetensity

bundles/org.openhab.binding.astro/src/main/java/org/openhab/binding/astro/internal/model/Range.java
```diff
@@ -58,7 +56,7 @@ public Calendar getEnd() {
     /**
      * Returns the duration in minutes.
      */
-    public QuantityType<Time> getDuration() {
+    public QuantityType<?> getDuration() {
```
lolodomo, line 59: ... or Time

bundles/org.openhab.binding.astro/src/main/java/org/openhab/binding/astro/internal/util/DateTimeUtils.java
```diff
@@ -79,7 +98,11 @@ public static Calendar toCalendar(double julianDate) {
         long millis = (long) ((julianDate + 0.5 - J1970) * MILLISECONDS_PER_DAY);
         Calendar cal = Calendar.getInstance();
         cal.setTimeInMillis(millis);
-        return DateUtils.round(cal, Calendar.MINUTE);
+        int second = cal.get(Calendar.SECOND);
+        if (second > 30) {
+            cal.set(Calendar.MINUTE, cal.get(Calendar.MINUTE) + 1);
```
lolodomo, line 103: Probably wrong when current minutes are 59.

lolodomo (changes requested):

bundles/org.openhab.binding.astro/src/main/java/org/openhab/binding/astro/internal/action/AstroActions.java
```diff
@@ -61,7 +59,7 @@ public void setThingHandler(@Nullable ThingHandler handler) {
     }
 
     @RuleAction(label = "get the azimuth", description = "Get the azimuth for a given time.")
-    public @Nullable @ActionOutput(name = "getAzimuth", label = "Azimuth", type = "org.openhab.core.library.types.QuantityType<javax.measure.quantity.Angle>") QuantityType<Angle> getAzimuth(
+    public @Nullable @ActionOutput(name = "getAzimuth", label = "Azimuth", type = "org.openhab.core.library.types.QuantityType<javax.measure.quantity.Angle>") QuantityType<?> getAzimuth(
```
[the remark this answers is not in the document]
clinique: You're correct, I made this modifications because the IDE complained with : 
`Unsafe null type conversion (type annotations): The value of type '@NonNull QuantityType<@NonNull Angle>' is made accessible using the less-annotated type 'QuantityType<Angle>'`
...but in reality the underlaying cause comes from not @NonNullByDefault annotated classes. I will revert these, as I have a plan to increase non null annotations in coming PR (as well as removing Calendar usage later).

bundles/org.openhab.binding.astro/src/main/java/org/openhab/binding/astro/internal/util/DateTimeUtils.java
clinique: Yes, perhaps, will change.

lolodomo (approved):
LGTM

lolodomo:
@gerrieg : do you want to have a look before I merge ?

clinique:
@lolodomo : I've started to work on extending non null annotations, would it be better to wait a second PR ?

lolodomo:
No please, this is now fully reviewed, let me first merge.
I prefer small PRs.

lolodomo:
@clinique : this is now merged, you can start with a new PR.

bundles/org.openhab.binding.astro/src/main/java/org/openhab/binding/astro/internal/util/DateTimeUtils.java
Nadahar: @lolodomo  What I don't understand is why it should be rounded to the closest minute at all. 

The method is defined as:

```java
    /**
     * Returns a calendar object from a julian date.
     */
    public static Calendar toCalendar(double julianDate) {
```
I certainly wouldn't expect that to do any rounding at all.

Nadahar: The rounding seems to come from here: 

https://github.com/openhab/openhab1-addons/pull/1323/files#diff-30f366fcdb6799621f6cb88cf45c6125c694ad0ec35422ca1265e9ca9f1b67bc

...but before that, it was simply truncated, which certainly wasn't any better.

Nadahar: The truncation seems to have been there from the very beginning:

https://github.com/openhab/openhab1-addons/commit/9f449754574b7412ea028b0d2170bd59cf086b7e#diff-30f366fcdb6799621f6cb88cf45c6125c694ad0ec35422ca1265e9ca9f1b67bc

I'm just wondering why - isn't it better to have a more precise conversion?

PR #11321, opened 2021-09-28
[miele] Fix supercool/superfreeze for fridges/fridge-freezers

jlaur:
Fix channels **supercool** and **superfreeze** for fridges/fridge-freezers.

Both channels were declared as read-only strings, but should have been writable switches, and are in fact handled as switches in handleCommand:

```
case SUPERCOOL: {
    if (command.equals(OnOffType.ON)) {
        result = bridgeHandler.invokeOperation(applianceIdentifier, modelID, "startSuperCooling");
    } else if (command.equals(OnOffType.OFF)) {
        result = bridgeHandler.invokeOperation(applianceIdentifier, modelID, "stopSuperCooling");
    }
```

Furthermore, the channels were not updated according to current operation of the fridge/fridge-freezer. This fix enables two-way synchronization of the supercool/superfreeze state by using the **state** property.

See more details in linked issue.

## Testing
Tested with fridge.

Initial sitemap:
image

Enable:
image

Verify state in Miele app:
image

Disable:
image

Verify state in sitemap:
image

## Pull request note
This fix was provided as part of some planned channel quality improvements. See #11317 for intentions/possible next steps.

Fixes #11320 

Signed-off-by: Jacob Laursen <jacob-github@vindvejr.dk>

Issue #11317, opened 2021-09-27
[miele] Channel quality improvements


jlaur:
The channels provided by this Miele@home binding could use some love:
- ~~Some are of wrong type (e.g. fridge SuperCool is provided as read-only string as opposed to writable switch).~~ Extracted to #11320
- [Units of Measurement](https://www.openhab.org/docs/concepts/units-of-measurement.html)
- Categories where applicable.
- Semantic tags where applicable.
- Spinning speed for washing machines is reported as 10x actual value.
- Some channels are marked as advanced for no apparent reason.
- ~~Strings for phases, programs etc. where not provided by the XGW 3000. Example, "Eco" instead of "program.28" for dishwasher.~~ Extracted to #11335 
- ~~Channels for raw values (better suited for rules).~~ Extracted to #11335

PR #11329, opened 2021-09-30
[miele] Temperature channel improvements: UoM, categories and descriptions

jlaur:
This PR improves channels provided by the binding in different ways.

### Temperatures
- Changed type from **Number** to **Number:Temperature**.
- Now supports Unit of Measurement.
- Fixed descriptions for specific appliances. For example, previously target temperature channels were described as "Target temperature to be reached by the oven", but in fact applied also to fridge and washing machine.
- Special value 32768 °C is now correctly translated into **Undefined**.
- Added category **Temperature** for default icons.

### Spinning speed
For washing machines spinning speed was multiplied by 10. Probably mislead by description in response JSON:

```
                {
                    "Name": "spinningSpeed",
                    "Value": 1600,
                    "Polling": 0,
                    "Metadata": {
                        "Filter": "true",
                        "description": "Spinning speed divided by 10!, 0x00 = without spinning, 0xFF = rinse hold.",
                        "LocalizedID": "Spin Speed",
                        "LocalizedValue": "1600",
                        "access": "RE"
                    }
                },
```

16.000 is clearly not correct number of rotations per minute.

### Time
Added category **Time** for the following channels:
- start
- duration
- elapsed
- finish

**Elapsed** and **remaining** time channels are no longer marked as advanced.

### Next steps
See #11317. Should be linked by first commit, but isn't.

Added myself as binding maintainer, see #11321.

### DIFF
--- a/bundles/org.openhab.binding.miele/src/main/java/org/openhab/binding/miele/internal/ExtendedDeviceStateUtil.java
+++ b/bundles/org.openhab.binding.miele/src/main/java/org/openhab/binding/miele/internal/ExtendedDeviceStateUtil.java
@@ -57,7 +57,7 @@ public class ExtendedDeviceStateUtil {
      * Convert string to Number:Temperature state with unit Celcius
      */
     public static State getTemperatureState(String s) throws NumberFormatException {
-        if (s.equals(TEMPERATURE_UNDEFINED)) {
+        if (TEMPERATURE_UNDEFINED.equals(s)) {
             return UnDefType.UNDEF;
         }
         int temperature = Integer.parseInt(s);
--- a/bundles/org.openhab.binding.miele/src/main/java/org/openhab/binding/miele/internal/ExtendedDeviceStateUtil.java
+++ b/bundles/org.openhab.binding.miele/src/main/java/org/openhab/binding/miele/internal/ExtendedDeviceStateUtil.java
@@ -14,6 +14,11 @@ package org.openhab.binding.miele.internal;
 
 import java.nio.charset.StandardCharsets;
 
+import org.openhab.core.library.types.QuantityType;
+import org.openhab.core.library.unit.SIUnits;
+import org.openhab.core.types.State;
+import org.openhab.core.types.UnDefType;
+
 /**
  * The {@link ExtendedDeviceStateUtil} class contains utility methods for parsing
  * ExtendedDeviceState information
@@ -22,6 +27,7 @@ import java.nio.charset.StandardCharsets;
  */
 public class ExtendedDeviceStateUtil {
     private static final byte[] HEX_ARRAY = "0123456789ABCDEF".getBytes(StandardCharsets.US_ASCII);
+    private static final String TEMPERATURE_UNDEFINED = "32768";
 
     /**
      * Convert byte array to hex representation.
@@ -45,5 +51,16 @@ public class ExtendedDeviceStateUtil {
      */
     public static byte[] stringToBytes(String input) {
         return input.getBytes(StandardCharsets.ISO_8859_1);
+    }
+
+    /**
+     * Convert string to Number:Temperature state with unit Celcius
+     */
+    public static State getTemperatureState(String s) throws NumberFormatException {
+        if (s.equals(TEMPERATURE_UNDEFINED)) {
+            return UnDefType.UNDEF;
+        }
+        int temperature = Integer.parseInt(s);
+        return new QuantityType<>(temperature, SIUnits.CELSIUS);
     }
 }
--- a/bundles/org.openhab.binding.miele/src/main/java/org/openhab/binding/miele/internal/handler/FridgeChannelSelector.java
+++ b/bundles/org.openhab.binding.miele/src/main/java/org/openhab/binding/miele/internal/handler/FridgeChannelSelector.java
@@ -17,10 +17,11 @@ import static org.openhab.binding.miele.internal.MieleBindingConstants.SUPERCOOL
 import java.lang.reflect.Method;
 import java.util.Map.Entry;
 
+import org.openhab.binding.miele.internal.ExtendedDeviceStateUtil;
 import org.openhab.binding.miele.internal.handler.MieleBridgeHandler.DeviceMetaData;
-import org.openhab.core.library.types.DecimalType;
 import org.openhab.core.library.types.OnOffType;
 import org.openhab.core.library.types.OpenClosedType;
+import org.openhab.core.library.types.QuantityType;
 import org.openhab.core.library.types.StringType;
 import org.openhab.core.types.State;
 import org.openhab.core.types.Type;
@@ -43,16 +44,16 @@ public enum FridgeChannelSelector implements ApplianceChannelSelector {
     COMPANY_ID("companyId", "companyId", StringType.class, true),
     STATE("state", "state", StringType.class, false),
     SUPERCOOL(null, SUPERCOOL_CHANNEL_ID, OnOffType.class, false),
-    FRIDGECURRENTTEMP("currentTemperature", "current", DecimalType.class, false) {
+    FRIDGECURRENTTEMP("currentTemperature", "current", QuantityType.class, false) {
         @Override
         public State getState(String s, DeviceMetaData dmd) {
-            return getState(s);
+            return getTemperatureState(s);
         }
     },
-    FRIDGETARGETTEMP("targetTemperature", "target", DecimalType.class, false) {
+    FRIDGETARGETTEMP("targetTemperature", "target", QuantityType.class, false) {
         @Override
         public State getState(String s, DeviceMetaData dmd) {
-            return getState(s);
+            return getTemperatureState(s);
         }
     },
     DOOR("signalDoor", "door", OpenClosedType.class, false) {
@@ -140,6 +141,15 @@ public enum FridgeChannelSelector implements ApplianceChannelSelector {
         }
 
         return null;
+    }
+
+    public State getTemperatureState(String s) {
+        try {
+            return ExtendedDeviceStateUtil.getTemperatureState(s);
+        } catch (Exception e) {
+            logger.error("An exception occurred while converting '{}' into a State", s);
+            return null;
+        }
     }
 
     public String getMieleEnum(String s, DeviceMetaData dmd) {
--- a/bundles/org.openhab.binding.miele/src/main/java/org/openhab/binding/miele/internal/handler/FridgeChannelSelector.java
+++ b/bundles/org.openhab.binding.miele/src/main/java/org/openhab/binding/miele/internal/handler/FridgeChannelSelector.java
@@ -146,9 +146,9 @@ public enum FridgeChannelSelector implements ApplianceChannelSelector {
     public State getTemperatureState(String s) {
         try {
             return ExtendedDeviceStateUtil.getTemperatureState(s);
-        } catch (Exception e) {
-            logger.error("An exception occurred while converting '{}' into a State", s);
-            return null;
+        } catch (NumberFormatException e) {
+            logger.warn("An exception occurred while converting '{}' into a State", s);
+            return UnDefType.UNDEF;
         }
     }
 
--- a/bundles/org.openhab.binding.miele/src/main/java/org/openhab/binding/miele/internal/handler/FridgeFreezerChannelSelector.java
+++ b/bundles/org.openhab.binding.miele/src/main/java/org/openhab/binding/miele/internal/handler/FridgeFreezerChannelSelector.java
@@ -18,10 +18,11 @@ import static org.openhab.binding.miele.internal.MieleBindingConstants.SUPERFREE
 import java.lang.reflect.Method;
 import java.util.Map.Entry;
 
+import org.openhab.binding.miele.internal.ExtendedDeviceStateUtil;
 import org.openhab.binding.miele.internal.handler.MieleBridgeHandler.DeviceMetaData;
-import org.openhab.core.library.types.DecimalType;
 import org.openhab.core.library.types.OnOffType;
 import org.openhab.core.library.types.OpenClosedType;
+import org.openhab.core.library.types.QuantityType;
 import org.openhab.core.library.types.StringType;
 import org.openhab.core.types.State;
 import org.openhab.core.types.Type;
@@ -48,28 +49,28 @@ public enum FridgeFreezerChannelSelector implements ApplianceChannelSelector {
     FRIDGESTATE("fridgeState", "fridgestate", StringType.class, false),
     SUPERCOOL(null, SUPERCOOL_CHANNEL_ID, OnOffType.class, false),
     SUPERFREEZE(null, SUPERFREEZE_CHANNEL_ID, OnOffType.class, false),
-    FREEZERCURRENTTEMP("freezerCurrentTemperature", "freezercurrent", DecimalType.class, false) {
+    FREEZERCURRENTTEMP("freezerCurrentTemperature", "freezercurrent", QuantityType.class, false) {
         @Override
         public State getState(String s, DeviceMetaData dmd) {
-            return getState(s);
+            return getTemperatureState(s);
         }
     },
-    FREEZERTARGETTEMP("freezerTargetTemperature", "freezertarget", DecimalType.class, false) {
+    FREEZERTARGETTEMP("freezerTargetTemperature", "freezertarget", QuantityType.class, false) {
         @Override
         public State getState(String s, DeviceMetaData dmd) {
-            return getState(s);
+            return getTemperatureState(s);
         }
     },
-    FRIDGECURRENTTEMP("fridgeCurrentTemperature", "fridgecurrent", DecimalType.class, false) {
+    FRIDGECURRENTTEMP("fridgeCurrentTemperature", "fridgecurrent", QuantityType.class, false) {
         @Override
         public State getState(String s, DeviceMetaData dmd) {
-            return getState(s);
+            return getTemperatureState(s);
         }
     },
-    FRIDGETARGETTEMP("fridgeTargetTemperature", "fridgetarget", DecimalType.class, false) {
+    FRIDGETARGETTEMP("fridgeTargetTemperature", "fridgetarget", QuantityType.class, false) {
         @Override
         public State getState(String s, DeviceMetaData dmd) {
-            return getState(s);
+            return getTemperatureState(s);
         }
     },
     DOOR("signalDoor", "door", OpenClosedType.class, false) {
@@ -158,6 +159,15 @@ public enum FridgeFreezerChannelSelector implements ApplianceChannelSelector {
         }
 
         return null;
+    }
+
+    public State getTemperatureState(String s) {
+        try {
+            return ExtendedDeviceStateUtil.getTemperatureState(s);
+        } catch (Exception e) {
+            logger.error("An exception occurred while converting '{}' into a State", s);
+            return null;
+        }
     }
 
     public String getMieleEnum(String s, DeviceMetaData dmd) {
--- a/bundles/org.openhab.binding.miele/src/main/java/org/openhab/binding/miele/internal/handler/FridgeFreezerChannelSelector.java
+++ b/bundles/org.openhab.binding.miele/src/main/java/org/openhab/binding/miele/internal/handler/FridgeFreezerChannelSelector.java
@@ -164,9 +164,9 @@ public enum FridgeFreezerChannelSelector implements ApplianceChannelSelector {
     public State getTemperatureState(String s) {
         try {
             return ExtendedDeviceStateUtil.getTemperatureState(s);
-        } catch (Exception e) {
-            logger.error("An exception occurred while converting '{}' into a State", s);
-            return null;
+        } catch (NumberFormatException e) {
+            logger.warn("An exception occurred while converting '{}' into a State", s);
+            return UnDefType.UNDEF;
         }
     }
 
--- a/bundles/org.openhab.binding.miele/src/main/java/org/openhab/binding/miele/internal/handler/OvenChannelSelector.java
+++ b/bundles/org.openhab.binding.miele/src/main/java/org/openhab/binding/miele/internal/handler/OvenChannelSelector.java
@@ -18,11 +18,12 @@ import java.util.Date;
 import java.util.Map.Entry;
 import java.util.TimeZone;
 
+import org.openhab.binding.miele.internal.ExtendedDeviceStateUtil;
 import org.openhab.binding.miele.internal.handler.MieleBridgeHandler.DeviceMetaData;
 import org.openhab.core.library.types.DateTimeType;
-import org.openhab.core.library.types.DecimalType;
 import org.openhab.core.library.types.OnOffType;
 import org.openhab.core.library.types.OpenClosedType;
+import org.openhab.core.library.types.QuantityType;
 import org.openhab.core.library.types.StringType;
 import org.openhab.core.types.State;
 import org.openhab.core.types.Type;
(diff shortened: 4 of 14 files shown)


bundles/org.openhab.binding.miele/src/main/java/org/openhab/binding/miele/internal/ExtendedDeviceStateUtil.java
```diff
@@ -46,4 +52,15 @@ public static String bytesToHex(byte[] bytes) {
     public static byte[] stringToBytes(String input) {
         return input.getBytes(StandardCharsets.ISO_8859_1);
     }
+
+    /**
+     * Convert string to Number:Temperature state with unit Celcius
+     */
+    public static State getTemperatureState(String s) throws NumberFormatException {
+        if (s.equals(TEMPERATURE_UNDEFINED)) {
```
lolodomo, line 60: Can you please reverse ? You should certainly have a SAT warning on this line.

bundles/org.openhab.binding.miele/src/main/java/org/openhab/binding/miele/internal/handler/FridgeChannelSelector.java
```diff
@@ -142,6 +143,15 @@ public State getState(String s) {
         return null;
     }
 
+    public State getTemperatureState(String s) {
+        try {
+            return ExtendedDeviceStateUtil.getTemperatureState(s);
+        } catch (Exception e) {
```
lolodomo, line 149: Please avoid catching a so generic exception.
Isn't `NumberFormatException ` enough ?

bundles/org.openhab.binding.miele/src/main/java/org/openhab/binding/miele/internal/handler/FridgeChannelSelector.java
```diff
@@ -142,6 +143,15 @@ public State getState(String s) {
...
+            logger.error("An exception occurred while converting '{}' into a State", s);
```
lolodomo, line 150: WARN level is certainly more adapted.

bundles/org.openhab.binding.miele/src/main/java/org/openhab/binding/miele/internal/handler/FridgeFreezerChannelSelector.java
```diff
@@ -160,6 +161,15 @@ public State getState(String s) {
         return null;
     }
 
+    public State getTemperatureState(String s) {
+        try {
+            return ExtendedDeviceStateUtil.getTemperatureState(s);
+        } catch (Exception e) {
```
lolodomo, line 167: Please avoid catching a so generic exception.
Isn't `NumberFormatException ` enough ?

bundles/org.openhab.binding.miele/src/main/java/org/openhab/binding/miele/internal/handler/FridgeFreezerChannelSelector.java
```diff
@@ -160,6 +161,15 @@ public State getState(String s) {
...
+            logger.error("An exception occurred while converting '{}' into a State", s);
```
lolodomo, line 168: WARN level is certainly more adapted.

bundles/org.openhab.binding.miele/src/main/java/org/openhab/binding/miele/internal/handler/FridgeChannelSelector.java
```diff
@@ -142,6 +143,15 @@ public State getState(String s) {
...
+            return null;
```
lolodomo, line 151: Rather than null, wouldn't it be better to return UndefType.UNDEF ?
In case there is a valid reason to return null, please add `@Nullable` before State in the method declaration.

bundles/org.openhab.binding.miele/src/main/java/org/openhab/binding/miele/internal/handler/FridgeFreezerChannelSelector.java
```diff
@@ -160,6 +161,15 @@ public State getState(String s) {
...
+            return null;
```
lolodomo, line 169: Rather than null, wouldn't it be better to return UndefType.UNDEF ?
In case there is a valid reason to return null, please add `@Nullable` before State in the method declaration.

bundles/org.openhab.binding.miele/src/main/java/org/openhab/binding/miele/internal/handler/OvenChannelSelector.java
```diff
@@ -216,6 +217,15 @@ public State getState(String s) {
         return null;
     }
 
+    public State getTemperatureState(String s) {
+        try {
+            return ExtendedDeviceStateUtil.getTemperatureState(s);
+        } catch (Exception e) {
```
lolodomo, line 223: Please avoid catching a so generic exception.
Isn't `NumberFormatException ` enough ?

bundles/org.openhab.binding.miele/src/main/java/org/openhab/binding/miele/internal/handler/OvenChannelSelector.java
```diff
@@ -216,6 +217,15 @@ public State getState(String s) {
...
+            logger.error("An exception occurred while converting '{}' into a State", s);
```
lolodomo, line 224: WARN level is certainly more adapted.

bundles/org.openhab.binding.miele/src/main/java/org/openhab/binding/miele/internal/handler/OvenChannelSelector.java
```diff
@@ -216,6 +217,15 @@ public State getState(String s) {
...
+            return null;
```
lolodomo, line 225: Rather than null, wouldn't it be better to return UndefType.UNDEF ?
In case there is a valid reason to return null, please add `@Nullable` before State in the method declaration.

bundles/org.openhab.binding.miele/src/main/java/org/openhab/binding/miele/internal/handler/WashingMachineChannelSelector.java
```diff
@@ -223,6 +223,15 @@ public State getState(String s) {
         return null;
     }
 
+    public State getTemperatureState(String s) {
+        try {
+            return ExtendedDeviceStateUtil.getTemperatureState(s);
+        } catch (Exception e) {
+            logger.error("An exception occurred while converting '{}' into a State", s);
+            return null;
```
lolodomo, line 231: Rather than null, wouldn't it be better to return UndefType.UNDEF ?
In case there is a valid reason to return null, please add `@Nullable` before State in the method declaration.

bundles/org.openhab.binding.miele/src/main/java/org/openhab/binding/miele/internal/handler/WashingMachineChannelSelector.java
```diff
@@ -223,6 +223,15 @@ public State getState(String s) {
         return null;
     }
 
+    public State getTemperatureState(String s) {
+        try {
+            return ExtendedDeviceStateUtil.getTemperatureState(s);
+        } catch (Exception e) {
+            logger.error("An exception occurred while converting '{}' into a State", s);
```
lolodomo, line 230: WARN level is certainly more adapted.

bundles/org.openhab.binding.miele/src/main/java/org/openhab/binding/miele/internal/handler/WashingMachineChannelSelector.java
```diff
@@ -223,6 +223,15 @@ public State getState(String s) {
         return null;
     }
 
+    public State getTemperatureState(String s) {
+        try {
+            return ExtendedDeviceStateUtil.getTemperatureState(s);
+        } catch (Exception e) {
```
lolodomo, line 229: Please avoid catching a so generic exception.
Isn't `NumberFormatException ` enough ?

bundles/org.openhab.binding.miele/src/main/resources/OH-INF/thing/channeltypes.xml
```diff
@@ -86,98 +90,50 @@
...
-	<channel-type id="measured" advanced="true">
-		<item-type>Number</item-type>
-		<label>Measured Temperature</label>
-		<description>Actual measured temperature in the oven</description>
-		<state readOnly="true"></state>
-	</channel-type>
-
-	<channel-type id="temp" advanced="true">
-		<item-type>Number</item-type>
+	<channel-type id="temperature">
```
lolodomo, line 93: Defining so generic channel type is not recommended.
At least, you could have `temperature` and `TargetTemperature` which are two very different notions.

bundles/org.openhab.binding.miele/src/main/resources/OH-INF/thing/channeltypes.xml
```diff
@@ -212,14 +168,14 @@
 		<state readOnly="true"></state>
 	</channel-type>
 
-	<channel-type id="powerConsumption" advanced="false">
+	<channel-type id="powerConsumption">
```
lolodomo, line 171: You suppressed "advanced" to all channel ?
When you have many channels on a thing, this is reasonable to have some of them being advanced.
In general, you don't set a channel as advanced when you are sure that any user will use/link it.

bundles/org.openhab.binding.miele/src/main/resources/OH-INF/thing/fridge.xml
```diff
@@ -16,8 +16,14 @@
...
-			<channel id="current" typeId="current"/>
-			<channel id="target" typeId="target"/>
+			<channel id="current" typeId="temperature">
+				<label>Current Temperature</label>
+				<description>Current temperature in the fridge</description>
+			</channel>
+			<channel id="target" typeId="temperature">
+				<label>Target Temperature</label>
+				<description>Target temperature to be reached by the fridge</description>
+			</channel>
```
lolodomo, line 26: As already mentioned, I don't thing this is a good idea to have the same channel type. Having both will then lead to no need to redefine the label and description here.

bundles/org.openhab.binding.miele/src/main/resources/OH-INF/thing/fridgefreezer.xml
```diff
@@ -19,10 +19,22 @@
...
+				<description>Target temperature to be reached by the freezer compartment</description>
+			</channel>
+			<channel id="fridgecurrent" typeId="temperature">
+				<label>Current Temperature</label>
+				<description>Current temperature in the fridge compartment</description>
+			</channel>
+			<channel id="fridgetarget" typeId="temperature">
+				<label>Target Temperature</label>
+				<description>Target temperature to be reached by the fridge compartment</description>
+			</channel>
```
lolodomo, line 37: Same remark here.
But here, it makes sense to have only one common "targetTemperature channel type for 2 different channels."

bundles/org.openhab.binding.miele/src/main/resources/OH-INF/thing/oven.xml
```diff
@@ -22,10 +22,22 @@
...
+				<description>Actual measured temperature in the oven</description>
+			</channel>
+			<channel id="temp1" typeId="temperature">
+				<label>Temperature</label>
+				<description>Program temperature in the oven</description>
+			</channel>
+			<channel id="temp2" typeId="temperature">
+				<label>Temperature</label>
+				<description>Program temperature in the oven</description>
+			</channel>
```
lolodomo, line 40: Same remark

lolodomo (changes requested):

bundles/org.openhab.binding.miele/src/main/java/org/openhab/binding/miele/internal/handler/FridgeChannelSelector.java
jlaur: Yes, fixed.

bundles/org.openhab.binding.miele/src/main/java/org/openhab/binding/miele/internal/handler/FridgeFreezerChannelSelector.java
jlaur: Yes, fixed.

bundles/org.openhab.binding.miele/src/main/java/org/openhab/binding/miele/internal/handler/OvenChannelSelector.java
jlaur: Yes, fixed.

bundles/org.openhab.binding.miele/src/main/java/org/openhab/binding/miele/internal/handler/WashingMachineChannelSelector.java
jlaur: Yes, fixed.

bundles/org.openhab.binding.miele/src/main/java/org/openhab/binding/miele/internal/handler/OvenChannelSelector.java
jlaur: Agreed, fixed.

bundles/org.openhab.binding.miele/src/main/java/org/openhab/binding/miele/internal/handler/WashingMachineChannelSelector.java
jlaur: Agreed, fixed.

bundles/org.openhab.binding.miele/src/main/java/org/openhab/binding/miele/internal/handler/FridgeFreezerChannelSelector.java
jlaur: Agreed, fixed.

bundles/org.openhab.binding.miele/src/main/java/org/openhab/binding/miele/internal/handler/FridgeChannelSelector.java
jlaur: Agreed, fixed.

bundles/org.openhab.binding.miele/src/main/java/org/openhab/binding/miele/internal/handler/FridgeChannelSelector.java
jlaur: Good idea, fixed.

bundles/org.openhab.binding.miele/src/main/java/org/openhab/binding/miele/internal/handler/OvenChannelSelector.java
jlaur: Good idea, fixed.

bundles/org.openhab.binding.miele/src/main/java/org/openhab/binding/miele/internal/handler/WashingMachineChannelSelector.java
jlaur: Good idea, fixed.

bundles/org.openhab.binding.miele/src/main/java/org/openhab/binding/miele/internal/handler/FridgeFreezerChannelSelector.java
jlaur: Good idea, fixed.

bundles/org.openhab.binding.miele/src/main/resources/OH-INF/thing/channeltypes.xml
jlaur: No, only for remaining/elapsed time as I think these channels are quite common. almost the primary/most important ones. For the rest I just removed explicit attribute as **advanced=false** is default. This makes it more readable IMHO.

bundles/org.openhab.binding.miele/src/main/resources/OH-INF/thing/channeltypes.xml
jlaur: I considered it, and it would make sense if target temperature was actually settable. Unfortunately it's not, so decided to go with shared definition as all properties are identical, so defining two different types seems arbitrary and redundant. Arbitrary because previously we had a channel type for each individual channel. Having two is better than having nine, but still - why two?

bundles/org.openhab.binding.miele/src/main/resources/OH-INF/thing/fridge.xml
jlaur: The description is needed as it's tailored for each appliance type. But you are right about label, so will reconsider.

bundles/org.openhab.binding.miele/src/main/java/org/openhab/binding/miele/internal/ExtendedDeviceStateUtil.java
jlaur: Can you elaborate, I don't understand?

bundles/org.openhab.binding.miele/src/main/resources/OH-INF/thing/channeltypes.xml
lolodomo: @kaikreuzer : please comment, I remember you told me to not do that (defining ulmtra generic chan,nel types like "temperature") a very long time ago. Is it still the case ?

bundles/org.openhab.binding.miele/src/main/resources/OH-INF/thing/channeltypes.xml
lolodomo: Oh you're right, I did not take care that advanced attribute was set to false.
In this case, this is OK for me.

bundles/org.openhab.binding.miele/src/main/java/org/openhab/binding/miele/internal/ExtendedDeviceStateUtil.java
lolodomo: Constant at left: TEMPERATURE_UNDEFINED.equals(s)

Maybe the SAT check is more clever than I imagine and only alerts when the variable can be null ?

bundles/org.openhab.binding.miele/src/main/resources/OH-INF/thing/fridge.xml
jlaur: You convinced me to split into current and target for reuse of same label - fixed.

bundles/org.openhab.binding.miele/src/main/resources/OH-INF/thing/channeltypes.xml
jlaur: You convinced me to split into current and target for reuse of same label - fixed. Still interested in comments about finding right balance between "too generic" and "too redundant". :-)

bundles/org.openhab.binding.miele/src/main/resources/OH-INF/thing/oven.xml
jlaur: You convinced me to split into current and target for reuse of same label - fixed.

bundles/org.openhab.binding.miele/src/main/resources/OH-INF/thing/fridgefreezer.xml
jlaur: You convinced me to split into current and target for reuse of same label - fixed.

bundles/org.openhab.binding.miele/src/main/java/org/openhab/binding/miele/internal/ExtendedDeviceStateUtil.java
Skinah: go to the bindings \target\code-analysis folder after you compile and read the report for warnings.

Yes I believe it is more clever as I made a suggestion to someone else and got told it did not apply.

jlaur: Thanks, fixed.

Skinah:
When this is merged can you create an Alert for it in this file? See other bindings that have Alerts for changing to use UOM...

https://github.com/openhab/openhab-distro/blob/34c684842beec47b1e51febd5ffb670f537504e7/distributions/openhab/src/main/resources/bin/update.lst

bundles/org.openhab.binding.miele/src/main/resources/OH-INF/thing/oven.xml
```diff
@@ -22,10 +22,21 @@
...
```
lolodomo, line 39: Maybe use "Program Temperature" as label ?
And now way to give a hint of the difference between these 2 temperatures ?

lolodomo (commented):
A last comment/remark

lolodomo:
@Skinah : why did you add the WIP label ?
I think there is no WIP, @jlaur is just considering the review comments...

Skinah:
Sorry label removed, saw `draft` and assumed it meant the same thing.

bundles/org.openhab.binding.miele/src/main/resources/OH-INF/thing/oven.xml
jlaur: Existing labels, not changed. Unfortunately I don't know what they are as my oven doesn't support them. I have changed to "Program Temperature 1" and "Program Temperature 2".

lolodomo (approved):
LGTM. Thanks

lolodomo:
Will merge if the build is green.

jlaur:
@Skinah - the "(potentially) not backward compatible", is that because of the change from Number to Number:Temperature? It shouldn't break compatibility, but some custom formatting might change of course.

jlaur:
@lolodomo - can you link to #11317? I think there is some problem with auto-linking based on commit messages when PR is born as draft.

Skinah:
@jlaur Yes from what I understand (happy for someone to tell me i am wrong) when you make this change from Number to Number:Temperature you will need to do a delete of the thing and recreate items. This is why I tagged it and also suggested the Alert be added to the file above. For example...

ALERT;TP-Link Smart Home Binding: The energy and rssi channels are now using Units of Measurements. Items must be adapted and the things must be recreated.

Thanks for updating the binding and adding UOM.

jlaur:
> @jlaur Yes from what I understand (happy for someone to tell me i am wrong) when you make this change from Number to Number:Temperature you will need to do a delete of the thing and recreate items. 

For sure you don't need to delete and recreate the thing. I will perform a new test of item created from UI. For me everything worked right out of the box, and I could simply change the type of my item in .items file for proper formatting in sitemaps etc.

jlaur:
> For sure you don't need to delete and recreate the thing. I will perform a new test of item created from UI. For me everything worked right out of the box, and I could simply change the type of my item in .items file for proper formatting in sitemaps etc.

@Skinah - I went back to version from main branch and created a new item from UI:
image
(same formatting in sitemap with no custom formatting: `Text item=Fridge_TargetTemperature`)

Then installed version from this branch, and changed target temperature to different value to see if still updated correctly:
image

Notice that item type doesn't match anymore, but without causing any problems. Last step, changed type of item in UI. Result:
image

So you don't need to recreate either things or items, you can simply update items to have UoM formatting. So this should be fully backwards compatible.

PR #11333, opened 2021-10-01
Proteus Eco Meter Binding

2chilled:
[Proteus EcoMeter] Initial contribution

Support Proteus EcoMeter (S). Main motivation is I want to make correlations between my Smart Irrigation System and the actual amount of water used by each watering run.

Thread: https://community.openhab.org/t/proteus-ecometer-binding/125903

Note that initially I used the openHAB serial port abstraction for connecting to the device. However I received corrupt byte messages. Therefore I've chosen another lib, which seems to work with absolutely zero issues so far.


bundles/org.openhab.binding.proteusecometer/src/main/java/org/openhab/binding/proteusecometer/internal/ProteusEcoMeterBindingConstants.java
```diff
@@ -0,0 +1,41 @@
...
+
+    public static final String temperature = "temperature";
+
+    public static final String sensorLevel = "sensorLevel";
+
+    public static final String usableLevelInLitre = "usableLevelInLitre";
+
+    public static final String usableLevelInPercent = "usableLevelInPercent";
+
+    public static final String totalCapacity = "totalCapacity";
```
Skinah, line 40: See https://www.openhab.org/docs/developer/guidelines.html#java-coding-style

CONSTANTS_SHOULD_BE_ALL_CAPS

See the BINDING_ID as an example.

bundles/org.openhab.binding.proteusecometer/src/main/resources/OH-INF/thing/thing-types.xml
```diff
@@ -0,0 +1,62 @@
...
+			<parameter name="usbPort" type="text" required="true">
+				<context>serial-port</context>
+				<label>USB Port</label>
+				<description>USB port the device is connected to i.e. /dev/ttyUSB0</description>
+			</parameter>
+		</config-description>
+	</thing-type>
+
+	<channel-type id="Temperature">
+		<item-type>Number:CELSIUS</item-type>
```
Skinah, line 29:
```suggestion
		<item-type>Number:Temperature</item-type>
```
See
https://www.openhab.org/docs/concepts/units-of-measurement.html#defining-channeltypes

if you try and create an item in OH3's main UI, this needs to match what is offered as an option. Please do a self review of the others as more than this one needs changing.

Skinah (changes requested):

bundles/org.openhab.binding.proteusecometer/NOTICE
```diff
@@ -0,0 +1,13 @@
...
+
+== Declared Project Licenses
+
+This program and the accompanying materials are made available under the terms
+of the Eclipse Public License 2.0 which is available at
+https://www.eclipse.org/legal/epl-2.0/.
+
+== Source Code
+
+https://github.com/openhab/openhab-addons
```
Skinah, line 13: Have a look at other NOTICE files as you will need to add some lines for jSerialComm library.

example:

== Third-party Content

oshi-core
* License: MIT License
* Project: https://github.com/oshi/oshi
* Source:  https://github.com/oshi/oshi

bundles/create_openhab_binding_skeleton.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
```
Skinah, line 1: Is this change a mistake? are you able to remove it from showing up and being tracked?

bundles/org.openhab.binding.proteusecometer/README.md
```diff
@@ -0,0 +1,134 @@
+# ProteusEcoMeter Binding
+
+This is the binding for the Proteus EcoMeter S, which is able to report the level of a cistern or tank.
+
+Note that this binding currently supports no write channels. This means you have to configure your sensor by
+considering the manual of the product (using wireless display). After doing that the binding comes into play
+and helps you to get your measured values into openHAB. Please be patient while waiting for the first received
+data. The sensor reports at an interval of approx. 1h, except when the water level changes relatively fast.
```
Skinah, line 8:
```suggestion
Note that this binding currently supports no write channels.
This means you have to configure your sensor by considering the manual of the product (using wireless display).
After doing that the binding comes into play and helps you to get your measured values into openHAB.
Please be patient while waiting for the first received data.
The sensor reports at an interval of approx. 1h, except when the water level changes relatively fast.
```

Please do this for the rest of the readme. Each sentence on its own line so it is cleaner if a change is needed to be made and reviewed.

If you accept this change or any others, don't forget to add your signoff on the last line on the comment field.

bundles/org.openhab.binding.proteusecometer/src/main/java/org/openhab/binding/proteusecometer/internal/ProteusEcoMeterConfiguration.java
```diff
@@ -0,0 +1,22 @@
...
+ */
+package org.openhab.binding.proteusecometer.internal;
+
+/**
+ * The {@link ProteusEcoMeterConfiguration} class contains fields mapping thing configuration parameters.
+ *
+ * @author Matthias Herrmann - Initial contribution
+ */
+public class ProteusEcoMeterConfiguration {
+    public String usbPort;
```
Skinah, line 21:
```suggestion
@NonNullByDefault
public class ProteusEcoMeterConfiguration {
    public String usbPort="";
```
Removes the need to check for null and NPE can not occur.

bundles/org.openhab.binding.proteusecometer/src/main/java/org/openhab/binding/proteusecometer/internal/ecometers/handler/ProteusEcoMeterSHandler.java
```diff
@@ -0,0 +1,157 @@
...
+ * The {@link ProteusEcoMeterSHandler} updates thing channels when receiving data
+ *
+ * @author Matthias Herrmann - Initial contribution
+ */
+@NonNullByDefault
+public class ProteusEcoMeterSHandler extends BaseThingHandler {
+
+    private final Logger logger = LoggerFactory.getLogger(ProteusEcoMeterSHandler.class);
+    private @Nullable SerialPort serialPort;
+    private @Nullable ProteusEcoMeterConfiguration config;
```
Skinah, line 55:
```suggestion
    private ProteusEcoMeterConfiguration config = new ProteusEcoMeterConfiguration ();
```
Remove the need to check for null all the time.

bundles/org.openhab.binding.proteusecometer/src/main/java/org/openhab/binding/proteusecometer/internal/ecometers/handler/ProteusEcoMeterSHandler.java
```diff
@@ -0,0 +1,157 @@
...
+
+    @Override
+    public void initialize() {
+        config = getConfigAs(ProteusEcoMeterConfiguration.class);
+        final ProteusEcoMeterConfiguration localConfig = config;
+        if (localConfig == null) {
+            final String msg = "Expected initialize to be called ONLY when config is not null";
+            logger.error(msg);
+            throw new IllegalStateException(msg);
+        }
```
Skinah, line 70:
```suggestion
```

If you followed the other suggestions you can just use `config` directly.

bundles/org.openhab.binding.proteusecometer/src/main/java/org/openhab/binding/proteusecometer/internal/ecometers/handler/ProteusEcoMeterSHandler.java
```diff
@@ -0,0 +1,157 @@
...
+        job = scheduler.schedule(() -> {
+
+            try {
+                final ProteusEcoMeterSService ecoMeterSService = new ProteusEcoMeterSService();
+                final Stream<ProteusEcoMeterSReply> replyStream = ecoMeterSService.read(localConfig.usbPort,
+                        serialPortService);
+                updateStatus(ThingStatus.ONLINE);
+
+                replyStream.forEach(reply -> {
+                    updateState(ProteusEcoMeterBindingConstants.SENSOR_LEVEL,
```
Skinah, line 81:
```suggestion
                    updateState(SENSOR_LEVEL,
```
Use an import at the top and you can remove all of these. Please do to all of them.

Skinah (changes requested):
This should do for a first pass and looks pretty close. Whilst you wait for someone to do an in depth review, can you make sure you have fixed any warnings in the /target/code analysis/ folder after you compile and read the report file.

bundles/create_openhab_binding_skeleton.sh
2chilled: This abstracts over the concrete install path of the bash binary. It's needed for distros using an advanced package manager (nix) and it still works in traditional envs. You should really consider keeping it.

bundles/org.openhab.binding.proteusecometer/src/main/java/org/openhab/binding/proteusecometer/internal/ProteusEcoMeterConfiguration.java
2chilled: .. in exchange for a non type universal way of representing "this is not defined", hindering parametric polymorphism and therefore the most general (and reusable) class of algorithms we know of.

But you are right, in this system it's probably the better tradeoff and I will apply it. I just come from a universe where "is not defined" is represented explicitly and in a principled type universal fashion such that problems like this one just don't occur. This is not meant as an offend :)

Skinah: No offence taken. I don’t know much on this topic as I am self taught and only worked on this project take what I suggest as a suggestion and not a request as you probably know more then me.

bundles/create_openhab_binding_skeleton.sh
Skinah: Over my head. I feel it is best for a separate PR which would be great to have. A single line change like that will progress much quicker.

Skinah (approved):
LGTM

Just a note that there are two files changed not part of the bundle..
.gitignore
 bundles/create_openhab_binding_skeleton.sh
I can't make a call on those, someone else will. The rest looks good.

Thanks for the submission.

bundles/create_openhab_binding_skeleton.sh
2chilled: @Skinah Sure, I can make a separate PR for that

bundles/org.openhab.binding.proteusecometer/src/main/java/org/openhab/binding/proteusecometer/internal/ProteusEcoMeterConfiguration.java
2chilled: @Skinah I have lots of respect for self taught people. If you are interested in other ways to write typesafe programs, take a look at https://typelevel.org/ and especially http://typelevel.org/cats/

lolodomo:
@2chilled : if not yet done, you could add your binding's logo to the openHAB website. See https://www.openhab.org/docs/developer/bindings/#add-your-binding-s-logo-to-the-openhab-website

PR #11339, opened 2021-10-03
Abstract over bin paths using env

2chilled:
Abstract over bin paths using env

This makes the script usable without modifications on [NixOS](https://nixos.org/) without breaking it in other environments.


wborn (approved):
Thanks!

PR #11343, opened 2021-10-04
[myq] Fixes a serious issue that could wipe out all OH cookies 

digitaldan:
thanks to @lolodomo  for pointing this out, this fixes an issue where httpclient cookies for an entire OH instance could be wiped out .  This scopes the cookie deletion to just the MyQ auth URL . 
 

digitaldan:
@lolodomo @fwolter  this is kinda an emergency fix , i didn't realize i was deleting cookies across all bindings using the jetty http client !

lolodomo (approved):
Thank you for the quick fix

PR #11257, opened 2021-09-17
[feed] added channel for enclosure link on latest entry

juergenpabel:
[feed] added channel for enclosure link on latest entry

The feed plugin doesn't support (most) podcasts, as the audio link is provided in an enclosure tag; this PR adds the new channel "lastest-enclosure" to the thing. Albeit multiple enclosures are supported, this thing exposes only the first enclosure (if any at all - else the channel will be UNDEF). This covers the vast majority of podcasts sufficiently while minimizing the changes to the binding (a general re-refactoring/improvement of this binding would seem to be reasonable, though).

PR #11345, opened 2021-10-04
[feed] added channel for enclosure link on latest entry

juergenpabel:
[feed] added channel for enclosure link on latest entry

The feed plugin doesn't support (most) podcasts, as the audio link is provided in an enclosure tag; this PR adds the new channel "lastest-enclosure" to the thing. Albeit multiple enclosures are supported, this thing exposes only the first enclosure (if any at all - else the channel will be UNDEF). This covers the vast majority of podcasts sufficiently while minimizing the changes to the binding (a general re-refactoring/improvement of this binding would seem to be reasonable, though).

Recreated the PR for cleanliness sake (replaces https://github.com/openhab/openhab-addons/pull/11257/)

### DIFF
--- a/bundles/org.openhab.binding.feed/README.md
+++ b/bundles/org.openhab.binding.feed/README.md
@@ -32,17 +32,18 @@ Optional configuration:
 
 The binding supports following channels
 
-| Channel Type ID    | Item Type | Description                                         |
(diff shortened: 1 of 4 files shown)


juergenpabel:
@lolodomo & @svilenvul: replaced the original PR with this one (includes feed README doc update)

PR #11351, opened 2021-10-06
[systeminfo] Fix heap graph too high on axis.

Skinah:
Was calculating the “used heap” based on max heap size and not on the currently allocated size.


Skinah:
@wborn Thanks, it is now fixed.

Skinah:
Yes, not sure what happened but will be more careful of this.

PR #11352, opened 2021-10-06
[upb] Handle device state reports

marcusb:
Device state reports are sent out by devices when their state is changed manually,
such as when pressing a light switch. This handles the messages by updating
the thing state in openHAB accordingly.


marcusb:
@lolodomo Let's wait for feedback.

lolodomo:
@marcusb / what's the status ? Is it ready for merging ? If not, please add the tag "work in progress".
I don't know why the DCO is not green. If you could fix that, that would be perfect.

marcusb:
@lolodomo Still waiting for an update from the tester.

I do not have permissions to set labels, and I don't know what's up with the CI status.

marcusb:
@lolodomo User confirmed it works, please merge.

PR #11353, opened 2021-10-06
Ensure trailing zeroes are removed before updating channels

pacive:
Avoids issues with e.g map transformations.

Signed-off-by: Anders Alfredsson <andersb86@gmail.com>


pacive:
> Are you sure that it will be ok in case your value is 200.0 for example ?

Did a bit of testing, and it should work as intended, but decided to use `setScale()` instead, to get a more accurate value regardless.

PR #11360, opened 2021-10-09
[openweathermap] Bridge status set to ONLINE when no attached things

### LINKED ISSUE #11191: [openweathermap] OpenWeatherMap Account offline until Weather and Forecast thing created
## Expected Behavior
OpenWeatherMap Account thing should show as Online after creation even without dependent things.

## Current Behavior
After properly configuring an OpenWeatherMap Account thing, the resultant thing stays offline until you create a new dependent Weather and forecast (or possibly other OWM thing).

1. Have no OWM things.
2. Create new OpenWeatherMap Account thing
3. do not have any dependent things

## Context
While not a show stopper, it burned a significant amount of time debugging and attempting to get the "Account" thing to come online thinking it was configured improperly.

## Your Environment
* Version used: OH 3.1
* Environment name and version (e.g. Chrome 76, Java 8, Node.js 12.9, ...): Chromium Edge Version 92.0.902.84 (Official build) (64-bit)
* Java: OpenJDK 64-Bit Server VM Zulu11.50+19-CA (build 11.0.12+7-LTS, mixed mode)
* Operating System and version (desktop or mobile, Windows 10, Raspbian Buster, ...): Debian 11 VM w/o desktop environment

lolodomo:
Fix #11191

### DIFF
--- a/bundles/org.openhab.binding.openweathermap/src/main/java/org/openhab/binding/openweathermap/internal/handler/OpenWeatherMapAPIHandler.java
+++ b/bundles/org.openhab.binding.openweathermap/src/main/java/org/openhab/binding/openweathermap/internal/handler/OpenWeatherMapAPIHandler.java
@@ -14,6 +14,7 @@ package org.openhab.binding.openweathermap.internal.handler;
 
 import static org.openhab.binding.openweathermap.internal.OpenWeatherMapBindingConstants.*;
 
+import java.util.List;
 import java.util.Set;
 import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.TimeUnit;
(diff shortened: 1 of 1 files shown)


lolodomo:
Or we simply decide that the bridge thing will always be ONLINE, whatever the status of the child's? 
It should not be so difficult to break the current computation of the bridge status.

kaikreuzer:
I think it is ok to keep it at UNKNOWN when there are child things and we can thus assume it will go to ONLINE or OFFLINE very soon after.
In the case here, I would indeed prefer to set it to ONLINE directly as a thing that stays forever in UNKNOWN state looks broken to me (and users probably wonder if anything is wrong with it).

PR #11362, opened 2021-10-09
Align translation filenames with Crowdin

wborn:
Crowdin is configured to generate translations for languages without countries.

See also https://github.com/openhab/openhab-addons/pull/11361.


lolodomo:
Please note that the documentation has to be fixed. 
https://www.openhab.org/docs/developer/utils/i18n.html

PR #11363, opened 2021-10-09
[ntp] Add all default translations to properties file

wborn:
This allows for translating the NTP Binding strings with Crowdin.


wborn:
Yet another approach (probably easier) could be some kind of bundle that you use runtime, with a console command to generate the translations properties file.

lolodomo:
> Yet another approach (probably easier) could be some kind of bundle that you use runtime, with a console command to generate the translations properties file.

Yes a console command is a great idea. As a parameter you will provide the binding id. Something to implement in the core framework. But the code behind this console command is probably not so trivial...

lolodomo:
And by the way bindings need often code changes in particular for thing status details which are often hardcoded in English in the things handlers. If not, the binding is not fully translated.

PR #11371, opened 2021-10-11
[danfossairunit] Add filter period channel

### LINKED ISSUE #11310: [danfossairunit] Filter lifetime management
I recently changed my filters as automatic response to a notification about remaining filter life being 0%. This time I had some additional time to be curious about the algorithm for determining when filter replacement is due, which turned out to be based on the most simple algorithm of all:
image

i.e. time-based and linear without any sensors or other more sophisticated logic (like volume of air going through since last filter replacement).

This time I purchased new filters with no plastic frame and more filter area, which should last longer than my previous filters. So I wondered how to manage this, as my Danfoss Air A2 will still ask me to replace filter after nine months.

I did some digging in this document:
https://docs.google.com/spreadsheets/d/1FnWGzeicXrBhMg_jaHFM7jzVb_5aiflNTMsUz-j_rRo/edit#gid=0

and found three interesting registers:

Name              | Endpoint | Number | Value | Type       | Flags     | CCM ver.
------------------|----------|--------|-------|------------|-----------|---------
Filter Period     |        1 |   5225 |     9 | TIME_MONTH | Writeable | 2
Filter Period Min |        1 |   5227 |     3 | TIME_MONTH | ReadOnly  | 2
Filter Period Max |        1 |   5228 |     9 | TIME_MONTH | ReadOnly  | 2

So I decided to examine them, which I did by experimentally adding new channels for them, respectively **DanfossHRV_Service_FilterPeriod**, **DanfossHRV_Service_FilterPeriodMin** and **DanfossHRV_Service_FilterPeriodMax**.

Initially:
- [openhab.event.ItemStateChangedEvent ] - Item 'DanfossHRV_Service_FilterPeriod' changed from NULL to 9
- [openhab.event.ItemStateChangedEvent ] - Item 'DanfossHRV_Service_FilterPeriodMin' changed from NULL to 3
- [openhab.event.ItemStateChangedEvent ] - Item 'DanfossHRV_Service_FilterPeriodMax' changed from NULL to 9

As min and max are readonly I thought they probably defined the boundaries for the period to be set. But let's try to probe and see if this is actually the case:

Find actual minimum:
- [openhab.event.ItemCommandEvent ] - Item 'DanfossHRV_Service_FilterPeriod' received command 1
- [penhab.event.ItemStatePredictedEvent] - Item 'DanfossHRV_Service_FilterPeriod' predicted to become 1
- [openhab.event.ItemStateChangedEvent ] - Item 'DanfossHRV_Service_FilterPeriod' changed from 9 to 1
- [openhab.event.ItemStateChangedEvent ] - Item 'DanfossHRV_Service_FilterPeriod' changed from 1 to 9
- [openhab.event.ItemCommandEvent ] - Item 'DanfossHRV_Service_FilterPeriod' received command 2
- [penhab.event.ItemStatePredictedEvent] - Item 'DanfossHRV_Service_FilterPeriod' predicted to become 2
- [openhab.event.ItemStateChangedEvent ] - Item 'DanfossHRV_Service_FilterPeriod' changed from 9 to 2
- [openhab.event.ItemStateChangedEvent ] - Item 'DanfossHRV_Service_FilterPeriod' changed from 2 to 9
- [openhab.event.ItemCommandEvent ] - Item 'DanfossHRV_Service_FilterPeriod' received command 3
- [penhab.event.ItemStatePredictedEvent] - Item 'DanfossHRV_Service_FilterPeriod' predicted to become 3
- [openhab.event.ItemStateChangedEvent ] - Item 'DanfossHRV_Service_FilterPeriod' changed from 9 to 3

So it seems that minimum number of months allowed to be set is 3.

Find actual maximum:
- [openhab.event.ItemCommandEvent ] - Item 'DanfossHRV_Service_FilterPeriod' received command 12
- [penhab.event.ItemStatePredictedEvent] - Item 'DanfossHRV_Service_FilterPeriod' predicted to become 12
- [openhab.event.ItemStateChangedEvent ] - Item 'DanfossHRV_Service_FilterPeriod' changed from 3 to 12
- [openhab.event.ItemCommandEvent ] - Item 'DanfossHRV_Service_FilterPeriod' received command 13
- [penhab.event.ItemStatePredictedEvent] - Item 'DanfossHRV_Service_FilterPeriod' predicted to become 13
- [openhab.event.ItemStateChangedEvent ] - Item 'DanfossHRV_Service_FilterPeriod' changed from 12 to 13
- [openhab.event.ItemStateChangedEvent ] - Item 'DanfossHRV_Service_FilterPeriod' changed from 13 to 12

So in this case it seems that maximum number of months allowed to be set is actually 12, at least for my A2 unit. Which is fortunate, since my intention was to be able to extend the period from the standard 9 months.

When changing the FilterPeriod value I noticed that RemainingFilterLife is recalculated accordingly. For example, as starting point I was at 99.60784149169922% with period of 9 months as filter was replaced yesterday. After changing period to 3 months, it was recalculated to 98.82353210449219%. When setting period to 12 months it was recalculated to 100%. And when setting back to 9 months it was again calculated to 99.60784149169922%. So nothing it reset, and period can be changed freely.

### Proposed solution

I would definitely like to be able to set the filter period through the binding. Hardcoding the boundary as between 3 and 12 seems reasonable as it doesn't seem possible to get this information from the CCM module.

Question now is how to implement it. My first thought is that thing configuration might be an appropriate place for this, while a channel seems less desirable, since it requires more from the user and is not changed frequently.

On the other hand, if this value can be set by the Air Dial, this could lead to some undesired results where Air Dial setting would override the value set when initializing the binding, and vice versa next time binding is initialized for whatever reason. So if this is the case it might be more appropriate to provide a channel, so rules could monitor and update it.

On my Link CC Controller this value cannot be set, but in the service menu I can read the property in device information: "Filter lifetime: 274 days". So now wondering if this value is hardcoded or read from the unit at some point. This was after changing it to 12 months, so for sure it didn't have any effect yet.

### Other considerations
Values for channel service#filter_life should be rounded to one decimal. It's a percentage value with 255 possible values, so if we absolutely don't want to lose any updates (last updates before replacement needed): 1.1764706373214722% -> 0.7843137383460999% -> 0.3921568691730499% -> 0%

one decimal should be fine.

jlaur:
Fixes #11310

Add new channel **filter_life** for number of months between filter replacements.

### DIFF
--- a/bundles/org.openhab.binding.danfossairunit/src/main/java/org/openhab/binding/danfossairunit/internal/discovery/DanfossAirUnitDiscoveryService.java
+++ b/bundles/org.openhab.binding.danfossairunit/src/main/java/org/openhab/binding/danfossairunit/internal/discovery/DanfossAirUnitDiscoveryService.java
@@ -80,7 +80,6 @@ public class DanfossAirUnitDiscoveryService extends AbstractDiscoveryService {
         logger.debug("Try to discover all Danfoss Air CCM devices");
 
         try (DatagramSocket socket = new DatagramSocket()) {
-
             Enumeration<NetworkInterface> interfaces = NetworkInterface.getNetworkInterfaces();
             while (interfaces.hasMoreElements()) {
                 NetworkInterface networkInterface = interfaces.nextElement();
@@ -96,7 +95,6 @@ public class DanfossAirUnitDiscoveryService extends AbstractDiscoveryService {
                     sendBroadcastToDiscoverThing(socket, interfaceAddress.getBroadcast());
                 }
             }
-
         } catch (IOException e) {
             logger.debug("No Danfoss Air CCM device found. Diagnostic: {}", e.getMessage());
         }


Skinah:
All the new PR's are failing at the moment so it is probably a build system issue, when new PR's pass again it will be worth triggering a rebuild then otherwise it will just chew up the servers time if we expect it to fail not due to your code.

lolodomo:
I tried after I was informed the build problem could have been solved.

PR #11372, opened 2021-10-11
Extend .gitignore

2chilled:
Extend .gitignore to git ignore files which should be ignored.


fwolter:
@openhab/add-ons-maintainers do you have any objections merging it?

PR #11380, opened 2021-10-13
[sunspec modbus] fix thing example

mstormi:
The Thing definition in () in l193 references the *name* that was assigned in l192, correct?
So they should match (they didn't). And "bridge" isn't quite a clever name in examples so I changed that in the first place.
Or do I misunderstand the mapping here ?

Signed-off-by: Markus Storm <markus.storm@gmx.net>


mrbig:
Thank you @mstormi you're absolute correct.

mrbig:
Seems like I have approval right, @ssalonen could you please approve this change?

ssalonen:
Code looks good and ci failure is probably unrelated.

Commit sign off is missing, please try to fix that.

PR #11389, opened 2021-10-15
[WLED] Fix Nightlight sleep timer to use default time

svjense:
[WLED] Fix Nightlight sleep timer to use default time

Based on WLED API (https://kno.wled.ge/interfaces/http-api/) current handler uses NL=1 so sleep timer is hard coded to 1 minute. Should be default (ND) to use configured time in WLED.

### DIFF
--- a/bundles/org.openhab.binding.wled/src/main/java/org/openhab/binding/wled/internal/WLedHandler.java
+++ b/bundles/org.openhab.binding.wled/src/main/java/org/openhab/binding/wled/internal/WLedHandler.java
@@ -387,7 +387,7 @@ public class WLedHandler extends BaseThingHandler {
                 break;
             case CHANNEL_SLEEP:
                 if (OnOffType.ON.equals(command)) {
-                    sendGetRequest("/win&NL=1");
+                    sendGetRequest("/win&ND");
                 } else {
                     sendGetRequest("/win&NL=0");
                 }


lolodomo:
Code looks good. 
Let wait for binding author approval first

Skinah:
What firmware version has it been tested to work on? I suspect it has changed how it operates since I tested with an older firmware as if you look at the API and the replies back from a camera you will see that NL=1 means it is enabled and not that the 1 is a time in minutes.

happy for it to be merged but I suspect it will fix a newer firmware whilst breaking anyone on the older firmware so if you can reply with version I can look at abstracting the Api to its own class and then extend the class to be different per version. I think this is going to be needed for this fast moving project.

svjense:
Sorry, forgot to mention I tested on the latest stable 0.12.0. Looking into Aircookies code i don't think &ND will break older versions but the oldest version I can find on github is 0.10.0 here: [https://github.com/Aircoookie/WLED/blob/71886c162b7df2ce027b32c0334b83152e39884c/wled00/set.cpp#L571](url)
Maybe you can use `&ND&NL=1` combined for backward compatibility. That should take the default value for time and ignore the 1 minute in firmware >=0.10.0 (probably older versions as well).

svjense:
Thanks for the git crash course. Done!

PR #11406, opened 2021-10-18
[linky] Correcting authentication bug 

clinique:
issue #10360


lolodomo:
One point remaining before I approve: the if condition related to username. Sorry for confusion I probably introduced.

PR #11244, opened 2021-09-12
[miele] Fix multicast and multi-protocol support (ZigBee/Wi-Fi)

jlaur:
Fixes #11242

Fixes #11243

Fixes #11252

Signed-off-by: Jacob Laursen <jacob-github@vindvejr.dk>

This PR fixes issues related to protocol support and broken multicast support. This is a continuation of #11218 which was a quickfix, while this one is more in-depth and trying to fix root cause, but with full backwards configuration compatibility:
- Protocol specification is no longer parsed/interpreted, but simply concatenated with configured appliance ID's which unfortunately are without this. This makes it also forwards-compatible with any new protocol.
- Hardcoded ZigBee protocol for all handlers except oven has been fixed, so that actions for Wi-Fi devices will now work.

Also fixed multicast support which was broken by an XGW 3000 firmware upgrade in 2016. This fix is backwards compatible, so if anyone is using firmware prior to 2.03, it will also work.

Last, fixed auto-discovery finding already configured things through things files.

Manual testing performed for:
- ZigBee device configured through UI.
- ZigBee device configured through file.
- Wi-Fi device configured through UI.
- Wi-Fi device configured through file.

See also #8902

PR #11423, opened 2021-10-23
[miele] Clean up properties and improve reliability and performance

### LINKED ISSUE #11422: [miele] Improve performance and reliability
After getting acquainted with the Miele binding and fixing some ZigBee/Wi-Fi issues, I would like to take next steps and improve reliability and performance.

#11244 introduced a dependency to the **protocol** property for correctly handling both ZigBee and Wi-Fi appliances. This has two drawbacks:
- The **protocol** property is a technical protocol prefix ("hdm:ZigBee:" or "hdm:LAN:"). #11244 did not change this, but it would make more sense to expose a user-oriented protocol name like "ZigBee" or "Wi-Fi". This is not possible with current technical dependency.
- In rare scenarios properties are not correctly set. This breaks channel updates as correct API calls cannot be made without protocol prefix.

Furthermore, some data structures could be replaced for better performance.

Previous related work:
- #11218
- #11244

jlaur:
Fixes #11422

### Reliability

#11244 introduced a dependency to the **protocol** property for correctly handling both ZigBee and Wi-Fi appliances. This had two drawbacks:
- The **protocol** property is a technical protocol prefix ("hdm:ZigBee:" or "hdm:LAN:"). #11244 did not change this, but it would make more sense to expose a user-oriented protocol name like "ZigBee" or "Wi-Fi". This is not possible with current technical dependency.
- In rare scenarios properties are not correctly set. This breaks channel updates as correct API calls cannot be made without protocol prefix. I'm not sure exactly when this happens, but it could be timing related if properties are not initialized before receiving state updates.

Similarly a dependency to the **serialNumber** property was introduced for correct handling multicast updates.

To fix both problems, two internal appliance caches are now used for fast lookup of UID from appliance id (i.e. "0123456789#0" -> "hdm:ZigBee:0123456789#0") as well as from remote UID/serial number. After removing this property dependency, the **protocol** property has been renamed to **protocolAdapter** which now contains only the actual protocol adapter name, i.e. **ZigBee** or **LAN**. Additionally, a new property, **connectionType**, has been introduced. For Wi-Fi devices protocolAdapter is **LAN** and connectionType is **WiFi**.

### Performance

Appliance cache has been changed from List to HashMap for faster lookups, and is also being used more actively now for looking up UID's.

### Property cleanup

Properties removed:
- **brandId** always containing value MI.
- **companyId** always containing value MI.

Properties renamed:
- **dc** renamed to **deviceClass**. Now also set for things configured in files, where previously only auto-discovered things had this property set.
- **protocol** renamed to **protocolAdapter** as this is more correct.

Properties added:
- **model** containing the appliance model.
- **connectionType** (not available for ZigBee appliances).

Before:
image

After:
image

### Refactoring

Refer to "gateway" instead of "ZigBee network" in configuration, since both ZigBee and LAN/Wi-Fi appliances are supported by the gateway.

Added constants for remaining handlers with hardcoded device classes.

Changed obscure string operations for skipping this result of the _getDeviceClassObjects_ response:
`"DeviceClass": "com.prosyst.mbs.services.zigbee.hdm.deviceclasses.ReportingControl"`

while processing only this result:
`"DeviceClass": "com.miele.xgw3000.gateway.hdm.deviceclasses.MieleWashingMachine"`
(or other device class name for appliance)

Fixed all warnings:
- Potential null pointer access
- Redundant null check
- Import never used

Fixed some Static Code Analysis issues:
- The package org.apache.commons.lang3.StringUtils should not be used.
- 'static' modifier out of order with the JLS suggestions.
- First javadoc author should have "Initial contribution" contribution description.

### Translation

Added default i18n properties file and partial Danish translation as baseline. Next steps will be adding custom strings for states, programs and phases so these can be provided in desired language. This is not planned within the context of this PR.

### DIFF
--- a/bundles/org.openhab.binding.miele/src/main/java/org/openhab/binding/miele/internal/DeviceUtil.java
+++ b/bundles/org.openhab.binding.miele/src/main/java/org/openhab/binding/miele/internal/DeviceUtil.java
@@ -63,13 +63,4 @@ public class DeviceUtil {
         int temperature = Integer.parseInt(s);
         return new QuantityType<>(temperature, SIUnits.CELSIUS);
     }
-
-    /**
-     * Get device class from fully qualified string
-     */
-    public static String getDeviceClassFromFullyQualifiedString(String fullString) {
-        // com.miele.xgw3000.gateway.hdm.deviceclasses.MieleWashingMachine
-        // Position of "WashingMachine" in above string:
-        return fullString.substring(49);
-    }
 }
--- a/bundles/org.openhab.binding.miele/src/main/java/org/openhab/binding/miele/internal/DeviceUtil.java
+++ b/bundles/org.openhab.binding.miele/src/main/java/org/openhab/binding/miele/internal/DeviceUtil.java
@@ -14,6 +14,7 @@ package org.openhab.binding.miele.internal;
 
 import java.nio.charset.StandardCharsets;
 
+import org.apache.commons.lang3.StringUtils;
 import org.openhab.core.library.types.QuantityType;
 import org.openhab.core.library.unit.SIUnits;
 import org.openhab.core.types.State;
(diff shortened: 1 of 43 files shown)


jlaur:
@kgoderis, @kaikreuzer - would it be possible for you to review this PR? It would be nice if it could be included in next milestone as it fixes an issue with dependency to protocol and seriaNumber properties which really would be nice to solve sooner than later.

kaikreuzer (approved):
Many thanks @jlaur for this nice refactoring and for cleaning up so many places in the code along the way!
Looks all good to me, so let's merge!

PR #11429, opened 2021-10-23
[homematic] Improve (re)connect handling to Homematic gateways

MHerbst:
The PR improves the connect and especially the re-connect handling to Homematic gateways through the following measures:

1. by using a global unique id for each bridge installation it is possible to use two or more OH installation with the same Homematic gateway in parallel.
2. the HmIP daemon normally need much longer than the HM daemon until it accepts event registration. Therefore a configurable retry mechanismn has been implemented.

Signed-off-by: Martin Herbst <develop@mherbst.de>

### DIFF
--- a/bundles/org.openhab.binding.homematic/src/main/java/org/openhab/binding/homematic/internal/communicator/client/RpcClient.java
+++ b/bundles/org.openhab.binding.homematic/src/main/java/org/openhab/binding/homematic/internal/communicator/client/RpcClient.java
@@ -74,11 +74,6 @@ public abstract class RpcClient<T> {
     }
 
     /**
-     * Disposes the client.
-     */
-    public abstract void dispose();
-
-    /**
      * Returns a RpcRequest for this client.
      */
     protected abstract RpcRequest<T> createRpcRequest(String methodName);
@@ -128,7 +123,16 @@ public abstract class RpcClient<T> {
                 logger.error("Callback registration for interface {} timed out", hmInterface.getName());
                 throw new IOException("Unable to reconnect in time");
             }
+            future = null;
         }
+    }
+
+    /**
+     * Disposes the client.
+     */
+    public void dispose() {
+        if (future != null)
+            future.cancel(true);
     }
 
     /**
--- a/bundles/org.openhab.binding.homematic/src/main/java/org/openhab/binding/homematic/internal/communicator/client/RpcClient.java
+++ b/bundles/org.openhab.binding.homematic/src/main/java/org/openhab/binding/homematic/internal/communicator/client/RpcClient.java
@@ -19,6 +19,7 @@ import java.util.Collection;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.UUID;
 
 import org.openhab.binding.homematic.internal.HomematicBindingConstants;
 import org.openhab.binding.homematic.internal.common.HomematicConfig;
@@ -55,6 +56,7 @@ public abstract class RpcClient<T> {
     protected static final int RESP_BUFFER_SIZE = 8192;
 
     protected HomematicConfig config;
+    private String thisUID = UUID.randomUUID().toString();
 
     public RpcClient(HomematicConfig config) {
         this.config = config;
@@ -83,10 +85,10 @@ public abstract class RpcClient<T> {
     /**
      * Register a callback for the specified interface where the Homematic gateway can send its events.
      */
-    public void init(HmInterface hmInterface, String clientId) throws IOException {
+    public void init(HmInterface hmInterface) throws IOException {
         RpcRequest<T> request = createRpcRequest("init");
         request.addArg(getRpcCallbackUrl());
-        request.addArg(clientId);
+        request.addArg(thisUID);
         if (config.getGatewayInfo().isHomegear()) {
             request.addArg(Integer.valueOf(0x22));
         }
--- a/bundles/org.openhab.binding.homematic/src/main/java/org/openhab/binding/homematic/internal/communicator/client/RpcClient.java
+++ b/bundles/org.openhab.binding.homematic/src/main/java/org/openhab/binding/homematic/internal/communicator/client/RpcClient.java
@@ -92,7 +92,28 @@ public abstract class RpcClient<T> {
         if (config.getGatewayInfo().isHomegear()) {
             request.addArg(Integer.valueOf(0x22));
         }
-        sendMessage(config.getRpcPort(hmInterface), request);
+        int maxRetryAttempts = config.getRetryConnects();
+        int waitTime = config.getRetryWaitTime() * 1000;
+        for (int attempt = 1; attempt <= maxRetryAttempts; attempt++) {
+            try {
+                logger.debug("Register callback for interface {}, attempt {}/{}", hmInterface.getName(), attempt,
+                        maxRetryAttempts);
+                sendMessage(config.getRpcPort(hmInterface), request);
+                break;
+            } catch (IOException e) {
+                if (attempt == maxRetryAttempts) {
+                    logger.warn(
+                            "All attempts to register a callback for interface '{}' failed, last exception message: {}",
+                            hmInterface.getName(), e.getMessage());
+                    throw (e);
+                }
+                try {
+                    Thread.sleep(waitTime);
+                } catch (InterruptedException e1) {
+                    // ignore
+                }
+            }
+        }
     }
 
     /**
--- a/bundles/org.openhab.binding.homematic/src/main/java/org/openhab/binding/homematic/internal/communicator/client/RpcClient.java
+++ b/bundles/org.openhab.binding.homematic/src/main/java/org/openhab/binding/homematic/internal/communicator/client/RpcClient.java
@@ -20,6 +20,12 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.UUID;
+import java.util.concurrent.CancellationException;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.ScheduledFuture;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
 
 import org.openhab.binding.homematic.internal.HomematicBindingConstants;
 import org.openhab.binding.homematic.internal.common.HomematicConfig;
@@ -42,6 +48,7 @@ import org.openhab.binding.homematic.internal.model.HmGatewayInfo;
 import org.openhab.binding.homematic.internal.model.HmInterface;
 import org.openhab.binding.homematic.internal.model.HmParamsetType;
 import org.openhab.binding.homematic.internal.model.HmRssiInfo;
+import org.openhab.core.common.ThreadPoolManager;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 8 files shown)


Flole998:
I don't really like the idea of limiting the amount of reconnects. I want it to attempt to reconnect forever in case the connection is dropped. If my gateway crashes and I'm not at home and don't notice it right away then after some time a simple restart of the gateway won't help, did I understand that correctly? So in my opinion there should be a way to specify "forever" in some way, even if it's just as "65535 means retry forever".

MHerbst:
I think my description and probably the naming of the new options is a bit misleading. After the connection to the CCU was lost, the binding attempts the reconnect forever. This has not been changes. I have added a retry for the registration of callbacks. After the binding finds again a running CCU it immediately tries to register itself for events. For HM-RF components this worked without problems. But for HmIP components this did not work because the HmIP daemon on the CCU needs much longer to start (on a CCU2 about 2 minutes). Therefore I have added the retry for the registration of the callbacks. 
I will improve the naming of the options and the description to make this clearer. I don't think that it makes sense to retry the callback registration forever. If it does not work within some minutes it will no work until the CCU is restartet.

fwolter:
Is this ready to be reviewed?

MHerbst:
Yes, would be nice if you could have a look.

bundles/org.openhab.binding.homematic/src/main/java/org/openhab/binding/homematic/internal/communicator/client/RpcClient.java
```diff
@@ -83,14 +85,35 @@ public RpcClient(HomematicConfig config) {
...
+                break;
+            } catch (IOException e) {
+                if (attempt == maxRetryAttempts) {
+                    logger.warn(
+                            "All attempts to register a callback for interface '{}' failed, last exception message: {}",
+                            hmInterface.getName(), e.getMessage());
+                    throw (e);
+                }
+                try {
+                    Thread.sleep(waitTime);
```
fwolter, line 111: The ThingHandler's `initialize()` method must return fast. As this code is called via that method, it must also return fast. Despite that, sleep should be avoided, because it blocks a thread in the schedulers thread pool. You could schedule another try via `scheduler.scheduleWithFixedDelay()`, if `sendMessage()` fails.

bundles/org.openhab.binding.homematic/src/main/java/org/openhab/binding/homematic/internal/communicator/client/RpcClient.java
```diff
@@ -83,14 +85,35 @@ public RpcClient(HomematicConfig config) {
...
+                } catch (InterruptedException e1) {
+                    // ignore
```
fwolter, line 113: The `InterruptedException` is thrown when OH is shut-down. That means the current operation should be finished as fast as possible and no further operations should be done. You can achieve this by not catching the exception here but throw the InterruptedException and catch it at the highest level you can.

bundles/org.openhab.binding.homematic/src/main/resources/OH-INF/thing/bridge.xml
```diff
@@ -95,6 +95,18 @@
 				<advanced>true</advanced>
 				<default>9292</default>
 			</parameter>
+			<parameter name="callbackRegistrationRetries" type="integer" min="1">
+				<label>Number of Callback Registration Retries </label>
```
fwolter, line 99: Labels are expected to be as short as possible. Guideline is 2-3 words with max 20-25 chars. See https://www.openhab.org/docs/developer/bindings/thing-xml.html#formatting-labels-and-descriptions

fwolter (changes requested):

bundles/org.openhab.binding.homematic/src/main/resources/OH-INF/thing/bridge.xml
Flole998: Also there seems to be a space at the end that should be removed IMO

bundles/org.openhab.binding.homematic/src/main/java/org/openhab/binding/homematic/internal/communicator/client/RpcClient.java
MHerbst: Thanks for the review and the tips.
The "fast return" is guaranteed in my opinion, because this method is called via `scheduler.submit`: 
https://github.com/openhab/openhab-addons/blob/e24d43aecbcfc5af551c8e9087d9be6992e5c69a/bundles/org.openhab.binding.homematic/src/main/java/org/openhab/binding/homematic/internal/handler/HomematicBridgeHandler.java#L95

I will replace the `sleep`.

bundles/org.openhab.binding.homematic/src/main/java/org/openhab/binding/homematic/internal/communicator/client/RpcClient.java
MHerbst: If I replace the `sleep`, the catch is no longer needed

bundles/org.openhab.binding.homematic/src/main/resources/OH-INF/thing/bridge.xml
MHerbst: OK, will find a shorter label name.

bundles/org.openhab.binding.homematic/src/main/java/org/openhab/binding/homematic/internal/communicator/client/RpcClient.java
MHerbst: @fwolter I have changed the implementation and replaced the sleep and implemented it as you suggested.

bundles/org.openhab.binding.homematic/src/main/java/org/openhab/binding/homematic/internal/communicator/client/RpcClient.java
MHerbst: @fwolter changed it

bundles/org.openhab.binding.homematic/src/main/resources/OH-INF/thing/bridge.xml
MHerbst: I have changed the configuration and this parameter has been removed

bundles/org.openhab.binding.homematic/src/main/java/org/openhab/binding/homematic/internal/communicator/client/RpcClient.java
```diff
@@ -83,14 +96,39 @@ public RpcClient(HomematicConfig config) {
...
+                logger.debug("Register callback for interface {}, attempt {}", hmInterface.getName(), ++attempt);
+                try {
+                    sendMessage(config.getRpcPort(hmInterface), request);
+                    future.cancel(true);
+                } catch (IOException ex) {
+                    // Ignore, retry
+                }
+            }, INITIAL_CALLBACK_REG_DELAY, CALLBACK_REG_DELAY, TimeUnit.SECONDS);
+            try {
+                future.get(config.getCallbackRegTimeout(), TimeUnit.SECONDS);
```
fwolter, line 122: `get()` will block forever when called on a future returned by `scheduleWithFixedDelay()`. Excerpt from the Javadoc:

```
Returns: a ScheduledFuture representing pending completion ofthe series of repeated tasks. The future's get() method will never return normally,and will throw an exception upon task cancellation orabnormal termination of a task execution.
```

This is called from the handler's `initialize()`, thus `initialize()` will block forever. Why not removing the entire `try` block?

You need to cancel the task when the handler is disposed.

Are you registering the callbacks repeatedly by intention?

fwolter (changes requested):

bundles/org.openhab.binding.homematic/src/main/java/org/openhab/binding/homematic/internal/communicator/client/RpcClient.java
MHerbst: > This is called from the handler's initialize(), thus initialize() will block forever. Why not removing the entire try block?
I have tried it, but without the blocking get, there can be up to 4 parallel requests to different ports (and services) on the Homematic gateway and the gateway simply does not like this. I have tried it this way and it did not work. In some tests I even had to hard reset the gateway because it was completely stuck.

The current implementation of the binding needs a more or less synchronous execution. 
I am not happy with the solution, but with the current implementation, I really see no other way to implement it :-(.

> This is called from the handler's initialize(), thus initialize() 
As far as I can see the initialize in the Bridge handler (HomematicThingHandler#initialize) is not blocked and finishes fast (I have verified it in my tests).

> Are you registering the callbacks repeatedly by intention?
The callback registration is executed after the binding was started or after the connection to the HM gateway was lost (e.g. if the gateway was restarted). In the first case, the first registration request will be successful and the scheduler is not used.

In the latter case, the binding performs a complete reconnect to the Homematic gateway. This includes a new registration of the callbacks. For each possible connection type (HM-RF, HmIP, Groups, CuxD) a separate callback must be registered because on the gateway, a service (daemon) is executed for each connection type listening on different ports. The HmIP service e.g. needs up to 2 min. until it is ready for callback registration requests. Therefore, the binding needs to perform these retries. There is no other way to figure out when a service on the gateway is ready to accept requests. 

It is even worse.  In the first seconds after the gateway restart you even don't get a correct reply to the XML RPC requests. Instead the gateway returns a HTML document with a message "please wait ...".

In my opinion, major parts of the binding's implementation need a complete overhaul. I would also be better to create a bridge for each of the possible connection types (would be a breaking change :-()
But such an overhaul requires time especially for testing.

> You need to cancel the task when the handler is disposed.
OK, will check this.

Flole998: > In my opinion, major parts of the binding's implementation need a complete overhaul.

I agree on that one. Also the channel creation stuff that is "fixed" in another PR is kinda broken by design and fixing that properly would require huge changes. So now we seem to have 2 changes/fixes that would ideally be handled with a complete rewrite of most of the binding....

fwolter: After looking into the code again, I saw that the Future cancels itself when the connection has been established successfully. So, `initialize()` will block as long as the remote side is unavailable. The current implementation calls `sendMessage()` only once (which is also bad in `initialize()`).

So, if I see correctly your code will block `initialize()` forever when the remote side is not available.

MHerbst: @fwolter A `sendMessage` during the Bridge initialization it is not blocking the handler initialization, as you can see in `HomematicBridgeHandler#initialize`:

```
    public void initialize() {
        synchronized (initDisposeLock) {
            isDisposed = false;
            initializeFuture = scheduler.submit(this::initializeInternal);
        }
    }
```
You can believe me, I have tested it (several times).

The binding works correctly. If the gateway is not accessible when the binding is started, all things are set to state "ERROR:BRIDGE". After the gateway comes up this is correctly detected and the thing's state changes to ONLINE.

I know, this solution is at least in parts "ugly as hell". But I have invested really hours of testing and trying to find a better solution without having to rewrite larger part of the binding. Now am I am really fed up with these connection problems and I for the next couple of weeks I don't have the time to work on it.

If you can't accept the PR in its current state, it is OK to me. But in this case, it will remain in this state for at least a month. Or someone else continues to work on this problems.

fwolter: I didn't notice that `initialize()` wasn't the `initialize()` from the ThingHandler, but from a custom class. Would be awesome if this code could be re-written sometime!

fwolter (approved):

PR #11437, opened 2021-10-23
[jsscripting] Update GraalJS to 21.3 to allow method selection via JavaScript

LukasA83:
To allow method selection via javascript in case of graaljs unable to choose right method by itself, Graaljs needed to be updated to >21, e.g. for events.postUpdate with item and state when state is a DecimalType.

Signed-off-by: Lukas Agethen lukas83@gmx.de


LukasA83:
Anything else needed for review from my side?

digitaldan:
I just tested this on my system and everything works as expected, so @LukasA83 the upgrade allows Graaljs to pick `foo(String bar)` or `foo(int bar)` correctly when calling from JS ?

LukasA83:
Well it allows you to define in case there is a conflict, see
https://github.com/oracle/graaljs/issues/37

digitaldan:
Maybe @jpg0 can give his review, but this LGTM

jpg0 (approved):
Looks good to me

LukasA83:
What else is required to get this PR merged?

wborn (approved):
Thanks! :+1: 
We only need to click a few buttons to get it merged. :wink:

PR #11459, opened 2021-10-27
[miio] force start cloudconnector after credentials are updated

marcelrv:


lolodomo (approved):
LGTM

PR #11463, opened 2021-10-27
[zoneminder] Only update state options if they've changed

mhilbush:
There are situations when the binding updates the state options even if the options haven't changed. This results in an unnecessary reload of the page in Basic UI. With this change, the state options are updated only if they change.

### DIFF
--- a/bundles/org.openhab.binding.zoneminder/src/main/java/org/openhab/binding/zoneminder/internal/handler/ZmBridgeHandler.java
+++ b/bundles/org.openhab.binding.zoneminder/src/main/java/org/openhab/binding/zoneminder/internal/handler/ZmBridgeHandler.java
@@ -100,7 +100,6 @@ public class ZmBridgeHandler extends BaseBridgeHandler {
     private @Nullable Future<?> refreshMonitorsJob;
 
     private List<Monitor> savedMonitors = new ArrayList<>();
-    private List<StateOption> savedStateOptions = new ArrayList<>();
 
     private String host = "";
     private boolean useSSL;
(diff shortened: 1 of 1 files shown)


lolodomo:
My immediate thinking is that'll it should rather be done / checked by the core framework. 
@cweitkamp : WDYT?

cweitkamp:
Definitely yes. And IIRC it is already implemented in `BaseDynamicStateDescriptionProvider::setStateOptions(...)`.

https://github.com/openhab/openhab-core/blob/40e23e23ca58fdb39be3b656430e8dcf8b18a344/bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/binding/BaseDynamicStateDescriptionProvider.java#L79

cweitkamp:
Do you think the ordering of the list elements might be the culprit?

mhilbush:
> Do you think the ordering of the list elements might be the culprit?

No. 😉 

Look at my original code. I had the call to `setStateOptions` in the wrong part of the code. It was inside the for loop. 🙄

mhilbush:
I'll update the PR to move those calls to after the for loop. That should resolve the issue. Agree?

cweitkamp:
> Agree?

I do.

mhilbush:
Ok, I believe this is the correct fix. Thanks for helping me find my error.

cweitkamp (approved):
LGTM. Thanks.

@lolodomo Fine for you?

lolodomo (approved):
LGTM

lolodomo:
I realized the PR title is wrong. Can you fix it?

mhilbush:
Good catch. fixed.

Issue #11465, opened 2021-10-28
Renault-api binding.


dougculnane:
Electric car charging rules can be improved if the battery state of the car is known.  There was a Renault ZE service API but this is no longer available and a new renault-api has a complex authentication flow.  There is python implemented code that mimics the mobile app and allows a command line interface to pull data from the Renault API. (https://github.com/hacf-fr/renault-api)

It would be nice to have this available as an easy to use binding so that openHAB users can get car battery (and other) status information in openHAB.

This is a long discussion documenting the issue:
https://community.openhab.org/t/betatest-renault-ze-services-binding/72226

I have implemented a basic binding that translates the python code to a Java.  It works on at least 3 cars according the the forum. https://community.openhab.org/t/betatest-renault-ze-services-binding/72226/79?u=doug_culnane

I will prepare a pull request for the code.

PR #11467, opened 2021-10-28
[renault] Initial Contribution

dougculnane:
Initial renualt-api binding Issue: #11465

Electric car charging rules can be improved if the battery state of the car is known. There was a Renault ZE service API but this is no longer available and a new renault-api has a complex authentication flow. There is python implemented code that mimics the mobile app and allows a command line interface to pull data from the Renault API. (https://github.com/hacf-fr/renault-api)

It would be nice to have this available as an easy to use binding so that openHAB users can get car battery (and other) status information in openHAB.

This is a long discussion documenting the issue:
https://community.openhab.org/t/betatest-renault-ze-services-binding/72226

I have implemented a basic binding that translates the python code to a Java. It works on at least 3 cars according the the forum. https://community.openhab.org/t/betatest-renault-ze-services-binding/72226/79?u=doug_culnane


lolodomo:
I am going to review your binding. 
As a first request, please add the missing entries in the BOM POM and CODE OWNERS files.

bundles/org.openhab.binding.renault/README.md
```diff
@@ -0,0 +1,45 @@
...
+Works on my car (Renault Zoe 50) but I only have one car to test.
+
+
+## Discovery
+
+No discovery
+
+## Binding Configuration
+
+You require your MyRenault credential, locale and VIN for your MyRenault registered car.
```
lolodomo, line 19: There is no binding configuration. This sentence has to be moved to the next section (thing configuration).

bundles/org.openhab.binding.renault/README.md
```diff
@@ -0,0 +1,45 @@
...
+
+The thing has these configuration parameters:
+
+| Parameter         | Description                            | Required |
+|-------------------|----------------------------------------|----------|
+| myRenaultUsername | MyRenault Username.                    | yes      |
+| myRenaultPassword | MyRenault Password.                    | yes      |
+| locale            | MyRenault Location (language_country). | yes      |
+| vin               | Vehicle Identification Number.         | yes      |
+| refreshInterval   | Interval the car is polled in minutes. | yes      |
```
lolodomo, line 31: This one is not required as you defined a default value.

bundles/org.openhab.binding.renault/README.md
```diff
@@ -0,0 +1,45 @@
...
+
+Currently all available channels are read only:
+
+| Channel ID   | Type     | Description                     |
+|--------------|----------|---------------------------------|
+| batterylevel | Number   | State of the battery in %       |
+| hvacstatus   | Switch   | HVAC status switch              |
+| image        | String   | Image URL of MyRenault          |
+| location     | Location | The GPS position of the vehicle |
+| odometer     | Number   | Total distance travelled        |
```
lolodomo, line 43: I would expect `Number:Length`

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/RenaultHandler.java
```diff
@@ -0,0 +1,113 @@
...
+            pollingJob = null;
+        }
+        super.dispose();
+    }
+
+    private void getStatus() {
+        MyRenaultHttpSession httpSession;
+        try {
+            httpSession = new MyRenaultHttpSession(this.config);
+            updateStatus(ThingStatus.ONLINE, ThingStatusDetail.NONE);
```
lolodomo, line 85: Just `updateStatus(ThingStatus.ONLINE);`
In case httpSession.updateCarData can trigger an exception, I will move this line just after.

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/RenaultHandler.java
```diff
@@ -0,0 +1,113 @@
...
+            updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.COMMUNICATION_ERROR, e.getMessage());
+        }
+    }
+
+    private void updateState(Car car) {
+        if (car.batteryLevel != null) {
+            updateState(CHANNEL_BATTERY_LEVEL, new DecimalType(car.batteryLevel));
+        }
+        if (car.hvacstatus != null) {
+            updateState(CHANNEL_HVAC_STATUS, (car.hvacstatus ? OnOffType.ON : OnOffType.OFF));
```
lolodomo, line 100: You can rahter use: `updateState(CHANNEL_HVAC_STATUS, OnOffType.from(car.hvacstatus.booleanValue));`

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/RenaultHandler.java
```diff
@@ -0,0 +1,113 @@
...
+        }
+        if (car.imageURL != null) {
+            updateState(CHANNEL_IMAGE, new StringType(car.imageURL));
+        }
+        if (car.gpsLatitude != null && car.gpsLongitude != null) {
+            updateState(CHANNEL_LOCATION,
+                    new PointType(new DecimalType(car.gpsLatitude), new DecimalType(car.gpsLongitude)));
+        }
+        if (car.odometer != null) {
+            updateState(CHANNEL_ODOMETER, new DecimalType(car.odometer));
```
lolodomo, line 110: To use UoM, I suppose the value is in kilometers, so you case do it like that:
`updateState(CHANNEL_ODOMETER, new QuantityType<Length>(car.odometer.doubleValue, KILO(METRE)));`

lolodomo (changes requested):
Review part 1 / 2

bundles/org.openhab.binding.renault/src/main/resources/OH-INF/thing/thing-types.xml
```diff
@@ -0,0 +1,115 @@
...
+		<state readOnly="true"></state>
+	</channel-type>
+	<channel-type id="location" advanced="false">
+		<item-type>Location</item-type>
+		<label>Location</label>
+		<description>The GPS position of the vehicle</description>
+		<state readOnly="true"></state>
+	</channel-type>
+	<channel-type id="odometer">
+		<item-type>Number</item-type>
```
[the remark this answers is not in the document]
dougculnane: Uom? Units of Measure?  

I understand this:
<item-type>Number:Length</item-type>
but I am not sure what you mean by "Please use UoM" thanks.

FYI: I think the odometer is in km (or at least it is in Austria).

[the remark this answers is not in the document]
lolodomo: Yes, Units of Measure

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/RenaultHandler.java
dougculnane: Establishing the session has to work or else nothing works.  The "get car data" can fail at each stage for reasons that are not clear.  So from testing in the openhab forum it seams better for some cars to get the session and get the data you can and let some channels be null.  Some features like HVAC or Location do not work for some accounts but ther users are happy to get the battery level.  So the binding tries to get what it can and ignore (but log) errors once the session is established.

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/RenaultHandler.java
dougculnane: The JSON structure from which this parameter is collected is called "totalMileage" but my de_AT car value is km.  So I think km is the standard but I can not be sure for all locations.

dougculnane: I guess I will set KM and let the US and UK refactor to the better system of units. ;-)

dougculnane:
Thanks @lolodomo for some great feedback.  I think I have got all your points and I will build and test the changes locally this weekend.  I hope to push a tested version ASAP.

Thanks again.

lolodomo:
I will review the last part (comm part) probably tomorrow.

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/RenaultHandler.java
```diff
@@ -0,0 +1,113 @@
...
+    }
+
+    private void getStatus() {
+        MyRenaultHttpSession httpSession;
+        try {
+            httpSession = new MyRenaultHttpSession(this.config);
+            updateStatus(ThingStatus.ONLINE, ThingStatusDetail.NONE);
+            httpSession.updateCarData(this.config);
+            updateState(httpSession.getCar());
+        } catch (Exception e) {
```
lolodomo, line 88: Note: I am ok with such high level exception catching here because this method is called in scheduleWithFixedDelay

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/renault/api/MyRenaultHttpSession.java
```diff
@@ -0,0 +1,266 @@
...
+        car = new Car();
+        this.httpClient = new HttpClient(new SslContextFactory(true));
+        this.constants = new Constants(config.locale);
+
+        httpClient.start();
+        login(config);
+        getAccountInfo(config);
+        getJWT(config);
+        getAccountID(config);
+        getVehicle(config);
```
lolodomo, line 62: Maybe it would be better to move these calls in a separate method and keep a constructor that just init variables and throws no exception.

lolodomo (changes requested):
Review part 2 of 2

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/RenaultHandler.java
lolodomo: in case httpSession.updateCarData throws an exception, with your current code, you will first change the state to ONLINE and then, to OFFLINE.
I believe `getStatus` should update the thing status only one time in all cases.
If you don't care about exceptions thrown by `httpSession.updateCarData`, maybe do 2 separate try/catch and do not set the thing status to OFFLINE for `updateCarData`.

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/renault/api/MyRenaultHttpSession.java
```diff
@@ -0,0 +1,266 @@
...
+    private String personId;
+    private String gigyaDataCenter;
+    private String jwt;
+
+    private final Logger logger = LoggerFactory.getLogger(MyRenaultHttpSession.class);
+
+    public MyRenaultHttpSession(RenaultConfiguration config) throws Exception {
+
+        car = new Car();
+        this.httpClient = new HttpClient(new SslContextFactory(true));
```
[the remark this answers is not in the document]
dougculnane: No reason other than it was in the class path and worked.  However the "SslContextFactory(true)" hides a sin I forgot I had used to get things working inititally.  I will change this as I need to cert check in on a proper binding.  (sorry) 

What is the best (as in most openHAB friendly) client HTTPClient to use?

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/renault/api/Constants.java
```diff
@@ -0,0 +1,237 @@
...
+    private String gigyaApiKey = "gigya-api-key";
+    private String gigyaRootUrl = "gigya-root-url";
+    private String kamereonApiKey = "kamereon-api-key";
+    private String kamereonRootUrl = "kamereon-root-url";
+
+    static private final String GIGYA_URL_EU = "https://accounts.eu1.gigya.com";
+    static private final String GIGYA_URL_US = "https://accounts.us1.gigya.com";
+    static private final String KAMEREON_APIKEY = "Ae9FDWugRxZQAGm3Sxgk7uJn6Q4CGEA2";
+    static private final String KAMEREON_URL_EU = "https://api-wired-prod-1-euw1.wrd-aws.com";
+    static private final String KAMEREON_URL_US = "https://api-wired-prod-1-usw2.wrd-aws.com";
```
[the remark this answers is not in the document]
dougculnane: Is the "SAT report" this : /target/summary_report.html ?  I will check it now that I am aware of it.  However if you mean something else please let me know.

dougculnane:
> Review part 2 of 2

Thanks @lolodomo I have a couple of questions in the comments and some work to do.  I will do and test these changes and get back with an updated version.  

Thanks again.

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/renault/api/Constants.java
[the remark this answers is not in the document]
lolodomo: Yes,  it is what I mean.

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/renault/api/MyRenaultHttpSession.java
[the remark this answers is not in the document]
lolodomo: In your thing handler, initialize your HTTP client like this:
```
this.httpClient = httpClientFactory.getCommonHttpClient();
```
And then pass it to your thing handler.
Most of bindings are doing that, look at the remote openHAB binding as an example.
You create your own HTTP client only when you need a particular HTTP client setup, because of course you must not change the setup of the common client.
In openHAB, the recommended HTTP client is the Jetty client (what is provided by `httpClientFactory.getCommonHttpClient()`. So you will then have no other code to change.

bundles/org.openhab.binding.renault/src/main/resources/OH-INF/thing/thing-types.xml
```diff
@@ -0,0 +1,115 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<thing:thing-descriptions bindingId="renault"
+	xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+	xmlns:thing="https://openhab.org/schemas/thing-description/v1.0.0"
+	xsi:schemaLocation="https://openhab.org/schemas/thing-description/v1.0.0 https://openhab.org/schemas/thing-description-1.0.0.xsd">
+
+	<!-- Sample Thing Type -->
```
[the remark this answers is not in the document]
dougculnane: Done

bundles/org.openhab.binding.renault/src/main/resources/OH-INF/thing/thing-types.xml
```diff
@@ -0,0 +1,115 @@
...
+	<thing-type id="car">
+		<!-- Comment this in for Things which need a Bridge to function.
+			<supported-bridge-type-refs>
+			<bridge-type-ref id="renaultBridge" />
+			</supported-bridge-type-refs>
+		-->
+
```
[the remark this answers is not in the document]
dougculnane: Done

bundles/org.openhab.binding.renault/src/main/resources/OH-INF/thing/thing-types.xml
```diff
@@ -0,0 +1,115 @@
...
+		<label>Renault Binding</label>
```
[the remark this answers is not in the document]
dougculnane: Done

bundles/org.openhab.binding.renault/src/main/resources/OH-INF/thing/thing-types.xml
```diff
@@ -0,0 +1,115 @@
...
+		<description>Renault Binding for MyRenault registered car</description>
```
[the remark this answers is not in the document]
dougculnane: Done

bundles/org.openhab.binding.renault/src/main/resources/OH-INF/thing/thing-types.xml
```diff
@@ -0,0 +1,115 @@
...
+
+		<channels>
+			<channel id="batterylevel" typeId="batterylevel"/>
```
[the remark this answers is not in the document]
dougculnane: Done

bundles/org.openhab.binding.renault/src/main/resources/OH-INF/thing/thing-types.xml
```diff
@@ -0,0 +1,115 @@
...
+		</config-description>
+	</thing-type>
+
+	<!-- Sample Channel Type -->
+	<channel-type id="batterylevel">
+		<item-type>Number</item-type>
+		<label>Battery Level</label>
+		<description>State of the battery in %</description>
+		<state pattern="%.1f %%" readOnly="true"></state>
+	</channel-type>
```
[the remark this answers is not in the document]
dougculnane: Done

bundles/org.openhab.binding.renault/src/main/resources/OH-INF/thing/thing-types.xml
```diff
@@ -0,0 +1,115 @@
...
+	<channel-type id="hvacstatus">
+		<item-type>Switch</item-type>
+		<label>HVAC</label>
+		<description>HVAC status</description>
```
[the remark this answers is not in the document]
dougculnane: "HVAC Status" for the label
 removed description.

bundles/org.openhab.binding.renault/src/main/resources/OH-INF/thing/thing-types.xml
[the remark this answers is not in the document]
dougculnane: Not sure how to add the Uof to the channel-type.

pattern="%.1f %unit%" ?

bundles/org.openhab.binding.renault/src/main/resources/OH-INF/thing/thing-types.xml
```diff
@@ -0,0 +1,115 @@
...
+		-->
+
+		<label>Renault Binding</label>
+		<description>Renault Binding for MyRenault registered car</description>
+
+		<channels>
+			<channel id="batterylevel" typeId="batterylevel"/>
+			<channel id="hvacstatus" typeId="hvacstatus"/>
+			<channel id="image" typeId="image"/>
+			<channel id="location" typeId="location"/>
```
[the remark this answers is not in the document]
dougculnane: Done

bundles/org.openhab.binding.renault/src/main/resources/OH-INF/thing/thing-types.xml
```diff
@@ -0,0 +1,115 @@
...
+		<label>Image URL</label>
+		<description>Image URL of MyRenault</description>
+		<state readOnly="true"></state>
+	</channel-type>
+	<channel-type id="location" advanced="false">
+		<item-type>Location</item-type>
+		<label>Location</label>
+		<description>The GPS position of the vehicle</description>
+		<state readOnly="true"></state>
+	</channel-type>
```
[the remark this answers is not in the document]
dougculnane: Done

bundles/org.openhab.binding.renault/src/main/resources/OH-INF/thing/thing-types.xml
```diff
@@ -0,0 +1,115 @@
...
+			</parameter>
+			<parameter name="vin" type="text" required="true">
+				<label>VIN</label>
+				<description>Vehicle Identification Number</description>
+			</parameter>
+			<parameter name="refreshInterval" type="integer" unit="s" min="1">
+				<label>Refresh Interval</label>
+				<description>Interval the car is polled in minutes.</description>
+				<default>10</default>
+			</parameter>
```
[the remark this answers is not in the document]
dougculnane: Changed to unit="m"

bundles/org.openhab.binding.renault/src/main/resources/OH-INF/thing/thing-types.xml
```diff
@@ -0,0 +1,115 @@
...
+			<channel id="hvacstatus" typeId="hvacstatus"/>
+			<channel id="image" typeId="image"/>
+			<channel id="location" typeId="location"/>
+			<channel id="odometer" typeId="odometer"/>
+		</channels>
+		<config-description>
+
+			<parameter name="myRenaultUsername" type="text" required="true">
+				<label>MyRenault Username</label>
+				<description>MyRenault Username</description>
```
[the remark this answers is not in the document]
dougculnane: Removed description

bundles/org.openhab.binding.renault/src/main/resources/OH-INF/thing/thing-types.xml
```diff
@@ -0,0 +1,115 @@
...
+			</parameter>
+			<parameter name="myRenaultPassword" type="text" required="true">
+				<context>password</context>
+				<label>MyRenault Password</label>
+				<description>MyRenault Password</description>
```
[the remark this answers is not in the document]
dougculnane: Removed description

bundles/org.openhab.binding.renault/src/main/resources/OH-INF/thing/thing-types.xml
```diff
@@ -0,0 +1,115 @@
...
+			</parameter>
+			<parameter name="locale" type="text" required="true">
+				<label>Locale</label>
+				<description>MyRenault Location</description>
```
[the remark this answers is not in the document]
dougculnane: Done.  Better description now.

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/RenaultConfiguration.java
```diff
@@ -0,0 +1,27 @@
...
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package org.openhab.binding.renault.internal;
+
+/**
+ * The {@link RenaultConfiguration} class contains fields mapping thing configuration parameters.
+ *
+ * @author Doug Culnane - Initial contribution
+ */
+public class RenaultConfiguration {
```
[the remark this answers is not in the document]
dougculnane: Done.  @NonNullByDefault is new to me so this was interesting.  I hope i got the implementation right.

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/RenaultConfiguration.java
```diff
@@ -0,0 +1,27 @@
...
+
+    public String myRenaultUsername;
+    public String myRenaultPassword;
+    public String locale;
+    public String vin;
+    public int refreshInterval;
```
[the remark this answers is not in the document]
dougculnane: Done.  Set other to empty string to keep the @NonNullByDefault happy.

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/RenaultConfiguration.java
```diff
@@ -0,0 +1,27 @@
...
+ * The {@link RenaultConfiguration} class contains fields mapping thing configuration parameters.
+ *
+ * @author Doug Culnane - Initial contribution
+ */
+public class RenaultConfiguration {
+
+    public String myRenaultUsername;
+    public String myRenaultPassword;
+    public String locale;
+    public String vin;
```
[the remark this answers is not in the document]
dougculnane: Done.

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/RenaultHandler.java
```diff
@@ -0,0 +1,113 @@
...
+    @Override
+    public void handleCommand(ChannelUID channelUID, Command command) {
+        // This binding only polls status data automatically.
+    }
+
+    @Override
+    public void initialize() {
+        updateStatus(ThingStatus.UNKNOWN);
+        this.config = getConfigAs(RenaultConfiguration.class);
+
```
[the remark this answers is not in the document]
dougculnane: Added a Validate configuration code section.

bundles/org.openhab.binding.renault/README.md
dougculnane: Done.

bundles/org.openhab.binding.renault/README.md
dougculnane: Done.

bundles/org.openhab.binding.renault/README.md
dougculnane: Done

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/RenaultHandler.java
dougculnane: Done.  Thanks for the advice and pointer to an example implementation.

dougculnane: Done. 2 stages are clearer now.

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/RenaultHandler.java
dougculnane: Done

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/RenaultHandler.java
dougculnane: Done.

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/renault/api/Car.java
```diff
@@ -0,0 +1,98 @@
...
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.google.gson.JsonArray;
+import com.google.gson.JsonElement;
+import com.google.gson.JsonObject;
+
+/**
+ * @author Doug Culnane - Initial contribution
```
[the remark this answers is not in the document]
dougculnane: Done

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/renault/api/Car.java
```diff
@@ -0,0 +1,98 @@
...
+    public Double odometer;
+    public String imageURL;
+    public Double gpsLatitude;
+    public Double gpsLongitude;
+
+    public void setBatteryStatus(JsonObject responseJson) {
+        try {
+            batteryLevel = responseJson.get("data").getAsJsonObject().get("attributes").getAsJsonObject()
+                    .get("batteryLevel").getAsDouble();
+        } catch (Exception e) {
```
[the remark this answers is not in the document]
dougculnane: Done.  I have used more specific exceptions in the implementation.

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/renault/api/Car.java
```diff
@@ -0,0 +1,98 @@
...
+            logger.error("Error {} parsing Battery Status: {}", e, responseJson);
```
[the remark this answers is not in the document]
dougculnane: Done.

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/renault/api/Car.java
```diff
@@ -0,0 +1,98 @@
...
+        try {
+            JsonArray assetsJson = responseJson.get("assets").getAsJsonArray();
+            for (JsonElement asset : assetsJson) {
+                if (asset.getAsJsonObject().get("assetType").getAsString().equals("PICTURE")) {
+                    JsonArray renditions = asset.getAsJsonObject().get("renditions").getAsJsonArray();
+                    for (JsonElement rendition : renditions) {
+                        if (rendition.getAsJsonObject().get("resolutionType").getAsString()
+                                .equals("ONE_MYRENAULT_SMALL")) {
+                            imageURL = rendition.getAsJsonObject().get("url").getAsString();
+                        }
```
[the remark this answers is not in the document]
dougculnane: Done.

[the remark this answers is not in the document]
dougculnane: In fact I only need to get the image once.  So I will skip this if it is set to a non empty value.

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/renault/api/Constants.java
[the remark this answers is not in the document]
dougculnane: Checked and cleared the SAT this time.

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/renault/api/Constants.java
```diff
@@ -0,0 +1,237 @@
...
+
+    static final String LOCALE_BASE_URL = "https://renault-wrd-prod-1-euw1-myrapp-one.s3-eu-west-1.amazonaws.com";
```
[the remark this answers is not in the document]
dougculnane: Deleted it.

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/renault/api/Constants.java
```diff
@@ -0,0 +1,237 @@
...
+package org.openhab.binding.renault.internal.renault.api;
+
+/**
+ * Constants for Renault API.
+ * 
+ * https://github.com/hacf-fr/renault-api/blob/main/src/renault_api/const.py
+ * 
+ * @author Doug Culnane - Initial contribution
+ */
+public class Constants {
```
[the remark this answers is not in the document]
dougculnane: Done

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/renault/api/Constants.java
```diff
@@ -0,0 +1,237 @@
...
+
+    private String gigyaApiKey = "gigya-api-key";
+    private String gigyaRootUrl = "gigya-root-url";
+    private String kamereonApiKey = "kamereon-api-key";
+    private String kamereonRootUrl = "kamereon-root-url";
```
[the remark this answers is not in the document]
dougculnane: Done

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/renault/api/MyRenaultHttpSession.java
```diff
@@ -0,0 +1,266 @@
...
+import org.openhab.binding.renault.internal.RenaultConfiguration;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.google.gson.JsonArray;
+import com.google.gson.JsonObject;
+import com.google.gson.JsonParser;
+
+/**
+ * Translation of this python code: https://github.com/hacf-fr/renault-api
```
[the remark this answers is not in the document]
dougculnane: "This is a Java version of the python renault-api project developed here: https://github.com/hacf-fr/renault-api"

I hope this is clearer.

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/renault/api/MyRenaultHttpSession.java
```diff
@@ -0,0 +1,266 @@
...
+ *
+ * @author Doug Culnane - Initial contribution
+ */
+public class MyRenaultHttpSession {
```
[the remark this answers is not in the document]
dougculnane: Done.

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/renault/api/MyRenaultHttpSession.java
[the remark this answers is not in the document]
dougculnane: Done.  Thanks for the pointer to an implementation to copy.

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/renault/api/MyRenaultHttpSession.java
```diff
@@ -0,0 +1,266 @@
...
+        logger.debug("URL: {}/accounts.login", this.constants.getGigyaRootUrl());
+        ContentResponse response = this.httpClient.FORM(this.constants.getGigyaRootUrl() + "/accounts.login", fields);
+        if (HttpStatus.OK_200 == response.getStatus()) {
+            JsonObject responseJson = new JsonParser().parse(response.getContentAsString()).getAsJsonObject();
+            JsonObject sessionInfoJson = responseJson.getAsJsonObject("sessionInfo");
+            cookieValue = sessionInfoJson.get("cookieValue").getAsString();
+            logger.debug("Cookie: {}", cookieValue);
+        } else {
+            logger.error("Response: [{}] {}\n{}", response.getStatus(), response.getReason(),
+                    response.getContentAsString());
```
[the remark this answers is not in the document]
dougculnane: Done set to warning.  Not all features are supported so some stuff will not work.  I would like to be verbose about this so the user has a chance to understand that the reason it does not work is because the server said "no".

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/renault/api/MyRenaultHttpSession.java
```diff
@@ -0,0 +1,266 @@
...
+
+    private void getVehicle(RenaultConfiguration config) throws Exception {
+
+        Request request = getKamereonRequest("/commerce/v1/accounts/" + kamereonaccountId + "/vehicles/" + config.vin
+                + "/details?country=" + getCountry(config));
+
+        ContentResponse response = request.send();
+        if (HttpStatus.OK_200 == response.getStatus()) {
+            JsonObject responseJson = new JsonParser().parse(response.getContentAsString()).getAsJsonObject();
+            logger.debug("responseJson: {} ", responseJson.toString());
```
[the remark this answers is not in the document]
dougculnane: New getKamereonResponse method has better log statements with URL etc.

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/renault/api/MyRenaultHttpSession.java
```diff
@@ -0,0 +1,266 @@
...
+        }
+    }
+
+    private Request getKamereonRequest(String path) {
+        Request request = this.httpClient.newRequest(this.constants.getKamereonRootUrl() + path);
+        request.method(HttpMethod.GET);
+        request.getHeaders().put(new HttpField("Content-type", "application/vnd.api+json"));
+        request.getHeaders().put(new HttpField("apikey", this.constants.getKamereonApiKey()));
+        request.getHeaders().put(new HttpField("x-kamereon-authorization", "Bearer " + kamereonToken));
+        request.getHeaders().put(new HttpField("x-gigya-id_token", jwt));
```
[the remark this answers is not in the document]
dougculnane: Done

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/RenaultHandler.java
lolodomo: You forgot to add try/catch around the second stage (updateCarData). Just do nothing in the catch.

bundles/org.openhab.binding.renault/src/main/resources/OH-INF/thing/thing-types.xml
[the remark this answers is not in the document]
lolodomo: Yes, exactly, please add the `%unit%` in the pattern, as you propose in your comment.

bundles/org.openhab.binding.renault/src/main/resources/OH-INF/thing/thing-types.xml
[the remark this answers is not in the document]
lolodomo: I have a big doubt. Did you check the documentation ?

bundles/org.openhab.binding.renault/src/main/resources/OH-INF/thing/thing-types.xml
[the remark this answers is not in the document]
lolodomo: Please keep the label very short, so "MyRenault Location" rather than "MyRenault Location (language_country)".

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/RenaultHandler.java
```diff
@@ -0,0 +1,147 @@
...
+    }
+
+    @Override
+    public void handleCommand(ChannelUID channelUID, Command command) {
+        // This binding only polls status data automatically.
+    }
+
+    @Override
+    public void initialize() {
+        updateStatus(ThingStatus.UNKNOWN);
```
lolodomo, line 72: Please move this just after all your cchecks, this will avoid multiple status changes.

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/RenaultHandler.java
```diff
@@ -0,0 +1,147 @@
...
+        this.config = getConfigAs(RenaultConfiguration.class);
+
+        // Validate configuration
+        if (this.config.myRenaultUsername.trim().length() == 0) {
```
lolodomo, line 76: I suggest using `isBlank()` instead of `.trim().length() == 0`
Same for others below.

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/RenaultHandlerFactory.java
```diff
@@ -0,0 +1,66 @@
...
+ * The {@link RenaultHandlerFactory} is responsible for creating things and thing
+ * handlers.
+ *
+ * @author Doug Culnane - Initial contribution
+ */
+@NonNullByDefault
+@Component(configurationPid = "binding.renault", service = ThingHandlerFactory.class)
+public class RenaultHandlerFactory extends BaseThingHandlerFactory {
+
+    private final HttpClient httpClient;
```
lolodomo, line 42: Minor: please move just after the constant.

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/renault/api/MyRenaultHttpSession.java
```diff
@@ -0,0 +1,224 @@
...
+    private @Nullable String jwt;
+
+    private final Logger logger = LoggerFactory.getLogger(MyRenaultHttpSession.class);
+
+    public MyRenaultHttpSession(RenaultConfiguration config, HttpClient httpClient) {
+        this.config = config;
+        this.httpClient = httpClient;
+    }
+
+    public void initSesssion() throws Exception {
```
lolodomo, line 63: Please throws RenaultException, InterruptedException, ExecutionException, TimeoutException rather than Exception.

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/renault/api/Car.java
```diff
@@ -0,0 +1,127 @@
...
+public class Car {
+
+    private final Logger logger = LoggerFactory.getLogger(Car.class);
+
+    public Double batteryLevel = Double.valueOf(-1);
+    public Boolean hvacstatus = false;
+    public Double odometer = Double.valueOf(0);
+    public String imageURL = "";
+    public Double gpsLatitude = Double.valueOf(0);
+    public Double gpsLongitude = Double.valueOf(0);
```
lolodomo, line 40: I do not understand why you are now setting a default value, non null. Because in your thing handler (method updateState), the null value is your trigger to update the channel.
This look inconsistent to me.
You should probably add `@Nullable`

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/renault/api/Car.java
```diff
@@ -0,0 +1,127 @@
...
+            if (attributes != null) {
+                if (attributes.get("gpsLatitude") != null) {
+                    gpsLatitude = attributes.get("gpsLatitude").getAsDouble();
+                }
+                if (attributes.get("gpsLongitude") != null) {
+                    gpsLongitude = attributes.get("gpsLongitude").getAsDouble();
+                }
+            }
+        } catch (IllegalStateException | ClassCastException e) {
+            logger.warn("Error {} parsing Cockpit: {}", e.getMessage(), responseJson);
```
lolodomo, line 87: ... parsing Location

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/RenaultHandler.java
```diff
@@ -0,0 +1,147 @@
...
+    }
+
+    private void getStatus() {
+        MyRenaultHttpSession httpSession = new MyRenaultHttpSession(this.config, httpClient);
+        try {
+            httpSession.initSesssion();
+            updateStatus(ThingStatus.ONLINE);
+        } catch (Exception e) {
+            httpSession = null;
+            logger.error("Error My Renault Http Session.", e);
```
lolodomo, line 120: Please do not use ERROR level, so rather WARNING.

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/renault/api/Car.java
```diff
@@ -0,0 +1,127 @@
...
+                                if (rendition.getAsJsonObject().get("resolutionType") != null
+                                        && rendition.getAsJsonObject().get("resolutionType").getAsString()
+                                                .equals("ONE_MYRENAULT_SMALL")) {
+                                    imageURL = rendition.getAsJsonObject().get("url").getAsString();
+                                    break;
+                                }
+                            }
+                        }
+                    }
+                    if (imageURL.length() > 0) {
```
lolodomo, line 110: First, it is better to use `!imageURL.isEmpty()`
But the real problem is that imageURL is a class member that could already have a value before this method is called. So you should probably use a local variable first and then finally set imageURL.

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/renault/api/MyRenaultHttpSession.java
```diff
@@ -0,0 +1,224 @@
...
+        this.constants = new Constants(config.locale);
```
lolodomo, line 64: You should move initializing `constants` in your constructor. Like that, no need to "tag" it as @Nullable.

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/renault/api/MyRenaultHttpSession.java
```diff
@@ -0,0 +1,224 @@
...
+        httpClient.start();
```
lolodomo, line 65: You must not start/stop the common HTTP client (it will be already started).

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/renault/api/MyRenaultHttpSession.java
```diff
@@ -0,0 +1,224 @@
...
+        Fields fields = new Fields();
+        fields.add("ApiKey", this.constants.getGigyaApiKey());
+        fields.add("loginID", config.myRenaultUsername);
+        fields.add("password", config.myRenaultPassword);
+        logger.debug("URL: {}/accounts.login", this.constants.getGigyaRootUrl());
+        ContentResponse response = httpClient.FORM(this.constants.getGigyaRootUrl() + "/accounts.login", fields);
+        if (HttpStatus.OK_200 == response.getStatus()) {
+            JsonObject responseJson = JsonParser.parseString(response.getContentAsString()).getAsJsonObject();
+            JsonObject sessionInfoJson = responseJson.getAsJsonObject("sessionInfo");
+            cookieValue = sessionInfoJson.get("cookieValue").getAsString();
```
lolodomo, line 90: I assume there are possible exceptions for these 3 lines that are not listed in the method prototype ? To be added in this case.
Same remark for the other methods.

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/renault/api/exceptions/RenaultException.java
```diff
@@ -0,0 +1,28 @@
...
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package org.openhab.binding.renault.internal.renault.api.exceptions;
+
+import org.eclipse.jdt.annotation.NonNullByDefault;
+
+/**
+ * @author Doug Culnane - Initial contribution
```
lolodomo, line 18: Please add a short description

lolodomo (changes requested):
New comments after your last set of changes.

bundles/org.openhab.binding.renault/src/main/resources/OH-INF/thing/thing-types.xml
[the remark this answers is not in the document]
lolodomo: This is clearly not easy to find !
https://www.openhab.org/schemas/config-description-1.0.0.xsd
The proper unit value to be used is "min".

lolodomo:
@dougculnane : thank you for all your changes. I marked as "resolved" most of my old comments. Please take a look to the ones which are not yet marked as resolved and please consider the new ones.
We are not too far from something clean.

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/RenaultHandler.java
dougculnane: I used StringUtils.isBlank() but the SAT report said that org.apache.commons.lang3.StringUtils is not to be used for some reason so I did this in stead.  Guessing that the NonNullByDefault handled the null check.  Can I use StringUtils or is there some other isBlank() method I should use?

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/renault/api/Car.java
dougculnane: I did this because I do not really know how to use the @NonNullByDefault properly.  Setting @Nullable caused a compile error for this code...

`        if (car.odometer != null) {
            updateState(CHANNEL_ODOMETER, new QuantityType<Length>(car.odometer, KILO(METRE)));
        }
`

required 'java.lang.@NonNull Number' but this expression has type 'java.lang.@Nullable Double'


I would like to use plain old Java and have the car values starting as null and then set properly.  I do not want defaults and I only want to update status if the value is set properly (not NULL).  However the SAT report say use the @NonNullByDefault on the car.  I would rather not or I need to know the proper trick here?

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/renault/api/Car.java
dougculnane: OK that is the isEmpty() method... Got it.

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/RenaultHandler.java
dougculnane: Got it.  String Object... isEmpty() !!!

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/renault/api/Car.java
dougculnane: Got a fix for this.  I will push @Nullable code. I am starting to understand better now.

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/RenaultHandler.java
lolodomo: IsBlank from java String (not StringUtils). It is part of java String in java 11.
https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/String.html#isBlank()

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/RenaultHandler.java
dougculnane: Done

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/RenaultHandlerFactory.java
dougculnane: Done

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/renault/api/MyRenaultHttpSession.java
dougculnane: Done

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/renault/api/Car.java
dougculnane: Done

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/RenaultHandler.java
dougculnane: Done

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/renault/api/Car.java
dougculnane: Done

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/renault/api/MyRenaultHttpSession.java
dougculnane: Nice.  Done.

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/renault/api/MyRenaultHttpSession.java
dougculnane: Done

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/renault/api/MyRenaultHttpSession.java
dougculnane: I think I got then all now.

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/renault/api/exceptions/RenaultException.java
dougculnane: Done

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/RenaultHandler.java
dougculnane: done

dougculnane:
@lolodomo Thanks for some good tips and feedback.  I have got all currently open issues resolved or attempted to resolve them now.

Thanks again.

bundles/org.openhab.binding.renault/src/main/resources/OH-INF/thing/thing-types.xml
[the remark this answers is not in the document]
dougculnane: Done

bundles/org.openhab.binding.renault/src/main/resources/OH-INF/thing/thing-types.xml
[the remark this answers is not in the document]
dougculnane: Done

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/RenaultHandler.java
```diff
@@ -0,0 +1,147 @@
...
+        }
+        if (car.imageURL != null && !car.imageURL.isEmpty()) {
+            updateState(CHANNEL_IMAGE, new StringType(car.imageURL));
+        }
+        if (car.gpsLatitude != null && car.gpsLongitude != null) {
+            updateState(CHANNEL_LOCATION,
+                    new PointType(new DecimalType(car.gpsLatitude), new DecimalType(car.gpsLongitude)));
+        }
+        if (car.odometer != null) {
+            updateState(CHANNEL_ODOMETER, new QuantityType<Length>(Double.valueOf(car.odometer), KILO(METRE)));
```
lolodomo, line 144: Here is the full method to avoid compilation warnings (NPE):
```
    private void updateState(Car car) {
        Double batteryLevel = car.batteryLevel;
        if (batteryLevel != null) {
            updateState(CHANNEL_BATTERY_LEVEL, new DecimalType(batteryLevel.doubleValue()));
        }
        Boolean hvacstatus = car.hvacstatus;
        if (hvacstatus != null) {
            updateState(CHANNEL_HVAC_STATUS, OnOffType.from(hvacstatus.booleanValue()));
        }
        String imageURL = car.imageURL;
        if (imageURL != null && !imageURL.isEmpty()) {
            updateState(CHANNEL_IMAGE, new StringType(imageURL));
        }
        Double latitude = car.gpsLatitude;
        Double longitude = car.gpsLongitude;
        if (latitude != null && longitude != null) {
            updateState(CHANNEL_LOCATION,
                    new PointType(new DecimalType(latitude.doubleValue()), new DecimalType(longitude.doubleValue())));
        }
        Double odometer = car.odometer;
        if (odometer != null) {
            updateState(CHANNEL_ODOMETER, new QuantityType<Length>(odometer.doubleValue(), KILO(METRE)));
        }
    }
```

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/RenaultHandler.java
```diff
@@ -0,0 +1,147 @@
...
+    }
+
+    @Override
+    public void dispose() {
+        if (pollingJob != null) {
+            pollingJob.cancel(true);
+            pollingJob = null;
+        }
+        super.dispose();
+    }
```
lolodomo, line 111: Here is the full method to avoid compilation warnings (NPE):
```
    public void dispose() {
        ScheduledFuture<?> job = pollingJob;
        if (job != null) {
            job.cancel(true);
            pollingJob = null;
        }
        super.dispose();
    }
```

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/RenaultHandler.java
```diff
@@ -0,0 +1,147 @@
...
+            updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.CONFIGURATION_ERROR,
+                    "The refresh interval mush to be larger than 1");
+            return;
+        }
+        updateStatus(ThingStatus.UNKNOWN);
+
+        // Background initialization:
+        if (pollingJob == null || pollingJob.isCancelled()) {
+            pollingJob = scheduler.scheduleWithFixedDelay(this::getStatus, 0, config.refreshInterval, TimeUnit.MINUTES);
+        }
```
lolodomo, line 101: Here is how to avoid compilation warnings (NPE):
```
        ScheduledFuture<?> job = pollingJob;
        if (job == null || job.isCancelled()) {
            pollingJob = scheduler.scheduleWithFixedDelay(this::getStatus, 0, config.refreshInterval, TimeUnit.MINUTES);
        }
```

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/renault/api/Car.java
```diff
@@ -0,0 +1,127 @@
...
+                                if (rendition.getAsJsonObject().get("resolutionType") != null
+                                        && rendition.getAsJsonObject().get("resolutionType").getAsString()
+                                                .equals("ONE_MYRENAULT_SMALL")) {
+                                    imageURL = rendition.getAsJsonObject().get("url").getAsString();
+                                    break;
+                                }
+                            }
+                        }
+                    }
+                    if (!imageURL.isEmpty()) {
```
lolodomo, line 110: There is a compilation warning here and your code is not good as imageURL could have been set previously.
Here is my proposal using a local variable:
```
                JsonArray assetsJson = responseJson.get("assets").getAsJsonArray();
                String url = null;
                for (JsonElement asset : assetsJson) {
                    if (asset.getAsJsonObject().get("assetType") != null
                            && asset.getAsJsonObject().get("assetType").getAsString().equals("PICTURE")) {
                        if (asset.getAsJsonObject().get("renditions") != null) {
                            JsonArray renditions = asset.getAsJsonObject().get("renditions").getAsJsonArray();
                            for (JsonElement rendition : renditions) {
                                if (rendition.getAsJsonObject().get("resolutionType") != null
                                        && rendition.getAsJsonObject().get("resolutionType").getAsString()
                                                .equals("ONE_MYRENAULT_SMALL")) {
                                    url = rendition.getAsJsonObject().get("url").getAsString();
                                    break;
                                }
                            }
                        }
                    }
                    if (url != null && !url.isEmpty()) {
                        imageURL = url;
                        break;
                    }
```

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/RenaultHandler.java
```diff
@@ -0,0 +1,147 @@
...
+ * See the NOTICE file(s) distributed with this work for additional
+ * information.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package org.openhab.binding.renault.internal;
```
lolodomo, line 13: A good practice (only a suggestion) is to put the thing handler(s) is a package internal.handler

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/renault/api/MyRenaultHttpSession.java
```diff
@@ -0,0 +1,227 @@
...
+
+    public void updateCarData() {
+        getCockpit();
+        getBatteryStatus();
+        getLocation();
+        getHvacStatus();
+    }
+
+    private void login() throws RenaultException, InterruptedException, ExecutionException, TimeoutException,
+            JsonParseException, ClassCastException, IllegalStateException, NullPointerException {
```
lolodomo, line 81: Please try avoiding throwing NullPointerException .
Rather test if null in your code.
Something like that:
```
            JsonObject responseJson = JsonParser.parseString(response.getContentAsString()).getAsJsonObject();
            String value = null;
            JsonObject sessionInfoJson = responseJson.getAsJsonObject("sessionInfo");
            if (sessionInfoJson != null) {
                value = sessionInfoJson.get("cookieValue").getAsString();
            }
            if (value != null) {
                cookieValue = value;
            } else {
                throw new RenaultException("Cookie value not found in JSON response");
            }
            logger.debug("Cookie: {}", cookieValue);
```
And do something similar for your other methods in this class.

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/renault/api/MyRenaultHttpSession.java
```diff
@@ -0,0 +1,227 @@
...
+    private final Logger logger = LoggerFactory.getLogger(MyRenaultHttpSession.class);
+
+    public MyRenaultHttpSession(RenaultConfiguration config, HttpClient httpClient) {
+        this.config = config;
+        this.httpClient = httpClient;
+        this.constants = new Constants(config.locale);
+    }
+
+    public void initSesssion() throws RenaultException, InterruptedException, ExecutionException, TimeoutException,
+            JsonParseException, ClassCastException, IllegalStateException, NullPointerException {
```
lolodomo, line 65: Look at my other comment about NullPointerException.

lolodomo (changes requested):
Here are my very final comments leading a clean compilation of the binding.

lolodomo:
@dougculnane : a last effort and this is finished.
After all these changes, please check that the binding is still working well ;)

lolodomo:
I am going to push my changes to your PR.

lolodomo:
@dougculnane : look at your Git repo, I submitted a PR with all my proposed changes. That is your turn to review and comment :)

lolodomo:
I will update my PR this evening to simplify even more exception handling.

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/RenaultHandler.java
dougculnane: @lolodomo Thanks for the tip but I am confused how this works?  You create a new reference to a nullable object then check the ref is not null.  How is this different to checking the original ref to the nullable object?  Happy to adopt this technique but I would like to understand it.

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/renault/api/Car.java
dougculnane: There is a some data we could persist in memory and do not to waste electricity collecting again. Account ID car image url....  However I copied the python code and run through the collection steps ever 10 mins and discard the collected Session and Car.  This is simple and robust but inefficient.  Very happy to improve it but not sure how to do this in an openHAB compatible way.

What is the best way to store the collected data?  Can I get the channel and see if there is an image?  Can I store the account ID etc in memory in the Session and not discard it each run?

Or we just get this finished and see if anyone uses it and what the feedback is before we optimise?

dougculnane:
@lolodomo I merged in your changes but now there is a check fail.  I guess the wrong author signed it off?  Not sure how to fix this.  

I added some questions to your comments above. I have this code running on my openhab instance for testing. It got the initial data so it works and I guess it will update too.

lolodomo:
> I merged in your changes but now there is a check fail. I guess the wrong author signed it off? Not sure how to fix this.

I made a stupid mistake in my email! ree.fr instead of free.fr
Maybe you can update it in your local branch using command "git commit --amend" and then do a force push using argument "--force"?

dougculnane:
@lolodomo  Looks like I managed to fix the commit message.

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/RenaultHandler.java
lolodomo: As pollingJob is a class member, the compiler assumes that it could be updated at any time by any other thread.
Using a local variable solves this problem.

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/renault/api/Car.java
lolodomo: You are already saving this data in memory in class member `imageURL`.
So what you have to do is to avoid recomputing it as soon as `imageURL` is not null.
This could be done in initSession by calling getVehicle() only if car.getImageUrl() is null ?

lolodomo (approved):
LGTM

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/renault/api/Car.java
lolodomo: I am wrong because you are instantiating a new `MyRenaultHttpSession` at each refresh cycle and so with an empty `Car `structure.
So in this case, this is more the thing handler that should keep a trace (boolean) if you already got the imageUrl and if this is the case you should request `MyRenaultHttpSession` to not request `getVehicle`

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/RenaultHandler.java
dougculnane: @lolodomo Merci! I wrote a unit test to test my understanding of the "=" char.  Turns out I got it wrong.  I thought it was a pointer to the memory of the object and adding a new var to the the old pointer was just a new ref to the memory location but your comment and my unit test show that it copies the value too!    ( I have only been programming Java for 25 years ! )  Now I understand this properly this so thanks.

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/renault/api/Car.java
dougculnane: OK.  I will add an hasBeenInitialized flag or something to the handler and push a new version that only http gets what it need to.  Hope to get that done this week.  New features at this stage are maybe not great but it will improve the implementation to make it more efficient.

dougculnane:
@lolodomo Let me do this properly and try to reduce a couple of HTTP gets if I can.  I will try to get this done and tested this week.  Thanks again.

dougculnane:
I have refactored this to make errors more transparent.  I am not sure if this is openHAB friendly?  I want the thing to stay online but say there is a com error with the location (which happens in some cars).  Or if you car gets up paired I want this to be more transparent.  I hope this is OK?

Screenshot_20211203_193820

dougculnane:
@lolodomo  I hope this is OK now?  Thanks again for your time.

lolodomo:
> I have refactored this to make errors more transparent. I am not sure if this is openHAB friendly? I want the thing to stay online but say there is a com error with the location (which happens in some cars). Or if you car gets un-paired I want this to be more transparent. I hope this is OK?

Honestly, this looks a little weird to me. Displaying a communication error but keeping the thing ONLINE.
If I correctly understand, you just want to ignore these communication errors because you already know that some of the cars will trigger these errors. Why not just keeping the thing status unchanged and add a DEBUG log that a user can display if he would like to investigate why a particular channel is not set ?
Additionally, your solution could even lead to several status changes (several communication errors) and the user will see in UI only one communication error.

dougculnane:
@lolodomo I know it is not great.  

The problem I am trying to solve is that if the car gets up-paired or some channels have errors there is no way of telling unless you look in the log files.  The the binding need to ignore errors and keep polling so it should stay ONLINE (as far as I understand or it will stop polling).  Is there a better way of reporting errors or should I create a new channel for a status message or something?  Thanks again.

dougculnane:
@lolodomo 
There is a good suggestion at the bottom of this tread that I can do.  Ignore not implemented channels:
https://community.openhab.org/t/betatest-renault-ze-services-binding/72226/115?u=doug_culnane

Is the status message a good idea or is there some other error viability idea?  Or are you getting board of this?

Thanks again again.

lolodomo:
For requests that are failing systematically (comm error) because not supported by a car, my opinion is that you should just log something and do nothing regarding the thing status. Avoiding next similar requests is of course a good idea. You could have  "disableXXX" booleans in your car structure that you will set to true in case a first request is failing.

For requests that are failing only in not normal situations but you know that the request is applicable for the car, your current solution could be acceptable. Keep status ONLINE if you are really sure that there is not general comm problem, else set it to OFFLINE.
Note that your scheduled task will continue to be called even if your thing status is OFFLINE.

We have to conclude quickly now, even if not perfect, to have your binding part of the final 3.2 distribution.

Let me know if you want to add a new final optimization (avoid recalling for ever requests that you know they will fail).

dougculnane:
@lolodomo Thanks for the advice.  I will try to push an update this evening.

dougculnane:
@lolodomo I added a not not Implemented error and disableXXX flags for the services. I keep the errors to the log files.  I recreate the car Object on init so that you can reset it...  Thanks again.

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/api/Car.java
```diff
@@ -0,0 +1,134 @@
...
+ */
+@NonNullByDefault
+public class Car {
+
+    private final Logger logger = LoggerFactory.getLogger(Car.class);
+
+    public boolean disableLocation = false;
+    public boolean disableBattery = false;
+    public boolean disableCockpit = false;
+    public boolean disableHvac = false;
```
lolodomo, line 38: Please use private members and add getters/setters
No need to set the member to false as it is the implicit default.

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/api/Car.java
```diff
@@ -0,0 +1,134 @@
...
+
+    public @Nullable Double batteryLevel;
+    public @Nullable Boolean hvacstatus;
+    public @Nullable Double odometer;
+    public @Nullable String imageURL;
+    public @Nullable Double gpsLatitude;
+    public @Nullable Double gpsLongitude;
```
lolodomo, line 45: Please use private members and add getters

lolodomo (changes requested):
Very final comment

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/api/Car.java
dougculnane: Done.

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/api/Car.java
dougculnane: Done

lolodomo (approved):
LGTM

dougculnane:
@lolodomo Getters and setter are done.  I have tested the binding on my openHAB installation.  Thanks again.

lolodomo:
@dougculnane : thank you for your new binding contribution.

dougculnane:
@lolodomo Merci beaucoup for your help and time getting this ready for release.

lolodomo:
@dougculnane : if not yet done, you could add your binding's logo to the openHAB website. See https://www.openhab.org/docs/developer/bindings/#add-your-binding-s-logo-to-the-openhab-website

dougculnane:
@lolodomo I have created a PR for the logo here:  https://github.com/openhab/openhab-docs/pull/1730

All the best, Doug

PR #11484, opened 2021-10-31
[openweathermap] Internationalization of discovery results

lolodomo:

### DIFF
--- a/bundles/org.openhab.binding.openweathermap/src/main/resources/OH-INF/i18n/openweathermap_fr.properties
+++ b/bundles/org.openhab.binding.openweathermap/src/main/resources/OH-INF/i18n/openweathermap_fr.properties
@@ -257,5 +257,5 @@ offline.conf-error-not-supported-onecall-number-of-hours = Le paramètre 'foreca
 offline.conf-error-not-supported-onecall-number-of-days = Le paramètre 'forecastDays' doit être compris entre 0 et 7.
 
 # discovery result
-discovery.openweathermap.weather-and-forecast.api.local.label = Météo locale et prévisions
-discovery.openweathermap.uvindex.api.local.label = Indice UV local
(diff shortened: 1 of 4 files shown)


bundles/org.openhab.binding.openweathermap/src/main/resources/OH-INF/i18n/openweathermap_fr.properties
```diff
@@ -257,5 +257,5 @@ offline.conf-error-not-supported-onecall-number-of-hours = Le paramètre 'foreca
 offline.conf-error-not-supported-onecall-number-of-days = Le paramètre 'forecastDays' doit être compris entre 0 et 7.
 
 # discovery result
-discovery.openweathermap.weather-and-forecast.api.local.label = Météo locale et prévisions
-discovery.openweathermap.uvindex.api.local.label = Indice UV local
+discovery.weather-and-forecast.local.label = Météo locale et prévisions
+discovery.air-pollution.local.label = Indice UV local
```
cweitkamp, line 261: Can you verify if this translation is correct? If not remove it from the French properties and let it be translated via Crowdin.

cweitkamp (changes requested):
Thanks. Looks much better and will probably work if someone does not choose id `api` for his bridge.

bundles/org.openhab.binding.openweathermap/src/main/resources/OH-INF/i18n/openweathermap_fr.properties
lolodomo: Translation is correct.

lolodomo:
> Looks much better and will probably work if someone does not choose id `api` for his bridge

Yes, that is the purpose and of course I tested it.

cweitkamp (approved):
LGTM. Thanks.

Issue #11214, opened 2021-09-06
[Homematic] Race condition causing items without channels to be created


Flole998:
Since quite some time there is an issue with items without channels being created: https://community.openhab.org/t/homematic-binding-channel-not-found-for-datapoint-errors-for-definitely-existing-channels/26209

What I observed: The thing type does correctly contain the channels, the actual thing doesn't. 

I did some investigation and discovered the following: If the thing type is not available yet when the thing is created then the thing will be missing the channels afterwards. Unfortunately I don’t yet know what to do in that case, a changeThingType if no channels found is a dirty hack but it ends in a MISSING_HANDLER status (but with the channels populated properly).

If someone has any idea on how to address this I would be very grateful for suggestions so this issue can finally be resolved.

Daniel-42:
Could this one be related? #11000 (Hue channels missing)

Flole998:
> Could this one be related? #11000 (Hue channels missing)

Most likely not

PR #11493, opened 2021-11-01
[Homematic] Fix "Channel not found for Datapoint"-Errors

Flole998:
Issue described and tested in https://community.openhab.org/t/homematic-binding-channel-not-found-for-datapoint-errors-for-definitely-existing-channels/26209/113 

Fixes: #11214
Signed-off-by: Flole <flole@flole.de>

### DIFF
--- a/bundles/org.openhab.binding.homematic/src/main/java/org/openhab/binding/homematic/internal/handler/HomematicThingHandler.java
+++ b/bundles/org.openhab.binding.homematic/src/main/java/org/openhab/binding/homematic/internal/handler/HomematicThingHandler.java
@@ -147,10 +147,52 @@ public class HomematicThingHandler extends BaseThingHandler {
         }
         updateConfiguration(config);
 
+        boolean channelsChanged = false;
+
         // update thing channel list for reconfigurable channels (relies on the new value of the
         // CHANNEL_FUNCTION datapoint fetched during configuration update)
         List<Channel> thingChannels = new ArrayList<>(getThing().getChannels());
+
+        if (thingChannels.isEmpty()) {
+            for (HmChannel channel : device.getChannels()) {
+                for (HmDatapoint dp : channel.getDatapoints()) {
+                    if (HomematicTypeGeneratorImpl.isIgnoredDatapoint(dp)
+                            || dp.getParamsetType() != HmParamsetType.VALUES) {
+                        continue;
+                    }
+                    ChannelUID channelUID = UidUtils.generateChannelUID(dp, getThing().getUID());
+                    if (containsChannel(thingChannels, channelUID)) {
+                        // Channel is already present
+                        continue;
+                    }
+
+                    ChannelTypeUID channelTypeUID = UidUtils.generateChannelTypeUID(dp);
+                    ChannelType channelType = channelTypeProvider.getInternalChannelType(channelTypeUID);
+                    if (channelType == null) {
+                        channelType = HomematicTypeGeneratorImpl.createChannelType(dp, channelTypeUID);
+                        channelTypeProvider.addChannelType(channelType);
+                    }
+
+                    Channel thingChannel = ChannelBuilder.create(channelUID, MetadataUtils.getItemType(dp))
+                            .withLabel(MetadataUtils.getLabel(dp))
+                            .withDescription(MetadataUtils.getDatapointDescription(dp)).withType(channelType.getUID())
+                            .build();
+                    thingChannels.add(thingChannel);
+
+                    logger.debug(
+                            "Updated value datapoints for channel {} of device '{}' (function {}), now has {} datapoints",
+                            channel, channel.getDevice().getAddress(), channel.getCurrentFunction(),
+                            channel.getDatapoints().size());
+                }
+            }
+            channelsChanged = true;
+        }
+
         if (updateDynamicChannelList(device, thingChannels)) {
+            channelsChanged = true;
+        }
+
+        if (channelsChanged) {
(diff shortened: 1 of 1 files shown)


Flole998:
@FStolte @gerrieg @mdicke2s Can anybody review please?

bundles/org.openhab.binding.homematic/src/main/java/org/openhab/binding/homematic/internal/handler/HomematicThingHandler.java
```diff
@@ -150,6 +150,42 @@ private void doInitializeInBackground() throws GatewayNotAvailableException, Hom
         // update thing channel list for reconfigurable channels (relies on the new value of the
         // CHANNEL_FUNCTION datapoint fetched during configuration update)
         List<Channel> thingChannels = new ArrayList<>(getThing().getChannels());
+
+        if (thingChannels.size() == 0) {
```
maniac103, line 154: Using `isEmpty()` seems more clear.

bundles/org.openhab.binding.homematic/src/main/java/org/openhab/binding/homematic/internal/handler/HomematicThingHandler.java
```diff
@@ -150,6 +150,42 @@ private void doInitializeInBackground() throws GatewayNotAvailableException, Hom
...
+                            "Updated value datapoints for channel {} of device '{}' (function {}), now has {} datapoints",
+                            channel, channel.getDevice().getAddress(), channel.getCurrentFunction(),
+                            channel.getDatapoints().size());
+                }
+            }
+            updateThing(editThing().withChannels(thingChannels).build());
+        }
+
         if (updateDynamicChannelList(device, thingChannels)) {
             updateThing(editThing().withChannels(thingChannels).build());
```
maniac103, line 190: Should collapse both `updateThing()` calls into one, e.g. by having a boolean variable `thingListChanged` which is set to `true` in your new block and updated here as well (`changed |= updateDynamicChannelList(...)`), and then `updateThing` is called when `changed` is `true`.

maniac103:
What still irks me about this is that this change is just a workaround for what should be a bug elsewhere. Did you try adding some breakpoints in `ThingFactory` or `BaseThingHandlerFactory` to check the state of the `ThingType` at thing creation time? 
AFAICT, the flow is
- inbox contains thing UID and thing type UID
- inbox calls `ThingFactory.createThing()`, passing thing UID, thing type UID, and bridge UID
- `ThingFactory.createThing()` tries to lookup a `ThingHandlerFactory` based on the thing type UID; if this fails, no thing is created
- `ThingFactory.craeteThing()` calls `createThing()` in the found factory (here, it's `HomematicThingHandlerFactory`)
- `HomematicThingHandlerFactory.createThing` (or rather the base class implementation in `BaseThingHandlerFactory`) looks up the `ThingType` by the thing type UID via `ThingTypeRegistry`, which in turn looks through all registered `ThingTypeProvider`s. In our case, `HomematicThingTypeProviderImpl` should match. If no `ThingType` can be found, no thing is created.
- `ThingFactory.createThing` is called, passing in the found `ThingType`, which in turn calls `ThingFactoryHelper.createChannels`
- `ThingFactoryHelper.createChannels` loops over `thingType.getChannelDefinitions()`, calling `ThingFactoryHelper.createChannel` for each one, which in turn calls `ThingFactoryHelper.createChannelBuilder`
- That one looks up a `ChannelType` from the channel type UID in the channel definition using a call to `ChannelTypeRegistry.getChannelType()`
- That one looks through all `ChannelTypeProvider`s (here: `HomematicChannelTypeProviderImpl` for a matching channel type

AFAICT, the *only* case where a thing can be created without channels is if the channel type lookup fails, so it would be great if you could verify this actually being the case. If it is, we need to check what's off in channel type registration. Unfortunately it's hard to follow git history due to the code being moved between repositories several times.

Flole998:
The reason this is necessary is, that (at least) when using text based config it could happen that things are created before the bridge is online or the thing type has been populated with the channels. In that case an empty thing type is used and the thing is missing the channels as when a thing type is updated the things that use it aren't automatically. So this will add the channels in that case.

maniac103:
So this essentially means that `createThing()` is called before `generate()` is called, right?

Flole998:
Yes iirc that was the issue. I also tried to re-create (basically reload) the thing if the channels are empty and that created the channels as well (but caused other issues).

If you ask me the current attempt is a wrong approach anyways as there's another Limit: If there are multiple versions of the same device with different channels it will cause issues. It would be better to request the channels for each thing and populate them on initialization.

bundles/org.openhab.binding.homematic/src/main/java/org/openhab/binding/homematic/internal/handler/HomematicThingHandler.java
Flole998: Fixed

bundles/org.openhab.binding.homematic/src/main/java/org/openhab/binding/homematic/internal/handler/HomematicThingHandler.java
Flole998: Fixed

Flole998:
@maniac103 I just fixed the 2 things you commented.

fwolter (approved):

fwolter:
The sign-off message should contain your real name. For the next contribution...

PR #11505, opened 2021-11-03
Configure default timeout of 15 minutes for all testable and lifecycle methods

wborn:
This helps to identify what tests cause builds to get stuck and it will more quickly end such builds.
When builds get stuck in tests, they would keep running for hours and then eventually when a timeout occurs, the job is killed without knowing why it got stuck.
Furhermore precious Jenkins executors will not keep being occupied by such jobs.

See: https://junit.org/junit5/docs/current/user-guide/#writing-tests-declarative-timeouts

It helps with identifying the root cause of issues like https://github.com/openhab/openhab-core/pull/2551


cweitkamp (approved):
LGTM. Thanks.

PR #11506, opened 2021-11-03
[openweathermap] Thing handler service registration

lolodomo:


lolodomo:
I close it until usage of ThingHandlerService is clarified.

Issue #11350, opened 2021-10-05
[mielecloud] ConfigFlowTest unstable


wborn:
This test failed in: https://ci.openhab.org/job/PR-openHAB-Addons/5743/

```
TEST org.openhab.binding.mielecloud.internal.config.ConfigFlowTest#configFlowHappyPathCreatesABridge() <<< ERROR: expected: <200> but was: <404>
org.opentest4j.AssertionFailedError: expected: <200> but was: <404>
        at org.junit.jupiter.api.AssertionUtils.fail(AssertionUtils.java:55)
        at org.junit.jupiter.api.AssertionUtils.failNotEqual(AssertionUtils.java:62)
        at org.junit.jupiter.api.AssertEquals.assertEquals(AssertEquals.java:150)
        at org.junit.jupiter.api.AssertEquals.assertEquals(AssertEquals.java:145)
        at org.junit.jupiter.api.Assertions.assertEquals(Assertions.java:510)
        at org.openhab.binding.mielecloud.internal.util.WebsiteCrawler.doGetRelative(WebsiteCrawler.java:44)
        at org.openhab.binding.mielecloud.internal.config.ConfigFlowTest.configureBridgeWithConfigFlow(ConfigFlowTest.java:76)
        at org.openhab.binding.mielecloud.internal.config.ConfigFlowTest.configFlowHappyPathCreatesABridge(ConfigFlowTest.java:111)
```

Can you have a look at it @BjoernLange?

BjoernLange:
Yes.

BjoernLange:
I had a look at this. One can make the tests fail locally by binding another server to port 8080 but that isn't the case in the CI build (different error message). I suspect that the parallel execution in the CI build causes failures when the binding tries to register it's website. That would match with the 404 error and the fact that the tests only fail sometimes.

I see multiple options here:
1. Remove the tests
2. Only run the tests in local / singlethreaded builds (maybe via some JUnit annotation?)
3. Don't run the integration tests in parallel in the CI build (is this possible?)

I would strongly prefer solution 2 or 3 but am not aware whether this is possible at all. Are there any experiences with that @wborn? Did other bindings have similar problems and if so, how was it fixed?

wborn:
In you use the HTTP server in tests, you should always run it on a random unoccupied port.

[...]

PR #11523, opened 2021-11-05
Always run OSGi HTTP service on random port in itests

wborn:
It is probably always a good idea to run the OSGi HTTP service on a random available port in itests.
So when this is always done it prevents future issues and removes a bit of duplication.

See also: https://github.com/openhab/openhab-addons/issues/11350


cweitkamp (approved):
LGTM. Thanks.

PR #11528, opened 2021-11-05
[twitter] Extend properties file with all default translations

lolodomo:


cweitkamp (approved):
LGTM. Thanks.

PR #11547, opened 2021-11-08
[openhabcloud] Added default i18n properties file

cweitkamp:

@wborn Thanks for adding support in your [i18n plugin](https://github.com/openhab/openhab-core/pull/2544#issuecomment-962652082) to translate other bundles too. Lets try if it is working like expected.


bundles/org.openhab.io.openhabcloud/src/main/resources/OH-INF/i18n/openhabcloud.properties
```diff
@@ -0,0 +1,10 @@
+# bundle config
```
cweitkamp, line 1: ftr: this comment was added by me manually.

wborn (approved):
Thanks! Now you can hopefully translate Cloud to Wolke everywhere, although I don't know if that is a good idea. 😉

PR #11548, opened 2021-11-08
[nikobus] added option to set rollershutter position

crnjan:
As also discussed [here](https://community.openhab.org/t/nikobus-v2/80241/324), an option to set rollershutter's position (in %) was added. Since Nikobus's rollerhutter module does not provide positioning information, required movement is calculated based on estimated position.


lolodomo (approved):
LGTN

Issue #11543, opened 2021-11-08
[hdpowerview] Shade secondary position calculated differently when getting vs setting


arroyoj:
## Expected Behavior
When setting the position of the secondary rail of a top-down/bottom-up Powerview shade, the secondary rail should move to the set position and the OpenHAB state of the secondary rail should consistently reflect that position.

## Current Behavior
After setting the position of the secondary rail using a PercentType command, the rail moves properly to the correct position, and the OpenHAB state is updated appropriately. However, after the add-on refreshes the position from the raw internal Powerview value, the OpenHAB state is incorrectly updated to the the "inverted" value within the 0-100 range (ie, 0 become 100, 70 becomes 30).

Here is an example from my logs after sending a secondary rail (with Item name `LivingRoom_Shade1_Top`) the command to move to a position value of 70:
```
2021-11-07 21:46:28.100 [INFO ] [openhab.event.ItemCommandEvent      ] - Item 'LivingRoom_Shade1_Top' received command 70
2021-11-07 21:46:28.101 [INFO ] [penhab.event.ItemStatePredictedEvent] - Item 'LivingRoom_Shade1_Top' predicted to become 70
2021-11-07 21:46:28.102 [INFO ] [openhab.event.ItemStateChangedEvent ] - Item 'LivingRoom_Shade1_Top' changed from 0 to 70
2021-11-07 21:47:17.739 [INFO ] [openhab.event.ItemStateChangedEvent ] - Item 'LivingRoom_Shade1_Top' changed from 70 to 30
```
The state is initially set to the correct value of 70, and the shade moves appropriately, but then after the add-on refreshes the state of all of the shades, the value is incorrectly set to 30.

## Possible Solution
First of all, thanks to the team for the HD Powerview add-on and especially @andrewfg for adding support for the dual action shades. Looking at the code, it appears that the set and get functions that convert between OpenHAB coordinates and Powerview coordinates are not reciprocal. 

Here is the code for the `setPosition2` function, in which setting an input PercentType value of 100 for the secondary rail will set the rail to the Powerview `MAX_SHADE` value and move the rail to the bottom (see line 214):
https://github.com/openhab/openhab-addons/blob/8dd4559443422ea9e034ea480a425154a7292cd5/bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/api/ShadePosition.java#L201-L219

[...]

PR #11552, opened 2021-11-10
[hdpowerview] Secondary shade position was inverted

andrewfg:
**Background**
This PR resolves issue #11543 

**Description**
The shade secondary position Getter method was using inverted logic compared to the respective Setter method.

**Solution**
The shade secondary position Getter and Setter methods are now compatible.


lolodomo (approved):
LGTM

Issue #11531, opened 2021-11-06
[ojelectronics] Binding stops working after server timeout


seime:
Problem: Bridge hits an unexpected timeout response from the server, then stops working.

OpenHab version: 3.1.0

Thing configuration:

```
Bridge ojelectronics:ojcloud:account "Micromatic Bridge" @ "Hele huset" [ 
  userName="xxxxx", 
  password="xxxx", 
  apiKey="xxxx", 
  customerId="15" 
  ] {
    Thing owd5 gulvtermostatBadNede "Bad nede" [ serialNumber="xxx" ]
  ....
```
Logs:
```
2021-11-05 18:56:16.386 [TRACE] [jelectronics.internal.OJCloudHandler] - OJElectronicsCloudHandler.handleRefreshDone(org.openhab.binding.ojelectronics.internal.models.groups.GroupContentResponseModel@f88730)
2021-11-05 19:00:27.083 [TRACE] [ics.internal.services.RefreshService] - refresh <html><head><title>500 - The request timed out.</title></head><body>  <font color ="#aa0000">         <h2>500 - The request timed out.</h2></font>  The web server failed to respond within the specified time.</body></html>
2021-11-05 19:00:27.084 [DEBUG] [ics.internal.services.RefreshService] - Error mapping Result to model
com.google.gson.JsonSyntaxException: java.lang.IllegalStateException: Expected BEGIN_OBJECT but was STRING at line 1 column 1 path $
        at com.google.gson.internal.bind.ReflectiveTypeAdapterFactory$Adapter.read(ReflectiveTypeAdapterFactory.java:226) ~[bundleFile:?]
        at com.google.gson.Gson.fromJson(Gson.java:932) ~[bundleFile:?]
        at com.google.gson.Gson.fromJson(Gson.java:897) ~[bundleFile:?]
        at com.google.gson.Gson.fromJson(Gson.java:846) ~[bundleFile:?]
        at com.google.gson.Gson.fromJson(Gson.java:817) ~[bundleFile:?]
        at org.openhab.binding.ojelectronics.internal.services.RefreshService.handleRefreshDone(RefreshService.java:143) [bundleFile:?]
        at org.openhab.binding.ojelectronics.internal.services.RefreshService$1.onComplete(RefreshService.java:125) [bundleFile:?]
        at org.eclipse.jetty.client.ResponseNotifier.notifyComplete(ResponseNotifier.java:218) [bundleFile:9.4.40.v20210413]
[...]

PR #11555, opened 2021-11-10
[ojelectronics] Fixing server timeout is stopping the extension

EvilPingu:
This is a bugfix for #11531 

Now the extention do not stop working if the server throw not an HTTP-Code 200.


lolodomo (approved):
LGTM

PR #11562, opened 2021-11-11
[MQTT.Homeassistant] make the mqtt.vacuum compliant

antroids:
[MQTT.Homeassistant] make the mqtt.vacuum compliant with the Specification

The mqtt.vacuum HA component has a Specification at https://www.home-assistant.io/integrations/vacuum.mqtt/
The goal of this PR is to make OpenHAB implementation compliant with the Specification, to be sure that any vacuum can be supported by Homeassistant integration out of the box.

I added support for legacy schema and all missing configuration attributes.
Unfortunately, time and usage status attributes are out of specification and they are specific for old valetudo versions only. So, I removed them from implementation. The new valetudo firmware should be supported by Homie binding. Any firmware specific mqtt messages can also be handled by generic mqtt binding.


fwolter (approved):

fwolter:
You could set yourself as the code owner in the `CODEOWNERS` file.

antroids:
> You could set yourself as the code owner in the `CODEOWNERS` file.

Do I need to create a PR for that?

fwolter:
Yes.

PR #11563, opened 2021-11-11
[miio] Initial i18n properties file for Xiaomi miio binding

marcelrv:


lolodomo:
@marcelrv : ideally you should use @text rather than a hardcoded English text when calling updateStatus in your thing handler.

marcelrv:
@lolodomo I see this indeed as a very first step in making the binding i18n compatible.
e.g. the 1250+ dynamically created channels defined infor the things also need a solution. (I did not yet see easy solution for those...)

lolodomo:
It looks like all channel types are not there in the properties file. An example: MijiaVacuumV2_fan-level

marcelrv:
> It looks like all channel types are not there in the properties file. An example: MijiaVacuumV2_fan-level

Yes, because they are about to be removed https://github.com/openhab/openhab-addons/pull/11568

lolodomo (approved):
LGTM, hoping there are no forgotten channel types.

PR #11574, opened 2021-11-12
[transform][map] Added default i18n properties file

cweitkamp:

For Transformation Services we do not have bundle related information to be translated. They are not configured as `@ConfigurableService`. The name is taken from the feature file.

grafik


lolodomo (approved):
LGTM

PR #11586, opened 2021-11-14
[amplipi] Add discovery and PA support

kaikreuzer:
This PR adds the possibility to find the AmpliPi on the local network through mDNS.

Additionally, it now acts as an audio sink, so that audio streams can be sent to it, which are then announced as a PA on all available zones (either with current volume or a specified volume level). This is a great feature for doing announcements in every room, such as doorbell, telephone call notifications etc.

### DIFF
--- a/bundles/org.openhab.binding.amplipi/src/main/java/org/openhab/binding/amplipi/internal/audio/PAAudioSink.java
+++ b/bundles/org.openhab.binding.amplipi/src/main/java/org/openhab/binding/amplipi/internal/audio/PAAudioSink.java
@@ -1,0 +1,153 @@
+/**
+ * Copyright (c) 2010-2021 Contributors to the openHAB project
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package org.openhab.binding.amplipi.internal.audio;
+
+import java.io.IOException;
+import java.util.Locale;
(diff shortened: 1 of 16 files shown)


kaikreuzer:
@linknum23 FYI!

bundles/org.openhab.binding.amplipi/src/main/java/org/openhab/binding/amplipi/internal/audio/PAAudioSink.java
```diff
@@ -0,0 +1,154 @@
...
+    private static final Set<AudioFormat> SUPPORTED_AUDIO_FORMATS = Set.of(AudioFormat.MP3, AudioFormat.WAV);
+    private static final Set<Class<? extends AudioStream>> SUPPORTED_AUDIO_STREAMS = Set
+            .of(FixedLengthAudioStream.class, URLAudioStream.class);
+
+    private @Nullable AmpliPiHandler handler;
+
+    private @Nullable PercentType volume;
+
+    public PAAudioSink() {
+    }
```
wborn, line 55: Isn't it possible to use the default constructor here and remove this default constructor?

wborn (commented):
Thanks for adding this nice enhancement!
It looked like a lot of code but most of it is generated. :wink: 
I only have one small remark and also a question.

bundles/org.openhab.binding.amplipi/src/main/java/org/openhab/binding/amplipi/internal/audio/PAAudioSink.java
kaikreuzer: Indeed, not sure why I had it there...

kaikreuzer:
Thanks @wborn, your remarks are addressed now!

wborn (approved):
Thank you!

PR #11591, opened 2021-11-15
[monopriceaudio] Clarify supported models and documentation

mlobstein:
In preparation to release a beta version of this binding to the marketplace which will support additional models, the mainline version is updated via this PR to better identify which models are supported. No functional changes are present in this PR, only documentation and name/label changes are made.


lolodomo (approved):
LGTM

PR #11594, opened 2021-11-16
[XMLTV] Preparing for Crowdin and code refining.

clinique:


bundles/org.openhab.binding.xmltv/src/main/resources/OH-INF/binding/binding.xml
```diff
@@ -4,6 +4,5 @@
 	xsi:schemaLocation="https://openhab.org/schemas/binding/v1.0.0 https://openhab.org/schemas/binding-1.0.0.xsd">
 
 	<name>XmlTV Binding</name>
-	<description>This is the binding for reading and parsing XmlTV files</description>
```
lolodomo, line 7: Why do you suppress the description ?

bundles/org.openhab.binding.xmltv/src/main/resources/OH-INF/thing/thing-types.xml
```diff
@@ -6,21 +6,17 @@
 
 	<bridge-type id="xmltvfile">
 		<label>XmlTVFile</label>
-		<description>This is the interface to a XmlTV file</description>
```
lolodomo, line 9: Why do you suppress all the descriptions ?

bundles/org.openhab.binding.xmltv/src/main/java/org/openhab/binding/xmltv/internal/handler/ChannelHandler.java
```diff
@@ -233,17 +226,21 @@ private void updateChannel(ChannelUID channelUID) {
...
-        return new QuantityType<>(secondsElapsed, Units.SECOND);
+        long elapsed = Duration.between(from, to).toSeconds();
+        return new QuantityType<>(elapsed, Units.SECOND);
     }
 
-    private @Nullable RawType downloadIcon(List<Icon> icons) {
+    private State downloadIcon(List<Icon> icons) {
         if (!icons.isEmpty()) {
             String url = icons.get(0).getSrc();
             return HttpUtil.downloadImage(url);
```
lolodomo, line 242: Take care, HttpUtil.downloadImage can return null

lolodomo (changes requested):

bundles/org.openhab.binding.xmltv/src/main/resources/OH-INF/binding/binding.xml
clinique: Descriptions are superfluous and redundant as soon as they are present in the properties file.

bundles/org.openhab.binding.xmltv/src/main/java/org/openhab/binding/xmltv/internal/XmlTVHandlerFactory.java
```diff
@@ -110,6 +120,8 @@ private synchronized void registerDeviceDiscoveryService(XmlTVHandler bridgeHand
 
     private synchronized void unregisterDeviceDiscoveryService(Thing thing) {
```
lolodomo, line 121: Rather use ThingHanflerService to simplify all this.

lolodomo (changes requested):

bundles/org.openhab.binding.xmltv/src/main/java/org/openhab/binding/xmltv/internal/XmlTVHandlerFactory.java
clinique: Done

bundles/org.openhab.binding.xmltv/src/main/resources/OH-INF/binding/binding.xml
lolodomo: @cweitkamp @wborn : what is your opinion about removing all description tags from XML files?

wborn: I don't have any plans to do the same for the add-ons I maintain. It takes effort and if the descriptions are good, they provide some context to things/channels that usually only have very short labels. So IMHO the descriptions help with understanding how add-ons are modeled.

lolodomo: That is also my opinion.

clinique: Don't get me wrong : removing description does not impact their presentation as soon as they are present in the properties file : 
image

I was surprised by this when I discovered it and thought it would avoid having two places to maintain the same string.

lolodomo: In this case, the description in the XML is more the original vision of the author and it is easy for any code maintainer to find it in the main place with everything else from the thing, that is the XML file.

clinique: Ok, I understand, but should I also reintroduce descriptions on things, channels ... ?

lolodomo: Yes ;)

clinique: Here you are.

lolodomo (approved):
LGTM

lolodomo:
Thank you

clinique:
You're welcome !

PR #11603, opened 2021-11-18
[miele] Localization of state, program and phase

### LINKED ISSUE #11602: [miele] Localization of states, programs and phases
As a continuation of:
- #11341
- #11377

I would like to support localization of text channels for states, programs and phases. This sounds simple, but some background information is needed:
- The openHAB language is configurable.
- The XGW 3000 gateway language is configurable (English, German, French or Dutch).
- Translations for the XGW 3000 are partial, even English strings are lacking. French is the least supported language mentioned.
- Translations for the openHAB binding will also be partial.
- The JSON-RPC protocol provides a LocalizedValue and an enum with all values. Those values are in English, no matter which gateway language is configured, but currently they override LocalizedValue in the binding.
- Multicast is supported, but only numeric values are available (no texts).
- The user might prefer a different language for the Miele texts than the global openHAB language configured.

So several factors needs to be taken into consideration, and the binding needs to come up with the best possible strings.

### LocalizedValue

German example:
```
                {
                    "Name": "state",
                    "Value": 2,
                    "Polling": 0,
                    "Metadata": {
                        "Filter": "true",
                        "description": "Property name for the appliance's current state.",
                        "LocalizedID": "Gerätestatus",
                        "enum": {
                            "End": "7",
                            "Abort": "9",
                            "Service": "12",
                            "Off": "1",
                            "Stand-By": "2",
                            "Programmed": "3",
                            "Paused": "6",
                            "Unknown": "0",
                            "Running": "5",
                            "Failure": "8",
                            "Locked": "145",
                            "Not Connected": "255",
                            "Waiting to Start": "4"
                        },
                        "LocalizedValue": "Bereit",
                        "access": "RE"
                    }
                },

```

Until now the binding would provide the text "Stand-By" for the state channel, even though this text is actually available in German as "Bereit".

### Missing gateway texts

Some texts are missing on the gateway. In this case the numeric value is still available. This is encoded into texts like this: "program.<value>" or "phase.<value>" for program and phase respectively - for example "program.1".

Some specific texts don't follow this rule, but this seems more like a bug. For washing machine, these are the texts for program 4 (Delicates):
program 4
- fr: réservé (0x0004)
- nl: gereserveerd (0x0004)

In this case it would be preferable to fallback to binding default language when French or Dutch text is not translated - than to fallback to gateway text.

### Approach and fallback strategy

I haven't found a way to check for existence of a text in a specific language. Comparing texts for English and some other language is not bullet-proof as they could actually be the same, which would mean there is no need to fallback to gateway text (which might be broken or missing). An example could be "QuickPowerWash".

To avoid having to fallback to gateway language, one strategy would be to supply all (or as many as possible) of the supported texts for all languages supported by the gateway. At this point this means:
- States (cross-appliance)
- Washing machine: programs and phases.
- Tumble dryer: programs and phases.
- Dishwasher: programs and phases.
- Oven: phases.

For:
- English
- German
- Dutch
- French

The reason for these specific appliance constraints is the lack of additional appliances from my side. Any help providing additional texts would be much appreciated. They can be harvested from the gateway as well as from the cloud binding. It's a time consuming process though.

jlaur:
Fixes #11602

This pull request provides localization of state, program and phase texts. Please see linked issue for details.

Continuation of:
- #11341 added support for built-in English texts where no texts are available from the XGW 3000 gateway (i.e. multicast updates and some missing texts from specific appliances and/or languages). This pull request extends this by adding full localization support.
- #11423 added initial support for localization.

### Primary enhancement

The XGW 3000 gateway supports English, German, French and Dutch, but with many limitations:
- Most dishwasher texts are missing (all languages).
- A lot of French texts are missing.
- Some texts are provided, but actually of no use - for example "réservé (0x0004)" for washing machine program 4 (delicates) in French.

This pull request includes as many texts in these languages as possible (and additionally Danish texts) in order to avoid fallback to English in situations where gateway could have provided a localized text.

Texts provided by the binding are delivered in the configured openHAB system language, but can also be overridden in the bridge configuration.

The provided texts are not translated, but taken from one of these sources:
- Response from the gateway when configured to target language.
- Corresponding text from the Miele Cloud binding when configured to target language.
- Text on appliance display when appliance is configured to target language.

In order to get a specific text, the appliance needs to be in the specific state or phase, so harvesting all texts is time consuming, and some texts are still lacking.

### Related other changes
- Bridge status messages are now internationalized and more specific.
- **LocalizedValue** from gateway takes precedence over enum texts. Enum texts are always in English even when gateway is configured to some other language, so this resulted in English state texts.
- Magic temperature value -32760 °C meaning "cold" is now interpreted as 10 °C.
- Some missing strings for channel labels and descriptions have been added.

### Additional changes
- **Configuration reload:** Previously, after changing bridge configuration, bridge thing would change status to UNKNOWN and never recover from that. Probably it was never noticed because bridge configuration is quite static. However, after adding language configuration, this problem appeared quite quickly and consistently. This has been fixed.
- **Code duplication refactoring:** Removed some duplicated code by extracting `DeviceMetaData` to separate file and implemented `getMieleEnum` here (previously implemented in each channel selector).


jlaur:
@kgoderis, @kaikreuzer - FYI, this PR is ready for review. It got kind of big, but finishes the work previously done in #11341 and #11423.

It would be nice for consideration in 3.2 as IMHO it more or less makes this binding feature complete. At least everything supported by the Miele@home cloud binding is now also supported by this binding, and all features I could find in the API's are supported. This concludes all the fixes and enhancements I intended to do for the Miele binding (and some more) - at least for now. :-)

kaikreuzer (approved):
Wow, what a great work - thanks @jlaur for polishing this binding in all aspects!

PR #11607, opened 2021-11-19
[SNCF] A binding to get French railways arrivals and departures

clinique:


lolodomo:
I will review your PR.

bundles/org.openhab.binding.sncf/README.md
```diff
@@ -0,0 +1,70 @@
+# SNCF Binding
+
+The SNCF binding provides real-time data for each train, bus, tramway... station in France.
+This is based on live API provided by DIGITALSNCF.
+
+Get you API key at https://www.digital.sncf.com/startup/api/token-developpeur
+
+Note : SNCF Api is based on the open [API Navitia](https://doc.navitia.io/#getting-started). This binding uses a very small subset of it, restricted to its primary purpose.
```
lolodomo, line 8: Please cut to have one line per sentence.

bundles/org.openhab.binding.sncf/README.md
```diff
@@ -0,0 +1,70 @@
+# SNCF Binding
+
+The SNCF binding provides real-time data for each train, bus, tramway... station in France.
+This is based on live API provided by DIGITALSNCF.
+
+Get you API key at https://www.digital.sncf.com/startup/api/token-developpeur
```
lolodomo, line 6: Plase add a link, for example 
`Get you API key on [DIGITALSNCF web site](https://www.digital.sncf.com/startup/api/token-developpeur)`

bundles/org.openhab.binding.sncf/README.md
```diff
@@ -0,0 +1,70 @@
...
+
+Note : SNCF Api is based on the open [API Navitia](https://doc.navitia.io/#getting-started). This binding uses a very small subset of it, restricted to its primary purpose.
+
+## Supported Things
+
+Bridge: The binding supports a bridge to connect to the [DIGITALSNCF service](https://www.digital.sncf.com/startup/api/token-developpeur). A bridge uses the thing ID "api".
+
+Station: Represents a given bus, train station.
+
+Of course, you can add multiple as many stations as needed.
```
lolodomo, line 16: I believe you can remove "multiple".

bundles/org.openhab.binding.sncf/README.md
```diff
@@ -0,0 +1,70 @@
...
+
+
+## Discovery
+
+This binding takes care of auto discovery. This method is strongly recommended as its the only way to get proper station ID depending upon transportation type.
```
lolodomo, line 21: as it is ...

bundles/org.openhab.binding.sncf/README.md
```diff
@@ -0,0 +1,70 @@
...
+
+To enable auto-discovery, your location system setting must be defined. Once done, at first launch discovery will search every station in a radius of 2500 m around the system, extending it by step of 500 m until it finds a first set of results.
```
lolodomo, line 23: I like this idea.
Please one line per sentence.
I think a comma is missing just after "launch".
Looking at the code, my feeling is that the initial radius is 2000 m (1500+500). To be clarified.

bundles/org.openhab.binding.sncf/README.md
```diff
@@ -0,0 +1,70 @@
...
+|-----------------------|-----------|--------------------------------------------------|
+| direction             | String    | The direction of the route                       |
+| code                  | String    | Code name of the line                            |
+| commercialMode        | String    | Commercial name of the line (RER, Transilien...) |
+| name                  | String    | Name of the line                                 |
+| network               | String    | Name of the network ruling the line              |
+| timestamp             | DateTime  | Timestamp of the event (departure, arrival)      |
+
+## Full Example
+
```
lolodomo, line 63: A full example of items would be appreciated.

bundles/org.openhab.binding.sncf/src/main/java/org/openhab/binding/sncf/internal/SncfBindingConstants.java
```diff
@@ -0,0 +1,55 @@
...
+    public static final String COMMERCIAL_MODE = "commercialMode";
+    public static final String NAME = "name";
+    public static final String NETWORK = "network";
+    public static final String TIMESTAMP = "timestamp";
+
+    // List of Thing Type UIDs
+    public static final ThingTypeUID APIBRIDGE_THING_TYPE = new ThingTypeUID(BINDING_ID, "api");
+    public static final ThingTypeUID STATION_THING_TYPE = new ThingTypeUID(BINDING_ID, "station");
+
+    // List of all adressable things in OH = SUPPORTED_DEVICE_THING_TYPES_UIDS + the virtual bridge
```
lolodomo, line 53: The last part of the comment looks not exactly appropriate to me.

bundles/org.openhab.binding.sncf/src/main/java/org/openhab/binding/sncf/internal/discovery/SncfDiscoveryService.java
```diff
@@ -0,0 +1,111 @@
...
+/**
+ * The {@link SncfDiscoveryService} searches for available
+ * station discoverable through API
+ *
+ * @author Gaël L'hopital - Initial contribution
+ */
+@Component(service = ThingHandlerService.class)
+@NonNullByDefault
+public class SncfDiscoveryService extends AbstractDiscoveryService implements ThingHandlerService {
+    private final Logger logger = LoggerFactory.getLogger(SncfDiscoveryService.class);
```
lolodomo, line 45: Move after the constant.

bundles/org.openhab.binding.sncf/src/main/java/org/openhab/binding/sncf/internal/discovery/SncfDiscoveryService.java
```diff
@@ -0,0 +1,111 @@
...
+    }
+
+    @Override
+    public void startScan() {
+        SncfBridgeHandler handler = bridgeHandler;
+        LocationProvider provider = locationProvider;
+        if (provider != null && handler != null) {
+            PointType location = provider.getLocation();
+            if (location != null) {
+                ThingUID bridgeId = handler.getThing().getUID();
```
lolodomo, line 71: Better to name the variable `bridgeUID`

bundles/org.openhab.binding.sncf/src/main/java/org/openhab/binding/sncf/internal/discovery/SncfDiscoveryService.java
```diff
@@ -0,0 +1,111 @@
...
+                        startScan();
+                    }
+                } catch (SncfException e) {
+                    logger.warn("Error calling Sncf Api : {}", e.getMessage());
+                }
+            } else {
+                logger.info("Please set a system location to enable station discovery");
+            }
+        }
+        stopScan();
```
lolodomo, line 96: I am a little surprise to see `stopScan` called in `startScan`

bundles/org.openhab.binding.sncf/src/main/java/org/openhab/binding/sncf/internal/dto/Coord.java
```diff
@@ -0,0 +1,22 @@
...
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package org.openhab.binding.sncf.internal.dto;
+
+/**
+ *
```
lolodomo, line 16: Please add a short description

bundles/org.openhab.binding.sncf/src/main/java/org/openhab/binding/sncf/internal/dto/Route.java
```diff
@@ -0,0 +1,21 @@
...
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package org.openhab.binding.sncf.internal.dto;
+
+/**
+ *
```
lolodomo, line 16: Please add a short description

bundles/org.openhab.binding.sncf/src/main/java/org/openhab/binding/sncf/internal/dto/StopArea.java
```diff
@@ -0,0 +1,21 @@
...
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package org.openhab.binding.sncf.internal.dto;
+
+/**
+ *
```
lolodomo, line 16: Please add a short description

bundles/org.openhab.binding.sncf/src/main/java/org/openhab/binding/sncf/internal/dto/StopDateTime.java
```diff
@@ -0,0 +1,22 @@
...
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package org.openhab.binding.sncf.internal.dto;
+
+/**
+ *
```
lolodomo, line 16: Please add a short description

bundles/org.openhab.binding.sncf/src/main/java/org/openhab/binding/sncf/internal/dto/StopPoint.java
```diff
@@ -0,0 +1,22 @@
...
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package org.openhab.binding.sncf.internal.dto;
+
+/**
+ *
```
lolodomo, line 16: Please add a short description

bundles/org.openhab.binding.sncf/src/main/java/org/openhab/binding/sncf/internal/dto/VJDisplayInformation.java
```diff
@@ -0,0 +1,24 @@
...
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package org.openhab.binding.sncf.internal.dto;
+
+/**
+ *
```
lolodomo, line 16: Please add a short description

lolodomo (changes requested):
Review part 1 of 3

bundles/org.openhab.binding.sncf/README.md
clinique: It's really bothering to do, but I'll do it... :)

bundles/org.openhab.binding.sncf/src/main/java/org/openhab/binding/sncf/internal/handler/SncfBridgeHandler.java
```diff
@@ -0,0 +1,177 @@
...
+        this.locationProvider = locationProvider;
+        this.httpClient = httpClient;
+        this.gson = gson;
+    }
+
+    @Override
+    public void initialize() {
+        logger.debug("Initializing SNCF API bridge handler.");
+        apiId = (String) getConfig().get("apiID");
+        if (apiId != null && apiId.length() != 0) {
```
lolodomo, line 87: `... && !apiId.isBlank()`

bundles/org.openhab.binding.sncf/src/main/java/org/openhab/binding/sncf/internal/handler/SncfBridgeHandler.java
```diff
@@ -0,0 +1,177 @@
...
+    @Override
+    public void handleCommand(ChannelUID channelUID, Command command) {
+        logger.debug("SNCF API Bridge is read-only and does not handle commands");
+    }
+
+    private <T extends SncfAnswer> T getResponseFromCache(String url, Class<T> objectClass) throws SncfException {
+        String answer = cache.get(url);
+        if (answer == null) {
+            final String newAnswer = getResponse(url);
+            cache.put(url, () -> newAnswer);
```
lolodomo, line 103: I am not sure to understand. You don't provide a real supplier that will get the value from the URL ?

bundles/org.openhab.binding.sncf/src/main/java/org/openhab/binding/sncf/internal/handler/SncfBridgeHandler.java
```diff
@@ -0,0 +1,177 @@
...
+            answer = newAnswer;
+        }
+        @Nullable
+        T response = gson.fromJson(answer, objectClass);
```
lolodomo, line 107: Are you sure that answer cannot be null at this place ? 
Maybe you should catch the JSON exception.

bundles/org.openhab.binding.sncf/src/main/java/org/openhab/binding/sncf/internal/handler/SncfBridgeHandler.java
```diff
@@ -0,0 +1,177 @@
...
+            String content = contentResponse.getContentAsString();
+            logger.debug("SNCF Api response: status = {}, content = '{}'", httpStatus, content);
+            switch (httpStatus) {
+                case OK_200:
+                    return content;
+                case BAD_REQUEST_400:
+                case UNAUTHORIZED_401:
+                case NOT_FOUND_404:
+                    logger.debug("SNCF Api server responded with status code {}: {}", httpStatus, content);
+                    throw new SncfException(content);
```
lolodomo, line 132: As the code is the same as the default, the code for these 3 cases can be removed.

bundles/org.openhab.binding.sncf/src/main/java/org/openhab/binding/sncf/internal/handler/SncfBridgeHandler.java
```diff
@@ -0,0 +1,177 @@
...
+    public @Nullable List<PlaceNearby> discoverNearby(PointType location, int distance) throws SncfException {
+        String URL = String.format(Locale.US,
+                "%s/coord/%.5f;%.5f/places_nearby?distance=%d&type[]=stop_point&count=100", SERVICE_URL,
+                location.getLongitude().floatValue(), location.getLatitude().floatValue(), distance);
+        PlacesNearby places = getResponseFromCache(URL, PlacesNearby.class);
+        return places.placesNearby;
+    }
+
+    public Optional<StopPoint> stopPointDetail(String stopPointId) throws SncfException {
+        String URL = String.format(Locale.US, "%s/stop_points/%s", SERVICE_URL, stopPointId);
```
lolodomo, line 156: Is it necessary to pass a locale in this case ?

bundles/org.openhab.binding.sncf/src/main/java/org/openhab/binding/sncf/internal/handler/SncfBridgeHandler.java
```diff
@@ -0,0 +1,177 @@
...
+        List<StopPoint> points = getResponseFromCache(URL, StopPoints.class).stopPoints;
+        return points != null ? Optional.ofNullable(points.get(0)) : Optional.empty();
+    }
+
+    public Optional<Passage> getNextPassage(String stopPointId, String expected) throws SncfException {
+        String URL = String.format(Locale.US, "%s/stop_points/%s/%s?disable_geojson=true&count=1", SERVICE_URL,
+                stopPointId, expected);
```
lolodomo, line 163: Is it necessary to pass a locale in this case ?

bundles/org.openhab.binding.sncf/src/main/java/org/openhab/binding/sncf/internal/handler/SncfBridgeHandler.java
```diff
@@ -0,0 +1,177 @@
...
+        List<Passage> passages = getResponseFromCache(URL, Passages.class).passages;
+        return passages != null ? !passages.isEmpty() ? Optional.ofNullable(passages.get(0)) : Optional.empty()
+                : Optional.empty();
```
lolodomo, line 166: return passages != null && !passages.isEmpty() ? Optional.ofNullable(passages.get(0)) : Optional.empty();

bundles/org.openhab.binding.sncf/src/main/java/org/openhab/binding/sncf/internal/handler/SncfBridgeHandler.java
```diff
@@ -0,0 +1,177 @@
...
+                "%s/coord/%.5f;%.5f/places_nearby?distance=%d&type[]=stop_point&count=100", SERVICE_URL,
+                location.getLongitude().floatValue(), location.getLatitude().floatValue(), distance);
+        PlacesNearby places = getResponseFromCache(URL, PlacesNearby.class);
+        return places.placesNearby;
+    }
+
+    public Optional<StopPoint> stopPointDetail(String stopPointId) throws SncfException {
+        String URL = String.format(Locale.US, "%s/stop_points/%s", SERVICE_URL, stopPointId);
+        List<StopPoint> points = getResponseFromCache(URL, StopPoints.class).stopPoints;
+        return points != null ? Optional.ofNullable(points.get(0)) : Optional.empty();
```
lolodomo, line 158: I believe you have to checks that `points` is not empty.

return points != null && !points.isEmpty() ? Optional.ofNullable(points.get(0)) : Optional.empty();

bundles/org.openhab.binding.sncf/src/main/java/org/openhab/binding/sncf/internal/handler/SncfBridgeHandler.java
```diff
@@ -0,0 +1,177 @@
...
+                    logger.debug("SNCF Api server responded with status code {}: {}", httpStatus, content);
+                    throw new SncfException(content);
+            }
+        } catch (TimeoutException | ExecutionException e) {
+            logger.debug("Exception occurred during execution: {}", e.getLocalizedMessage(), e);
+            throw new SncfException(e.getLocalizedMessage(), e.getCause());
+        } catch (InterruptedException e) {
+            logger.debug("Execution interrupted: {}", e.getLocalizedMessage(), e);
+            Thread.currentThread().interrupt();
+            throw new SncfException(e.getLocalizedMessage(), e.getCause());
```
lolodomo, line 143: I am not fully convinced by using `e.getLocalizedMessage()`, in particular for logger.
What is expected is to have log entries in English. Only messages displayed in UI should be translated.

bundles/org.openhab.binding.sncf/src/main/resources/OH-INF/config/config.xml
```diff
@@ -0,0 +1,16 @@
...
+<config-description:config-descriptions
+	xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+	xmlns:config-description="https://openhab.org/schemas/config-description/v1.0.0"
+	xsi:schemaLocation="https://openhab.org/schemas/config-description/v1.0.0
+		https://openhab.org/schemas/config-description-1.0.0.xsd">
+
+	<config-description uri="thing-type:sncf:api">
+		<parameter name="apiID" type="text">
+			<label>API ID</label>
+			<required>true</required>
```
lolodomo, line 11: This tag is now deprecated. Instead please add a "required" attribute to the parameter tag.

bundles/org.openhab.binding.sncf/src/main/resources/OH-INF/i18n/sncf.properties
```diff
@@ -0,0 +1,29 @@
...
+
+config.thing-type.sncf.api.apiID.label = API ID
+config.thing-type.sncf.api.apiID.description = Your Sncf API ID
+
+thing-type.sncf.api.label = SNCF API
+thing-type.sncf.api.description = This bridge represents the gateway to Sncf API.
+
+thing-type.sncf.station.label = Station
+thing-type.sncf.station.description = This represents a station hosting some transportation mode.
+
```
lolodomo, line 13: Config stopPointId is missing.

bundles/org.openhab.binding.sncf/src/main/resources/OH-INF/thing/station.xml
```diff
@@ -0,0 +1,82 @@
...
+	<channel-group-type id="passage">
+		<label>Other</label>
+		<channels>
+			<channel id="direction" typeId="direction"/>
+			<channel id="code" typeId="code"/>
+			<channel id="commercialMode" typeId="commercialMode"/>
+			<channel id="name" typeId="name"/>
+			<channel id="network" typeId="network"/>
+			<channel id="timestamp" typeId="timestamp">
+				<label>Horodatage</label>
```
lolodomo, line 41: To be removed

bundles/org.openhab.binding.sncf/src/main/resources/OH-INF/i18n/sncf.properties
```diff
@@ -0,0 +1,29 @@
...
+channel-type.sncf.code.description = Code of the line.
+channel-type.sncf.name.label = Name
+channel-type.sncf.name.description = Name of the line.
+channel-type.sncf.network.label = Network
+channel-type.sncf.network.description = Name of the transportation network.
+channel-type.sncf.commercialMode.label = Commercial Mode
+channel-type.sncf.commercialMode.description = Commercial modes are close from physical modes, but not normalized and can refer to a brand, something that can be specific to a network, and known to the traveler.
+channel-type.sncf.timestamp.label = Timestamp
+channel-type.sncf.timestamp.description = Timestamp of the future event.
+
```
lolodomo, line 26: Group labels are missing.

The best is that you use @wborn tool to generate this file.

lolodomo (changes requested):
Review part 2 of 3

bundles/org.openhab.binding.sncf/src/main/resources/OH-INF/i18n/sncf.properties
wborn: Yes it's a handy tool! :slightly_smiling_face:

lolodomo:
Remains for me the class `StationHandler` to review.

bundles/org.openhab.binding.sncf/src/main/java/org/openhab/binding/sncf/internal/handler/SncfBridgeHandler.java
clinique: ok, switched to ``e.getMessage()``

bundles/org.openhab.binding.sncf/src/main/java/org/openhab/binding/sncf/internal/handler/SncfBridgeHandler.java
clinique: You're correct, not needed.

bundles/org.openhab.binding.sncf/src/main/java/org/openhab/binding/sncf/internal/handler/SncfBridgeHandler.java
clinique: You're correct, not needed.

bundles/org.openhab.binding.sncf/src/main/java/org/openhab/binding/sncf/internal/handler/SncfBridgeHandler.java
clinique: Done.

bundles/org.openhab.binding.sncf/src/main/java/org/openhab/binding/sncf/internal/handler/SncfBridgeHandler.java
clinique: newAnswer is the supplier.

lolodomo: But this is only a string !

bundles/org.openhab.binding.sncf/src/main/java/org/openhab/binding/sncf/internal/handler/SncfBridgeHandler.java
clinique: Changed.

bundles/org.openhab.binding.sncf/src/main/java/org/openhab/binding/sncf/internal/handler/SncfBridgeHandler.java
clinique: Done

bundles/org.openhab.binding.sncf/src/main/java/org/openhab/binding/sncf/internal/handler/SncfBridgeHandler.java
clinique: Very valid.

bundles/org.openhab.binding.sncf/src/main/java/org/openhab/binding/sncf/internal/handler/SncfBridgeHandler.java
clinique: So true.

bundles/org.openhab.binding.sncf/src/main/java/org/openhab/binding/sncf/internal/handler/SncfBridgeHandler.java
clinique: ...sure, but it's a valid provider, and thus it works. But never mind, I did not like this and changed it.

lolodomo: Ok

bundles/org.openhab.binding.sncf/src/main/java/org/openhab/binding/sncf/internal/discovery/SncfDiscoveryService.java
```diff
@@ -0,0 +1,110 @@
...
+/**
+ * The {@link SncfDiscoveryService} searches for available
+ * station discoverable through API
+ *
+ * @author Gaël L'hopital - Initial contribution
+ */
+@Component(service = ThingHandlerService.class)
+@NonNullByDefault
+public class SncfDiscoveryService extends AbstractDiscoveryService implements ThingHandlerService {
+    private static final int SEARCH_TIME = 2;
```
lolodomo, line 45: I think 2 seconds is a little too short, especially in case you will run several requests with increasing radius.
10 seconds would be a more reasonable time.
With two seconds, the risk is that your scan could be cancelled too early.

bundles/org.openhab.binding.sncf/src/main/java/org/openhab/binding/sncf/internal/handler/StationHandler.java
```diff
@@ -0,0 +1,223 @@
...
+    public StationHandler(Thing thing, LocationProvider locationProvider) {
+        super(thing);
+        this.locationProvider = locationProvider;
+    }
+
+    @Override
+    public void initialize() {
+        logger.trace("Initializing the Station handler for {}", getThing().getUID());
+
+        stationId = (String) getConfig().get("stopPointId");
```
lolodomo, line 79: Please add a check that `stationId` is not blank and if it is change thing status to OFFLINE (CONFIGURATION_ERROR) .

bundles/org.openhab.binding.sncf/src/main/java/org/openhab/binding/sncf/internal/handler/StationHandler.java
```diff
@@ -0,0 +1,223 @@
...
+
+        if (thing.getProperties().isEmpty()) {
+            discoverAttributes(stationId);
```
lolodomo, line 82: discoverAttributes could have already set the thing status to OFFLINE, either because the bridge is not ready or because the request failed. You should avoid continue running the rest of this method in this case.
So continue only if the method succeeded.

bundles/org.openhab.binding.sncf/src/main/java/org/openhab/binding/sncf/internal/handler/StationHandler.java
```diff
@@ -0,0 +1,223 @@
...
+        }
+        zoneOffset = ZoneId.of(thing.getProperties().get(TIMEZONE)).getRules().getOffset(Instant.now()).getId()
+                .replace(":", "");
```
lolodomo, line 85: You should at least test that property "Tiemzone" is present. If not, set the thing to OFFLINE.

bundles/org.openhab.binding.sncf/src/main/java/org/openhab/binding/sncf/internal/handler/StationHandler.java
```diff
@@ -0,0 +1,223 @@
...
+                            });
+                    scheduleRefresh(fromDTO(arrival.stopDateTime.arrivalDateTime));
+                }, () -> {
+                    logger.debug("No {} available", GROUP_ARRIVAL);
+                    scheduleRefresh(ZonedDateTime.now().plusMinutes(5));
+                });
+                updateStatus(ThingStatus.ONLINE);
+            } catch (SncfException e) {
+                updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.COMMUNICATION_ERROR, e.getMessage());
+                freeRefreshJob();
```
lolodomo, line 160: That means in case of exception, you will then never check again ?
I don't know if it is or not the expected behaviour.

bundles/org.openhab.binding.sncf/src/main/java/org/openhab/binding/sncf/internal/handler/StationHandler.java
```diff
@@ -0,0 +1,223 @@
...
+            case DIRECTION:
+                return new StringType(passage.route.direction.name);
+            case CODE:
+                return new StringType(passage.displayInformations.code);
+            case COMMERCIAL_MODE:
+                return new StringType(passage.displayInformations.commercialMode);
+            case NAME:
+                return new StringType(passage.displayInformations.name);
+            case NETWORK:
+                return new StringType(passage.displayInformations.network);
```
lolodomo, line 176: Maybe you should consider the case the string is null ?

bundles/org.openhab.binding.sncf/src/main/java/org/openhab/binding/sncf/internal/handler/StationHandler.java
```diff
@@ -0,0 +1,223 @@
...
+                });
+                updateStatus(ThingStatus.ONLINE);
+            } catch (SncfException e) {
+                updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.COMMUNICATION_ERROR, e.getMessage());
+                freeRefreshJob();
+            }
+        }
+    }
+
+    private State getValue(String channelId, Passage passage, String groupDeparture) {
```
lolodomo, line 165: Maybe `group` rather than `groupDeparture` as it can be departure or arrival

bundles/org.openhab.binding.sncf/src/main/java/org/openhab/binding/sncf/internal/handler/StationHandler.java
```diff
@@ -0,0 +1,223 @@
...
+            queryApiAndUpdateChannels();
+        }
+    }
+
+    private @Nullable SncfBridgeHandler getBridgeHandler() {
+        Bridge bridge = getBridge();
+        if (bridge != null) {
+            BridgeHandler handler = bridge.getHandler();
+            if (handler != null) {
+                if (handler.getThing().getStatus() == ThingStatus.ONLINE) {
```
lolodomo, line 215: If bridge status is not ONLINE, it would be better to set BRIDGE_OFFLINE rather than BRIDGE_UNINITIALIZED

bundles/org.openhab.binding.sncf/src/main/java/org/openhab/binding/sncf/internal/handler/StationHandler.java
```diff
@@ -0,0 +1,223 @@
...
+        if (thing.getProperties().isEmpty()) {
+            discoverAttributes(stationId);
+        }
+        zoneOffset = ZoneId.of(thing.getProperties().get(TIMEZONE)).getRules().getOffset(Instant.now()).getId()
+                .replace(":", "");
+
+        updateStatus(ThingStatus.ONLINE);
+        scheduleRefresh(ZonedDateTime.now().plusSeconds(2));
+    }
+
```
lolodomo, line 90: As usual, you should override `bridgeStatusChanged` because the default behaviour is only to update the thing status but your code in initialize() will not be run again.

bundles/org.openhab.binding.sncf/src/main/resources/OH-INF/thing/station.xml
```diff
@@ -0,0 +1,80 @@
...
+	<thing-type id="station">
+		<supported-bridge-type-refs>
+			<bridge-type-ref id="api"/>
+		</supported-bridge-type-refs>
+
+		<label>Station</label>
+
+		<channel-groups>
+			<channel-group id="arrivals" typeId="passage">
+				<label>Next Arrival</label>
```
lolodomo, line 16: Next Arrival

bundles/org.openhab.binding.sncf/src/main/resources/OH-INF/thing/station.xml
```diff
@@ -0,0 +1,80 @@
...
+			</channel-group>
+			<channel-group id="departures" typeId="passage">
+				<label>Next Departure</label>
```
lolodomo, line 19: Next Departure

lolodomo (changes requested):
Review part 3 of 3

bundles/org.openhab.binding.sncf/src/main/java/org/openhab/binding/sncf/internal/discovery/SncfDiscoveryService.java
clinique: You're totally right.

bundles/org.openhab.binding.sncf/src/main/java/org/openhab/binding/sncf/internal/handler/StationHandler.java
clinique: Done.

bundles/org.openhab.binding.sncf/src/main/java/org/openhab/binding/sncf/internal/handler/StationHandler.java
clinique: Done.

bundles/org.openhab.binding.sncf/src/main/java/org/openhab/binding/sncf/internal/handler/StationHandler.java
clinique: Done.

bundles/org.openhab.binding.sncf/src/main/java/org/openhab/binding/sncf/internal/handler/StationHandler.java
clinique: Well, if there is a communication error, I think the user has to figure whats going on and reactivate the thing. I could reschedule in a few minutes...WDYT ?

bundles/org.openhab.binding.sncf/src/main/java/org/openhab/binding/sncf/internal/handler/StationHandler.java
clinique: Done.

bundles/org.openhab.binding.sncf/src/main/java/org/openhab/binding/sncf/internal/handler/StationHandler.java
clinique: Yes

bundles/org.openhab.binding.sncf/src/main/java/org/openhab/binding/sncf/internal/handler/StationHandler.java
clinique: Correct.

bundles/org.openhab.binding.sncf/src/main/java/org/openhab/binding/sncf/internal/handler/StationHandler.java
clinique: Done

bundles/org.openhab.binding.sncf/src/main/resources/OH-INF/thing/station.xml
clinique: Sorry, I do not understand your comment.

bundles/org.openhab.binding.sncf/src/main/resources/OH-INF/thing/station.xml
clinique: Sorry, I do not understand your comment.

bundles/org.openhab.binding.sncf/src/main/resources/OH-INF/thing/station.xml
lolodomo: Forget my stupid comment, I thought you used "arrivals" here!

bundles/org.openhab.binding.sncf/src/main/resources/OH-INF/thing/station.xml
lolodomo: Forget my stupid comment, I thought you used "departures" here!

bundles/org.openhab.binding.sncf/src/main/java/org/openhab/binding/sncf/internal/handler/StationHandler.java
lolodomo: Ok, let's keep it like that.

bundles/org.openhab.binding.sncf/src/main/java/org/openhab/binding/sncf/internal/handler/StationHandler.java
lolodomo: You misunderstood my comment: this was appropriate to use BRIDGE_UNINITIALIZED when there is no bridge or no bridge handler. 
Use BRIDGE_OFFLINE when there is a bridge handler but the bridge is OFFLINE.

clinique: Ok, corrected.

bundles/org.openhab.binding.sncf/src/main/java/org/openhab/binding/sncf/internal/handler/StationHandler.java
lolodomo: Not exactly.
Like in `initialize`, you should calll `discoverAttributes` to get the timezone.
In fact, the easier for you would be to just call `initialize` in `bridgeStatusChanged `.

lolodomo:
@clinique : I am going to test the binding.
What remains to do now:

1. enhance the example in documentation
2. fix `bridgeStatusChanged`
3. consider what reports SAT (this is very easy to fix like replacing `URL` by `url` and cutting the catch exception)

clinique:
Fine for me @lolodomo

lolodomo:
Reading the FAQ, I see that for Transilien, thtis is unfortunately not real-time:

> L'API propose les horaires planifiés et temps réel des trains SNCF suivants : TGV, Intercités, TER, Lyria et Eurostar. Les horaires des trains Transilien sont disponibles en planifiés uniquement. Pour avoir accès aux horaires des prochains passages des trains Transilien, rendez-vous sur data.sncf.com.

lolodomo:
Doing my first tests, I see that the values for channel code and name are identical. The values for channels commercial modes and network are identical too. Do you thing it is really necessary to have all these channels ?

In your discovery service, please call the superclass with false as a last parameter as there is no background discovery.

Regarding the discovery, 2 things with the same label are found, one is a bus station (property ending by ":Bus") and the other a train station (property ending by ":RapidTransit"). May you could include the type of station in the label.
Maybe "RapidTransit" is not the only type for a train station.

The refresh of data is not working at all for me.

lolodomo:
> The refresh of data is not working at all for me.

After adding few logs:
```
14:26:22.132 [DEBUG] [sncf.internal.handler.StationHandler] - wishedDelay is 157 seconds
14:26:22.132 [DEBUG] [sncf.internal.handler.StationHandler] - existingDelay is 0 seconds
14:26:22.132 [DEBUG] [sncf.internal.handler.StationHandler] - Do nothing, existingDelay earlier than wishedDelay
```
So the problem is your retrieval of the existing delay which is 0.
Maybe you have  to reschedule when the exiusting delay is 0.

bundles/org.openhab.binding.sncf/src/main/java/org/openhab/binding/sncf/internal/discovery/SncfDiscoveryService.java
```diff
@@ -0,0 +1,110 @@
...
+    private final Logger logger = LoggerFactory.getLogger(SncfDiscoveryService.class);
+
+    private @Nullable LocationProvider locationProvider;
+    private @Nullable SncfBridgeHandler bridgeHandler;
+
+    private int searchRange = 1500;
+
+    @Activate
+    public SncfDiscoveryService() {
+        super(SUPPORTED_THING_TYPES_UIDS, SEARCH_TIME);
```
lolodomo, line 56: Please add a last parameter with value "false" to indicate there is no background discovery.

bundles/org.openhab.binding.sncf/src/main/java/org/openhab/binding/sncf/internal/handler/StationHandler.java
```diff
@@ -0,0 +1,256 @@
...
+    private void scheduleRefresh(@Nullable ZonedDateTime when) {
+        // Ensure we'll try to refresh in one minute if no valid timestamp is provided
+        long wishedDelay = ZonedDateTime.now().until(when != null ? when : ZonedDateTime.now().plusMinutes(1),
+                ChronoUnit.SECONDS);
+        ScheduledFuture<?> job = refreshJob;
+        if (job != null) {
+            long existingDelay = job.getDelay(TimeUnit.SECONDS);
+            if (existingDelay < wishedDelay) {
```
lolodomo, line 143: Take care, if run by the current job, existingDelay will be 0. In this case, you want to reschedule.

bundles/org.openhab.binding.sncf/src/main/resources/OH-INF/thing/station.xml
```diff
@@ -0,0 +1,80 @@
...
+	<channel-type id="commercialMode">
+		<item-type>String</item-type>
+		<label>Commercial Mode</label>
+		<state readOnly="true"></state>
+	</channel-type>
+
+	<channel-type id="timestamp">
+		<item-type>DateTime</item-type>
+		<label>Timestamp</label>
+		<state readOnly="true"/>
```
lolodomo, line 77: Please add a default pattern with only hours and minutes.

bundles/org.openhab.binding.sncf/src/main/java/org/openhab/binding/sncf/internal/discovery/SncfDiscoveryService.java
```diff
@@ -0,0 +1,110 @@
...
+                searchRange += 500;
+                try {
+                    List<PlaceNearby> places = handler.discoverNearby(location, searchRange);
+                    if (places != null && !places.isEmpty()) {
+                        places.forEach(place -> {
+                            String placeId = place.id;
+                            String thingId = placeId.replace(":", "_").replace("-", "_").replace("stop_point_", "");
+                            thingDiscovered(
+                                    DiscoveryResultBuilder.create(new ThingUID(STATION_THING_TYPE, bridgeUID, thingId))
+                                            .withLabel(place.stopPoint.name).withBridge(bridgeUID)
```
lolodomo, line 81: You can have several places with the same `place.stopPoint.name`, like a bus station, a train station, ...
Maybe your label could include the kind of station ?

Alternative: having different thing types for each kind of stations ?

lolodomo (changes requested):

bundles/org.openhab.binding.sncf/src/main/java/org/openhab/binding/sncf/internal/handler/StationHandler.java
clinique: Done.

clinique:
> @clinique : I am going to test the binding. What remains to do now:
> 
>     1. enhance the example in documentation (if you are ok)
> 
>     2. fix `bridgeStatusChanged`
> 
>     3. consider what reports SAT (this is very easy to fix like replacing `URL` by `url` and cutting the catch exception)

Done for the last two points

bundles/org.openhab.binding.sncf/src/main/java/org/openhab/binding/sncf/internal/discovery/SncfDiscoveryService.java
clinique: Done

bundles/org.openhab.binding.sncf/src/main/java/org/openhab/binding/sncf/internal/handler/StationHandler.java
clinique: Yes, identified and circumvented.

bundles/org.openhab.binding.sncf/src/main/resources/OH-INF/thing/station.xml
clinique: Done

bundles/org.openhab.binding.sncf/src/main/java/org/openhab/binding/sncf/internal/discovery/SncfDiscoveryService.java
clinique: The trick is apparently the kind of stop_point is the last element of the placeId (I did not found any alternate API call that could provide this information in a clean way) - but there is no 100% certainty that it is always true. So I was not happy to pull this information in the label.

bundles/org.openhab.binding.sncf/src/main/java/org/openhab/binding/sncf/internal/handler/StationHandler.java
lolodomo: Refresh seems to work well now.

bundles/org.openhab.binding.sncf/src/main/java/org/openhab/binding/sncf/internal/handler/StationHandler.java
lolodomo: For me, the best implementation would have been to start from the super class code, something like that:
```
    @Override
            initialize();
        }
    }
```

bundles/org.openhab.binding.sncf/src/main/java/org/openhab/binding/sncf/internal/discovery/SncfDiscoveryService.java
lolodomo: Your new code is better.

bundles/org.openhab.binding.sncf/src/main/resources/OH-INF/thing/station.xml
lolodomo: You followed my comment, thank you.
But now I see there is one mine between the arrival and the departure time with the default pattern, like 19:16 and 19h17. With the full pattern, I remember the difference was 40 seconds and so the departure time was in the same minute.
All that to say that maybe you should include seconds in the default pattern ?

bundles/org.openhab.binding.sncf/src/main/java/org/openhab/binding/sncf/internal/handler/StationHandler.java
```diff
@@ -0,0 +1,257 @@
...
+        }
+        logger.debug("Scheduling update in {} seconds.", wishedDelay);
+        refreshJob = scheduler.schedule(() -> updateThing(), wishedDelay, TimeUnit.SECONDS);
+    }
+
+    private void updateThing() {
+        SncfBridgeHandler bridgeHandler = getBridgeHandler();
+        if (bridgeHandler != null) {
+            scheduler.submit(() -> updatePassage(bridgeHandler, GROUP_ARRIVAL));
+            scheduler.submit(() -> updatePassage(bridgeHandler, GROUP_DEPARTURE));
```
lolodomo, line 161: What annoys me with these calls called in parallel in separate threads is that you have to consider concurrent calls to `scheduleRefresh` and `freeRefreshJob`. I am not sure this is something you took into consideration, did you ?
Is it really necessary to do that ? Two calls in sequence could be ok, no ?

bundles/org.openhab.binding.sncf/src/main/java/org/openhab/binding/sncf/internal/handler/StationHandler.java
```diff
@@ -0,0 +1,257 @@
...
+        }
+        return UnDefType.NULL;
+    }
+
+    private State fromNullableString(@Nullable String aValue) {
+        return aValue != null ? StringType.valueOf(aValue) : UnDefType.NULL;
+    }
+
+    private @Nullable ZonedDateTime fromDTO(@Nullable String dateTime) {
+        return dateTime != null ? ZonedDateTime.parse(dateTime + zoneOffset, NAVITIA_DATE_FORMAT).plusMinutes(1) : null;
```
lolodomo, line 212: Do you really need to add one minute ?
I will compare timestamps with my SNCF app as soon as data are available again !

bundles/org.openhab.binding.sncf/src/main/resources/OH-INF/thing/station.xml
```diff
@@ -0,0 +1,81 @@
...
+			</parameter>
+		</config-description>
+	</thing-type>
+
+	<channel-group-type id="passage">
+		<label>Other</label>
+		<channels>
+			<channel id="direction" typeId="direction"/>
+			<channel id="code" typeId="code"/>
+			<channel id="commercialMode" typeId="commercialMode"/>
```
lolodomo, line 37: I am not sure the user needs two channels. What he needs is the "line". As an example, having "RER" as commercial mode + "A" as code in two different channels could be replaced by a simple channel that will deliver "RER A".
WDYT ?

bundles/org.openhab.binding.sncf/src/main/resources/OH-INF/thing/station.xml
lolodomo: Sometimes I have the same minute with the current code, so first I should check the raw data.

lolodomo:
I just saw a case while next arrival was at 19:45 and at 19:45 it changed to 19:46 (same train). So it would mean this is real-time.

bundles/org.openhab.binding.sncf/src/main/java/org/openhab/binding/sncf/internal/handler/StationHandler.java
lolodomo: In fact, this is probably to avoid retrieving the same train with the same timestamp ?
You refresh one minute later  so that either you will switch to another train or the same train but with a new timestamp.

lolodomo:
I will extend my search radius to see how the channels are set in a bug train station like Paris gare de Lyon for example.

bundles/org.openhab.binding.sncf/src/main/java/org/openhab/binding/sncf/internal/handler/StationHandler.java
clinique: Well, I had to do this because I faced a concurrent exception at jetty level with the previous successive calls. It is the only I found to solve the problem.

bundles/org.openhab.binding.sncf/src/main/java/org/openhab/binding/sncf/internal/handler/StationHandler.java
clinique: You're correct. I introduced that to be sure I will not be requesting for the same train that left the train station before the updated information is available. On the other hand I did not figure it will impact channel data publications. I've got to correct his.

bundles/org.openhab.binding.sncf/src/main/resources/OH-INF/thing/station.xml
clinique: I will check if it works also for buses, but basically, you're right.

lolodomo:
@clinique : it is strange, it looks like I got only trains going in one direction, not both.

bundles/org.openhab.binding.sncf/src/main/java/org/openhab/binding/sncf/internal/handler/StationHandler.java
clinique: Changed.

bundles/org.openhab.binding.sncf/src/main/resources/OH-INF/thing/station.xml
clinique: Changed.

clinique:
> @clinique : it is strange, it looks like I got only trains going in one direction, not both.

Apparently it works fine : 
`12:41:19.301 [INFO ] [openhab.event.ItemStateChangedEvent ] - Item 'PoissyRapidTransit_Departures_Direction' changed from Boissy-Saint-Léger (Boissy-Saint-Léger) to Paris Saint-Lazare (Paris)
12:41:19.301 [INFO ] [openhab.event.ItemStateChangedEvent ] - Item 'PoissyRapidTransit_Departures_LineName' changed from RER A to TRANSILIEN J
12:41:19.302 [INFO ] [openhab.event.ItemStateChangedEvent ] - Item 'PoissyRapidTransit_Departures_Timestamp' changed from 2021-12-01T12:40:20.000+0100 to 2021-12-01T12:54:00.000+0100`

clinique:
@lolodomo : I think I've addressed your last comments / remarks. I'm going to push a new update.

bundles/org.openhab.binding.sncf/src/main/resources/OH-INF/thing/station.xml
clinique: I think it will be clearer  if I reintroduce seconds in the pattern.

lolodomo:
@clinique, I am running your last version.
Unfortunately, I can confirm there is a major issue with the direction which is often wrong. Good timestamps and good line but wrong direction.
Maybe the wrong field is retrieved in the JSON ?

lolodomo:
I found the solution, the direction has to be retrieved in display_information.
In route, you have direction + direction_type being either "forward" or "backward".
So this code is working after adding the field `direction`:
```
            case DIRECTION:
                // return fromNullableString(passage.route.direction.name);
                return fromNullableString(passage.displayInformations.direction);
```

clinique:
> I found the solution, the direction has to be retrieved in display_information. So this code is working after adding the field `direction`:
> 
> ```
>             case DIRECTION:
>                 // return fromNullableString(passage.route.direction.name);
>                 return fromNullableString(passage.displayInformations.direction);
> ```

Thanks for your bug hunting. Pushing the update.

bundles/org.openhab.binding.sncf/README.md
clinique: Done

bundles/org.openhab.binding.sncf/src/main/java/org/openhab/binding/sncf/internal/handler/StationHandler.java
lolodomo: I still do not understand why you create a new thread while `updateThing` is already executed by a scheduled thread ?

clinique: If/when I don't do it in a thread, I get this error on second call : 
``
17:32:17.180 [DEBUG] [f.internal.handler.SncfBridgeHandler] - Execution interrupted : null
java.lang.InterruptedException: null
	at java.util.concurrent.locks.AbstractQueuedSynchronizer.acquireSharedInterruptibly(AbstractQueuedSynchronizer.java:1343) ~[?:?]
``

lolodomo:
@clinique : still one comment not fully resolved (about creating additional threads).

My tests are positive.

New general comments:

1. In case you have two trains at the same time (same minute), for example each in one direction, you will loose one of them. Maybe refreshing ONE minute after the planed arrival could be reduced to 10 seconds after the planed arrival to reduce this risk ? Another option would be to have channels for the two next trains. WDYT ?

2. Regarding default pattern for timestamps, I still think it should rather be only hours + minutes but it should rounds to minutes, that is the real value is 18:30:50 should be displayed as 18:31. I compared to the the times provided by the SNCF Android app. It could be improved later.

3. You did not find the train platform in the returned data (I did not see it myself but have only a quick look on data) ?

4. After my tests, I have the feeling that for Ile-de-France, the data are planned data, oot real-time data.

lolodomo:
I forgot one point: now that `route` is no more used, unsued classes could be removed ?

clinique:
> @clinique : still one comment not fully resolved (about creating additional threads).

Just answered it a few minutes ago and retested (see above in the conversation thread).
> 
> My tests are positive.
> 
> New general comments:
> 
>     1. In case you have two trains at the same time (same minute), for example each in one direction, you will loose one of them. Maybe refreshing ONE minute after the planned arrival could be reduced to 10 seconds after the planned arrival to reduce this risk ? Another option would be to have channels for the two next trains. WDYT ?
> 

I'm going to reduce to 10 seconds and see if it does the job.

>     2. Regarding default pattern for timestamps, I still think it should rather be only hours + minutes but it should round to minutes, that is the real value is 18:30:50 should be displayed as 18:31. I compared to the the times provided by the SNCF Android app. It could be improved later.
> 

I quite do not get the point of rounding a timestamp provided sharp. If used in automation, the sharper the better. Ok to come back on this on return of users experience.

>     3. You did not find the train platform in the returned data (I did not see it myself but have only a quick look on data) ?
> 

Yes, I searched but did not find it neither.

>     4. After my tests, I have the feeling that for Ile-de-France, the data are planned data, oot real-time data.

I have a plan to take care of announced service disruptions (not time to dive on this currently), that could generate difference between planned and real-time data.

lolodomo:
> I quite do not get the point of rounding a timestamp provided sharp. If used in automation, the sharper the better. Ok to come back on this on return of users experience.

I don't want to round the raw data, only the default display value in UI to show only hours+minutes.

clinique:
> > I quite do not get the point of rounding a timestamp provided sharp. If used in automation, the sharper the better. Ok to come back on this on return of users experience.
> 
> I don't want to round the raw data, only the default display value in UI to show only hours+minutes.

Don't you see it'll be weird for users to see a given time in logs and a different one in the interface ?

lolodomo:
> > > I quite do not get the point of rounding a timestamp provided sharp. If used in automation, the sharper the better. Ok to come back on this on return of users experience.
> > 
> > 
> > I don't want to round the raw data, only the default display value in UI to show only hours+minutes.
> 
> Don't you see it'll be weird for users to see a given time in logs and a different one in the interface ?

Yes, maybe.
But in SNCF app, they display hours+minutes. They clearly do not have a precision at the second ;)
But ok, keep it as it is.

bundles/org.openhab.binding.sncf/src/main/java/org/openhab/binding/sncf/internal/handler/StationHandler.java
lolodomo: Ok I think I understand. It is probably due to your task cancelation (freeRefreshJob).
By creating this additional, you avoid canceling yourself ;)
I am not sure this is a clean implementation but let's keep it like that.

lolodomo (approved):
LGTM now

lolodomo:
@clinique: let me know if your test is positive and I will then merge. I am going to run my own test in parallel.

clinique:
> @clinique: let me know if your test is positive and I will then merge. I am going to run my own test in parallel.

Yes, you can merge. Thanks for your attentive and patient review !

lolodomo:
Yes, it seems to work well for me too.
Thank you for your new binding contribution Gael.

lolodomo:
@clinique: if not yet done, you could add your binding's logo to the openHAB website. See https://www.openhab.org/docs/developer/bindings/#add-your-binding-s-logo-to-the-openhab-website

PR #11613, opened 2021-11-20
[Souliss] Fix t14 typical

lucacalcaterra:
Fix t14 (item type must be of Switch instead of Trigger)


bundles/org.openhab.binding.souliss/src/main/resources/OH-INF/thing/thing-types.xml
```diff
@@ -1057,7 +1057,7 @@
 
 
 	<channel-type id="pulse">
-		<kind>trigger</kind>
+        <item-type>Switch</item-type>
```
lolodomo, line 1060: You need to apply spotless: mvn spotless:apply

lolodomo (changes requested):

lolodomo:
This PR looks terribly strange. 
For example, all your changes in SoulissGatewayDiscovery are in fact already present in repo. 
Can you provide a new PR based on current main branch with only what you would like to fix?

lucacalcaterra:
> This PR looks terribly strange. For example, all your changes in SoulissGatewayDiscovery are in fact already present in repo. Can you provide a new PR based on current main branch with only what you would like to fix?

I created a PR to openhab:main from lucacalcaterra:souliss (the branch i worked from begin)... probably my error is this... should i work on main? Or i can continue to work on my souliss branch ?

lolodomo:
Update your main branch and then create a new branch from your main branch. Apply your changes to this new branch and finally push your PR.

lucacalcaterra:
I know, in fact I'm doing it ..
however it was a curiosity, I wanted to understand if I could make a pr from the old branch

PR #11614, opened 2021-11-20
[Souliss] Fix T14 Typical

lucacalcaterra:
fix T14 typical - must be switch instead of trigger
Signed-off-by: Luca Calcaterra <calcaterra.luca@gmail.com>


lolodomo (approved):
LGTM

PR #11619, opened 2021-11-21
[tr064] Added default i18n properties file

cweitkamp:


wborn (approved):
Thanks!

PR #11626, opened 2021-11-21
[max] Add default translation properties file for MAX!

marcelrv:


cweitkamp (approved):
LGTM. Thanks.

PR #11534, opened 2021-11-06
[hdpowerview] Add support for scene groups

jlaur:
Fixes #11533

Signed-off-by: Jacob Laursen <jacob-github@vindvejr.dk>

The Hunter Douglas PowerView Hub supports two different scene concepts:
- Scenes
- Scene Collections

A scene collection is a collection of scenes and can be used to group multiple scenes and activate them all with a single click or automation. When this feature was available in the Hunter Douglas PowerView app they were called scene groups. This terminology it kept in user-oriented strings, while API name is scene collections.

This PR adds support for activation of scene groups. Scene groups will appear as channels the same way as scenes, like this:
image

PR #11637, opened 2021-11-24
[hdpowerview] Add support for enabling/disabling automations

### LINKED ISSUE #11516: [hdpowerview] Provide channels for switching automations on and off
A few times per year I go into the PowerView app and switch one automation **on** and another one **off**. In the summer I want the shades to go up at a specific time to wake up the kids with sunlight. In the fall when the sun rises later than wake-up time, I disable this automation and enable another one for having the shades up at sunrise. And again the opposite in the spring when the sun will start to rise too early again. Same scenario in bedroom, just different fixed time, so also different date for switching automations. Same for sunset, so again different dates.

The reason for using the PowerView automations is to have fewer dependencies. The Hunter Douglas Hub provides the needed options for activating scenes at a fixed time, sunrise/sunset or offsets from sunrise/sunset. So this works fine without a dependency to openHAB and the hardware it runs on.

What I would like the Hunter Douglas PowerView binding to do for me is managing the automations, i.e. with the Astro binding and some rules to enable/disable specific automations when needed.

With Packet Capture for Android it was a matter of seconds to find the API calls needed for this.

**GET /api/scheduledevents/**

to find all automations. Each of them can refer to either a sceneId or sceneCollectionId, for example:
```
		{
			"enabled": true,
			"sceneCollectionId": 27119,
			"daySunday": false,
			"dayMonday": true,
			"dayTuesday": true,
			"dayWednesday": true,
			"dayThursday": true,
			"dayFriday": true,
			"daySaturday": false,
			"eventType": 1,
			"hour": 0,
			"minute": 0,
			"id": 1262
		},
```

To disable this:
**PUT /api/scheduledevents/1262**

with payload
```
{
	"scheduledEvent": {
		"dayFriday": true,
		"dayMonday": true,
		"daySaturday": false,
		"daySunday": false,
		"dayThursday": true,
		"dayTuesday": true,
		"dayWednesday": true,
		"enabled": false,
		"eventType": 1,
		"hour": 0,
		"id": 1262,
		"minute": 0,
		"sceneCollectionId": 27119
	}
}
```

Scenes are fetched with **GET /api/scenes/** and scene collections with **GET /api/sceneCollections/**. Names are Base64-encoded.

jlaur:
Fixes #11516

This pull request adds automation support to the Hunter Douglas PowerView binding. This can be useful for keeping as much logic in the PowerView Hub as possible to reduce dependencies. A typical use-case would be switching between winter and summer automations. In the summer one might want the shades up on weekdays at a fixed time. In the winter when still dark at this time, an offset relative to sunrise might be more appropriate. It will now be possible with rules to switch between such two automations when the sun will start to rise before or after some fixed time.

Channel groups have been introduced for:
- **Scenes:** Items will need to be relinked to new channel. Backwards compatible channels are created as long as items are linked to them and new channels are unlinked. See also: [Naming convention for channels?](https://community.openhab.org/t/naming-convention-for-channels/128841/10)
- **Scene Groups:** Introduced in #11534 so no backwards compatibility needed if PR will be included in 3.2.
- **Automations:** New.

Visual example:

image

### Additional changes
- Scene and scene group channels are now refreshed automatically, so renaming will be reflected.
- Scene and scene group channels are now sorted by order provided by the API, i.e. same order as in app, except that room is not taken into consideration.
- Scene and scene group channel updates have been optimized for the two most common scenarios: Create all of them initially and skip when nothing has changed.
- Dynamic channels are flushed during initialization (and rebuilt after), so changed system language will be reflected at least when binding is restarted. Previously this required openHAB to be restarted.


jlaur:
JAR: https://drive.google.com/file/d/15Wgt_cuNb3XUEbHejoAeEj31scXqgBCL/view?usp=sharing

jlaur:
@andrewfg, @arroyoj - tagging you directly in case you would be interested in this (reviewing and/or testing).

arroyoj:
@jlaur, this looks like a nice set of new features. I don't use automations, but the improvements to the dynamic channels, like automatic renaming, are things I am looking forward to. I had one thought about how to address the breaking change of moving scene channels into a channel group. I don't fully understand OpenHAB's internals and channel handling, so sorry if this doesn't make sense, but if a user has an item linked to the old (non-grouped) scene channel, will commands from that item still get sent to the hub Thing, but then dropped/ignored? I guess what I'm wondering is rather than making deprecated channels to handle the old item links (which I think you tested but didn't work?), could you still handle the old channels and issue a warning?

The section of the code that I'm thinking about is in `handleCommand` here: https://github.com/jlaur/openhab-addons/blob/f6b6ec5dff81344f91248b93e6cfdf742f392655/bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/handler/HDPowerViewHubHandler.java#L112-L142, particularly this part:
```
            HDPowerViewWebTargets webTargets = this.webTargets;
            if (webTargets == null) {
                throw new ProcessingException("Web targets not initialized");
            }
            int id = Integer.parseInt(channelUID.getIdWithoutGroup());
            if (sceneChannelTypeUID.equals(channel.getChannelTypeUID()) && OnOffType.ON.equals(command)) {
                webTargets.activateScene(id);
            } else if (sceneGroupChannelTypeUID.equals(channel.getChannelTypeUID()) && OnOffType.ON.equals(command)) {
                webTargets.activateSceneCollection(id);
            } else if (automationChannelTypeUID.equals(channel.getChannelTypeUID())) {
                webTargets.enableScheduledEvent(id, OnOffType.ON.equals(command));
            }
```

In `handleCommand`, it looks like you get the ID associated with the channel, and then send the ID to the appropriate handler based on which channel group it came from. I'm wondering if you can assume that any ID that comes from a channel *without* a channel group is an old scene channel link, and so have a final `else` clause that uses that ID to activate the corresponding scene and issue a warning or something that the scene is being activated through a deprecated link. This way, a user's old scene links would still work, assuming those channels were still considered "valid" by OpenHAB and made it to the `handleCommand` method. I'm not sure what this would do to the UI for links, but maybe when they see the link is missing, they would check the logs and see the message. If this approach worked, though, it would be nice that user's scene items kept working after the update.

jlaur:
@arroyoj - thanks for your comments. If you are right, the current version of the code would already work. However, `handleCommand` receives `ChannelUID`, so I'm not sure the framework will call it for a channel that doesn't exist, even if an item is linked to it. I'm not 100% sure right now, but can test it tomorrow as unfortunately I won't have any time today. You are also welcome to give the JAR a try. :-) It's updated after latest commit and includes all previous changes, including the getter/setter consistency fix.

jlaur:
@arroyoj - unfortunately it doesn't work, `handleCommand` will not be called when channel is unknown:
image

arroyoj:
@jlaur, thanks for testing it. I tried it out too, and I realized that I was misunderstanding how channels are handled. I was thinking any channel with the right "prefix" would be forwarded to the Thing, but clearly not. It would be great if there was some way to create a "catch-all" channel, but I could find no evidence of that in the core code.

I did have another thought about how you could structure the logic of channel creation to maintain backwards compatibility while avoiding the situation described in [Naming convention for channels?](https://community.openhab.org/t/naming-convention-for-channels/128841/10). If I understand the problem you described in that post, items linked to channels *with* a group ID (ie, `scenes#123`) will also be detected as linked to a similar channel *without* the group ID (ie, `123`). So, if you create an "old"/deprecated no-group channel, you cannot tell if an item linked to it is a deprecated link or a correct new link. In contrast, items previously linked to old channels (ie, `123`) will *not* be detected as linked to new with-group channels (ie, `scenes#123`). Am I understanding the problem right?

If that is correct, could the following logic (expressed as pseudocode) work for dynamic channel creation while keeping backwards compatibility?
```
for (scene_id in scene_list_from_hub) {
  channel_id_with_group = new ChannelUID(group id, scene_id)
  deprecated_channel_id = new ChannelUID(scene_id)
  if ( !isLinked(channel_id_with_group) ) {
    // This user does not have an item linked to the new channel, so maybe they have an old link
    if ( isLinked(deprecated_channel_id) ) {
      // Old link detected, need deprecated channel
      createChannel(deprecated_channel_id)
      logger.warn("Deprecated channel being used")
    }
  }
  // Always create the correct new channel
  createChannel(channel_id_with_group)
}
```

The two `if` statements could possibly be combined, but I kept them separate to try to show the logic. The second `if` statement might incorrectly be "true" if the user has new channel links, but the first `if` statement should guard against that because it will only be true if there are no new channel links.

The assumption here is that if the user has created at least one new channel link for a given scene, they no longer need the deprecated channel. On the other hand, if they have only an old channel link for the scene, a deprecated channel is needed to keep it working, but the new channel should also be available. If someone had two items linked to the same scene, it is possible only one would be updated and the other could stop working, but my guess is most people make a single item for each scene. And, if someone has not linked a scene at all, only the new channel should be created.

Sorry if you have tried this already or if this won't work out as I thought. Unfortunately, I don't have a way to test this myself.

jlaur:
@arroyoj - thanks, that's a simple and quite reasonable approach - the missing piece in the puzzle, I would say. I've implemented it and this logic works and seems straight-forward. JAR updated.

andrewfg:
Two comments from my side..
1. I have not checked your code, but I fully support your modifications to ensure that on existing user's systems, the links of pre-existing Items to pre-existing scene Channels will not be broken. (Indeed I would insist that they would not be broken).
2. I note that my just opened PR #11698 is touching some of the same files as this PR does, so it is possible that the one or the other PR might cause merge conflicts; I hope not, but lets see; It probably depends on whose PR gets merged first..

jlaur:
> 1. I have not checked your code, but I fully support your modifications to ensure that on existing user's systems, the links of pre-existing Items to pre-existing scene Channels will not be broken. (Indeed I would insist that they would not be broken).

I'm preparing to create an issue for openhab-core for supporting channel migrations. See also:
https://community.openhab.org/t/naming-convention-for-channels/128841/7

Currently it's hard to manage, but with @arroyoj's suggestion I believe the current logic is quite acceptable. See also #11668.

> 2. I note that my just opened PR [[hdpowerview] Fix secondary position bug. Add shade database and properties. #11698](https://github.com/openhab/openhab-addons/pull/11698) is touching some of the same files as this PR does, so it is possible that the one or the other PR might cause merge conflicts; I hope not, but lets see; It probably depends on whose PR gets merged first..

I did a first quick around of your PR, and as far as I can see we only collide in `HDPowerViewBindingConstants` which will be easy to resolve. Let's hope it will stay that way and we can have the PR's reviewed and merged sometime soon, so we can be ready for the next rounds. :-)

jlaur:
> I'm preparing to create an issue for openhab-core for supporting channel migrations. See also: https://community.openhab.org/t/naming-convention-for-channels/128841/7

Created openhab/openhab-core#2590

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/handler/HDPowerViewHubHandler.java
```diff
@@ -215,8 +239,14 @@ private synchronized void poll() {
         try {
             logger.debug("Polling for state");
             pollShades();
-            pollScenes();
-            pollSceneCollections();
+            List<Scene> scenes = pollScenes();
+            updateSceneChannels(scenes);
+            createDeprecatedSceneChannels(scenes);
+            List<SceneCollection> sceneCollections = pollSceneCollections();
```
andrewfg, line 245: Shouldn't this code be within its own method?

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/handler/HDPowerViewHubHandler.java
```diff
@@ -287,41 +317,74 @@ private void pollScenes() throws JsonParseException, HubProcessingException, Hub
...
+                HDPowerViewBindingConstants.CHANNEL_GROUP_SCENES);
+        ChannelUID channelUid = new ChannelUID(channelGroupUid, Integer.toString(scene.id));
+        String description = translationProvider.getText("dynamic-channel.scene-activate.description", scene.getName());
+        Channel channel = ChannelBuilder.create(channelUid, CoreItemFactory.SWITCH).withType(sceneChannelTypeUID)
+                .withLabel(scene.getName()).withDescription(description).build();
+
+        return channel;
+    }
+
+    private void createDeprecatedSceneChannels(List<Scene> scenes) {
```
andrewfg, line 350: You use the word 'deprecated'; which is usually applied to describe code that will at some time in future be removed entirely from the code-base. So I wonder if perhaps you should use the word 'legacy' instead; which is usually applied to describe old implementations that will be maintained (non breaking code) in existing implementations, but which would/should not be be used in virgin implementations.

andrewfg: PS I think you should change 'deprecated' => 'legacy' also in the Issue that you opened in OH core. As I think that the core maintainers will otherwise not understand the purpose of your Issue.

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/handler/HDPowerViewHubHandler.java
jlaur: Which code exactly, all four lines? Why?

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/handler/HDPowerViewHubHandler.java
jlaur: It's deprecated with the intention of being removed sometime in the future. If you insist, I'm okay with renaming, but for user-oriented texts I would still prefer "deprecated", as they should migrate to the new channels. Any objections to that?

jlaur: Instead, maybe I could add method documentation stating explicitly that this is for backwards compatibility and might break some users installations if/when removed, in case they have not migrated?

jlaur: Documentation added as it explains the purpose in any case.

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/handler/HDPowerViewHubHandler.java
andrewfg: No. I meant all the whole block of lines that you had added at that point. Because they comprise a separate block of functionality Which is not actually covered under the title of @poll()').

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/handler/HDPowerViewHubHandler.java
andrewfg: Urrh. Ok.

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/handler/HDPowerViewHubHandler.java
andrewfg: Lines 243 thru 249. Maybe into **two** sub functions: updateSceneChannels() and updateScheduledEvents()..

jlaur: Ah, in that way. Well, actually it's unchanged, I just divided polling and channel updating into separate methods to make them more distinct and smaller. Do you think it would be more appropriate to rename the `poll` method in your opinion? We fetch scenes (A), scene collections (B) and scheduled events (C) and I'd prefer to fetch A, update channels for A, fetch B, update channels for B, fetch C and update channels for C. So if fetching automations, for example, would fail for any reason, we would still have the scenes and scene collections updated. Any suggestions much welcome. :)

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/handler/HDPowerViewHubHandler.java
jlaur: I did some additional reading, and I believe we are in some grey zone.

[Wikipedia: Deprecation](https://en.wikipedia.org/wiki/Deprecation#Software) - applies here:
- "Discouragement of use"
- "It has been superseded"
- "without completely removing it or prohibiting its use"

Also this thread: [Stack Overflow: Difference between a deprecated and a legacy API?](https://stackoverflow.com/questions/2873254/difference-between-a-deprecated-and-a-legacy-api)

> Perhaps an old code using legacy but non-deprecated class requires no action, since for now at least, they aren't in danger of ceasing to exist in future version."
> 
> In contrast, deprecation explicitly warns that they may cease to exist, so action should be taken to migrate to the replacement.

Action is required by the users here - they are encouraged to migrate to the new channels by issuing a warning.

So IMHO deprecation is at least as good as lesser defined "legacy" here. Closing comment.

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/handler/HDPowerViewHubHandler.java
jlaur: @andrewfg - I have cleaned up the `poll` method as I agree it had become a bit cluttered. Methods are still divided, but responsibility to call data fetching methods has now been propagated to the channel update methods. Also, `createDeprecatedSceneChannels` is now called from `updateSceneChannels`, so now it won't be called that often anymore.

jlaur:
@andrewfg - do you have more comments, or do you think it's in a state ready for being merged?

andrewfg:
> do you have more comments

I don't have any more comments. 

(But -- to be honest -- since I do not actually have automations in my system, I have not fully tested it. Therefore please take my lack of comments as a 'no objection' rather than an 'active approval')..

jlaur:
I have been using this version in production all the way, and have tested the following:
- Scenes can be activated from old channel not belonging to any channel group (backwards compatibility).
- Scenes can be activated from new channel belonging to channel group.
- Old scene channels not belonging to any channel group and not linked to any items are not created.
- Scene Groups can be activated from new channel belonging to channel group. Scene Groups were not part of 3.1, so no backwards compatibility provided.
- Automations can be enabled/disabled.
- When automations are enabled/disabled from the outside (e.g. PowerView app) this is reflected correctly.
- When scenes, scene groups or automations are changed from the outside, this is reflected (for example if a scene is renamed).
- If locale is changed and binding is restarted, channels are updated.

@lolodomo - would you consider this PR for 3.2? The advantage of including it would be less backwards compatibility issues as scene groups were not part of 3.1. Otherwise will have to create some more additional logic for scene group backwards compatibility.

arroyoj:
I just tested out the latest JAR from @jlaur on my system. From a user perspective, it all looks good to me. Here is what I tested:

- Backwards compatible scene channels were created and worked. All of my old scene links were maintained, and scene items continued to function properly. The warning messages to upgrade the links showed up in the logs. On the Channels page, I saw both the old scene channels (marked as DEPRECATED) and the new scene channels.
- The new scene channels showed up in a separate "Scenes" channel group.
- Creating a new item and linking it to a new scene channel (ie, `scenes#123`) worked. The item linked to the new scene channel behaved the same as if it was linked to the deprecated channel.
- After restarting my system with a new scene channel linked (ie, `scenes#123`), the corresponding deprecated channel (ie, `#123`) was not created, as expected, and no warning message about that channel was logged, again as expected.
- Renaming a scene in the Powerview App updated the name for the corresponding scene channel! This is great! When I first set up my system, I actually went through and deleted scenes in the Powerview App and then recreated them with new names just because it bothered me that the names in OpenHAB were not the same after editing. Thanks for incorporating this.
- Although I don't use automations, I created an automation in the Powerview App for testing and it promptly showed up in OpenHAB as a new automations channel, within an "Automations" channel group.
- Linking an item to this automation channel allowed me to enable and disable the automation, and the enabled/disabled state set in OpenHAB was reflected in the Powerview App, and vice versa.

I think that covers the features described in the PR, and I didn't see any issues arise. Thanks @jlaur for adding these features.

jlaur:
Thanks a lot for verifying the functionality, @arroyoj.

andrewfg (approved):
LGTM

jlaur:
@lolodomo, @kaikreuzer - any chance of this getting into 3.2?

andrewfg:
@lolodomo / @kaikreuzer -- it would be super cool if you could squeeze in #11698 as well ;)

fwolter (approved):

PR #11644, opened 2021-11-28
[openwebnet] Fixes discovery of devices on local bus. Added it translation

### LINKED ISSUE #11582: [openwebnet] discovery on local bus sets wrong where address
## Expected Behavior
Devices configured on local bus should be discovered with correct where address parameter , for example `where=12*4*03`

## Current Behavior
Devices configured on local bus are discovered with wrong where address parameter, for example `12h4h03`

## Possible solution
After discovery a light device on local bus, change its where by changing all `h` with `*`

Discover a device on local bus and add it as thing. It will have a wrong where parameter and it will not work with OH

mvalla:
This PR fixes #11582: while discovering devices on local bus the where parameter is now set correctly, not using normalizeWhere().
In addition this PR adds translations string for `it ` locale.


wborn (approved):
Thanks for the fix!

PR #11646, opened 2021-11-28
[transform][map] Fixed wrong i18n properties and added label of Profile

cweitkamp:
- Fixed wrong i18n properties and added label of Profile

The i18n properties for Profiles were not created correctly in https://github.com/openhab/openhab-addons/pull/11574. Additional the label of the Profile was not added automatically.


wborn:
I made some changes in the plugin that should fix this. :+1: 
It is probably also better to name the file `map.properties`

cweitkamp:
I will rename the file and apply all the changes to other languages directly in Crowdin.

wborn (approved):
Thanks!

PR #11648, opened 2021-11-28
Add i18n-maven-plugin dependency

wborn:
When the plugin dependency is managed you can also use the plugin without adding GAV parameters to commands.

E.g. it allows for using it with:

```
mvn i18n:generate-default-translations
```

Depends on https://github.com/openhab/openhab-core/pull/2544


cweitkamp (approved):
LGTM. Thanks.

PR #11653, opened 2021-11-28
[lifx] Improve InterruptedException handling

wborn:
When the binding is stopped sleeping threads are interrupted by design.
By throwing the InterruptedException, it should be caught in LifxSelectorUtil.sendPacket (which is waiting for the packet interval to elapse), which will then abort sending a packet.

This prevents:

```
[ERROR] [lifx.internal.util.LifxThrottlingUtil] - An exception occurred while putting the thread to sleep : 'sleep interrupted'
```


lolodomo (approved):
LGTM

PR #11657, opened 2021-11-28
[miio] add support for BT Gateway switch on chuangmi.plug.212a01

marcelrv:

### DIFF
--- a/bundles/org.openhab.binding.miio/src/main/java/org/openhab/binding/miio/internal/basic/Conversions.java
+++ b/bundles/org.openhab.binding.miio/src/main/java/org/openhab/binding/miio/internal/basic/Conversions.java
@@ -24,6 +24,7 @@ import org.slf4j.LoggerFactory;
 
 import com.google.gson.JsonElement;
 import com.google.gson.JsonObject;
+import com.google.gson.JsonParseException;
 import com.google.gson.JsonParser;
 import com.google.gson.JsonPrimitive;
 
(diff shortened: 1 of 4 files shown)


bundles/org.openhab.binding.miio/src/main/java/org/openhab/binding/miio/internal/basic/Conversions.java
```diff
@@ -121,9 +123,28 @@ public static JsonElement tankLevel(JsonElement value12) throws ClassCastExcepti
         }
     }
 
+    public static JsonElement getJsonElement(String element, JsonElement responseValue)
+            throws ClassCastException, IllegalStateException {
```
fwolter, line 128: You'd better know before accessing it if it's the right type in the right state. Catching these is bad.

fwolter (changes requested):

bundles/org.openhab.binding.miio/src/main/java/org/openhab/binding/miio/internal/basic/Conversions.java
marcelrv: maybe I'm overly causious here.. as the responsevalue is something that comes from the device and can have rather unexpected values I normally rather have a friendly message than an ugly stacktrace.

That said, I think the code indeed has little chance to generate these errors even in case of strange inputs so i'll remove it

fwolter (approved):
LGTM

PR #11658, opened 2021-11-28
[miio] fix Error parsing miot data null for new miot devices

marcelrv:


lolodomo:
I do not see an obvious link between the code change and your PR title.

marcelrv:
It is the error in the screen you see..
Seems they no longer allow http, which results in a null, which gives this message

wborn (approved):
Thanks for the fix!

PR #11666, opened 2021-11-29
[homewizard] Fix issue with missing gas values

Daniel-42:
This commit is a bugfix fixes a crash in the HomeWizard add on that happens when the smart meter does not provide gas values. The crash was caused by the empty timestamp.

Signed-off-by: Daniël van Os <daniel@supercell.nl>

### DIFF
--- a/bundles/org.openhab.binding.homewizard/src/main/java/org/openhab/binding/homewizard/internal/HomeWizardHandler.java
+++ b/bundles/org.openhab.binding.homewizard/src/main/java/org/openhab/binding/homewizard/internal/HomeWizardHandler.java
@@ -13,6 +13,9 @@
 package org.openhab.binding.homewizard.internal;
 
 import java.io.IOException;
+import java.time.DateTimeException;
+import java.time.ZoneId;
+import java.time.ZonedDateTime;
 import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.TimeUnit;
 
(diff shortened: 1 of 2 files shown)


bundles/org.openhab.binding.homewizard/src/main/java/org/openhab/binding/homewizard/internal/HomeWizardHandler.java
```diff
@@ -171,30 +174,41 @@ private void pollingCode() {
...
+            long year = dtv + 2000;
+
+            String dateString = String.format("%04d-%02d-%02dT%02d:%02d:%02d", year, month, day, hours, minutes,
+                    seconds);
+            try {
+                DateTimeType dtt = DateTimeType.valueOf(dateString);
+                updateState(HomeWizardBindingConstants.CHANNEL_GAS_TIMESTAMP, dtt);
+                updateState(HomeWizardBindingConstants.CHANNEL_TOTAL_GAS,
+                        new QuantityType<>(payload.getTotalGasM3(), SIUnits.CUBIC_METRE));
+            } catch (java.time.format.DateTimeParseException e) {
```
wborn, line 209: Thanks for the fix! I have one small suggestion, can you import the class instead?

```suggestion
            } catch (DateTimeParseException e) {
```

It would be even better if you'd create a `DateTimeType` using a `ZonedDateTime`, then you don't need to format the values and parse them again, which will save some CPU cycles. :slightly_smiling_face:

Daniel-42: Thanks! I've applied the suggestions.

bundles/org.openhab.binding.homewizard/src/main/java/org/openhab/binding/homewizard/internal/HomeWizardHandler.java
```diff
@@ -171,30 +177,40 @@ private void pollingCode() {
...
-        updateState(HomeWizardBindingConstants.CHANNEL_GAS_TIMESTAMP, dtt);
+            try {
+                DateTimeType dtt = new DateTimeType(
+                        ZonedDateTime.of(year, month, day, hours, minutes, seconds, 0, ZoneId.systemDefault()));
+                updateState(HomeWizardBindingConstants.CHANNEL_GAS_TIMESTAMP, dtt);
+                updateState(HomeWizardBindingConstants.CHANNEL_TOTAL_GAS,
+                        new QuantityType<>(payload.getTotalGasM3(), SIUnits.CUBIC_METRE));
+            } catch (DateTimeException e) {
+                logger.warn("Unable to parse Gas timestamp: {}", payload.getGasTimestamp());
+            }
```
wborn, line 213: Thanks for addressing my comment. :+1: 

You can also remove this try/catch around this code because the `DateTimeException` will no longer be thrown now that the formatted date/time string is no longer parsed

wborn: I now see it can be thrown if a field value of the ZonedDateTime is out of range.

```
     * @throws DateTimeException if the value of any field is out of range, or
     *  if the day-of-month is invalid for the month-year
```

So let's merge it. :slightly_smiling_face:

wborn (approved):
Thanks for the fix. :+1:

PR #11674, opened 2021-11-30
[nuvo] Update thing label for better matching in Add-On search

mlobstein:
Update the thing label name so searching by the model name returns a thing result in the Add-On search page.

### DIFF
--- a/bundles/org.openhab.binding.nuvo/src/main/resources/OH-INF/thing/channels.xml
+++ b/bundles/org.openhab.binding.nuvo/src/main/resources/OH-INF/thing/channels.xml
@@ -6,7 +6,7 @@
 
 	<!-- Nuvo Whole House Amplifier Thing -->
 	<thing-type id="amplifier">
-		<label>Nuvo Whole House Amplifier</label>
+		<label>Grand Concerto or Essentia G Whole House Amplifier</label>
 		<description>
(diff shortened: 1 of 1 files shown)


bundles/org.openhab.binding.nuvo/src/main/resources/OH-INF/thing/channels.xml
```diff
@@ -6,7 +6,7 @@
 
 	<!-- Nuvo Whole House Amplifier Thing -->
 	<thing-type id="amplifier">
-		<label>Nuvo Whole House Amplifier</label>
+		<label>Grand Concerto or Essentia G Whole House Amplifier</label>
```
wborn, line 9: Labels are kept short intentionally:

See: https://www.openhab.org/docs/developer/bindings/thing-xml.html#formatting-labels-and-descriptions

> The label should be short so that for most UIs it does not spread across multiple lines. Guideline is 2-3 words with up to 25 chars.

Maybe it also works by updating the description instead?

mlobstein: It will come up as a binding match based on text in the description but not a thing match. Anyway I moved the new text to the description tag.

wborn (approved):
Thanks!

PR #11675, opened 2021-11-30
[radiothermostat] Update thing label for better matching in Add-On search

mlobstein:
Update the thing label name so searching by the thermostat model name returns a thing result in the Add-On search page.

### DIFF
--- a/bundles/org.openhab.binding.radiothermostat/src/main/resources/OH-INF/thing/thing-types.xml
+++ b/bundles/org.openhab.binding.radiothermostat/src/main/resources/OH-INF/thing/thing-types.xml
@@ -6,9 +6,9 @@
 
 	<!-- RadioThemostat Thing -->
 	<thing-type id="rtherm">
-		<label>CT30, CT50/3M50 or CT80 Wi-Fi Thermostat</label>
+		<label>Thermostat</label>
 		<description>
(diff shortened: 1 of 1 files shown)


bundles/org.openhab.binding.radiothermostat/src/main/resources/OH-INF/thing/thing-types.xml
```diff
@@ -6,7 +6,7 @@
 
 	<!-- RadioThemostat Thing -->
 	<thing-type id="rtherm">
-		<label>Thermostat</label>
+		<label>CT30, CT50/3M50 or CT80 Wi-Fi Thermostat</label>
```
wborn, line 9: Labels are kept short intentionally:

See: https://www.openhab.org/docs/developer/bindings/thing-xml.html#formatting-labels-and-descriptions

> The label should be short so that for most UIs it does not spread across multiple lines. Guideline is 2-3 words with up to 25 chars. Labels should be capitalized using the following rules:

mlobstein: Ditto the comment from #11674

wborn (approved):
Thanks!

PR #11682, opened 2021-12-01
[avmfritz] Exclude other Powerline products from discovery

cweitkamp:
- Exclude other Powerline products from discovery as only 546E is supported by the binding

See https://community.openhab.org/t/support-of-avm-fritz-powerline-540e/128672


marcelrv (approved):
LGTM

wborn (approved):
Thank you!

PR #11689, opened 2021-12-02
[chromecast] Added configuration flag to disable background discovery

cweitkamp:
- Added configuration flag to disable background discovery


jlaur:
Would it make sense to configure it from here?
image

image

cweitkamp:
In general it does make sense. But unfortunately it will not work. The discovery for Chromecast devices is a separate OSGi service decoupled from the binding itself. The location you are pointing to allows to add more configurations for the binding itself - it uses configuration prefix `binding.chromecast`. It is not allowed to reuse this configuration is for another service.

lolodomo (approved):
LGTM

PR #11692, opened 2021-12-02
[androiddebugbridge] Add channels for record events, open urls and doc improvements

GiviMAD:
Minor additions for the androiddebugbridge binding:
* Add the ability to record/send input events.
* Add channel to open url
* Allow start package with activity info.
* Minor doc improvements and fixes.

### DIFF
--- a/bundles/org.openhab.binding.androiddebugbridge/README.md
+++ b/bundles/org.openhab.binding.androiddebugbridge/README.md
@@ -1,7 +1,9 @@
 # Android Debug Bridge Binding
 
-This binding allows to connect to android devices through the adb protocol. 
-The device needs to have **usb debugging enabled** and **allow debugging over tcp**, some devices allow to enable this in the device options but others need a previous connection through adb or even be rooted. 
+This binding allows to connect to android devices through the adb protocol.
+
+The device needs to have **usb debugging enabled** and **allow debugging over tcp**, some devices allow to enable this in the device options but others need a previous connection through adb or even be rooted.
+
 If you are not familiar with adb I suggest you to search "How to enable adb over wifi on \<device name\>" or something like that.
 
 ## Supported Things
@@ -10,7 +12,7 @@ This binding was tested on the Fire TV Stick (android version 7.1.2, volume cont
 
 ## Discovery
 
-As I can not find a way to identify android devices in the network the discovery will try to connect through adb to all the reachable ip in the defined range, you could customize the discovery process through the binding options. **Your device will prop a message requesting you to authorize the connection, you should check the option "Always allow connections from this device" (or something similar) and accept**.
+As I can not find a way to identify android devices in the network the discovery will try to connect through adb to all the reachable ip in the defined range, you could customize the discovery process through the binding options. **Your device will prompt a message requesting you to authorize the connection, you should check the option "Always allow connections from this device" (or something similar) and accept**.
 
 ## Binding Configuration
 
@@ -33,6 +35,7 @@ As I can not find a way to identify android devices in the network the discovery
 | port | int | Device port listening to adb connections (default: 5555) |
 | refreshTime | int | Seconds between device status refreshes (default: 30) |
 | timeout | int | Command timeout in seconds (default: 5) |
+| recordSeconds | int | Record input time in seconds |
 | mediaStateJSONConfig | String | Expects a JSON array. Allow to configure the media state detection method per app. Described in the following section |
 
 ## Media State Detection
@@ -52,6 +55,15 @@ This is a sample of the mediaStateJSONConfig thing configuration:
 
 `[{"name": "com.amazon.tv.launcher", "mode": "idle"},{"name": "org.jellyfin.androidtv", "mode": "wake_lock", "wakeLockPlayStates": [2,3]},{"name": "com.amazon.firetv.youtube", "mode": "wake_lock", "wakeLockPlayStates": [2]}]`
 
+## Record/Send input events
+As the execution of key events takes a while you can use input events an alternative way to control the device you could use input events. As they are pretty device specific you can use the record-input and recorded-input to store/send those events. 
+An example of what you can do:
+* You can send the command `UP` to the `record-input` channel the binding will them capture the events you send through your remote for the defined recordSeconds config for the thing, so press once the UP key on your remote and wait a while.
+* Now that you have recorded your input you can send the `UP` command to the `recorded-input` event and it will send the recorded event to the android device. 
+
+Please note that events could fail if the input method is removed, for example it fails if you clone the events of a bluetooth controller and the remote is offline.
+
+
 ## Channels
 
 | channel  | type   | description                  |
@@ -59,12 +71,17 @@ This is a sample of the mediaStateJSONConfig thing configuration:
 | key-event  | String | Send key event to android device. Possible values listed below |
 | text  | String | Send text to android device |
 | tap  | String | Send tap event to android device (format x,y) |
+| url  | String | Open url in browser |
 | media-volume  | Dimmer | Set or get media volume level on android device |
 | media-control  | Player | Control media on android device |
 | start-package  | String | Run application by package name |
 | stop-package  | String | Stop application by package name |
+| stop-current-package  | String | Stop current application |
 | current-package  | String | Package name of the top application in screen |
+| record-input  | String | Capture events, generate the equivalent command and store it under the provided name |
+| recorded-input  | String | Emulates previously captured input events by name |
 | shutdown  | String | Power off/reboot device (allowed values POWER_OFF, REBOOT) |
+| awake-state  | OnOff | Awake state value. |
 | wake-lock  | Number | Power wake lock value |
 | screen-state  | Switch | Screen power state |
 
--- a/bundles/org.openhab.binding.androiddebugbridge/src/main/java/org/openhab/binding/androiddebugbridge/internal/AndroidDebugBridgeDevice.java
+++ b/bundles/org.openhab.binding.androiddebugbridge/src/main/java/org/openhab/binding/androiddebugbridge/internal/AndroidDebugBridgeDevice.java
@@ -22,6 +22,7 @@ import java.nio.charset.Charset;
 import java.nio.charset.StandardCharsets;
 import java.security.NoSuchAlgorithmException;
 import java.security.spec.InvalidKeySpecException;
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Base64;
 import java.util.concurrent.*;
@@ -55,6 +56,11 @@ public class AndroidDebugBridgeDevice {
     private static final Pattern TAP_EVENT_PATTERN = Pattern.compile("(?<x>\\d+),(?<y>\\d+)");
     private static final Pattern PACKAGE_NAME_PATTERN = Pattern
             .compile("^([A-Za-z]{1}[A-Za-z\\d_]*\\.)+[A-Za-z][A-Za-z\\d_]*$");
+    private static final Pattern ACTIVITY_NAME_PATTERN = Pattern.compile("^[A-Za-z\\.]*");
+    private static final Pattern URL_PATTERN = Pattern.compile(
+            "https?:\\/\\/(www\\.)?[-a-zA-Z0-9@:%._\\+~#=]{2,256}\\.[a-z]{2,4}\\b([-a-zA-Z0-9@:%_\\+.~#?&//=]*)$");
+    private static final Pattern INPUT_EVENT_PATTERN = Pattern
+            .compile("/(?<input>\\S+): (?<n1>\\S+) (?<n2>\\S+) (?<n3>\\S+)$", Pattern.MULTILINE);
 
     private static @Nullable AdbCrypto adbCrypto;
 
@@ -78,6 +84,7 @@ public class AndroidDebugBridgeDevice {
     private String ip = "127.0.0.1";
     private int port = 5555;
     private int timeoutSec = 5;
+    private int recordSeconds;
     private @Nullable Socket socket;
     private @Nullable AdbConnection connection;
     private @Nullable Future<String> commandFuture;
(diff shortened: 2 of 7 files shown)


bundles/org.openhab.binding.androiddebugbridge/src/main/resources/OH-INF/thing/thing-types.xml
```diff
@@ -44,6 +47,11 @@
 				<description>Command timeout seconds.</description>
 				<default>5</default>
 			</parameter>
+			<parameter name="recordSeconds" type="integer" required="true" min="3" max="20">
```
lolodomo, line 50: recordTime looks more appropriate to me. 
And in the description: how much time in seconds the... 
If you agree, it has to be changed at different places including documentation.

bundles/org.openhab.binding.androiddebugbridge/README.md
```diff
@@ -10,7 +12,7 @@ This binding was tested on the Fire TV Stick (android version 7.1.2, volume cont
 
 ## Discovery
 
-As I can not find a way to identify android devices in the network the discovery will try to connect through adb to all the reachable ip in the defined range, you could customize the discovery process through the binding options. **Your device will prop a message requesting you to authorize the connection, you should check the option "Always allow connections from this device" (or something similar) and accept**.
+As I can not find a way to identify android devices in the network the discovery will try to connect through adb to all the reachable ip in the defined range, you could customize the discovery process through the binding options. **Your device will prompt a message requesting you to authorize the connection, you should check the option "Always allow connections from this device" (or something similar) and accept**.
```
lolodomo, line 15: I believe you should end your sentence after "range". And please put each sentence on a different line.

bundles/org.openhab.binding.androiddebugbridge/README.md
```diff
@@ -52,19 +55,33 @@ This is a sample of the mediaStateJSONConfig thing configuration:
 
 `[{"name": "com.amazon.tv.launcher", "mode": "idle"},{"name": "org.jellyfin.androidtv", "mode": "wake_lock", "wakeLockPlayStates": [2,3]},{"name": "com.amazon.firetv.youtube", "mode": "wake_lock", "wakeLockPlayStates": [2]}]`
 
+## Record/Send input events
+As the execution of key events takes a while you can use input events an alternative way to control the device you could use input events. As they are pretty device specific you can use the record-input and recorded-input to store/send those events.
```
lolodomo, line 59: ... as an alternative.. 
Something strange in the first sentence with the end that looks duplucated. 
Please add comma at right places, that is before the 2 "you can use"
Each sentence one line.

bundles/org.openhab.binding.androiddebugbridge/README.md
```diff
@@ -52,19 +55,33 @@ This is a sample of the mediaStateJSONConfig thing configuration:
 
 `[{"name": "com.amazon.tv.launcher", "mode": "idle"},{"name": "org.jellyfin.androidtv", "mode": "wake_lock", "wakeLockPlayStates": [2,3]},{"name": "com.amazon.firetv.youtube", "mode": "wake_lock", "wakeLockPlayStates": [2]}]`
 
+## Record/Send input events
+As the execution of key events takes a while you can use input events an alternative way to control the device you could use input events. As they are pretty device specific you can use the record-input and recorded-input to store/send those events. 
+An example of what you can do:
+* You can send the command `UP` to the `record-input` channel the binding will them capture the events you send through your remote for the defined recordSeconds config for the thing, so press once the UP key on your remote and wait a while.
+* Now that you have recorded your input you can send the `UP` command to the `recorded-input` event and it will send the recorded event to the android device.
```
lolodomo, line 62: Please add a comma after "your input"

bundles/org.openhab.binding.androiddebugbridge/README.md
```diff
@@ -52,19 +55,33 @@ This is a sample of the mediaStateJSONConfig thing configuration:
 
 `[{"name": "com.amazon.tv.launcher", "mode": "idle"},{"name": "org.jellyfin.androidtv", "mode": "wake_lock", "wakeLockPlayStates": [2,3]},{"name": "com.amazon.firetv.youtube", "mode": "wake_lock", "wakeLockPlayStates": [2]}]`
 
+## Record/Send input events
+As the execution of key events takes a while you can use input events an alternative way to control the device you could use input events. As they are pretty device specific you can use the record-input and recorded-input to store/send those events. 
+An example of what you can do:
+* You can send the command `UP` to the `record-input` channel the binding will them capture the events you send through your remote for the defined recordSeconds config for the thing, so press once the UP key on your remote and wait a while.
```
lolodomo, line 61: ... will then... 
Please add commas

lolodomo (changes requested):

bundles/org.openhab.binding.androiddebugbridge/src/main/java/org/openhab/binding/androiddebugbridge/internal/AndroidDebugBridgeDevice.java
```diff
@@ -258,6 +315,31 @@ private VolumeInfo getVolume(int stream) throws AndroidDebugBridgeDeviceExceptio
...
+            throws AndroidDebugBridgeDeviceException, InterruptedException, TimeoutException, ExecutionException {
+        String out = runAdbShell(recordDuration * 2, "getevent", "&", "sleep", Integer.toString(recordDuration), "&&",
+                "exit");
+        var matcher = INPUT_EVENT_PATTERN.matcher(out);
+        var commandList = new ArrayList<String>();
+        while (matcher.find()) {
+            String inputPath = matcher.group("input");
+            int n1 = Integer.parseInt(matcher.group("n1"), 16);
+            int n2 = Integer.parseInt(matcher.group("n2"), 16);
+            int n3 = Integer.parseInt(matcher.group("n3"), 16);
```
lolodomo, line 328: Maybe you should catch parse exception

bundles/org.openhab.binding.androiddebugbridge/src/main/java/org/openhab/binding/androiddebugbridge/internal/AndroidDebugBridgeHandler.java
```diff
@@ -170,7 +175,51 @@ private void handleCommandInternal(ChannelUID channelUID, Command command)
...
         }
+        String recordPropertyName = getRecordPropertyName(recordName);
+        logger.debug("RECORD: {}", recordPropertyName);
+        if (!recordPropertyName.isEmpty() && !recordPropertyName.equals("NULL")) {
+            var eventCommand = adbConnection.recordInputEvents();
+            if (eventCommand.isEmpty()) {
+                logger.debug("No events recorded");
+                if (this.getThing().getProperties().containsKey(recordPropertyName)) {
+                    this.getThing().setProperty(recordPropertyName, null);
+                    updateProperties(editProperties());
```
lolodomo, line 207: Why not using updateProperty rather than these people lines?

bundles/org.openhab.binding.androiddebugbridge/src/main/resources/OH-INF/thing/thing-types.xml
```diff
@@ -44,6 +47,11 @@
 				<description>Command timeout seconds.</description>
 				<default>5</default>
 			</parameter>
+			<parameter name="recordDuration" type="integer" required="true" min="3" max="20">
+				<label>Record Seconds</label>
```
lolodomo, line 51: Record Duration

lolodomo (changes requested):

bundles/org.openhab.binding.androiddebugbridge/src/main/java/org/openhab/binding/androiddebugbridge/internal/AndroidDebugBridgeDevice.java
GiviMAD: I don't think is needed because the regex is quite explicit and the service will not return other values in that place other than alphanumerical numbers. 
Either way I can add it if you want but wrapping the complete while because the command recorded have no sense if we miss a part. 
Let me know what I should do.

lolodomo: Your pattern matching does not guarantee that your string will match a hexa number. This can be any string without white spaces.
So either your update your pattern to be sure to match a hexa number and you do not need catching exception, or your keep your pattern unchallenged and it will be safer to catch exception.

bundles/org.openhab.binding.androiddebugbridge/src/main/java/org/openhab/binding/androiddebugbridge/internal/AndroidDebugBridgeHandler.java
lolodomo: Looking at core framework code, I have the feeling that it should work even with a null value. You can try ignoring the compiler warning and see if it works. 
Other and maybe better option would be to set the value to an empty string.

GiviMAD: It gives me a compiler error when I pass a null value to the updateProperty fn. And if I pass and empty string, the property doesn't get removed which is what I want.
So I can not find other way to remove the property.

bundles/org.openhab.binding.androiddebugbridge/src/main/java/org/openhab/binding/androiddebugbridge/internal/AndroidDebugBridgeDevice.java
GiviMAD: I have added a NumberFormatException catch around the while loop.

bundles/org.openhab.binding.androiddebugbridge/src/main/java/org/openhab/binding/androiddebugbridge/internal/AndroidDebugBridgeHandler.java
GiviMAD: Even removing the property instead of setting if to null doesn't work. Setting it to null seems to be the only way to get it removed from the thing correctly.

GiviMAD: @lolodomo, Is it ok to leave it as is?

lolodomo: No at least until I understand and validate your strange code and understand a simple and normal call is not possible. 
I will download your branch and see.

GiviMAD: Sure.
I think the problem itself is in the BaseThingHandler class it makes no sense to make the value non nullable on the updateProperty function and all the related ones when the updateProperty method in the thing accepts null. 
So I'm having to bypass this by calling the editProperties method to get and instance of a non nullable map which contains in fact a null value, or at least that's what I think is happening.
Either way I'll open and issue on the core with this info to see if I'm missing something.
Let me know if you find something and thanks for the review.

lolodomo: Why not simply removing the property? 
It makes no real sense to set a property to NULL.

bundles/org.openhab.binding.androiddebugbridge/src/main/java/org/openhab/binding/androiddebugbridge/internal/AndroidDebugBridgeHandler.java
```diff
@@ -170,7 +175,51 @@ private void handleCommandInternal(ChannelUID channelUID, Command command)
...
+    private void recordDeviceInput(Command recordNameCommand)
+            throws AndroidDebugBridgeDeviceException, InterruptedException, TimeoutException, ExecutionException {
+        var recordName = recordNameCommand.toFullString();
+        if (!RECORD_NAME_PATTERN.matcher(recordName).matches()) {
+            logger.warn("Invalid record name, accepts alphanumeric values with '_'.");
+            return;
         }
+        String recordPropertyName = getRecordPropertyName(recordName);
+        logger.debug("RECORD: {}", recordPropertyName);
+        if (!recordPropertyName.isEmpty() && !recordPropertyName.equals("NULL")) {
```
lolodomo, line 201: I believe this test is useless as the condition will always be true.
No?

lolodomo (changes requested):

bundles/org.openhab.binding.androiddebugbridge/src/main/java/org/openhab/binding/androiddebugbridge/internal/AndroidDebugBridgeHandler.java
GiviMAD: I already tried that also but it makes non effect. 
But event if that had worked, I need to call the updateProperties/updateProperty function for the changes to be persisted because internal they notify the binding core that there is property changes. That's what I discover after I saw my the properties disappear after a reboot.

GiviMAD: I discovered that I have to set the properties to null to delete them reading the forum, documentation or a comment on other binding, I can't remember, I agree it's weird. A deleteProperty method on the BaseThingHandler class will be great.

lolodomo: My idea was to use the code you proposed in your core issue but you're right, it cannot work. 
The solution would be to add a method `removeProperty` in the core framework. 
So I see no proper solution now. If your solution is working, that's fine for me.

lolodomo (approved):
LGTM

lolodomo:
@GiviMAD : sorry for the delay I needed to understand your strange code to delete a thing property. Unfortunately I can't merge this PR until 3.2 is released, except if @kaikreuzer accepts a derogation.

PR #11693, opened 2021-12-03
[jsscripting] Added simple Shared Cache into scope

jpg0:
Add a very basic shared cache into the JS add-on to allow JS scripts to cache objects between runs and/or scripts.

### DIFF
--- a/bundles/org.openhab.automation.jsscripting/src/main/java/org/openhab/automation/jsscripting/internal/scope/SharedCache.java
+++ b/bundles/org.openhab.automation.jsscripting/src/main/java/org/openhab/automation/jsscripting/internal/scope/SharedCache.java
@@ -1,0 +1,101 @@
+/**
+ * Copyright (c) 2010-2021 Contributors to the openHAB project
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package org.openhab.automation.jsscripting.internal.scope;
+
+import java.util.Collection;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Set;
+import java.util.function.Supplier;
+
+import org.eclipse.jdt.annotation.NonNullByDefault;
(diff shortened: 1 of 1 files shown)


jpg0:
Can be used like this:

```
let counter = require("@runtime").sharedcache.getOrCreate("counter", () => ({"times":0}))

counter.times = counter.times + 1
 
Java.type('org.slf4j.LoggerFactory').getLogger('script.js.cachetest').warn("count: " + counter.times); 
```

Which will increment the counter each time the script is run.

bundles/org.openhab.automation.jsscripting/src/main/java/org/openhab/automation/jsscripting/internal/scope/SharedCache.java
```diff
@@ -0,0 +1,97 @@
...
+
+        public void put(String k, Object v) {
+            backingMap.put(k, v);
+        }
+
+        public @Nullable Object get(String k) {
+            return backingMap.get(k);
+        }
+
+        public @Nullable Object getOrCreate(String k, Supplier<Object> supplier) {
```
digitaldan, line 93: Any reason this would not just overload the `get` call?  This would seem more natural to use and a bit nicer syntax in scripts.  Not sure if overloaded methods poses a problem with JS -> Java invocation , but it seems to work nicely in my tests.

digitaldan:
Should we also expose `remove(Object key)` ?

bundles/org.openhab.automation.jsscripting/src/main/java/org/openhab/automation/jsscripting/internal/scope/SharedCache.java
jpg0: I guess I did it this way because it is named differently in Java, but happy to overload instead of you prefer.

jpg0:
> Should we also expose `remove(Object key)` ?  

Yes I think that's sensible.

bundles/org.openhab.automation.jsscripting/src/main/java/org/openhab/automation/jsscripting/internal/scope/SharedCache.java
digitaldan: I feel like it would fit better with JavaScript patterns where arguments tend to be optional , also it's less verbose which make the scripts look nice :-)

digitaldan (approved):
LGTM!

lolodomo (approved):
LGTM

PR #11694, opened 2021-12-03
[chromecast] Added pattern for DateTime channel types

cweitkamp:
- Added pattern for DateTime channel types

### DIFF
--- a/bundles/org.openhab.binding.chromecast/src/main/resources/OH-INF/thing/thing-types.xml
+++ b/bundles/org.openhab.binding.chromecast/src/main/resources/OH-INF/thing/thing-types.xml
@@ -244,7 +244,7 @@
 		<item-type>DateTime</item-type>
 		<label>Broadcast Date</label>
 		<description>The broadcast date of the currently playing media</description>
-		<category>time</category>
+		<category>Calendar</category>
 		<state readOnly="true" pattern="%1$tY-%1$tm-%1$td"/>
 	</channel-type>
 
@@ -259,7 +259,7 @@
 		<item-type>DateTime</item-type>
 		<label>Creation Date</label>
 		<description>The creation date of the currently playing media</description>
-		<category>time</category>
+		<category>Calendar</category>
 		<state readOnly="true" pattern="%1$tY-%1$tm-%1$td"/>
 	</channel-type>
 
@@ -302,7 +302,7 @@
 		<item-type>DateTime</item-type>
 		<label>Release Date</label>
 		<description>The release date of the currently playing media</description>
-		<category>time</category>
+		<category>Calendar</category>
 		<state readOnly="true" pattern="%1$tY-%1$tm-%1$td"/>
 	</channel-type>
 
--- a/bundles/org.openhab.binding.chromecast/src/main/resources/OH-INF/thing/thing-types.xml
+++ b/bundles/org.openhab.binding.chromecast/src/main/resources/OH-INF/thing/thing-types.xml
@@ -244,7 +244,8 @@
 		<item-type>DateTime</item-type>
 		<label>Broadcast Date</label>
 		<description>The broadcast date of the currently playing media</description>
-		<state readOnly="true"/>
+		<category>time</category>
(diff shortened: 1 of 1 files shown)


bundles/org.openhab.binding.chromecast/src/main/resources/OH-INF/thing/thing-types.xml
```diff
@@ -244,7 +244,8 @@
 		<item-type>DateTime</item-type>
 		<label>Broadcast Date</label>
 		<description>The broadcast date of the currently playing media</description>
-		<state readOnly="true"/>
+		<category>time</category>
```
lolodomo, line 247: The first letter is always in uppercase.
I don't know if this is a problem with a lowercase first letter.
Can you update to be consistent with all other bindings ?
And by the way, `Date` looks more appropriate.

lolodomo (changes requested):

bundles/org.openhab.binding.chromecast/src/main/resources/OH-INF/thing/thing-types.xml
cweitkamp: The category is used to pick the default icon rendered in UIs. So I do not see a reason to write it in camel case.

We do not have an icon for "date". What I can offer is "calendar".

lolodomo: I have a doubt, I don"t find `Date` in the documentation
https://www.openhab.org/docs/concepts/categories.html#channel-categories
But `Date` as channel category is used for example in the NTP binding.

cweitkamp: > > The first letter is always in uppercase.

> So I do not see a reason to write it in camel case.

Hm I see you are right. I will change it.

lolodomo: Regarding `Date` as channel category, we have to fix the MI IO and the NTP bindings:
```
$ find ./*/src -name "*.xml" -exec grep -H "<category>Date" {} \;
./org.openhab.binding.miio/src/main/resources/OH-INF/thing/vacuumThing.xml:             <category>Date</category>
./org.openhab.binding.miio/src/main/resources/OH-INF/thing/vacuumThing.xml:             <category>Date</category>
./org.openhab.binding.ntp/src/main/resources/OH-INF/thing/thing-types.xml:              <category>Date</category>
./org.openhab.binding.ntp/src/main/resources/OH-INF/thing/thing-types.xml:              <category>Date</category>
```

cweitkamp: I changed the category to "Calendar".

bundles/org.openhab.binding.chromecast/src/main/resources/OH-INF/thing/thing-types.xml
```diff
@@ -244,7 +244,7 @@
 		<item-type>DateTime</item-type>
 		<label>Broadcast Date</label>
 		<description>The broadcast date of the currently playing media</description>
-		<category>time</category>
+		<category>Calendar</category>
```
lolodomo, line 247: `Calendar` is not listed as a valid channel category in the documentation: https://www.openhab.org/docs/concepts/categories.html#channel-categories
Searching this channel category in all bindings, it is not found.
In case it is really a valid category, at least the documentation should be updated.

lolodomo: Icon is available in the classic iconset: https://github.com/openhab/openhab-webui/blob/main/bundles/org.openhab.ui.iconset.classic/src/main/resources/icons/calendar.png
Documentation is just outdated !
So using it as channel category is I guess valid.

lolodomo (approved):

PR #11698, opened 2021-12-04
[hdpowerview] Fix secondary position bug. Add shade database and properties.

### LINKED ISSUE #11557: [hdpowerview] Setting primary rail position of dual-rail shade causes secondary rail position to be undefined
## Expected Behavior
Moving/setting the primary rail of a dual-rail shade should not change the binding's state for the secondary rail position.

## Current Behavior
When the primary rail of a dual-rail shade is moved (either using up/down commands or by sending a PercentType command), the secondary rail position becomes UNDEF within OpenHAB. As expected, the secondary rail itself does not move. However, the secondary rail state is lost. Eventually, the binding will tell the hub to do a "hard refresh" of each shade, and then the secondary rail state will be properly restored.

This bug seems to be due in part to a quirk of the Powerview Hub. When a shade position command is PUT to the hub, it seems to cache that exact position object, and returns it on future polling. So, if a command is sent with only the primary rail position set, the hub will return only the primary rail position with each poll until the hard refresh is performed.

I captured a TRACE level log showing this behavior. To make the logging a bit easier, I disabled the Thing for all but one of my shades, so only shade ID 58131 is getting updated. In addition, to speed up the state changes, I set the hard refresh interval to 5 minutes. I also set the polling interval to 2 minutes to reduce the number of log lines. The full log is attached ([powerview_secondary_undef_trace.txt](https://github.com/openhab/openhab-addons/files/7517630/powerview_secondary_undef_trace.txt)), but an annotated and truncated version is included below. Also, this log was captured using the JAR file that fixed issue #11543, but I confirmed that this bug existed in the 3.1.0 version of the add-on.

I started capturing the log the log after a hard refresh. I then issued a command to move the secondary rail (`LivingRoom_Shade5_Top`) to position 49. This worked as expected, and the JSON response included both rail positions.
```
21:38:27.755 [TRACE] [erview.internal.HDPowerViewWebTargets] - API command GET http://192.168.1.104/api/shades/58131
21:38:30.910 [TRACE] [erview.internal.HDPowerViewWebTargets] - JSON response = {"shade":{"id":58131,"type":9,"capabilities":7,"batteryKind":2,"smartPowerSupply":{"status":0,"id":0,"port":0},"batteryStatus":3,"batteryStrength":158,"roomId":1877,"firmware":{"revision":2,"subRevision":3,"build":3147,"index":3},"name":"TGl2aW5nIFJvb20gNQ==","groupId":27580,"signalStrength":4,"positions":{"posKind1":1,"position1":0,"posKind2":2,"position2":65535},"timedOut":false}}
21:38:47.931 [INFO ] [openhab.event.ItemCommandEvent       ] - Item 'LivingRoom_Shade5_Top' received command 49
21:38:47.933 [INFO ] [openhab.event.ItemStatePredictedEvent] - Item 'LivingRoom_Shade5_Top' predicted to become 49
21:38:47.934 [TRACE] [erview.internal.HDPowerViewWebTargets] - API command GET http://192.168.1.104/api/shades/58131
21:38:47.940 [INFO ] [openhab.event.ItemStateChangedEvent  ] - Item 'LivingRoom_Shade5_Top' changed from 100 to 49
21:38:47.984 [TRACE] [erview.internal.HDPowerViewWebTargets] - JSON response = {"shade":{"id":58131,"type":9,"capabilities":7,"batteryKind":2,"smartPowerSupply":{"status":0,"id":0,"port":0},"batteryStatus":3,"batteryStrength":158,"roomId":1877,"firmware":{"revision":2,"subRevision":3,"build":3147,"index":3},"name":"TGl2aW5nIFJvb20gNQ==","groupId":27580,"signalStrength":4,"positions":{"posKind1":1,"position1":0,"posKind2":2,"position2":65535}}}
21:38:47.987 [TRACE] [erview.internal.HDPowerViewWebTargets] - API command PUT http://192.168.1.104/api/shades/58131
21:38:47.988 [TRACE] [erview.internal.HDPowerViewWebTargets] - JSON command = {"shade":{"id":58131,"positions":{"posKind1":1,"position1":0,"posKind2":2,"position2":32112}}}
21:38:49.625 [TRACE] [erview.internal.HDPowerViewWebTargets] - JSON response = {"shade":{"id":58131,"type":9,"capabilities":7,"batteryKind":2,"smartPowerSupply":{"status":0,"id":0,"port":0},"batteryStatus":3,"batteryStrength":158,"roomId":1877,"firmware":{"revision":2,"subRevision":3,"build":3147,"index":3},"name":"TGl2aW5nIFJvb20gNQ==","groupId":27580,"signalStrength":4,"positions":{"posKind1":1,"position1":0,"posKind2":2,"position2":32112}}}
```
I then issued a command to move the primary rail (`LivingRoom_Shade5_Bottom`) to position 80. This time, only a primary rail position was sent, and only a primary rail position was received in the JSON response:
```
21:39:02.785 [INFO ] [openhab.event.ItemCommandEvent       ] - Item 'LivingRoom_Shade5_Bottom' received command 80
21:39:02.787 [INFO ] [openhab.event.ItemStatePredictedEvent] - Item 'LivingRoom_Shade5_Bottom' predicted to become 80
21:39:02.789 [TRACE] [erview.internal.HDPowerViewWebTargets] - API command PUT http://192.168.1.104/api/shades/58131
21:39:02.792 [TRACE] [erview.internal.HDPowerViewWebTargets] - JSON command = {"shade":{"id":58131,"positions":{"posKind1":1,"position1":13107}}}
21:39:02.797 [INFO ] [openhab.event.ItemStateChangedEvent  ] - Item 'LivingRoom_Shade5_Bottom' changed from 100 to 80
21:39:04.425 [TRACE] [erview.internal.HDPowerViewWebTargets] - JSON response = {"shade":{"id":58131,"type":9,"capabilities":7,"batteryKind":2,"smartPowerSupply":{"status":0,"id":0,"port":0},"batteryStatus":3,"batteryStrength":158,"roomId":1877,"firmware":{"revision":2,"subRevision":3,"build":3147,"index":3},"name":"TGl2aW5nIFJvb20gNQ==","groupId":27580,"signalStrength":4,"positions":{"posKind1":1,"position1":13107}}}
```
Then, the binding polled for state. For simplicity, I'm truncating the JSON response of all 11 shades to show only 2 shades: the shade with id 58131 that was just moved and another shade that was not recently moved for comparison. Shade 58131 is only returning the primary rail position that was just set, not both rails, and the update leads to the secondary rail (`LivingRoom_Shade5_Top`) being set to UNDEF. I've removed some lines for simplicity, but see attached for full log:
```
21:39:23.009 [DEBUG] [nternal.handler.HDPowerViewHubHandler] - Polling for state
21:39:23.010 [TRACE] [erview.internal.HDPowerViewWebTargets] - API command GET http://192.168.1.104/api/shades/
21:39:23.122 [TRACE] [erview.internal.HDPowerViewWebTargets] - JSON response = {"shadeIds":[44940,58131,10734,35688,49916,40481,13306,43539,48128,57893,57040],"shadeData":[{"id":44940,"type":8,"capabilities":7,"batteryKind":2,"smartPowerSupply":{"status":0,"id":0,"port":0},"batteryStatus":3,"batteryStrength":161,"roomId":23864,"firmware":{"revision":2,"subRevision":3,"build":3147,"index":3},"name":"TWFpbiBCZWRyb29tIDU=","groupId":24609,"signalStrength":4,"positions":{"posKind1":1,"position1":0,"posKind2":2,"position2":19911}},{"id":58131,"type":9,"capabilities":7,"batteryKind":2,"smartPowerSupply":{"status":0,"id":0,"port":0},"batteryStatus":3,"batteryStrength":158,"roomId":1877,"firmware":{"revision":2,"subRevision":3,"build":3147,"index":3},"name":"TGl2aW5nIFJvb20gNQ==","groupId":27580,"signalStrength":4,"positions":{"posKind1":1,"position1":13107}},<RESPONSE TRUNCATED>]}
21:39:23.136 [DEBUG] [nternal.handler.HDPowerViewHubHandler] - Received data for 11 shades
<SNIP>
21:39:23.152 [DEBUG] [nternal.handler.HDPowerViewHubHandler] - Updating shade '58131'
<SNIP>
21:39:23.173 [INFO ] [openhab.event.ItemStateChangedEvent  ] - Item 'LivingRoom_Shade5_Top' changed from 49 to UNDEF
```
Additional polling continues to return only the primary position for this shade. Eventually, a hard refresh is triggered, and the secondary rail position is once again returned on the next polling. Response and logs are truncated as above for simplicity:
```
21:43:27.762 [TRACE] [erview.internal.HDPowerViewWebTargets] - API command GET http://192.168.1.104/api/shades/58131
21:43:30.909 [TRACE] [erview.internal.HDPowerViewWebTargets] - JSON response = {"shade":{"id":58131,"type":9,"capabilities":7,"batteryKind":2,"smartPowerSupply":{"status":0,"id":0,"port":0},"batteryStatus":3,"batteryStrength":158,"roomId":1877,"firmware":{"revision":2,"subRevision":3,"build":3147,"index":3},"name":"TGl2aW5nIFJvb20gNQ==","groupId":27580,"signalStrength":4,"positions":{"posKind1":1,"position1":13155,"posKind2":2,"position2":32062},"timedOut":false}}
<SNIP>
21:45:23.774 [DEBUG] [nternal.handler.HDPowerViewHubHandler] - Polling for state
21:45:23.775 [TRACE] [erview.internal.HDPowerViewWebTargets] - API command GET http://192.168.1.104/api/shades/
21:45:23.838 [TRACE] [erview.internal.HDPowerViewWebTargets] - JSON response = {"shadeIds":[44940,58131,10734,35688,49916,40481,13306,43539,48128,57893,57040],"shadeData":[{"id":44940,"type":8,"capabilities":7,"batteryKind":2,"smartPowerSupply":{"status":0,"id":0,"port":0},"batteryStatus":3,"batteryStrength":161,"roomId":23864,"firmware":{"revision":2,"subRevision":3,"build":3147,"index":3},"name":"TWFpbiBCZWRyb29tIDU=","groupId":24609,"signalStrength":4,"positions":{"posKind1":1,"position1":0,"posKind2":2,"position2":19911}},{"id":58131,"type":9,"capabilities":7,"batteryKind":2,"smartPowerSupply":{"status":0,"id":0,"port":0},"batteryStatus":3,"batteryStrength":158,"roomId":1877,"firmware":{"revision":2,"subRevision":3,"build":3147,"index":3},"name":"TGl2aW5nIFJvb20gNQ==","groupId":27580,"signalStrength":4,"positions":{"posKind1":1,"position1":13155,"posKind2":2,"position2":32062}},<RESPONSE TRUNCATED>]}
21:45:23.845 [DEBUG] [nternal.handler.HDPowerViewHubHandler] - Received data for 11 shades
<SNIP>
21:45:23.848 [DEBUG] [nternal.handler.HDPowerViewHubHandler] - Updating shade '58131'
21:45:23.852 [INFO ] [openhab.event.ItemStateChangedEvent  ] - Item 'LivingRoom_Shade5_Top' changed from UNDEF to 49
```

## Possible Solution
I think there are a few approaches to fix this. One workaround would be to increase the hard refresh frequency, but hard refreshes are expensive/slow because they involve RF communications with all shades on the network. Another option would be to perform a hard refresh before polling whenever the binding knows a shade has recently been moved, but I'm not sure how easy that would be to implement. The approach that seems cleanest to me is to make setting the primary rail similar to setting the secondary rail in this shade handler code:
https://github.com/openhab/openhab-addons/blob/062f45440663b431d6213d2854c17027787c9269/bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/handler/HDPowerViewShadeHandler.java#L204-L227
The binding could first get the current state of the shade regardless of which rail was being moved, then only update the position of the desired rail and send the move shade command with both positions set. This would ensure the hub keeps the state of both rails in its cache. In the case of single rail shades, I think the get operation would already have the secondary rail position recorded as null, so updating the primary rail position and resending should still work, but I don't have any single rail shades to test that on.

1. Set the primary rail position of a dual-rail shade
2. Wait for the next hub polling to take place
3. Check the secondary rail position, which will be UNDEF until the next hard refresh or secondary rail command

## Context
I have been setting up slider UIs for my dual-rail shades, so that I have vertical sliders for the top and bottom rails that visually show their physical position. When I use the slider to change the bottom (primary) rail, the slider for the top (secondary) rail will move to 0 (the location when the value is UNDEF), so the UI no longer reflects the physical position of my shades.

## Your Environment
* Version used: OpenHAB 3.1.0; HD PowerView Binding 3.1.0 (logs collected with version 3.2.0-SNAPSHOT from #11543)

[powerview_secondary_undef_trace.txt](https://github.com/openhab/openhab-addons/files/7517630/powerview_secondary_undef_trace.txt)

### LINKED ISSUE #11593: [hdpowerview] Resolve differences between OH dual-rail shade behavior, HD remote behavior, and documentation
After fixing issue #11543 in PR #11552, the behavior of the secondary channel for dual-rail shades no longer matches the documentation in the README. However, there is also confusion about what the "correct" behavior should be for dual-rail shades, which is not helped by HunterDouglas's poorly worded documentation. This issue is to continue the discussion with @andrewfg and @jlaur that began in #11543 after that issue was closed. The goal is to try to resolve the differences between OpenHAB's behavior with dual-rail shades, the HunterDouglas Pebble Remote behavior, and the OpenHAB documentation.

I think there are a few questions that should be addressed, but I hope others will help to refine this list:
1. How to map PowerView Hub raw coordinates for the primary and secondary rails to OpenHAB coordinates? That is, should both rails exist in the SAME OpenHAB coordinates with the top of the window always 0% and bottom always 100% (v3.1.0 setter behavior), or should the primary and secondary rails exist in INVERTED OpenHAB coordinates (v3.1.0 getter behavior)? See #11543 for description of setter vs getter behavior in v3.1.0.

2. How should OpenHAB's UP/DOWN commands be mapped to primary and secondary rail motion?

3. How to define the OPEN and CLOSED state for the primary and secondary rails?

4. How to ensure that the dual-rail shade behavior is still compatible with venetian blinds?

Since I found it hard to understand HunterDouglas's manuals and [PowerView API documentation](https://github.com/openhab/openhab-addons/files/3479287/PowerView.HUB.Over.IP.pdf), I made some visuals for myself to help map the coordinate systems used for single and dual-rail shades. In this first single-rail figure, grey indicates shade fabric, and the heavy line is the primary (single) rail. Here are how the raw Powerview Hub coordinates and OpenHAB coordinates map for single-rail shades, based on the HDPowerview add-on version 3.1.0 (and unaffected by PR #11552):
image

Since I struggled to get this straight in my mind, "bottom-up" shades are traditional single-rail shades, where the fabric is fixed at the top of the window and the bottom (primary) rail can be moved up to open the shade. "Top-down" shades are single rail shades where the fabric is fixed at the bottom of the window and the top (primary) rail can be moved down to open the shade. The single rail in both of these shades exists in the same hub coordinate space, but the open vs closed coordinates differ because of where the fabric is located. Hopefully the above diagrams help explain this very confusing text from the PowerView API docs: "Top-down shades are in the same coordinate space as bottom-up shades.  Shade  position  values  for  top-down  shades  would  be  reversed  for  bottom-up  shades.  For example, since 65535 is the open value for a bottom-up shade, it is the closed value for a top-down shade." 

Next, using the version of the hdpowerview add-on from PR #11552, I made a similar figure for my dual-rail shades at various states. In this figure, grey again indicates the shade fabric, the heavy black line is the *primary* rail, and the heavy red line is the *secondary* rail (when at the same location, only the primary rail is shown). The primary rail coordinate systems are shown on the left in black, and the secondary rail coordinate systems are shown on the right in red. For each shade position (labeled A-H at the bottom), the corresponding raw and OpenHAB coordinates for both rails are provided:
image

While the dual-rail shades have a single Closed state, they have two Open states (unlike single-rail shades, which have a single open state). I think this makes it hard to define "OPEN" in OpenHAB.

The HunterDouglas manual has this image describing the remote operation for dual-rail shades that I have found confusing:
image

To help me understand how the HunterDouglas Remote works to control dual-rail shades, and so that I could wrap my head around how the OpenHAB commands should behave to mimic it, I set my dual-rail shades to 6 different starting positions that seemed to correspond to all the possible "equivalent" shade states (OH coordinates are as in the previous figure). I then pushed each of the 4 buttons on the remote and wrote down what happened with each rail in the table below. I reset the shade to the initial position before each button push, and waited until the shade fully stopped moving, as sometimes one rail moves and then another. The buttons are color coded by whether they are associated with the primary rail (black) or secondary rail (red), and the rail effects are similarly color coded. In *italics* I noted any time a rail seemed to move in the opposite direction of the arrow on the button that I pushed. Hopefully this is helpful for understanding these dual-rail shades and the remote.
image

From the first table, I came up with a summary of how the buttons work (second table above). As @andrewfg has pointed out, pushing an up arrow does not always mean a rail moves up, as the secondary rail's up arrow can move the primary rail down (and vice versa, the primary rail's down arrow can move the secondary rail up). However, unfortunately Up does not always mean Open, either. The primary "Up arrow / Open" button always means Open, but the secondary "right button with up arrow" means Close. That really confused me, and was not what I was expecting.

Based on the above analysis, these are my suggestions for how OpenHAB should behave to make it match the HD Remote, and the README could be updated to match. @andrewfg and @jlaur, please let me know what you think and whether this makes sense, or whether there should be other behavior. It would also be great to get feedback from @berland and @mstroeve, who were involved in testing the original dual-rail code.

Suggestions:
- OpenHAB Items should always move the channel they are bound to in the canonical direction (ie, sending UP to primary should move the bottom rail up, while sending UP to secondary should move the top rail up). This seems to be current behavior, but does not match what is documented in the README. This current behavior also matches how the remote buttons work.
- OpenHAB Items *could* move the other channel in the opposite direction to fully mimic the HD PowerView remote. Currently, this is not implemented (there is no cross-talk between the primary and secondary channels in the binding). My preference would be for this behavior to be optional at the Thing level, since personally I dislike the cross-talk present in the PowerView remote.
- I prefer to keep both rails in the same OpenHAB coordinate space, as they are with PR #11552. This would be useful for a dual-slider UI to set the position of both rails, and also seems intuitive to me. However, this would be a change from v3.1.0 "getter" behavior. It might also impact how OPEN and CLOSED states are determined.

Once we agree on behavior, I'm happy to help test out the code to implement it. Also, my suggestions are only for how the primary and secondary channels should behave with dual-rail shades. I don't have any venetian blinds, so I couldn't test that to try to understand how they might fit into the above model. One thing that is not clear to me with venetian blinds is whether they only use position1 in the hub JSON or whether they can be controlled with both position1 and position2 (ie, could position2 be used to control the vane, while position1 always controls the primary rail for single-rail, dual-rail, and venetian blinds?).

Sorry for cramming so much into a single issue, but I thought it would be helpful to gather everything in one place.

andrewfg:
### Background

Issue #11557 reports a bug where, on i) dual action shades, if ii) the Primary position is changed, then iii) the Secondary position becomes undefined, until iv) the next hard refresh occurs.

In addition while exploring solutions to this bug, the following additional issues were identified in the existing code..
- It doesn't identify shade types and their respective specific capabilities.
- It doesn't handle the case where primary and secondary positions are reversed in the hub JSON.
- It doesn't prevent the lower rail from trying to move above the upper rail, or vice-versa.
- After a shade move, the position Channels aren't updated until the next soft refresh.

### Solution

This PR makes the following changes..
1. Fixes #11557.
2. Adds a shade type and capabilities database to identify known shade types and their respective specific capabilities.
3. Produces `logger.warn()` messages when a shade with new (unknown) type or capabilities is encountered; prompting maintainers to update/extend the code to support it.
4. Produces `logger.warn()` messages when the JSON returned by the hub seems incompatible with the shade capabilities; prompting maintainers of potential errors in the known types / capabilities database.
5. Added shade Thing properties so that shade type and capabilities are visible to the user.
6. Handle the case where primary and secondary positions are reversed in the JSON.
7. Prevent the lower rail from trying to move above the upper rail, or vice-versa.
8. After a shade move, the position Channels are updated immediately.
9. Updated / extended JUnit tests to test the above new features.
10. Updated JUnit tests to eliminate MVN compiler potential null warnings.
11. Updates the ReadMe so that it fixes #11593.

### Acknowledgements

Many thanks to @arroyoj and @jlaur for testing, feedback and support.


bundles/org.openhab.binding.hdpowerview/src/test/java/org/openhab/binding/hdpowerview/HDPowerViewJUnitTests.java
```diff
@@ -286,12 +272,62 @@ public void testOnlineCommunication() {
     }
 
     /**
-     * Test generic JSON shades response
+     * Test parsing of ShadePosition.
+     *
+     */
+    @Test
+    public void testShadePositionParsing() {
```
jlaur, line 279: I propose to split this method into separate methods, one for each distinct testcase.

jlaur:
@andrewfg - I suppose file **bundles/org.openhab.binding.hdpowerview/org.openhab.binding.hdpowerview-3.2.0-SNAPSHOT.jar** was added by mistake?

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/api/ShadePosition.java
```diff
@@ -190,56 +154,196 @@ private State getPosition1(CoordinateSystem coordSys) {
...
+                        setPosition1(shadeCapabilities, posKindCoords, percent);
+                    } else if (shadeCapabilities.supportsTiltAnywhere()) {
+                        setPosition2(shadeCapabilities, posKindCoords, percent);
+                    }
+                } else if (shadeCapabilities.supportsTiltAnywhere()) {
+                    setPosition1(shadeCapabilities, posKindCoords, percent);
+                }
+                break;
+
             default:
```
jlaur, line 345: Remove empty `default`.

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/database/ShadeCapabilitiesDatabase.java
```diff
@@ -0,0 +1,350 @@
...
+        @Override
+        public String toString() {
+            return String.format("%d ( %s )", intValue, text);
+        }
+    }
+
+    /**
+     * Describes a shade type entry in the database; implements 'capabilities' parameter.
+     *
+     * @author AndrewFG - Initial contribution
```
jlaur, line 98: Full name missing here, but I think author tag is only needed on file level?

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/HDPowerViewBindingConstants.java
```diff
@@ -49,6 +49,11 @@
     public static final String CHANNELTYPE_SCENE_ACTIVATE = "scene-activate";
     public static final String CHANNELTYPE_SCENE_GROUP_ACTIVATE = "scene-group-activate";
 
+    public static final String PROPERTY_SHADE_TYPE = "type";
+    public static final String PROPERTY_SHADE_CAPABILITIES = "capabilities";
+    public static final String PROPERTY_SHADE_JSON_HAS_SECONDARY = "jsonHasSecondary";
```
jlaur, line 54: Any reason for having "json" in the name? These properties will be seen by users, so seems kind of internal/technical.

jlaur:
@andrewfg - really a lot of cool work in here! I'm currently scratching the surface, will try to go deeper into soon, hopefully today and/or tomorrow.

arroyoj:
@andrewfg, thanks for the PR to address this issue. I've tested the latest JAR and it is working well for me.

andrewfg:
> I suppose file bundles/org.openhab.binding.hdpowerview/org.openhab.binding.hdpowerview-3.2.0-SNAPSHOT.jar was added by mistake

The JAR is there just so that people can drop it in addons without needing to build it themselves. I would delete it from the repo branch as the last act before it is merged..

bundles/org.openhab.binding.hdpowerview/src/test/java/org/openhab/binding/hdpowerview/HDPowerViewJUnitTests.java
andrewfg: Ok.

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/api/ShadePosition.java
andrewfg: The `switch()` statement runs over an enum which includes values `NONE` and `ERROR` that shall not be handled; and I simply put in the empty `default` to get rid of the compiler warnings; the actual code runs no differently with or without it.

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/database/ShadeCapabilitiesDatabase.java
andrewfg: I will change it to my full name. I'm not sure, but I think (maybe) it is required on all classes (included embedded ones), rather than just on outer classes?? But anyway it does no harm.

jlaur:
> The JAR is there just so that people can drop it in addons without needing to build it themselves. I would delete it from the repo branch as the last act before it is merged..

Ah, okay. In that case I would suggest to mark the PR as draft, or publish the JAR somewhere else. Otherwise the PR could be completed at any time with the JAR.

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/api/ShadePosition.java
jlaur: Sorry, I missed that, just looked strange. Would it be an idea to explicitly provide cases for NONE and ERROR with a comment that they are not handled on purpose? In that case the compiler would provide a warning if a new enum value would be added without adapting this code accordingly.

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/HDPowerViewBindingConstants.java
andrewfg: Hmm. There are two types of properties resp. capabilities -- namely a) those that are **_declared_** through the shade's `type` and `capabilties` attributes, and b) those that are **_inferred_** from the structure of the JSON `positions` response (e.g. if it returns both a `posKind=1` and a `posKind=2` entry then we can infer that it might be a shade with top down and bottom up capabilities). The background is that I created the known types database via some "inspired guesses", and the code is logging any differences between declared and inferred capabilities to provoke users to give us feedback if any of those guesses are wrong. Nevertheless I do agree with you that the 'json' prefix could be confusing for users, and I would be happy to rename the properties. If properties would have long texts, then they would be something like 'potential support for a secondary rail has been detected' and 'potential support for tilt anywhere functionality has been detected' -- and I can't think of a good short text for either of those. => Do you have any suggestions?

andrewfg:
> publish the JAR somewhere else

Yes I will do that.

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/HDPowerViewBindingConstants.java
arroyoj: I agree that removing the "json" prefix would be good. For new names, what about `secondaryRailDetected` and `tiltAnywhereDetected`?

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/api/ShadePosition.java
andrewfg: > explicitly provide cases for NONE and ERROR

Good idea. Will do.

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/HDPowerViewBindingConstants.java
andrewfg: Ok. Done.

andrewfg:
> publish the JAR somewhere else

The JAR is [HERE](https://github.com/andrewfg/openhab-addons/tree/main/bundles/org.openhab.binding.hdpowerview)..

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/api/ShadePosition.java
```diff
@@ -190,56 +154,203 @@ private State getPosition1(CoordinateSystem coordSys) {
...
+        }
+        return UnDefType.UNDEF;
+    }
+
+    /**
+     * Detect if the ShadePosition has a posKindN value indicating support for a secondary rail.
+     *
+     * @return true if the ShadePosition supports a secondary rail.
+     */
+    public boolean jsonSecondary() {
```
jlaur, line 280: Naming: `hasSecondary`?

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/api/ShadePosition.java
```diff
@@ -190,56 +154,203 @@ private State getPosition1(CoordinateSystem coordSys) {
...
+        return SECONDARY_ZERO_IS_OPEN.equals(posKind1) || SECONDARY_ZERO_IS_OPEN.equals(posKind2);
+    }
+
+    /**
+     * Detect if the ShadePosition has both a posKindN value indicating support for tilt, AND a posKindN indicating
+     * support for a primary rail. i.e. it supports tilt anywhere functionality.
+     *
+     * @return true if the ShadePosition supports tilt anywhere.
+     */
+    public boolean jsonTiltAnywhere() {
```
jlaur, line 290: Naming: `canTiltAnywhere` or something similar?

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/api/ShadePosition.java
andrewfg: secondaryDetected

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/api/ShadePosition.java
andrewfg: tiltAnywhereDetected

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/handler/HDPowerViewShadeHandler.java
```diff
@@ -150,16 +154,107 @@ protected void onReceiveUpdate(@Nullable ShadeData shadeData) {
...
+        // update 'capabilities' property
+        final Integer temp = shadeData.capabilities;
+        final int capabilitiesVal = temp != null ? temp.intValue() : -1;
+        capabilities = db.getCapabilities(capabilitiesVal);
+        propKey = HDPowerViewBindingConstants.PROPERTY_SHADE_CAPABILITIES;
+        propOldVal = properties.getOrDefault(propKey, "");
+        propNewVal = capabilities.toString();
+        if (!propNewVal.equals(propOldVal)) {
+            propChanged = true;
+            getThing().setProperty(propKey, propNewVal);
```
jlaur, line 197: Consider updating both properties at the same time with `getThing().setProperties(properties);`

jlaur:
I have noticed a changed behaviour - after starting binding:

```
2021-12-05 21:28:36.657 [INFO ] [ab.event.ThingStatusInfoChangedEvent] - Thing 'hdpowerview:shade:hub:blind2' changed from ONLINE to OFFLINE (COMMUNICATION_ERROR)
2021-12-05 21:28:36.664 [INFO ] [openhab.event.ItemStateChangedEvent ] - Item 'Blind2_Vane' changed from 0 to UNDEF
2021-12-05 21:29:36.905 [INFO ] [ab.event.ThingStatusInfoChangedEvent] - Thing 'hdpowerview:shade:hub:blind2' changed from OFFLINE (COMMUNICATION_ERROR) to ONLINE
2021-12-05 21:29:36.917 [INFO ] [openhab.event.ItemStateChangedEvent ] - Item 'Blind2_Vane' changed from UNDEF to 0
```

The communication error and one minute delay to go online again is fixed by #11707. But still, is it intentional that vane is updated to **Unspecified** when thing is offline? In that case, for consistency, positions should probably also be updated?

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/database/ShadeCapabilitiesDatabase.java
```diff
@@ -0,0 +1,350 @@
...
+     *
+     * @author Andrew Fiddian-Green - Initial Contribution
+     */
+    private static class Base {
+        protected int intValue = -1;
+        protected String text = "-- not in database --";
+
+        @Override
+        public String toString() {
+            return String.format("%d ( %s )", intValue, text);
```
jlaur, line 91: Excess spaces, will become something like "44 ( Twist )" as opposed to "44 (Twist)".

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/database/ShadeCapabilitiesDatabase.java
```diff
@@ -0,0 +1,350 @@
...
+            new Capabilities(0).primary().tiltOnClosed()                      .text("Bottom Up"),
+            new Capabilities(1).primary().tiltAnywhere()                      .text("Bottom Up Tilt 90°"),
+            new Capabilities(2).primary().tiltAnywhere().tilt180()            .text("Bottom Up Tilt 180°"),
+            new Capabilities(3).primary().tiltOnClosed()                      .text("Vertical"),
+            new Capabilities(4).primary().tiltAnywhere().tilt180()            .text("Vertical Tilt 180°"),
+            new Capabilities(5)          .tiltAnywhere().tilt180()            .text("Tilt Only 180°"),
+            new Capabilities(6).primary()                                     .text("Top Down")                 .primaryStateInverted(),
+            new Capabilities(7).primary()                         .secondary().text("Top Down Bottom Up"),
+            new Capabilities(8).primary()                                     .text("Duolite Lift"),
+            new Capabilities(9).primary().tiltAnywhere()                      .text("Duolite Lift and Tilt 90°"),
```
jlaur, line 47: As these texts are exposed to users as properties, do we want to support i18n?

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/database/ShadeCapabilitiesDatabase.java
jlaur: Also, consider if swapping type and name would be more user-oriented, i.e. "Twist (44)".

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/database/ShadeCapabilitiesDatabase.java
```diff
@@ -0,0 +1,350 @@
...
+ */
+@NonNullByDefault
+public class ShadeCapabilitiesDatabase {
+
+    private final Logger logger = LoggerFactory.getLogger(ShadeCapabilitiesDatabase.class);
+
+    /*
+     * Database of known shade capabilities.
+     */
+    private static final Capabilities[] CAPABILITIES_DATABASE = {
```
jlaur, line 36: Consider if using a `Map` or something similar would make sense for _O(1)_ lookup in getCapabilities/isCapabilitiesInDatabase.

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/discovery/HDPowerViewShadeDiscoveryService.java
```diff
@@ -96,12 +97,35 @@ private Runnable createScanner() {
...
+                                    }
+                                }
+
+                                Integer capabilities = shadeData.capabilities;
+                                int caps = (capabilities != null) ? capabilities.intValue() : -1;
+                                if (caps >= 0) {
+                                    if (db.isCapabilitiesInDatabase(caps)) {
+                                        builder = builder.withProperty(
+                                                HDPowerViewBindingConstants.PROPERTY_SHADE_CAPABILITIES,
+                                                db.getCapabilities(caps).toString());
```
jlaur, line 121: Two _O(n)_ lookups for the same capabilities right after each other. Consider dropping `isCapabilitiesInDatabase` and letting `getCapabilities` return some distinct value (or even null) if capabilities are not known to avoid two lookups.

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/database/ShadeCapabilitiesDatabase.java
andrewfg: No. These are capabilities declared by HD; in English.

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/handler/HDPowerViewShadeHandler.java
andrewfg: I considered that, but see no point. Maybe one, maybe the other, or maybe both would be updated (most likely not the latter at the same time). So why bother.

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/database/ShadeCapabilitiesDatabase.java
andrewfg: Ok.

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/handler/HDPowerViewShadeHandler.java
jlaur: Ah, yes, you're right.

jlaur:
@andrewfg - I don't think I have more comments. The only thing left for me would be testing getters/setters for my Twist (with vane), Duette and Duette top down/bottom up. Do you have anything in particular you would like tested?

andrewfg:
> don't think I have more comments

I still have some more work to address your comments of yesterday.

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/database/ShadeCapabilitiesDatabase.java
andrewfg: Ok. Love the very "hairy" code..

```
private static final Map<Integer, Capabilities> CAPABILITIES_DATABASE = Arrays.asList(
    new Capabilities(0).primary().tiltOnClosed().text("Bottom Up"),
    ... etc.
    new Capabilities()).stream().collect(Collectors.toMap(Capabilities::getValue, Function.identity()));
```

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/discovery/HDPowerViewShadeDiscoveryService.java
andrewfg: OK.

andrewfg:
> I have noticed a changed behaviour - after starting binding:

I think that this was due to a timing error --  `updateSoftProperties()` must initialize the shade capabilities before 'updateBindingStates()` gets called..

andrewfg:
@jlaur I finished the code changes that you requested. I also made a few other minor tweaks. And I fixed the timing issue with the state update 'UNDEF'. So it is basically all ready to go again. But perhaps you can do a final check regarding the state update 'UNDEF' -- just in case.

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/database/ShadeCapabilitiesDatabase.java
```diff
@@ -45,14 +50,14 @@
             new Capabilities(7).primary()                         .secondary().text("Top Down Bottom Up"),
             new Capabilities(8).primary()                                     .text("Duolite Lift"),
             new Capabilities(9).primary().tiltAnywhere()                      .text("Duolite Lift and Tilt 90°"),
-        // @formatter:on
-            new Capabilities() };
+    // @formatter:on
+            new Capabilities()).stream().collect(Collectors.toMap(Capabilities::getValue, Function.identity()));
```
jlaur, line 54: I believe this can be simplified and optimized (by avoiding some unneeded conversions) into something like:

```
    private static final Map<Integer, Capabilities> CAPABILITIES_DATABASE = Map.ofEntries(
    // @formatter:off
            Map.entry(0, new Capabilities(0).primary().tiltOnClosed()                      .text("Bottom Up")),
            Map.entry(1, new Capabilities(1).primary().tiltAnywhere()                      .text("Bottom Up Tilt 90°")),
[...]
    // @formatter:on
            );
```

andrewfg: Agreed. But why do you want a human to refactor it, when you can do it with a computer? ;)

andrewfg: In particular i want to avoid the risk that the map's key and the value's value may be different. It is anyway a static final field so any minor code overhead would be one off on loading the JAR..

jlaur: @andrewfg - hmm, damn, I feel a joke here, but not getting it. :)

jlaur: Well, problem is that the key is also part of the value. But maybe it doesn't need to be. You have two usages:
- **isCapabilitiesInDatabase:** Just checks existence, key not needed.
- **getCapabilities:** Returns Capabilities from key. Which means caller already has the key, so should not be needed.

andrewfg: It's not a simple as that. In some places it uses 'naked' Capabilities instances and/or 'naked' Type instances (ones which are explicitly not in the Map) that have an intValue of -1, and a specific text, to indicate an error. So it is not possible to  relocate the Value intValue to the Map Key. If we are to use a Map, then we must to duplicate the Map key and Value intValue. (That is the original reason why I did not use a Map). If you insist on having a Map, the I cannot accept your proposed method of manual Map entry creation because it introduces the risk of creating a entries where the Map Key and the Value intValue are not equal. The advantage of the `Arrays.asList().stream().collect(Collectors.toMap(Capabilities::getValue, Function.identity()))` technique, is that Value intValue is declared just once, and the class constructor creates the Map Key directly from it.

arroyoj (approved):
LGTM

bundles/org.openhab.binding.hdpowerview/README.md
```diff
@@ -93,20 +93,34 @@ And for horizontal shades, it maps the horizontal position of the "truck" to the
...
+| Single action<br>bottom-up  | `position`        | ▲                     | Up               | `OPEN`         | 0%      | ▲                            |
+|                             |                   | ▼                     | Down             | `CLOSED`       | 100%    | ▼                            |
+| Single action<br>top-down   | `position`        | ▲                     | Up               | ***`CLOSED`*** | 0%      | ▲                            |
+|                             |                   | ▼                     | Down             | ***`OPEN`***   | 100%    | ▼                            |
+| Single action<br>right-left | `position`        | ▲                     | ***Left***       | `OPEN`         | 0%      | ▲                            |
+|                             |                   | ▼                     | ***Right***      | `CLOSED`       | 100%    | ▼                            |
+| Single action<br>eft-right | `position`        | ▲                     | ***Right***      | `OPEN`         | 0%      | ▲                            |
```
arroyoj, line 104: The updated ReadMe looks great! Just found one very small typo: missing l (L) in "left", after the <br>.

andrewfg:
@jlaur I just want to check with you that you have no more issues with this PR. I rebased it so that it includes your recent changes on scene collections and automations, so if you build the JAR you should have the full functionality. And if you are happy with it, could you please close your review, so the maintainers can proceed?

jlaur:
@andrewfg - cool! I'll switch to your version now and perform a few verifications. You closed all me comments, but I don't see any distinct option to "close the review"?

jlaur:
@andrewfg - there's a problem reading vane position from Twist. I'm investigating it and will also try to compare with main branch.

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/database/ShadeCapabilitiesDatabase.java
```diff
@@ -0,0 +1,339 @@
...
+         * Check if the Capabilities class instance supports a primary shade.
+         *
+         * @return true if it supports a primary shade.
+         */
+        public boolean supportsPrimary() {
+            return supportsPrimary;
+        }
+
+        /**
+         * Check if the Capabilities class instance supports a vane.
```
jlaur, line 204: Method description mentions vane support, but name mentions tilt anywhere support. Is it the same? My Twist (44) with capabilities "Bottom Up (0)" has a vane/slat, but property is: "tiltAnywhereDetected: false"

jlaur:
> @andrewfg - there's a problem reading vane position from Twist. I'm investigating it and will also try to compare with main branch.

I was not able to reproduce this. Confirmed now that this branch and main works the same way:
- Vane fully open: **100**
- Vane fully closed: **0**
- When position is < 100: **UNDEF**

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/database/ShadeCapabilitiesDatabase.java
andrewfg: A couple of comments..
- In general 'Vane' and 'Tilt' mean the same function -- namely to adjust the light transmission without changing the vertical position; albeit perhaps making a different physical movement on different types of shade.
- There are two types of vane/tilt -- namely a) the shade has one motor that influences both the position **or** vane/tilt, and b) the shade has two motors that can fully independently move the position **and** vane/tilt. The former strictly can only adjust the vane/tilt when the (e.g. venetian) blind has been fully opened; and any movement after it is open, adjusts the vanes; typically they only have 90 degrees of vane/tilt (dark .. light). The latter can literally move the vane/tilt at any shade position; and typically they have 180 degrees of vane/tilt (dark .. light .. dark). The former type of vane/tilt is included in `capabilities:0` whereas `tiltAnywhere` is declared in another class of `capabilities`.
- And yes the Twist blinds are indded "special" :) (I have some, so I know them well). They have just one motor, so they are declared as `capabilities:0` i.e. NOT supporting tilt anywhere. But the oddity of Twist is that you can indeed adjust the overlap of the bars (dark .. light .. dark) at any shade position by means of its single motor. Nevertheless electronically the hub reports only a `position1` value that can be either `posKind:1` (shade) OR `posKind:3` (vane/tilt) [to be specific it does NOT report a separate `position2`]. And the interpretation of the `position1` / `posKind:1` / `posKind:3` values is indeed as you describe in your own post below, and indeed as described in the ReadMe.

However I will update the comments to eliminate any confusion over vane/tilt.

andrewfg:
> I don't see any distinct option to "close the review"?

Hmm. See this https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/reviewing-changes-in-pull-requests/reviewing-proposed-changes-in-a-pull-request

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/database/ShadeCapabilitiesDatabase.java
jlaur: Thanks for the detailed explanation - and for the clarification in the method doc.

jlaur:
> > I don't see any distinct option to "close the review"?

> Hmm. See this https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/reviewing-changes-in-pull-requests/reviewing-proposed-changes-in-a-pull-request

I think I didn't "open" a review, but just left single comments. Or maybe it's missing privileges as I'm not a maintainer. In any case, can't close the review. But for what it's worth, I'm okay with the changes and didn't observe any problems running this version.

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/database/ShadeCapabilitiesDatabase.java
```diff
@@ -0,0 +1,339 @@
...
+    /**
+     * Base class that is extended by Type and Capabilities classes.
+     *
+     * @author Andrew Fiddian-Green - Initial Contribution
+     */
+    private static class Base {
+        protected int intValue = -1;
+        protected String text = "-- not in database --";
+
+        protected Integer getValue() {
```
lolodomo, line 94: Is there a reason to not return int ?

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/database/ShadeCapabilitiesDatabase.java
```diff
@@ -0,0 +1,339 @@
...
+     * @param type
+     * @param capabilities
+     */
+    public void logCapabilitiesMismatch(int type, int capabilities) {
+        logger.warn("The 'capabilities:{}' reported by shade 'type:{}' don't match the database!{}", capabilities, type,
+                REQUEST_DEVELOPERS_TO_UPDATE);
+    }
+
+    /**
+     * Log a message indicating that a shade's secondary/vanes support, as observed via its actual JSON pay-load, does
```
lolodomo, line 325: payload I believe

lolodomo (changes requested):
Review part 1 of 2

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/api/ShadePosition.java
```diff
@@ -28,159 +31,126 @@
...
-        switch (actuatorClass) {
-            case PRIMARY_ACTUATOR:
-                return getPosition1(coordSys);
-            case SECONDARY_ACTUATOR:
-                return getPosition2(coordSys);
-            default:
-                return UnDefType.UNDEF;
+    public State getState(Capabilities shadeCapabilities, CoordinateSystem posKindCoords) {
+        State result = getPosition1(shadeCapabilities, posKindCoords);
+        if (UnDefType.UNDEF.equals(result)) {
```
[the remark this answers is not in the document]
jlaur: For enum I think `==` is preferable. See also https://github.com/openhab/openhab-addons/pull/11768#issuecomment-991970074.

bundles/org.openhab.binding.hdpowerview/README.md
```diff
@@ -95,20 +95,34 @@ And for horizontal shades, it maps the horizontal position of the "truck" to the
...
-|                          |                   | ▼                     | ***Right***      | `CLOSED`       | 100%    |
-| Single action left-right | `position`        | ▲                     | ***Right***      | `OPEN`         | 0%      |
-|                          |                   | ▼                     | ***Left***       | `CLOSED`       | 100%    |
-| Dual action (lower rail) | `position`        | ▲                     | Up               | `OPEN`         | 0%      |
-|                          |                   | ▼                     | Down             | `CLOSED`       | 100%    |
-| Dual action (upper rail) | ***`secondary`*** | ▲                     | ***Down***       | `OPEN`         | 0%      |
-|                          |                   | ▼                     | ***Up***         | `CLOSED`       | 100%    |
+| Type of Shade            | Channel           | Rollershutter Command | Motion direction | Shade State    | Percent | Pebble Remote Button         |
+|--------------------------|-------------------|-----------------------|------------------|----------------|---------|------------------------------|
```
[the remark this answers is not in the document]
andrewfg: Ok I will do that. (But I suppose that you are aware that MarkDown renderers don’t care about such trivia in the source code, and will anyway align tables like this automatically? You can see how it renders [at the bottom of this page](https://github.com/andrewfg/openhab-addons/tree/hdpowerview-database-position-bugfix/bundles/org.openhab.binding.hdpowerview#hunter-douglas-luxaflex-powerview-binding))

bundles/org.openhab.binding.hdpowerview/README.md
```diff
@@ -95,20 +95,34 @@ And for horizontal shades, it maps the horizontal position of the "truck" to the
...
+| Single action<br>bottom-up  | `position`        | ▲                     | Up               | `OPEN`         | 0%      | ▲                            |
```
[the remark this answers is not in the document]
andrewfg: Yes that is intended. It makes the MarkDown renderer produce better looking column widths in the table. See how it looks in the link in the comment above.

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/api/ShadePosition.java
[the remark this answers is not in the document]
andrewfg: Ok. I will change this.

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/database/ShadeCapabilitiesDatabase.java
andrewfg: Not sure. I will check.

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/database/ShadeCapabilitiesDatabase.java
andrewfg: Ok, I will change.

bundles/org.openhab.binding.hdpowerview/README.md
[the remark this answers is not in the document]
lolodomo: Yes, I am aware

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/database/ShadeCapabilitiesDatabase.java
lolodomo: Maybe for the map ?
I don't remember if a Map requires objects as key or if a simple basic type is ok too.

bundles/org.openhab.binding.hdpowerview/src/test/java/org/openhab/binding/hdpowerview/HDPowerViewJUnitTests.java
```diff
@@ -300,93 +352,245 @@ public void shadeResponseIsParsedCorrectly() throws JsonParseException {
...
+         * for (CoordinateSystem coordinate2 : coordinates) {
+         * System.out.println(String.format("%s|%s|%d|%s|%s|%s", capabilities.toString(),
+         * coordinate.toString(), position, gson.toJson(shade), coordinate2.toString(),
+         * shade.getState(capabilities, coordinate2)));
+         * }
+         * }
+         * }
+         * }
+         * }
+         */
```
lolodomo, line 594: As fully commented, I propose to remove this test.

lolodomo (changes requested):
Review part 2 of 2

lolodomo:
To be honest, IU have not reviewed in details the unit tests.

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/database/ShadeCapabilitiesDatabase.java
andrewfg: That could be it. Maps do require objects rather than primitives.

bundles/org.openhab.binding.hdpowerview/src/test/java/org/openhab/binding/hdpowerview/HDPowerViewJUnitTests.java
andrewfg: Let me think about that. It is useful code so it may make more sense to convert ut to an actual test case.

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/database/ShadeCapabilitiesDatabase.java
lolodomo: Ok, I close this comment.

lolodomo (approved):
LGTM

jlaur:
@andrewfg - congrats. Finally! :-) And of course - shortly after the merge I discovered a minor inconvenience which I unfortunately didn't observe/realize before. I've been running a version based on #11853. So after this PR was merged into main I rebased towards main to include your changes also.

After updating the binding I wanted to check the new shade properties. They were not added, so I restarted openHAB entirely. Now the slight problem: As soon as the things appeared I checked again, and this time it looked like this:
image

Then after quite a while (because my system was still busy loading everything) it was updated correctly:
image

I believe it is caused by this definition:
```
		<properties>
			<property name="vendor">Hunter Douglas (Luxaflex)</property>
			<property name="modelId">PowerView Motorized Shade</property>
			<property name="type"></property>
			<property name="capabilities"></property>
			<property name="secondaryRailDetected"></property>
			<property name="tiltAnywhereDetected"></property>
		</properties>
```

It would probably be most correct to leave them out and only add them dynamically when they are available. What do you think? It's a small thing, so just wanted to share this observation.

arroyoj:
@andrewfg, thanks so much for fixing the bug, updating the README, and all the other improvements. It's great to have it all merged in.

@jlaur, I agree with @andrewfg that having the properties blank for a short time before the hub has been polled isn't too bad. I'm fine with it as is.

jlaur:
@andrewfg, @arroyoj - that's fine, no rush as it's just a minor glitch. I'll probably fix it next time I'm doing something in that area. @andrewfg - just minor correction: It's not only an issue when new version is installed, it will always happen while starting openHAB. At least for file-based configuration.

PR #11705, opened 2021-12-05
[openwebnet] reduced log messages during UPnP bridge discovery

### LINKED ISSUE #11701: [openwebnet] reduce INFO log messages for upnp discovery
Logged messages while discovering new devices in the network via UPNP should be reduced to just those related to BTicino devices, all other messages should be logged at DEBUG level.

mvalla:
Closes #11701 by changing 2 log messages to DEBUG level.

### DIFF
--- a/bundles/org.openhab.binding.openwebnet/src/main/java/org/openhab/binding/openwebnet/internal/discovery/BusGatewayUpnpDiscovery.java
+++ b/bundles/org.openhab.binding.openwebnet/src/main/java/org/openhab/binding/openwebnet/internal/discovery/BusGatewayUpnpDiscovery.java
@@ -150,10 +150,10 @@ public class BusGatewayUpnpDiscovery implements UpnpDiscoveryParticipant {
 
     @Override
     public @Nullable DiscoveryResult createResult(RemoteDevice device) {
-        logger.info("Found device {}", device.getType());
+        logger.debug("Found device {}", device.getType());
         DeviceInfo devInfo = new DeviceInfo(device);
         if (!devInfo.manufacturer.matches("<unknown>")) {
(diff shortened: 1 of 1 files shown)


bundles/org.openhab.binding.openwebnet/src/main/java/org/openhab/binding/openwebnet/internal/discovery/BusGatewayUpnpDiscovery.java
```diff
@@ -226,7 +226,7 @@ private DeviceInfo(RemoteDevice device) {
                     }
                 }
             }
-            logger.info("Found BTicino device: not a OpenWebNet gateway or is not supported (UDN={})", idString);
+            logger.info("Found BTicino device: not a OpenWebNet gateway or not supported (UDN={})", idString);
```
lolodomo, line 229: Even this one could be rather DEBUG? 
We should have the minimum of INFO logs.

mvalla: I would find useful to tell the user via INFO message that the BTicino device that was found is NOT a gateway that can be used by this binding for integration. Therefore I would leave the level to INFO.

lolodomo: I let another maintainer decide.
@kaikreuzer ?

lolodomo:
@fwolter : please have a look, if this is ok for you to leave an INFO level log, you can merge.

fwolter (approved):

PR #11706, opened 2021-12-05
[homeconnect] Add oven control

bruestel:
Allow to control oven via binding. It is now possible to control the oven without an additional agreement (start the program, etc.

Signed-off-by: Jonas Brüstel <jonas@bruestel.net>


lolodomo (approved):
LGTM

kl8ter:
It would be nice if you could set the “duration time” as well as the “Fast Heat” over the binding.

PR #11713, opened 2021-12-05
[dsmr] Add missing English translations

wborn:
Without these English defaults Crowdin would remove Dutch translations.

See: https://github.com/openhab/openhab-addons/pull/11712


Hilbrand (commented):
Maybe it's better to remove the label and description from the XML files or use the @ reference notation, otherwise the label and description exist in 2 places, meaning undesirable maintenance burden, and will be likely cause mismatches if some forgot to update both if needed.

lolodomo (approved):
LGTM

PR #11714, opened 2021-12-05
[mqtt] Add missing English translations

wborn:
Without these English defaults Crowdin would remove German translations.

See: https://github.com/openhab/openhab-addons/pull/11712/


lolodomo (approved):
LGTM

PR #11719, opened 2021-12-06
GraalJS now uses automation/js

jpg0:
This change modifies the 'new' GraalJS plugin to monitor `<openHAB-conf>/automation/js for scripts`, with libraries in `<openHAB-conf>/automation/js/node_modules` - as per NodeJS standard spec.

It also removes itself from monitoring the original `automation/jsr233/javascript/*` directories, and instead leaves this for Nashorn. This allows users to move scripts over to GraalJS from Nashorn one-by-one.

Note that this change **breaks any ES6+ scripts which are in automation/jsr233/javascript/\* directories**. These will need to be moved to `automation/js`.

Note also that this PR depends on https://github.com/openhab/openhab-core/pull/2593 which enables it to use it's own script directory.

### DIFF
--- a/bundles/org.openhab.automation.jsscripting/src/main/java/org/openhab/automation/jsscripting/internal/GraalJSScriptEngineFactory.java
+++ b/bundles/org.openhab.automation.jsscripting/src/main/java/org/openhab/automation/jsscripting/internal/GraalJSScriptEngineFactory.java
@@ -22,8 +22,6 @@ import javax.script.ScriptEngine;
 import org.openhab.core.automation.module.script.ScriptEngineFactory;
 import org.osgi.service.component.annotations.Component;
 
-import com.oracle.truffle.js.scriptengine.GraalJSEngineFactory;
-
 /**
  * An implementation of {@link ScriptEngineFactory} with customizations for GraalJS ScriptEngines.
  *
@@ -32,13 +30,24 @@ import com.oracle.truffle.js.scriptengine.GraalJSEngineFactory;
 @Component(service = ScriptEngineFactory.class)
 public final class GraalJSScriptEngineFactory implements ScriptEngineFactory {
 
+    public static final String MIME_TYPE = "application/javascript;version=ECMAScript-2021";
+
     @Override
     public List<String> getScriptTypes() {
         List<String> scriptTypes = new ArrayList<>();
-        GraalJSEngineFactory graalJSEngineFactory = new GraalJSEngineFactory();
 
-        scriptTypes.addAll(graalJSEngineFactory.getMimeTypes());
-        scriptTypes.addAll(graalJSEngineFactory.getExtensions());
+        /*
+         * Whilst we run in parallel with Nashorn, we use a custom mime-type to avoid
+         * disrupting Nashorn scripts. When Nashorn is removed, we take over the standard
+         * JS runtime.
+         */
+
+        // GraalJSEngineFactory graalJSEngineFactory = new GraalJSEngineFactory();
+        //
+        // scriptTypes.addAll(graalJSEngineFactory.getMimeTypes());
+        // scriptTypes.addAll(graalJSEngineFactory.getExtensions());
+
+        scriptTypes.add(MIME_TYPE);
 
         return Collections.unmodifiableList(scriptTypes);
     }
--- a/bundles/org.openhab.automation.jsscripting/src/main/java/org/openhab/automation/jsscripting/internal/fs/watch/JSDependencyTracker.java
+++ b/bundles/org.openhab.automation.jsscripting/src/main/java/org/openhab/automation/jsscripting/internal/fs/watch/JSDependencyTracker.java
@@ -1,0 +1,61 @@
+/**
+ * Copyright (c) 2010-2021 Contributors to the openHAB project
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package org.openhab.automation.jsscripting.internal.fs.watch;
+
+import java.io.File;
+
+import org.openhab.core.OpenHAB;
+import org.openhab.core.automation.module.script.rulesupport.loader.DependencyTracker;
+import org.osgi.service.component.annotations.Activate;
+import org.osgi.service.component.annotations.Component;
+import org.osgi.service.component.annotations.Deactivate;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Tracks JS module dependencies
+ *
+ * @author Jonathan Gilbert - Initial contribution
+ */
+@Component(immediate = true, service = JSDependencyTracker.class)
+public class JSDependencyTracker extends DependencyTracker {
+
+    private final Logger logger = LoggerFactory.getLogger(JSDependencyTracker.class);
+
+    public static final String LIB_PATH = String.join(File.separator, OpenHAB.getConfigFolder(), "automation", "js",
+            "node_modules");
+
+    public JSDependencyTracker() {
+        super(LIB_PATH);
+    }
+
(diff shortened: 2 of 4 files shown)


bundles/org.openhab.automation.jsscripting/src/main/java/org/openhab/automation/jsscripting/internal/fs/watch/JSDependencyTracker.java
```diff
@@ -0,0 +1,48 @@
...
+
+/**
+ * Tracks JS module dependencies
+ *
+ * @author Jonathan Gilbert - Initial contribution
+ */
+@Component(immediate = true, service = JSDependencyTracker.class)
+public class JSDependencyTracker extends DependencyTracker {
+
+    public static final String LIB_PATH = String.join(File.separator, OpenHAB.getConfigFolder(), "automation", "js",
```
digitaldan, line 32: Since this directory does not exist yet, should we attempt to create it when this bundle is activated?  I think if the user creates it after the script watcher starts to monitor, they would need to restart OH (or the related bundles) to have the watcher see it.

bundles/org.openhab.automation.jsscripting/src/main/java/org/openhab/automation/jsscripting/internal/fs/watch/JSScriptFileWatcher.java
```diff
@@ -0,0 +1,66 @@
...
+import org.osgi.service.component.annotations.Reference;
+
+/**
+ * Monitors <openHAB-conf>/automation/js for Javascript files
+ *
+ * @author Jonathan Gilbert - Initial contribution
+ */
+@Component(immediate = true)
+public class JSScriptFileWatcher extends ScriptFileWatcher {
+    private static final String FILE_DIRECTORY = "automation" + File.separator + "js";
```
digitaldan, line 36: Since this directory does not exist yet, should we attempt to create it when this bundle is activated?

bundles/org.openhab.automation.jsscripting/src/main/java/org/openhab/automation/jsscripting/internal/GraalJSScriptEngineFactory.java
```diff
@@ -32,13 +30,24 @@
 @Component(service = ScriptEngineFactory.class)
 public final class GraalJSScriptEngineFactory implements ScriptEngineFactory {
 
+    public static final String MIME_TYPE = "x-application/javascript.new";
```
digitaldan, line 33:
```suggestion
    public static final String MIME_TYPE = "application/javascript;version=ECMAScript-2021";
```

As suggested by @jpg0 in another conversation

bundles/org.openhab.automation.jsscripting/src/main/java/org/openhab/automation/jsscripting/internal/fs/watch/JSScriptFileWatcher.java
jpg0: So I had a quick look at the code to do this and it raised some questions:

 - This is just an implementation of ScriptFileWatcher, so should this base class create all the directories it watches? This would be `automation/js` and `automation/jsr223`.
 - There is also a ScriptLibraryWatcher, which watches `automation/js/node_modules` - should it create that directory?

I am leaning towards creating the two directories, so putting the creation into ScriptFileWatcher. `node_modules` would get created by npm anyway.

digitaldan: Sounds good!

jpg0: I've added it to the PR in core: https://github.com/openhab/openhab-core/pull/2593

bundles/org.openhab.automation.jsscripting/src/main/java/org/openhab/automation/jsscripting/internal/fs/watch/JSDependencyTracker.java
jpg0: Hmm, good point about the restarting. I wondered if it would start watching once created, but it doesn't look like it does.

bundles/org.openhab.automation.jsscripting/src/main/java/org/openhab/automation/jsscripting/internal/fs/watch/JSScriptFileWatcher.java
digitaldan: Awesome !

bundles/org.openhab.automation.jsscripting/src/main/java/org/openhab/automation/jsscripting/internal/fs/watch/JSDependencyTracker.java
jpg0: I've added this in now, just like for the script watching.

digitaldan: Awesome.

cweitkamp (commented):
Waiting for https://github.com/openhab/openhab-core/runs/4497892590 to succeed. Then we can rerun the build for this PR.

bundles/org.openhab.automation.jsscripting/src/main/java/org/openhab/automation/jsscripting/internal/GraalJSScriptEngineFactory.java
```diff
@@ -32,13 +30,24 @@
...
 
-        scriptTypes.addAll(graalJSEngineFactory.getMimeTypes());
-        scriptTypes.addAll(graalJSEngineFactory.getExtensions());
+        /*
+         * Whilst we run in parallel with Nashorn, we use a custom mime-type to avoid
+         * disrupting Nashorn scripts. When Nashorn is removed, we take over the standard
+         * JS runtime.
+         */
+
+        // GraalJSEngineFactory graalJSEngineFactory = new GraalJSEngineFactory();
```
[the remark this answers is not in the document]
digitaldan: They are left there as a reminder of what the code should be when we eventually remove nashorn in a future release.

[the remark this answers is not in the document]
jpg0: @digitaldan  is correct. They are commented out now to allow this JS addon to coexist with Nashorn. At the time we remove Nashorn, we uncomment these lines and this will take over Nashorn scripts.

If there is another process to do this though, I'm happy to use it.

jpg0:
I am not sure why this is failing; has it not yet caught up with the change in core?

bundles/org.openhab.automation.jsscripting/src/main/java/org/openhab/automation/jsscripting/internal/GraalJSScriptEngineFactory.java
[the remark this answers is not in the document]
cweitkamp: Alright. Keep it in place.

digitaldan:
@jpg0 https://ci.openhab.org/job/openHAB-Core/880/ is building right now which i think will push updated core bundles to the repos.

bundles/org.openhab.automation.jsscripting/src/main/java/org/openhab/automation/jsscripting/internal/fs/watch/JSDependencyTracker.java
```diff
@@ -0,0 +1,61 @@
...
+            "node_modules");
+
+    public JSDependencyTracker() {
+        super(LIB_PATH);
+    }
+
+    @Activate
+    public void activate() {
+
```
kaikreuzer, line 44:
```suggestion
```

kaikreuzer (approved):

PR #11726, opened 2021-12-07
[lutron] Explicitly enable proper monitoring types for HomeWorks connections

### LINKED ISSUE #11604: [lutron] Explicitly enable the proper monitoring types for HomeWorks connections
HomeWorks logins don't always have the proper monitoring types enabled by default. This prevents the binding's ipbridge from working properly, since it will not receive the notifications it needs. This seems to have become a problem since Lutron started discouraging installers from enabling the default "lutron" login for HomeWorks processors, which did have the proper monitoring types enabled by default.

An update to ipbridge is needed so that it will explicitly enable the proper monitoring types for HomeWorks connections at connection init time rather than relying on the installer to configure the monitoring defaults properly for the login being used.

This may be an issue for some RadioRA2 installations as well, although it hasn't been reported.

bobadair:
Update to ipbridge to explicitly enable the proper monitoring types for HomeWorks LIP connections at connection init time.

Closes #11604

### DIFF
--- a/bundles/org.openhab.binding.lutron/src/main/java/org/openhab/binding/lutron/internal/protocol/lip/Monitoring.java
+++ b/bundles/org.openhab.binding.lutron/src/main/java/org/openhab/binding/lutron/internal/protocol/lip/Monitoring.java
@@ -1,0 +1,65 @@
+/**
+ * Copyright (c) 2010-2021 Contributors to the openHAB project
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package org.openhab.binding.lutron.internal.protocol.lip;
+
+import java.util.Collections;
+import java.util.Set;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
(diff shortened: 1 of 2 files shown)


bundles/org.openhab.binding.lutron/src/main/java/org/openhab/binding/lutron/internal/protocol/lip/Monitoring.java
```diff
@@ -0,0 +1,65 @@
...
+    public static final Integer DIAGERROR = 22;
+    public static final Integer SHADEGRP = 23;
+    public static final Integer PARTITION = 24;
+    public static final Integer SYSTEM = 25;
+    public static final Integer SENSORGROUP = 26;
+    public static final Integer TEMPSENSOR = 27;
+    public static final Integer ALL = 255;
+
+    /** Set of monitoring types which must be enabled */
+    public static final Set<Integer> REQUIRED_SET = Collections.unmodifiableSet(
```
lolodomo, line 63: You could simply use Set.of()

lolodomo (changes requested):

lolodomo (approved):
LGTM

bundles/org.openhab.binding.lutron/src/main/java/org/openhab/binding/lutron/internal/protocol/lip/Monitoring.java
bobadair: Thanks. Not sure what I was thinking there.

PR #11728, opened 2021-12-07
[pollytts] Add default translations properties file

wborn:
This is the only remaining voice add-on which isn't yet translatable with Crowdin.


lolodomo (approved):
LGTM

PR #11729, opened 2021-12-07
[hdpowerview] Introduce 'new' secondary display mode.

### LINKED ISSUE #11593: [hdpowerview] Resolve differences between OH dual-rail shade behavior, HD remote behavior, and documentation

andrewfg:
Fixes #11593 

This PR does two things..
1. Corrects errors in the ReadMe describing the motion of the secondary rail.
2. Introduces a `legacySecondaryMode` configuration parameter, whose value is false on new Things that are created via auto- discovery in the UI.
3. Introduces a new secondary mode that reverses the secondary percentage display if `legacySecondaryMode` is neither `true` nor `null` missing.

Depends on #11698


andrewfg:
@arroyoj ..
- You can download the new JAR is [HERE](https://github.com/andrewfg/openhab-addons/tree/main/bundles/org.openhab.binding.hdpowerview)
- You can read the new README is [HERE](https://github.com/andrewfg/openhab-addons/tree/hdpowerview-change-up-down/bundles/org.openhab.binding.hdpowerview)

arroyoj:
@andrewfg, thanks for putting together the PR. I downloaded the JAR and started testing it, and I came across 2 issues so far:

1. ReadMe: in the "legacy" table for the secondary rail, the pebble remote button images are swapped. The up arrow command corresponds to "right.png", even under "legacy".
2. The legacySecondaryMode config parameter logic appears to be backwards. After installing the JAR, making no changes to my shade Things (so legacySecondaryMode was null), my shades were behaving like "modern", **not** like v3.1.0. I then set a single shade to legacySecondaryMode: true. That shade continued to act like "modern". When I finally set legacySecondaryMode to false, then it actually acted like the legacy v3.1.0 behavior. I'm guessing this is related to this "newSecondaryMode" boolean:
```
       /*
         * Set newSecondaryMode if the legacySecondaryMode configuration parameter is not explicitly FALSE.
         * For the avoidance of doubt: this means legacySecondaryMode is NEITHER..
         * - NULL i.e. the real legacy case where the configuration parameter is missing, NOR
         * - TRUE i.e. where the configuration parameter has been manually overridden.
         */
        newSecondaryMode = Boolean.FALSE.equals(getConfigAs(HDPowerViewShadeConfiguration.class).legacySecondaryMode);
```
First, the logic described in the comments really confused me, and I'm struggling to wrap my head around it. I think it would be less confusing if the "legacySecondaryMode" config parameter set a "legacySecondary" boolean that enabled the old behavior, rather than enabling/disabling the "new" behavior. Is there a namespace issue that the config parameter and boolean cannot have the same name, since having the same name would be the easiest to understand.

Second, although the boolean is named "newSecondaryMode", when this boolean is TRUE, then the "legacy" behavior is being enabled in the getter and setter code. My suggestion would be a) rename the boolean to refer to the legacy mode and b) set the boolean to TRUE if either legacySecondaryMode is NULL or TRUE.

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/api/CoordinateSystem.java
```diff
@@ -41,71 +43,69 @@
...
-     *      posKind: 1 {ZERO_IS_CLOSED}
+     * Shade and vane fully down: (top-down: closed, bottom-up: open)
+     *      posKind: 1
      *      position1: 0
-     * 
+     *
      * ALTERNATE: Shade and vane fully down: (top-down: closed, bottom-up: open)
-     *      posKind: 3 {VANE_COORDS}
+     *      posKind: 3
      *      position: 0
```
arroyoj, line 64: There is a small error (pre-dates this PR, but probably worth fixing) in these comments. When the shade (primary rail) is fully up, a "top-down" shade is closed and a "bottom-up" shade is open, so the descriptions on line 54 are backwards. Similarly, when the shade/primary rail is fully down, the top-down shade is open and the bottom-up shade is closed, so the descriptions on lines 58 and 62 are also backwards. These changes will make this set of comments consistent with the comments on lines 37-41 that were taken from the Hunter Douglas API docs.

andrewfg:
> the logic described in the comments really confused me, and I'm struggling to wrap my head around it

Indeed. The challenge is how to make the Config Param initialize properly on legacy resp. new Things for all of the following six cases. And I was playing around to see if double-de-referencing might fix it. ... It doesn't..
1. Legacy Thing created from an existing .things file:  param absent (null). 
2. Legacy Thing previously created from the Inbox: param absent (null).
3. Legacy Thing previously created manually in the UI: param absent (null).
4. New Thing created a newly written .things file:  param is either absent (null), or present with value explicitly true/false; user **MUST** decide.
5. New Thing created from the Inbox:  new discovery code pre-sets param explicitly to true; user **_may_** override in the UI.
6. New Thing previously created manually in the UI: param should be pre-set to true, but actually defaults to absent (null); user **MUST** override in the UI.

Cases 1. 2. 3. and 5. are 100% Ok; case 4. is not ideal, but probably Ok too; and I think case 6. is not solvable.

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/api/CoordinateSystem.java
andrewfg: Ok.

andrewfg:
Hmmm.

I am wondering if instead of having a Config Param to differentiate the behaviour of a single class of shades, we might instead have two classes of shade -- namely "shade" (legacy) and (say) "shade2". The latter's Thing handler would extend the former's having 99% the same functionality, but just overriding percent inversion code in handleCommand() and updateBindingStates(). Then all new shades could be created as instances of "shade2" whereas legacy shades would have already been created as instances of "shade".

We might still have a config param on "shade" which, if set, would force `migrateThingType()` if the config param is expressly set. Or something like that... ??

```
    /**
     * Informs the framework that the ThingType of the given {@link Thing} should be changed.
     *
     * @param thing thing that should be migrated to another ThingType (must not be null)
     * @param thingTypeUID the new type of the thing (must not be null)
     * @param configuration a configuration that should be applied to the given {@link Thing}
     */
    void migrateThingType(Thing thing, ThingTypeUID thingTypeUID, Configuration configuration);
```
Note: this would require a bit of extra work on the ReadMe..

arroyoj:
Having two classes of nearly identical shades seems a bit overkill, especially for a feature that probably affects a very small set of users. Users with single rail shades, which I'm guessing is the majority, would see two types of shades that are actually identical for them, and it seems more confusing. A config setting that explicitly applies only to the secondary rail should be clear.

Have you tried seeing the default value in thing-types.xml? It seems like that should solve case 6, and then the logic of legacy mode should be something like this:
```legacySecondary = !(Boolean.FALSE.equals(getConfigAs(HDPowerViewShadeConfiguration.class).legacySecondaryMode));```
In the code above, I negated what was previously called newSecondaryMode, and renamed it to reflect legacy mode.

**Comment by andrewfg** [comment]
> Have you tried setting the default value in thing-types.xml

Yes. It forces null values to the default value. So it fixes 6. but breaks 1. 2. 3.

**Comment by andrewfg** [comment]
A few updates as follows. Please let me know how you get on.
- Fixed the `newShadeMode` null/false/true logic.
- Implemented a second Thing type `shade2` which overrides `shade` to invert the percentages.
- The JAR and ReadMe are on the same links as in my prior post above.
- This resolves all of the creation / migration cases 1. thru 6. above.
- If your `shade` things were created in the UI you can use the UI to set `newSecondaryMode` to migrate them the `shade2`.
- If your `shade` things were created in a .things file you can edit it to change `shade` to `shade2` (or add newSecondaryMode="true").

**Comment by arroyoj** [comment]
Sorry I didn't get a chance to respond earlier. I was going to suggest that we simplify this by making "legacy" the default config value for now (even with newly created Things) and having a period of "opt in" to the new behavior, at which point we could more slowly transition from legacy to new.

In any case, I won't be able to test it until tonight, but looking at the code, I'm >90% sure that "shade2" is actually the legacy behavior, not "new". The current dev code since #11552 was merged is actually "new" behavior. "Legacy" is essentially reverting that merge to v3.1.0 behavior. Again, when I was testing the earlier JAR from this PR, when legacyMode was false was when I actually got legacy behavior, but I'll test more tonight.

**Comment by andrewfg** [comment]
> I'm >90% sure that "shade2" is actually the legacy behavior, not "new"

I can't test it so you need to let me know. The only difference is in `fixupPosition()` as follows (and these can obviously be swapped if you want).

### Shade

```
    protected int fixupPosition(PercentType position) {
        // Legacy Secondary Mode: return the position value.
        return position.intValue();
    }
```

### Shade2

```
    protected int fixupPosition(PercentType position) {
        // New Secondary Mode: return the inverse of the position value.
        return 100 - position.intValue();
    }
```

**Comment by andrewfg** [comment]
Just for the avoidance of doubt:
- In the legacy version the setter and the getter were opposite
- In the last bug fix we changed the setter to match the getter

So when you say the current PR is reversed for legacy vs. new mode, you must tell me very specifically whether you are testing vs. the getter or the setter.

**Comment by arroyoj** [comment]
@andrewfg, sorry, I should have been more clear about what I meant by "legacy" mode. I consider "legacy" to be what is described in the new ReadMe as "legacy", which corresponds to the v3.1.0 "getter".

I tested the latest JAR and found the issues described below. I'd also like to suggest that we go back to having a single "shade" Thing and just a config param, but switch to "legacy" as the default. Details below.

1. Shade2 is behaving like "legacy"
Using the updated ReadMe from this PR as a reference, the "shade2" Thing is behaving like a "legacy" shade (same as v3.1.0 getter). The "shade" Thing is behaving like the "new" shade (same as v3.1.0 setter). As you mentioned above, though, this would be an easy fix of flipping `fixupPosition` for shade vs shade2.


2. Shade --> Shade2 conversion cannot easily be reversed
Setting the "New Secondary Mode" config to True successfully converted a "shade" Thing to a "shade2" Thing. However, once done, there was no way to go backward to the other behavior. I think from a user perspective, that could be confusing. In all other cases, boolean config parameters are reversible. If you flip the switch and then are not happy with the new behavior, or just do it by accident, you have to delete the Thing and add it back in to get back the old behavior. Actually, you have to add it in manually if shade2 is created by auto discovery and you don't want that.


3. Shade2 doesn't work
This is certainly the biggest issue. After I converted one of my shades to a "shade2" Thing, it stopped responding to commands. Here is a copy of some logs (I haven't run TRACE logs yet). Note: LivingRoom_Shade5 is a "shade2" (not working), while "LivingRoom_Shade4" is still a "shade" (working):
```
2021-12-08 20:04:04.975 [INFO ] [openhab.event.ItemCommandEvent      ] - Item 'LivingRoom_Shade5_Top' received command 20
2021-12-08 20:04:04.976 [INFO ] [penhab.event.ItemStatePredictedEvent] - Item 'LivingRoom_Shade5_Top' predicted to become 0
2021-12-08 20:04:11.050 [INFO ] [openhab.event.ItemCommandEvent      ] - Item 'LivingRoom_Shade5_Top' received command 21
2021-12-08 20:04:11.051 [INFO ] [penhab.event.ItemStatePredictedEvent] - Item 'LivingRoom_Shade5_Top' predicted to become 0
2021-12-08 20:04:19.506 [INFO ] [openhab.event.ItemCommandEvent      ] - Item 'LivingRoom_Shade5_Bottom' received command 80
2021-12-08 20:04:19.507 [INFO ] [penhab.event.ItemStatePredictedEvent] - Item 'LivingRoom_Shade5_Bottom' predicted to become 100
2021-12-08 20:04:43.869 [INFO ] [openhab.event.ItemCommandEvent      ] - Item 'LivingRoom_Shade5_Top' received command 20
2021-12-08 20:04:43.870 [INFO ] [penhab.event.ItemStatePredictedEvent] - Item 'LivingRoom_Shade5_Top' predicted to become 0
2021-12-08 20:05:00.910 [INFO ] [openhab.event.ItemCommandEvent      ] - Item 'LivingRoom_Shade4_Top' received command 80
2021-12-08 20:05:00.911 [INFO ] [penhab.event.ItemStatePredictedEvent] - Item 'LivingRoom_Shade4_Top' predicted to become 80
2021-12-08 20:05:00.922 [INFO ] [openhab.event.ItemStateChangedEvent ] - Item 'LivingRoom_Shade4_Top' changed from 97 to 80
2021-12-08 20:05:57.431 [INFO ] [openhab.event.ItemCommandEvent      ] - Item 'LivingRoom_Shade5_Top' received command UP
2021-12-08 20:05:57.432 [INFO ] [penhab.event.ItemStatePredictedEvent] - Item 'LivingRoom_Shade5_Top' predicted to become 0
```
Basically, with a shade2, the command is received, but nothing happens after that. The shade does not physically move, and the state never changes. I tried both top (secondary) and bottom (primary) rails, and both PercentType and UpDownType commands, but neither worked. I restarted OpenHAB to make sure this wasn't just from the Thing conversion, but the "shade2" Thing still didn't work.


### Suggestion - single shade Thing, but default to "legacy"

Rather than fixing the shade2 issues above, I would suggest we go back to a single shade Thing and the `legacySecondaryMode` config parameter. However, to make the logic and path forward easier, I think we should go with "legacySecondaryMode" defaulting to TRUE in all cases. That way, cases 1-6 would all work (a "legacy" shade would be created by default automatically or manually, and existing shades would be annotated as "legacy" by setting the default to TRUE in thing-types.xml). All users by default would move forward with current "getter" behavior, which is what would be expected for anyone not explicitly using the PercentType setter. Users like me who want the new behavior could manually enable the new behavior by explicitly setting `legacySecondaryMode` to FALSE. Then, in the next version *after* the coming v3.2, we could make a clean transition to the "new" behavior (which I still think is "better"). That transition would be easier, though, because there would be no "null" legacy case to deal with, just an explicit old `legacySecondaryMode` = TRUE. For the updated ReadMe, we could swap the legacy table rows and the main table rows for the secondary channel, and have a new "preview" table for the modern behavior describing what would happen if `legacySecondaryMode` is disabled. Thoughts?

I'm also hoping that by simplifying it, we can make it more likely this makes it in before v3.2 is released. I think it is important there is a smooth upgrade path for v3.2 that minimizes the disruption to current users, and defaulting to "legacy" might be the best bet given the implementation challenges.

andrewfg:
Ok I agree with your suggestion above. I will revert shade2 and push a new commit today.

On a procedural note: this PR builds on the previous bug fix and database PR, so it looks much more messy than it should do. I would like to clean that up, so it would help to get the prior PR merged asap. So could you formally close your review with LGTM, and hopefully we can motivate one of the maintainers to second it.

andrewfg:
> Using the updated ReadMe from this PR as a reference, the "shade2" Thing is behaving like a "legacy" shade (same as v3.1.0 getter). The "shade" Thing is behaving like the "new" shade (same as v3.1.0 setter).

@arroyoj I am still horribly concerned about your statement above. Below is a table of changes that we have recently made on the Setter and the Getter. Basically over the whole time span we never changed the Setter, and #11522 we reversed the Getter. Now if you claim that the current iteration has the legacy and new modes reversed, then it seems to that post ##1152 legacy mode is actually the one you were looking for all along, and that the new mode is not actually required. So can you please check again: 

- Is the behaviour of #11522 as you wanted it?
- Is the behaviour of #11698 as you wanted it?
- Is the behaviour of #11729 (legacy) as you wanted it?
- So do you need of #11729 (new) at all?

image

arroyoj:
Yes, this has become very confusing, and I'm sorry about that, but over all of these issues, I have been consistent in my definition of legacy vs new.

In #11552, we changed/broke the v3.1.x getter by making it match the setter. This prompted @jlaur to point out that his rules would need to be updated. Thus, I consider the original v3.1.x getter to be legacy. Changing the getter to match the setter was "new", even though the setter never changed. If you look at the ReadMe in this PR, the "legacy" table describes the v3.1.x getter. The main table describes the v3.1.x setter. I have been trying to get the code to match this ReadMe.

The confusion is that what you are currently calling "new mode" in your table, I have been consistently been calling "legacy".

To answer your specific questions:
> Is the behaviour of #11552 as you wanted it?

Yes. That was exactly what I wanted.

> Is the behaviour of #11698 as you wanted it?

Yes.

> Is the behaviour of #11729 (legacy) as you wanted it?

What you are calling legacy (and I'm calling new) is what I wanted.

> So do you need of #11729 (new) at all?

What you are calling new (and I am calling legacy) is needed for backwards compatibility for situations like @jlaur's rules. Personally, I do not need it, but others probably do.

For my uses, I would be fine without this PR, but I suspect many users would think their system switched on them. They might be fine with the new behavior after #11552, but it would be a noticeable change that could appear broken at first.

In my mind, the getter defines the legacy behavior. If you change your shades outside OH and then check OH, you *only* see the getter behavior. In v3.1, if you set your shade in OH, walk away, and then check 5 min later, you see the getter behavior. Therefore, I think v3.1 getter is legacy. Since the setter and getter in v3.1 were opposite, only one can be legacy, and I think it should be the getter.

@jlaur, as another dual rail shade user, I'd love to get your thoughts on what is "legacy".

andrewfg:
Ok. But then I think the real title of this PR is "add option to allow users to revert to pre #11552 behaviour". Or?

arroyoj:
Thanks @andrewfg.

Assuming we go forward with my definition of "legacy" and make it the default in all cases, then how about something liket this as the PR title: "Revert secondary position to v3.1 getter behavior and provide transition path to corrected position mapping". If "corrected" is too strong, we could just say "new". Or another idea: "Revert secondary position to v3.1 getter behavior and provide option for new mapping".

Then, after v3.2 is released (assuming this makes it in), it would be straight forward to make the v3.3 default be the same as #11552 and provide a clean transition from the old default.

andrewfg:
No. PR #11552 was a bug fix. So the correct behaviour has (apparently) now been achieved. So if we implement a new PR then it is a reversion to a buggy situation. It may be simplest to just add a footnote in the ReadMe that there was a bug prior to v3.2 and that it’s fix might change the behaviour of pre-existing installations.

andrewfg:
So my (hopefully final) proposal is to cancel this current PR and instead simply incorporate a modified version of the ReadMe (see it [HERE](https://github.com/andrewfg/openhab-addons/tree/hdpowerview-change-up-down/bundles/org.openhab.binding.hdpowerview)) into our other currently open PR #11698 => Is that OK with you?

arroyoj:
That makes sense. I was thinking of #11552 as a bug fix that could have been resolved in two different ways. While I'm personally happy with how it was resolved, the subsequent discussion made me wonder whether the other way would have been less jarring to users. In any case, I'm fine with the suggestion to just note it in the Readme.

In retrospect, I think "legacy" became too confusing of a term, because the true legacy was a broken getter/setter combo that clearly shouldn't be maintained. What if we wrap up this PR with Readme changes only, so the Readme matches the #11552 code, like you proposed above. Then, after v3.2 is released, we could consider a separate PR providing a config option for an "alternative" secondary position mapping (ie, no judgement on which is better or legacy vs modern, just giving the user options). In that new PR, the "enable alternative" config option would default to FALSE in all 6 cases you previously described, but any user who was dissatisfied with the bug fix could manually enable it.

Of course, if you'd rather just leave it all alone after this, I'm fine with that, too. My shades have been great since #11552 and #11698. I opened #11593 primarily as a doc fix, and it seems like that's what this PR will be, so I'm good. Thanks!

arroyoj:
> So my (hopefully final) proposal is to cancel this current PR and instead simply incorporate a modified version of the ReadMe (see it [HERE](https://github.com/andrewfg/openhab-addons/tree/hdpowerview-change-up-down/bundles/org.openhab.binding.hdpowerview)) into our other currently open PR #11698 => Is that OK with you?

Sorry, didn't see your comment when I wrote mine. Yes, totally fine with that. Happy to double check the updated readme, but that sounds like a good plan. Could still consider an option for alternative mapping, but that would definitely be post-v3.2.

andrewfg:
Closing by incorporating new ReadMe into #11698

PR #11730, opened 2021-12-07
Update README.md

raveydavies:
Changed FanState channel to Switch and corrected small typing error

[venstarthermostat] README.md update to change fanState to Switch

This corrects documentation following recent binding update in which fanState was added as a Switch but incorrectly documented as a String.


raveydavies:
@lolodomo I deleted the old PR because I was having a nightmare signing it off. It worked this time, although I did get the same DCO error again and needed to correct it. Not sure what is going on.

lolodomo (approved):
LGTM

PR #11734, opened 2021-12-08
[nzwateralerts] Fix smart water alert level.

Cossey:
Small fix for smartwater.org.nz displaying a new image when water needs to be conserved (restriction season). This tweak will fix `UNDEF` (it should be `0`).


lolodomo (approved):
LGTM

PR #11737, opened 2021-12-08
[nikohomecontrol] Fix dimmer control sequence

### LINKED ISSUE #11695: [nikohomecontrol] When a dimmable light circuit is set from off to a value, the previous value is also received
## Expected Behavior
When setting a dimmable light value, this should be the only value received by the binding.

## Current Behavior
When using a condition in Niko Home Control to set a dimmable light circuit to a certain value, the latest known value of the dimmable light circuit will be received on the binding only milliseconds before the actual value. Example:

When motion detected by PIR, set value of dimmable light to 60%. (In this example, the dimmer memory is disabled, thus the default value is always 100).
The following is logged in the binding:
```
2021-12-03 18:01:17.801 [DEBUG] [econtrol.internal.protocol.NhcAction] - update channel state for 9a297128-9ad6-41d7-9e2f-f341832d21fa with 100
2021-12-03 18:01:17.803 [DEBUG] [l.nhc2.NikoHomeControlCommunication2] - setting action 9a297128-9ad6-41d7-9e2f-f341832d21fa internally to ON
2021-12-03 18:01:17.805 [DEBUG] [econtrol.internal.protocol.NhcAction] - update channel state for 9a297128-9ad6-41d7-9e2f-f341832d21fa with 60
2021-12-03 18:01:17.818 [DEBUG] [l.nhc2.NikoHomeControlCommunication2] - setting action 9a297128-9ad6-41d7-9e2f-f341832d21fa internally to 60
```

## Possible Solution
Is this a bug in Niko Home Control? Then I will report it to them.

Otherwise only react on latest received value on dimmer within 15(?) milliseconds.

## Your Environment
Niko Home Control 2.13.31085
openHAB 3.2.0.M4

@mherwege

mherwege:

Closes #11695

The sequence of on/off and brightness commands and status updates caused flickering with lamps on other binding channels following the binding channels. This fix adjusts the sequence to avoid setting the wrong brightness value for a short instance before setting the correct value.

### DIFF
--- a/bundles/org.openhab.binding.nikohomecontrol/src/main/java/org/openhab/binding/nikohomecontrol/internal/protocol/nhc2/NikoHomeControlCommunication2.java
+++ b/bundles/org.openhab.binding.nikohomecontrol/src/main/java/org/openhab/binding/nikohomecontrol/internal/protocol/nhc2/NikoHomeControlCommunication2.java
@@ -480,7 +480,7 @@ public class NikoHomeControlCommunication2 extends NikoHomeControlCommunication
             booleanState = basicStateProperty.get().basicState;
         }
 
-        if ((booleanState != null) && NHCOFF.equals(booleanState)) {
+        if (NHCOFF.equals(booleanState)) {
             action.setBooleanState(false);
             logger.debug("setting action {} internally to OFF", action.getId());
         }
@@ -488,12 +488,16 @@ public class NikoHomeControlCommunication2 extends NikoHomeControlCommunication
         if (dimmerProperty.isPresent()) {
             String brightness = dimmerProperty.get().brightness;
             if (brightness != null) {
-                action.setState(Integer.parseInt(brightness));
-                logger.debug("setting action {} internally to {}", action.getId(), dimmerProperty.get().brightness);
+                try {
+                    action.setState(Integer.parseInt(brightness));
+                    logger.debug("setting action {} internally to {}", action.getId(), dimmerProperty.get().brightness);
+                } catch (NumberFormatException e) {
+                    logger.debug("received invalid brightness value {} for dimmer {}", brightness, action.getId());
+                }
             }
         }
 
-        if ((booleanState != null) && NHCON.equals(booleanState)) {
+        if (NHCON.equals(booleanState)) {
             action.setBooleanState(true);
             logger.debug("setting action {} internally to ON", action.getId());
         }
@@ -505,7 +509,7 @@ public class NikoHomeControlCommunication2 extends NikoHomeControlCommunication
                 action.setState(Integer.parseInt(position));
                 logger.debug("setting action {} internally to {}", action.getId(), position);
             } catch (NumberFormatException e) {
-                logger.trace("received empty rollershutter {} position info", action.getId());
+                logger.trace("received empty or invalid rollershutter {} position info {}", action.getId(), position);
             }
         });
     }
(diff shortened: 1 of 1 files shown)


bundles/org.openhab.binding.nikohomecontrol/src/main/java/org/openhab/binding/nikohomecontrol/internal/protocol/nhc2/NikoHomeControlCommunication2.java
```diff
@@ -480,14 +480,9 @@ private void updateLightState(NhcAction2 action, List<NhcProperty> devicePropert
...
 
-        if (booleanState != null) {
-            if (NHCON.equals(booleanState)) {
-                action.setBooleanState(true);
-                logger.debug("setting action {} internally to ON", action.getId());
-            } else if (NHCOFF.equals(booleanState)) {
-                action.setBooleanState(false);
-                logger.debug("setting action {} internally to OFF", action.getId());
-            }
+        if ((booleanState != null) && NHCOFF.equals(booleanState)) {
```
lolodomo, line 483: You can suppress your first test condition as it is already covered by your second condition.

bundles/org.openhab.binding.nikohomecontrol/src/main/java/org/openhab/binding/nikohomecontrol/internal/protocol/nhc2/NikoHomeControlCommunication2.java
```diff
@@ -497,6 +492,11 @@ private void updateLightState(NhcAction2 action, List<NhcProperty> devicePropert
                 logger.debug("setting action {} internally to {}", action.getId(), dimmerProperty.get().brightness);
             }
         }
+
+        if ((booleanState != null) && NHCON.equals(booleanState)) {
```
lolodomo, line 496: You can suppress your first test condition as it is already covered by your second condition.

bundles/org.openhab.binding.nikohomecontrol/src/main/java/org/openhab/binding/nikohomecontrol/internal/protocol/nhc2/NikoHomeControlCommunication2.java
```diff
@@ -627,6 +631,9 @@ public void executeAction(String actionId, String value) {
                 } else if (NHCOFF.equals(value)) {
                     property.status = value;
                 } else {
+                    action.setState(Integer.parseInt(value)); // set cached state to new brightness value to avoid
```
lolodomo, line 634: No need to catch exception?

lolodomo (changes requested):

bundles/org.openhab.binding.nikohomecontrol/src/main/java/org/openhab/binding/nikohomecontrol/internal/protocol/nhc2/NikoHomeControlCommunication2.java
mherwege: It would be a bug if the value is not an int here, but I will put in a catch and log it (and do it at a few more places in the same class when I am at it).

mherwege:
@lolodomo Thanks for your review. Comments should have been addressed.

lolodomo (approved):
LGTM

PR #11739, opened 2021-12-09
[hpprinter] Change status channel from hash map values to state description options.

Cossey:
This changes how both the status channels work. Originally, they referenced a hash table with the name of the status value. It seems more appropriate to have these as state description options as it would allow the statuses to be localised.


lolodomo (approved):
LGTM

PR #11746, opened 2021-12-10
[plugwise] Fix 'power' channel not correctly updated with power production

wborn:
This fixes the issue that the 'power' channel would not update with the correct state because the number of pulses in the `PowerInformationResponseMessage` is signed instead of unsigned.

When the binding detected these strange readings it would normally log: "Circle (...) is in a kind of error state ...".

See also: https://community.openhab.org/t/plugwise-cirle-connected-to-pv-panel-gives-no-power-data-in-oh3/123001


BasvanH:
Thank you @wborn for the fix, I'm still busy to capture the packets for the energy part.

wborn:
...and thank you for testing the changes! :+1: 
I think that also warrants for a "community approved" label. :wink:

lolodomo:
I am not sure about the cast to short but if it is tested and approved, that is fine for me.

lolodomo (approved):
LGTM

PR #11751, opened 2021-12-11
[nikohomecontrol] Added ability to directly use virtual flags as switches

### LINKED ISSUE #11740: [nikohomecontrol] Ability to add virtual flags without the need for a basic action
## Expected behavior
The ability to add virtual devices without the need for a basic action to be assigned.

## Current behavior
When having a PIR device added in OpenHAB, that PIR does not emit motion events. It's only possible to control as a push button. The PIR in Niko Home Control requires a device to control. So because the PIR, that is available in the OpenHAB binding, does not emit motion events, then one need to add the device that the PIR controls. When one sets this to a virtual flag, without a basic action, because one does not want this to show up as a switch in either the Niko Home App or Google Home, then it is not possible to add this virtual flag to OpenHAB.

image

## Proposed solution
Allow the virtual flag to be added from the inbox.

The virtual flag, that has no basic action, seems to be available in the `received topic hobby/control/devices/evt`:
```
{
	"Name": "PIR",
	"Traits": [],
	"Model": "flag",
	"Type": "virtual",
	"Uuid": "c69b4f89-681e-4894-9871-6606e7802d96",
	"Identifier": "1e4482e6-e241-4be6-b808-ade90b64ad37",
	"Parameters": [
		{
			"LocationName": "Walk-in"
		},
		{
			"LocationId": "6941a084-5383-41e4-9df4-2df1191798f2"
		},
		{
			"LocationIcon": "dressing"
		}
	],
	"Technology": "nikohomecontrol"
}
```
And according to the logs, the binding does in fact receives updates from the virtual flag with no basic action, when in this example motion is detected by the PIR and it turns on the virtual flag:
```
2021-12-09 19:22:01.762 [TRACE] [l.nhc2.NikoHomeControlCommunication2] - received topic hobby/control/devices/evt, payload {"Method":"devices.status","Params":[{"Devices":[{"Properties":[{"Status":"True"}],"Uuid":"c69b4f89-681e-4894-9871-6606e7802d96"}]}]}
```

I have tried manually adding the device and refer to the Uuid in the Action ID, but the binding says `Configured action ID does not match an action in controller`

@mherwege

mherwege:
Closes #11740

Virtual flags are now directly visible in OH, without needing a workaround of creating free start/stop actions in NHC.

### DIFF
--- a/bundles/org.openhab.binding.nikohomecontrol/src/main/java/org/openhab/binding/nikohomecontrol/internal/protocol/NikoHomeControlConstants.java
+++ b/bundles/org.openhab.binding.nikohomecontrol/src/main/java/org/openhab/binding/nikohomecontrol/internal/protocol/NikoHomeControlConstants.java
@@ -35,6 +35,9 @@ public class NikoHomeControlConstants {
     public static final String NHCON = "On";
     public static final String NHCOFF = "Off";
 
+    public static final String NHCTRUE = "True";
+    public static final String NHCFALSE = "False";
+
     public static final String NHCTRIGGERED = "Triggered";
 
     // rollershutter constants in the Nhc layer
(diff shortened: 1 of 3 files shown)


lolodomo (approved):
LGTM

PR #11753, opened 2021-12-11
Add default translations for io add-ons

wborn:

### DIFF
--- a/bundles/org.openhab.io.neeo/src/main/resources/OH-INF/i18n/neeo.properties
+++ b/bundles/org.openhab.io.neeo/src/main/resources/OH-INF/i18n/neeo.properties
@@ -6,3 +6,7 @@ io.config.neeo.exposeNeeoBinding.label = Expose NEEO Binding
 io.config.neeo.exposeNeeoBinding.description = Expose things found by the NEEO Binding
 io.config.neeo.searchLimit.label = Search Limit
 io.config.neeo.searchLimit.description = The maximum number of results to return for a search
+
+# service
+
+service.io.neeo.label = NEEO Integration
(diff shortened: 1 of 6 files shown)


yfre (approved):
looks good for homekit

bundles/org.openhab.io.neeo/src/main/resources/OH-INF/i18n/neeo.properties
```diff
@@ -0,0 +1,8 @@
+io.config.neeo.checkStatusInterval.label = Check Status Interval (seconds)
+io.config.neeo.checkStatusInterval.description = The interval (in seconds) to check the status of the brain
+io.config.neeo.exposeAll.label = Expose All
+io.config.neeo.exposeAll.description = Expose all things
+io.config.neeo.exposeNeeoBinding.label = Expose NEEO Binding
+io.config.neeo.exposeNeeoBinding.description = Expose things found by the NEEO Binding
+io.config.neeo.searchLimit.label = Search Limit
+io.config.neeo.searchLimit.description = The maximum number of results to return for a search
```
lolodomo, line 8: We don't have an entry for the binding label

lolodomo (changes requested):
Note that there is nothing for the ImperiHome integration service.

bundles/org.openhab.io.neeo/src/main/resources/OH-INF/i18n/neeo.properties
wborn: I have made a fix for it @lolodomo!

wborn:
> Note that there is nothing for the ImperiHome integration service.

It doesn't seem to have any configuration parameters:

https://github.com/openhab/openhab-addons/blob/main/bundles/org.openhab.io.imperihome/src/main/resources/OH-INF/config/config.xml

But it is a configurable service:

https://github.com/openhab/openhab-addons/blob/cafed142e7dfac916eda8e6d893b5817a4a89e0b/bundles/org.openhab.io.imperihome/src/main/java/org/openhab/io/imperihome/internal/ImperiHomeApiServlet.java#L66

Looks like it can all be removed and readded when neccessary in the future?

lolodomo:
Ok, so we can probably forget Imperihome.

lolodomo (approved):
LGTM

PR #11754, opened 2021-12-11
Add default translations for persistence add-ons

wborn:


lolodomo (approved):
LGTM

lolodomo:
Note that some persistence services still have no file, those without any configuration parameters.

PR #11755, opened 2021-12-11
Add default translations for transform add-ons

wborn:


lolodomo (approved):
LGTM

PR #11757, opened 2021-12-11
[urtsi] Fixed leading spaces in translations

cweitkamp:
- Fixed leading spaces in translations

### DIFF
--- a/bundles/org.openhab.binding.urtsi/src/main/resources/OH-INF/i18n/urtsi.properties
+++ b/bundles/org.openhab.binding.urtsi/src/main/resources/OH-INF/i18n/urtsi.properties
@@ -1,18 +1,22 @@
 # binding
+
+binding.urtsi.name = Somfy URTSI II Binding
+
+# binding
+
 bindingDescription = This is the binding for Somfy Universal RTS Interface II (URTSI II).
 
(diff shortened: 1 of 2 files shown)


bundles/org.openhab.binding.urtsi/src/main/resources/OH-INF/i18n/urtsi.properties
```diff
@@ -1,18 +1,22 @@
 # binding
+
+binding.urtsi.name = Somfy URTSI II Binding
```
lolodomo, line 3: I suggest a change in binding.xml to use the same approach for label and description, either both using `@text` or none.

lolodomo (changes requested):

bundles/org.openhab.binding.urtsi/src/main/resources/OH-INF/i18n/urtsi.properties
cweitkamp: Done.

lolodomo (approved):
LGTM

lolodomo:
@cweitkamp : thank you, I let you fix the German translation (leading spaces).

PR #11758, opened 2021-12-11
Add GitHub Action Annotations for compiler warnings

fwolter:
Signed-off-by: Fabian Wolter <github@fabian-wolter.de>

### DIFF
--- a/.github/openhab-compile-problems.json
+++ b/.github/openhab-compile-problems.json
@@ -16,6 +16,19 @@
           "loop": true
         }
       ]
+    },
+    {
+      "owner": "openhab-compile-warnings",
+      "severity": "warning",
+      "pattern": [
+        {
+          "regexp": "^\\[WARNING\\] (.+\\.java):\\[(.+),(.+)\\] (.*)$",
+          "file": 1,
+          "line": 2,
+          "col": 3,
+          "message": 4
+        }
+      ]
     }
   ]
 }


wborn:
It will probably annotate all files with warnings instead of only changed lines. That's why only annotation were added for errors.

fwolter:
It reports the warnings for all addons that are built. As most of the time only one addon is built in a particular PR, there will only be warnings for one addon. See https://github.com/fwolter/openhab-addons/actions/runs/1567745895

fwolter:
Ah, you mean unchanged files in the addon, which has changes. Yes, that's right. But I think that's not too bad, if the codeowner sees all warnings of his/her binding in the PR. Maybe the 10 warnings limit is prioritized to start with the changed files? I take a look.

fwolter:
There can be 10 warnings per step and 50 warnings per job. See https://github.community/t/annotation-limitation/17998 Compiler, Checkstyle, PMD and Spotbugs are separate steps, so there will be room for some more warnings. I think that's OK for a single binding. For sure it won't be enough when all addons are built.

I think it's better to have warnings for all files of a binding, which might be truncated therefore, than no warnings at all.

wborn:
> I think that's OK for a single binding. For sure it won't be enough when all addons are built.

I think it will be annoying and everybody will just start ignoring all these unrelated warnings which will then also clutter up reviews.

fwolter:
I find it makes code reviews easier and more efficient. Especially to see the SAT warnings which are newly introduced in contrast to the existing. You'll see it at a glance with the annotations:
grafik
Maybe another maintainer can join the discussion @openhab/add-ons-maintainers  If I'm alone with my opinion it's okay to close this.

fwolter:
Seems like I'm alone with my opinion. That's ok.

kaikreuzer:
Hey @fwolter, sorry for not commenting earlier. I definitely also see value in annotating warnings, but knowing the code base and the amount of warnings we still have in there, I'm also a bit worried that it might be too distracting/annoying and thus think that we should go without at least for now.

PR #11766, opened 2021-12-12
 [denonmarantz] Fix xml unmarshalling exception

pieterbuts:
[denonmarantz] Fix xml unmarshalling problem

This bugfix fixes issue 11502.

XML unmarshalling failed with an exception which had as inner exception: javax.xml.bind.UnmarshalException: Namespace URIs and local names to the unmarshaller needs to be interned.

The decapitalization of attirbutes and names in PropertyRenamerDelegate caused these elements to be not interned.


lolodomo:
PR not correctly signed.
Making an exception as the change is small and the delay for 3.2 is now very reduced,

lolodomo (approved):

PR #11768, opened 2021-12-12
[hdpowerview] Update positions after triggering scene/scene group

### LINKED ISSUE #11697: [hdpowerview] Update position Channels immediately when a Scene (Collection) is triggered.
In [Issue #11557](https://github.com/openhab/openhab-addons/issues/11557) @arroyoj suggested the following:

> I would .. suggest doing a soft refresh after any scene is triggered. As far as I can tell from my testing, the hub responds to a `GET /shades` call with the correct JSON data for all the shade states immediately after a scene is triggered, even while the shades are still moving, so a soft refresh should be sufficient.

In addition logically a soft refresh should probably also be done when a **_Scene Collection_** is triggered (see merged PR #11534 from @jlaur).

jlaur:
When triggering a scene or scene group, some shades will have new positions. These new positions are available in the hub immediately after triggering the scene or scene group, i.e. before the shades will reach those new positions.

Previously it would take up to one minute to have the positions refreshed. Now it will happen a few seconds after the scene or scene group has been triggered.

Fixes #11697

### Test documentation

Before change:
```
2021-12-12 21:26:42.228 [INFO ] [openhab.event.ItemCommandEvent      ] - Item 'PowerViewHub_Scene_Office_Up' received command ON
2021-12-12 21:27:08.782 [INFO ] [openhab.event.ItemStateChangedEvent ] - Item 'Blind9_Position' changed from 100 to 0

```
After change:
```
2021-12-12 21:50:22.834 [INFO ] [openhab.event.ItemCommandEvent      ] - Item 'PowerViewHub_Scene_Office_Up' received command ON
2021-12-12 21:50:26.105 [INFO ] [openhab.event.ItemStateChangedEvent ] - Item 'Blind9_Position' changed from 100 to 0

```
And back:
```
2021-12-12 21:50:29.706 [INFO ] [openhab.event.ItemCommandEvent      ] - Item 'PowerViewHub_Scene_Office_Down' received command ON
2021-12-12 21:50:32.955 [INFO ] [openhab.event.ItemStateChangedEvent ] - Item 'Blind9_Position' changed from 0 to 100

```

### DIFF
--- a/bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/handler/HDPowerViewHubHandler.java
+++ b/bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/handler/HDPowerViewHubHandler.java
@@ -113,7 +113,7 @@ public class HDPowerViewHubHandler extends BaseBridgeHandler {
 
     @Override
     public void handleCommand(ChannelUID channelUID, Command command) {
-        if (RefreshType.REFRESH.equals(command)) {
+        if (RefreshType.REFRESH == command) {
             requestRefreshShadePositions();
             return;
         }
@@ -129,14 +129,14 @@ public class HDPowerViewHubHandler extends BaseBridgeHandler {
                 throw new ProcessingException("Web targets not initialized");
             }
             int id = Integer.parseInt(channelUID.getIdWithoutGroup());
-            if (sceneChannelTypeUID.equals(channel.getChannelTypeUID()) && OnOffType.ON.equals(command)) {
+            if (sceneChannelTypeUID.equals(channel.getChannelTypeUID()) && OnOffType.ON == command) {
                 webTargets.activateScene(id);
                 pollShades();
-            } else if (sceneGroupChannelTypeUID.equals(channel.getChannelTypeUID()) && OnOffType.ON.equals(command)) {
+            } else if (sceneGroupChannelTypeUID.equals(channel.getChannelTypeUID()) && OnOffType.ON == command) {
                 webTargets.activateSceneCollection(id);
                 pollShades();
             } else if (automationChannelTypeUID.equals(channel.getChannelTypeUID())) {
-                webTargets.enableScheduledEvent(id, OnOffType.ON.equals(command));
+                webTargets.enableScheduledEvent(id, OnOffType.ON == command);
             }
         } catch (HubMaintenanceException e) {
             // exceptions are logged in HDPowerViewWebTargets
--- a/bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/handler/HDPowerViewHubHandler.java
+++ b/bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/handler/HDPowerViewHubHandler.java
@@ -155,7 +155,7 @@ public class HDPowerViewHubHandler extends BaseBridgeHandler {
                 logger.warn("Bridge returned a bad JSON response: {}", e.getMessage());
             } catch (HubProcessingException e) {
                 logger.warn("Error connecting to bridge: {}", e.getMessage());
-                updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.BRIDGE_OFFLINE, e.getMessage());
+                updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.COMMUNICATION_ERROR, e.getMessage());
             } catch (HubMaintenanceException e) {
                 // exceptions are logged in HDPowerViewWebTargets
             }
--- a/bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/handler/HDPowerViewHubHandler.java
+++ b/bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/handler/HDPowerViewHubHandler.java
@@ -131,8 +131,10 @@ public class HDPowerViewHubHandler extends BaseBridgeHandler {
             int id = Integer.parseInt(channelUID.getIdWithoutGroup());
             if (sceneChannelTypeUID.equals(channel.getChannelTypeUID()) && OnOffType.ON.equals(command)) {
                 webTargets.activateScene(id);
+                pollShades();
             } else if (sceneGroupChannelTypeUID.equals(channel.getChannelTypeUID()) && OnOffType.ON.equals(command)) {
                 webTargets.activateSceneCollection(id);
+                pollShades();
             } else if (automationChannelTypeUID.equals(channel.getChannelTypeUID())) {
                 webTargets.enableScheduledEvent(id, OnOffType.ON.equals(command));
             }
--- a/bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/handler/HDPowerViewHubHandler.java
+++ b/bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/handler/HDPowerViewHubHandler.java
@@ -23,7 +23,6 @@ import java.util.Map;
 import java.util.Map.Entry;
 import java.util.StringJoiner;
 import java.util.concurrent.CopyOnWriteArrayList;
-import java.util.concurrent.Future;
 import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.TimeUnit;
 
@@ -87,7 +86,6 @@ public class HDPowerViewHubHandler extends BaseBridgeHandler {
     private long hardRefreshBatteryLevelInterval;
 
     private @Nullable HDPowerViewWebTargets webTargets;
-    private @Nullable Future<?> pollShadesOnceFuture;
     private @Nullable ScheduledFuture<?> pollFuture;
     private @Nullable ScheduledFuture<?> hardRefreshPositionFuture;
     private @Nullable ScheduledFuture<?> hardRefreshBatteryLevelFuture;
@@ -133,10 +131,10 @@ public class HDPowerViewHubHandler extends BaseBridgeHandler {
             int id = Integer.parseInt(channelUID.getIdWithoutGroup());
             if (sceneChannelTypeUID.equals(channel.getChannelTypeUID()) && OnOffType.ON == command) {
                 webTargets.activateScene(id);
-                pollShadesAsync();
+                pollShades();
             } else if (sceneGroupChannelTypeUID.equals(channel.getChannelTypeUID()) && OnOffType.ON == command) {
                 webTargets.activateSceneCollection(id);
-                pollShadesAsync();
+                pollShades();
             } else if (automationChannelTypeUID.equals(channel.getChannelTypeUID())) {
                 webTargets.enableScheduledEvent(id, OnOffType.ON == command);
             }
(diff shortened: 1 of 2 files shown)


jlaur:
FYI @andrewfg, @arroyoj

andrewfg:
I think equals() and == compile to the same thing. ??

andrewfg:
Also wondering if you should replace the pollshades() calls with scheduler.submit([ -> | pollshades()]) ??

jlaur:
> I think equals() and == compile to the same thing. ??

See https://stackoverflow.com/questions/1750435/comparing-java-enum-members-or-equals

So it's faster and safer at compile-time as well as runtime.

jlaur:
> Also wondering if you should replace the pollshades() calls with scheduler.submit([ -> | pollshades()]) ??

Done.

arroyoj:
@jlaur, thanks for putting this together. It all looks good to me. Do you have a JAR available so that I can try using it?

jlaur:
@arroyoj - of course: https://drive.google.com/file/d/15Wgt_cuNb3XUEbHejoAeEj31scXqgBCL/view?usp=sharing

arroyoj:
@jlaur, thanks for sharing the JAR. I tested it out on my machine, and it worked great. Similar to your testing, I saw that the shade positions got updated in ~4 seconds after triggering a scene, rather than up to 1 min like before. I didn't see any issues. Thanks again for this fix.

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/handler/HDPowerViewHubHandler.java
```diff
@@ -143,6 +147,21 @@ public void handleCommand(ChannelUID channelUID, Command command) {
         }
     }
 
+    private synchronized void pollShadesAsync() {
+        pollShadesOnceFuture = scheduler.submit(() -> {
```
lolodomo, line 151: No need to check if one is still scheduled ?

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/handler/HDPowerViewHubHandler.java
```diff
@@ -143,6 +147,21 @@ public void handleCommand(ChannelUID channelUID, Command command) {
...
+            try {
+                pollShades();
+            } catch (JsonParseException e) {
+                logger.warn("Bridge returned a bad JSON response: {}", e.getMessage());
+            } catch (HubProcessingException e) {
+                logger.warn("Error connecting to bridge: {}", e.getMessage());
+                updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.BRIDGE_OFFLINE, e.getMessage());
```
lolodomo, line 158: The detailed status is probably inappropriate.
If your bridge thing is really OFFLINE, if the thing handler is properly implemented, it will switch the thing status to OFLLINE / BRIDGE_OFFLINE elsewhere.
Maybe COMMUNICATION_ERROR is more appropriate ?

lolodomo (changes requested):

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/handler/HDPowerViewHubHandler.java
jlaur: Good point!

I'm in doubt about best approach here. Was reacting on @andrewfg's comment, since probably it's a bad idea to block the handler from `handleCommand`? But actually have mixed feelings about it. We already have blocking calls in same method for activating scenes etc. from same method, so it's already inconsistent. Only polling job is scheduled. Do you know if it's not allowed to have blocking network calls in `handleCommand`? It would certainly simplify the implementation if it wouldn't be needed.

It's fire-and-forget operations, so only keeping the reference for being cancelled correctly in dispose(), but now only cancels the last operation if multiple scenes are triggered very quickly as you've found.

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/handler/HDPowerViewHubHandler.java
jlaur: It seems logical, yes. I've changed to COMMUNICATION_ERROR. But currently the only place where bridge if marked BRIDGE_OFFLINE is from method `poll` under the same conditions, i.e. as soon as any call would return HTTP code != 200, no response, network timeout etc. This seems similar, but I'm fine with letting only the polling job take responsibility for this.

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/handler/HDPowerViewHubHandler.java
lolodomo: It depends what you mean by "blocking" :)
I think it is not allowed to have a really blocking call in `handleCommand`.
But you can run a HTTP request for example, without the need to create a separate thread.

jlaur: @lolodomo - I meant blocking as in I/O (network) operation, rendering thread blocked for possibly several hundreds milliseconds. :) Okay, if this is common, I'll revert commit d99aa701395ff28e29c259ee0ff01c45094c3fac to simplify the code and avoid unwanted behaviour.

jlaur: Reverted.

jlaur:
> Also wondering if you should replace the pollshades() calls with scheduler.submit([ -> | pollshades()]) ??

@andrewfg - please see thread from @lolodomo's comment. Decided to revert and go back to synchronous HTTP call for simplicity. Only difference is that we have two sync calls where previously we had one. With the previous changes we would have one sync and one async, and it became messy.

andrewfg:
> Decided to revert and go back to synchronous HTTP call for simplicity.

I don’t agree. I am not so much concerned about an extra HTTP call. What concerns me is that you are making re-entrant calls to the OH core; whilst you are still within handleCommand() for one channel, you run pollShades which makes multiple updateChannel callbacks to the core; in fact one call back from each channel and each shade in your system. So your handleCommand could potentially cause the OH core to block on itself. Not good.

jlaur:
> > Decided to revert and go back to synchronous HTTP call for simplicity.
> 
> I don’t agree. I am not so much concerned about an extra HTTP call. What concerns me is that you are making re-entrant calls to the OH core; whilst you are still within handleCommand() for one channel, you run pollShades which makes multiple updateChannel callbacks to the core; in fact one call back from each channel and each shade in your system. So your handleCommand could potentially cause the OH core to block on itself. Not good.

Ahh - thanks for this explanation. New approach needed, marking PR as draft.

andrewfg:
Perhaps your prior attempt was overkill; I understand why you had to wrap the asynch pollShades call in an exception handler, but perhaps it is easier to let the existing code do the heavy lifting rather than reinventing the wheel; so after handleCommand triggered the scene, you could just 'submit' a call to `poll()` as follows..

```
scheduler.submit(() -> {
    poll();
});
```

jlaur:
> Perhaps your prior attempt was overkill; I understand why you had to wrap the asynch pollShades call in an exception handler, but perhaps it is easier to let the existing code do the heavy lifting rather than reinventing the wheel; so after handleCommand triggered the scene, you could just 'submit' a call to `poll()` as follows..
> 
> ```
> scheduler.submit(() -> {
>     poll();
> });
> ```

`poll` will perform four calls in total:
- fetch shades
- fetch scenes
- fetch scene collections
- fetch scheduled events

and we need only one of them. Still, it would be much simpler like you described, so could be refactored into something like that. My concern is that if we schedule a fire-and-forget call like that, we might have a problem in case `dispose()` is called right after while operation is still ongoing in future, for which we don't have a hook for cancelling it?

jlaur:
> My concern is that if we schedule a fire-and-forget call like that, we might have a problem in case `dispose()` is called right after while operation is still ongoing in future, for which we don't have a hook for cancelling it?

Perhaps we could reschedule existing future as a compromise? We would still fetch everything, but at least we would postpone next ordinary poll for another minute:

```
    private void rescheduleSoftPoll() {
        ScheduledFuture<?> future = this.pollFuture;
        if (future != null) {
            future.cancel(false);
        }
        logger.debug("Rescheduling poll every {}ms", refreshInterval);
        this.pollFuture = scheduler.scheduleWithFixedDelay(this::poll, 0, refreshInterval, TimeUnit.MILLISECONDS);
    }
```

And we would have the future cancelled from `dispose`.

andrewfg:
I think there is no need to worry about that: firstly the sheduler is part of the core, so if the core is shutting down, it will do so cleanly (it won't start un-run tasks, and it won't interrupt already running ones); and secondly the poll method is marked as synchronized so it cannot be called in parrallel.

Obviously if you unpack the poll() method, you will need to think carefully about placing the synchronized marker on any of its parts, since you would be introducing a risk of self blocking between regular polls and this special poll.

In summary I would still just submit poll; who cares if its updating other channels? It is essentially just doing another regular poll a bit earlier than the normal polling cycle would do so..

andrewfg:
> Perhaps we could reschedule existing future

Yes, that would work.

jlaur:
> > Perhaps we could reschedule existing future
> 
> Yes, that would work.

Done.

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/handler/HDPowerViewHubHandler.java
```diff
@@ -129,12 +131,16 @@ public void handleCommand(ChannelUID channelUID, Command command) {
...
+                // Reschedule soft poll for immediate shade position update.
+                scheduleSoftPoll(0);
+            } else if (sceneGroupChannelTypeUID.equals(channel.getChannelTypeUID()) && OnOffType.ON == command) {
                 webTargets.activateSceneCollection(id);
+                // Reschedule soft poll for immediate shade position update.
+                scheduleSoftPoll(0);
             } else if (automationChannelTypeUID.equals(channel.getChannelTypeUID())) {
-                webTargets.enableScheduledEvent(id, OnOffType.ON.equals(command));
+                webTargets.enableScheduledEvent(id, OnOffType.ON == command);
             }
```
andrewfg, line 144: LGTM

andrewfg (approved):
LGTM

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/handler/HDPowerViewHubHandler.java
jlaur: @lolodomo - with @andrewfg found a better solution (see other comments). Ready again.

lolodomo (approved):
LGTM

andrewfg:
@lolodomo thanks for reviewing and merging the HDPowerView PR's of @jlaur (which I had approved); but could you please also review my own PR #11698 so that we can keep things in sync?

lolodomo:
> @lolodomo thanks for reviewing and merging the HDPowerView PR's of @jlaur (which I had approved); but could you please also review my own PR #11698 so that we can keep things in sync?

Really big one.  Probably too much to expect a fast review from myself

PR #11785, opened 2021-12-15
[wled] add configuration to sort state options of channels

lordjaxom:
As discussed in the forum, it would be nice to have the state options for the channels "effects" and "palettes" sorted like they are in the WLED mobile app, i.e. the first option at the top and the rest alphabetically. This PR adds two thing options that allows to enable exactly that on both channels separately.

https://community.openhab.org/t/mainui-sorting-choices-provided-by-thing-channel/129883/5?u=lordjaxom

This allows for displaying the channels' state options in a sorted list in the various UIs without having to maintain that list manually by adding state option metadata to items.


lolodomo (approved):
LGTM

PR #11786, opened 2021-12-15
[wled] Fix two events being logged when light is off.

Skinah:
Because colours come from RGB values the controls were jumping to the colour that was set and then jumping to OFF as the binding was parsing the two values separately. This PR fixes it so the event.log only shows the channel getting the off state and not a HSB that is on, followed by an OFF.


lolodomo (approved):
LGTM

Issue #11789, opened 2021-12-15
[avmfritz] Handler tries to set brightness for blinds in version 3.2.0.M5


quidam:
## Expected Behavior
The binding does not attempt to set the brightness of blinds.

## Current Behavior
As HAN-FUN blinds (bit mask 335888) identify themselves not only as blinds (bit 18 set), but also as dimmable lights (bit 16 set), the AVMFritzBaseThingHandler tries to update the channel "brightness", which doesn't exist. This leads to infinite messages like 
`Channel 'brightness' in thing 'avmfritz:HAN_FUN_BLINDS:192_168_nnn_n:nnnnnnnnn' does not exist, recreating thing.`
This not only results in a multitude of log entries, but also prevents openHAB to correctly initialize the blinds resulting in an unknown state of the HAN-FUN things following the blinds.

## Possible Solution
only set brightness when the thing is identified as a dimmable light and NOT as blinds. (I'm presently working on a PR)

## Steps to Reproduce (for Bugs)
1. Install avmfritz binding
2. add a blinds thing and other HAN-FUN things
3. set the log level to DEBUG for the binding
4. analyze the openhab log

## Your Environment
* Version used: openHAB and add-on version 3.2.0.M5
* Operating System and version: Raspbian Buster

lolodomo:
Closed by #11790

PR #11773, opened 2021-12-13
[avmfritz] Fixed update of data for blinds

cweitkamp:
- Fixed update of data for blinds

See https://community.openhab.org/t/avmfritz-support-for-han-fun-dect-ule-devices/38950/31

Has been introduced with https://github.com/openhab/openhab-addons/pull/11348.

Signed-off-by: Christoph Weitkamp <github@christophweitkamp.de>

PR #11790, opened 2021-12-15
[avmfritz] Prevent attempt to set brightness of blinds

quidam:
This is a bug fix for issue #11789. It prevents the AVMFritzBaseThingHandler from attempting to set the (non-existing) brightness channel for HAN-FUN blinds.

Signed-off-by: Ulrich Mertin <mail@ulrich-mertin.de>


lolodomo (approved):
LGTM

cweitkamp:
Thanks for looking into this. With my fix in https://github.com/openhab/openhab-addons/pull/11773 this check is not needed anymore. But I think it will not hurt too. 😉 .

Two people one thought.

PR #11794, opened 2021-12-15
[miio] fix action channel miot for empty input with parameters

marcelrv:


lolodomo (approved):
LGTN

PR #11656, opened 2021-11-28
[jsscripting] openhab-js integration 

digitaldan:
This PR integrates @jpg0 excellent OHJ helper library, which is now part of https://github.com/openhab/openhab-js and provides an out of the box environment for users to write Javascript based rules without needing to access raw Java types. Scripts can now be written like  
```javascript
items.getItem("KitchenLight").sendCommand("ON");
```
without any other imports or logic

This pr also enables node.js style console logging (console.log, console.debug...) as well as common javascript setTimeout support . 

Todo
- [ ] Update documentation 
- [x] Update common logging name for  user scripts and add to openhab-core default log config (script.js right now) 
- [x] Add a simpler path for user scripts (deprecate the jsr223 path)

PR #11798, opened 2021-12-16
[jsscripting] Add default translations

wborn:
This add-on recently also became configurable with #11656.


cweitkamp (approved):
LGTM. Thanks.

PR #11799, opened 2021-12-16
[samsungtv] Rename German translations file

wborn:
While uploading the German translations I noticed the file had a different name.
Renaming it prevents us ending up with 2 German translations files once all the texts are translated in German.


cweitkamp (approved):
LGTM. Thanks.

Issue #9722, opened 2021-01-06
[Homekit] Detected value out of range for light sensor


EjvindHald:
In the evening it is completely dark outside my house, and my light sensor therefore correctly registers a light level of 0 lux.

Expected behaviour
I would then expect that 0 lux could be sent to Homekit without problems, because this is normal with a good sensor.

Current Behavior
In the case 0 in in light level from the sensor, this appears in the log:

> Detected value out of range 0.0. Returning min value instead. Characteristic io.github.hapjava.characteristics.impl.lightsensor.CurrentAmbientLightLevelCharacteristic

and my Homekit app shows 1 lux in light level.

Solution
If possible, one solution could be to add custom min and max values. If that is not possible then simply change it to min value as today but without an entry in the log file. This is because it is correct measurement, when it is completely dark.

yfre:
interesting. for some reason the default min value defined by apple is 0.0001.
the HAP library we use for homekit integration does not allow custom min/max value for this characteristic.  
so, i would not to expect a quick fix for this issue. maybe as workaround you could replace 0 with 0.0001 as value using rules or so

diRk262:
I got the same error with several lightsensors, but for me HomeKit is showing 0 Lux in preview and 0,0001 Lux in detailed view.
So when HomeKit min value is 0,0001 and OH sends 0.0, the error will be shown because 0,0<0,0001 Lux.
At the moment I got no idea how to fix this.

yfre:
probably the easiest option right now is to disable logging.
e.g.in karaf console

`set:log ERROR io.github.hapjava.characteristics.impl.lightsensor.CurrentAmbientLightLevelCharacteristic`

this will stop logging for that class and logic will replace 0,0 with 0, 0001

diRk262:
Ok, I changed my log level to error and now the warning will not be shown anymore. Good idea, thank you. Now my log looks way cleaner :)

By the way, you got a typo. The command is `log:set` and not `set:log`.

diRk262:
@yfre 
This is really strange, the warning log entry is back even though my log level is still set to error.

`date
Jan Fri 8 20:08:58 2021
openhab> log:get io.github.hapjava.characteristics.impl.lightsensor.CurrentAmbientLightLevelCharacteristic
ERROR
openhab> log:tail
......
[...]

PR #11803, opened 2021-12-17
[homekit] return min value of characteristic instead of 0

yfre:
a characteristic can have min value, defined either by HAP standard or via openHAB configuration, different to 0.

the current implementation was returning 0 in case openHAB item state is null, independent of defined min value. 

this PR make sure that min value is returned in such cases.  
fix for https://github.com/openhab/openhab-addons/issues/9722


lolodomo (approved):
LGTM

PR #11806, opened 2021-12-18
[wled] Fix Global OFF not turning lights/segments off.

Skinah:
Fixes a bug where the global on/off control is not reflected in each light/segments true state.


lolodomo (approved):
LGTM

lolodomo:
I suppose you agree it does not fix a critical issue ?
I plan to merge it after final 3.2 is released.

Skinah:
Thank you for the multiple quick approvals, apprieated. Critical no, but it will be the source of confusion for people and cause them to post for help. I do have the jar with all fixes in the marketplace already, but I would prefer to delete it from marketplace if this gets merged into 3.2 stable.

PR #11813, opened 2021-12-18
[onkyo] Add channels for media info

marcelrv:
Add channels for audio & video in & out information similar as in the onkyo app

### DIFF
--- a/bundles/org.openhab.binding.onkyo/src/main/java/org/openhab/binding/onkyo/internal/OnkyoParserHelper.java
+++ b/bundles/org.openhab.binding.onkyo/src/main/java/org/openhab/binding/onkyo/internal/OnkyoParserHelper.java
@@ -12,9 +12,12 @@
  */
 package org.openhab.binding.onkyo.internal;
 
+import java.util.Arrays;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
 import org.eclipse.jdt.annotation.NonNullByDefault;
 import org.openhab.core.library.types.StringType;
-import org.openhab.core.types.State;
 
 /**
  * Helper to parse messages.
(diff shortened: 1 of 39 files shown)


lolodomo:
Please update README.

lolodomo:
Please also run the maven command to update your default i18n properties file

bundles/org.openhab.binding.onkyo/src/main/java/org/openhab/binding/onkyo/internal/OnkyoParserHelper.java
```diff
@@ -0,0 +1,47 @@
...
+ * @author Marcel Verpaalen - Initial contribution
+ */
+@NonNullByDefault
+public final class OnkyoParserHelper {
+
+    public static State infoBuilder(String data, int from, int to) {
+        StringBuilder builder = new StringBuilder();
+        int end = to;
+        String[] element = data.split(",");
+        if (element.length < from) {
```
lolodomo, line 31: Please check but I believe it should be `<=`

bundles/org.openhab.binding.onkyo/src/main/java/org/openhab/binding/onkyo/internal/OnkyoParserHelper.java
```diff
@@ -0,0 +1,47 @@
...
+            return StringType.EMPTY;
+        }
+        if (element.length < end) {
+            end = element.length;
+        }
```
lolodomo, line 36: Do you want to include or exclude "end" index ?

bundles/org.openhab.binding.onkyo/src/main/java/org/openhab/binding/onkyo/internal/OnkyoParserHelper.java
```diff
@@ -0,0 +1,47 @@
...
+        boolean firstElement = true;
+        for (int i = from; i < end; i++) {
+            if (!element[i].isEmpty() && !firstElement) {
+                builder.append(", ");
+            }
+            builder.append(element[i]);
+            firstElement = false;
+        }
```
lolodomo, line 44: You can do this with less lines of code.
Here is an example I used in one of my recent PR (after I got the same comment :) ). You can adapt it to your need.
```
        Object @Nullable [] params = msgParams;
        if (params != null && params.length > 0) {
            result += Stream.of(params).map(param -> String.format("\"%s\"", param == null ? "" : param.toString()))
                    .collect(Collectors.joining(", ", " [ ", " ]"));
        }
```

bundles/org.openhab.binding.onkyo/src/main/resources/OH-INF/thing/channels.xml
```diff
@@ -188,5 +188,29 @@
...
+		<item-type>String</item-type>
+		<label>Audio Output Info</label>
+		<description>Audio Output Stream Information</description>
+		<state readOnly="true"></state>
+	</channel-type>
+	<channel-type id="videoIn" advanced="true">
+		<item-type>String</item-type>
+		<label>Video Input Info</label>
+		<description>Video Input Stream Information</description>
+		<state readOnly="true" pattern="%s"></state>
```
lolodomo, line 207: Is there a reason we specify a pattern for some of the new channels but not all ?

lolodomo (changes requested):

bundles/org.openhab.binding.onkyo/src/main/resources/OH-INF/thing/channels.xml
marcelrv: None at all...  Did not notice it... 
I wrote the code 5years ago... and needed to download the  onkyo app just for this info and remembered I did write this way back. time to dust if off

bundles/org.openhab.binding.onkyo/src/main/java/org/openhab/binding/onkyo/internal/OnkyoParserHelper.java
marcelrv: yes, sure can. 
I'm still not too familiar with labda/streams so it is not my typical first way to go... though it is indeed quite compact.

lolodomo (approved):
LGTM, thanks

PR #11816, opened 2021-12-19
[opensprinkler] Add Veto and some default times to state options.

### LINKED ISSUE #11815: [opensprinkler] Sequencing station commands can lead to unexpected behaviour unless autoupdate='false'
Suggest editing opensprinkler binding to set autoupdate to **off** by default for station control channels.

## Expected Behavior
Rule created to switch on station 2 when station 1 is commanded off.
Should result in station 2 turning on whenever station 1 is turned off.

## Current Behavior
station2 appears (in the log) to turn on and back off again. In the opensprinkler interface, station 2 is not turned on at all, but instead results in a "scheduled" status to run 36 hours in the future.

## Possible Solution
After adding ```autoupdate="false"``` to the items, the rule operates as expected. I suspect the problem arises when the binding almost simultaneously sends the commands to stop station 1 and start station 2. However, when autoupdate is forced off, the rule doesn't trigger until after the opensprinkler device has confirmed successful completion of stopping station 1, so the command to start station 2 no longer conflicts.

1. Create rule as follows:
```
rule "run station2 after running station1"
when
	Item station1 changed from ON to OFF
then
	station2.sendCommand(ON)
end
```
2. using the openhab interface (BasicUI), turn on station 1.
3. still using the openhab interface, turn off station 1.

This should result in the incorrect behaviour.

## Context
https://community.openhab.org/t/opensprinkler-stations-jumping-straight-back-to-off-when-on-command-sent/130178

## Your Environment
* Versions used: openhab 3.1.0, 3.2.0M5
* Environment name and version: opensprinkler hardware version 3.1 - DC; firmware 2.1.9 (9)
* Operating System and version: openhab running on MacOS 10.15.7

Skinah:
PR does two things.

1. Gives default times to the duration channels so a new OH3 user does not need to setup metadata for the state options to get a working list widget.
2. Adds Veto to the autoUpdates on a single channel as it was causing issues with a user here on the forum. closes #11815


bundles/org.openhab.binding.opensprinkler/src/main/resources/OH-INF/thing/thing-types.xml
```diff
@@ -180,15 +181,48 @@
 		<label>Next Duration</label>
 		<description>The duration the station will be opened the next time it is switched on.</description>
 		<category>Time</category>
-		<state readOnly="false" pattern="%.0f min"/>
+		<state readOnly="false" pattern="%.0f min">
+			<options>
+				<option value="15s">15 Seconds</option>
+				<option value="1min">1 Minute</option>
+				<option value="5min">5 Minutes</option>
```
lolodomo, line 188: Do your options cover all possible values or only few of them ?
Is the pattern still necessary ? For other values ?
I have a doubt about the values in your options with different units. What unit is used when building the channel state in the binding.
Is it something you tested ?
Same comments for the other channel (rainDelay).

lolodomo:
And please also run the maven command to update your default i18n properties file.

`mvn i18n:generate-default-translations`

bundles/org.openhab.binding.opensprinkler/src/main/resources/OH-INF/thing/thing-types.xml
Skinah: 1. Covers values that most users should be happy with. Metadata can be added to change these default values if they do not suit and I have tested doing this. The issue IMHO is when no defaults are provided the control/widgets default to ANALIZE and showing a graph instead of offering a list of suitable commands to send. For a new openHAB user these changes are far better and also saves time for experienced users if the defaults are suitable.
2. RE pattern: I have removed it for the `nextDuration` as the pattern was not necessary. For the rainDelay channel the API uses hours so I would prefer this gets displayed by default so if people see the TRACE logs they are less likely to get confused. Users can over ride the pattern with metadata. Originally when the binding was merged the ANALIZE feature did not take UOM into account when stored in persistence and this worked around that, not sure if this is fixed now in newer openHAB cores as I have seen it mentioned a few times.
3. The rainDelay has a state that the API will count down in hours. The code does handle all the different units and this is something I have tested for 10+ months now as I have been using Metadata to add the options in different units.

Skinah:
I tried to run `mvn i18n:generate-default-translations` guessing the build system is not yet ready for building 3.3.0-SNAPSHOT?

I got these lines...

```
[WARNING] The POM for org.openhab.core.tools:i18n-maven-plugin:jar:3.3.0-SNAPSHOT is missing, no dependency information available

.....

[ERROR] No plugin found for prefix 'i18n' in the current project and in the plugin groups [org.apache.maven.plugins, org.codehaus.mojo] available from the repositories [local (C:\Users\nospa\.m2\repository), central (https://repo1.maven.org/maven2), openhab-release (https://openhab.jfrog.io/openhab/libs-release)] -> [Help 1]

.......

```

lolodomo:
Build failed. Please apply spotless: `mvn spotless:apply`

lolodomo:
A soon as you fixed the spotless problem, I will run the i18n tool for you and create a PR to your repo.

Skinah:
Spotless was fine on my system, so I did another rebase and force push and now it shows as sorted.

bundles/org.openhab.binding.opensprinkler/src/main/resources/OH-INF/thing/thing-types.xml
lolodomo: I have never played with Number:Time but if you set 24 hours as value, your final value as a string will be 1d and not 24h ? 

And for the other channel, are you sure you will get 1min rather than 60s for example?

lolodomo:
@Skinah : I pushed a PR to your repo with the result of the run of the i18n translation tool.

bundles/org.openhab.binding.opensprinkler/src/main/resources/OH-INF/thing/thing-types.xml
Skinah: All I know is that I have tested it, and it works as I don't know how the UOM works in depth internally. When the command is sent, the device is queried and the response back from the device updates the channels state in HOURS as per the pattern. The channel that has had the pattern removed can at times display the result in seconds if for example a time of 4 days is selected it can change to be shown in seconds. Most of the time it displays the selected label as the nextDuration, as the device does not have a state for this. It is only a virtual control used when you turn a watering zone on.

Skinah:
thanks, changes have been merged.

bundles/org.openhab.binding.opensprinkler/src/main/resources/OH-INF/thing/thing-types.xml
lolodomo: I am not clear with that. Can another maintainer please comment & review? @fwolter @cweitkamp @wborn @kaikreuzer

fwolter: I've never tested if a StringType sent to a Channel, defined in state options and the Channel linked to an UoM Item converts correctly when the string is suffixed with the unit. There's a QuantityType constructor that accepts strings in the format defined in above state options. I think it can work and I trust in @Skinah if he says he tested it, that it works.

lolodomo: In this case, please merge this PR @fwolter

lolodomo:
@fwolter : please take the lead to review this PR

fwolter (approved):

PR #11817, opened 2021-12-19
[map] Rename transform.properties to map.properties

cweitkamp:
- Rename `transform.properties` to `map.properties`

This is necessary because otherwise Crowdin will not pick-up changes.


lolodomo (approved):
LGTM

wborn (approved):

PR #11820, opened 2021-12-20
Update OH version to 3.3.0 in skeleton scripts

cweitkamp:


kaikreuzer (approved):
Ah, very well, I would have forgotten those - thanks!

PR #11828, opened 2021-12-21
correct i18n translations containing spaces in keys in several addons

holgerfriedrich:
[networkupstools] correct i18n translations containing spaces in keys
    
[bmwconnecteddrive] correct i18n translations containing spaces in keys
    
[omnilink] correct i18n translations containing spaces in keys
    
[roku] correct i18n translations containing spaces in keys
    
[somfytahoma] correct i18n translations containing spaces in keys
    
[wlanthermo] correct i18n translations containing spaces in keys
    
[yamahareceiver] correct i18n translations containing spaces in keys
    
This is a workaround for openhab/openhab-core#2636.


holgerfriedrich:
Without this change, state OL is displayed as "CHRG LB = Online, charging, low battery"
I did not create an issue, only the fix.

Hilbrand:
Is this the correct change? It will fix displaying the wrong message for OL, But I'm not sure if this is correct for the changed keys. Because what I read spaces should be escaped with a backslash, not replaced with an underscore. The cause of this is that the new `i18:generate-default-translations` plugin doesn't handle spaces correctly and it generats invalid keys (@wborn).

holgerfriedrich:
I think so. When I updated the string in openhab-cli `openhab:update ... "OL CHRG"`  it translated correctly.

I did not find any documentation how to escape the spaces, and a look into other addons did not show any special escaping besides the underscore.

holgerfriedrich:
@Hilbrand  you are probably right, this fixes the effect, but not the root cause.
The following addons are affected as well: yamahareceiver, roku, wlanthermo.

lolodomo:
As the value is with a space, I don't understand how adding a _ could properly resolve the problem. The property will not be found if option is defined with a _

holgerfriedrich:
@lolodomo I have tested the change, it actually works.
Updating the string value e.g. to "OL CHRG" leads to the correct items text shown in the UI. "OL" is also processed correctly.

In OH3.1 the items showed correctly, in OH3.2 it is currently broken.

@Hilbrand It seems I have to create an issue about the handling of spaces in the i18:generate-default-translations (which currently causes problems in yamahareceiver, roku, wlanthermo, networkupstools). Maybe then I can get a direction if I should replace the spaces in ~20 keys to make it work or someone does a fix of the i18 handling. Looking at other occurrences "state.option" in all property files, I actually see a lot of underscores.... WDYT?

lolodomo:
@wborn @cweitkamp ; what is the expected syntax in the properties file for an option value containing a space? 
Is a _ expected to match a space?! 

I suggest a manual fix of the concerned properties files until the i18n tool is fixed.

Hilbrand:
It works with this change because when the value for the key with spaces is retrieved. It doesn't find the value of the properties file. because there is no key that matches, only one with underscores, but that doesn't match. Then it falls back to the key in the xml file, which still contains the correct value and that is displayed. There for it looks like this change here works, but it only works because it basically adds none existing keys. i.e. giving them a random other name would give in the same result.

According to the [documentation](https://docs.oracle.com/javase/6/docs/api/java/util/Properties.html#load(java.io.Reader)) it states: `All of these key termination characters may be included in the key by escaping them with a preceding backslash character;`

I would suggest to fix this in the tool and than rerun the plugin on the bindings in error. This could firstly be done with a local fix in the core project even before a fix would actually be merged. I've created a issue on the core repository: https://github.com/openhab/openhab-core/issues/2636

lolodomo:
Ok, so the proposed change in this PR is, as I guessed, inappropriate.

bundles/org.openhab.binding.networkupstools/src/main/resources/OH-INF/i18n/networkupstools.properties
```diff
@@ -82,9 +82,9 @@ channel-type.networkupstools.ups-status.state.option.OL = Online
 channel-type.networkupstools.ups-status.state.option.OB = On battery
 channel-type.networkupstools.ups-status.state.option.LB = Low battery
 channel-type.networkupstools.ups-status.state.option.CHRG = Charging
-channel-type.networkupstools.ups-status.state.option.OL CHRG = Online, charging
-channel-type.networkupstools.ups-status.state.option.OL LB = Online, low battery
-channel-type.networkupstools.ups-status.state.option.OL CHRG LB = Online, charging, low battery
+channel-type.networkupstools.ups-status.state.option.OL_CHRG = Online, charging
+channel-type.networkupstools.ups-status.state.option.OL_LB = Online, low battery
+channel-type.networkupstools.ups-status.state.option.OL_CHRG_LB = Online, charging, low battery
```
Hilbrand, line 87: This should fix the issue I would expect. You can test this by changing the text to something random, build the binding, and see if it shows the random text as option description:

```suggestion
channel-type.networkupstools.ups-status.state.option.OL\ CHRG = Online, charging
channel-type.networkupstools.ups-status.state.option.OL\ LB = Online, low battery
channel-type.networkupstools.ups-status.state.option.OL\ CHRG\ LB = Online, charging, low battery
```

holgerfriedrich: Yes, backslash space actually works. 
My random text shows up.

@hilbrand is it worth to proceed here? Shall I extend the PR to the other affected plugins?

lolodomo: I think it would be a good idea if you can fix all. 
Just rename your PR in this case.

lolodomo:
As I am myself using this binding, I can confirm the bug.

ecdye (approved):
LGTM, for OmniLink

wborn (approved):
Sorry for breaking the strings with the plugin.
I had hoped we had ironed out all the :bug: :lady_beetle:!

lolodomo (approved):
LGTM
Thanks for finding that.

holgerfriedrich:
@lolodomo @Hilbrand thanks for your help getting this PR done

PR #11830, opened 2021-12-21
[jsscripting] JS script engine no longer watches node_modules for scripts

### LINKED ISSUE #11831: [jsscripting] Files from node_modules are loaded by FileWatcher
## Expected Behavior

When using jsscripting, the JavaScript files from ``node_modules`` should not be loaded on startup. These files should be called by JavaScript files in the root.

## Current Behavior

When starting openHAB 3.2 all existing JavaScript files in directory ``node_modules`` are also loaded. This leads to hundreds of errors at startup.

## Possible Solution

Exclude ``node_modules`` folder from FileWatcher service.

1. Install fresh openHAB 3.2 Win Release
2. Install jsscripting Binding via UI
3. Add/goto folder ``conf/automation/js``
4. Install ``npm install openhab``
5. Restart openHAB

## Context
It's just a startup issue, bit it could waste memory etc.

## Your Environment

* Version used: openHAB 3.2 Win Release
* Java: OpenJDK 64-Bit Server VM 18.9 (build 11+28, mixed mode)
* Operating System and version: Windows 11

jpg0:
Fixes #11831 where `node_modules` is watched for scripts. This change prevents this subdir being watched (library changes are still watched elsewhere).

### DIFF
--- a/bundles/org.openhab.automation.jsscripting/src/main/java/org/openhab/automation/jsscripting/internal/fs/watch/JSFileWatcher.java
+++ b/bundles/org.openhab.automation.jsscripting/src/main/java/org/openhab/automation/jsscripting/internal/fs/watch/JSFileWatcher.java
@@ -1,0 +1,57 @@
+/**
+ * Copyright (c) 2010-2022 Contributors to the openHAB project
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package org.openhab.automation.jsscripting.internal.fs.watch;
+
+import org.openhab.core.automation.module.script.ScriptEngineManager;
+import org.openhab.core.service.ReadyService;
+import org.osgi.service.component.annotations.Activate;
+import org.osgi.service.component.annotations.Component;
+import org.osgi.service.component.annotations.Deactivate;
+import org.osgi.service.component.annotations.Reference;
+
+/**
+ * Monitors <openHAB-conf>/automation/js for Javascript files & libraries.
+ *
+ * This class is required to ensure that the *order* of set up is correct. Specifically, the dependency tracker must
+ * be activated after the script file watcher. This is because AbstractWatchService only allows a single service to
+ * watch a single directory, and given that the watchers are nested and the last registration wins, the one we want to
+ * monitor the libraries must be registered last.
+ *
+ * @author Jonathan Gilbert - Initial contribution
+ */
+@Component(immediate = true, service = JSFileWatcher.class)
+public class JSFileWatcher {
+
+    private final JSScriptFileWatcher jsScriptFileWatcher;
+    private final JSDependencyTracker jsDependencyTracker;
+
+    @Activate
+    public JSFileWatcher(final @Reference ScriptEngineManager manager, final @Reference ReadyService readyService) {
+        jsDependencyTracker = new JSDependencyTracker();
+        jsScriptFileWatcher = new JSScriptFileWatcher(manager, readyService, jsDependencyTracker);
+    }
+
+    @Activate
+    public void activate() {
+        jsScriptFileWatcher.activate();
+        jsDependencyTracker.activate();
+        jsDependencyTracker.addChangeTracker(jsScriptFileWatcher);
+    }
+
+    @Deactivate
+    void deactivate() {
+        jsDependencyTracker.removeChangeTracker(jsScriptFileWatcher);
+        jsDependencyTracker.deactivate();
+        jsScriptFileWatcher.deactivate();
+    }
+}
--- a/bundles/org.openhab.automation.jsscripting/src/main/java/org/openhab/automation/jsscripting/internal/fs/watch/JSDependencyTracker.java
+++ b/bundles/org.openhab.automation.jsscripting/src/main/java/org/openhab/automation/jsscripting/internal/fs/watch/JSDependencyTracker.java
@@ -16,9 +16,6 @@ import java.io.File;
 
 import org.openhab.core.OpenHAB;
 import org.openhab.core.automation.module.script.rulesupport.loader.DependencyTracker;
-import org.osgi.service.component.annotations.Activate;
-import org.osgi.service.component.annotations.Component;
-import org.osgi.service.component.annotations.Deactivate;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -27,7 +24,6 @@ import org.slf4j.LoggerFactory;
  *
  * @author Jonathan Gilbert - Initial contribution
  */
-@Component(immediate = true, service = JSDependencyTracker.class)
 public class JSDependencyTracker extends DependencyTracker {
 
     private final Logger logger = LoggerFactory.getLogger(JSDependencyTracker.class);
@@ -39,7 +35,6 @@ public class JSDependencyTracker extends DependencyTracker {
         super(LIB_PATH);
     }
 
-    @Activate
     public void activate() {
         File directory = new File(LIB_PATH);
         if (!directory.exists()) {
@@ -51,10 +46,5 @@ public class JSDependencyTracker extends DependencyTracker {
         }
 
         super.activate();
-    }
-
-    @Deactivate
-    public void deactivate() {
-        super.deactivate();
     }
 }
--- a/bundles/org.openhab.automation.jsscripting/src/main/java/org/openhab/automation/jsscripting/internal/fs/watch/JSScriptFileWatcher.java
+++ b/bundles/org.openhab.automation.jsscripting/src/main/java/org/openhab/automation/jsscripting/internal/fs/watch/JSScriptFileWatcher.java
@@ -13,43 +13,41 @@
 package org.openhab.automation.jsscripting.internal.fs.watch;
 
 import java.io.File;
+import java.nio.file.Path;
+import java.nio.file.WatchEvent;
 import java.util.Optional;
 
+import org.eclipse.jdt.annotation.Nullable;
 import org.openhab.automation.jsscripting.internal.GraalJSScriptEngineFactory;
 import org.openhab.core.automation.module.script.ScriptEngineManager;
 import org.openhab.core.automation.module.script.rulesupport.loader.ScriptFileReference;
 import org.openhab.core.automation.module.script.rulesupport.loader.ScriptFileWatcher;
 import org.openhab.core.service.ReadyService;
-import org.osgi.service.component.annotations.Activate;
-import org.osgi.service.component.annotations.Component;
-import org.osgi.service.component.annotations.Deactivate;
-import org.osgi.service.component.annotations.Reference;
 
 /**
- * Monitors <openHAB-conf>/automation/js for Javascript files
+ * Monitors <openHAB-conf>/automation/js for Javascript files, but not libraries
  *
  * @author Jonathan Gilbert - Initial contribution
  */
-@Component(immediate = true)
 public class JSScriptFileWatcher extends ScriptFileWatcher {
     private static final String FILE_DIRECTORY = "automation" + File.separator + "js";
+    private static final String IGNORE_DIR_NAME = "node_modules";
 
-    @Activate
-    public JSScriptFileWatcher(final @Reference ScriptEngineManager manager, final @Reference ReadyService readyService,
-            final @Reference JSDependencyTracker jsDependencyTracker) {
-        super(manager, jsDependencyTracker, readyService, FILE_DIRECTORY);
+    private final String ignorePath;
+
+    public JSScriptFileWatcher(final ScriptEngineManager manager, final ReadyService readyService,
+            JSDependencyTracker dependencyTracker) {
+        super(manager, dependencyTracker, readyService, FILE_DIRECTORY);
+
+        ignorePath = pathToWatch + File.separator + "node_modules";
     }
 
-    @Activate
     @Override
-    public void activate() {
-        super.activate();
-    }
-
-    @Deactivate
-    @Override
-    public void deactivate() {
-        super.deactivate();
+    protected void processWatchEvent(@Nullable WatchEvent<?> event, WatchEvent.@Nullable Kind<?> kind,
+            @Nullable Path path) {
+        if (!path.startsWith(ignorePath)) {
+            super.processWatchEvent(event, kind, path);
+        }
     }
 
     @Override
@@ -61,5 +59,10 @@ public class JSScriptFileWatcher extends ScriptFileWatcher {
                 return Optional.of(GraalJSScriptEngineFactory.MIME_TYPE);
             }
         });
+    }
+
+    @Override
+    protected boolean watchSubDirectories() {
+        return false;
     }
 }


csowada:
Here the bug report #11831

digitaldan (approved):
LGTM

JamieTemple:
a bit out of my depth here, but this change looks like it will stop watching all folders under `automation/js` - I have a lot of subfolders for organising my scripts - will these no longer be loaded?

jpg0:
Yeah, this does remove watching of all subdirectories under `automation/js`.

I guess the two options are:
- only watch the root, meaning that all scripts need to go in this single directory
- watch whole tree except for `node_modules` and everything under it

I submitted this PR as it was the simplest way forward to fix the (major imo) bug, and I personally don't use scripts in multiple directories, so didn't really think about it. However, I understand that this was the way original scripts works, so maybe the second option would be better. It also looks like its possible to do without changing code in core (override `processWatchEvent`).

Happy to resubmit with the second option if this would be preferred?

JamieTemple:
> Yeah, this does remove watching of all subdirectories under `automation/js`.
> 
> I guess the two options are:
> 
> * only watch the root, meaning that all scripts need to go in this single directory
> * watch whole tree except for `node_modules` and everything under it
> 
> Happy to resubmit with the second option if this would be preferred?

Second option def sounds right to me :)

digitaldan (changes requested):
Waiting on changes to support sub dirs, but exclude "node_modules"

jpg0:
I've updated the PR to the new behaviour.

Whilst doing this I noticed that the library change watching doesn't seem to be working, I'm guessing that it was broken at some point in all this recent activity.

moonraker46:
How can I fix this issue with a workaround? 
I have downloaded the 3.2 sources of the add-on, then changed the mentioned file https://raw.githubusercontent.com/openhab/openhab-addons/a130a224485520a0218e1ee8237c9eba792350c3/bundles/org.openhab.automation.jsscripting/src/main/java/org/openhab/automation/jsscripting/internal/fs/watch/JSScriptFileWatcher.java within the src directory and did a build of the 3.2*.jar with "man clean install" and dropped it into the OH  addons folder to install it.

But after a OH restart there was still the same behavior, every *.js from automation/js/node_modules was imported as a rule?

What have I to do to fix this? This behavior is crashing my complete OH system due to the resulting Java memory leaks. 

Thanks!

jpg0:
@moonraker46 I'm not sure why this is occuring; can you use the OH cli to ensure that you have the correct version (and only one!) running?

jpg0:
I've modified this change to ensure that module reloading is functioning. Unfortunately the issue was due to problems much lower level in `WatchQueueReader` so has required more complexity at this level (without needing to make changes in core).

digitaldan:
So i'm not seeing any change in behavior either.  I added some logging and it doesn't seem like the JSScriptFileWatcher is ever called, at least "processWatchEvent" never gets called.  I'm still debugging on my end to see why the service is not registering

jpg0:
Something that I found out is that you cannot have two things watching the same path; the last one will replace the first. The updated code that I pushed ensured that the library listener gets activated after the script listener. However if there is anything else listening to that path it would cause problems.

Weird that it's not working at all though; I have it running against 3.3.0-SNAPSHOT on a clean docker install.

J-N-K:
@jpg0: https://github.com/openhab/openhab-core/issues/2466

jpg0:
@J-N-K that is precisely the behaviour I observed. My workaround was to explicitly control the registration behaviour, however I suspect the state of individual systems (with additional watchers) may subvert this.

florian-h05:
Another idea of how to fix this:
What about having two folders in `automation/js`?

E.g., having `node_modules` for the libs and sth like `scripts` for the scripts?

jpg0:
@florian-h05 good idea, however if you move scripts to `automation/js/scripts`, then libraries need to also move to `automation/js/scripts/node_modules`. Don't worry about it though, it can certainly be fixed. First I'd like to see why my fix is not working for others, as it may require a simple change to get it working for all.

@digitaldan the problem I had before is exactly as @J-N-K describes - in the WatchQueueReader it could only handle a single service per directory. Try looking into it's execution as soon as it sees a change in `automation/js` and see why it's not propagating the event to the watcher (it's in the loop inside `WatchQueueReader.run()`).

digitaldan:
@jpg0 will do, i have some time today to look.

moonraker46:
> @jpg0 will do, i have some time today to look.

Did you have success with fixing the root cause of this issue?

florian-h05:
@digitaldan @jpg0 

Did you hace any success?

digitaldan:
I'm  working on it right now actually.  I modified the WatchQueueReader service to support multiple Watch Services as suggested, but am still running into issues where we try and load everything under node_modules.  Its a little slow going.

jpg0:
Do you _need_ to make the fix to WatchQueueReader? Have you figured out why my PR is not working for you?

digitaldan:
@jpg0 I have a temporary fix while i continue trying to work on the WatchQueueReader

In JSScriptFileWatcher.java
```java
@Override
    protected boolean createAndLoad(ScriptFileReference ref) {
        if (ref.getScriptFileURL().getPath().startsWith(ignorePath)) {
            return false;
        }
        return super.createAndLoad(new ScriptFileReference(ref.getScriptFileURL()) {
            @Override
            public Optional<String> getScriptType() {
                assert super.getScriptType().get().equalsIgnoreCase("js");
                assert !super.getScriptFileURL().getPath().startsWith(ignorePath);
                return Optional.of(GraalJSScriptEngineFactory.MIME_TYPE);
            }
        });
    }
```

I'm assuming this should not be called if processWatchEvents is set to ignore that path, but even with my changes to WatchQueueReader , its still happening (likely i still need to tweak it) .  This is new territory  for me, so its a little slow going to debug.

digitaldan:
We posted at the same time, as you can see from above, for me processWatchEvent is called, but then so is createAndLoad, event though processWatchEvent ignored it.

jpg0:
Can you see why it would have worked for me but not you? If so I could certainly modify the PR.

digitaldan:
I could never get it to work for me, or rather it might have worked at system startup, but if the binding gets restarted, you install a new version, etc... Then the issue would happen.   I tend to restart stuff a lot when debugging , maybe thats the difference?

digitaldan:
So i think we should add at a minimum the following (or something like it) as a fix for the time being to make the binding useable when using libraries.  I'm sure we will be doing a patch release to 3.2, and this would be a good one to get in. 
```java
 if (ref.getScriptFileURL().getPath().startsWith(ignorePath)) {
            return false;
        }
```

jpg0:
@digitaldan the behaviour that you're seeing certainly seems strange. Do you have a configuration that I may be able to replicate? I ask because I'm loathe to support adding in code to avoid a situation we don't understand. Technically this code block should not be necessary as the existing code explicitly requests to not monitor subdirectories. Stepping though this stuff in a debugger should highlight the problem very quickly? (this is how I diagnosed the initial problems)

digitaldan:
Yeah, i have not run it through a debugger yet, i don't usually have my IDE setup for running core , so I'll probably do that soon.  

To reproduce I download a nightly snapshot
`wget "https://ci.openhab.org/job/openHAB3-Distribution/lastSuccessfulBuild/artifact/distributions/openhab/target/openhab-3.3.0-SNAPSHOT.zip"`

 unzip it and start it up.  I then install the shipped JSScripting binding through the UI which ensures any dependencies are installed, then i use the karaf console to install my compiled version of the binding like:

`bundle:update org.openhab.core file:///Users/daniel/openhab-main/git/openhab-core/bundles/org.openhab.core/target/org.openhab.core-3.3.0-SNAPSHOT.jar`

If you then move a library (say openhab-js) into the node_modules folder, i think things look ok until you  restart the binding (or the system)  then it tries to load everything.

csowada:
Maybe it is helpful, but I've not really checked it. On start of the binding the modified function ```processWatchEvent``` is not called. With the first update of a file it works as expected.

I think that we must also filter the root folder in ```ScriptFileWatcher.java``` 

```js
    private synchronized void onStartLevelChanged(int newLevel) {
        int previousLevel = currentStartLevel;
        currentStartLevel = newLevel;

        if (previousLevel < StartLevelService.STARTLEVEL_MODEL) { // not yet started
            if (newLevel >= StartLevelService.STARTLEVEL_MODEL) { // start
                ScheduledExecutorService localScheduler = executorFactory.get();
                scheduler = localScheduler;

                // ************** LOADS all incl. node.modules ?????
                // ***************************************************************
                localScheduler.submit(() -> importResources(new File(pathToWatch)));


                localScheduler.scheduleWithFixedDelay(() -> checkFiles(currentStartLevel), 0, RECHECK_INTERVAL,
                        TimeUnit.SECONDS);
            }
        } else { // already started
            assert scheduler != null;
            if (newLevel < StartLevelService.STARTLEVEL_MODEL) { // stop
                scheduler.shutdown();
                scheduler = null;
            } else if (newLevel > previousLevel) {
                scheduler.submit(() -> checkFiles(newLevel));
            }
        }
    }
```

jpg0:
Ok, I just had another quick look at this. @csowada you were on the right track, that code is required to load scripts at startup, but there was other code in ScriptFileWatcher which assumed that subdirectories should be always processed.

I've submitted a PR to fix this: https://github.com/openhab/openhab-core/pull/2664

(I also noticed that in an update to this PR I had also managed to lose the code to ignore watching subdirs which I just readded.)

@digitaldan I would appreciate it if you could have another go at verifying this.

digitaldan:
Hi, just getting some time now to look. I'll give this a try now that the weekend is here and let you know, thanks!

digitaldan:
Nice Work!  Yes, along with https://github.com/openhab/openhab-core/pull/2664  i am not seeing these errors any more.

digitaldan (approved):

bundles/org.openhab.automation.jsscripting/src/main/java/org/openhab/automation/jsscripting/internal/fs/watch/JSFileWatcher.java
```diff
@@ -0,0 +1,57 @@
+/**
+ * Copyright (c) 2010-2021 Contributors to the openHAB project
```
wborn, line 2:
```suggestion
 * Copyright (c) 2010-2022 Contributors to the openHAB project
```

janwo:
@wborn Can we retrigger the build? I want to try this out ♥️

wborn:
> Can we retrigger the build?

That build would fail because of the invalid license header. :wink:

bundles/org.openhab.automation.jsscripting/src/main/java/org/openhab/automation/jsscripting/internal/fs/watch/JSFileWatcher.java
jpg0: @wborn done

digitaldan:
👍

moonraker46:
So basically there is no ready jsscripting  jar and depending core jar with included bug fix for OH3.2 to download yet?

moonraker46:
> So basically there is no ready jsscripting jar and depending core jar with included bug fix for OH3.2 to download yet?

Builded the core org.openhab.core.automation.module.script.rulesupport-3.2.0.jar package https://file.io/F49ZdSzJcrq3 and and the scripting add-on  org.openhab.automation.jsscripting-3.2.0.jar https://file.io/BhvZpkGukD8x by myself (for OH 3.2.0).

Everything seems to work like a charm now.

janwo:
> > Can we retrigger the build?
> 
> That build would fail because of the invalid license header. 😉

Can we merge it now?😁

digitaldan:
@wborn this looks good to me 👍

wborn (approved):
Thanks for the fix! :+1:

digitaldan:
@jpg0 So this  is strange, i just downloaded the latest snapshot and am having issue loading any rules.  When starting up, no file based rules are loaded, i see nothing in the logs, and no rules show up in the UI.  If i edit a file based rule, then the script watcher does see it and loads the rule file, but i get the following exception:

```
15:49:31.500 [INFO ] [.rulesupport.loader.ScriptFileWatcher] - Loading script '/Users/daniel/openhab-main/openhab-3.3.0-SNAPSHOT/conf/automation/js/test.js'
15:49:31.501 [WARN ] [ommon.WrappedScheduledExecutorService] - Scheduled runnable ended with an exception:
java.lang.IllegalStateException: The Context is already closed.
	at com.oracle.truffle.polyglot.PolyglotEngineException.illegalState(PolyglotEngineException.java:129) ~[?:?]
	at com.oracle.truffle.polyglot.PolyglotContextImpl.checkClosed(PolyglotContextImpl.java:1025) ~[?:?]
	at com.oracle.truffle.polyglot.PolyglotContextImpl.enterThreadChanged(PolyglotContextImpl.java:608) ~[?:?]
	at com.oracle.truffle.polyglot.PolyglotEngineImpl.enterCached(PolyglotEngineImpl.java:1885) ~[?:?]
	at com.oracle.truffle.polyglot.PolyglotEngineImpl.enterIfNeeded(PolyglotEngineImpl.java:1813) ~[?:?]
	at com.oracle.truffle.polyglot.PolyglotValueDispatch.hostEnter(PolyglotValueDispatch.java:1228) ~[?:?]
	at com.oracle.truffle.polyglot.PolyglotContextImpl.getBindings(PolyglotContextImpl.java:966) ~[?:?]
	at com.oracle.truffle.polyglot.PolyglotContextDispatch.getBindings(PolyglotContextDispatch.java:97) ~[?:?]
	at org.graalvm.polyglot.Context.getBindings(Context.java:540) ~[?:?]
	at com.oracle.truffle.js.scriptengine.GraalJSScriptEngine.invokeFunction(GraalJSScriptEngine.java:548) ~[?:?]
	at org.openhab.automation.jsscripting.internal.scriptengine.DelegatingScriptEngineWithInvocableAndAutocloseable.invokeFunction(DelegatingScriptEngineWithInvocableAndAutocloseable.java:116) ~[?:?]
	at org.openhab.automation.jsscripting.internal.scriptengine.InvocationInterceptingScriptEngineWithInvocableAndAutoCloseable.invokeFunction(InvocationInterceptingScriptEngineWithInvocableAndAutoCloseable.java:119) ~[?:?]
	at org.openhab.automation.jsscripting.internal.scriptengine.DelegatingScriptEngineWithInvocableAndAutocloseable.invokeFunction(DelegatingScriptEngineWithInvocableAndAutocloseable.java:116) ~[?:?]
	at org.openhab.automation.jsscripting.internal.scriptengine.InvocationInterceptingScriptEngineWithInvocableAndAutoCloseable.invokeFunction(InvocationInterceptingScriptEngineWithInvocableAndAutoCloseable.java:119) ~[?:?]
	at org.openhab.core.automation.module.script.internal.ScriptEngineManagerImpl.removeEngine(ScriptEngineManagerImpl.java:207) ~[?:?]
	at org.openhab.core.automation.module.script.internal.ScriptEngineManagerImpl.createScriptEngine(ScriptEngineManagerImpl.java:127) ~[?:?]
	at org.openhab.core.automation.module.script.rulesupport.loader.ScriptFileWatcher.createAndLoad(ScriptFileWatcher.java:229) ~[?:?]
	at org.openhab.automation.jsscripting.internal.fs.watch.JSScriptFileWatcher.createAndLoad(JSScriptFileWatcher.java:55) ~[?:?]
	at org.openhab.core.automation.module.script.rulesupport.loader.ScriptFileWatcher.importFile(ScriptFileWatcher.java:213) ~[?:?]
	at org.openhab.core.automation.module.script.rulesupport.loader.ScriptFileWatcher.lambda$2(ScriptFileWatcher.java:205) ~[?:?]
	at java.util.Optional.ifPresent(Optional.java:183) ~[?:?]
	at org.openhab.core.automation.module.script.rulesupport.loader.ScriptFileWatcher.importFileWhenReady(ScriptFileWatcher.java:203) ~[?:?]
	at org.openhab.core.automation.module.script.rulesupport.loader.ScriptFileWatcher.processWatchEvent(ScriptFileWatcher.java:180) ~[?:?]
	at org.openhab.automation.jsscripting.internal.fs.watch.JSScriptFileWatcher.processWatchEvent(JSScriptFileWatcher.java:49) ~[?:?]
	at org.openhab.core.service.WatchQueueReader.lambda$3(WatchQueueReader.java:319) ~[?:?]
```

Further edits will not be picked up by the script watcher at this point.   I'm not sure what is going on.

attilabalazsy:
Guys, what is the good place putting my own library? Previously it was inside the node_modules/myfolder, now it is not loading even if I move the folder a level up (to the same level as node_modules). The library is simply not loading ...

PR #11839, opened 2021-12-21
[tplinksmarthome] KP405 channel in xml should be dimmer not switch

Hilbrand:


lolodomo (approved):
LGTM

lolodomo:
This is not a breaking change as your enhancement that introduced this thing was merged in version 3.3.

PR #11840, opened 2021-12-21
[tradfri] Fix UTF-8 characters in i18n properties file

cweitkamp:
- Fix UTF-8 characters in i18n properties file


lolodomo (approved):
LGTM

PR #11842, opened 2021-12-22
[hue] Changed discovery to mDNS; added HTTPS handling; refactor HTTPClient to use jetty shared client

cweitkamp:
- Changed discovery to mDNS
- Added HTTPS handling
- Refactor HTTPClient to use Jetty shared client
- Refactor to reduce code and simplify implementation

Depends on https://github.com/openhab/openhab-core/pull/2622 and https://github.com/openhab/openhab-core/pull/2635

Philips announced a brand new HUE API v2 being available. These are the first steps to add support for it.

The new API supports some really interesting new features like streaming SSEs. It uses HTTPS only. Additional, they will ~deprecate~ disable UPnP in Q2 2022 in favor of mDNS.

See their [Migration Guide to new API](https://developers.meethue.com/develop/hue-api-v2/migration-guide-to-the-new-hue-api/).


bundles/org.openhab.binding.hue/src/main/java/org/openhab/binding/hue/internal/discovery/HueBridgeUPnPDiscoveryParticipant.java
```diff
@@ -43,31 +41,31 @@
...
+ * central {@link UpnpDiscoveryService}.
  *
  * @author Kai Kreuzer - Initial contribution
  * @author Thomas Höfer - Added representation
  */
-@NonNullByDefault
 @Component(service = UpnpDiscoveryParticipant.class)
-public class HueBridgeDiscoveryParticipant implements UpnpDiscoveryParticipant {
+@NonNullByDefault
+public class HueBridgeUPnPDiscoveryParticipant implements UpnpDiscoveryParticipant {
```
cweitkamp, line 52: TODO: Can be removed ...

bundles/org.openhab.binding.hue/src/main/java/org/openhab/binding/hue/internal/discovery/HueBridgeMDNSDiscoveryParticipant.java
```diff
@@ -0,0 +1,147 @@
...
+    public @Nullable ThingUID getThingUID(ServiceInfo service) {
+        String id = service.getPropertyString(MDNS_PROPERTY_BRIDGE_ID);
+        if (id != null && !id.isBlank()) {
+            return new ThingUID(THING_TYPE_BRIDGE, id);
+        }
+        return null;
+    }
+
+    // @Override
+    // public long getRemovalGracePeriodSeconds(RemoteDevice device) {
```
cweitkamp, line 131: TODO: https://github.com/openhab/openhab-core/pull/2635

bundles/org.openhab.binding.hue/src/main/java/org/openhab/binding/hue/internal/connection/HueTlsTrustManagerProvider.java
```diff
@@ -0,0 +1,101 @@
...
+import java.net.MalformedURLException;
+import java.net.URL;
+import java.nio.charset.StandardCharsets;
+import java.security.cert.CertificateException;
+
+import javax.net.ssl.X509ExtendedTrustManager;
+
+import org.eclipse.jdt.annotation.NonNullByDefault;
+import org.eclipse.jdt.annotation.Nullable;
+import org.openhab.core.io.net.http.PEMTrustManager;
```
cweitkamp, line 26: TODO: https://github.com/openhab/openhab-core/pull/2622

bundles/org.openhab.binding.hue/src/main/java/org/openhab/binding/hue/internal/handler/HueBridgeHandler.java
```diff
@@ -689,20 +705,22 @@ public void initialize() {
                     "@text/offline.conf-error-no-ip-address");
         } else {
             if (hueBridge == null) {
-                hueBridge = new HueBridge(ip, hueBridgeConfig.getPort(), hueBridgeConfig.getProtocol(), scheduler);
+                if (HueBridgeConfig.HTTPS.equals(hueBridgeConfig.getProtocol())) {
+                    // register trustmanager service
+                    HueTlsTrustManagerProvider tlsTrustManagerProvider = new HueTlsTrustManagerProvider(
+                            ip + ":" + hueBridgeConfig.getPort(), hueBridgeConfig.useSelfSignedCertificate());
+                    serviceRegistration = FrameworkUtil.getBundle(getClass()).getBundleContext()
+                            .registerService(TlsTrustManagerProvider.class.getName(), tlsTrustManagerProvider, null);
```
cweitkamp, line 713: This is a really ugly implementation and we do not want it. But unfortunately I did not yet found a better solution.

Implementing the `HueTlsTrustManagerProvider` to be a `ThingHandlerService` would be my favorite on but when doing so there will be a race condition between activation of the `HueTlsTrustManagerProvider` and initialization of the `HueBridgeHandler`.

Registering the ``HueTlsTrustManagerProvider` in the `HueThingHandlerFactory` does not work because I need access to the configuration of the `Thing` to retrieve IP address, port and protocol.

kaikreuzer:
Wow, that's awesome, thank you - I already hoped that you would look into the new hue APIs! 😄 
Especially the new SSE support is really nice to finally get instant state updates.

cweitkamp:
I am planning to separate the implementation into three steps. This is the first one.

1. Basic refactoring
2. Restore current behavior on v2 API - meaning new endpoints, new model - and keep v1 for users who own older Hue Bridges in the background
3. Add handling of SSE

bundles/org.openhab.binding.hue/src/main/java/org/openhab/binding/hue/internal/discovery/HueBridgeMDNSDiscoveryParticipant.java
andrewfg: TODO: see example Miele https://github.com/openhab/openhab-addons/pull/11834

bundles/org.openhab.binding.hue/src/main/java/org/openhab/binding/hue/internal/connection/HueTlsTrustManagerProvider.java
```diff
@@ -0,0 +1,101 @@
...
+     * @return a {@link PEMTrustManager} instance
+     * @throws CertificateInstantiationException
+     */
+    private PEMTrustManager getInstanceFromResource(String fileName) throws CertificateException {
+        String pemCert = readPEMCertificateStringFromResource(fileName);
+        if (pemCert != null) {
+            return new PEMTrustManager(pemCert);
+        }
+        throw new CertificateInstantiationException(
+                String.format("Certifcate resource '%s' not found or not accessible.", fileName));
```
jlaur, line 87: Typo: Certif**i**cate

bundles/org.openhab.binding.hue/src/main/java/org/openhab/binding/hue/internal/handler/HueBridgeHandler.java
```diff
@@ -689,20 +705,22 @@ public void initialize() {
...
+                }
+
+                hueBridge = new HueBridge(httpClient, ip, hueBridgeConfig.getPort(), hueBridgeConfig.getProtocol(),
+                        scheduler);
                 hueBridge.setTimeout(5000);
```
jlaur, line 718: Introduce constant for this value?

bundles/org.openhab.binding.hue/src/main/java/org/openhab/binding/hue/internal/HueBridge.java
```diff
@@ -67,25 +75,31 @@
...
+    private @Nullable Config cachedConfig;
 
     /**
      * Connect with a bridge as a new user.
      *
+     * @param httpClient
      * @param ip ip address of bridge
      * @param port port of bridge
      * @param protocol protocol to connect to the bridge
+     * @param scheduler
```
jlaur, line 99: Parameter description? Same for `httpClient` and other constructor.

andrewfg:
Continuing discussion with @lolodomo from this https://github.com/openhab/openhab-addons/pull/11834#issuecomment-1000286907

> So would make sense in case to set it to 0

You could still do that by means of an entry in the '.cfg' file (rather than via the UI).

> My concern was just to keep the parameter for the hue binding.

@lolodomo the main reason to remove the parameter was the brilliant idea of @kaikreuzer quoted below. Which would enable new systems to pre-discover potential Things on the network without any bindings having been installed. However if you insist that it must be a binding parameter (instead of an entry in a `.cfg` file) then we could not develop such a great feature.

> I would have loved for a long long time already to extract different UPnP+mDNS discovery services into an "installed by default" bundle, so that the setup wizard could directly identify devices in your home without the need of installing all the bindings first.)

lolodomo:
My concern was to have a parameter, whatever the file in which to define it. 

By the way, now that we move from UPnP to mDNS, is this grace period still useful?

bundles/org.openhab.binding.hue/src/main/java/org/openhab/binding/hue/internal/discovery/HueBridgeMDNSDiscoveryParticipant.java
```diff
@@ -0,0 +1,136 @@
...
+        String removalGracePeriodPropertyValue = (String) properties.get(REMOVAL_GRACE_PERIOD);
+        if (removalGracePeriodPropertyValue != null && !removalGracePeriodPropertyValue.isBlank()) {
+            try {
+                removalGracePeriodSeconds = Long.parseLong(removalGracePeriodPropertyValue);
+            } catch (NumberFormatException e) {
+                logger.warn("Configuration property '{}' has invalid value: {}", REMOVAL_GRACE_PERIOD,
+                        removalGracePeriodPropertyValue);
+            }
+        }
+    }
```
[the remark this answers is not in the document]
cweitkamp: I am wondering if the mDNS discovery for Hue Bridges suffer from this behavior at all. In https://github.com/openhab/openhab-addons/pull/9985 we introduced the grace period because of a wrong (or too low) `maxAge` value for UPnP discovery. What is the matching criteria for mDNS? Maybe we should spent some more time to test my replacement and decide afterwards if we still need this configuration for the Hue binding? Wdyt?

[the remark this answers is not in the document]
lolodomo: Exactly.

lolodomo:
@cweitkamp : do you know if mDMD discovery is available for old v1 bridge?

bundles/org.openhab.binding.hue/src/main/java/org/openhab/binding/hue/internal/discovery/HueBridgeMDNSDiscoveryParticipant.java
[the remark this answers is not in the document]
jlaur: @cweitkamp - good point. It could also be kept, but with a default value of 0.

andrewfg:
> now that we move from UPnP to mDNS, is this grace period still useful for mDNS discovery?

I will install the JAR from this PR and test it on my Hue bridge to see if the mDNS disco suffers from the same problem as the UPNP disco.

BTW this time of year is a perfect time to test it. In my case the problem (on UPNP disco) only occurs when the Hue bridge is running "heavy" automations, and presumably the automation causes delays in sending the NOTIFY messages. This time of year is when I run a Christmas light colour loop in the front window of my house. This is quite a "heavy" automation.

cweitkamp:
@lolodomo I am afraid not.

> mDNS is available Hue bridge v2.x (squared-shaped) and supported with firmware version 1.23 and beyond. 

For older bridges there still is the `HueBridgeNupnpDiscovery` which retrieves results from https://discovery.meethue.com/. Both methods are recommended by Philips.

bundles/org.openhab.binding.hue/src/main/java/org/openhab/binding/hue/internal/discovery/HueBridgeMDNSDiscoveryParticipant.java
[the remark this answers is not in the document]
jlaur: Hmm, now just brainstorming: Would it be possible for MDNSDiscoveryParticipant (in core) to simply read that configuration taking the binding id into consideration? I.e. get value of `discovery.<binding id>:removalGracePeriod` and apply that if it exists? Then bindings would only have to implement anything if they need a default value != 0. In all cases the .cfg value would override binding default (which would be 0 if binding has no default).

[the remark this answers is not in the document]
andrewfg: Hmm.. I am not sure if the MDNSDiscoveryParticipant in core has a way (currently) to know the binding Id. Especially if we want to implement @kaikreuzer's suggestion of pre-discovering Things on new systems before any bindings have been installed. And if it would be possible then we would obviously have to revert both mDNS and UPnP disco..

andrewfg:
> I will install the JAR from this PR and test it on my Hue bridge

Unfortunately I could not build the binding jar because there is some dependency on PEMTrustManager in the core, which I cannot find. => So @cweitkamp could you please give me a link to the built JAR?

cweitkamp:
Here you are: [org.openhab.core.io.net-3.3.0-SNAPSHOT.zip](https://github.com/openhab/openhab-addons/files/7770900/org.openhab.core.io.net-3.3.0-SNAPSHOT.zip)

andrewfg:
> Here you are

@cweitkamp many thanks. I have it loaded on my test system; it discovered the Bridge Ok (presumably via mDNS); and when I instantiated the Bridge as a Thing, it discovered all the lamp Things Ok too. So far so good. I then deleted the Bridge Thing, and deleted everything from the Inbox; loaded up the Bridge with two "heavy" Hue Labs Christmas Automations; and I will let the test system run throughout this evening to see if the Bridge a) disappears from and reappears in the Inbox, or b) remains online stable. I will report back later.

andrewfg:
@cweitkamp I see that your JAR does still implement the config param for removalGracePeriod with default at 50 seconds, but for the avoidance of doubt, can you please advise if your code is recognising that param in the mDNS disco participant?

cweitkamp:
@andrewfg Yes, the current implementation has full support for mDNS grace period configuration via `services.cfg` file. Set `discovery.hue:removalGracePeriod = 0` to disable it.

andrewfg:
^
Unfortunately so far I only got an indeterminate result..
- Whether removalGracePeriod has a value of 50 or 1, the Bridge is never removed from the Inbox.
- **_HOWEVER_** even if I completely disconnect the Ethernet cable, the Bridge is still not removed from the Inbox!!

So unfortunately I cannot say (yet) if the Inbox coming/going problem is actually resolved in Hue API v2. I suspect that my router or other device on my LAN is doing (m)DNS caching / re-broadcasting, and I suppose that is why the problem "appears" to be resolved.

Next Steps: Rather than trying to debug my working LAN, or flush the respective mDNS cache, I will assemble a rudimentary LAN and try to test again on that instead. To do that I think I need to by myself a spare basic router/switch, and it being Christmas, this might take me a day or two.

andrewfg:
In the meantime, I think I discovered some bugs.

To be specific, when the core does an mDNS broadcast, the Hue Bridge returns an infinite Time To Live value (-1). Which I think means that once the device has entered into the Inbox, it will NEVER be removed, even if the device is physically disconnected.

So IMHO there are probably two bugs as follows..
1. There is a bug in the Hue Bridge firmware: It should not return an infinite Time To Live value. Probably this should be reported to Philips.
2. There is a bug in the OH MDNSDiscoveryService: It should never accept an infinite Time To Live value from a device. Instead it should impose some non infinite time after which it will delete/refresh its cache. So that physically lost devices are finally actually deleted from the Inbox.

Obviously this issue potentially goes beyond the scope of just the Hue binding. => But its probably best to collect comments here first.

andrewfg:
^
More..

I just looked at the network traffic using WireShark. In fact the Hue Bridge **IS** returning valid TTL values (of 120 seconds) in its mDNS response telegrams. => So it seems that the core MDNSDiscoveryService is wrongly parsing the TTL value to INFINITE (-1).

=> I obviously need to dig deeper into this!!

cweitkamp:
The magic behind mDNS discovery on the local network is done by a third-party library located here: https://github.com/jmdns/jmdns. Without knowing much about the internals of that library I tend to say your observed issue has to be looked up over there. OHC implements the JmDNS `ServiceListener` and reacts upon the events accordingly.

andrewfg:
@cweitkamp thanks for the pointer. In fact it seems more complex than that: when the OH mDNS discovery service calls for devices to report in, the Hue actually sends **eight** different types mDNS answer records (types PTR, TXT, SRV, A, AAAA x2, NSEC x 2). And the `HueBridgeMDNSDiscoveryParticipant.getThingUID()` method recognises and hits on six of these. So in other words, after one mDNS ping, the Hue binding currently makes six `thingDiscovered()` calls-backs. They all have the same ThingUID so the core only adds one thing to the Inbox. But I suppose that the JmDNS library caches each answer record separately according to its own TTL. And the specific difficulty is that some of the six/seven answer records have a TTL of 120 seconds (which seems reasonable), and some of them have a TTL of 4500 seconds (75 minutes, which seems too long).

So I will try to refine your mDNS code in this PR to 1) to be more selective so that it only hits on just one of the eight types of mDNS answer records, and (therefore) 2) will use the proper TTL value on the cache.

andrewfg:
> be more selective so that it only hits on just one of the eight types of mDNS answer records

I have been digging through the code, and I think this is impossible; so suggest not changing anything.

> use the proper TTL value on the cache.

Nevertheless, I think the JMDNS cache is taking the 4500 second TTL value on the "main" discovery resource record, rather than the 120 second TTL value on the discovery resource record that contains the IP address. This means that a `removalGracePeriod` would certainly never be needed. But it also means that if the Hue is unplugged or powered off, it will still remain in the Inbox for 75 minutes. 

So my proposed next steps are as follows..
1. I suggest to keep your existing removalGracePeriod code "just in case"; but you could make it so that if the config param is absent the default value is 0.
2. I suggest add a '.withTTL()` clause to the result builder instantiation (see below); and I suggest the value 120 seconds since this is exactly the value that the Hue bridge provides in the TTL value it provides on its discovery resource record containing the IPv4 address. (i.e. hard code the value that is actually observed, even though it apparently cannot be read programmatically via calls to the JmDNS bundle). This means that if the bridge is unplugged or powered off, it will disappear from the Inbox in a reasonable time.

image

cweitkamp:
@andrewfg Thank you very much for the analysis. I incorporated your suggestions from https://github.com/openhab/openhab-addons/pull/11842#issuecomment-1001633159.

jaywiseman1971:
Will this fix address the bridge going offline/online within milliseconds of each other?

https://community.openhab.org/t/oh-3-x-bug-things-going-offline-online-within-milliseconds/131356

Best, Jay

bundles/org.openhab.binding.hue/src/main/java/org/openhab/binding/hue/internal/connection/HueBridge.java
```diff
@@ -67,25 +98,31 @@
...
+    private @Nullable Config cachedConfig;
 
     /**
      * Connect with a bridge as a new user.
      *
+     * @param httpClient
      * @param ip ip address of bridge
      * @param port port of bridge
      * @param protocol protocol to connect to the bridge
+     * @param scheduler
```
lolodomo, line 122: You could add a short description to the 2 new parameters

bundles/org.openhab.binding.hue/src/main/java/org/openhab/binding/hue/internal/connection/HueBridge.java
```diff
@@ -106,35 +143,26 @@ public HueBridge(String ip, int port, String protocol, ScheduledExecutorService
      * Use the ip only constructor and authenticate() function if
      * you don't want to connect right now.
      *
+     * @param httpClient
      * @param ip ip address of bridge
      * @param port port of bridge
      * @param protocol protocol to connect to the bridge
      * @param username username to authenticate with
+     * @param scheduler
```
lolodomo, line 151: Same here

bundles/org.openhab.binding.hue/src/main/java/org/openhab/binding/hue/internal/connection/HueBridge.java
```diff
@@ -438,17 +453,17 @@ public Group getAllGroup() {
...
-        Result result = http.get(getRelativeURL("groups"));
+        HueResult result = get(getRelativeURL("groups"));
 
         handleErrors(result);
 
-        Map<String, FullGroup> groupMap = safeFromJson(result.getBody(), FullGroup.GSON_TYPE);
-        ArrayList<FullGroup> groupList = new ArrayList<>();
+        Map<String, FullGroup> groupMap = safeFromJson(result.body, FullGroup.GSON_TYPE);
 
+        ArrayList<FullGroup> groups = new ArrayList<>();
```
lolodomo, line 462: Just `List<FullGroup> groups =` ?

bundles/org.openhab.binding.hue/src/main/java/org/openhab/binding/hue/internal/connection/HueBridge.java
```diff
@@ -67,25 +98,31 @@
 
     private static final String DATE_FORMAT = "yyyy-MM-dd'T'HH:mm:ss";
 
+    private final HttpClient httpClient;
     private final String ip;
     private final String baseUrl;
     private @Nullable String username;
+    private int timeout = 1000;
```
lolodomo, line 105: Looks like a strange default value.
Is it a value in seconds ?

bundles/org.openhab.binding.hue/src/main/java/org/openhab/binding/hue/internal/connection/HueBridge.java
```diff
@@ -1129,4 +984,117 @@ private String getRelativeURL(String path) {
...
+                request.content(new StringContentProvider(body), "application/json");
+            }
+
+            final ContentResponse contentResponse = request.send();
+
+            final int httpStatus = contentResponse.getStatus();
+            final String content = contentResponse.getContentAsString();
+            logger.trace("Hue response: status = {}, content = '{}'", httpStatus, content);
+            return new HueResult(content, httpStatus);
+        } catch (ExecutionException | InterruptedException | TimeoutException e) {
```
lolodomo, line 1026: Please execute `Thread.currentThread().interrupt();` in case of InterruptedException before logging and throwing `CommunicationException`

lolodomo (changes requested):
Review part 1 of 3

bundles/org.openhab.binding.hue/src/main/java/org/openhab/binding/hue/internal/discovery/HueBridgeMDNSDiscoveryParticipant.java
```diff
@@ -0,0 +1,138 @@
...
+
+/**
+ * The {@link HueBridgeMDNSDiscoveryParticipant} is responsible for discovering new and removed Hue Bridges. It uses the
+ * central {@link MDNSDiscoveryService}.
+ *
+ * @author Kai Kreuzer - Initial contribution
+ * @author Thomas Höfer - Added representation
+ * @author Christoph Weitkamp - Change discovery protocol to mDNS
+ */
+@Component(configurationPid = "discovery.hue")
```
lolodomo, line 49: Looking at other bidnings, sometimes `service = MDNSDiscoveryParticipant.class` is added.
Is it necessary or not ?

bundles/org.openhab.binding.hue/src/main/java/org/openhab/binding/hue/internal/discovery/HueBridgeMDNSDiscoveryParticipant.java
```diff
@@ -0,0 +1,138 @@
...
+            }
+        }
+        return null;
+    }
+
+    @Override
+    public @Nullable ThingUID getThingUID(ServiceInfo service) {
+        String id = service.getPropertyString(MDNS_PROPERTY_BRIDGE_ID);
+        if (id != null && !id.isBlank()) {
+            return new ThingUID(THING_TYPE_BRIDGE, id.toLowerCase());
```
lolodomo, line 129: Maybe you should replace possible unexpected characters in id before using it as thing ID ?

bundles/org.openhab.binding.hue/src/main/java/org/openhab/binding/hue/internal/discovery/HueBridgeNupnpDiscovery.java
```diff
@@ -87,37 +75,25 @@ private void discoverHueBridges() {
         for (BridgeJsonParameters bridge : getBridgeList()) {
             if (isReachableAndValidHueBridge(bridge)) {
                 String host = bridge.getInternalIpAddress();
-                String serialNumber = bridge.getId().substring(0, 6) + bridge.getId().substring(10);
-                serialNumber = serialNumber.toLowerCase();
+                String serialNumber = bridge.getId().toLowerCase();
```
lolodomo, line 78: You are changing the way to get the serial number. Do you mean this was not working well ?

bundles/org.openhab.binding.hue/src/main/java/org/openhab/binding/hue/internal/dto/ScheduleCommand.java
```diff
@@ -55,7 +58,7 @@ public String getMethod() {
      *
      * @return request body
      */
-    public String getBody() {
-        return body.toString();
+    public @Nullable String getBody() {
+        return body != null ? body.toString() : "";
```
lolodomo, line 62: Apparently, your return cannot be null

lolodomo (changes requested):
Review part 2 of 3

bundles/org.openhab.binding.hue/README.md
```diff
@@ -1,23 +1,23 @@
 # Philips Hue Binding
 
 This binding integrates the [Philips Hue Lighting system](https://www.meethue.com).
-The integration happens through the Hue bridge, which acts as an IP gateway to the ZigBee devices.
+The integration happens through the Hue Bridge, which acts as an IP gateway to the ZigBee devices.
```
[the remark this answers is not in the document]
cweitkamp: "Hue Bridge" is the official product name.

bundles/org.openhab.binding.hue/src/main/java/org/openhab/binding/hue/internal/discovery/HueBridgeMDNSDiscoveryParticipant.java
cweitkamp: No, it is not necessary.

bundles/org.openhab.binding.hue/src/main/java/org/openhab/binding/hue/internal/connection/HueBridge.java
cweitkamp: Yes, we are talking about seconds. I will decrease it to 10.

bundles/org.openhab.binding.hue/src/main/java/org/openhab/binding/hue/internal/dto/ScheduleCommand.java
cweitkamp: Luckily this class is not needed at all. I removed all references.

bundles/org.openhab.binding.hue/src/main/java/org/openhab/binding/hue/internal/discovery/HueBridgeMDNSDiscoveryParticipant.java
cweitkamp: AFAIK the `bridgeid` is a hexadecimal which cannot contain unexpected characters.

bundles/org.openhab.binding.hue/src/main/java/org/openhab/binding/hue/internal/discovery/HueBridgeNupnpDiscovery.java
cweitkamp: The removed UPnP discovery used a shortened `bridgeid` without the bridge indicator "fffe" in it. This has changed in mDNS discovery and thus we do not need this logic anymore to align the two bridge discovery services.

cweitkamp:
@lolodomo Thanks for your review. I incorporated your suggested changes and pushed a new version.

bundles/org.openhab.binding.hue/src/main/java/org/openhab/binding/hue/internal/connection/HueBridge.java
lolodomo: But it is then used as seconds in the code if I correctly remember when I reviewed.

cweitkamp: Yes, i changed that as well. See https://github.com/openhab/openhab-addons/pull/11842/commits/0ad986ad41599d61bbf707b11b12e15e3d828ba6.

lolodomo:
@cweitkamp / I will try to finish my last part review very soon. Note that there is now a conflict to solve.
Do you still expect more tests ? I could test the new discovery stuff with my old bridge but that's all.
Would you like this PR merged and be included in 3.3M1 or do you prefer we wait ?

cweitkamp:
I'd prefer to see it in milestone 3.3.M1. But in advance we have to merge https://github.com/openhab/openhab-core/pull/2695 too. Maybe @openhab/core-maintainers can take care of that soon.

andrewfg:
^
I was planning on giving this new version a good testing on my operative system. But I can’t do that until the core has been updated to remove the dependency. So in the meantime (speaking for myself) the new binding version is operationally untested. Except for the discovery issue, that seems to have been resolved.

bundles/org.openhab.binding.hue/src/main/java/org/openhab/binding/hue/internal/handler/HueBridgeHandler.java
```diff
@@ -1043,8 +1069,8 @@ private void notifyGroupSceneUpdate(List<Scene> scenes) {
 
         String ip = hueBridgeConfig.getIpAddress();
```
lolodomo, line 1070: As detected GitHub actions, there should be a problem here as you removed the method `getIpAddress` from the config class.

bundles/org.openhab.binding.hue/src/main/java/org/openhab/binding/hue/internal/handler/HueBridgeHandler.java
```diff
@@ -676,38 +686,45 @@ public void dispose() {
...
-                // Try a first connection that will fail, then try to authenticate,
-                // and finally change the bridge status to ONLINE
-                initJob = scheduler.submit(new PollingRunnable() {
-                    @Override
-                    protected void doConnectedRun() throws IOException, ApiException {
-                    }
-                });
+                onUpdate();
             }
-            onUpdate();
```
lolodomo, line 705: I am not sure in which case initialize could be called while hueBridge is not null, but with your changes the different polling will no more be started in this case.

lolodomo (changes requested):
Review part 3 of 3

lolodomo:
@cweitkamp : review finished

bundles/org.openhab.binding.hue/src/main/java/org/openhab/binding/hue/internal/connection/HueBridge.java
```diff
@@ -1129,4 +971,133 @@ private String getRelativeURL(String path) {
...
+                synchronized (commandsQueue) {
+                    AsyncPutParameters payloadCallbackPair = commandsQueue.poll();
+                    if (payloadCallbackPair != null) {
+                        logger.debug("Async sending put to address: {} delay: {} body: {}", payloadCallbackPair.address,
+                                payloadCallbackPair.delay, payloadCallbackPair.body);
+                        try {
+                            HueResult result = doNetwork(payloadCallbackPair.address, HttpMethod.PUT,
+                                    payloadCallbackPair.body);
+                            payloadCallbackPair.future.complete(result);
+                        } catch (CommunicationException e) {
```
lolodomo, line 1065: Are you not catching also ConfigurationException ?

bundles/org.openhab.binding.hue/src/main/java/org/openhab/binding/hue/internal/connection/HueBridge.java
```diff
@@ -106,35 +147,26 @@ public HueBridge(String ip, int port, String protocol, ScheduledExecutorService
...
      * @param port port of bridge
      * @param protocol protocol to connect to the bridge
      * @param username username to authenticate with
+     * @param scheduler the ExecutorService to schedule commands
      */
-    public HueBridge(String ip, int port, String protocol, String username, ScheduledExecutorService scheduler)
-            throws IOException, ApiException {
-        this(ip, port, protocol, scheduler);
+    public HueBridge(HttpClient httpClient, String ip, int port, String protocol, String username,
+            ScheduledExecutorService scheduler) throws IOException, ApiException {
```
lolodomo, line 158: Other exceptions can be thrown by the call of method authenticate : ConfigurationException, UnauthorizedException

lolodomo (changes requested):

lolodomo (changes requested):

jlaur:
@cweitkamp - considering the migration to mDNS in Q2/2022, would it be possible to have a look at the few remaining comments from @lolodomo, so we could have this merged soon? :)

bundles/org.openhab.binding.hue/src/main/java/org/openhab/binding/hue/internal/discovery/HueDeviceDiscoveryService.java
```diff
@@ -70,13 +69,13 @@
 public class HueDeviceDiscoveryService extends AbstractDiscoveryService
         implements DiscoveryService, ThingHandlerService {
```
jlaur, line 70: Old, but: `AbstractDiscoveryService` implements `DiscoveryService`, so this seems redundant here.

lolodomo:
@cweitkamp : could you please at least comment ? If you have no time, someone can probably finish, starting from your current code.

cweitkamp:
Sry for the delay. I incorporated your comments. In case I did not hit all of them please feel free to add more commits.

bundles/org.openhab.binding.hue/src/main/java/org/openhab/binding/hue/internal/discovery/HueBridgeMDNSDiscoveryParticipant.java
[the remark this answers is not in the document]
jlaur: @andrewfg - can this comment be closed? I'm not sure if anything is left to do in this PR, putting aside some general ideas and future wishes. :-)

jlaur:
@lolodomo - gentle ping, can you check your comments?

jlaur:
> @lolodomo - gentle ping, can you check your comments?

Maybe before tomorrow? :-)

jlaur (approved):
Thank you for all these improvements! @lolodomo, @andrewfg - please check your comments.

andrewfg (approved):
LGTM

bundles/org.openhab.binding.hue/src/main/java/org/openhab/binding/hue/internal/connection/HueBridge.java
jlaur: I've checked that it's caught in current code.

bundles/org.openhab.binding.hue/src/main/java/org/openhab/binding/hue/internal/connection/HueBridge.java
jlaur: I've checked that they were added.

lolodomo:
I should be able to have a look today

bundles/org.openhab.binding.hue/src/main/java/org/openhab/binding/hue/internal/connection/HueBridge.java
```diff
@@ -438,17 +464,16 @@ public Group getAllGroup() {
     public List<FullGroup> getGroups() throws IOException, ApiException {
```
jlaur, line 464: ConfigurationException, CommunicationException is missing here (for consistency).

bundles/org.openhab.binding.hue/src/main/java/org/openhab/binding/hue/internal/connection/HueBridge.java
```diff
@@ -1129,4 +971,133 @@ private String getRelativeURL(String path) {
...
+
+    public HueResult post(String address, String body) throws ConfigurationException, CommunicationException {
+        return doNetwork(address, HttpMethod.POST, body);
+    }
+
+    public HueResult put(String address, String body) throws ConfigurationException, CommunicationException {
+        return doNetwork(address, HttpMethod.PUT, body);
+    }
+
+    public HueResult delete(String address) throws ConfigurationException, CommunicationException {
```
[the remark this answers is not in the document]
jlaur: I have checked, only one method lacking. I've added this as a separate comment. Not strictly needed as they are unchecked runtime exceptions, but for consistency it should be added. I wouldn't let it block the PR though.

bundles/org.openhab.binding.hue/src/main/java/org/openhab/binding/hue/internal/connection/HueBridge.java
jlaur: Added.

bundles/org.openhab.binding.hue/src/main/java/org/openhab/binding/hue/internal/discovery/HueDeviceDiscoveryService.java
jlaur: Removed.

bundles/org.openhab.binding.hue/src/main/java/org/openhab/binding/hue/internal/handler/HueBridgeHandler.java
jlaur: This has been addressed.

bundles/org.openhab.binding.hue/src/main/java/org/openhab/binding/hue/internal/connection/HueBridge.java
lolodomo: But method updateGroups in is calling this method and so should also be marked as throwing `ConfigurationException and, `CommunicationException `.
And then lightPollingRunnable...

If you search a little, you will find that these exceptions should be mentioned at more places.

lolodomo:
Additionally, I see the tag requesting more tests. Was there at least someone testing the binding with these changes ?

lolodomo:
If no additional tests are required, we could merge the PR and then later I will try to check and cleanup the exception handling in a next PR.

jlaur:
> If no additional tests are required, we could merge the PR and then later I will try to check and cleanup the exception handling in a next PR.

Fine for me, I will also install it on my productive system and monitor it. The exception handling is mostly cosmetic, I believe.

bundles/org.openhab.binding.hue/src/main/java/org/openhab/binding/hue/internal/connection/HueBridge.java
jlaur: You are right. This seems to be an actual bug since these exceptions are not handled anywhere. So in case of time-out, invalid certificate, etc. the binding might stop working.

jlaur: @cweitkamp - since we didn't make it for 3.3, this comment should probably be looked into.

bundles/org.openhab.binding.hue/src/main/java/org/openhab/binding/hue/internal/handler/HueBridgeHandler.java
mlobstein: Is there a way for HueTlsTrustManagerProvider to return a status in the case that getting the external cert fails (i.e. the server is momentarily unreachable)? It seems like the current code has no provision to retry the cert retrieval process since it attempted only once during initialize().

lolodomo:
@jlaur : it would be good to finish this PR. What was the result of your tests, any problem ?

jlaur:
> it would be good to finish this PR. What was the result of your tests, any problem ?

I think there is one important open comment regarding possible lack of exception handling + conflicts that need to be resolved. I don't remember if I tested it, but I currently don't have this version installed. I will install it tonight to get started.

andrewfg:
> If you would be able to find time 

Apropos time..  ideally we should also implement the SSE (albeit in a separate PR) before the API v2 binding passes into a full OH release version.

As a matter of fact I just wrote the SSE implementation for the HDPowerView Generation 3 API (#13355), and the Hunter Douglas API Gen 3 SSE specification is almost identical to Philips API V2 SSE specification. So I would be happy to contribute that part for Hue -- unless someone else is already working on it. => @jlaur / @lolodomo / @cweitkamp can you confirm if you are happy for me to contribute the SSE implementation for Hue?

lolodomo:
> can you confirm if you are happy for me to contribute the SSE implementation for Hue?

Any contribution is welcome. 
Just two requests from myself:
1. Use the OH standards for the SSE implementation (JAX-RS +Jettyl.
2. Keep the compatibility withe old bridge in case it does not support SSE.

andrewfg:
> Any contribution is welcome.

Oops! Just digging into @cweitkamp 's code in this PR .. please correct me if I am wrong .. but it seems that this PR does NOT implement the API V2 at all!!

Apparently this PR just PREPARES for an API V2 implementation (only). So actually before we/I even start to implement SSE, we/I would first need to implement the regular API V2 endpoints via regular HTTP(S) GET to the new API V2 url endpoints.

So in fact there is much more work to be done than I originally expected. There are two further PRs required..
- Implement API V2 GET/PUT endpoint urls, and respective json parsing
- Implement API V2 SSE

---

> Just two requests from myself:

For sure. :)

cweitkamp:
Hi all, I will try to finalize this PR in the next two weeks.

Please correct me if I am wrong but in my opinion the newly introduced exceptions are handled in the `run()` method of the private `PollingRunnable` class inside the `HueBridgeHandler`. The only place where the currently existing `ApiException` is handled too. Adding `throws` declaration for uncatchable runtime exceptions is a question of code-style. If you like to see it everywhere I am going to add it - even if we are editing a nearly deprecated class. Before starting on this contribution I thought about a complete rewrite of the binding instead of a refactoring it but decided against it because of the long time stability. From my point of view the code partly is in a very bad - or let's call it old fashioned - state. Sry to be so rude here.

@andrewfg You are right. This PR was meant as a preparation for further improvements and implementations of V2 API and SSE events. Feel free to start working on it. It will be much appreciated. I am not seeing a chance for me to jump in for such a big improvement in the near future. But I will volunteer for testing.

bundles/org.openhab.binding.hue/src/main/java/org/openhab/binding/hue/internal/connection/HueBridge.java
jlaur: As clarified by [this comment](https://github.com/openhab/openhab-addons/pull/11842#issuecomment-1259970498) the exceptions are handled. So only the consistency matter is left. We could either remove all the explicit declarations of these unchecked exceptions, or add them everywhere for consistency. I would opt for the first option. However, it's only a matter of style and it doesn't make any difference runtime, so I would approve this PR even without changing anything.

jlaur:
> Hi all, I will try to finalize this PR in the next two weeks.

Thanks, much appreciated! I will install it on my productive system now.

> Please correct me if I am wrong but in my opinion the newly introduced exceptions are handled in the `run()` method of the private `PollingRunnable` class inside the `HueBridgeHandler`. The only place where the currently existing `ApiException` is handled too.

Thanks for clarifying this, I missed that previously.

> Adding `throws` declaration for uncatchable runtime exceptions is a question of code-style. If you like to see it everywhere I am going to add it - even if we are editing a nearly deprecated class.

IMHO don't bother.

> Before starting on this contribution I thought about a complete rewrite of the binding instead of a refactoring it but decided against it because of the long time stability. From my point of view the code partly is in a very bad - or let's call it old fashioned - state. Sry to be so rude here.

It's better to be transparent about this, thanks. I think this could be valuable input for @andrewfg also before starting some major changes.

andrewfg:
> I thought about a complete rewrite of the binding 
> the code partly is in a very bad - or let's call it old fashioned - state

Yup. I agree with the above points! :)

Notwithstanding the above, I am tending to follow the same approach as I took with the HDPowerView API generation change as follows..
- Refactor the OH facing parts of the binding to use abstract base DTO classes (such as `Scene`, `Light`, etc.) and access the Bridge via an abstract base communication class `HueBridge`
- Relocate the current API V1 implementation code from `Scene`, `Light`, etc., and `HueBridge` into new `SceneV1`, `LightV1`, etc., and `HueBridgeV1` implementation classes which implement/extend the abovementioned base abstract classes
- Write new `SceneV2`, `LightV2`, etc., and `HueBridgeV2` implementation classes that implement/extend the base abstract classes too.
- Use a Config Parameter in the the thing handler to select whether the OH facing code instantiates/consumes the `xxxV1` or `xxxV2` implementation classes.

This allows the the OH facing parts of the binding to remain almost unchanged. And the V1 code would remain functionally unchanged too; although the actual code has to be split between the `xxx` base/abstract class, and the `xxxV1` implementation class.

In the case of the HDPowerView migration, we have the good luck that the manufacturer chose to keep the V1 and V2 DTOs fairly similar, with many fields being common, and only a subset of fields having changed names and meanings. However in the case of the Hue migration, I can already see that the DTOs have changed MASSIVELY between V1 and V2, so I can already say that unlike the HDPowerView migration which I was able to write in a week or so, this Hue migration will certainly take a lot longer.

In any case, before I start writing code, I will first write a Class Structure diagram similar to the one I wrote for [HDPowerView](https://github.com/openhab/openhab-addons/issues/12678#issuecomment-1239297302) -- this should help to clarify if my above mentioned approach would work (or not)..

andrewfg:
@cweitkamp @jlaur @lolodomo rather than discussing the next steps for the Hue binding in this PR, I have opened a new issue #13456 where we can discuss the potential next PRs for implementing the API V2 functionality.

cweitkamp:
Rebase done. Conflicts are resolved.

andrewfg:
There is a status update on the work I have already done, and my planned next steps, [**here**](https://github.com/openhab/openhab-addons/issues/13456#issuecomment-1268570899)

jlaur:
> I have been running this version for a week now without any major issues. I have experienced this a few times though, which I'm not used to

With debug log level I have now observed a few times that it happens because of timeouts:

```
2022-10-07 05:22:35.414 [DEBUG] [ng.hue.internal.connection.HueBridge] - TimeoutException occurred during execution: Total timeout 5000 ms elapsed
java.util.concurrent.TimeoutException: Total timeout 5000 ms elapsed
	at org.eclipse.jetty.client.HttpConnection$RequestTimeouts.onExpired(HttpConnection.java:334) ~[?:?]
	at org.eclipse.jetty.client.HttpConnection$RequestTimeouts.onExpired(HttpConnection.java:314) ~[?:?]
	at org.eclipse.jetty.io.CyclicTimeouts.onTimeoutExpired(CyclicTimeouts.java:110) ~[?:?]
	at org.eclipse.jetty.io.CyclicTimeouts.access$100(CyclicTimeouts.java:50) ~[?:?]
	at org.eclipse.jetty.io.CyclicTimeouts$Timeouts.onTimeoutExpired(CyclicTimeouts.java:196) ~[?:?]
	at org.eclipse.jetty.io.CyclicTimeout$Wakeup.run(CyclicTimeout.java:298) ~[?:?]
2022-10-07 05:22:35.538 [DEBUG] [ue.internal.handler.HueBridgeHandler] - Connection to Hue Bridge 192.168.0.251 lost: Total timeout 5000 ms elapsed
org.openhab.core.i18n.CommunicationException: Total timeout 5000 ms elapsed
	at org.openhab.binding.hue.internal.connection.HueBridge.doNetwork(HueBridge.java:1048) ~[bundleFile:?]
	at org.openhab.binding.hue.internal.connection.HueBridge.doNetwork(HueBridge.java:1014) ~[bundleFile:?]
	at org.openhab.binding.hue.internal.connection.HueBridge.get(HueBridge.java:997) ~[bundleFile:?]
	at org.openhab.binding.hue.internal.connection.HueBridge.getTypedLights(HueBridge.java:251) ~[bundleFile:?]
	at org.openhab.binding.hue.internal.connection.HueBridge.getFullLights(HueBridge.java:230) ~[bundleFile:?]
	at org.openhab.binding.hue.internal.handler.HueBridgeHandler$2.updateLights(HueBridgeHandler.java:248) ~[bundleFile:?]
	at org.openhab.binding.hue.internal.handler.HueBridgeHandler$2.doConnectedRun(HueBridgeHandler.java:239) ~[bundleFile:?]
	at org.openhab.binding.hue.internal.handler.HueBridgeHandler$PollingRunnable.run(HueBridgeHandler.java:136) [bundleFile:?]
	at java.util.concurrent.FutureTask.runAndReset(FutureTask.java:305) [?:?]
	at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:305) [?:?]
2022-10-07 05:22:35.544 [DEBUG] [ue.internal.handler.HueBridgeHandler] - Bridge connection lost. Updating thing status to OFFLINE.
2022-10-07 05:22:35.567 [DEBUG] [ue.internal.handler.HueSensorHandler] - bridgeStatusChanged OFFLINE: @text/offline.bridge-connection-lost
[...]
2022-10-07 05:22:35.602 [DEBUG] [ue.internal.handler.HueBridgeHandler] - Connection to Hue Bridge 192.168.0.251 established.
2022-10-07 05:22:35.604 [DEBUG] [ue.internal.handler.HueBridgeHandler] - Bridge connection resumed.
```

I don't know why, but previously I didn't experience this. I even restarted openHAB yesterday to get rid of any slowness caused by memory leaks. I see in the code that timeout was previously also 5 seconds, so I don't have an idea right now why this behavior seems different. How many concurrent connections can be made from shared Jetty client instance?

@cweitkamp - do you have any ideas?

andrewfg:
^
Just to report that I am also running the latest build, and I am NOT seeing any exceptions..

cweitkamp:
> How many concurrent connections can be made from shared Jetty client instance?

The client is configured to use at [maximum 2 connections per destination](https://github.com/openhab/openhab-core/blob/4096df1c6433f10e332632ced9f2e679dd8c340d/bundles/org.openhab.core.io.net/src/main/java/org/openhab/core/io/net/http/internal/WebClientFactoryImpl.java#L273). It uses 5 up to 10 threads. Those values are user specific (see https://github.com/openhab/openhab-core/blob/4096df1c6433f10e332632ced9f2e679dd8c340d/bundles/org.openhab.core.io.net/src/main/java/org/openhab/core/io/net/http/internal/WebClientFactoryImpl.java#L156-L163).

andrewfg:
Just to mention that `HueTlsTrustManagerProvider.java` would still need some work in order for it to support the certificate validation for SSE connections (as I have found in #13570 ) .. however I think that such work could be done in the latter PR rather than this one. And also I would need @cweitkamp to help me please :)

jlaur:
> > How many concurrent connections can be made from shared Jetty client instance?
> 
> The client is configured to use at [maximum 2 connections per destination](https://github.com/openhab/openhab-core/blob/4096df1c6433f10e332632ced9f2e679dd8c340d/bundles/org.openhab.core.io.net/src/main/java/org/openhab/core/io/net/http/internal/WebClientFactoryImpl.java#L273). It uses 5 up to 10 threads. Those values are user specific (see https://github.com/openhab/openhab-core/blob/4096df1c6433f10e332632ced9f2e679dd8c340d/bundles/org.openhab.core.io.net/src/main/java/org/openhab/core/io/net/http/internal/WebClientFactoryImpl.java#L156-L163).

Thanks. I will need to figure out how to add some more detailed logging when this happens, as it does happen consistently a few times per week now. It wouldn't be much of a problem if my rules were more well-written - currently I monitor all my devices and get push messages when they are offline/unreachable, but I don't consider if the reason is the bridge being offline, so that triggers are lot of messages. :-) It might be the number of threads being the limitation as some of my other bindings are quite busy all the time.

jlaur:
> Just to report that I am also running the latest build, and I am NOT seeing any exceptions..

To see the exception, DEBUG log level would need to be set. Have your checked in logs if the bridge has come **OFFLINE** at any moment, even just for a few seconds? I'm suspecting (for me) if may be caused by activity from other bindings using the shared http client.

jlaur:
@lolodomo, @cweitkamp - although the switch to shared Jetty client has changed some behavior related to timeouts in my setup, I believe this is a general problem for which this binding is not to blame, probably it's just a symptom. So IMHO it would be okay to merge if no one else but me experienced any issues. I will still try to track this further.

lolodomo:
My only remaining comments were about exception handling but I believe you concluded there is in fact no problem. As a consequence, that is fine for me,. Feel free to merge @jlaur

jaywiseman1971:
Can someone post the JAR link so I can test this?  

I have 2 HUE bridges with over 50 devices across them.  I'm running OH 3.2 and the same HUE version binding, my issue is the HUE bridges go offline/online throughout the day very often (just offline for a few milliseconds).  Reading this thread, is this because it's using a shared HTTP client?  If this is solved, this would be great!

jlaur:
> Can someone post the JAR link so I can test this?

This version of the binding requires core 3.3 or higher, so you would need to upgrade before being able to test it.

I actually experienced the opposite problem. Shared Jetty client was introduced with this PR and I didn't have timeouts before I began testing this.

cweitkamp:
@jlaur Which other bindings do you use in your environment based on the shared Jetty client?

jlaur:
> @jlaur Which other bindings do you use in your environment based on the shared Jetty client?

* **miele:** Polling interval 15 seconds, each time one call to get devices and one call per device - I have 5. So 18 calls per minute when idle, but they are relatively fast.
* **hdpowerview:** Polling interval 1 minute, 4 calls per iteration. Relatively fast. However, every 15 minutes a hard refresh per shade. I have 11, and these calls are quite expensive. Additionally every 12 hours 11 very expensive calls to fresh battery status.
* **boschindego:** Polling interval 3 minutes, one call per iteration. Quite cheap.
* **meater:** Polling interval 30 seconds, so 2 calls per minute.
* **wemo:** One call per minute through `HttpUtil.executeUrl`, I believe.
* netatmo
* harmonyhub
* unifi: Actually doesn't. Mentioned only because I found this interesting comment: https://github.com/openhab/openhab-addons/blob/c9481d5c36a658439f96410ab0132ef754c9c28f/bundles/org.openhab.binding.unifi/src/main/java/org/openhab/binding/unifi/internal/UniFiThingHandlerFactory.java#L59-L61

cweitkamp:
Interesting. And what about the polling configuration of the hue binding itself? IIRC the binding has three polling jobs. One for lights including groups - frequency is configurable by user in seconds , one for sensors - interval configurable too (in ms, disabled, if 0 or less) and one for scenes - fixed value (10 minutes). Plus one request for each command sent towards Hue Bridge.

jlaur:
> And what about the polling configuration of the hue binding itself? IIRC the binding has three polling jobs. One for lights including groups - frequency is configurable by user in seconds , one for sensors - interval configurable too (in ms, disabled, if 0 or less) and one for scenes - fixed value (10 minutes). Plus one request for each command sent towards Hue Bridge.

Polling interval is 10 seconds (default). I remember now that I lowered Sensor Polling Interval to 250 ms. I believe that was an attempt to slightly improve responsiveness of some FoH buttons (in lack of SSE).

andrewfg:
> in lack of SSE

=> FYI I am working on SSE for Hue and HDPowerview :)

=> also the Gardena binding uses the shared Jetty client; although it has WebSocket call backs (similar to SSE) so the load is low; but one user has complained that when the WebSocket drops, it breaks other bindings; which may or may not be related ??

https://community.openhab.org/t/gardena-binding-throws-connection-errors-and-seems-to-break-some-bindings/140085

jlaur:
@cweitkamp - thanks again!

PR #11849, opened 2021-12-23
[MQTT] Add additional logging information

LukasA83:
Add additional logging information when command is not supported by type.


lolodomo (approved):
LGTM

PR #11851, opened 2021-12-23
[boschshc] Add default translations

wborn:
This makes more texts translatable which helps to prevent some translations from being removed in #11832.


cweitkamp (approved):
LGTM. Thanks.

PR #11854, opened 2021-12-24
[myStrom] Request info is not supported by the first generation of plug

### LINKED ISSUE #10432: [myStrom] Smart Plug with the firmware 2.68 state Unknown"
The old myStrom Smart Plug with the firmware 2.68 remain on Unknown. The new Mystrom Smart Plug in with the Frimware 3.82 work fine.
myStrom 

## Expected Behavior
 If you're describing a bug, tell us what should happen 
 If you're suggesting a change/improvement, tell us how it should work 

## Current Behavior
The Thing status remains on Unknown and not on Online. 

## Context
The status is not updated., Thing deleted and recreated. 

## Your Environment
Opnehab 3.1.0  Debian 10, Openhabian, Java 8 Nodes.js 12.

Fredo70:
closes #10432

If the request to "http://[Switch IP]/api/v1/info" fails (not supported by first generatino of plug), it will not prevent the plug from working because it does not require any information. For the Bulb, the information "mac" is required. With this Thing, it will fail as before.

https://github.com/openhab/openhab-addons/issues/10432
https://community.openhab.org/t/mystrom-smart-plugs-with-firmware-2-68-not-work-in-oh3/124427/6

Signed-off-by: Frederic Chastagnol <fchastagnol@fredoware.ch>

### DIFF
--- a/bundles/org.openhab.binding.mystrom/src/main/java/org/openhab/binding/mystrom/internal/AbstractMyStromHandler.java
+++ b/bundles/org.openhab.binding.mystrom/src/main/java/org/openhab/binding/mystrom/internal/AbstractMyStromHandler.java
@@ -47,8 +47,10 @@ import org.openhab.core.thing.ThingStatus;
 import org.openhab.core.thing.ThingStatusDetail;
 import org.openhab.core.thing.binding.BaseThingHandler;
 import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import com.google.gson.Gson;
+import com.google.gson.JsonSyntaxException;
 
 /**
  * The {@link AbstractMyStromHandler} is responsible for handling commands, which are
@@ -65,6 +67,7 @@ public abstract class AbstractMyStromHandler extends BaseThingHandler {
     protected String hostname = "";
     protected String mac = "";
 
+    private final Logger logger = LoggerFactory.getLogger(AbstractMyStromHandler.class);
     private @Nullable ScheduledFuture<?> pollingJob;
     protected final Gson gson = new Gson();
 
@@ -116,12 +119,10 @@ public abstract class AbstractMyStromHandler extends BaseThingHandler {
                     Locale.getDefault());
             properties.put(PROPERTY_LAST_REFRESH, formatter.format(calendar.getTime()));
             updateProperties(properties);
-        } catch (Exception ex) {
-            getLogger().debug("Updating properties failed: ", ex);
+        } catch (JsonSyntaxException | MyStromException ex) {
+            logger.debug("Updating properties failed: ", ex);
         }
     }
-
-    protected abstract Logger getLogger();
 
     /**
      * Calls the API with the given http method, request path and actual data.
--- a/bundles/org.openhab.binding.mystrom/src/main/java/org/openhab/binding/mystrom/internal/AbstractMyStromHandler.java
+++ b/bundles/org.openhab.binding.mystrom/src/main/java/org/openhab/binding/mystrom/internal/AbstractMyStromHandler.java
@@ -46,6 +46,7 @@ import org.openhab.core.thing.Thing;
 import org.openhab.core.thing.ThingStatus;
 import org.openhab.core.thing.ThingStatusDetail;
 import org.openhab.core.thing.binding.BaseThingHandler;
+import org.slf4j.Logger;
 
 import com.google.gson.Gson;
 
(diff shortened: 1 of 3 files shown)


bundles/org.openhab.binding.mystrom/src/main/java/org/openhab/binding/mystrom/internal/MyStromBulbHandler.java
```diff
@@ -154,6 +154,18 @@ public void handleCommand(ChannelUID channelUID, Command command) {
...
     }
 
+    @Override
+    protected final Logger getLogger() {
+        return logger;
+    }
+
+    @Override
+    protected void checkRequiredInfo() throws MyStromException {
+        if (mac.equals("")) {
```
lolodomo, line 164: Please rather use `isBlank` method

bundles/org.openhab.binding.mystrom/src/main/java/org/openhab/binding/mystrom/internal/AbstractMyStromHandler.java
```diff
@@ -91,30 +92,37 @@ public final void dispose() {
...
-        properties.put(PROPERTY_DNS, deviceInfo.dns);
-        properties.put(PROPERTY_STATIC, Boolean.toString(deviceInfo.staticState));
-        properties.put(PROPERTY_CONNECTED, Boolean.toString(deviceInfo.connected));
-        Calendar calendar = Calendar.getInstance();
-        DateFormat formatter = DateFormat.getDateTimeInstance(DateFormat.SHORT, DateFormat.MEDIUM, Locale.getDefault());
-        properties.put(PROPERTY_LAST_REFRESH, formatter.format(calendar.getTime()));
-        updateProperties(properties);
     }
 
+    protected abstract Logger getLogger();
```
lolodomo, line 124: This is not really a problem to have your logger in the abstract class.
Doing that like you did is not wrong but I never saw that in openHAB bindings.

bundles/org.openhab.binding.mystrom/src/main/java/org/openhab/binding/mystrom/internal/AbstractMyStromHandler.java
```diff
@@ -91,30 +92,37 @@ public final void dispose() {
...
+            properties.put(PROPERTY_GW, deviceInfo.gw);
+            properties.put(PROPERTY_DNS, deviceInfo.dns);
+            properties.put(PROPERTY_STATIC, Boolean.toString(deviceInfo.staticState));
+            properties.put(PROPERTY_CONNECTED, Boolean.toString(deviceInfo.connected));
+            Calendar calendar = Calendar.getInstance();
+            DateFormat formatter = DateFormat.getDateTimeInstance(DateFormat.SHORT, DateFormat.MEDIUM,
+                    Locale.getDefault());
+            properties.put(PROPERTY_LAST_REFRESH, formatter.format(calendar.getTime()));
+            updateProperties(properties);
+        } catch (Exception ex) {
```
lolodomo, line 119: Please catch specific exceptions that can be thrown by your code.

lolodomo (changes requested):

bundles/org.openhab.binding.mystrom/src/main/java/org/openhab/binding/mystrom/internal/AbstractMyStromHandler.java
Fredo70: Ok. Isn't it important that each handler class (MyStromBulbHandler and MyStromPlugHandler) has its own logger?

bundles/org.openhab.binding.mystrom/src/main/java/org/openhab/binding/mystrom/internal/AbstractMyStromHandler.java
Fredo70: is this ok?
```
    } catch (JsonSyntaxException | MyStromException ex) {
        getLogger().debug("Updating properties failed: ", ex);
    }
```
Or should each exception be treated separately?

bundles/org.openhab.binding.mystrom/src/main/java/org/openhab/binding/mystrom/internal/AbstractMyStromHandler.java
lolodomo: No.
In fact, this is even a bad idea to do like you did, because the logger name is generally used to know in which class the log was issued. We loose this information with the you implement it.

bundles/org.openhab.binding.mystrom/src/main/java/org/openhab/binding/mystrom/internal/AbstractMyStromHandler.java
lolodomo: It is ok

lolodomo (approved):
LGTM

PR #11860, opened 2021-12-26
[LuxtronikHeatpump] Adds additional channels

sgiehl:
Based on some forum discussions like https://community.openhab.org/t/luxtronikheatpump-finding-the-currently-drawn-power/121605 and some private feedback. We were able to figure out that for heat pumps with a software version > 3.X the socket serves up to 260 values.
As those heat pumps do no longer serve a java applet, but use a web ui, that uses a web socket connection instead it's not possible to look up the channels in the applet code (like it was done for all other channels). 

This PR adds all those additional channels (232 - 260). 

By now we were able to determine three channels based on their values and which value they matched on a heat pump.
The other channels are currently named based in their number.

I'm not sure if it's suitable to add such "unknown" channels. If not I can also adjust the PR to only contain the new channels we were able to determine yet.


bundles/org.openhab.binding.luxtronikheatpump/src/main/resources/OH-INF/thing/channels.xml
```diff
@@ -1947,6 +1947,181 @@
 	</channel-type>
 
 
+	<channel-type id="channel232" advanced="true">
+		<item-type>Number</item-type>
+		<label>Channel 232</label>
+		<state readOnly="true"/>
+	</channel-type>
```
lolodomo, line 1954: Maybe create only one channel type for all unknown channels.

bundles/org.openhab.binding.luxtronikheatpump/src/main/resources/OH-INF/i18n/luxtronikheatpump_de.properties
```diff
@@ -271,6 +271,37 @@ channel-type.luxtronikheatpump.temperatureRoom.label = Raumtemp. Istwert
 channel-type.luxtronikheatpump.temperatureRoomTarget.label = Raumtemp. Sollwert
 channel-type.luxtronikheatpump.temperatureHotWaterTop.label = Temp. Brauchwasser Oben
 channel-type.luxtronikheatpump.frequencyCompressor.label = Verdichterfrequenz
+channel-type.luxtronikheatpump.channel232.label = Kanal 232
```
lolodomo, line 274: Please update the default translation file : luxtronikheatpump.properties

I would recommend to not update the German translation in this PR but to do it later in Crowdin when the PR is merged.
If you do it in that PR, this will require a manual action in Crowdin to load your changes ; there is a risk to loose translations.

lolodomo (changes requested):

lolodomo:
> I'm not sure if it's suitable to add such "unknown" channels. If not I can also adjust the PR to only contain the new channels we were able to determine yet.

I cannot really answer, I let you decide but my feeling is that it is a little strange to add channels you even do not know what purpose they are.

bundles/org.openhab.binding.luxtronikheatpump/src/main/resources/OH-INF/i18n/luxtronikheatpump_de.properties
sgiehl: Ah ok. I'm not very familiar with crowdin yet. We are using weblate in our projects, which has some kind of sync in both directions. But I can also add them to Crowdin once this PR was merged.

bundles/org.openhab.binding.luxtronikheatpump/src/main/resources/OH-INF/thing/channels.xml
sgiehl: I'm not sure what exactly needs to be changed here. Could you explain that a bit more?

lolodomo: Rather having many channel types with a number, you could just define a channel type for unknown data with a description explaining the situation, like for example:
```
	<channel-type id="unknownChannel" advanced="true">
		<item-type>Number</item-type>
		<label>Unknown Channel</label>
		<description>blablabla</description>
		<state readOnly="true"/>
	</channel-type>
```

bundles/org.openhab.binding.luxtronikheatpump/src/main/resources/OH-INF/i18n/luxtronikheatpump_de.properties
lolodomo: But update the default translation file with English properties.

lolodomo: You just have to run this command: `mvn i18n:generate-default-translations` at the root of your binding.

sgiehl: ok. did that and pushed the update

bundles/org.openhab.binding.luxtronikheatpump/src/main/resources/OH-INF/thing/channels.xml
sgiehl: tried to change that accordingly. Hope that is what you were referring to.

lolodomo: Yes, exactly.
But as you want a different channel label for each channel, something is missing.
I am going to push a PR to your repo.

lolodomo:
PR submitted to your repo/branch.

Note that I would recommend to not change the German translations at all in this PR, and do it later in Crowdin.
If you don't do that, German translations in Crowdin will be outdated, and the outdated  texts will then be pushed here the next time one of the text is changed in the German translation.

bundles/org.openhab.binding.luxtronikheatpump/src/main/resources/OH-INF/i18n/luxtronikheatpump_de.properties
lolodomo: Just remove all changes in the German properties.
You (or someone) will translate in German later in Crowdin.

sgiehl:
@lolodomo thanks for that. I've reverted the changes to the german language file

lolodomo (approved):
LGTM

lolodomo:
> @lolodomo thanks for that. I've reverted the changes to the german language file

Thank you.
I will merge as soon as the build succeeds.

lolodomo:
Looking now in Crowdin, I see a lot of translations are missing in German for this binding, not only the ones you just added.

sgiehl:
@lolodomo thanks for merging. I'll check crowdin and add the missing translations there...

tmp-hallenser:
Might not be the right place for feedback

The Luxtronik binding is not working properly anymore in the latest docker snapshot built. As this is the only change to the binding, I though this might be the root cause. The Thing remains in state 'unknown' (in OH3.2 all was working fine). The Thing created in OH3.2 even survives the migration to OH3.3, but lacks the additional channels. My heatpump is running V3.85.8 (r7193) if that's relevant.

The are no DEBUG, INFO or ERROR messages in the log hinting at a connection issue.

sgiehl:
Hi @tmp-hallenser 
Thanks for the feedback.
I was able to reproduce that locally. Will try to create a fix for that tomorrow.

PR #11862, opened 2021-12-26
[dscalarm] Update README.md

yfaway:
Fix 1 spelling error. Mark several channels as writable. Add link to the PDF file that contains the list of EnvisaLink commands.


lolodomo:
@RSStephens for review

lolodomo (changes requested):
Looking at the code, your changes are correct but you missed two channels which are writtable too: zone_bypass_mode and panel_command. This second channel is apparently even not documented at all.

lolodomo (approved):
LGTM

lolodomo:
Not correctly signed off but it is only a minor change in the documentation.

PR #11873, opened 2021-12-28
[miele] Refactor mDNS discovery

jlaur:

Brought back some minor stashed refactoring done while investigating #11837, which was fixed by openhab/openhab-core#2635 and #11834:
- Added nullness annotations, and with that fixed: _Unsafe null type conversion (type annotations)_
- Fixed warning: _The import org.openhab.core.config.discovery.mdns.internal.MDNSDiscoveryService is never used_
- Fixed error handling problem that could lead to discovery result without any properties.
- Localized discovery result label.
- Simplified some code by having constants on left-side.
- Reduced nesting.
- Reduce log level from **error** to **warning** when failing to create discovery result.

### DIFF
--- a/bundles/org.openhab.binding.miele/src/main/java/org/openhab/binding/miele/internal/discovery/MieleMDNSDiscoveryParticipant.java
+++ b/bundles/org.openhab.binding.miele/src/main/java/org/openhab/binding/miele/internal/discovery/MieleMDNSDiscoveryParticipant.java
@@ -131,7 +131,7 @@ public class MieleMDNSDiscoveryParticipant implements MDNSDiscoveryParticipant {
             logger.debug("Discovered Miele@home gateway with IP address {} and interface IP address {}", ipAddress,
                     interfaceIpAddress);
         } catch (IOException e) {
-            logger.error("An exception occurred while connecting to the Miele Gateway: '{}'", e.getMessage());
+            logger.warn("An exception occurred while connecting to the Miele Gateway: '{}'", e.getMessage());
             return null;
         }
 
--- a/bundles/org.openhab.binding.miele/src/main/java/org/openhab/binding/miele/internal/discovery/MieleMDNSDiscoveryParticipant.java
+++ b/bundles/org.openhab.binding.miele/src/main/java/org/openhab/binding/miele/internal/discovery/MieleMDNSDiscoveryParticipant.java
@@ -15,19 +15,19 @@ package org.openhab.binding.miele.internal.discovery;
 import java.io.IOException;
 import java.net.InetAddress;
 import java.net.Socket;
-import java.util.Collections;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.Set;
 
 import javax.jmdns.ServiceInfo;
 
+import org.eclipse.jdt.annotation.NonNullByDefault;
 import org.eclipse.jdt.annotation.Nullable;
 import org.openhab.binding.miele.internal.MieleBindingConstants;
+import org.openhab.binding.miele.internal.handler.MieleBridgeHandler;
 import org.openhab.core.config.discovery.DiscoveryResult;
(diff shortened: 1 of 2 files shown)


bundles/org.openhab.binding.miele/src/main/java/org/openhab/binding/miele/internal/discovery/MieleMDNSDiscoveryParticipant.java
```diff
@@ -70,83 +79,88 @@ public void modified(@Nullable Map<String, Object> configProperties) {
      * @param configProperties the passed configuration parameters.
      */
     private void updateRemovalGracePeriod(Map<String, Object> configProperties) {
```
lolodomo, line 81: You are now calling this method with a nullable map so you should mark the param as @nullable and keep your test.

bundles/org.openhab.binding.miele/src/main/java/org/openhab/binding/miele/internal/discovery/MieleMDNSDiscoveryParticipant.java
```diff
@@ -70,83 +79,88 @@ public void modified(@Nullable Map<String, Object> configProperties) {
...
+            socket = new Socket(addresses[0], 80);
+            InetAddress ourAddress = socket.getLocalAddress();
+            String interfaceIpAddress = ourAddress.getHostAddress();
+            socket.close();
+
+            properties.put(MieleBindingConstants.INTERFACE, interfaceIpAddress);
+            logger.debug("Discovered Miele@home gateway with IP address {} and interface IP address {}", ipAddress,
+                    interfaceIpAddress);
+        } catch (IOException e) {
+            logger.error("An exception occurred while connecting to the Miele Gateway: '{}'", e.getMessage());
```
lolodomo, line 134: warn is more adapted. error is reserved for critical problem impacting the full openHAB system.

lolodomo (changes requested):

bundles/org.openhab.binding.miele/src/main/java/org/openhab/binding/miele/internal/discovery/MieleMDNSDiscoveryParticipant.java
jlaur: Fixed.

bundles/org.openhab.binding.miele/src/main/java/org/openhab/binding/miele/internal/discovery/MieleMDNSDiscoveryParticipant.java
jlaur: I don't see a point in calling a method with null just so it can do nothing? `activate` and `modified` are declared with `@Nullable` so can't do anything about that, but I can at least avoid propagating that further.

jlaur: Oh wait, now read your comment again. I think you didn't see the changes in `activate` and `modifed`? `updateRemovalGracePeriod` is no longer called with a nullable map. This would result in warnings, as its not declared nullable anymore.

lolodomo: activate and modified could be called with a null parameter.

`public void activate(@Nullable Map<String, Object> configProperties) {`

lolodomo: Ok, I thought the null test in `activate` and `modified` was removed, while in fact it is added.
So of course, it is ok.

lolodomo (approved):
LGTM

PR #11877, opened 2021-12-28
[googletts] Use returned sound to get play informations

### LINKED ISSUE #10015: [googletts] Only white noise
Platform information:
Hardware: Raspberry Pi 4B 2GB
OS: Openhabian 1.6.2 64Bit
Java Runtime Environment: Java 11
openHAB version: 3.0.1 release build

When openhab service is stopped the voice played bij Google TTS service via 3.5mm jack output, i only hear white noise. Sometimes rebooting the raspberry pi solves the problem. Every time I restart the openhab service, this problem occurs again.

What I checked:
- All google voices are present
- I can select them in the UI, i tried a couple Wavenet and the ‘default-ones’, doesn’t change
- VoiceRSS works fine
- The wav files (temp folder) play back nicely when opened on another device
- Google limits is ok
- I’m just using openhab:voice say test on karaf > no difference
- openhab:audio play barking.mp3 < works fine
- playback via chromecast is fine

Not sure how to fix it.

***-  At this very moment, I had to reboot the Pi > White noise >restarted openhab > white noise when 'saying' something... I cleared the cache with [openhab-cli clean-cache] >>> This solved the problem this time

dalgwen:
When google tts returns a wav file, it is now parsed to get correct informations about it.
(It is mandatory for playing it with, at least, the System speaker audio sink)

This is the fix for the issue [discuted here](https://community.openhab.org/t/oh3-google-tts-only-white-noise/112714).

[Snapshot release](https://github.com/dalgwen/openhab-addons/releases/tag/3.3.0-snapshot-googleTTS-fix_white_noise)

Close #10015

Signed-off-by: Gwendal Roulleau <gwendal.roulleau@gmail.com>

### DIFF
--- a/bundles/org.openhab.voice.googletts/src/main/java/org/openhab/voice/googletts/internal/GoogleTTSService.java
+++ b/bundles/org.openhab.voice.googletts/src/main/java/org/openhab/voice/googletts/internal/GoogleTTSService.java
@@ -24,17 +24,13 @@ import java.util.Locale;
 import java.util.Map;
 import java.util.Set;
 
-import javax.sound.sampled.AudioFormat.Encoding;
-import javax.sound.sampled.AudioInputStream;
-import javax.sound.sampled.AudioSystem;
-import javax.sound.sampled.UnsupportedAudioFileException;
-
 import org.eclipse.jdt.annotation.NonNullByDefault;
 import org.eclipse.jdt.annotation.Nullable;
 import org.openhab.core.OpenHAB;
 import org.openhab.core.audio.AudioFormat;
 import org.openhab.core.audio.AudioStream;
 import org.openhab.core.audio.ByteArrayAudioStream;
+import org.openhab.core.audio.utils.AudioWaveUtils;
 import org.openhab.core.auth.client.oauth2.OAuthFactory;
 import org.openhab.core.config.core.ConfigurableService;
 import org.openhab.core.voice.TTSException;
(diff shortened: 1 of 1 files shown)


bundles/org.openhab.voice.googletts/src/main/java/org/openhab/voice/googletts/internal/GoogleTTSService.java
```diff
@@ -332,6 +340,38 @@ public AudioStream synthesize(String text, Voice voice, AudioFormat requestedFor
...
+            if (javaSoundencoding.equals(Encoding.PCM_SIGNED.toString())) {
+                codecPCMSignedOrUnsigned = AudioFormat.CODEC_PCM_SIGNED;
+            } else if (javaSoundencoding.equals(Encoding.PCM_UNSIGNED.toString())) {
+                codecPCMSignedOrUnsigned = AudioFormat.CODEC_PCM_UNSIGNED;
+            } else if (javaSoundencoding.equals(Encoding.ULAW.toString())) {
+                codecPCMSignedOrUnsigned = AudioFormat.CODEC_PCM_ULAW;
+            } else if (javaSoundencoding.equals(Encoding.ALAW.toString())) {
+                codecPCMSignedOrUnsigned = AudioFormat.CODEC_PCM_ALAW;
+            } else {
+                codecPCMSignedOrUnsigned = null;
```
lolodomo, line 365: In case codecPCMSignedOrUnsigned  is null (not identified), do you think your computations of bitRate and frequency are relevant ?
Or should you consider the default format in this case ?

bundles/org.openhab.voice.googletts/src/main/java/org/openhab/voice/googletts/internal/GoogleTTSService.java
```diff
@@ -332,6 +340,38 @@ public AudioStream synthesize(String text, Voice voice, AudioFormat requestedFor
...
         }
-        return new ByteArrayAudioStream(audio, requestedFormat);
+
+        // compute the real format returned by google
+        AudioFormat realAudioFormat = parseAudioFormat(audio);
+
+        return new ByteArrayAudioStream(audio, realAudioFormat);
+    }
+
+    private AudioFormat parseAudioFormat(byte[] audio) {
```
cweitkamp, line 350: Thanks for looking into this.

Two questions:
1. Should we wait for https://github.com/openhab/openhab-core/pull/2650 and use your new constructor for passing channels to the generated `AudioFormat`?
2. This is nearly the same code as used in https://github.com/openhab/openhab-core/pull/2650 and thus a util method in OHC makes a lot of sense to reduce duplicate code. Doesn't it?

dalgwen: 1 & 2 : Yes it seems logical. Good idea.
The googletts addon won't be compatible with the 3.2 branch tough. Is it an issue ?
I'm not familiar with "coupling" openhab-core and openhab-addons. Is there any special precaution to take ?

bundles/org.openhab.voice.googletts/src/main/java/org/openhab/voice/googletts/internal/GoogleTTSService.java
dalgwen: I honestly don't know.
If the codec is not recognized, or not one of the few mentionned here, the probability that the sink will play the sound without errors is pretty low. The value computed, or the default one, won't matter in the end.
I suppose it depends on the sink responsible for playing the sound and it's robustness.
Do you have a better educated guess ?

bundles/org.openhab.voice.googletts/src/main/java/org/openhab/voice/googletts/internal/GoogleTTSService.java
cweitkamp: With https://github.com/openhab/openhab-core/pull/2650 in place you might want to reconsider your code and use the new features and util classes in this PR.

dalgwen: Thanks ! Done.
But the checks (continuous development build) failed.
Maybe I was too fast, and the change in openhab-core was not ready for openhab-addon ? If so, do you know when, and if I can trigger another build (apart from force pushing something) ?

lolodomo (approved):
LGTM

bundles/org.openhab.voice.googletts/src/main/java/org/openhab/voice/googletts/internal/GoogleTTSService.java
cweitkamp: The code is built and published nightly. After a merge I OHC it will be available in the next day for openHAB Add-ons and other repositories.

PR #11884, opened 2021-12-29
[shelly] Fix comment in translations file

wborn:
This fixes the comment showing up in the translations.
The code seems to use the 'message.discovery.failed' instead of 'discovery.failed'.


cweitkamp (approved):
LGTM. Thanks.

PR #11889, opened 2021-12-29
[astro] Fixed timeLeft calculation at the end of the year

### LINKED ISSUE #11859: [astro]:sun:Sun:season#timeLeft returns -300 days for Summer on 26 Dec 2021 in Australia
## Expected Behavior

On 26 December 2021,  it should return 65 days until the end of Summer on 1 March 2022 for Australia in the Southern hemisphere

## Current Behavior

Channel returns -300 days which is the numbers of days from the start of the previous Autumn season.

Australia uses metorological seasons and not astronomical seasons, so Summer runs from 1 December until 28 February the following year.

## Possible Solution

In `org.openhab.binding.astro.internal.model`

    public Calendar getNextSeason() {
        return DateTimeUtils.getNextFromToday(spring, summer, autumn, winter);
    }

assumes `spring` is the first season of the year, when in Australia `summer` is both the first and last season of the year.

## Context

Bug returns invalid number of days remaining for current season.

Country/Region: Australia

`2021-12-26 00:00:30.338 [INFO ] [openhab.event.ItemStateChangedEvent ] - Item 'Season_Time_Left' changed from -299.0003522800926 day to -300.00035108796294 day`

openHAB Season_Time_Left

## Your Environment
openHAB 3.2.0.M5
Ubuntu Linux 21.10

Hilbrand:
Season timeleft calculated the wrong time at the end of the year, because the seasons it measures the time with are the seasons within the same year. So for times at the end of the year it computed against the season at the beginning of the year, resulting in negative number of days.
Also changed the time of timeLeft to actual days instead of quantity unit days. Because with the milliseconds internal resolution it will generate events each and every time this channel is updated as milliseconds change often.

Closes  #11859

### DIFF
--- a/bundles/org.openhab.binding.astro/src/main/java/org/openhab/binding/astro/internal/model/Season.java
+++ b/bundles/org.openhab.binding.astro/src/main/java/org/openhab/binding/astro/internal/model/Season.java
@@ -12,8 +12,8 @@
  */
 package org.openhab.binding.astro.internal.model;
 
-import static org.openhab.core.library.unit.MetricPrefix.MILLI;
-
+import java.time.Duration;
+import java.time.temporal.ChronoUnit;
 import java.util.Calendar;
 
 import javax.measure.quantity.Time;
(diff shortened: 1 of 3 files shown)


lolodomo:
@clinique: as you were involved in astro calculations at a time, can you please tell us if these changes look ok to you ?
At least, it looks ok to me.

clinique (approved):
LGTM

lolodomo (approved):
LGTM

PR #11891, opened 2021-12-30
[danfossairunit] Fix inability to recover from network issues on initialization

### LINKED ISSUE #11304: [danfossairunit] When CCM module is unreachable on binding initialization it will never recover
When working on #11172 I uncovered some network reliability issues with this binding. Some were fixed, but during my tests I noticed a distinct scenario still not working correctly (not a regression, it never did): When the CCM module is unreachable during the handler's `initialize` method, the binding will stay offline forever.

Additionally, it seems like `initialize` might not return quickly in this scenario, as required by the framework.

## Expected Behavior
Binding should continuously try to establish a connection to recover from this.

## Current Behavior
Binding will try once to connect in order for fetch unit name and serial number. If this fails, the polling job will not be scheduled and the binding is essentially dead.

## Possible Solution
The problem is in the initialize method:

```
            scheduler.execute(() -> {
                try {
                    thing.setProperty(PROPERTY_UNIT_NAME, danfossAirUnit.getUnitName());
                    thing.setProperty(PROPERTY_SERIAL, danfossAirUnit.getUnitSerialNumber());
                    pollingJob = scheduler.scheduleWithFixedDelay(this::updateAllChannels, 5, config.refreshInterval,
                            TimeUnit.SECONDS);
                } catch (IOException e) {
                    updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.OFFLINE.COMMUNICATION_ERROR, e.getMessage());
                }
            });
```

getUnitName() and getUnitSerialNumber are blocking network calls. In case of network issues they might be slow. To my understanding scheduler.execute() may execute in callers thread, in which case initialize will not return quickly:
https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/concurrent/Executor.html#execute(java.lang.Runnable)

"Executes the given command at some time in the future. **The command may execute** in a new thread, in a pooled thread, or **in the calling thread**, at the discretion of the Executor implementation."

After timeout an IOException is thrown and caught here, causing the thing to be marked as offline and the polling job scheduling to be skipped.

## Steps to Reproduce
Unplug network cable to CCM module or turn off the air unit. Restart the binding. Connect the CCM module again or turn on the air unit.

jlaur:
Fixes #11304

When working on #11172 I uncovered some network reliability issues with this binding. Some were fixed, but during my tests I noticed a distinct scenario still not working correctly (not a regression, it never did): When the CCM module is unreachable during the handler's `initialize` method, the binding will stay offline forever.

Binding will try once to connect in order for fetch unit name and serial number. If this fails, the polling job will not be scheduled and the binding is essentially dead.

Tests performed:
- Restart openHAB while CCM module is unplugged. Thing will remain **OFFLINE** until CCM module is connected, then will go **ONLINE**, properties will be initialized and channels will start to update periodically. This scenario is fixed by this PR.
- While everything is running, unplug the CCM module. Thing will go **OFFLINE**. When CCM module is connected again, thing will be **ONLINE** and channels will start to update periodically again. This scenario was fixed by #11172 and still works.

### DIFF
--- a/bundles/org.openhab.binding.danfossairunit/src/main/java/org/openhab/binding/danfossairunit/internal/handler/DanfossAirUnitHandler.java
+++ b/bundles/org.openhab.binding.danfossairunit/src/main/java/org/openhab/binding/danfossairunit/internal/handler/DanfossAirUnitHandler.java
@@ -17,6 +17,8 @@ import static org.openhab.binding.danfossairunit.internal.DanfossAirUnitBindingC
 import java.io.IOException;
 import java.net.InetAddress;
 import java.net.UnknownHostException;
+import java.util.HashMap;
+import java.util.Map;
 import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.TimeUnit;
 
@@ -62,6 +64,7 @@ public class DanfossAirUnitHandler extends BaseThingHandler {
     private @Nullable ScheduledFuture<?> pollingJob;
     private @Nullable DanfossAirUnitCommunicationController communicationController;
     private @Nullable DanfossAirUnit airUnit;
+    private boolean propertiesInitializedSuccessfully = false;
 
     public DanfossAirUnitHandler(Thing thing) {
         super(thing);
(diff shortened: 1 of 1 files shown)


bundles/org.openhab.binding.danfossairunit/src/main/java/org/openhab/binding/danfossairunit/internal/handler/DanfossAirUnitHandler.java
```diff
@@ -62,6 +64,7 @@
     private @Nullable ScheduledFuture<?> pollingJob;
     private @Nullable DanfossAirUnitCommunicationController communicationController;
     private @Nullable DanfossAirUnit airUnit;
+    private boolean propertiesInitializedSuccessfully = false;
```
lolodomo, line 67: Would it make sense to also set propertiesInitializedSuccessfully to false in dispose() ?

lolodomo (commented):
Just one comment/question.

bundles/org.openhab.binding.danfossairunit/src/main/java/org/openhab/binding/danfossairunit/internal/handler/DanfossAirUnitHandler.java
jlaur: I don't think so. The properties are static (unit name and serial number), so once initialized there will be no need for doing that again.

lolodomo: Yes, considering the kind of properties, you're right.

lolodomo (approved):
LGTM

Issue #11892, opened 2021-12-30
[shelly] Allow updateInterval values of 1s


richiew:
The Shelly Binding should allow minimum updateInterval values of 1s.

### Example 1
With the minimum value of 10s on the Shelly 1 PM it's not possible to implement a consumption based rule which relies on current data (e.g. when starting a saw in our workshop we want to activate the suction system; but currently we have to wait for 10 seconds).

### Example 2
With the minimum value of 60s on the Shelly DW 2 it's not possible to immediately react to alarms raised by this component. When implementing a security system based on this sensor, lower update intervals are needed.

## My Environment
* OpenHAB 3.2.0 (Release Build)
* Operating System: openHABian 3.2.0 - Release Build

richiew:
This improvement is no longer needed for our case.

Because of the [documentation](https://www.openhab.org/addons/bindings/shelly/#binding-configuration) (thanks for the helpful comments in the forum entries) we found out, that the issue in our case was the different subnet of the openHAB server and the Shelly devies. We placed the Shelly devices in their own subnet and therefore the CoIoT did not work. As soon as we moved the devices in question into the same subnet as the openHAB server, we received immediate updates. We will also try to switch Shelly devices from multicast to unicast and register the openHAB server to figure out if this way the devices update the openHAB server as immediately as if they are in the same subnet.

PR #11900, opened 2021-12-30
[shelly] set min value to 1s for updateInterval in config.xml

rose78ch:
https://github.com/openhab/openhab-addons/issues/11892

### DIFF
--- a/bundles/org.openhab.binding.shelly/src/main/resources/OH-INF/config/config.xml
+++ b/bundles/org.openhab.binding.shelly/src/main/resources/OH-INF/config/config.xml
@@ -43,7 +43,7 @@
 			<advanced>true</advanced>
 			<default>true</default>
 		</parameter>
-		<parameter name="updateInterval" type="integer" min="10" required="true" unit="s">
+		<parameter name="updateInterval" type="integer" min="1" required="true" unit="s">
 			<label>General Update Interval</label>
 			<description>Interval to query an update from the device.</description>
 			<default>60</default>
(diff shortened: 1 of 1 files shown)


lolodomo:
@markus7017 for review. 

I guess polling each second could lead to a lot of network traffic and potential problems (for example if the poll takes more than 1 second).

lolodomo:
By the way, your change does not compile and your PR is not correctly signed.

rose78ch:
> By the way, your change does not compile and your PR is not correctly signed.

How can I do that?

rose78ch:
> @markus7017 for review.
> 
> I guess polling each second could lead to a lot of network traffic and potential problems (for example if the poll takes more than 1 second).

In the mystrom binding the minimum interval is also 1s and I don't have some problems.

lolodomo:
> > @markus7017 for review.
> > I guess polling each second could lead to a lot of network traffic and potential problems (for example if the poll takes more than 1 second).
> 
> In the mystrom binding the minimum interval is also 1s and I don't have some problems with network traffic.

I was just telling that if 10s was chosen, there could be a reason. I have myself no idea if it is reasonable or not, I do not even know/analyze what is done during the poll.
Ideally, we need the comment from the author/maintainer of this binding. That is the reason why I ping him.

lolodomo:
Note that patching the code in GitHub WEB without checking that the code is even compiling (and so of course without any test) is not the minimal quality for PRs expected by the openHAB project.
As you can see, your changes lead to the project no more compiling.

PR #11907, opened 2021-12-30
[LuxtronikHeatpump] Adds discovery service

sgiehl:
This PR adds an discovery class to the LuxtronikHeatmap binding.

As such heat pumps seem not to be discoverable using upnp or mdns (at least not the one I own), the discovery currently looks through the IPs in the current IPv4 subnet and tries connecting to the socket using port 8889 and 8888.
If the answer can be parsed the heat pump will be addable as thing using the heat pumps type as name.

Most parts of the discovery are based on / copied from the detection of `LGHomBot`, which is doing similar steps for discovery.

### DIFF
--- a/bundles/org.openhab.binding.luxtronikheatpump/README.md
+++ b/bundles/org.openhab.binding.luxtronikheatpump/README.md
@@ -25,7 +25,7 @@ This binding only supports one thing type "Luxtronik Heatpump" (heatpump).
 
 ## Discovery
 
-This binding will try to detect heat pumps that a reachable in the same IPv4 subnet.
+This binding will try to detect heat pumps that are reachable in the same IPv4 subnet.
 
 ## Thing Configuration
 
--- a/bundles/org.openhab.binding.luxtronikheatpump/README.md
+++ b/bundles/org.openhab.binding.luxtronikheatpump/README.md
@@ -23,6 +23,10 @@ Note: The whole functionality is based on data that was reverse engineered, so u
 
 This binding only supports one thing type "Luxtronik Heatpump" (heatpump).
 
+## Discovery
+
+This binding will try to detect heat pumps that a reachable in the same IPv4 subnet.
+
 ## Thing Configuration
 
 Each heatpump requires the following configuration parameters:
--- a/bundles/org.openhab.binding.luxtronikheatpump/src/main/java/org/openhab/binding/luxtronikheatpump/internal/ChannelUpdaterJob.java
+++ b/bundles/org.openhab.binding.luxtronikheatpump/src/main/java/org/openhab/binding/luxtronikheatpump/internal/ChannelUpdaterJob.java
@@ -248,7 +248,6 @@ public class ChannelUpdaterJob implements SchedulerRunnable, Runnable {
     }
 
     private void updateProperties(Integer[] heatpumpValues) {
-
         Map<String, Object> properties = getProperties(heatpumpValues);
 
         for (Map.Entry<String, Object> property : properties.entrySet()) {
@@ -264,10 +263,6 @@ public class ChannelUpdaterJob implements SchedulerRunnable, Runnable {
         String translation = translationProvider
                 .getText("channel-type.luxtronikheatpump." + name + ".state.option." + option);
         return translation == null ? "" : translation;
-    }
-
-    private void setProperty(String name, String value) {
-        handler.updateProperty(name, value);
     }
 
     private String formatHours(@Nullable Integer value) {
--- a/bundles/org.openhab.binding.luxtronikheatpump/src/main/java/org/openhab/binding/luxtronikheatpump/internal/ChannelUpdaterJob.java
+++ b/bundles/org.openhab.binding.luxtronikheatpump/src/main/java/org/openhab/binding/luxtronikheatpump/internal/ChannelUpdaterJob.java
@@ -16,6 +16,8 @@ import java.io.IOException;
 import java.time.DateTimeException;
 import java.time.Instant;
 import java.time.ZoneId;
+import java.util.HashMap;
+import java.util.Map;
 
 import javax.measure.Unit;
 
@@ -179,7 +181,7 @@ public class ChannelUpdaterJob implements SchedulerRunnable, Runnable {
         return rawValue;
     }
 
-    private String getSoftwareVersion(Integer[] heatpumpValues) {
+    private static String getSoftwareVersion(Integer[] heatpumpValues) {
         StringBuffer softwareVersion = new StringBuffer("");
 
         for (int i = 81; i <= 90; i++) {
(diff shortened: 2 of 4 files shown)


bundles/org.openhab.binding.luxtronikheatpump/README.md
```diff
@@ -23,6 +23,10 @@ Note: The whole functionality is based on data that was reverse engineered, so u
 
 This binding only supports one thing type "Luxtronik Heatpump" (heatpump).
 
+## Discovery
+
+This binding will try to detect heat pumps that a reachable in the same IPv4 subnet.
```
lsiepel, line 28: that are reachable

bundles/org.openhab.binding.luxtronikheatpump/src/main/java/org/openhab/binding/luxtronikheatpump/internal/discovery/LuxtronikHeatpumpDiscovery.java
```diff
@@ -0,0 +1,231 @@
...
+            ThingTypeUID typeId = LuxtronikHeatpumpBindingConstants.THING_TYPE_HEATPUMP;
+            ThingUID uid = new ThingUID(typeId, type);
+
+            DiscoveryResult result = DiscoveryResultBuilder.create(uid).withProperties(properties).withLabel(type)
+                    .build();
+            thingDiscovered(result);
+
+            return true;
+        } catch (IOException e) {
+            // Ignore, this is the expected behavior.
```
lsiepel, line 186: A trace log entry could be written.

lsiepel (commented):
Very minor comments about the code.

Maybe it makes perfectly sense, buit i'm not sure if this ip/port scanning is the desired way of doing autodiscovery. I leave that to other more experienced reviewer :)

Is this discovery tested by others? Was there a jar file or community thread about it?

bundles/org.openhab.binding.luxtronikheatpump/src/main/java/org/openhab/binding/luxtronikheatpump/internal/discovery/LuxtronikHeatpumpDiscovery.java
```diff
@@ -0,0 +1,231 @@
...
+        scanning = true;
+        ExecutorService localExecutorService = Executors.newFixedThreadPool(SCAN_THREADS);
+        executorService = localExecutorService;
+        for (int i = 0; i < addressCount; i++) {
+
+            localExecutorService.execute(() -> {
+                if (scanning && baseAdr != null) {
+                    String ipAdd = getNextIPAddress(baseAdr);
+
+                    boolean detection = (discoverFromIp(ipAdd, 8889) || discoverFromIp(ipAdd, 8888));
```
lsiepel, line 161: This var does not seem to be used

bundles/org.openhab.binding.luxtronikheatpump/src/main/java/org/openhab/binding/luxtronikheatpump/internal/ChannelUpdaterJob.java
```diff
@@ -225,20 +227,33 @@ private void setExtendedState(Integer[] heatpumpValues, Integer[] heatpumpParams
...
+        properties.put("softwareVersion", getSoftwareVersion(heatpumpValues));
+        properties.put("ipAddress", transformIpAddress(heatpumpValues[91]));
+        properties.put("subnetMask", transformIpAddress(heatpumpValues[92]));
+        properties.put("broadcastAddress", transformIpAddress(heatpumpValues[93]));
+        properties.put("gateway", transformIpAddress(heatpumpValues[94]));
+
+        return properties;
+    }
 
-        setProperty("softwareVersion", getSoftwareVersion(heatpumpValues));
```
lsiepel, line 237: private void setProperty method in this class is no longer used anywhere, could you drop it?

bundles/org.openhab.binding.luxtronikheatpump/src/main/java/org/openhab/binding/luxtronikheatpump/internal/ChannelUpdaterJob.java
```diff
@@ -225,20 +227,33 @@ private void setExtendedState(Integer[] heatpumpValues, Integer[] heatpumpParams
...
-        setProperty("ipAddress", transformIpAddress(heatpumpValues[91]));
-        setProperty("subnetMask", transformIpAddress(heatpumpValues[92]));
-        setProperty("broadcastAddress", transformIpAddress(heatpumpValues[93]));
-        setProperty("gateway", transformIpAddress(heatpumpValues[94]));
+    private void updateProperties(Integer[] heatpumpValues) {
+
```
lsiepel, line 251: Remove empty line

bundles/org.openhab.binding.luxtronikheatpump/src/main/java/org/openhab/binding/luxtronikheatpump/internal/discovery/LuxtronikHeatpumpDiscovery.java
sgiehl: I don't think that would help much. At this point an exception is thrown if no heatpump could be found for the given IP / port. As this is executed for each IP that is discovered in the network it might spam the log with useless entries.

bundles/org.openhab.binding.luxtronikheatpump/src/main/java/org/openhab/binding/luxtronikheatpump/internal/discovery/LuxtronikHeatpumpDiscovery.java
sgiehl: Yes. The variable isn't needed. But that way it could be written into one line. But can convert it into a if clause instead.

bundles/org.openhab.binding.luxtronikheatpump/src/main/java/org/openhab/binding/luxtronikheatpump/internal/ChannelUpdaterJob.java
sgiehl: yes. good find :+1:

sgiehl:
@lsiepel thanks for the review. I've applied some changes. I don't think the discovery was tested by anyone besides myself yet. It works fine with my heat pump in my local network. It seems there is no other way to discover such heat pumps and there are other bindings using similar methods.

bundles/org.openhab.binding.luxtronikheatpump/src/main/java/org/openhab/binding/luxtronikheatpump/internal/discovery/LuxtronikHeatpumpDiscovery.java
lsiepel: That is where a trace log is used for. But if you think it will not help anyone ever, leave it as it is.

lsiepel (approved):
Thanks

lolodomo (approved):
LGTM, thank you

PR #11911, opened 2021-12-31
[rotel] Fix default properties

lolodomo:
Also reorder properties by running mvn i18n:generate-default-translations


cweitkamp (approved):
LGTM. Thanks.

🎆 Happy New Year.

PR #11916, opened 2021-12-31
[voicerss] Add support for WAV audio format

abrenk:
This pull request adds support for WAV / PCM audio in all frequencies and bit depths supported by VoiceRSS.

Support for MP3, Ogg Vorbis and AAC was kept as it was (44.1 kHz, 16 bit) but could be extended based on this work.

The file-based cache was kept backwards compatible and only adds a format specifier to the file name if it is different from the default.

### DIFF
--- a/bundles/org.openhab.voice.voicerss/src/main/java/org/openhab/voice/voicerss/internal/VoiceRSSTTSService.java
+++ b/bundles/org.openhab.voice.voicerss/src/main/java/org/openhab/voice/voicerss/internal/VoiceRSSTTSService.java
@@ -50,6 +50,20 @@ public class VoiceRSSTTSService implements TTSService {
 
     // API Key comes from ConfigAdmin
     private static final String CONFIG_API_KEY = "apiKey";
+
+    /**
+     * Map from openHAB AudioFormat Codec to VoiceRSS API Audio Codec
+     */
+    private static final Map<String, String> CODEC_MAP = Map.of(AudioFormat.CODEC_PCM_SIGNED, "WAV",
+            AudioFormat.CODEC_PCM_UNSIGNED, "WAV", AudioFormat.CODEC_PCM_ALAW, "WAV", AudioFormat.CODEC_PCM_ULAW, "WAV",
+            AudioFormat.CODEC_MP3, "MP3", AudioFormat.CODEC_VORBIS, "OGG", AudioFormat.CODEC_AAC, "AAC");
+
+    /**
+     * Map from openHAB AudioFormat Frequency to VoiceRSS API Audio Frequency
+     */
+    private static final Map<Long, String> FREQUENCY_MAP = Map.of(8_000L, "8khz", 11_025L, "11khz", 12_000L, "12khz",
+            16_000L, "16khz", 22_050L, "22khz", 24_000L, "24khz", 32_000L, "32khz", 44_100L, "44khz", 48_000L, "48khz");
+
     private String apiKey;
 
     private final Logger logger = LoggerFactory.getLogger(VoiceRSSTTSService.class);
@@ -121,22 +135,12 @@ public class VoiceRSSTTSService implements TTSService {
         if (!voices.contains(voice)) {
             throw new TTSException("The passed voice is unsupported");
         }
-        boolean isAudioFormatSupported = false;
-        for (AudioFormat currentAudioFormat : audioFormats) {
-            if (currentAudioFormat.isCompatible(requestedFormat)) {
-                isAudioFormatSupported = true;
-                break;
-            }
-        }
-        if (!isAudioFormatSupported) {
-            throw new TTSException("The passed AudioFormat is unsupported");
-        }
 
-        // now create the input stream for given text, locale, format. There is
-        // only a default voice
+        // now create the input stream for given text, locale, voice, codec and format.
         try {
             File cacheAudioFile = voiceRssImpl.getTextToSpeechAsFile(apiKey, trimmedText,
-                    voice.getLocale().toLanguageTag(), voice.getLabel(), getApiAudioFormat(requestedFormat));
+                    voice.getLocale().toLanguageTag(), voice.getLabel(), getApiAudioCodec(requestedFormat),
+                    getApiAudioFormat(requestedFormat));
             if (cacheAudioFile == null) {
                 throw new TTSException("Could not read from VoiceRSS service");
             }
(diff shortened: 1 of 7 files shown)


bundles/org.openhab.voice.voicerss/src/main/java/org/openhab/voice/voicerss/internal/VoiceRSSTTSService.java
```diff
@@ -121,22 +135,12 @@ public AudioStream synthesize(String text, Voice voice, AudioFormat requestedFor
         if (!voices.contains(voice)) {
             throw new TTSException("The passed voice is unsupported");
         }
-        boolean isAudioFormatSupported = false;
```
abrenk, line 124: This check isn't needed here any longer because the methods getApiAudioCodec() and getApiAudioFormat() called later on throw a TTSException if a format isn't supported.

lolodomo:
Does the service let you choose the output format (I do not remember and I have not checked) ?

WAV files are probably bigger, so will take more place in RAM (for RPI) and will mean probably a bigger delay to start the playback. Does it provide a better audio quality in practice ?

Please fix the conflict. Then I will review.

abrenk:
The URL contains the requested audio format (see [VoiceRSSCloudImpl.java#L259](https://github.com/openhab/openhab-addons/blob/46ba275edbbd66b89a6b9231873cd9dbabb1e1c6/bundles/org.openhab.voice.voicerss/src/main/java/org/openhab/voice/voicerss/internal/cloudapi/VoiceRSSCloudImpl.java#L259)). Supported formats are listed in the [VoiceRSS API documentation](https://www.voicerss.org/api/) in the "Audio Codecs" (MP3,WAV, AAC, OGG, CAF) and "Audio Formats" (8 kHz - 48 kHz, 8 or 16 Bit) sections.

Anything above 8-Bit, 16 kHz is pretty much the same quality, it's just generated speech after all. Below that it sounds muffled.

But it's not a question of quality, rather than what formats the audio sink supports.

In my case I need the audio in WAV, 22 kHz, 16 Bit to stream to my [u::Lux switches](https://www.u-lux.com/en/ulux-switch/overview/) (smart wall switch with integrated speaker).

I don't think that even a Raspberry Pi would have any trouble with the file size difference. Converting the audio format between audio source and sink would add latency and processing time on the other hand. So it's best to let VoiceRSS generate the needed format.

I have rebased to main (fixing the conflict and updating copyright headers in the two new files).

lolodomo:
> But it's not a question of quality, rather than what formats the audio sink supports.

Ok, I understand now the main purpose of this enhancement.

bundles/org.openhab.voice.voicerss/src/main/java/org/openhab/voice/voicerss/internal/cloudapi/VoiceRSSCloudImpl.java
```diff
@@ -41,21 +39,51 @@
...
  */
 public class VoiceRSSCloudImpl implements VoiceRSSCloudAPI {
 
     public static final String DEFAULT_VOICE = "default";
 
     private final Logger logger = LoggerFactory.getLogger(VoiceRSSCloudImpl.class);
 
-    private static final Set<String> SUPPORTED_AUDIO_FORMATS = Stream.of("MP3", "OGG", "AAC").collect(toSet());
+    private static final Set<AudioFormat> SUPPORTED_AUDIO_FORMATS = Set.of(
+            new AudioFormat(AudioFormat.CONTAINER_NONE, AudioFormat.CODEC_MP3, null, 16, null, 44_100L),
```
lolodomo, line 58: Bitrate was previously set to 64000 for MP3. isn't it any problem to set it to null now ?

bundles/org.openhab.voice.voicerss/src/main/java/org/openhab/voice/voicerss/tool/CreateTTSCache.java
```diff
@@ -106,7 +106,7 @@ private void generateCacheForMessage(String apiKey, String cacheDir, String loca
             return;
         }
         CachedVoiceRSSCloudImpl impl = new CachedVoiceRSSCloudImpl(cacheDir);
-        File cachedFile = impl.getTextToSpeechAsFile(apiKey, trimmedMsg, locale, voice, "MP3");
+        File cachedFile = impl.getTextToSpeechAsFile(apiKey, trimmedMsg, locale, voice, "MP3", null);
```
lolodomo, line 109: I am a little lost. Why is MP3 hardcoded here ?
And using null as last new parameter can not lead to problems (NPE) ?

more generally, now, using VoiceRSS, what will be with your changes the default codec used ? Will it be MP3 if the sink supports MP3 ?

lolodomo (changes requested):

bundles/org.openhab.voice.voicerss/src/main/java/org/openhab/voice/voicerss/internal/cloudapi/VoiceRSSCloudImpl.java
abrenk: It was specified, yes, but as it was not sent to the VoiceRSS API it was not guaranteed to be correct. And AudioFormat's interpretation of "null" is "any value", so I think it is better this way.

I don't know if the bitRate field in AudioFormat is used anywhere at all, really. It's not used in openhab-addons and only copied around in openhab-core...

lolodomo: If not used, of course it is ok.

bundles/org.openhab.voice.voicerss/src/main/java/org/openhab/voice/voicerss/tool/CreateTTSCache.java
abrenk: > I am a little lost. Why is MP3 hardcoded here ?

"MP3" was already hardcoded before and not parameterized in the commit to "add support for OGG and AAC audio formats" because it's probably unused. So I also didn't touch it.

This file is meant to be called from the CLI, see [Caching in README.md](https://github.com/openhab/openhab-addons/tree/main/bundles/org.openhab.voice.voicerss#caching). Perhaps it should be deleted as calling the [`openhab:voice say <text>` command](https://www.openhab.org/docs/administration/runtime.html#voice) in the runtime console has the same effect.

> And using null as last new parameter can not lead to problems (NPE) ?

No risk of a NPE: ` if (!Objects.equals(format, "44khz_16bit_mono")) { ... }`

> More generally, using VoiceRSS with your changes, what will be the default codec used ? Will it be MP3 if the sink supports MP3?

Hm. Good question. I think so, if the sink does not support WAV. That's decided in [`VoiceManagerImpl`](https://github.com/openhab/openhab-core/blob/0a778df0adddd90f3db219662c1d7c7369bed14f/bundles/org.openhab.core.voice/src/main/java/org/openhab/core/voice/internal/VoiceManagerImpl.java#L217):

```
Set<AudioFormat> audioFormats = tts.getSupportedFormats();
...
AudioFormat audioFormat = getBestMatch(audioFormats, sink.getSupportedFormats());
```

So I think that's not really an implementation concern of the VoiceRSS binding but the openhab-core classes.

lolodomo: I see that it was previously hardcoded but I did not remember why. As I was probably the one who added OGG/AAC, if I left MP3 hardcoded here, there is certainly a good reason, I hope :) But I will check that again, considering your comments about the CLI command to build the cache.

I would like to be sure that these changes will not lead to users "loosing" their cache in practice because now the binding is now using another codec and other files in the cache.
Currently, I see that my cached files have all the MP3 extension.

Let me the time to analyse more in details the code and to test your changes.

abrenk: I made sure not to change the filename for previous uses of the default 44khz_16bit_mono format:

```
/**
 * Gets a unique filename [...] suffixed by the format if it is not the default of "44khz_16bit_mono".
 */
private String getUniqueFilenameForText(String text, String locale, String voice, String format) {
    ...
    if (!Objects.equals(format, "44khz_16bit_mono")) {
        filename += "_" + format;
    }
    ...
}
```

Thanks for your review so far.

abrenk: A lot of audio sinks support MP3 and WAV: javasound, webaudio, chromecast, kodi, squeezebox, ...

`VoiceManagerImpl#getPreferredFormat` seems to prefer WAV, so I guess you're right that this change might lead to cache misses.

But it would be somewhat unfitting to have special "prefer MP3" code only in this binding. Better to implement something in openhab-core.

There's code duplication between `VoiceManagerImpl#getPreferredFormat` and `AudioFormat#getPreferredFormat`, the latter has a strange `defaultFrequency = 16384` setting, that is not contained in the [list of common sampling frequencies in Wikipedia](https://en.m.wikipedia.org/wiki/Sampling_(signal_processing)) (the correct value would be 16000 Hz). This is corrected in `VoiceManagerImpl` to 44100 Hz. I already wrote that bitRate is practically unused. bigEndian only makes sense for WAV with a bit depth of 16 bit but not for 8 bit WAV or MP3 or other compressed formats. Maybe it's time for a little cleanup and refactoring there.

lolodomo: The duplicated code is something I have detected several years ago, I mentioned it somewhere in an issue but it remained like that as I got no answer.

lolodomo: In case, WAV iis preferred with your changes, it would impact mainly users who built the cache with predefined texts. I don't think this is something very common. In this case, this should be mentioned in the release notes as a breaking change (but I assume this will impact really a small number of users).

lolodomo: Ok this class concerns only the external tool to create cached files. 
It could be certainly improved to support other audio format through a new command line argument, probably in a separate PR. 

What remains important to understand and test is what will become the new default format after your changes.

abrenk: If I change the default audio sink to "webaudio" and call `voice say This is test` than a .wav file is created in the cache folder.

How can I mention this change in the release notes? Can this be part of this pull request?

lolodomo: Not in this PR as the file to be changed in another repo. 

No way to give to MP3 a higher priority so that we keep a full backward compatibility?

lolodomo: I will test with Sonos as renderer.

lolodomo: Maybe we should just add an advanced parameter to let the user force an audio format. 
By default, the binding will support WAV, NP3  OGG, ... If set to MP3 for example, the binding will only consider MP3 format. 
Like that, er have a way to keep backward  compatibility + compatibility with the cache tool. 
WDYT?

abrenk: A config option to opt-in to WAV support could be implemented. The return value of `getSupportedFormats()` could be without WAV in the default case then. But I think this is not something a single binding should do as a special case.

The code that decides what audio format to use is in openhab-core. I can't say if a sink can prefer a format by returning it first in the of list of supported formats. Oh, I see that it's a `Set` and not a list, so I guess returning a `LinkedHashSet`  would be possible but isn't really part of the contract of `AudioSink`.

Another possibility is to mention how to batch convert the cache in the release notes:
```
for i in userdata/voicerss/cache/*.mp3; do sox "$i" "${i%.mp3}.wav"; done
```

abrenk: > By default, the binding will support WAV, MP3 , OGG, ... If set to MP3 for example, the binding will only consider MP3 format.
> Like that, we have a way to keep backward compatibility

Listing the preferred audio formats to use as a config option in openhab-core would be a good idea. But would not solve our problem here, because -core preferred wav and should continue to do so when such an option is introduced, I guess.

So if you do not think a remark in the release notes is enough we should start to implement an option for the voicerss binding and maybe later on do something about it in -core. Perhaps even changing `AudioSink` to return a `List`. A sink should have the possibility to specify a preference.

lolodomo: What I was thinking of is clearly an option in voicerss. My concern is users who created cached files for voicerss using the provided tool. Currently all these files are MP3 files. 

My idea was a binding parameter that would limit the audio formats handled by the binding to a particular audio format, MP3 as an example. 

But now I think the best option is just to enhance the tool to let the user select the output audio format. Like that, to switch to WAV, files, the user will have only to rerurn his command but asking WAV format. 

Can you enhance the tool with one additional command line argument that will be the audio codec to be used?

abrenk: Okay, I will do that!

lolodomo:
@abrenk : any progress ?

lolodomo (approved):
LGTM

lolodomo:
In fact, I am going to merge your PR now.
The cache tool can be updated in a separate PR.

lolodomo:
I just tested and even if a MP3 file is still produced in my case, I see "_44khz_16_mono" added to my files.
So your test is probably not working. I will see if I find why.

lolodomo:
I think I found the problem: you were building an audio format code like `44khz_16_mono `instead of expected `44khz_16bit_mono`

PR #11918, opened 2021-12-31
[shelly] set min value to 1s for updateInterval in config.xml

rose78ch:
https://github.com/openhab/openhab-addons/issues/11892


rose78ch:
I have found the solution in the documentation (only beginners read the instructions :joy: ):
OH and the shellys are in a different subnet. Now I put the 2 shelly (who should faster react) in the same subnet as OH. The state and value change in openHab takes now 1-2 seconds.

PR #11921, opened 2022-01-01
[ipcamera] Fix multiple mjpeg issues and allow stream to stay alive

Skinah:
These changes fix/improve:

- Flickering that some users found when the FIFO buffer drops elements during times of network stress. Reported on forum.
- Fix bug stopping multiple mjpeg streams from working from a RTSP source.
- Improve ipcamera.mjpeg to keep streams alive when cameras go offline or reboot.


lolodomo (approved):
LGTM

PR #11922, opened 2022-01-01
[jdbc] Implement 'ModifiablePersistenceService' interface

cweitkamp:
- Implement `ModifiablePersistenceService` interface for MySQL and MariaDB

See https://github.com/openhab/openhab-core/issues/2618

### DIFF
--- a/bundles/org.openhab.persistence.jdbc/src/main/java/org/openhab/persistence/jdbc/db/JdbcBaseDAO.java
+++ b/bundles/org.openhab.persistence.jdbc/src/main/java/org/openhab/persistence/jdbc/db/JdbcBaseDAO.java
@@ -13,11 +13,11 @@
 package org.openhab.persistence.jdbc.db;
 
 import java.math.BigDecimal;
-import java.sql.Timestamp;
 import java.time.Instant;
 import java.time.ZoneId;
 import java.time.ZonedDateTime;
 import java.time.format.DateTimeFormatter;
+import java.util.Date;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
@@ -327,13 +327,23 @@ public class JdbcBaseDAO {
         Yank.execute(sql, null);
     }
 
-    public void doStoreItemValue(Item item, ItemVO vo) {
-        ItemVO storedVO = storeItemValueProvider(item, vo);
+    public void doStoreItemValue(Item item, State itemState, ItemVO vo) {
+        ItemVO storedVO = storeItemValueProvider(item, itemState, vo);
         String sql = StringUtilsExt.replaceArrayMerge(sqlInsertItemValue,
                 new String[] { "#tableName#", "#tablePrimaryValue#" },
                 new String[] { storedVO.getTableName(), sqlTypes.get("tablePrimaryValue") });
         Object[] params = new Object[] { storedVO.getValue(), storedVO.getValue() };
         logger.debug("JDBC::doStoreItemValue sql={} value='{}'", sql, storedVO.getValue());
+        Yank.execute(sql, params);
+    }
+
+    public void doStoreItemValue(Item item, State itemState, ItemVO vo, Date date) {
+        ItemVO storedVO = storeItemValueProvider(item, itemState, vo);
+        String sql = StringUtilsExt.replaceArrayMerge(sqlInsertItemValue,
+                new String[] { "#tableName#", "#tablePrimaryValue#" }, new String[] { storedVO.getTableName(), "?" });
+        java.sql.Timestamp timestamp = new java.sql.Timestamp(date.getTime());
+        Object[] params = new Object[] { storedVO.getValue(), timestamp, storedVO.getValue() };
+        logger.debug("JDBC::doStoreItemValue sql={} timestamp={} value='{}'", sql, timestamp, storedVO.getValue());
         Yank.execute(sql, params);
     }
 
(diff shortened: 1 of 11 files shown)


bundles/org.openhab.persistence.jdbc/src/main/java/org/openhab/persistence/jdbc/db/JdbcBaseDAO.java
```diff
@@ -398,14 +407,14 @@ private String updateItemTableNamesProvider(List<ItemVO> namesList) {
...
+    protected ItemVO storeItemValueProvider(Item item, State itemState, ItemVO vo) {
         String itemType = getItemType(item);
 
         logger.debug("JDBC::storeItemValueProvider: item '{}' as Type '{}' in '{}' with state '{}'", item.getName(),
-                itemType, vo.getTableName(), item.getState());
+                itemType, vo.getTableName(), itemState);
 
         // insertItemValue
-        logger.debug("JDBC::storeItemValueProvider: getState: '{}'", item.getState());
+        logger.debug("JDBC::storeItemValueProvider: getState: '{}'", itemState);
```
wborn, line 417:
```suggestion
        logger.debug("JDBC::storeItemValueProvider: itemState: '{}'", itemState);
```

bundles/org.openhab.persistence.jdbc/src/main/java/org/openhab/persistence/jdbc/db/JdbcBaseDAO.java
```diff
@@ -465,22 +473,22 @@ protected ItemVO storeItemValueProvider(Item item, ItemVO vo) {
...
+                        ((DateTimeType) itemState).getZonedDateTime().toInstant().toEpochMilli());
                 logger.debug("JDBC::storeItemValueProvider: DateTimeItem: '{}'", d);
                 vo.setValue(d);
                 break;
             default:
                 // All other items should return the best format by default
                 vo.setValueTypes(getSqlTypes().get(itemType), java.lang.String.class);
-                logger.debug("JDBC::storeItemValueProvider: other: item.getState().toString(): '{}'", item.getState());
-                vo.setValue(item.getState().toString());
+                logger.debug("JDBC::storeItemValueProvider: other: item.getState().toString(): '{}'", itemState);
```
wborn, line 490:
```suggestion
                logger.debug("JDBC::storeItemValueProvider: other: itemState.toString(): '{}'", itemState);
```

wborn (commented):
Thanks for implementing the interface!
Can you also update the logging for your changes?

bundles/org.openhab.persistence.jdbc/src/main/java/org/openhab/persistence/jdbc/internal/JdbcPersistenceService.java
```diff
@@ -110,39 +116,46 @@ public void deactivate(final int reason) {
...
+                    "JDBC::store: No connection to database. Cannot persist state '{}' for item '{}'! Will retry connecting to database when error count:{} equals errReconnectThreshold:{}",
+                    state, item, errCnt, conf.getErrReconnectThreshold());
             return;
         }
         long timerStart = System.currentTimeMillis();
-        storeItemValue(item);
-        logger.debug("JDBC: Stored item '{}' as '{}' in SQL database at {} in {} ms.", item.getName(), item.getState(),
-                new java.util.Date(), System.currentTimeMillis() - timerStart);
+        storeItemValue(item, state, date);
+        logger.debug("JDBC: Stored item '{}' as '{}' in SQL database at {} in {} ms.", item.getName(), state,
```
wborn, line 157: It would also make sense to make this non trivial logging conditional with `if (logger.isDebugEnabled()) {`

bundles/org.openhab.persistence.jdbc/src/main/java/org/openhab/persistence/jdbc/internal/JdbcPersistenceService.java
```diff
@@ -231,4 +244,12 @@ public void updateConfig(Map<Object, Object> configuration) {
     public List<PersistenceStrategy> getDefaultStrategies() {
         return List.of(PersistenceStrategy.Globals.CHANGE);
     }
+
+    @Override
+    public boolean remove(FilterCriteria filter) throws IllegalArgumentException {
+        if (filter.getItemName() == null) {
+            throw new IllegalArgumentException("Item name must not be null");
+        }
+        throw new UnsupportedOperationException("Not implemented yet");
```
wborn, line 253: Is it a lot of work to implement this? If not implemented I'd always throw the `UnsupportedOperationException` so users will not waste time on adding a missing item name and then eventually run into `UnsupportedOperationException`. :wink:

cweitkamp: I am not sure. will look into it.

cweitkamp: @wborn I implemented the missing piece of code and added basic unit tests for the generation of queries.

wborn:
Thanks for also implementing the delete part @cweitkamp!

I did notice that the `FilterCriteria` also allows for [filtering on state](https://github.com/openhab/openhab-core/blob/d6933021901ca28a4122f7057be3fbd4ef3d873d/bundles/org.openhab.core.persistence/src/main/java/org/openhab/core/persistence/FilterCriteria.java#L88-L89) with an operator which does not seem to be used at all by the JDBC persistence service? 

It might cause more state to be removed when this filter is used for removing garbage state like null or extremely low/high values. If that is hard to implement you could also throw an exception when it is specified in the `FilterCriteria` parameter.

cweitkamp:
Yes, I agree. It is a nice feature and I think it is not too hard to implement either.

But as long as the REST API is the only consumer of a `ModifiablePersistenceService` and the delete method does not support parameters like state or operator I do not see the "real" need to implement it right away. I guess we found another future improvement to be implemented. 😉

wborn:
Yes I now see it's not possible to create such a filter using the API so everyones persisted state should be safe! :slightly_smiling_face:

wborn (approved):
Thank you very much! :+1:

PR #11923, opened 2022-01-01
[WIP][mapdb] Implement 'ModifiablePersistenceService' interface

cweitkamp:
- Implement `ModifiablePersistenceService` interface

Does it really make sense to implement this interface for MapDB as it only can store the last value?

See openhab/openhab-core#2618

### DIFF
--- a/bundles/org.openhab.persistence.mapdb/src/main/java/org/openhab/persistence/mapdb/internal/MapDbPersistenceService.java
+++ b/bundles/org.openhab.persistence.mapdb/src/main/java/org/openhab/persistence/mapdb/internal/MapDbPersistenceService.java
@@ -37,6 +37,7 @@ import org.openhab.core.common.ThreadPoolManager;
 import org.openhab.core.items.Item;
 import org.openhab.core.persistence.FilterCriteria;
 import org.openhab.core.persistence.HistoricItem;
+import org.openhab.core.persistence.ModifiablePersistenceService;
 import org.openhab.core.persistence.PersistenceItemInfo;
 import org.openhab.core.persistence.PersistenceService;
 import org.openhab.core.persistence.QueryablePersistenceService;
@@ -54,14 +55,14 @@ import com.google.gson.GsonBuilder;
 
 /**
  * This is the implementation of the MapDB {@link PersistenceService}. To learn more about MapDB please visit their
- * <a href="http://www.mapdb.org/">website</a>.
+ * <a href="https://www.mapdb.org/">website</a>.
  *
  * @author Jens Viebig - Initial contribution
  * @author Martin Kühl - Port to 3.x
  */
 @NonNullByDefault
 @Component(service = { PersistenceService.class, QueryablePersistenceService.class })
-public class MapDbPersistenceService implements QueryablePersistenceService {
+public class MapDbPersistenceService implements ModifiablePersistenceService {
 
     private static final String SERVICE_ID = "mapdb";
     private static final String SERVICE_LABEL = "MapDB";
@@ -169,7 +170,16 @@ public class MapDbPersistenceService implements QueryablePersistenceService {
 
     @Override
     public void store(Item item, @Nullable String alias) {
-        if (item.getState() instanceof UnDefType) {
+        internalStore(item, alias, new Date(), item.getState());
+    }
+
+    @Override
+    public void store(Item item, Date date, State state) {
+        internalStore(item, null, date, state);
+    }
+
+    private void internalStore(Item item, @Nullable String alias, Date date, State state) {
+        if (state instanceof UnDefType) {
             return;
         }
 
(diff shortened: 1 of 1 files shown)


TBail:
Hmm good question.

The difference between mapdb and other persistence solutions is from my poit of view the fact that mapdb i only able to store one value. So it may be a good idea to have the function to save this one value on purpose with a defined timestamp.

But to be honest i have no idea how to use the feature with mapdb. To realise my approach i have to wait for the influxdb update.

spacemanspiff2007:
The only use case I can think of is that mapdb does **not** persist on every change but the user has a script in place that only persists some states but not all of them.

On startup mapdb is then used to restore the state.

kaikreuzer:
I'd say that this is rather counter-intuitive. While for a "real" modifiable persistence service this is meant to update a historical value, for mapDB this now replaces the current (possibly more recent) value by a different one. This might not be the intended behavior.

As mapDB does not store any time series, I would say that there is no sense in offering methods to modify the stored time series.

My vote would thus go against this PR.

cweitkamp:
This strengthen my initial feeling.

This PR shows that the implementation for MapDB is easy to archive. But implementing the `ModifiablePersistenceService` does not fit into the idea behind this persistence service. I will close this PR but leave the code behind for history.

spacemanspiff2007:
@kaikreuzer 
How would I then set a value in mapdb through the rest api?

The user has to directly write "mapdb" in the api field so it should be clear that the value will be overwritten since it was a conscious decision for mapdb.
At least when the value is read back to see if it was successful only one value will be returned which is the new one.
So overwriting is nothing that happens implicitly or unexpected.

I'd hate to not have this implemented because mapdb is an easy way to quickly test the persistence api calls.
If that call already fails it's most likely implementation is buggy and the user will give up (as I did).

kaikreuzer:
> How would I then set a value in mapdb through the rest api?

Hm, well: Sending a state update for the item, so that it gets persisted the usual way.

spacemanspiff2007:
So if I want to have a defined state from mapdb on startup I have to:
- configure on-change persistence for the item
- post an update to the item
- disable on-change persistence for the item
- post the original state to the item

That's quite a workaround if I want to restore a value that the item currently doesn't have.

---

I think it's much better to have consistent api behavior than to decide for the user which api call makes sense and which don't.
It's not clear why for some persistence service the endpoint would work and for some it wouldn't. This just creates confusion.

@kaikreuzer: Could you please rethink about about it from this angle and factor consistency into your vote?

Rossko57:
I can't see why this is any different to allowing `myItem.persist("mapdb")` from a rule.  Say I want to persist the result of some calculation or condition on my terms, not a predefined schedule or on every change.

spacemanspiff2007:
Correct, it's the same. Does it work from a rule?

Rossko57:
> Does it work from a rule?

Has done since OH1 I believe.  I assume persistence extensions to Items do not examine this poperty.

spacemanspiff2007:
Then wouldn't it make sense if it would work through the REST API, too?

kaikreuzer:
@spacemanspiff2007 The discussion here was about whether it makes sense to make MapDB a ModifiablePersistenceService and as this service does not store any history, it simply does not make sense to have it implement an interface that is meant for allowing changing its historic data.

If I get you right, all you wish is to be able to trigger a "persist" of the current item state through the REST API, right?
If so, I'd say this should be an issue in openhab-core. Imho, the operation "storeItemDataInPersistenceService" in the PersistenceResource could be adapted to work if the user does not provide a time with the request (and in this case simply work with ANY persistence service and not require a modifiable one). The swagger docs even claim that the time "Will default to current time." if not provided, but then again, it marks this parameter to be required. I'd think it should be an easy fix.

spacemanspiff2007:
Thank you @kaikreuzer for clarifying and explaining it to me. 
I really misunderstood - I thought this was about setting a value through the rest api since it was linked from the rest api issue  (see post 1).

> If I get you right, all you wish is to be able to trigger a "persist" of the current item state through the REST API, right?

Almost. I want to trigger a persist and I want to supply the item state that will be persisted, too.
Example call:
```
/rest/persistence/items/MY_ITEM?serviceId=map_db&state=MY_ITEM_STATE
```
I can gladly supply a timestamp, this is no dealbraker for me :wink: 



> I'd think it should be an easy fix.

Shall I create a new issue?

Rossko57:
> I want to trigger a persist and I want to supply the item state that will be persisted

Ah, that is quite different from the rules extension `myItem.persist` then.  We might call this "spoofing" a state.

kaikreuzer:
> We might call this "spoofing" a state.

That's an interesting comment. The question is probably, whether we actually set the item state to what we persist, so that it reflects the reality or if we indeed allow persisting something that has nothing to do with the current state...?

spacemanspiff2007:
> or if we indeed allow persisting something that has nothing to do with the current state...?

That is how I understand it works for the ``ModifiablePersistenceService``.
E.g.:
- Use rrd4j and have no persistence rule set up except ``restore on startup``
- Persist a value custom value with a timestamp e.g. 1 sec in the past. 
- If I restart openhab the item will now have the custom startup value

Persistence operations are not bound to item states and thus should not change it.
Otherwise one would use the item operations.

kaikreuzer:
> Persistence operations are not bound to item states

Well, we were talking about an equivalent to `myItem.persist("mapdb")` in rules and this is very much bound to item states.
And yes, persistence is only meant to persist item states.
The modifiable persistence is just an exception to the extend that it cannot check what was the correct item state in the past and thus has to trust the caller. Main reason was to allow cleaning history for outliers or invalid data.

spacemanspiff2007:
> I understood that the problem to solve here is that the persistence configuration might not pick this up and thus it is needed to trigger persistence manually through the REST call.

Shall I create a new issue to track a possible implementation?
How do you suggest we proceed?

PR #11926, opened 2022-01-01
[mihome] Update README.md

RadOle:
[mihome] Updated documentation for rules examples in area of receivedEvent implicit variable usage.


wborn (approved):
Thanks! It was indeed a breaking change in OH3, see https://github.com/openhab/openhab-core/pull/1770#issuecomment-734795276.

PR #11618, opened 2021-11-21
[feed] Changed pattern for DateTime channel types to be considered by i18n-tool

cweitkamp:
- Changed pattern for DateTime channel types to be considered by i18n-tool

Signed-off-by: Christoph Weitkamp <github@christophweitkamp.de>

PR #11928, opened 2022-01-02
[icalendar] fixes #11927: Added original texts for translation

### LINKED ISSUE #11927: [icalendar] Not all translations are in original language available
As mentioned in #11890, at least one translation has no original string, making updates via crowdin also removing the translation.

This report is only for the record, the PR should contain a issue-number. I'll fix it myself.

daMihe:
Just added originals so they can be edited via crowdin.


cweitkamp:
Thank you for taking care. Would be nice if you can add a pattern for channel types which can be linked to a `DateTime` item type - like I did in https://github.com/openhab/openhab-addons/pull/11618 for the feed binding. This way we can provide localized formats for date/time values too.

And do not forget to run `mvn i18n:generate-default-translations` afterwards.

daMihe:
Thank you for the hint. Built the addon and also rebuilt translation source.

bundles/org.openhab.binding.icalendar/src/main/resources/OH-INF/thing/thing-types.xml
```diff
@@ -156,6 +156,7 @@
 
 			<parameter name="maxEvents" type="integer" required="true" min="0" groupName="general">
 				<label>Maximum Matches</label>
+				<description>Maximum Matches found by this filter</description>
```
lolodomo, line 159: matches with a lowercase please.

bundles/org.openhab.binding.icalendar/src/main/resources/OH-INF/thing/thing-types.xml
```diff
@@ -171,7 +172,8 @@
 					<option value="WEEK">week</option>
 				</options>
 				<default>HOUR</default>
-				<label>Date or Time Unit for Start and End</label>
+				<label>Date/Time Unit</label>
+				<description>Date or Time Unit for Start and End</description>
```
lolodomo, line 176: For description, use uppercase only for the first letter of the first word, not for all words.

lolodomo (changes requested):

bundles/org.openhab.binding.icalendar/src/main/resources/OH-INF/thing/thing-types.xml
daMihe: OK.

bundles/org.openhab.binding.icalendar/src/main/resources/OH-INF/thing/thing-types.xml
daMihe: OK. Is there an english language based reason when to use capitalization or are labels nouns just capitalized by convention?

cweitkamp (approved):

Issue #11930, opened 2022-01-02
[wolfsmartset] NullPointerException updateing values for Wolf Smartset Pro


BoBiene:
Issue reported in the community forum: 
https://community.openhab.org/t/oh-3-2-wolf-smartset-binding-null-pointer-exception/130486

## Expected Behavior
Values should update properly

## Current Behavior
Due to the NullPointerException the Unit-Thing value update failes

PR #11931, opened 2022-01-02
[wolfsmartset] Fix NullPointerException when updating values

BoBiene:
Signed-off-by: Bo Biene <openhab.github@biene.eu>
Fix for #11930

### DIFF
--- a/bundles/org.openhab.binding.wolfsmartset/src/main/java/org/openhab/binding/wolfsmartset/internal/dto/GetParameterValuesDTO.java
+++ b/bundles/org.openhab.binding.wolfsmartset/src/main/java/org/openhab/binding/wolfsmartset/internal/dto/GetParameterValuesDTO.java
@@ -16,6 +16,8 @@ import java.util.List;
 
 import javax.annotation.Generated;
 
+import org.eclipse.jdt.annotation.NonNullByDefault;
+
 import com.google.gson.annotations.Expose;
 import com.google.gson.annotations.SerializedName;
 
@@ -53,8 +55,8 @@ public class GetParameterValuesDTO {
         this.values = values;
     }
 
-    public Boolean getIsNewJobCreated() {
-        return isNewJobCreated;
+    public @NonNullByDefault Boolean getIsNewJobCreated() {
+        return isNewJobCreated != null ? isNewJobCreated : false;
     }
 
     public void setIsNewJobCreated(Boolean isNewJobCreated) {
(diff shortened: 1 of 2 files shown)


bundles/org.openhab.binding.wolfsmartset/src/main/java/org/openhab/binding/wolfsmartset/internal/dto/GetParameterValuesDTO.java
```diff
@@ -53,8 +55,8 @@ public void setValues(List<ValueDTO> values) {
         this.values = values;
     }
 
-    public Boolean getIsNewJobCreated() {
-        return isNewJobCreated;
+    public @NonNullByDefault Boolean getIsNewJobCreated() {
```
lolodomo, line 58: A non nullable Boolean is for me simply a boolean. 
Why not returning type boolean?

lolodomo (changes requested):

bundles/org.openhab.binding.wolfsmartset/src/main/java/org/openhab/binding/wolfsmartset/internal/dto/GetParameterValuesDTO.java
BoBiene: I'm not familiar with java ;) My knowledge is based on .NET / C# and there a bool / boolean / Boolean is the same and is never null. The issue here is, that the ``isNewJobCreated`` is null, so i thought Boolean can get null, and i added the Nullcheck here to prevent the NullPointerException. 

So is Type boolean not the same as Type Boolean in Java?

lolodomo: No boolean is a primitive type can be true or false. 
Boolean is an object that can be null or having a value true or false.

bundles/org.openhab.binding.wolfsmartset/src/main/java/org/openhab/binding/wolfsmartset/internal/handler/WolfSmartsetUnitThingHandler.java
```diff
@@ -185,7 +185,7 @@ public void updateConfiguration(SubMenuEntryDTO submenu, MenuItemTabViewDTO tabm
     public void updateValues(@Nullable GetParameterValuesDTO values) {
         var thingId = thing.getUID();
         if (values != null && values.getValues() != null && values.getValues().size() > 0) {
-            if (!values.getIsNewJobCreated()) {
+            if (Boolean.FALSE.equals(values.getIsNewJobCreated())) {
```
lolodomo, line 188: Now that the method returns a boolean, you have to test like this:

if `(!values.getIsNewJobCreated())`

I am not sure if the correct fix would not have been to check null here rather than in the called method. I let you think about that and decide.

lolodomo (changes requested):

lolodomo (approved):
LGTM

PR #11206, opened 2021-09-04
[hdpowerview] Add new battery level channel

jlaur:
Fixes #11205

Signed-off-by: Jacob Laursen <jacob-github@vindvejr.dk>

This adds a new channel **batteryLevel** of standard type _system.battery-level_ extracting as much data as possible from the Hunter Douglas PowerView Hub. This is quite limited as only three different values are available: low, medium or high. Previously it was only possible through channel **lowBattery** to deduct if level was either low or medium/high. In the future this channel could be improved by calculations taking voltage into consideration also.

Also fixes a regression from #11201 where "no status available" when **lowBattery** was ON would lead to **lowBattery** being reset to OFF.

PR #11201, opened 2021-09-03
[hdpowerview] Fix low battery when no battery status available

jlaur:
Fixes #11200

Signed-off-by: Jacob Laursen <jacob-github@vindvejr.dk>

**BatteryStatus**
0 = No Status Available, 1 = Low, 2 = Medium, 3 = High, 4 = Plugged In
Type : enum (0, 1, 2, 3, 4)

PR #11155, opened 2021-08-24
[hdpowerview] Add new batteryVoltage channel

jlaur:
This draft PR adds a new advanced channel for reading battery voltage reported by shade.

Also fixes documentation of the **lowBattery** channel which was documented as **batteryLow**.

PR #11933, opened 2022-01-02
[hdpowerview] Refresh battery level when receiving refresh command

### LINKED ISSUE #11932: [hdpowerview] Refresh battery status also when receiving refresh command
Hunter Douglas PowerView motorized shades are **notoriously unreliable** when reporting battery status (level/voltage). When battery status is hard-refreshed it will often result in wrong value or no value at all. For example, very low voltages like 2.5 V might be reported when voltage is actually above 15 V. Worst example:
image

Currently sending the REFRESH command to a shade will only result in hard refresh of the position. Battery status should also be included. This would make it possible to create rule logic for retrying refresh to get a decent value when value is either missing or voltage has dropped more than a certain threshold value.

Logged when not able to get battery status:
```
2022-01-02 14:40:55.303 [WARN ] [rnal.handler.HDPowerViewShadeHandler] - Shade 36321 wireless refresh time out
```

jlaur:
Fixes #11932

As natural evolution after:
- #11155
- #11201 
- #11206
- #11260

we should hard-refresh battery level also when receiving REFRESH command.

### DIFF
--- a/bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/handler/HDPowerViewShadeHandler.java
+++ b/bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/handler/HDPowerViewShadeHandler.java
@@ -104,8 +104,9 @@ public class HDPowerViewShadeHandler extends AbstractHubbedThingHandler {
 
     @Override
     public void handleCommand(ChannelUID channelUID, Command command) {
-        if (RefreshType.REFRESH.equals(command)) {
+        if (RefreshType.REFRESH == command) {
             requestRefreshShadePosition();
+            requestRefreshShadeBatteryLevel();
             return;
         }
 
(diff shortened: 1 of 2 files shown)


bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/handler/HDPowerViewShadeHandler.java
```diff
@@ -104,8 +104,9 @@ public void initialize() {
 
     @Override
     public void handleCommand(ChannelUID channelUID, Command command) {
-        if (RefreshType.REFRESH.equals(command)) {
+        if (RefreshType.REFRESH == command) {
             requestRefreshShadePosition();
+            requestRefreshShadeBatteryLevel();
```
lolodomo, line 109: The REFRESH command should be at channel level, this is not a global command for all channels. 
Are all of the channels depending on each other?

lolodomo (changes requested):

lolodomo (approved):
LGTM

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/handler/HDPowerViewShadeHandler.java
jlaur: You are absolutely right - nice catch! I have updated the logic so that channel is taken into consideration. Documentation has been updated accordingly to clarify how it works.

PR #11936, opened 2022-01-02
[dali] Fix frame packing

rs22:
Signed-off-by: Robert Schmid <r.schmid@outlook.com>

This fixes an out-of-range bug that prevented DALI frames from being created.
(The counter variable that helps in copying bytes from a list into an array in reverse order should be 1 in the first loop iteration.)

This bug was introduced when I incorporated change requests as part of my original PR for the binding, and I somehow missed to test this afterwards.
With this fix, the binding works as expected. 😅 

Happy new year!


lolodomo (approved):
LGTM

PR #11937, opened 2022-01-02
[LuxtronikHeatpump] fix channel type of newly defined unknown channels

sgiehl:
Fixes an issue introduced with https://github.com/openhab/openhab-addons/pull/11860

The new "unknown" channels are all using the same channel type. When a new thing is created the channels are rebuilt based on the available channels. Before the channel id and the channel type were always the same. This fails for the new channels. 
This PR should fix this.

See https://github.com/openhab/openhab-addons/pull/11860#issuecomment-1003620849


lolodomo (approved):
LGTM

tmp-hallenser:
@sgiehl Is this fix already included in the latest docker snapshot image? (still remaining in 'unknown' state :( )

sgiehl:
I'm not sure how often that is built. But you will need to remove and readd the thing.

tmp-hallenser:
Docker image has been built like 2hrs ago; removed thing and added it again - still not working. Let's see if it's fixed in one of the next builds

sgiehl:
I actually don't know what the docker image contains and wether this also includes the latest checkout of the add-ons. Maybe @lolodomo can answer that.

lolodomo:
I believe it should be included in docker version 2472 built about 8 hours ago.

sgiehl:
@lolodomo my problem actually is, that my heatpump doesn't serve those additional channels. So I'm not able to test that properly. I modified the code locally so the channels are added even if they aren't available and compiled the binding. Adding that to the latest docker snapshot works for me.

tmp-hallenser:
I'm now seeing the new channels - seem like it wasn't immediately included into the docker image.

PR #7969, opened 2020-06-21
[amazonechocontrol] add support for smarthome devices

J-N-K:
This is a follow-up to the abandoned PR #6140 and includes mainly the changes from the PR and some improvements. All feature credits go to @lkn94 and @mgeramb, I merely fixed some codestyle issues and resolved merge conflicts. 

Also-By: Lukas Knoeller <lukask@hobbyblogging.de>
Also-By: Michael Geramb <mail@michael-geramb.at>

Signed-Off-By: Jan N. Klug <jan.n.klug@rub.de>

PR #6140, opened 2019-09-27
[amazonechocontrol] Add features to control smart devices connected to the Amazon Echo

lkn94:
Add smart home functions to the amazonechocontrol binding. Users are able to control smart devices connected to Amazon Alexa:

- turn on/off your lights
- change the color
- control groups of lights or just single bulbs
- receive the current state of the lights
- turn on/off smart plugs

A first beta version of these functions can be found in the [openhab community](https://community.openhab.org/t/preview-and-beta-amazon-echo-control/81721).

PR #11943, opened 2022-01-02
[amazonechocontrol] fix refresh time for skill connected devices

J-N-K:
It was recently discovered that during the refactoring from #6140 to #7969 I made a mistake that could lead to very frequent requests and high load on smarthome skills (ioBroker and openHAB). Since at least AWS hosted skills are paid per request this has the potential to cause high bills for the skill authors. I'm very sorry for that.


wborn (approved):
Thanks for the fix!

PR #11945, opened 2022-01-02
[jsscripting] Append the ruleID or file name to the logger when console logging

digitaldan:
See https://github.com/openhab/openhab-js/issues/50

### DIFF
--- a/bundles/org.openhab.automation.jsscripting/src/main/resources/node_modules/@jsscripting-globals.js
+++ b/bundles/org.openhab.automation.jsscripting/src/main/resources/node_modules/@jsscripting-globals.js
@@ -2,13 +2,18 @@
 (function (global) {
     'use strict';
 
-    const scriptId = globalThis["ruleUID"] !== undefined ? ".ui." + globalThis["ruleUID"] : globalThis["javax.script.filename"] !== undefined ? ".file." + globalThis["javax.script.filename"].replace(/^.*[\\\/]/, '') : "";
+    //Append the script file name OR rule UID depending on which is available  
+    const defaultLoggerName = "org.openhab.automation.script" + (globalThis["javax.script.filename"] ? ".file." + globalThis["javax.script.filename"].replace(/^.*[\\\/]/, '') : globalThis["ruleUID"] ? ".ui." + globalThis["ruleUID"] : "");
     const System = Java.type('java.lang.System');
-    const log = Java.type("org.slf4j.LoggerFactory").getLogger("org.openhab.automation.script" + scriptId);
     const ScriptExecution = Java.type('org.openhab.core.model.script.actions.ScriptExecution');
     const ZonedDateTime = Java.type('java.time.ZonedDateTime');
-
     const formatRegExp = /%[sdj%]/g;
+
+    function createLogger(name = defaultLoggerName) {
+        return Java.type("org.slf4j.LoggerFactory").getLogger(name);
+    }
+    //user configurable
+    let log = createLogger();
 
     function stringify(value) {
         try {
@@ -141,6 +146,7 @@
                 timers[label] = System.currentTimeMillis();
             }
         },
+
         timeEnd: function (label) {
             if (label) {
                 const now = System.currentTimeMillis();
@@ -151,6 +157,16 @@
                     log.info(format.apply(null, [label + ':', '<no timer>']));
                 }
             }
+        },
+
+        //Allow user customizable logging names
+        set loggerName(name) {
+            log = createLogger(name);
+            this._loggerName = name;
+        },
+
+        get loggerName() {
+            return this._loggerName || defaultLoggerName;
         }
     };
 
@@ -201,5 +217,4 @@
     //Support legacy NodeJS libraries 
     globalThis.global = globalThis;
     globalThis.process = { env: { NODE_ENV: '' } };
-
 })(this);
--- a/bundles/org.openhab.automation.jsscripting/src/main/resources/node_modules/@jsscripting-globals.js
+++ b/bundles/org.openhab.automation.jsscripting/src/main/resources/node_modules/@jsscripting-globals.js
@@ -12,6 +12,7 @@
     function createLogger(name = defaultLoggerName) {
         return Java.type("org.slf4j.LoggerFactory").getLogger(name);
     }
+
     //user configurable
     let log = createLogger();
 
--- a/bundles/org.openhab.automation.jsscripting/src/main/resources/node_modules/@jsscripting-globals.js
+++ b/bundles/org.openhab.automation.jsscripting/src/main/resources/node_modules/@jsscripting-globals.js
@@ -2,8 +2,9 @@
 (function (global) {
     'use strict';
 
+    const scriptId = globalThis["ruleUID"] !== undefined ? ".ui." + globalThis["ruleUID"] : globalThis["javax.script.filename"] !== undefined ? ".file." + globalThis["javax.script.filename"].replace(/^.*[\\\/]/, '') : "";
     const System = Java.type('java.lang.System');
-    const log = Java.type("org.slf4j.LoggerFactory").getLogger("org.openhab.automation.script");
+    const log = Java.type("org.slf4j.LoggerFactory").getLogger("org.openhab.automation.script" + scriptId);
     const ScriptExecution = Java.type('org.openhab.core.model.script.actions.ScriptExecution');
     const ZonedDateTime = Java.type('java.time.ZonedDateTime');
 


Redsandro:
Is it possible to figure out if an UI script comes from a "rule" or a "script"?

I know a UI "script" is just a "rule" without conditions, but the helpfulness of the logs will double if it differentiates for us which settings page to pick here:

image

i.e. `.script.ui.dae9651cb4` vs `.rule.ui.ae8561cb4d` so we don't have to do a coin flip.

Confectrician:
From the corresponding opanhab-js issue discussion written by @digitaldan :

> but i would also like to encourage using console logging over this as well.

Maybe we should also write down this encouragement in the corresponding documentation, then.

rkoshak:
I'm a little late to the party but would it make sense to use the rule name instead of the ruleUID? Both file based rules and UI rules have a name defined by the user that is also meaningful to the user. The name can be acquired from the RuleManager.

    const { ruleRegistry } = require('@runtime/RuleSupport');
    const ruleName = ruleRegistry.get(ruleUID).getName();

If I understand correctly this should add much if any extra load on logging but it allows the logger to behave a little more consistently between file and UI based rules, as well as make it easier to determine which rule it came from since it's the name that is front an center, not the UID.

I was going to ask over there but the last comment indicated to continue the discussion over here. Why should we use console over the OH logger? Can we define our own namespaces (even if it's just below default the main console namespace). I worry about making it impossible for users to turn up and down the logging levels of their rules on a rule by rule basis, or groups of rules in one go, if they use console instead of `log`.

> but i would also like to encourage using console logging over this as well.

Another good argument for using the rule name instead of the UID. Though what makes a Script a Script is just the tag. As with getting the name, once you have the Rule Object from the registry you can also get the tags. So this is possible too.

Redsandro:
> would it make sense to use the rule name instead of the ruleUID?

I think the most helpful thing the logger could do is the originally described behavior of:

1. Distinguish between GUI locations i.e. `rule` or `script`
2. Least amount of characters necessary to find the script in a search box (ctrl+f)

So, if there are many rules, among which the following:

* `Turn garden lighting on`
* `Turn garden lighting on for 1 hour`
* `Turn garden lighting off`
* `Turn garden lighting off in 1 hour`

Then the least amount of search terms you'd need to enter to uniquely find a single rule is probably `garden lighting on`. Now if you look only at the IDs, it would be something like this:

* `a0e33b0d61`
* `9005f6d6ce`
* `de41826d85`
* `d2f3aeeec6`

Then it would be enough to enter two or three characters in the search box to find the rule you are looking for.

Also, a user defined name can be a whole sentence, polluting the namespace.

J-N-K:
@rkoshak Are you sure the name is always unique? At least currently you can add two UI defined rules with the same name. The UID is unique, that is enforced by the framework.

rkoshak:
We are talking about logger names. Does it matter if they are not guaranteed to be unique? It would certainly be unusual if someone were to use the same name for multiple rules. They certainly would be making things difficult on themselves to manage their rules by doing so.

But it seems a bit harsh to not use something chosen by the user here, something they have control over and is human friendly, in the off chance that someone decided to name two rules the same. 

Again, since it's logging it won't break anything if two rules use the same logger. Right now *all* console logging goes to the same logger already.

However, what I don't know is whether the name is required in file based rules. It's been too long since I've done so.

J-N-K:
Sure, technically duplicate names for loggers are not an issue. But they should identify the rule as good as possible. And that's not the case if you omit the rule UID.

One reason for duplicate names: you group your rules in files, one file per room. A rule that controls the daily up/down of the rollershutters would be named "Rollershutter" and that name would be the same for each room.

Of course, you could always argue "the user should not do that" and that he's responsible for a good naming scheme. But reality proves that this will not happen. There will be duplicate names and user will struggle to find out who issued the log message.

rkoshak:
It still feels like putting a burden on users to have to map between rule UIDs and log statements. Especially if we will be encouraging users to use a logger where they have no ability to control the logging namespaces at all. I could see using both the rule name and UID but then the logger names start to get really long.

It's already frustrating to have to explain over and over again that "somename-3" that gets appended to an error caused by a rule being loaded from a file means that error is coming from the third rule defined in the file "somename.rules". Now we will have to explain this over and over for *all* logging, not just parsing errors. 

Then you have the problem that the user doesn't get to define the full ruleUID in file based rules (unless something changed from when I did Jython rules). It appears it uses the name (replacing spaces with underscores) and a really long randomly generated string:

```
Test-rule-from-rule-2c58167f-cd3f-422d-91f1-9866f7542a31
Test-rule-from-rule2-3f81e9f7-f7cc-4d02-98e2-ad68b9c7fcf5
```

The first is the UID from a rule created using builder. The second is from JSRule. 

So you *still* have your problem where you've named your rules in different files "Rollershutter". There is nothing there to tell you which one. You'll have to look it up on the console or in MainUI. I can say from experience way back in OH 2 that this is a real pain.

Oh, and that random string will change the next time that file is loaded. You can't even set up a mental map or chart to help you figure out which one it is. You'll have to relook it up every time you change the file. 

And as you can see, the UID is really long. 

So I argue that:

- Using the rule UID for file based rules doesn't actually address your scenario anyway, not without requiring a lookup.

- The rule UID is awfully long to be used in a logger namespace, with the current default logging config it will get truncated to the last 36 characters. So all you'll actually see in your logs is the randomly changing part and will lose the rule name part anyway. 

- Because the rule UID is randomly generated and changes every time, you will lose the ability to adjust the logging level (for example) for a specific rule. The next time the rule is loaded it will have a new UID and therefore a new logger name.

Unless I'm missing something, using the rule UID seems like a really bad idea to me. It will make logging much more difficult to read, adjust, and consume for file based users. In the UI the users have the option to set their own rule UID when they create the rule, but a distressingly few actually do so.

Redsandro:
When you use the default process conveyed by the OH3 UI, you'll end up with 10 character Rule ID's. I think it makes sense to focus on the implied workflow initially as this PR does.

If I understand correctly, you will get different and more lengthy identifiers when you deviate from this workflow, so it makes sense to allow users to manually deviate from the logging namespace for example by introducing a method not unlike the one [described here](https://github.com/openhab/openhab-js/issues/50#issuecomment-1003758766): `console.namespace('some.template')`

rkoshak:
>  it makes sense to allow users to manually deviate from the logging namespace

If we give the users the ability to define their own namespaces I don't think I have too much heart burn over this at all really. That provides a good deal of relief. If the default doesn't work, you can override it.

> When you use the default process conveyed by the OH3 UI
Even better, in the UI you can override the randomly generated UID and use something meaningful. I'm actually not concerned about using the UID for UI based rules.

You don't have that in files though and it's still the case that the UID changes from one loading of the file to the next which is a problem (unless things have changed in this regard and I missed that change). My concern is if we go down the path of UID in the logger namespace for file based users now, it will result in either a bifurcated logging behavior in the long run or only partially functional logging behavior for file based users.

rkoshak:
On relooking above I think there were a couple of topics I didn't address:

> Least amount of characters necessary to find the script in a search box (ctrl+f)

Why should we require search over browsing? Since the UID is autogenerated when using text based files, why *require* those users to search in MainUI to correlate their log statements with a rule? And it will be required every time since the UID changes every time the rule is loaded. 

> Least amount of characters necessary to find the script in a search box (ctrl+f)

This is a case where I would say it serves the them right for using the Name field to put in a description for which there is already defined a separate field. And since the name is appended to the UID this is still a problem if we do use the ruleUID.

Redsandro:
> Why should we require search over browsing?

Because ~

1. ~ distinct namespaces are an improvement over the current indiscernible log namespace
2. ~ it's cheap (PR exists, click to merge)
3. ~ in reality, as rules become plentiful, you'll end up [searching](#issuecomment-1009321689) anyway
4. ~ it's a _namespace_, not a _name_. The namespace is an identifier meant to offer uniqueness, it is not meant to contain the story all about how the rule's life got flipped-turned upside down. Consider it like a barcode, and until now, all rules have a unique barcode yet all logs have the same barcode.

> Since the UID is autogenerated when using text based files

This problem does not exist in the PR because it uses the filename in stead, correct?

> If we give the users the ability to define their own namespaces I don't think I have too much heart burn over this at all really.

Same. It's not too big of a deal for me. If I don't like it, I will use the old logger. 

However, I speak up because as a brand new OH3 user, I think the UX can be improved by sticking to one best practice workflow and have every tool in service of that workflow _by default_, while allowing seasoned users to deviate from that. Because when you think about it, I'm probably the most novice noob to use JSScripting and no one brought it up before I did.

So my advise: let's assume 10 char IDs. The namespaces stay short and readable. Filename as alternative seems like a good compromise. Tools (in this case the to-be-recommended console logger) should not be diluted in their usefulness to service alternate workflows in my opinion. Don't use namespaces for user sentences.

It's possible that while this issue is about **how to make the rule _namespace_ unique,** you are thinking about **how to make the rule _name_ more visible in logs.** I think in the latter case, what you really want is **control over the logging format.**

In stead of: `$date [$level] [$namespace] - $message`
You want: `$date [$level] [$name] - $message`

That's a different issue.

I don't know what the process is for OH development, who decides what to merge, if there is a vote, if stating my opinion and responding to someone's concerns is helpful enough, but consider this my last input on the matter. Arguing my point is becoming too expensive for me to care.

rkoshak:
> ~ distinct namespaces are an improvement over the current indiscernible log namespace

I'm not arguing to do nothing.

> ~ it's cheaper (PR exists, click to merge)

Cheaper for devs. Way more expensive to end users and those who help them. And any changes made in any direction is just one or two lines of code. 

> ~ in reality, as rules become plentiful, you'll end up searching anyway

So for those who have a few rules we will, say "too bad, gotta search." Very user unfriendly.

> ~ it's a namespace, not a name. The namespace is an identifier meant to offer uniqueness, it is not meant to contain the story all about how the rule's life got flipped-turned upside down. Consider it like a barcode, and until now, all rules have a unique barcode yet all logs have the same barcode.

That completely ignores how the logs are actually used by openHAB users. Frankly, it's counter to how almost any one who uses logging uses them. If I have a log statement, I don't expect it's namespace to *change* every time I restart my program. How could I configure that? How could I even understand my own logs?

> This problem does not exist in the PR because it uses the filename in stead, correct?

There is no filename for UI rules. So if we use the rule UID (which is greatly deemphasized on the UI) instead we now have two completely different behaviors. Even without the many problems I point out with using the ruleUID already, I'd be against that inconsistency. This leads to confusion for the end users.

> I think the UX can be improved by sticking to one best practice workflow and have every tool in service of that workflow by default

But we don't have that and we are a *very long* way from achieving that. As a self proclaimed noob, you were not here for the huge amount of discussion and arguments that erupted as a result of the mere suggestion of dropping support for file based configs. I'm talking about multiple forum threads running to thousands of posts, a number of users and developers quitting in protest, the whole nine yards. 

So I'm looking for *consistency* between files and UI based. I'm not willing to tell one or the other "sorry, we don't support you any more." This approach says that. 

> In stead of: $date [$level] [$namespace] - $message
> You want: $date [$level] [$name] - $message
> 
> That's a different issue.

But that's *not* a different issue. That's exactly what I'm saying I want, or at least recommend as an alternative. The problem is:

1. the loggers are actually configured by the namespaces
2. the logger doesn't have access to the $name, as far as the logger is concerned the $namespace *is* the $name

You can't get to the "You want" without using $name as the namespace.

I've said all I have to say on this. Honestly if the fact that the ruleUID changes for file based users every time the rule is reloaded isn't enough to give pause and try to find and alternative (I don't really care if it's the name of the rule, just not the ruleUID), nothing will.  But I simply cannot fall behind advocating the use of `console` for logging to end users in their rules if that becomes the case. 

And that's the crux of my objections. If we don't encourage everyone to use `console` over `log` I don't really care one way or the other. Most end users will continue to use `log` and they can define their own logger namespaces as they go just like has always been the case. But if `console` becomes the default, we need to make sure it's as usable as `log` is and using the ruleUID doesn't do that. For file based users it makes the logs almost unusable.

Redsandro:
FWIW every other OH JavaScript tutorial I've seen already has `getLogger('org.openhab.script.' + ctx.ruleUID)` that's why for consistency's sake it seemed like a good default to me.

rkoshak:
You are never going to convince me that unique and short somehow trumps consistent (between UI rules and file based rules), stable (doesn't change from one load to the next) and usable by all (i.e. file based users should not have to go to MainUI each and every time they look at their logs to figure out which rule a log statement came from). 

If you are hell bent on using the ruleUID, then fix *that* first (see my post above). Make it so the UID can be user set-able at the creation of the rule so it doesn't change every time the rule loads, and since the user gets to define it the ID will be meaningful and memorable. If the user makes it long, who are we to say they can't? Either they will see what problems that creates or they know what they are doing.

While we're at it, in the case when the ID isn't preselected by the user, generate a shorter random UID (though this still has the problem where the UID changes on every load). NOTE: as it currently stands, the randomly generated UID *includes the rule's name* so your concern about long namespaces is still going to be a problem for file based users, only worse because it's 500 characters followed by 36 randomly generated characters. 

Either we need to fix the ruleUID so it can be used to generate consistent and usable logs for *all* users, something else needs to be used, or we cannot encourage all users to use `console` over `log` by default. Don't like the rule's name for this? I'm open to other suggestions.

digitaldan:
Hi , sorry i'm late to the party, i was not able to work on openHAB this week, but have a 3 day weekend, so hoping to make up lost time.  

I hear the different opinions and will propose a solution this weekend.  I think it will incorporate changes to openhab-js as well to better lock down rule ids  and support custom ids.

Redsandro:
@digitaldan commented:

> I hear the different opinions and will propose a solution this weekend.

Great! :+1: 

@rkoshak commented:

> I'm open to other suggestions.

Let's see what DigitalDan comes up with. Alternatively, since apparently the `ruleID` can be and sometimes is huge, contain all kinds of alphanumerical characters and other characters like dashes, perhaps the `ruleID` can be retired entirely and replaced by an autogenerated namespace along the lines of `org.oh.aut.script.` followed by a random 10 digit number _unless provided by the user in its entirety_ where a maximum length of 36 chars is recommended but the user is free to enter a paragraph if it makes the debug logs more valuable to them.

This way, every script has a namespace in stead of a ruleID, and we don't have to have two things.

> If you are hell bent on using the ruleUID, then fix that first. Make it so the

I'm confused. It's clear you feel strongly about this. Some weeks ago you explained to me that having an opinion isn't enough. _"If you feel strongly about something, you'll probably need to code it yourself or find someone to code it for you."_ So in the interest of moving this along I thought the right approach was to be appreciative of this PR, CI this in and anyone feeling strongly about improving it further will provide their PR in the next CI. Instructions unclear.

digitaldan:
Couple changes, the openhab-js library now allows rule ID's to be passed into the rule builder,  so i think this makes a nice compromise to rule names and IDs.  I also added the ability to set the console log name, for those who want to customize it, for example
```javascript
console.loggerName = "oh.bar";
console.warn("Hello !");
```

would produce 
```
14:53:46.138 [WARN ] [oh.bar                               ] - Hello!
```

Regarding my view on custom logging, I don't think there is anything wrong with it at all, the openhab-js library supports that nicely, but my preference to push console logging by default is that its convenient , familiar, and probably meets most user's needs out of the box.  I'm hoping this tweak to supply the logger name meets more intermediate usage needs, and for the very advanced you can use the openHAB native logger https://github.com/openhab/openhab-js#log from our helper library.


Let me know what you think.

Redsandro:
`loggerName` stays within rule context: check.
Looks good to me! :shipit:

digitaldan:
Thanks everyone, this is ready then and i have removed the WIP tag

rkoshak:
> I'm confused. It's clear you feel strongly about this. Some weeks ago you explained to me that having an opinion isn't enough. "If you feel strongly about something, you'll probably need to code it yourself or find someone to code it for you." So in the interest of moving this along I thought the right approach was to be appreciative of this PR, CI this in and anyone feeling strongly about improving it further will provide their PR in the next CI. Instructions unclear.

I was in fact looking for ways around this with the limited time I had. I *was* looking for alternatives and was prepared to make changes here or elsewhere (I was beat to it). The solution to making this PR workable was somewhere else, namely either fixing the ruleUID as @digitaldan did, adding the ability for users to set the namespace for console logging which was also implemented, or choosing something else to use for the logger namespace which is no longer needed. 

I would not and could not have prevented the original PR from being accepted. It wasn't a problem with using the ruleUID per se , it was that the way the ruleUID was generated would make the logs unusable. But I couldn't sit back and allow that to happen without fighting for a better solution. There still is going to be a pretty big set of users where their logs will become unusable anyway because they don't provide UIDs for their rules (no matter how they are created), but we can address that through docs, which I'll even volunteer to write. It for sure will be in the Getting Started Tutorial docs I'm working on for rules.

Redsandro:
@rkoshak ok gotcha. Thank you for your work on Getting Started tutorials.

I don't think I agree with your phrasing that there still is a _"big set of users where their logs will **become unusable**"_ because as far as I understand `loggerName` is a syntactic shortcut for existing `getLogger` behavior at worst, but it's entirely plausible I'm not having a complete picture here.

rkoshak:
It's pregnant but hashing this out further but I'm the interest of clarity...

I've personally lived though something very much like why l what I describe. Back when I used for based Jython rules:

1. there was no way for the user to pick the ruleUID which was randomly generated.

2. That randomly generated uid was quite long.

3. Everytime a .py, .js, etc file is loaded the old rule is deleted and a new rule is created.

4. Because of 1 and 3, everytime the time it's loaded it is loaded with a new and different uid.

5. Back then (and sometimes even now) when errors are reported the only identifying information is the ruleUID.

So, back then every time I had an error I would have to open PaperUI search through the list of rules and see what rule was throwing the error. My rules weren't in PaperUI. My kids weren't in PaperUI. So if have to have three windows open at all times: one to see the error, one to coordinate the rule uid to the name of the rule, and then vscode to find the rule with that name and make corrections.

And because of 4, I couldn't even try to remember that correlation because as soon as I save the file, the uid changes. It was a huge pain.

Flash forward to today. Let's assume 1-4 we're still true, which was the case when this PR was first submitted.

If we appended the rule uid to the logger name we are right back at 5, only not just for errors, but for all logs made with console.log (which is to be the recommended way to log).  When I see a log statement, the only unique part of that statement to tell me what rule it came from is a randomly generated and constantly changing string of gibberish. The only way to correlate a log statement to file based rule would be to go to MainUI and search for the right uid, to find the rule's name, to find the rule in the file. And every time the file changes you have to do it all over again because the uid changes on every load.

File based users should not be forced to use the UI just to discover where a log statement came from. File based users should not be forced to use the UI at all. 

No user should be forced to have to run a search to correlate a log statement with the rule it came from as a matter of course, even UI users. Maybe the first time or two if the user chose not to define their own uids.

Because of 4, that also means you can't configure the logger settings on a per rule basis. It would either be the settings for all rules or nothing. With the id changing every time the rule loads, you'd have to change the logger levels every time too.

digitaldan:
@wborn any change to get this merged?  Its a small logging fix that should be very useful.

wborn (approved):
LGTM and thanks! :+1:

PR #11947, opened 2022-01-03
[Denon HEOS] Update Denon CLI reference in plugin documentation

MBizm:
The old reference to Denon HEOS CLI documentation was outdated and is not available anymore. Updating link to latest version.
Thanks for great work!
Michael


wborn (approved):
Thanks for fixing it!

PR #11950, opened 2022-01-03
[jsscripting] Fixed links to JS-Joda

LightIsLife:
Fixed path to JS-Joda github page to be formatter as external links by simply deleting the preceding "#"


wborn (approved):
Thanks for fixing the links!

PR #11954, opened 2022-01-04
[jdbc] Fixed NPE when SQL query failed

cweitkamp:
- Fixed NPE when SQL query failed

In case of an error the `Yank.queryObjectArrays(..., ...)` method returns a `null` value.

### DIFF
--- a/bundles/org.openhab.persistence.jdbc/src/main/java/org/openhab/persistence/jdbc/db/JdbcBaseDAO.java
+++ b/bundles/org.openhab.persistence.jdbc/src/main/java/org/openhab/persistence/jdbc/db/JdbcBaseDAO.java
@@ -342,10 +342,14 @@ public class JdbcBaseDAO {
         String sql = histItemFilterQueryProvider(filter, numberDecimalcount, table, name, timeZone);
         logger.debug("JDBC::doGetHistItemFilterQuery sql={}", sql);
         List<Object[]> m = Yank.queryObjectArrays(sql, null);
+        if (m == null) {
+            logger.debug("JDBC::doGetHistItemFilterQuery Query failed. Returning an empty list.");
+            return List.of();
+        }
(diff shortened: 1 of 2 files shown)


bundles/org.openhab.persistence.jdbc/src/main/java/org/openhab/persistence/jdbc/db/JdbcBaseDAO.java
```diff
@@ -342,10 +342,14 @@ public void doStoreItemValue(Item item, ItemVO vo) {
         String sql = histItemFilterQueryProvider(filter, numberDecimalcount, table, name, timeZone);
         logger.debug("JDBC::doGetHistItemFilterQuery sql={}", sql);
         List<Object[]> m = Yank.queryObjectArrays(sql, null);
+        if (m == null) {
```
jlaur, line 345: Is this a bug in Yank? Seems it would be more appropriate if it would have thrown `YankSQLException`, as this seems to be an error situation? Do we have correct error handling in place for this exception - if so, maybe throw instead of returning empty list? Sorry, no time for digging deeper right now, so for now just asking questions. :)

cweitkamp: `YankSQLException` have been introduced in version 3.3.0. We are using version 3.2.0 which handles all exceptions internally by logging an error. Thus the return value of this method is `null`.

jlaur (approved):
LGTM

cweitkamp:
@jlaur Thanks for merging. Can you please add the Milestone afterwards. We use this property to identify PRs for automatic generation of our changelogs. Thanks.

jlaur:
@cweitkamp - of course, thanks for reminding me.

PR #11966, opened 2022-01-05
[atlona] Correct spelling of atlona

mlobstein:
Fix misspelled word _atlona_ throughout the binding. No functional changes.

### DIFF
--- a/bundles/org.openhab.binding.atlona/src/main/java/org/openhab/binding/atlona/internal/pro3/AtlonaPro3PortocolHandler.java
+++ b/bundles/org.openhab.binding.atlona/src/main/java/org/openhab/binding/atlona/internal/pro3/AtlonaPro3PortocolHandler.java
@@ -200,7 +200,7 @@ class AtlonaPro3PortocolHandler {
         try {
             response = callback.getResponse();
             if (!response.equals("")) {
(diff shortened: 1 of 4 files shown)


lolodomo (approved):
LGTM

lolodomo:
I have a doubt if this or not a breaking change, at least for users using UI to manage things. 
It is not for those using config files. 
Did you check that?

mlobstein:
> I have a doubt if this or not a breaking change, at least for users using UI to manage things. It is not for those using config files. Did you check that?

I built a jar without the changes, dropped it into addons, created a thing through the UI and verified it was working. I then built another jar with the changes, deleted the original jar from addons, waited a minute, and then dropped in the new jar. The thing re-initialized and functioned correctly. I also changed the polling interval setting in the UI and saved without any issues.

lolodomo:
> I built a jar without the changes, dropped it into addons, created a thing through the UI and verified it was working. I then built another jar with the changes, deleted the original jar from addons, waited a minute, and then dropped in the new jar. The thing re-initialized and functioned correctly. I also changed the polling interval setting in the UI and saved without any issues.

That is exactly the test I was expecting, that is checking it is still possible to update the thing configuration parameters from an old thing created with UI. 
Thank you for the test.

PR #11968, opened 2022-01-05
[wemo] Fix issues if device is not properly registered as UPn…

hmerk:
…P participant. This has a breaking change, as all Wemo Things have to be deleted and rediscovered. Should not be troublesome, as the Things will get the same ID (UDN) again.

### DIFF
--- a/bundles/org.openhab.binding.wemo/src/main/java/org/openhab/binding/wemo/internal/WemoUtil.java
+++ b/bundles/org.openhab.binding.wemo/src/main/java/org/openhab/binding/wemo/internal/WemoUtil.java
@@ -137,7 +137,7 @@ public class WemoUtil {
         return port == null ? null : "http://" + host + ":" + port + "/upnp/control/" + actionService + "1";
     }
 
-    private static boolean servicePing(String host, int port) {
+    public static boolean servicePing(String host, int port) {
         try {
             HttpUtil.executeUrl("GET", "http://" + host + ":" + port, 250);
             return true;
--- a/bundles/org.openhab.binding.wemo/src/main/java/org/openhab/binding/wemo/internal/discovery/WemoDiscoveryParticipant.java
+++ b/bundles/org.openhab.binding.wemo/src/main/java/org/openhab/binding/wemo/internal/discovery/WemoDiscoveryParticipant.java
@@ -13,7 +13,9 @@
 package org.openhab.binding.wemo.internal.discovery;
 
 import static org.openhab.binding.wemo.internal.WemoBindingConstants.*;
+import static org.openhab.binding.wemo.internal.WemoUtil.substringBetween;
 
+import java.net.URL;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.Set;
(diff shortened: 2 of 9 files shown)


hmerk:
Arrgh, it is failiing on the tests, but i do not know anything about the tests...

jlaur:
@hmerk - with this version I'm now getting this logged every minute:

```
2022-01-05 17:37:17.689 [ERROR] [ng.wemo.internal.handler.WemoHandler] - Failed to get actual state for device 'wemo:insight:insight2': null
```

However, I'm still receiving updates for channel `currentPower`. I'm using file-based configuration.

**wemo.things:**
```
wemo:insight:insight2 "WeMo 2" [udn="Insight-1_0-123456K7890E01"]
```

**wemo.items:**
```
Switch WeMo2_Switch "Switch" <switch> ["Switch"] { channel="wemo:insight:insight2:state" }
Number:Power WeMo2_Power "Power" <energy> ["Measurement", "Power"] { channel="wemo:insight:insight2:currentPower" }
```

jaywiseman1971:
> @hmerk - with this version I'm now getting this logged every minute:
> 
> ```
> 2022-01-05 17:37:17.689 [ERROR] [ng.wemo.internal.handler.WemoHandler] - Failed to get actual state for device 'wemo:insight:insight2': null
> ```
> 
> However, I'm still receiving updates for channel `currentPower`. I'm using file-based configuration.
> 
> **wemo.things:**
> 
> ```
> wemo:insight:insight2 "WeMo 2" [udn="Insight-1_0-123456K7890E01"]
> ```
> 
> **wemo.items:**
> 
> ```
> Switch WeMo2_Switch "Switch" <switch> ["Switch"] { channel="wemo:insight:insight2:state" }
> Number:Power WeMo2_Power "Power" <energy> ["Measurement", "Power"] { channel="wemo:insight:insight2:currentPower" }
> ```

You MUST delete your original WeMo Things and rediscover them to have this binding work.  This will create 3 new properties within the WeMo Thing for the rotating IP address issue.

Best, Jay

hmerk:
This is an issue with text based config. I don't know if you can set the properties within the .things definition.

jlaur:
> This is an issue with text based config. I don't know if you can set the properties within the .things definition.

That's not possible. I'll have a closer look and see if I can come up with a suggestion later. I don't think it's acceptable breaking text-based configuration, and I'm quite sure it can be avoided.

jlaur:
> Please try to set the descriptorURL in your thing definition.

If you mean add it as configuration parameter like this:

```
wemo:insight:insight2 "WeMo 2" [udn="Insight-1_0-123456K7890E01", descriptorURL="http://192.168.0.59:49154/setup.xml"]
```
or even:
```
wemo:insight:insight2 "WeMo 2" [udn="Insight-1_0-123456K7890E01", descriptorURL="http://192.168.0.59:49154/setup.xml", ipAddress="192.168.0.59", port=49154]
```

then it doesn't work in current version, because it's a thing configuration parameter, not a property. But I think you are on the right track.

> In astro binding, geolocation is a property, but can be configured in the thing definition.

I think that's because it's a configuration parameter?

```
UID: astro:sun:home
label: Astro Sun Data
thingTypeUID: astro:sun
configuration:
  useMeteorologicalSeason: false
  interval: 60
  geolocation: 57,9
```

> If that works, I just need to remove host and port property.

If I understand this correctly from the PR description and code, the binding currently depends on UPnP not only for discovery, but also for correct initialization:
https://github.com/openhab/openhab-addons/blob/51b1f40a62a457717c4713c4d22a54efb55ec9a4/bundles/org.openhab.binding.wemo/src/main/java/org/openhab/binding/wemo/internal/handler/WemoHandler.java#L110-L116

So discovery would (previously) retrieve UDN and add this as a property. The handler would then use this to get descriptorURL (containing IP address and port) through UPnP. But if for some reason this registration didn't work in some scenarios (as reported), it wouldn't work.

So it seems that what is most important for the binding to work is IP address and port. These could be added as configuration parameters, which would remove the UPnP dependency, i.e. any need for registering participant.

That's not a big deal for file configurations. For UI configuration based on discovery, I believe that properties added during discovery will be defaulted as configuration parameters, or at least they can be (not yet into how it exactly works).

Last thing, probably descriptorURL is not needed for anything, neither as configuration parameter or property? Just extract IP address and port.

hmerk:
UPnP is needed for receiving GENA events from the devices, so we cannot leave this part out.
I will close this PR for now and come back with a better solution.

jlaur:
> UPnP is needed for receiving GENA events from the devices, so we cannot leave this part out.

OK! And in meantime I just verified that by adding this in **thing-types.xml** in your branch:

```
			<parameter name="ipaddress" type="text" required="true">
				<label>IP address</label>
				<context>network-address</context>
				<description>Host name or IP address of the WeMo device</description>
			</parameter>
```

it will automatically be transferred as a configuration parameter from discovery and not as a property:

image

Issue #11969, opened 2022-01-05
[homematic] some devices offline (CONFIGURATION_ERROR) after restart of openHAB when using large number of homematic devices


raykleibelt:
When using a large number of Homematic devices, the default scan duration of 60 seconds is not sufficient to read the device description for all devices from the CCU. Devices whose device description cannot be read are subsequently offline and show a CONFIG_ERROR (e.g.: "Device with address '0007DD123A3DD1' not found on gateway 'MEQ0123456'"). The scan duration can be extended by setting the bridge configuration 'Install Mode Duration' to a higher value (r.g. 120 seconds). This solves the problem.

However, after restarting openHAB, the HomematicDeviceDiscoveryService always uses the default value of 60 seconds as scan duration instead of the configured 'Install Mode Duration'. 
As a result, some Homematic devices are offline after a restart of openHAB and do not show any reaction.
When the device scan is started manually afterwards, all devices are online again because the HomematicDeviceDiscoveryService then uses the configured 'Install Mode Duration'.

also reported here:
https://community.openhab.org/t/homematic-binding-not-all-devices-read-from-ccu/130936

I intend to submit a PR to solve the problem.

PR #11971, opened 2022-01-05
[homematic] fixing incorrect scan duration after openHAB restart

raykleibelt:
Bug fix for #11969

### Issue:
After restart of openHAB the scan duration for homematic devices is read from a variable which is initialized with a default value instead of using the configuration of the binding.

### Fix:
The scan duration is read directly from the configuration of the binding.


lolodomo (approved):
LGTM

PR #11973, opened 2022-01-06
Update license headers to 2022

Hilbrand:


clinique (approved):

hmerk (approved):

kaikreuzer (approved):
Time for it, thank you!

lolodomo:
As we are merging PRs from last year, it is not impossible that some headers with 2021 were inserted after your PR was created.
Maybe it should be run again before each new milestone, at least the first milestones of 2022.

wborn:
> Maybe it should be run again before each new milestone, at least the first milestones of 2022

Probably much sooner because the build fails due to a SAT error when the wrong headers are used. 😉

lolodomo:
Ok, good point, at least we have a way to detect quickly such problem.

Issue #7542, opened 2020-05-03
[kodi] State sometimes not updated


radokristof:
Sometimes the current playing state is not updated in Kodi.

## Expected Behavior
The state should always represent the current state.

## Current Behavior
Sometimes the current state stucks in the PAUSE state, but the player is playing something. On the web interface the current state is correctly shown. In Debug mode, I can usually see this:

```
==> /var/log/openhab2/openhab.log <==

2020-05-03 22:12:43.228 [DEBUG] [odi.internal.protocol.KodiConnection] - checkConnection: try to connect to Kodi ws://10.0.0.131:9090/jsonrpc

2020-05-03 22:12:43.264 [DEBUG] [i.internal.protocol.KodiClientSocket] - callMethod returns: {"id":133315,"jsonrpc":"2.0","result":{"version":{"major":18,"minor":6,"revision":"20200320-nogitfound","tag":"stable"}}}

2020-05-03 22:12:48.269 [DEBUG] [i.internal.protocol.KodiClientSocket] - callMethod returns: {"id":133316,"jsonrpc":"2.0","result":{"version":{"major":18,"minor":6,"revision":"20200320-nogitfound","tag":"stable"}}}
```

It doesn't matter if I play/pause the video, until I restart the binding.
If I restart the binding via the console, the DEBUG shows much better responses.

```
[...]

PR #11984, opened 2022-01-07
[kodi] Add channels for screensaver state and virtual keyboard state and include OnAVStart for playing state detection

Flole998:
Fixes: #7542


bundles/org.openhab.binding.kodi/src/main/java/org/openhab/binding/kodi/internal/KodiEventListener.java
```diff
@@ -53,6 +53,8 @@
 
     void updateScreenSaverState(boolean screenSaveActive);
```
lolodomo, line 54: Please also fix the parameter here

bundles/org.openhab.binding.kodi/src/main/java/org/openhab/binding/kodi/internal/protocol/KodiConnection.java
```diff
@@ -68,6 +68,7 @@
     private static final String PROPERTY_FANART = "fanart";
     private static final String PROPERTY_THUMBNAIL = "thumbnail";
     private static final String PROPERTY_VERSION = "version";
+    private static final String PROPERTY_SCREENSAVER = "System.ScreensaverActive";
```
lolodomo, line 71: Why using this new kind of naming for property?

bundles/org.openhab.binding.kodi/src/main/resources/OH-INF/thing/thing-types.xml
```diff
@@ -560,4 +562,16 @@
 		<description>Profile of Kodi</description>
 		<state pattern="%s"/>
 	</channel-type>
+	<channel-type id="screensaver">
```
lolodomo, line 565: Please also update the default translation properties

lolodomo (changes requested):

lolodomo:
@pail23 and @cweitkamp : would you like to review too?

bundles/org.openhab.binding.kodi/src/main/java/org/openhab/binding/kodi/internal/protocol/KodiConnection.java
Flole998: That's simply how Kodi is calling it.

bundles/org.openhab.binding.kodi/src/main/resources/OH-INF/thing/thing-types.xml
Flole998: Fixed

bundles/org.openhab.binding.kodi/src/main/java/org/openhab/binding/kodi/internal/KodiEventListener.java
Flole998: Fixed

bundles/org.openhab.binding.kodi/src/main/java/org/openhab/binding/kodi/internal/protocol/KodiConnection.java
lolodomo: It is not a thing property?

Flole998: No, all these `PROPERTY_*` are properties of the JSON that is sent to/from Kodi.

lolodomo: Ok, thank you

lolodomo (approved):
Code LGTM

lolodomo:
I let a chance to binding maintainers to review before merging. 
@pail23 @cweitkamp

bundles/org.openhab.binding.kodi/src/main/resources/OH-INF/thing/thing-types.xml
```diff
@@ -560,4 +562,16 @@
...
+		<item-type>Switch</item-type>
+		<label>Screensaver</label>
+		<description>Status of the screensaver</description>
+		<state readOnly="true"/>
+	</channel-type>
+	<channel-type id="inputrequested">
```
cweitkamp, line 571: I propose to model this as a trigger channel which emits the requested input type (or "FINISHED"). This way a user can implement different reactions based on the requested type. Wdyt?

bundles/org.openhab.binding.kodi/src/main/java/org/openhab/binding/kodi/internal/protocol/KodiConnection.java
```diff
@@ -994,7 +997,7 @@ public void handleEvent(JsonObject json) {
     }
 
     private void processPlayerStateChanged(String method, JsonObject json) {
-        if ("Player.OnPlay".equals(method)) {
+        if ("Player.OnPlay".equals(method) || "Player.OnAVStart".equals(method)) {
```
cweitkamp, line 1000: Nice finding 👍🏼 .

cweitkamp (changes requested):
Thanks for this improvement. I am missing the documentation for this new feature.

And I left a comment regarding the modelling of the `inputrequested` channel.

bundles/org.openhab.binding.kodi/src/main/resources/OH-INF/thing/thing-types.xml
Flole998: I'm trying to get a change into Kodi which allows detecting if an Input is requested or not. Currently, if openHAB is started up after the input is requested it wouldn't notice it, even though the dialog is still open. The idea was to add a boolean for that, so that it could be implemented similar to the screensaver detection. If this would indeed be modeled as a trigger channel this future enhancement would be a lot harder.

Also we actually have "Title" (like "Enter search String") and the type, so ideally we would give access to both. If you still think it's better to make it a trigger channel (with the disadvantage that at least I won't be able to implement the polling on startup/reconnection) then it might make more sense to use the title and not the type.

lolodomo:
Oh yes, I forgot the missing update of the documentation. Hopefully I did not merge it.

bundles/org.openhab.binding.kodi/src/main/resources/OH-INF/thing/thing-types.xml
cweitkamp: This will not work either. The `Input.OnInputRequested` is a Kodi notification which will be triggered only once when the user input is necessary. Thus if openHAB is not available at that specific moment or the websocket of the binding is not connected it will never be recognized by OH or set after a restart. You are not polling for it. Even more the state of the item linked to the channel will be `NULL` after an restart.

Regarding the title I can think of an OH enhancement to define additional properties delivered by an event. Currently an event can only have a type no other payload.

Flole998: Exactly, there is currently no way of polling it in Kodi. But I am working on implementing that, it will just be a boolean though.

cweitkamp: Alright. Please ping me once you got it implemented.

bundles/org.openhab.binding.kodi/src/main/resources/OH-INF/thing/thing-types.xml
```diff
@@ -560,4 +562,16 @@
 		<description>Profile of Kodi</description>
 		<state pattern="%s"/>
 	</channel-type>
+	<channel-type id="screensaver">
+		<item-type>Switch</item-type>
+		<label>Screensaver</label>
+		<description>Status of the screensaver</description>
```
jlaur, line 568: These texts (and inputrequested) must be added to resources/OH-INF/i18n/kodi.properties as well:
```
channel-type.kodi.screensaver.label = Screensaver
channel-type.kodi.screensaver.description= Status of the screensaver
channel-type.kodi.inputrequested.label = InputRequested
channel-type.kodi.inputrequested.description= Status that shows if Kodi is asking for user input.
```

bundles/org.openhab.binding.kodi/src/main/resources/OH-INF/thing/thing-types.xml
```diff
@@ -560,4 +562,16 @@
...
+		<state readOnly="true"/>
+	</channel-type>
+	<channel-type id="inputrequested">
+		<item-type>Switch</item-type>
+		<label>InputRequested</label>
```
jlaur, line 573: "Input requested" in two words?

bundles/org.openhab.binding.kodi/src/main/resources/OH-INF/thing/thing-types.xml
Flole998: Just to confirm: You want to wait for the change before merging? That change needs to be done at Kodi first and the maintainers there have already made clear on a different PR that there won't be API-changes between versions, so before it could actually be used a new version needs to be released with those changes in. I don't think waiting for that is a good idea as that could very well take another year or so, so I'd rather kick out this change if it's a reason not to merge and put it in a diferent PR.

bundles/org.openhab.binding.kodi/src/main/resources/OH-INF/thing/thing-types.xml
Flole998: It's already there, isn't it?

bundles/org.openhab.binding.kodi/src/main/resources/OH-INF/thing/thing-types.xml
Flole998: Yes, that's probably better. Thanks!

bundles/org.openhab.binding.kodi/src/main/resources/OH-INF/thing/thing-types.xml
jlaur: Right. I somehow missed the file, sorry.

bundles/org.openhab.binding.kodi/src/main/resources/OH-INF/thing/thing-types.xml
Flole998: Fixed

Flole998:
I just updated the README.md aswell.

lolodomo:
@cweitkamp  and @jlaur : please have the lead to finish the review & merge of this PR.

bundles/org.openhab.binding.kodi/src/main/resources/OH-INF/thing/thing-types.xml
jlaur: @cweitkamp, @Flole998 - can this discussion be concluded? Anything preventing this PR for being merged?

cweitkamp: Yes, I am fine with it. We can merge this now.

cweitkamp (approved):

jlaur (approved):
Thanks for this contribution!

PR #11992, opened 2022-01-07
[danfossairunit] Consolidate properties for compliance

### LINKED ISSUE #11991: [danfossairunit] Consolidate properties
Properties for the Danfoss HRV thing doesn't comply to naming convention:
https://www.openhab.org/docs/developer/bindings/thing-xml.html#properties

The two defines properties are:
- **Unit Name:** should be renamed to **modelId**.
- **Serial Number:** should be renamed to **serialNumber**.

jlaur:
Fixes #11991

This pull request was made for compliance with naming convention for properties in general and for specific property types in particular. Changes:
- **serialNumber** renamed from **Serial Number**.
- **modelId** renamed from **Unit Name**.
- **vendor** introduced.

Also fixed small issue with properties showing as **unknown** during startup.

There is a slight problem when upgrading to this version, but I wouldn't consider it a breaking change as it's only cosmetic. For things created from UI, there will be duplicate properties until thing is recreated.

image

For file-based configuration this problem is resolved by openHAB restart:

image


lolodomo (approved):
LGTM

Issue #11878, opened 2021-12-28
[InfluxDB] Crashing influxDB 2.1 with a simpel rule: previousState


FutureCow:
## Expected Behavior
Getting the prevousState from an item with a simple DSL rule.

## Current Behavior
To make to problem as easy a possible i create a simple DSL rule:

```
configuration: {}
triggers:
  - id: "1"
    configuration:
      itemName: Schakel_Server_Power
    type: core.ItemStateChangeTrigger
conditions: []
actions:
  - inputs: {}
    id: "2"
    configuration:
      type: application/vnd.openhab.dsl.rule
      script: 'logInfo("Test", "recorded: " + Schakel_Server_Power.previousState(false,"influxdb").toString)'
    type: script.ScriptAction
```
This rules make my InfluxDB 2.1.1 crash! (do I need some more InfluxDB debug? how?)
```
Dec 22 22:57:44 server systemd[1]: influxdb.service: A process of this unit has been killed by the OOM killer.
Dec 22 22:57:44 server systemd[1]: influxdb.service: Main process exited, code=killed, status=9/KILL
Dec 22 22:57:44 server systemd[1]: influxdb.service: Failed with result 'oom-kill'.
Dec 22 22:57:44 server systemd[1]: influxdb.service: Consumed 58.711s CPU time.
Dec 22 22:57:46 server systemd[1]: influxdb.service: Scheduled restart job, restart counter is at 43.
Dec 22 22:57:46 server systemd[1]: Stopped InfluxDB is an open-source, distributed, time series database.
Dec 22 22:57:46 server systemd[1]: influxdb.service: Consumed 58.711s CPU time.
Dec 22 22:57:46 server systemd[1]: Starting InfluxDB is an open-source, distributed, time series database...
```
This a the error inside OH3.2:

```
Caused by: java.net.ConnectException: Failed to connect to /192.168.1.2:8086
        at okhttp3.internal.connection.RealConnection.connectSocket(RealConnection.java:265) ~[?:?]
        at okhttp3.internal.connection.RealConnection.connect(RealConnection.java:183) ~[?:?]
        at okhttp3.internal.connection.ExchangeFinder.findConnection(ExchangeFinder.java:224) ~[?:?]
        at okhttp3.internal.connection.ExchangeFinder.findHealthyConnection(ExchangeFinder.java:108) ~[?:?]
        at okhttp3.internal.connection.ExchangeFinder.find(ExchangeFinder.java:88) ~[?:?]
[...]

PR #11993, opened 2022-01-07
[influxdb] Fix previous item state performance issues

lujop:
Fix issue #11878 
Optimizing a filter with limit 1 and time descending order for `| last()`
More discussion in the issue thread

### DIFF
--- a/bundles/org.openhab.persistence.influxdb/src/main/java/org/openhab/persistence/influxdb/internal/influx2/Influx2FilterCriteriaQueryCreatorImpl.java
+++ b/bundles/org.openhab.persistence.influxdb/src/main/java/org/openhab/persistence/influxdb/internal/influx2/Influx2FilterCriteriaQueryCreatorImpl.java
@@ -87,17 +87,29 @@ public class Influx2FilterCriteriaQueryCreatorImpl implements FilterCriteriaQuer
             flux = flux.filter(restrictions);
         }
 
-        if (criteria.getOrdering() != null) {
-            boolean desc = criteria.getOrdering() == FilterCriteria.Ordering.DESCENDING;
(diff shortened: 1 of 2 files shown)


lujop:
Attached  a compiled version for easy test it
[org.openhab.persistence.influxdb-3.3.0-SNAPSHOT.zip](https://github.com/openhab/openhab-addons/files/7831884/org.openhab.persistence.influxdb-3.3.0-SNAPSHOT.zip)
g

wborn:
The fix looks good to me. Do you also know if a similar fix should be made in the `Influx1FilterCriteriaQueryCreatorImpl` for users that are still using InfluxDB 1.x?

lujop:
I think that it isn't needed in Inlux1 because it doesn't support sorting. In fact you can only choose TIME desc or normal TIME asc.
But to be 100% sure do you have access to a InfluxDB 1.0 with a lot of data?
It will be as easy as checking how much time needs a select order by time desc limit 1 in a table.

wborn:
> But to be 100% sure do you have access to a InfluxDB 1.0 with a lot of data?

I still have backups of my InfluxDB 1.8 data which I can easily revive in a Docker container. My biggest measurement seems to go back ~11 years and has 267449 points.

I've only recently migrated to 2.1.1 which I've postponed for quite some time. That's because I also wanted to combine measurements into one using the item metadata approach without losing any history. :wink:

lujop:
Then if you can do a select in one of the measurements with Order by time desc limit 1 and it's more or less instantaneous, or so fast as a limit 1 without order, we can be 100% sure no optimization is needed.

wborn:
After a brief test I didn't notice any changes in performance with InfluxDB 1.8.9. :slightly_smiling_face:

wborn (approved):
Thanks for the fix!

PR #11997, opened 2022-01-08
[miele] Consolidate properties for compliance

### LINKED ISSUE #11996: [miele] Consolidate properties
Property refactoring suggestions:
- **model** should be renamed to **modelId** for system compliance.
- **firmwareVersion** should be added.
- **vendor** should be added.

jlaur:
Fixes #11996

This pull request provides minor changes for property improvements:
- **modelId** renamed from **model**.
- **vendor** introduced.
- **firmwareVersion** introduced.
- **connectionBaudRate** introduced.

### DIFF
--- a/bundles/org.openhab.binding.miele/src/main/java/org/openhab/binding/miele/internal/handler/MieleBridgeHandler.java
+++ b/bundles/org.openhab.binding.miele/src/main/java/org/openhab/binding/miele/internal/handler/MieleBridgeHandler.java
@@ -144,6 +144,11 @@ public class MieleBridgeHandler extends BaseBridgeHandler {
         }
 
         @NonNull
+        public String getFirmwareVersion() {
+            return Properties.get("firmware.version").getAsString();
+        }
+
+        @NonNull
         public String getRemoteUid() {
             JsonElement remoteUid = Properties.get("remote.uid");
             if (remoteUid == null) {
@@ -161,6 +166,14 @@ public class MieleBridgeHandler extends BaseBridgeHandler {
                 return null;
             }
             return connectionType.getAsString();
+        }
+
+        public String getConnectionBaudRate() {
+            JsonElement baudRate = Properties.get("connection.baud.rate");
+            if (baudRate == null) {
+                return null;
+            }
+            return baudRate.getAsString();
         }
 
         @NonNull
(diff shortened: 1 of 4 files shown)


bundles/org.openhab.binding.miele/src/main/java/org/openhab/binding/miele/internal/handler/MieleBridgeHandler.java
```diff
@@ -163,6 +168,14 @@ public String getConnectionType() {
             return connectionType.getAsString();
         }
 
+        public String getConnectionBaudRate() {
+            JsonElement baudRate = Properties.get("connection.baud.rate");
```
lolodomo, line 172: Please add `@Nullable` to the method return. 

And if not too much work, switch the class to non null by default.

lolodomo (changes requested):

bundles/org.openhab.binding.miele/src/main/java/org/openhab/binding/miele/internal/handler/MieleBridgeHandler.java
jlaur: > Please add `@Nullable` to the method return. 

Currently that would be redundant as class is null by default.

> And if not too much work, switch the class to non null by default. 

The class is for JSON deserialization, so I'm not sure it makes much sense? It would require a massive amount of @Nullable decorations except for a few methods.

lolodomo: Ok, I see, it is in class HomeDevice, not in class MieleBridgeHandler.

I guess it is a very old binding ?

lolodomo (approved):
LGTM

bundles/org.openhab.binding.miele/src/main/java/org/openhab/binding/miele/internal/handler/MieleBridgeHandler.java
jlaur: > I guess it is a very old binding ?

Yes, since 1.x I think. I've refactored it a lot recently, but still much to do. For example extracting HomeDevice. :-)

Issue #7872, opened 2020-06-07
Echonet Lite Support? (for Mitsubishi MAC-568IF-E Wifi Controller)


jmone1:
<a alt="bounty" href="https://www.bountysource.com/issues/91653932"><img src="https://api.bountysource.com/badge/issue?issue_id=91653932"></a>

As discussed in the [OpenHAB Community Post](https://community.openhab.org/t/echonet-lite-support-for-mitsubishi-mac-568if-e-wifi-controller/100159):

I have a Mitsubishi Air Conditioner with a MAC-568IF-E Wifi Controller that supports Echonet Lite direct integration.  I see there was some work done for a binding years ago:
https://github.com/kaz260/org.openhab.binding.echonetlite

There is also a working integration being discuss at HA 
https://community.home-assistant.io/t/mitsubishi-aircon/37454/99

It seems that there is a whole range of HVAC equipment produced in Japan that uses Econet Lite so hopefully this would work with a wide range of such devices.  The full specification for Echonet Lite is here
https://echonet.jp/product_en/echonet_lite_specification/

Thanks
Nathan

jmone1:
FYI - I created a Bounty for this (not sure why it is not taged on this request however).
https://www.bountysource.com/issues/91653932-echonet-lite-support-for-mitsubishi-mac-568if-e-wifi-controller

Thanks
Nathan

wborn:
Thanks for adding the significant bounty! Adding the label is a manual operation. :wink:

ab3be6:
Hello Nathan,

I also would like OH to support the echonet lite device.
So I found a tool and solution for this and I share it with you.
I hope this information helps you.

[https://github.com/KAIT-HEMS/node-picogw](url)
This gateway runs on node.js and can convert the echonet device to the MQTT topic.
If your OH has the MQTT binding and broker, you can find the echonet device as the MQTT topic.

I tried to set up this GW and MQTT broker, but I don't have experience with MQTT binding.
Now my OH does not detect an echonet device yet...

just for your information, the PicoGW could not run on a raspberry pi which already has MQTT broker mosquitto.
of course, vanilla raspberry pi can run the PicoGW.

If you have interested in this configuration, and you have succeeded to set it up, could you share your work steps with me?

jmone1:
Thanks - I did see that, but was hoping by adding a bounty someone (with much more knowledge than me) would be able to create a native adapter for it for OH3

mikeb01:
[...]

PR #11999, opened 2022-01-09
[echonetlite] Initial contribution

mikeb01:
This pull request adds the initial structure for and add-on to support Echonet Lite based devices (https://echonet.jp/product_en/echonet_lite_specification/).  Echonet Lite supports a broad range of devices, however this only contains implementation for the properties and device values used by Mitsubishi heat pumps as that is the only device I currently have available for testing.  As per the feature request #7872.

Signed-off-by: Michael Barker <mikeb01@gmail.com>

Jar: https://openhab.jfrog.io/artifactory/libs-pullrequest-local/org/openhab/addons/bundles/org.openhab.binding.echonetlite/3.4.0-SNAPSHOT/org.openhab.binding.echonetlite-3.4.0-SNAPSHOT.jar


netwolfuk:
Hi @mikeb01 Thanks for putting all the effort into this binding. I am trying it out with a Mitsubishi mac-568if-e WiFi adaptor.

However my docker server, and my IoT vlan are on different subnets, so that multicast discovery if not finding it. Therefore,it does not discover the group code, class code or instance number. Can you make suggestions on what those values should be for the above adaptor?

I think I have enabled Echonet correctly in the Mitsi WiFi app, but it's hard to validate UDP connectivity.

netwolfuk:
Hmm, perhaps it's possible to "half configure" the binding somehow? eg, give it the IP and let it do the rest of the scanning? I'll have a dig through the code and see if I can figure that out too. Thanks again for all your work so far @mikeb01

mikeb01:
Hi @netwolfuk

For heat pumps/Home AC,
Group Code: 1
Class Code: 48

The value for 'instance' is almost certain to be 1 (it is for all of mine) and just trying that value won't hurt.  One way to check would be to connect a laptop to your IoT VLAN and run OpenHab and the echonetlite bindings to do a scan to find out the values, then enter them into the OpenHab instance that is running within the container.

It should be possible to do the half configuration you mention.  All devices run a host controller with fixed values for the group/class/instance triple, which can be queried for the devices that are attached.  That would return the specific values for the Home AC.

netwolfuk:
Thanks @mikeb01 Adding the Group and Class code values has the binding working.

I also found this line in `EchonetClass` which lines up with your reply...

    AIRCON_HOMEAC(0x01, 0x30, Epc.Device.values(), Epc.AcGroup.values(), Epc.HomeAc.values()),

Converting those first two values to decimal is as you stated.

I have also tagged IoT vLAN traffic to my VM host, and will add the Docker VM to the IoT subnet directly. This should avoid me having to forward multicast traffic in future. I suspect that would have solved it immediately, and will hopefully prevent more of these discovery issues going forward.

My next task if to copy your binding to add Airtouch4 support to OH. It looks like these AirCon protocols are all bitmasks and bytes, rather than JSON and XML like I'm used it.

Anyway, thanks again for your help.

mikeb01:
Does the Airtouch4 also support Echonetlite?  If so my existing binding could be extended to support it.  The only difference is that there may be additional properties that aren't yet implemented.  If you can point me to a specification I can have a look.

netwolfuk:
It doesn't mention echonet anywhere in the pdf I found online and I've not looked closely enough at echonet to find the differences.

I'll try to upload it when I get to a computer.

netwolfuk:
Regarding this PR. When I switched the AC off, the outside temp went from 17° to 126° C which seems unlikely.

mikeb01:
I've seen something similar with outside temp on my system.  I don't think that the device actually supports it, but it reports as an available property in the available meta data and sends out a value (that looks invalid) for it.  There's not really anything the bindings can do it in this situation, I just don't add that channel to the model.

netwolfuk:
I just noticed that in the WiFi app from Mitsi, the "Show outdoor temperature" option has a note that says "This may not be available when power is off".

I've somehow managed to break connectivity by moving the vlan connection directly onto the container. I'll keep looking once I figure out how to enable extra logging.

mikeb01:
Yes, it does look like very different protocol.  If you are going to write a custom binding for it I would recommend a couple of things:
- If you haven't written a OpenHab binding before, make a simple one (e.g. reads data from a file) to get a feel for the structure and set of classes used for a binding (also follow the docs as well as they have some scripts that simplify setup).
- If you can find an existing client for the protocol (better if the code for it is open source as well) that allows you poke at your device that can be really helpful.
- Get familiar with Wireshark for checking the exact bytes going out on the wire.

netwolfuk:
Yeah. Thanks for the tips. I did write a binding for OH1 the oppo bluray player over RS232. But my PR had build issues like this one so it never got merged.

I planned to write the driver layer first and exercise that via junit or similar. I might even publish that as a maven artifact for other projects to use.

Then move onto wrapping that in a binding.

My wireshark skills are rusty. That's a great tip. Thanks

mikeb01:
With regards to the build errors, it appears to be an unrelated module that is not compiling correctly (MQTT).  It appears to be broken in the `main` branch as well.

jlaur:
> With regards to the build errors, it appears to be an unrelated module that is not compiling correctly (MQTT). It appears to be broken in the `main` branch as well.

This problem should be fixed now (by disabling the broken test): #12793

netwolfuk:
Oh ok. I cloned your PR and merged the latest `main` into it, but it didn't fix it. I thought maybe I was holding it wrong. I'll check my CI server sometime.

netwolfuk:
Thanks for your reply @jlaur 

I've merged main onto my local branch, and my CI at home says it's a different failing test.
```
org.mockito.exceptions.verification.WantedButNotInvoked: 

Wanted but not invoked:
thingCallback.statusUpdated(
    thing,
    <custom argument matcher>
);
-> at org.openhab.binding.groupepsa.internal.things.GroupePSAHandlerTest.intializeAndCheckChannels(GroupePSAHandlerTest.java:149)

However, there were exactly 24 interactions with this mock
```
I also setup a build to run `openhab-addons/main` from github (rather than this PR), and that is failing with the same error.

Perhaps I am doing something wrong.

mikeb01:
> This problem should be fixed now (by disabling the broken test): https://github.com/openhab/openhab-addons/pull/12793

There still appears to be build errors in main (Checkstyle failures).

netwolfuk:
Hi @mikeb01 I have downloaded the updated jar (it has a different md5sum so I assume it is updated).

Now when I try to use my thing, I am getting...
```
[INFO ] [onetlite.internal.EchonetLiteHandler] - Initialising: EchonetLiteConfiguration{hostname='192.168.7.102', port=3610, groupCode=1, classCode=48, instance=1, pollIntervalMs=30000, retryTimeoutMs=1000}
[WARN ] [onetlite.internal.EchonetLiteHandler] - Unable to handle commands if bridge is null
```
I didn't previously have a bridge, so I created one (it says it's online). How do I associate the two? Do I need to have a bridge with the latest update, or have I done something wrong?

My thing code looks like this in OH3...
```
UID: echonetlite:device:d6d6871a99
label: EchonetLite Device
thingTypeUID: echonetlite:device
configuration:
  classCode: 48
  hostname: 192.168.7.102
  instance: 1
  port: 3610
  pollIntervalMs: 30000
  retryTimeoutMs: 1000
  groupCode: 1
```
and my bridge is...
```
UID: echonetlite:bridge:b3b64341f0
label: Echonet Bridge
thingTypeUID: echonetlite:bridge
configuration:
  port: 3610
  multicastAddress: 224.0.23.0
```
Note: currently my OH VM (I have moved off Docker) is not on the IOT network, so I suspect the bridge will never find the AC unit via multicast. I will look at re-adding it to both LAN and IOT networks, but last time I was flummoxed by a routing issue so disabled that ethernet device.

Can the new binding still work without a bridge connection? I don't quite understand what the bridge provides, as the description is `Virtual bridge to ensure that the a single binding to the echonet port`

mikeb01:
Yes, you do need to have a bridge.  Let me do some testing with adding a bridge then manually adding the thing.  I haven't tried that recently as scanning was working fine.

We need to have a bridge as all communications runs via the same port 3610, in both directions.  An Ephemeral port can't be used, so there can only be one network connection for all things, so I route all of the network requests through the bridge.

mikeb01:
Looks like I need to add some code for handling the case of manually adding an echonetlite thing to a bridge.

netwolfuk:
> Looks like I need to add some code for handling the case of manually adding an echonetlite thing to a bridge.

I would be very appreciative. 

I spent last last evening learning about avahi, mDNS, IGMP Snooping, Multicast, etc. My openhab machine is on my server network and my IOT network, and avahi can see the mdns items on the server network, but not the IOT network. I still have much to learn apparently. Somewhere between my switches, firewall and ubiqiti Wifi I am possibly blocking multicast on that subnet.

mikeb01:
> Somewhere between my switches, firewall and ubiqiti Wifi I am possibly blocking multicast on that subnet.

It also possible that the multicast TTL is not large enough.  This will define how manage networks the multicast traffic can hop through before the packet gets dropped.  I'll have look into this as well, but I'll fix the manual addition first.

netwolfuk:
> This will define how manage networks the multicast traffic can hop through before the packet gets dropped.

Ah, interesting. I have three switches, plus the 3 ubiqiti APs, which I presume act like switches. Plus, my firewall between each subnet.
However, I would have thought I'd see some items in `avahi-browse` from the command line, which running outside of Openhab.

mikeb01:
I've pushed a small update.

I missed a small setting in the thing metadata.  Now when a device is added there is a configuration option for the bridge which allows the selection of a bridge.  I would also recommend re-adding the device and ensuring that the timeout is 2000ms.  I changed this recently as otherwise you end up with too many invalid timeouts.  It should now be the default.

mikeb01:
It needs a clean build to upload the new artefact.  Looks like a build error at the moment.

netwolfuk:
Thanks for the update. I built it locally, and was able to edit my existing Thing and add the new Bridge to it.
I did notice the timeouts, and was going to mention that next. I assumed it was possibly because my AC has no internet, but I have now increased the timeout to 2000ms, and will see how that goes.

Thanks again for all your hard work on this. My thing is now green again in OH3.

.gitattributes
```diff
@@ -1,2 +1,4 @@
 *.java text=auto
 *.xml text=auto
+text eol=lf
+
```
wborn, line 4: This seems wrong for Windows users who use crlf with text file line endings.

```suggestion
```

Removing the changes to this file will also make your PR build a lot faster. :wink:

mikeb01:
@netwolfuk You will get occasional timeouts, which is expected when using UDP over WiFi.  I found locally that 1000ms was too short for a timeout, sometimes the device would take slightly longer than that to respond.  If the timeout is too short you will start seeing messages like "Unexpected XXX response: XXX".  This is a cue to increase the timeout.  If you enable debug logging for the echonetlite binding it will also report the time each response took and you can use that to get a gauge of the latency.

netwolfuk:
@mikeb01 Thanks for the info. Can see the following, and I presume that's what you're referring to...

`[DEBUG] [ng.echonetlite.internal.EchonetObject] - Get_Res response time: 6ms`
`[DEBUG] [ng.echonetlite.internal.EchonetObject] - Get_Res response time: 35ms`

Assuming that's the correct log entry, it's pretty quick, albeit varying from ~6 to ~35ms.

With regards to mDNS. Should I be able to see the echonet device with a Bonjior discovery tool? Connecting my Mac to the IOT wifi, I can see `_openhab-server._tcp.` and a few others. Interestingly, I can only find hard-wired machines. Nothing on the actual wifi is discoverable. 

I realise this is the wrong place to be asking. Perhaps you know of a good primer on this stuff I could read up on?

mikeb01:
@netwolfuk Yes, that is the logging I was talking about.  6-35ms is typical, but I found occasionally it would take an extra second (e.g. 1032ms) and keeping the timeout at 1000ms meant a lot of these were discarded unnecessarily.  2000ms was a nice sweet spot.  However, that could just be something particular to my WiFi.

AFAIK, the Mitsubishi devices don't support mDNS.  mDNS is an additional protocol on top of Multicast UDP.  Echonetlite doesn't require mDNS support.  I would not expect the devices to turn up in Bonjour or Avahi.

netwolfuk (commented):
Just a few typos or grammatical errors in the README text.

netwolfuk:
Hi @mikeb01 I'm not sure if you're aware of this, but I saw this comment which implies that the binding can be promoted to a community binding before it's merged.

https://community.openhab.org/t/new-binding-nad-a-v-receiver-binding/133353/2

bundles/org.openhab.binding.echonetlite/pom.xml
```diff
@@ -0,0 +1,17 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns="http://maven.apache.org/POM/4.0.0"
+  xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd">
+
+  <modelVersion>4.0.0</modelVersion>
+
+  <parent>
+    <groupId>org.openhab.addons.bundles</groupId>
+    <artifactId>org.openhab.addons.reactor.bundles</artifactId>
+    <version>3.3.0-SNAPSHOT</version>
```
wborn, line 10:
```suggestion
    <version>3.4.0-SNAPSHOT</version>
```

netwolfuk: Hi @wborn   I noticed that the skeletons are still on 3.3.0 as well. I have raised a PR for that. See #13107

wborn: Very good! :+1:

mikeb01:
@wborn @jlaur What are the next steps for getting this reviewed for merging?  It is fairly stable and there are a few users running it successfully now.

microneer:
This binding is working very well for me on my 3.3.0 M7 system, I suggest this would be a quick win for OH to accept the PR.

jlaur:
@mikeb01 - so the PR is ready for review and no longer work in progress? When you confirm I will remove the label and assign maintainers for reviewing, so all will be notified.

mikeb01:
@jlaur Yes I it is ready for review.

jlaur:
> Yes it is ready for review.

Did you push your changes? It seems my latest comments are not resolved.

bundles/org.openhab.binding.echonetlite/src/main/resources/OH-INF/thing/channel-types.xml
```diff
@@ -0,0 +1,366 @@
...
+				<option value="Fan">Fan</option>
+				<option value="Other">Other</option>
+			</options>
+		</state>
+	</channel-type>
+
+	<channel-type id="setTemperature">
+		<item-type>Number</item-type>
+		<label>Set Temperature</label>
+		<description>Set Temperature</description>
```
jlaur, line 345: Is it a setpoint, i.e. desired target temperature? More description could be added.

bundles/org.openhab.binding.echonetlite/src/main/resources/OH-INF/thing/channel-types.xml
```diff
@@ -0,0 +1,366 @@
...
+		<category>Heating</category>
```
jlaur, line 346: If this is a temperature setpoint, then category **Temperature** seems more appropriate:
```suggestion
		<category>Temperature</category>
```

bundles/org.openhab.binding.echonetlite/src/main/resources/OH-INF/i18n/echonetlite_xx.properties
```diff
@@ -0,0 +1,21 @@
+# FIXME: please substitute the xx with a proper locale, ie. de
```
jlaur, line 1: Please remove this file.

bundles/org.openhab.binding.echonetlite/src/main/java/org/openhab/binding/echonetlite/internal/HexUtil.java
```diff
@@ -0,0 +1,98 @@
...
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package org.openhab.binding.echonetlite.internal;
+
+import java.nio.ByteBuffer;
+
+/**
+ * @author Michael Barker - Initial contribution
+ */
+public class HexUtil {
```
jlaur, line 20: Please add `@NonNullByDefault` annotation.

bundles/org.openhab.binding.echonetlite/src/main/java/org/openhab/binding/echonetlite/internal/EchonetLiteHandler.java
```diff
@@ -0,0 +1,187 @@
...
+
+        @Nullable
+        final EchonetLiteBridgeHandler bridgeHandler = bridgeHandler();
+        if (null == bridgeHandler) {
+            updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.CONFIGURATION_ERROR, "Bridge is null");
+            return;
+        }
+
+        try {
+            // TOOD: Move DNS lookup out to background
```
jlaur, line 118: Can this be resolved?

bundles/org.openhab.binding.echonetlite/src/main/java/org/openhab/binding/echonetlite/internal/EchonetLiteHandler.java
```diff
@@ -0,0 +1,187 @@
...
+        final EchonetDeviceConfig config = getConfigAs(EchonetDeviceConfig.class);
+
+        logger.info("Initialising: {}", config);
+
+        updateStatus(ThingStatus.UNKNOWN);
+
+        @Nullable
+        final EchonetLiteBridgeHandler bridgeHandler = bridgeHandler();
+        if (null == bridgeHandler) {
+            updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.CONFIGURATION_ERROR, "Bridge is null");
```
jlaur, line 113: Please support localization by adding an i18n string here. See for example:
https://github.com/openhab/openhab-addons/blob/e4917c3b4a41a178c5176b75d786dfd383c807ee/bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/handler/HDPowerViewShadeHandler.java#L109-L110

and:
https://github.com/openhab/openhab-addons/blob/e4917c3b4a41a178c5176b75d786dfd383c807ee/bundles/org.openhab.binding.hdpowerview/src/main/resources/OH-INF/i18n/hdpowerview.properties#L70

bundles/org.openhab.binding.echonetlite/src/main/java/org/openhab/binding/echonetlite/internal/EchonetLiteBridgeHandler.java
```diff
@@ -0,0 +1,399 @@
...
+                final int pdc = echonetMessage.currentPdc();
+                ByteBuffer edt = echonetMessage.currentEdt();
+                echonetObject.applyProperty(instanceKey, esv, epc, pdc, edt);
+            }
+        }
+    }
+
+    private void poll() {
+
+        updateStatus(ThingStatus.ONLINE);
```
jlaur, line 266: This seems too early since the polling could fail, in which case it would immediately after switch to **OFFLINE**.

bundles/org.openhab.binding.echonetlite/src/main/java/org/openhab/binding/echonetlite/internal/EchonetLiteBridgeHandler.java
```diff
@@ -0,0 +1,399 @@
...
+        try {
+            while (!Thread.currentThread().isInterrupted()) {
+                final long nowMs = clock.millis();
+                pollRequests(nowMs);
+                pollDevices(nowMs, requireNonNull(echonetChannel));
+                pollNetwork(nowMs, requireNonNull(echonetChannel));
+            }
+        } catch (Exception e) {
+            updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.CONFIGURATION_ERROR, e.getMessage());
+            logger.error("Poll failed", e);
```
jlaur, line 276: This logging is not needed as the status change will already be logged by the framework.

bundles/org.openhab.binding.echonetlite/src/main/java/org/openhab/binding/echonetlite/internal/EchonetLiteBridgeHandler.java
```diff
@@ -0,0 +1,399 @@
...
+    }
+
+    public void removeDevice(final InstanceKey instanceKey) {
+        requests.add(new RemoveDevice(instanceKey));
+    }
+
+    private void removeDeviceInternal(final RemoveDevice removeDevice) {
+        final EchonetObject remove = devicesByKey.remove(removeDevice.instanceKey);
+
+        logger.info("Removing device: {}, {}", removeDevice.instanceKey, remove);
```
jlaur, line 131: Please reduce log level to DEBUG or TRACE. See https://next.openhab.org/docs/developer/guidelines.html#f-logging

bundles/org.openhab.binding.echonetlite/src/main/java/org/openhab/binding/echonetlite/internal/EchonetDiscoveryService.java
```diff
@@ -0,0 +1,105 @@
...
+        }
+
+        final DiscoveryResult discoveryResult = DiscoveryResultBuilder
+                .create(new ThingUID(THING_TYPE_ECHONET_DEVICE, bridgeHandler.getThing().getUID(), identifier))
+                .withProperty("instanceKey", instanceKey.representationProperty())
+                .withProperty("hostname", instanceKey.address.getAddress().getHostAddress())
+                .withProperty("port", instanceKey.address.getPort())
+                .withProperty("groupCode", instanceKey.klass.groupCode())
+                .withProperty("classCode", instanceKey.klass.classCode()).withProperty("instance", instanceKey.instance)
+                .withBridge(bridgeHandler.getThing().getUID()).withRepresentationProperty("instanceKey").build();
```
jlaur, line 80: Consider adding constants for all these properties within `EchonetLiteBindingConstants` since they are used in multiple places. This will guarantee that they are exactly the same.

bundles/org.openhab.binding.echonetlite/src/main/java/org/openhab/binding/echonetlite/internal/EchonetDiscoveryService.java
```diff
@@ -0,0 +1,105 @@
...
+    }
+
+    @Override
+    protected void startScan() {
+        final EchonetLiteBridgeHandler bridgeHandler = this.bridgeHandler;
+        logger.debug("startScan: {}", bridgeHandler);
+        if (null != bridgeHandler) {
+            bridgeHandler.startDiscovery(this);
+        } else {
+            logger.error("bridgeHandler not initialized");
```
jlaur, line 53: Log level ERROR should only be used if something is very wrong and the binding will basically stop working. See https://next.openhab.org/docs/developer/guidelines.html#f-logging

bundles/org.openhab.binding.echonetlite/src/main/java/org/openhab/binding/echonetlite/internal/EchonetDevice.java
```diff
@@ -0,0 +1,199 @@
...
+    private final LinkedHashMap<Epc, State> pendingSets = new LinkedHashMap<>();
+    private final HashMap<Epc, State> stateFields = new HashMap<>();
+    private final HashMap<String, Epc> epcByChannelId = new HashMap<>();
+    private final Logger logger = LoggerFactory.getLogger(EchonetDevice.class);
+    @Nullable
+    private EchonetPropertyMap getPropertyMap;
+    private EchonetDeviceListener listener;
+    private boolean initialised = false;
+
+    private long lastPollMs = 0;
```
jlaur, line 44: Consider using `Instant` instead of milliseconds from clock to be less susceptible to changes in system clock.

bundles/org.openhab.binding.echonetlite/src/main/java/org/openhab/binding/echonetlite/internal/EchonetChannel.java
```diff
@@ -0,0 +1,103 @@
...
+        channel.register(selector, SelectionKey.OP_READ);
+    }
+
+    private boolean hasIpV4Address(final NetworkInterface networkInterface) {
+        return networkInterface.inetAddresses().anyMatch(ia -> ia instanceof Inet4Address);
+    }
+
+    public void close() {
+        try {
+            logger.info("closing selector");
```
jlaur, line 63: Please reduce log level here, see https://next.openhab.org/docs/developer/guidelines.html#f-logging.

bundles/org.openhab.binding.echonetlite/src/main/java/org/openhab/binding/echonetlite/internal/EchonetChannel.java
```diff
@@ -0,0 +1,103 @@
...
+            selector.close();
+            logger.info("closing channel");
```
jlaur, line 65: Please reduce log level here, see https://next.openhab.org/docs/developer/guidelines.html#f-logging.

bundles/org.openhab.binding.echonetlite/src/main/java/org/openhab/binding/echonetlite/internal/EchonetChannel.java
```diff
@@ -0,0 +1,103 @@
...
+            channel.close();
+        } catch (IOException e) {
+            logger.error("Failed to close selector/channel", e);
```
jlaur, line 68: Please reduce log level here, see https://next.openhab.org/docs/developer/guidelines.html#f-logging.

jlaur (changes requested):
Thanks for your contribution! The code looks pretty good. I have provided some minor feedback for you.

There are some compiler warnings to be addressed.

There are some checkstyle warnings left. You could take a look at target/code-analysis/report.html.

Please generate `echonetlite.properties` with missing i18n strings. See https://www.openhab.org/docs/developer/utils/i18n.html#generating-i18n-properties-file

bundles/org.openhab.binding.echonetlite/src/main/java/org/openhab/binding/echonetlite/internal/EchonetLiteBridgeHandler.java
mikeb01: I've moved this until after one successful poll cycle has been completed.

bundles/org.openhab.binding.echonetlite/src/main/java/org/openhab/binding/echonetlite/internal/EchonetDevice.java
mikeb01: I can understand the need to use a monotonic time source here, but I'm not sure that `Instant` provides that.  I'm also loath to use `Instant` do to the polling frequency and the amount of garbage it will create.  Instead I'm going to derive a timestamp from `System.nanoTime()` which is monotonic where supported (which will be most OSs).  The timestamp is only used for relative timing (i.e. timeouts), so a actual time (e.g. since epoch) is not required.

mikeb01:
I think I've addressed most of the review feedback.  The only place where I've differed from the recommended approach is around the use of the clock (comments above).

The outstanding compiler warnings I'm confused by.  The are being reported as being "Redundant null check", but as far as I can tell the values can be null.  The two in `EchonetDevice` are referring to a variable that has been retrieved via a `Hashmap::get` call and the other in `EchonetLiteBridgeHandler` is the result of a call to `ArrayBlockingQueue::poll`.  Both of these are quite clearly stated in the Javadoc that they can be null.  Is there a trick to get Spotbugs to recognise this?  I tried to annotate one of the variables with @Nullable, but that didn't make any difference.

bundles/org.openhab.binding.echonetlite/src/main/java/org/openhab/binding/echonetlite/internal/EchonetDevice.java
```diff
@@ -0,0 +1,199 @@
...
+                    epcByChannelId.put(epc.channelId(), epc);
+                }
+
+                final State pendingState = pendingSets.get(epc);
+                if (null != pendingState && pendingState.equals(state)) {
+                    logger.debug("pendingSet - removing: {} {}", epc, state);
+                    pendingSets.remove(epc);
+                } else if (null != pendingState) {
+                    logger.debug("pendingSet - state mismatch: {} {} {}", epc, pendingState, state);
+                }
```
jlaur, line 74: To get rid of the warnings:

`EchonetDevice.java:[69,29] Redundant null check: The variable pendingState cannot be null at this location
EchonetDevice.java:[72,36] Redundant null check: The variable pendingState cannot be null at this location`

```suggestion
                if (pendingSets.containsKey(epc)) {
                    final State pendingState = pendingSets.get(epc);
                    if (pendingState.equals(state)) {
                        logger.debug("pendingSet - removing: {} {}", epc, state);
                        pendingSets.remove(epc);
                    } else {
                        logger.debug("pendingSet - state mismatch: {} {} {}", epc, pendingState, state);
                    }
                }
```

I think the problem is that `LinkedHashMap.get()` will return `null` in two cases: When key is not found and when map contains `null` for given key.

jlaur (changes requested):
Quick feedback for the warnings.

bundles/org.openhab.binding.echonetlite/src/main/java/org/openhab/binding/echonetlite/internal/EchonetChannel.java
```diff
@@ -0,0 +1,107 @@
...
+    private final Selector selector = Selector.open();
+
+    private short tid = 0;
+
+    public EchonetChannel(InetSocketAddress discoveryAddress) throws IOException {
+        channel = DatagramChannel.open(StandardProtocolFamily.INET);
+        channel.bind(new InetSocketAddress("0.0.0.0", discoveryAddress.getPort()));
+        final Enumeration<NetworkInterface> networkInterfaces = NetworkInterface.getNetworkInterfaces();
+        while (networkInterfaces.hasMoreElements()) {
+            final NetworkInterface networkInterface = networkInterfaces.nextElement();
```
jlaur, line 52: To fix `[INFO] EchonetClass.java:[22,31] Unsafe null type conversion (type annotations): The value of type 'org.openhab.binding.echonetlite.internal.Epc.@NonNull Device @org.eclipse.jdt.annotation.NonNull[]' is made accessible using the less-annotated type 'org.openhab.binding.echonetlite.internal.Epc @org.eclipse.jdt.annotation.NonNull[]'`:
```suggestion
            final @Nullable NetworkInterface networkInterface = networkInterfaces.nextElement();
```

bundles/org.openhab.binding.echonetlite/src/main/java/org/openhab/binding/echonetlite/internal/EchonetLiteHandler.java
```diff
@@ -0,0 +1,185 @@
...
+            final InetSocketAddress address = new InetSocketAddress(requireNonNull(config.hostname), config.port);
+            final InstanceKey instanceKey = new InstanceKey(address,
+                    EchonetClass.resolve(config.groupCode, config.classCode), config.instance);
+            this.instanceKey = instanceKey;
+
+            updateProperty(PROPERTY_NAME_INSTANCE_KEY, instanceKey.representationProperty());
+            bridgeHandler.newDevice(instanceKey, config.pollIntervalMs, config.retryTimeoutMs, this);
+        } catch (Exception e) {
+            updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.CONFIGURATION_ERROR);
+            logger.error("Failed to configure {}", config, e);
```
jlaur, line 128:
```suggestion
            updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.CONFIGURATION_ERROR, e.getMessage());
```

bundles/org.openhab.binding.echonetlite/src/main/java/org/openhab/binding/echonetlite/internal/EchonetLiteBridgeHandler.java
```diff
@@ -0,0 +1,400 @@
...
+ * we can only have a single listener for echonet lite messages. Hence, using a bridge model to handle communications
+ * and discovery.
+ *
+ * @author Michael Barker - Initial contribution
+ */
+@NonNullByDefault
+public class EchonetLiteBridgeHandler extends BaseBridgeHandler {
+
+    private final Logger logger = LoggerFactory.getLogger(EchonetLiteBridgeHandler.class);
+    private final ArrayBlockingQueue<Message> requests = new ArrayBlockingQueue<>(1024);
```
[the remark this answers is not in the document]
mikeb01: I found an alternative.  Wrapping the call in a method that returns @Nullable works around the warning.  I'm not a fan of setting the generic type that way.  The idea that the Message can be null is a behaviour of the `poll` method not the type used in the queue.  Using `<@Nullable Message>` would also imply that we want to allow `offer(null)` as a valid operation, which I don't think is what we want.

bundles/org.openhab.binding.echonetlite/README.md
```diff
@@ -0,0 +1,87 @@
...
+* __classCode__: Class code for the device, see __groupCode__ for reference information.
+The value for Home Air Conditioners is '48' (0x30).
+* __instance__: Instance identifier if multiple instances are running on the same IP address.
+Typically, this value will be '1'.
+* __pollIntervalMs__: Interval between polls of the device for its current status.
+If multicast is not working this will determine the latency at which changes made directly on the device will be propagated back to openHAB, default is 30 000ms.
+* __retryTimeoutMs__: Length of time the bridge will wait before resubmitting a request, default is 2 000ms.
+
+Because the binding uses UDP, packets can be lost on the network, so retries are necessary.
+Testing has shown that 2000ms is a reasonable default that allows for timely retries without rejecting slow, but legitimate responses.
```
jlaur, line 40: Consistency: A few lines before, space was used as thousands separator (30 000ms and 2 000ms). Consider whether to use no thousands separator or same one.

bundles/org.openhab.binding.echonetlite/src/main/java/org/openhab/binding/echonetlite/internal/EchonetChannel.java
```diff
@@ -0,0 +1,107 @@
...
+    }
+
+    public void close() {
+        try {
+            logger.debug("closing selector");
+            selector.close();
+            logger.debug("closing channel");
+            channel.close();
+        } catch (IOException e) {
+            logger.warn("Failed to close selector/channel", e);
```
jlaur, line 72: Can this be silently ignored - if not, does it need to be warning?

bundles/org.openhab.binding.echonetlite/src/main/java/org/openhab/binding/echonetlite/internal/EchonetChannel.java
```diff
@@ -0,0 +1,107 @@
...
+                final SocketAddress address = channel.receive(buffer);
+
+                echonetMessage.sourceAddress(address);
+                buffer.flip();
+                long t0 = System.currentTimeMillis();
+                consumer.accept(echonetMessage);
+                long t1 = System.currentTimeMillis();
+                final long processingTimeMs = t1 - t0;
+                if (500 < processingTimeMs) {
+                    logger.info("Message took {}ms to process", processingTimeMs);
```
jlaur, line 100: This seems like trace or debug information?

bundles/org.openhab.binding.echonetlite/src/main/java/org/openhab/binding/echonetlite/internal/EchonetChannel.java
```diff
@@ -0,0 +1,107 @@
...
+                }
+            } catch (IOException e) {
+                logger.error("Unable to select", e);
```
jlaur, line 103: Can log level be reduced, or will binding stop working after this?

bundles/org.openhab.binding.echonetlite/src/main/java/org/openhab/binding/echonetlite/internal/EchonetDiscoveryService.java
```diff
@@ -0,0 +1,113 @@
...
+    }
+
+    @Override
+    protected void startScan() {
+        final EchonetLiteBridgeHandler bridgeHandler = this.bridgeHandler;
+        logger.debug("startScan: {}", bridgeHandler);
+        if (null != bridgeHandler) {
+            bridgeHandler.startDiscovery(this);
+        } else {
+            logger.warn("bridgeHandler not initialized");
```
jlaur, line 59: What is the consequence of this for the user? Does it need to be logged?

bundles/org.openhab.binding.echonetlite/src/main/java/org/openhab/binding/echonetlite/internal/EchonetDevice.java
mikeb01: This passes spotbugs, but unfortunately looks a bit weird as we've declared a variable `@Nullable` the immediately proceed with using it without a null check.  It also leads to a warning in Intellij.  Using a wrapper method similar to the queue case seems to be the cleaner approach.

bundles/org.openhab.binding.echonetlite/src/main/java/org/openhab/binding/echonetlite/internal/EchonetChannel.java
mikeb01: Again this makes the code look weird as we mark the variable as `@Nullable` and use it without a null check (and adding the null check generates the redundant check warning).  The cleanest solution I've come across is to just cast it to the same type, e.g.

```suggestion
            final NetworkInterface networkInterface = (NetworkInterface) networkInterfaces.nextElement();
```

Which due to the `@NonNullByDefault` is effectively `(@NonNull NetworkInterface)` which is enough to get Spotbugs to stop complaining.  This does come across as a redundant cast in Intellij, but I can live with that.  If necessary we can use the `//NoInspection RedundantCast`, but I've left that out for now.  I tried using `@SupressWarnings("RedundantCast")`, but Spotbugs didn't like that much either.

jlaur: Well, not exactly. This was not a compiler warning, just info - caused by missing null annotation of the `NetworkInterface` classes. By adding `@Nullable` in declaration we have compatible annotation on both sides. This doesn't mean we need to check for null, since probably it was analyzed that `nextElement()` cannot return `null` when `hasMoreElement()` returns `true`.

bundles/org.openhab.binding.echonetlite/src/main/java/org/openhab/binding/echonetlite/internal/EchonetDevice.java
```diff
@@ -0,0 +1,203 @@
...
+
+                if (initialised) {
+                    listener.onUpdated(epc.channelId(), state);
+                } else if (pendingGets.isEmpty()) {
+                    initialised = true;
+                    listener.onInitialised(identifier(), instanceKey, channelIds());
+                    stateFields.forEach((e, s) -> listener.onUpdated(e.channelId(), s));
+                }
+
```
jlaur, line 83: Last checkstyle warning: Remove this empty line.

bundles/org.openhab.binding.echonetlite/README.md
```diff
@@ -0,0 +1,87 @@
...
+| manufacturerCode                   | String    | Manufacturer Code                                             |
+| businessFacilityCode               | String    | Business Facility Code                                        |
+| powerSavingOperationSetting        | Switch    | Power Saving Operation Setting                                |
+| cumulativeOperatingTime            | Number    | Cumulative Operating Time                                     |
+| airFlowRate                        | String    | Air Flow Rate                                                 |
+| automaticControlOfAirFlowDirection | String    | Automatic Control Of Air Flow Direction                       |
+| automaticSwingOfAirFlow            | String    | Automatic Swing Of Air Flow                                   |
+| airFlowDirectionVertical           | String    | Air Flow Direction Vertical                                   |
+| airFlowDirectionHorizontal         | String    | Air Flow Direction Horizontal                                 |
+| operationMode                      | String    | Operation Mode                                                |
```
jlaur, line 67: Probably should be updated to same description channel now has.

bundles/org.openhab.binding.echonetlite/src/main/resources/OH-INF/thing/channel-types.xml
```diff
@@ -0,0 +1,378 @@
...
+				<option value="Rate 6">Rate 6</option>
+				<option value="Rate 7">Rate 7</option>
+				<option value="Rate 8">Rate 8</option>
+			</options>
+		</state>
+	</channel-type>
+
+	<channel-type id="automaticControlOfAirFlowDirection">
+		<item-type>String</item-type>
+		<label>Automatic Control Of Air Flow Direction</label>
```
jlaur, line 244: Please try to shorten this label (< 25 characters, 2-3 words). Description will still explain the details. Perhaps "Air Flow Direction"?

bundles/org.openhab.binding.echonetlite/src/main/resources/OH-INF/thing/channel-types.xml
```diff
@@ -0,0 +1,378 @@
...
+		<label>Power Saving Operation Setting</label>
+		<description>Power Saving Operation Setting</description>
+		<category>Switch</category>
+	</channel-type>
+
+	<channel-type id="cumulativeOperatingTime">
+		<item-type>Number</item-type>
+		<label>Cumulative Operating Time</label>
+		<description>Cumulative Operating Time</description>
+		<category>Number</category>
```
jlaur, line 219:
```suggestion
		<category>Time</category>
```

bundles/org.openhab.binding.echonetlite/src/main/resources/OH-INF/thing/channel-types.xml
```diff
@@ -0,0 +1,378 @@
...
+		<item-type>Switch</item-type>
+		<label>Power Saving Operation Setting</label>
+		<description>Power Saving Operation Setting</description>
+		<category>Switch</category>
+	</channel-type>
+
+	<channel-type id="cumulativeOperatingTime">
+		<item-type>Number</item-type>
+		<label>Cumulative Operating Time</label>
+		<description>Cumulative Operating Time</description>
```
jlaur, line 218: In which unit? Can this be mentioned in description?

bundles/org.openhab.binding.echonetlite/src/main/resources/OH-INF/thing/channel-types.xml
```diff
@@ -0,0 +1,378 @@
...
+		<item-type>String</item-type>
+		<label>Business Facility Code</label>
+		<description>Business Facility Code</description>
+		<category>Text</category>
+		<state readOnly="true"/>
+	</channel-type>
+
+	<channel-type id="powerSavingOperationSetting">
+		<item-type>Switch</item-type>
+		<label>Power Saving Operation Setting</label>
```
jlaur, line 210: Please try to shorten this label (< 25 characters, 2-3 words).

jlaur (changes requested):
Unfortunately I didn't find time to go through everything, but a few additional comments were added. I hope to be able to do a final round tomorrow evening. I reopened the UoM comment, this is probably the most important one to look into before merging.

bundles/org.openhab.binding.echonetlite/src/main/resources/OH-INF/thing/channel-types.xml
```diff
@@ -0,0 +1,378 @@
...
+				<option value="Cooling">Cooling</option>
+				<option value="Heating">Heating</option>
+				<option value="Dry">Dry</option>
+				<option value="Fan">Fan</option>
+				<option value="Other">Other</option>
+			</options>
+		</state>
+	</channel-type>
+
+	<channel-type id="setTemperature">
```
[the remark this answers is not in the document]
mikeb01: Target temperature is probably a better name, but Ideally I'd like to leave this as is.  As much as possible I've tried to match the channel ids and labels with the property names used in the EchonetLite specification, where they have use "set temperatue" as their convention rather than "target temperature".

bundles/org.openhab.binding.echonetlite/src/main/resources/OH-INF/thing/channel-types.xml
mikeb01: I've reduced it to "Automatic Air Flow Direction".  Unfortunately "Air Flow Direction" is not correct in this case as the channel controls whether the direction control is automated, rather that what the air flow direction is (which are a separate fields).

bundles/org.openhab.binding.echonetlite/src/main/resources/OH-INF/thing/channel-types.xml
mikeb01: I've updated this.  This should probably use UoM too.

bundles/org.openhab.binding.echonetlite/README.md
```diff
@@ -0,0 +1,87 @@
...
+| setTemperature                     | Number    | Set Temperature                                               |
```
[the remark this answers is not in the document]
mikeb01: I missed the Java implementation side of the UoM implementation.

bundles/org.openhab.binding.echonetlite/src/main/java/org/openhab/binding/echonetlite/internal/EchonetChannel.java
jlaur: I now realize that `nextElement()` will throw `NoSuchElementException` when there are no more elements, so it cannot return null. Thus, my comment is invalid and your approach is solid. :-)

bundles/org.openhab.binding.echonetlite/src/main/resources/OH-INF/thing/thing-types.xml
```diff
@@ -0,0 +1,80 @@
...
+		<representation-property>instanceKey</representation-property>
+
+		<config-description>
+			<parameter name="hostname" type="text" required="true">
+				<context>network-address</context>
+				<label>Hostname</label>
+				<description>Hostname or IP address of the device</description>
+			</parameter>
+			<parameter name="port" type="integer" required="true">
+				<context>network-port</context>
```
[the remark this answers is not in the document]
jlaur: `network-port` is still invalid.

bundles/org.openhab.binding.echonetlite/src/main/java/org/openhab/binding/echonetlite/internal/EchonetLiteBridgeHandler.java
```diff
@@ -0,0 +1,401 @@
...
+
+    private void start(final InstanceKey managementControllerKey, InstanceKey discoveryKey) throws IOException {
+        this.managementControllerKey = managementControllerKey;
+        this.discoveryKey = discoveryKey;
+
+        logger.debug("Binding echonet channel");
+        echonetChannel = new EchonetChannel(discoveryKey.address);
+        logger.debug("Starting networking thread");
+
+        networkingThread.setName("Echonet Networking");
```
jlaur, line 80:
```suggestion
        networkingThread.setName("OH-binding-" + EchonetLiteBindingConstants.BINDING_ID);
```

bundles/org.openhab.binding.echonetlite/src/main/java/org/openhab/binding/echonetlite/internal/EchonetLiteBridgeHandler.java
```diff
@@ -0,0 +1,401 @@
...
+        } catch (IOException e) {
+            logger.warn("Failed to poll for messages", e);
+        }
+    }
+
+    private void onMessage(final EchonetMessage echonetMessage) {
+        final EchonetClass echonetClass = echonetMessage.sourceClass();
+        final SocketAddress sourceAddress = echonetMessage.sourceAddress();
+        if (null == echonetClass) {
+            logger.warn("Unable to find echonetClass for message: {}", echonetMessage.toDebug());
```
jlaur, line 232: When users see this, should they report it as a bug, or what is needed to mitigate this? Should the bridge remain **ONLINE**?

bundles/org.openhab.binding.echonetlite/src/main/java/org/openhab/binding/echonetlite/internal/EchonetLiteBridgeHandler.java
```diff
@@ -0,0 +1,401 @@
...
+            return;
+        }
+
+        if (null == sourceAddress) {
+            logger.warn("Unable to find sourceAddress for message: {}", echonetMessage.toDebug());
```
jlaur, line 237: Ditto.

bundles/org.openhab.binding.echonetlite/src/main/java/org/openhab/binding/echonetlite/internal/EchonetLiteBridgeHandler.java
```diff
@@ -0,0 +1,401 @@
...
+    }
+
+    @Override
+    public void dispose() {
+        if (networkingThread.isAlive()) {
+            networkingThread.interrupt();
+            try {
+                networkingThread.join(TimeUnit.SECONDS.toMillis(5));
+            } catch (InterruptedException e) {
+                logger.warn("Interrupted while closing", e);
```
jlaur, line 308: Is this something not expected to happen, which should be reported as a bug?

bundles/org.openhab.binding.echonetlite/src/main/java/org/openhab/binding/echonetlite/internal/EchonetLiteBridgeHandler.java
```diff
@@ -0,0 +1,401 @@
...
+        }
+    }
+
+    private void poll() {
+        try {
+            doPoll();
+            updateStatus(ThingStatus.ONLINE);
+
+            while (!Thread.currentThread().isInterrupted()) {
+                doPoll();
```
jlaur, line 268: Should status be updated to **ONLINE** here as well, or is it no possible to go **OFFLINE** within this loop?

bundles/org.openhab.binding.echonetlite/src/main/java/org/openhab/binding/echonetlite/internal/EchonetLiteHandler.java
```diff
@@ -0,0 +1,184 @@
...
+    public EchonetLiteHandler(final Thing thing) {
+        super(thing);
+    }
+
+    @Nullable
+    private EchonetLiteBridgeHandler bridgeHandler() {
+        @Nullable
+        final Bridge bridge = getBridge();
+        if (null == bridge) {
+            logger.warn("Unable to handle commands if bridge is null");
```
jlaur, line 64: This method is also called by `initialize()` which has its own handling and will also log it. This logging should probably be moved to `handleCommand()`.

bundles/org.openhab.binding.echonetlite/src/main/java/org/openhab/binding/echonetlite/internal/EchonetLiteHandler.java
```diff
@@ -0,0 +1,184 @@
...
+            return null;
+        }
+
+        @Nullable
+        final EchonetLiteBridgeHandler handler = (EchonetLiteBridgeHandler) bridge.getHandler();
+        if (null == handler) {
+            logger.warn("Unable to handle commands if bridge handler is null");
```
jlaur, line 71: This method is also called by `initialize()` which has its own handling and will also log it. This logging should probably be moved to `handleCommand()`.

bundles/org.openhab.binding.echonetlite/src/main/java/org/openhab/binding/echonetlite/internal/StateCodec.java
```diff
@@ -0,0 +1,216 @@
...
+            final int value = edt.get(); // Should expand to typed value (mask excluded)
+            return new DecimalType(value);
+        }
+
+        public void encodeState(final State state, final ByteBuffer edt) {
+            edt.put((byte) (((DecimalType) state).intValue()));
+        }
+    }
+
+    enum Temperature8bitCodec implements StateCodec {
```
jlaur, line 199: Consider unit test coverage for this as well.

bundles/org.openhab.binding.echonetlite/src/main/test/org/openhab/binding/echonetlite/internal/protocol/StateCodecTest.java
```diff
@@ -0,0 +1,101 @@
+package org.openhab.binding.echonetlite.internal.protocol;
```
jlaur, line 1: In order to be discovered and run by Maven, this should be moved from:
- bundles/org.openhab.binding.echonetlite/src/main/test/org/openhab/binding/echonetlite/internal/protocol/ to:
- bundles/org.openhab.binding.echonetlite/src/test/java/org/openhab/binding/echonetlite/internal/protocol/

bundles/org.openhab.binding.echonetlite/src/main/test/org/openhab/binding/echonetlite/internal/protocol/LangUtilTest.java
```diff
@@ -0,0 +1,14 @@
+package org.openhab.binding.echonetlite.internal.protocol;
```
jlaur, line 1: In order to be discovered and run by Maven, this should be moved from:
- bundles/org.openhab.binding.echonetlite/src/main/test/org/openhab/binding/echonetlite/internal/protocol/ to:
- bundles/org.openhab.binding.echonetlite/src/test/java/org/openhab/binding/echonetlite/internal/protocol/

bundles/org.openhab.binding.echonetlite/src/main/test/org/openhab/binding/echonetlite/internal/EpcTest.java
```diff
@@ -0,0 +1,41 @@
+/**
```
jlaur, line 1: In order to be discovered and run by Maven, this should be moved from:
- bundles/org.openhab.binding.echonetlite/src/main/test/org/openhab/binding/echonetlite/internal/ to:
- bundles/org.openhab.binding.echonetlite/src/test/java/org/openhab/binding/echonetlite/internal/

bundles/org.openhab.binding.echonetlite/src/main/test/org/openhab/binding/echonetlite/internal/protocol/StateCodecTest.java
```diff
@@ -0,0 +1,101 @@
...
+        final OptionCodec optionCodec = new OptionCodec(new Option("ABC", 123), new Option("DEF", 101));
+        assertEncodeDecode(optionCodec, new StringType("ABC"), new byte[] { 123 });
+        assertEncodeDecode(optionCodec, new StringType("DEF"), new byte[] { 101 });
+    }
+
+    @Test
+    void shouldEncodeAndDecode8Bit() {
+        assertEncodeDecode(Decimal8bitCodec.INSTANCE, new DecimalType(123), new byte[] { 123 });
+        assertEncodeDecode(Decimal8bitCodec.INSTANCE, new DecimalType(1), new byte[] { 1 });
+        assertEncodeDecode(Decimal8bitCodec.INSTANCE, new DecimalType(-1), new byte[] { b(255) });
```
jlaur, line 99:
```suggestion
        assertEncodeDecode(StateCodec.Decimal8bitCodec.INSTANCE, new DecimalType(123), new byte[] { 123 });
        assertEncodeDecode(StateCodec.Decimal8bitCodec.INSTANCE, new DecimalType(1), new byte[] { 1 });
        assertEncodeDecode(StateCodec.Decimal8bitCodec.INSTANCE, new DecimalType(-1), new byte[] { b(255) });
```

bundles/org.openhab.binding.echonetlite/src/main/test/org/openhab/binding/echonetlite/internal/protocol/StateCodecTest.java
```diff
@@ -0,0 +1,101 @@
...
+        buffer.order(ByteOrder.BIG_ENDIAN);
+
+        final int valueInSeconds = 484260;
+        final long valueInMinutes = TimeUnit.SECONDS.toMinutes(valueInSeconds);
+        buffer.put(b(0x42));
+        buffer.putInt((int) valueInMinutes);
+
+        buffer.flip();
+        buffer.order(ByteOrder.LITTLE_ENDIAN);
+        assertEquals(valueInSeconds, ((DecimalType) OperatingTimeDecode.INSTANCE.decodeState(buffer)).intValue());
```
jlaur, line 81: After introducing UoM, I believe something like this is needed:
```suggestion
        assertEquals(new QuantityType<>(valueInSeconds, Units.SECOND),
                OperatingTimeDecode.INSTANCE.decodeState(buffer));
```

jlaur (changes requested):
Final round I think. :-)

bundles/org.openhab.binding.echonetlite/src/main/test/org/openhab/binding/echonetlite/internal/EpcTest.java
mikeb01: That was weird not sure how the tests ended up where they did.

jlaur:
@mikeb01 - there are a few remaining open comments. It happens to be the ones hidden by GitHub, so maybe you missed them. If you can go through the remaining `logger.warn` lines and assess if they need user action or not, I think we are close to merging this PR.

bundles/org.openhab.binding.echonetlite/src/main/java/org/openhab/binding/echonetlite/internal/EchonetLiteBridgeHandler.java
mikeb01: Yes this should be a bug.

bundles/org.openhab.binding.echonetlite/src/main/java/org/openhab/binding/echonetlite/internal/EchonetLiteBridgeHandler.java
mikeb01: For the bridge, no.  Once the socket is bound then the bridge is online.

bundles/org.openhab.binding.echonetlite/src/main/resources/OH-INF/thing/channel-types.xml
jlaur: TBH, I'm not sure if this is a rule of thumb rather than a hard limit. Let's close comment and hope it will fit in all UI's.

bundles/org.openhab.binding.echonetlite/src/main/java/org/openhab/binding/echonetlite/internal/EchonetLiteBridgeHandler.java
mikeb01: However, I think I need to go back and re-look at the online/offline implementation for devices.  It should be based on timeliness of response messages.

bundles/org.openhab.binding.echonetlite/src/main/resources/OH-INF/thing/thing-types.xml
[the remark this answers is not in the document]
jlaur: Please remember to check this comment also. :)

bundles/org.openhab.binding.echonetlite/src/main/java/org/openhab/binding/echonetlite/internal/EchonetLiteBridgeHandler.java
mikeb01: I've actually removed this check completely.

mikeb01:
I've added better support for updating the devices' thing status.  After 2 timeouts it will go offline and will go back online once data is received again.

jlaur:
@mikeb01 - can you check [this comment](https://github.com/openhab/openhab-addons/pull/11999#discussion_r977981204)? I believe this is the only remaining one.

jlaur (approved):
Thanks again for your contribution.

netwolfuk:
Congratulations @mikeb01 on the merge.

PR #12003, opened 2022-01-09
[mqtt] Add documentation about using auto discovery with Tasmota

cmorty:
I got most of this information from https://community.openhab.org/t/oh3-tasmota-via-mqtt-autodiscovery/111630.
IMO at least the gist of it should be in the documentation to help others.
Should I also link the forum? Did I get everything right?


lolodomo:
You are mentioning a specific device for a very generic binding, I am not sure that it is the right place.
We also have this other README for HomeAssistant binding which already mention some of your points:
https://github.com/openhab/openhab-addons/blob/main/bundles/org.openhab.binding.mqtt.homeassistant/README.md
Maybe you should rather update the other file ?

cmorty:
@lolodomo : You are absolutely right. - I'll get to it. :)

cmorty:
@lolodomo : Ok. Fixed it. I also did some minor formatting. It's in an extra commit so I can remove it in case it's too much. :-)

lolodomo (approved):
LGTM

lolodomo:
Can you try to provide a proper "signed-off" ?

PR #12005, opened 2022-01-09
[nuki] Support for SmartLock 3.0 and SmartDoor

janvyb:
This PR adds support for recently released products - Smart Lock 3.0 and Smart Door.

This is pretty simple PR since both new devices behave the same as Smart Lock 2.0, the only change needed is to allow devices with new deviceTypes (3, 4) to be recognized by discovery as smartlock.

Added new `deviceType` read only property (since now there are 3 smartlock deviceTypes) on smartlock thing because all actions on bridge require deviceType to be sent with request. This is just to be sure, there does not seem to be any difference sending deviceType 0, 3 or 4 (my smartlock 2.0 works the same with any of these values), this property seems to only exist to distinguish smartlock and opener calls.

I also updated documentation on `smartlockDoorState` channel to include new values introduced in v 1.13.1 of the bridge API.


bundles/org.openhab.binding.nuki/src/main/java/org/openhab/binding/nuki/internal/configuration/NukiDeviceConfiguration.java
```diff
@@ -22,4 +22,5 @@
 @NonNullByDefault
 public class NukiDeviceConfiguration {
     public String nukiId = "";
+    public Integer deviceType = 0;
```
lolodomo, line 25: Using simply `int` would have been sufficient and even better.

bundles/org.openhab.binding.nuki/src/main/java/org/openhab/binding/nuki/internal/discovery/NukiDeviceDiscoveryService.java
```diff
@@ -69,14 +70,22 @@ private DiscoveryResult createDiscoveryResult(BridgeApiListDeviceDto device, Nuk
                 .withRepresentationProperty(NukiBindingConstants.PROPERTY_NUKI_ID)
                 .withProperty(NukiBindingConstants.PROPERTY_NAME, device.getName())
                 .withProperty(NukiBindingConstants.PROPERTY_NUKI_ID, device.getNukiId())
+                .withProperty(NukiBindingConstants.PROPERTY_DEVICE_TYPE, device.getDeviceType())
                 .withProperty(NukiBindingConstants.PROPERTY_FIRMWARE_VERSION, device.getFirmwareVersion()).build();
     }
 
     private ThingUID getUid(String nukiId, int deviceType, NukiBridgeHandler bridgeHandler) {
```
lolodomo, line 77: Please add `throw IllegalArgumentException` on this method and handle the exception in the method `createDiscoveryResult`

bundles/org.openhab.binding.nuki/src/main/resources/OH-INF/thing/thing-types.xml
```diff
@@ -92,6 +92,23 @@
 				<label>Nuki ID</label>
 				<description>The decimal string that identifies the Nuki Smart Lock.</description>
 			</parameter>
+			<parameter name="deviceType" type="integer" required="true" readOnly="true">
```
lolodomo, line 95: Please remove `required="true"` : as you provide a default, this parameter is in fact not "required".
And at my knowledge, the attribute readOnly is not a valid attribute. To be removed.

bundles/org.openhab.binding.nuki/src/main/resources/OH-INF/thing/thing-types.xml
```diff
@@ -191,12 +208,14 @@
...
-				<option value="0">Unavailable</option>
 				<option value="1">Deactivated</option>
 				<option value="2">Closed</option>
 				<option value="3">Open</option>
-				<option value="4">Unknown</option>
+				<option value="4">Door state unknown</option>
 				<option value="5">Calibrating</option>
+				<option value="16">Uncalibrated</option>
+				<option value="240">Removed</option>
+				<option value="255">Unknown</option>
```
lolodomo, line 218: Please update the file nuki.properties to take into account these new options AND the new configuration parameter.
The easiest is to simply run: `mvn i18n:generate-default-translations` or you can do it manually.

lolodomo (changes requested):

bundles/org.openhab.binding.nuki/src/main/java/org/openhab/binding/nuki/internal/discovery/NukiDeviceDiscoveryService.java
janvyb: I'm not sure I follow. Where should I put the throw and why?

bundles/org.openhab.binding.nuki/src/main/resources/OH-INF/thing/thing-types.xml
janvyb: Readonly works, without it user can edit the parameter in main UI. The value is required, as in the smartlock cannot function without it being specified since it must be sent with http requests. Making the deviceType `int` somewhat solves it since it can't be set to null, but still seems better to explicitly tell in the config that the propery is mandatory.

bundles/org.openhab.binding.nuki/src/main/resources/OH-INF/thing/thing-types.xml
janvyb: What plugin is that? It's not configured on the project, i get 
```
[ERROR] No plugin found for prefix 'i18n' in the current project and in the plugin groups [org.apache.maven.plugins, org.codehaus.mojo]
```
when I run it.

bundles/org.openhab.binding.nuki/src/main/java/org/openhab/binding/nuki/internal/discovery/NukiDeviceDiscoveryService.java
lolodomo: `private ThingUID getUid(String nukiId, int deviceType, NukiBridgeHandler bridgeHandler) throw IllegalArgumentException  {`

Catch this exception in method createDiscoveryResult which is calling the method getUid.

bundles/org.openhab.binding.nuki/src/main/resources/OH-INF/thing/thing-types.xml
lolodomo: If you create manually your device in MainUI (not through the discovery), the user must have the ability to select the device type.

While you are setting a default value for your parameter, you will always have a value, that's why "required" is not required :)
You need this default value for backward compatibility.
If you want to keep the "required" attribute, I will not block your PR but it just makes no sense IMHO.

bundles/org.openhab.binding.nuki/src/main/resources/OH-INF/thing/thing-types.xml
lolodomo: As you have not too many changes to do, can you please do the changes manually.
Note that config parameters are generally sorted in alphabetic order.

bundles/org.openhab.binding.nuki/src/main/resources/OH-INF/thing/thing-types.xml
janvyb: You actually can't create the thing manually, only using discovery (or file configuration). Marking the property as required is for documentation purposes more than anything. The idea is that optional property with default value implies that it can be explicitly set to null (which doesn't really apply here since it'll be `int`, but still).

bundles/org.openhab.binding.nuki/src/main/java/org/openhab/binding/nuki/internal/discovery/NukiDeviceDiscoveryService.java
janvyb: Right I get it now. That case can "theoretically" never happen, since the device list is first filtered and only devices with supported types are passed to `createDiscoveryResult`, see
```
                        .filter(device -> NukiBindingConstants.DEVICE_DISCOVERY_SUPPORTED_DEVICES
                                .contains(device.getDeviceType()))
```
But thinking about it it's not really nice to have that check in two places, I'll drop the filter.

bundles/org.openhab.binding.nuki/src/main/resources/OH-INF/thing/thing-types.xml
lolodomo: How MainUI can prevent creation of the thing  ? 
And why not creating the thing manually if the user prefers this way for any reason?

janvyb: It's the ` listed="false"` property on thing-type, this way if you add new thing in the UI, you'll only get Nuki bridge in the list of things you can add manually.
This is because all communication is done through the bridge and the same API endpoint that is used for discovery is also used for updating thing status (there's a /list endpoint that returns list of all connected devices and their properties). That means if the thing is not returned from this endpoint, you can't discover it but and adding it manually would not work either.

bundles/org.openhab.binding.nuki/src/main/java/org/openhab/binding/nuki/internal/discovery/NukiDeviceDiscoveryService.java
janvyb: I've removed the exception since I want to keep the stream and exception handling does not play well with it. Is this ok?

bundles/org.openhab.binding.nuki/src/main/resources/OH-INF/thing/thing-types.xml
lolodomo: Yes, I know this property `listed="false"`. But I am not really sure it was a good idea to use it for the reason you explained. We have in openHAB many and many bindings having the same constraint and which do not use this property. In this case, the created thing is just OFFLINE with a message explaining why. Forcing the user to use the discovery to create a thing  is not the normal/expected behaviour.
At my knowledge, this property was more to be used for thing types which become obsolete but we keep them for compatibility reasons, using this property to avoid users creating new things of such a type.

lolodomo: That is interesting, I am very curious to know what are the bindings using this property.

lolodomo: It is used in more bindings than I was expected: 16 bindings including the Sonos binding. So I have to change my mind on that point.
For the Sonos binding for example, the reason is that it is almost impossible for a standard user to guess the id to setup when creating the thing. I think this should be the correct trigger to use this property.

janvyb: This is similar to how for example zigbee binding works (which also use `listed` property extensively). I also did it because getting the nukiId property is not very user friendly. You can find it buried deep in some settings menu of nuki app, but it's in hexadecimal format and the bridge api uses decimal format.

bundles/org.openhab.binding.nuki/src/main/java/org/openhab/binding/nuki/internal/discovery/NukiDeviceDiscoveryService.java
lolodomo: Ok for me.

lolodomo (approved):
LGTM

bundles/org.openhab.binding.nuki/src/main/resources/OH-INF/thing/thing-types.xml
lolodomo: Ok.
I learned something today, this `readOnly` attribute for thing parameters.

Issue #9578, opened 2020-12-29
[influxdb] OH3.0.0 Persisted values are restored from Influx without UoM


schmidmuc:
Situation:
- Running a OH3.0.0 installation under Ubuntu 20.04 LTS
- 
- I have UoM items in my configuration which get persisted to Influx DB
- After restart of OH, values of these items get restored, but without any units
- As a consequence there are many errors because of wrong item type (Number insted of Number with Unit), many rules don't work until a new value (including UoM) is written to the item
- In addition it's seems not to be transparent with which unit the values are persisted to Influx

Preferred behavior:
- Values are always persisted using the base unit value, so Byte instead of GigaByte, Watt insted of KiloWatt, etc.
- Values are restored after restart including this base unit

Alternative behavior:
- Define the preferred unit in the item configuration which should get used for persistence and as the default unit (if not defined explicitly in sitemaps, etc.)

My influxdb.persist:
`Strategies {
	everyHour : 	"0 0 * * * ?"
	default = everyChange
}
Items {
    gInflux* : strategy = everyChange, everyHour, restoreOnStartup
}
`

cweitkamp:
Related to https://github.com/openhab/openhab-core/issues/1954

PR #12010, opened 2022-01-09
[influxdb] [WIP] Introduce support for QuantityTypes 

lujop:
Introduce support to persist/restore quantity types as tracked in #9578 
Related to https://github.com/openhab/openhab-core/issues/1954

Current implementation persist values with unit defined in persistem item if present, and restores with same unit.

It's a work in progress, test needs to be done and also manual tests with openhab.
Still


lolodomo:
@lujop : are you still working on that PR ?

J-N-K:
Since there was quite some refactoring which created a lot of merge conflicts and there has not been any activity for half a year I would suggest to close this PR.

Issue #11952, opened 2022-01-03
[JS scripting engine] Memory Leak


egoist6:
As per digitaldan's request I re-open this ticket here which seems to be related to JSscripting.
When changing and testing JS scripts (edition 11) which have syntax errors, after 10-20 runs the CPU load increases to 100% and you get the following error messages:
`2021-12-29 22:12:59.868 [WARN ] [org.eclipse.jetty.server.HttpChannel] - /rest/rules/SystemReadLog java.lang.OutOfMemoryError: Java heap space`
`2021-12-29 22:14:35.128 [WARN ] [ab.core.internal.events.EventHandler] - Dispatching event to subscriber 'org.openhab.core.io.monitor.internal.EventLogger@16f740b' takes more than 5000ms.`
`2021-12-29 22:19:26.790 [WARN ] [mmon.WrappedScheduledExecutorService] - Scheduled runnable ended with an exception: java.lang.OutOfMemoryError: Java heap space`

Let me be more precise: at least the memory leak can be reproduced with faulty JS scripts - I am not saying that the script engine is the root cause of the poblem. I just wanted to bring this to your attention.

This problem has been confirmed by a couple of users yet.
See [here ](https://community.openhab.org/t/jsscripting-error-message-java-lang-outofmemoryerror-java-heap-space/130846)
and [here](https://community.openhab.org/t/javascript-memory-stability-issues/130810)

Environment
latest openhab 3.2 and Zulu11-32 version installed by latest openhabian update option
Raspberry Pi4

jpg0:
Just leaving a note here:

From a small investigation I could not see anything in openHAB code which appears to leak upon repeated script load failure.

One possible source of leaking could be from inside Graal itself, because the engines (and hence the polyglot contexts) are not being closed (`ScriptEngine.close()`).  The best place to put this is probably the `ScriptEngineManagerImpl` in core, as it's currently possible to repeatedly load one script which will overwrite the previous load without ever attempting to unload it (which is probably where it should be closed, if `AutoCloseable`).

jpg0:
I have verified that the leak is due to my guess. Well, I repeatedly reloading an erroring script and did managed to push up the memory (by about 25%, although I couldn't get it any higher), however my closing engines I could not increase it at all.

This requires changes in both core and the add-on. The core change is https://github.com/openhab/openhab-core/pull/2681

jpg0:
The other change required is: https://github.com/openhab/openhab-addons/pull/12022

PR #12022, opened 2022-01-11
[jsscripting] Wrapped GraalJS ScriptEngines now also Autocloseable

jpg0:
Allows wrapped GraalJS scripts to also implement `AutoCloseable` so that they can be cleaned up when no longer used.

With https://github.com/openhab/openhab-core/pull/2681, fixes https://github.com/openhab/openhab-addons/issues/11952.

Note that these changes can be applied independently without issue, however both are required to fix the above issue.


wborn (approved):
Thanks!

Issue #10275, opened 2021-03-03
[homekit] Temperature sensors, possibly thermostats not returning correct values


ecdye:
@yfre 

There seems to be an issue after the milestone 2 build where the temperature sensor only returns the maximum value regardless of the fact that the temperature in openHAB appears correctly. Here is an example log line it says this no matter what temperature it actually is, and then displays 100 ˚F in the home app.

```
2021-03-03 07:44:48.780 [WARN ] [istics.impl.base.FloatCharacteristic] - Detected value out of range 69.0. Returning max value instead. Characteristic io.github.hapjava.characteristics.impl.thermostat.CurrentTemperatureCharacteristic@c3ad04
```

yfre:
is maxValue set at item?  have you made it using mainUi or text configuration? what is the actual temperature and is "use Fahrenheit" set to true in homekit configuration?

ecdye:
> is maxValue set at item? 

No, I did not set it I just left it to default.

> have you made it using mainUi or text configuration?

mainUI

>  what is the actual temperature 

The actual temperature is 69, the log message always says `Detected value out of range [insert actual temperature]....`.

> is "use Fahrenheit" set to true in homekit configuration?

yes

yfre:
i think i know what it is doing. 
100 is the default max temperature in Celsius.  
in the recent version i have added support for Fahrenheit for custom max/min. probably I did something wrong for default values. it takes default 100 as F and converts it to C, which is 37 C. compares it with 69 and decides that 69 is above limit and sends 37 C = > 100 F. 

it is definitely a bug. 

could you please try one thing  - add custom maxValue  at item and check whether it works correctly with custom maxValue, looks like this in mainUI (but do it as currentTemperature)

![image](https://user-images.githubusercontent.com/11726616/109877190-fc8ac180-7c72-11eb-9fe0-2bec900a8af6.png)

ecdye:
If I manually set the min and max values to 0 and 100 respectively yes it works. Which means it will serve as a workaround for the moment until the bug is fixed

thales-man:
I’m also having issues with temperature sensors not reporting correctly. External sensors showing 1.9˚C are being translated to 50˚F. This clearly isn’t right. I’d hazard having monitored this temperature gauge that 50 is some sort of default floor value as it hasn’t changed for hours, even though it’s changed outside. I’m not sure what the purpose of having min / max values is quite frankly. It should just translate correctly the given value from one format to the next. 
[...]

PR #12024, opened 2022-01-11
[homekit] fix Fahrenheit conversion

### LINKED ISSUE #11957: [homekit] threshold temperature get converted from Fahrenheit to Fahrenheit
reported here
https://community.openhab.org/t/venstar-binding-and-homekit-setpoints-incorrectly-converting/131127

probably, binding should convert threshold temperature to C for HomeKit as well

yfre:
Apple HomeKit protocol  works internally with Celsius for temperature, so that we need to convert temperature in case openHAB is using Fahrenheit. this was already implemented for actual temperature values but not for temperature threshold and min/max values 

This PR add temperature conversion to thresholds and min/max values and fixes 
#11957
#10275

the existing methods convertFromCelsius and convertToCelsius are moved from abstract class to a factory so that they can be used from different places. 

these methods read binding setting "useFahrenheit" via OSGI service as settings could not be injected into factory or propagated to it easily. 

### DIFF
--- a/bundles/org.openhab.io.homekit/src/main/java/org/openhab/io/homekit/internal/accessories/HomekitTemperatureSensorImpl.java
+++ b/bundles/org.openhab.io.homekit/src/main/java/org/openhab/io/homekit/internal/accessories/HomekitTemperatureSensorImpl.java
@@ -25,6 +25,7 @@ import org.openhab.io.homekit.internal.HomekitTaggedItem;
 
 import io.github.hapjava.accessories.TemperatureSensorAccessory;
 import io.github.hapjava.characteristics.HomekitCharacteristicChangeCallback;
+import io.github.hapjava.characteristics.impl.thermostat.CurrentTemperatureCharacteristic;
 import io.github.hapjava.characteristics.impl.thermostat.TargetTemperatureCharacteristic;
 import io.github.hapjava.services.impl.TemperatureSensorService;
 
@@ -46,7 +47,8 @@ class HomekitTemperatureSensorImpl extends AbstractHomekitAccessoryImpl implemen
         final @Nullable DecimalType state = getStateAs(HomekitCharacteristicType.CURRENT_TEMPERATURE,
                 DecimalType.class);
         return CompletableFuture
-                .completedFuture(state != null ? convertToCelsius(state.doubleValue()) : getMinCurrentTemperature());
+                .completedFuture(state != null ? HomekitCharacteristicFactory.convertToCelsius(state.doubleValue())
+                        : getMinCurrentTemperature());
     }
 
     @Override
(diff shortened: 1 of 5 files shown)


bundles/org.openhab.io.homekit/src/main/java/org/openhab/io/homekit/internal/accessories/HomekitTemperatureSensorImpl.java
```diff
@@ -56,16 +58,23 @@ public void subscribeCurrentTemperature(HomekitCharacteristicChangeCallback call
...
     public double getMinCurrentTemperature() {
-        return convertToCelsius(
+        // Apple defines default values in Celsius. We need to convert them to Fahrenheit if openHAB is using Fahrenheit
+        // convertToCelsius and convertFromCelsius are only converting if useFahrenheit is set to true, so no additional
+        // check here needed
+        return HomekitCharacteristicFactory.convertToCelsius(
                 getAccessoryConfiguration(HomekitCharacteristicType.CURRENT_TEMPERATURE, HomekitTaggedItem.MIN_VALUE,
-                        BigDecimal.valueOf(TargetTemperatureCharacteristic.DEFAULT_MIN_VALUE)).doubleValue());
+                        BigDecimal.valueOf(HomekitCharacteristicFactory
+                                .convertFromCelsius(CurrentTemperatureCharacteristic.DEFAULT_MIN_VALUE)))
```
lolodomo, line 67: In several places, you replaced `TargetTemperatureCharacteristic.DEFAULT_MIN_VALUE` by `CurrentTemperatureCharacteristic.DEFAULT_MIN_VALUE`. Is it expected ?

bundles/org.openhab.io.homekit/src/main/java/org/openhab/io/homekit/internal/accessories/HomekitTemperatureSensorImpl.java
```diff
@@ -56,16 +58,23 @@ public void subscribeCurrentTemperature(HomekitCharacteristicChangeCallback call
...
     }
 
     @Override
     public double getMaxCurrentTemperature() {
-        return convertToCelsius(
+        return HomekitCharacteristicFactory.convertToCelsius(
                 getAccessoryConfiguration(HomekitCharacteristicType.CURRENT_TEMPERATURE, HomekitTaggedItem.MAX_VALUE,
-                        BigDecimal.valueOf(TargetTemperatureCharacteristic.DEFAULT_MAX_VALUE)).doubleValue());
+                        BigDecimal.valueOf(HomekitCharacteristicFactory
+                                .convertFromCelsius(CurrentTemperatureCharacteristic.DEFAULT_MAX_VALUE)))
```
lolodomo, line 76: In several places, you replaced `TargetTemperatureCharacteristic.DEFAULT_MAX_VALUE` by `CurrentTemperatureCharacteristic.DEFAULT_MAX_VALUE`. Is it expected ?

lolodomo (changes requested):

bundles/org.openhab.io.homekit/src/main/java/org/openhab/io/homekit/internal/accessories/HomekitTemperatureSensorImpl.java
yfre: @lolodomo good catch.  but it was wrong before and CurrentTemperatureCharacteristic.DEFAULT_MIN_VALUE is correct one. 
there are 2 characteristics - current and target temperature and both of them have default min/max values defined by apple. 
in that method the current temperature is created so we need to use min/max values for current temperature. 
probably it has little or no impact at the end, but the CurrentTemperatureCharacteristic.DEFAULT_MIN_VALUE is correct one.

i should have mentioned this fix in the PR description.

lolodomo (approved):
LGTM

PR #12028, opened 2022-01-11
[porcupineks] Keyword Spotter, initial contribution

GiviMAD:

This voice service allows you to use the PicoVoice product Porcupine as your keyword spotter in openHAB.

### DIFF
--- a/bundles/org.openhab.voice.porcupineks/README.md
+++ b/bundles/org.openhab.voice.porcupineks/README.md
@@ -2,6 +2,8 @@
 
 This voice service allows you to use the PicoVoice product Porcupine as your keyword spotter in openHAB.
 
+This add-on use a platform specific binary, should work on all the platforms supported by porcupine, if you encounter a problem try to run one of the porcupine java demos on your machine.
+
 ## How to use it
 
 After installing you would be able to access the addon options through the openHAB configuration page under the 'Other Services' section. There you will need to provide your PicoVoice Api Key.
@@ -20,6 +22,6 @@ As a summary service will only work if it's able find the correct ppn for your m
 
 This service currently supports english, german, french and spanish. But only the english model binary is included with the addon and will be used if the one for your configured language is not found under '\<openHAB userdata\>/.porcupine'. 
 
-To get the language model files go to the porcupine github repo. As an example for spanish the file is called 'porcupine_params_es.pv'
+To get the language model files go to the [porcupine repo](https://github.com/Picovoice/porcupine/tree/v2.0/lib/common).
 
-Note that the keyword model in use should match the language model.
+Note that the keyword model you use should match the language model.
--- a/bundles/org.openhab.voice.porcupineks/README.md
+++ b/bundles/org.openhab.voice.porcupineks/README.md
@@ -2,26 +2,36 @@
 
 This voice service allows you to use the PicoVoice product Porcupine as your keyword spotter in openHAB.
 
-This add-on use a platform specific binary, should work on all the platforms supported by porcupine, if you encounter a problem try to run one of the porcupine java demos on your machine.
+This add-on use a platform specific binary, it should work on all the platforms supported by porcupine,
+if you encounter a problem try to run one of the porcupine java demos on your machine.
 
 ## How to use it
 
-After installing you would be able to access the addon options through the openHAB configuration page under the 'Other Services' section. There you will need to provide your PicoVoice Api Key.
+After installing, you would be able to access the addon options 
+through the openHAB configuration page under the 'Other Services' section.
+There you will need to provide your PicoVoice Api Key.
 
 After that you can select porcupine as your default Keyword Spotter in your 'Voice' settings.
 
 ## Magic Word Configuration
 
-The magic word to spot is gathered from your 'Voice' configuration. The default english keyword models are loaded in the addon (also the english language model) so you can use those without adding anything else.
+The magic word to spot is gathered from your 'Voice' configuration. 
+The default english keyword models are loaded in the addon (also the english language model) 
+so you can use those without adding anything else.
 
-Note that you can use the pico voice platform to generate your own keyword models, for use them you should place the generated file under '\<openHAB userdata\>/.porcupine' and configure your magic world to match the file name replacing spaces with '_' and adding the extension '.ppn'.
+Note that you can use the pico voice platform to generate your own keyword models. 
+To use them, you should place the generated file under '\<openHAB userdata\>/.porcupine' and 
+configure your magic word to match the file name replacing spaces with '_' and adding the extension '.ppn'.
 
-As a summary service will only work if it's able find the correct ppn for your magic work configuration.
+As a summary, the service will only work if it's able to find the correct ppn for your magic word configuration.
 
 ## Language support
 
-This service currently supports english, german, french and spanish. But only the english model binary is included with the addon and will be used if the one for your configured language is not found under '\<openHAB userdata\>/.porcupine'. 
+This service currently supports English, German, French and Spanish. 
 
-To get the language model files go to the [porcupine repo](https://github.com/Picovoice/porcupine/tree/v2.0/lib/common).
+Only the English model binary is included with the addon and will be used 
+if the one for your configured language is not found under '\<openHAB userdata\>/.porcupine'. 
+
+To get the language model files, go to the [porcupine repo](https://github.com/Picovoice/porcupine/tree/v2.0/lib/common).
 
 Note that the keyword model you use should match the language model.
--- a/bundles/org.openhab.voice.porcupineks/README.md
+++ b/bundles/org.openhab.voice.porcupineks/README.md
@@ -2,8 +2,7 @@
 
 This voice service allows you to use the PicoVoice product Porcupine as your keyword spotter in openHAB.
 
-This add-on use a platform specific binary, it should work on all the platforms supported by porcupine,
-if you encounter a problem try to run one of the porcupine java demos on your machine.
+Porcupine provides on-device wake word detection powered by deep learning. This add-on should work on all the platforms supported by Porcupine, if you encounter a problem you can try to run one of the Porcupine java demos on your machine.
 
 ## How to use it
 
@@ -20,18 +19,38 @@ The default english keyword models are loaded in the addon (also the english lan
 so you can use those without adding anything else.
 
 Note that you can use the pico voice platform to generate your own keyword models. 
-To use them, you should place the generated file under '\<openHAB userdata\>/.porcupine' and 
+To use them, you should place the generated file under '\<openHAB userdata\>/porcupine' and 
 configure your magic word to match the file name replacing spaces with '_' and adding the extension '.ppn'.
 
 As a summary, the service will only work if it's able to find the correct ppn for your magic word configuration.
+
+#### Build-in keywords
+
+Remember that they only work with the English language model. (read bellow section)
+
+* alexa
+* americano
+* blueberry
+* bumblebee
+* computer
+* grapefruits
+* grasshopper
+* hey google
+* hey siri
+* jarvis
+* ok google
+* picovoice
+* porcupine
+* terminator
+
 
 ## Language support
 
 This service currently supports English, German, French and Spanish. 
 
 Only the English model binary is included with the addon and will be used 
-if the one for your configured language is not found under '\<openHAB userdata\>/.porcupine'. 
+if the one for your configured language is not found under '\<openHAB userdata\>/porcupine'. 
 
-To get the language model files, go to the [porcupine repo](https://github.com/Picovoice/porcupine/tree/v2.0/lib/common).
+To get the language model files, go to the [Porcupine repo](https://github.com/Picovoice/porcupine/tree/v2.0/lib/common).
 
 Note that the keyword model you use should match the language model.
--- a/bundles/org.openhab.voice.porcupineks/README.md
+++ b/bundles/org.openhab.voice.porcupineks/README.md
@@ -2,27 +2,30 @@
 
 This voice service allows you to use the PicoVoice product Porcupine as your keyword spotter in openHAB.
 
-Porcupine provides on-device wake word detection powered by deep learning. This add-on should work on all the platforms supported by Porcupine, if you encounter a problem you can try to run one of the Porcupine java demos on your machine.
+Porcupine provides on-device wake word detection powered by deep learning.
+This add-on should work on all the platforms supported by Porcupine, if you encounter a problem you can try to run one of the Porcupine java demos on your machine.
+
+Important: No voice data listened by this service will be uploaded to the Cloud.
+The voice data is processed offline, locally on your openHAB server by Porcupine.
+Once in a while, access keys are validated to stay active and this requires an Internet connection.
 
 ## How to use it
 
-After installing, you would be able to access the addon options 
-through the openHAB configuration page under the 'Other Services' section.
+After installing, you will be able to access the addon options through the openHAB configuration page under the 'Other Services' section.
 There you will need to provide your PicoVoice Api Key.
 
-After that you can select porcupine as your default Keyword Spotter in your 'Voice' settings.
+After that, you can select Porcupine as your default Keyword Spotter in your 'Voice' settings.
 
 ## Magic Word Configuration
 
 The magic word to spot is gathered from your 'Voice' configuration. 
-The default english keyword models are loaded in the addon (also the english language model) 
-so you can use those without adding anything else.
+The default english keyword models are loaded in the addon (also the english language model) so you can use those without adding anything else.
 
 Note that you can use the pico voice platform to generate your own keyword models. 
-To use them, you should place the generated file under '\<openHAB userdata\>/porcupine' and 
-configure your magic word to match the file name replacing spaces with '_' and adding the extension '.ppn'.
+To use them, you should place the generated file under '\<openHAB userdata\>/porcupine' and configure your magic word to match the file name replacing spaces with '_' and adding the extension '.ppn'.
+As an example, the file generated for the keyword "ok openhab" will be named 'ok_openhab.ppn'.
 
-As a summary, the service will only work if it's able to find the correct ppn for your magic word configuration.
+The service will only work if it's able to find the correct ppn for your magic word configuration.
 
 #### Build-in keywords
 
@@ -48,8 +51,7 @@ Remember that they only work with the English language model. (read bellow secti
 
 This service currently supports English, German, French and Spanish. 
 
-Only the English model binary is included with the addon and will be used 
-if the one for your configured language is not found under '\<openHAB userdata\>/porcupine'. 
+Only the English model binary is included with the addon and will be used if the one for your configured language is not found under '\<openHAB userdata\>/porcupine'.
 
 To get the language model files, go to the [Porcupine repo](https://github.com/Picovoice/porcupine/tree/v2.0/lib/common).
 
--- a/bundles/org.openhab.voice.porcupineks/README.md
+++ b/bundles/org.openhab.voice.porcupineks/README.md
@@ -1,0 +1,25 @@
+# Porcupine Keyword Spotter
+
+This voice service allows you to use the PicoVoice product Porcupine as your keyword spotter in openHAB.
+
+## How to use it
+
+After installing you would be able to access the addon options through the openHAB configuration page under the 'Other Services' section. There you will need to provide your PicoVoice Api Key.
+
+After that you can select porcupine as your default Keyword Spotter in your 'Voice' settings.
+
+## Magic Word Configuration
+
+The magic word to spot is gathered from your 'Voice' configuration. The default english keyword models are loaded in the addon (also the english language model) so you can use those without adding anything else.
+
+Note that you can use the pico voice platform to generate your own keyword models, for use them you should place the generated file under '\<openHAB userdata\>/.porcupine' and configure your magic world to match the file name replacing spaces with '_' and adding the extension '.ppn'.
+
+As a summary service will only work if it's able find the correct ppn for your magic work configuration.
+
+## Language support
+
+This service currently supports english, german, french and spanish. But only the english model binary is included with the addon and will be used if the one for your configured language is not found under '\<openHAB userdata\>/.porcupine'. 
+
+To get the language model files go to the porcupine github repo. As an example for spanish the file is called 'porcupine_params_es.pv'
+
+Note that the keyword model in use should match the language model.
--- a/bundles/org.openhab.voice.porcupineks/pom.xml
+++ b/bundles/org.openhab.voice.porcupineks/pom.xml
@@ -18,6 +18,7 @@
       <groupId>ai.picovoice</groupId>
       <artifactId>porcupine-java</artifactId>
       <version>2.0.2</version>
+      <scope>compile</scope>
     </dependency>
   </dependencies>
 </project>
--- a/bundles/org.openhab.voice.porcupineks/pom.xml
+++ b/bundles/org.openhab.voice.porcupineks/pom.xml
@@ -12,7 +12,7 @@
 
   <artifactId>org.openhab.voice.porcupineks</artifactId>
 
-  <name>openHAB Add-ons :: Bundles :: Porcupine Keyword Spotter</name>
+  <name>openHAB Add-ons :: Bundles :: Voice :: Porcupine Keyword Spotter</name>
   <dependencies>
     <dependency>
       <groupId>ai.picovoice</groupId>
--- a/bundles/org.openhab.voice.porcupineks/pom.xml
+++ b/bundles/org.openhab.voice.porcupineks/pom.xml
@@ -1,0 +1,23 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns="http://maven.apache.org/POM/4.0.0"
+  xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd">
+
+  <modelVersion>4.0.0</modelVersion>
+
+  <parent>
+    <groupId>org.openhab.addons.bundles</groupId>
+    <artifactId>org.openhab.addons.reactor.bundles</artifactId>
+    <version>3.3.0-SNAPSHOT</version>
+  </parent>
+
+  <artifactId>org.openhab.voice.porcupineks</artifactId>
+
+  <name>openHAB Add-ons :: Bundles :: Porcupine Keyword Spotter</name>
+  <dependencies>
+    <dependency>
+      <groupId>ai.picovoice</groupId>
+      <artifactId>porcupine-java</artifactId>
+      <version>2.0.2</version>
+    </dependency>
+  </dependencies>
+</project>
--- a/bundles/org.openhab.voice.porcupineks/src/main/java/org/openhab/voice/porcupineks/internal/PorcupineKSConfiguration.java
+++ b/bundles/org.openhab.voice.porcupineks/src/main/java/org/openhab/voice/porcupineks/internal/PorcupineKSConfiguration.java
@@ -1,0 +1,33 @@
+/**
+ * Copyright (c) 2010-2022 Contributors to the openHAB project
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package org.openhab.voice.porcupineks.internal;
+
+import org.eclipse.jdt.annotation.NonNullByDefault;
+
+/**
+ * The {@link PorcupineKSConfiguration} class contains fields mapping thing configuration parameters.
+ *
+ * @author Miguel Álvarez - Initial contribution
+ */
+@NonNullByDefault
+public class PorcupineKSConfiguration {
+
+    /**
+     * Sample configuration parameters. Replace with your own.
+     */
+    public String apiKey = "";
+    /**
+     * Sample configuration parameters. Replace with your own.
+     */
+    public float sensitivity = 0.5f;
+}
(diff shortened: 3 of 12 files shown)


bundles/org.openhab.voice.porcupineks/pom.xml
```diff
@@ -0,0 +1,23 @@
...
+
+  <parent>
+    <groupId>org.openhab.addons.bundles</groupId>
+    <artifactId>org.openhab.addons.reactor.bundles</artifactId>
+    <version>3.3.0-SNAPSHOT</version>
+  </parent>
+
+  <artifactId>org.openhab.voice.porcupineks</artifactId>
+
+  <name>openHAB Add-ons :: Bundles :: Porcupine Keyword Spotter</name>
```
lolodomo, line 15: Rather `<name>openHAB Add-ons :: Bundles :: Voice :: Porcupine Keyword Spotter</name>`

bundles/org.openhab.voice.porcupineks/src/main/java/org/openhab/voice/porcupineks/internal/PorcupineKSConfiguration.java
```diff
@@ -0,0 +1,33 @@
...
+public class PorcupineKSConfiguration {
+
+    /**
+     * Sample configuration parameters. Replace with your own.
+     */
+    public String apiKey = "";
+    /**
+     * Sample configuration parameters. Replace with your own.
+     */
+    public float sensitivity = 0.5f;
```
lolodomo, line 32: I am not sure the javadoc for each of your constants has any value here. Either you suppress it, or you provide a clearer description.

bundles/org.openhab.voice.porcupineks/pom.xml
```diff
@@ -0,0 +1,23 @@
...
+  <dependencies>
+    <dependency>
+      <groupId>ai.picovoice</groupId>
+      <artifactId>porcupine-java</artifactId>
+      <version>2.0.2</version>
+    </dependency>
+  </dependencies>
```
lolodomo, line 22: To be checked by someone if this requires something in the feature.xml file.

bundles/org.openhab.voice.porcupineks/README.md
```diff
@@ -0,0 +1,25 @@
+# Porcupine Keyword Spotter
+
+This voice service allows you to use the PicoVoice product Porcupine as your keyword spotter in openHAB.
+
+## How to use it
+
+After installing you would be able to access the addon options through the openHAB configuration page under the 'Other Services' section. There you will need to provide your PicoVoice Api Key.
```
lolodomo, line 9: General comment for all the README: add a return after sentence, it helps for maintenance.

After installing you **will** be ...

bundles/org.openhab.voice.porcupineks/README.md
```diff
@@ -0,0 +1,25 @@
...
+
+After that you can select porcupine as your default Keyword Spotter in your 'Voice' settings.
+
+## Magic Word Configuration
+
+The magic word to spot is gathered from your 'Voice' configuration. The default english keyword models are loaded in the addon (also the english language model) so you can use those without adding anything else.
+
+Note that you can use the pico voice platform to generate your own keyword models, for use them you should place the generated file under '\<openHAB userdata\>/.porcupine' and configure your magic world to match the file name replacing spaces with '_' and adding the extension '.ppn'.
```
lolodomo, line 17: ... to generate your own keyword models.
**To** use them, you ...

... and configure your magic **word** to match the ...

bundles/org.openhab.voice.porcupineks/README.md
```diff
@@ -0,0 +1,25 @@
...
+
+As a summary service will only work if it's able find the correct ppn for your magic work configuration.
```
lolodomo, line 19: As a summary**, the** service will only work if it's able **to** find the correct ppn for your magic **word** configuration.

bundles/org.openhab.voice.porcupineks/README.md
```diff
@@ -0,0 +1,25 @@
...
+
+## Language support
+
+This service currently supports english, german, french and spanish. But only the english model binary is included with the addon and will be used if the one for your configured language is not found under '\<openHAB userdata\>/.porcupine'. 
+
+To get the language model files go to the porcupine github repo. As an example for spanish the file is called 'porcupine_params_es.pv'
```
lolodomo: To get the language model files**,** go to the porcupine github repo.
As an example for spanish**,** the file is called 'porcupine_params_es.pv'**.**

bundles/org.openhab.voice.porcupineks/src/main/resources/OH-INF/config/config.xml
```diff
@@ -0,0 +1,25 @@
...
+	xmlns:config-description="https://openhab.org/schemas/config-description/v1.0.0"
+	xsi:schemaLocation="https://openhab.org/schemas/config-description/v1.0.0
+		https://openhab.org/schemas/config-description-1.0.0.xsd">
+	<config-description uri="voice:porcupineks">
+		<parameter-group name="authentication">
+			<label>Pico Voice Credentials</label>
+		</parameter-group>
+		<parameter-group name="spot">
+			<label>Spot</label>
+		</parameter-group>
```
lolodomo, line 13: Using groups when you have only one param in each group and only a total of 2 params might be not necessary ?

bundles/org.openhab.voice.porcupineks/src/main/resources/OH-INF/config/config.xml
```diff
@@ -0,0 +1,25 @@
...
+		<parameter name="apiKey" type="text" required="true" groupName="authentication">
+			<label>Pico Voice API Key</label>
+			<description>API Key for using the underling binary.</description>
```
lolodomo, line 16: Maybe this is not the best description for a final user ?

bundles/org.openhab.voice.porcupineks/src/main/resources/OH-INF/config/config.xml
```diff
@@ -0,0 +1,25 @@
...
+		</parameter>
+		<parameter name="sensitivity" type="decimal" min="0" max="1" required="true" groupName="spot">
```
lolodomo, line 18: As you provide a default value, you should not set it as required.
I have a doubt with "decimal" type, did you check that ?

lolodomo (changes requested):
Review part 1 of 2

bundles/org.openhab.voice.porcupineks/src/main/java/org/openhab/voice/porcupineks/internal/PorcupineKSConfiguration.java
GiviMAD: Sorry I forgot to write those, ty for pointing at it.

bundles/org.openhab.voice.porcupineks/src/main/resources/OH-INF/config/config.xml
GiviMAD: Yes decimal type seems to work fine. I can change the value, it is correctly accessed from the binding, and also the validations for min and max work fine on the ui.

lolodomo: Verified: decimal is a valid value.

bundles/org.openhab.voice.porcupineks/src/main/java/org/openhab/voice/porcupineks/internal/PorcupineKSService.java
```diff
@@ -0,0 +1,283 @@
...
+        captureBuffer.order(ByteOrder.LITTLE_ENDIAN);
+        short[] porcupineBuffer = new short[frameLength];
+        this.loop = true;
+        while (loop) {
+            try {
+                // read a buffer of audio
+                numBytesRead = audioStream.read(captureBuffer.array(), 0, captureBuffer.capacity());
+                // don't pass to porcupine if we don't have a full buffer
+                if (numBytesRead != frameLength * 2) {
+                    continue;
```
lolodomo, line 267: Question: What if the read returned only part of the buffer ? You are ignoring it.

lolodomo (changes requested):
Review part 2 of 2

lolodomo:
Please also run `mvn i18:generate-default-translations` to produce the default translation properties.

lolodomo:
Is the underlying library compatible with Java 11 ?

bundles/org.openhab.voice.porcupineks/src/main/java/org/openhab/voice/porcupineks/internal/PorcupineKSService.java
```diff
@@ -0,0 +1,283 @@
...
+        if (!childDirectory.exists()) {
+            if (childDirectory.mkdir()) {
+                logger.info("porcupine extraction file dir created {}", EXTRACTION_FOLDER);
+            }
+        }
+    }
+
+    @Activate
+    protected void activate(ComponentContext componentContext, Map<String, Object> config) {
+        var serviceConfig = new org.openhab.core.config.core.Configuration(config).as(PorcupineKSConfiguration.class);
```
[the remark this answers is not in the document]
J-N-K: IMO `var` is the worst addition to Java ever made and should be forbidden.

lolodomo:
Maybe you could explain (in documentation) if the underlying service is in the cloud or purely locall. Are there audio data sent in the cloud. This is something most of users are very sensible.
If I correctly understand, there are few files automatically downloaded from the WEB when initializing the service, but then are the audio data sent in the cloud to be analyzed ?

bundles/org.openhab.voice.porcupineks/pom.xml
lolodomo: Reading documentation here https://www.openhab.org/docs/developer/buildsystem.html
I understand that in your case you have nothing to add in your feature.xml file (so your feature.xml is fine) but I also understand that you should add `<scope>compile</scope>` in your POM file.

@wborn or @J-N-K : please confirm my assumptions.

J-N-K: Correct. It‘ll probably work without the scope since `compile` is the default scope, but should be added for clarity. Nothing needs to be added to the feature if the dependency is embedded.

lolodomo:
And please clarify in the documentation if the service supports by default any keyword that is provided in the openHAB configuration, or if only a particular (English) keyword is supported by default.

bundles/org.openhab.voice.porcupineks/pom.xml
lolodomo: Thank you for the explanation.

GiviMAD: Thank you, I'll add the compile scope for clarity.

bundles/org.openhab.voice.porcupineks/src/main/java/org/openhab/voice/porcupineks/internal/PorcupineKSService.java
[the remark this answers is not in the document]
GiviMAD: As I work as a NodeJS developer I'm used to living in the dark but I understand your hate.
I was sticking to not use var for primitive types or when the property name do not contains the class name, but I can use it less, let me know.
I'm going to change this line for now. I think the reason for using org.openhab.core.config.core.Configuration was that there is another Configuration class in the scope but I'll check.

bundles/org.openhab.voice.porcupineks/src/main/java/org/openhab/voice/porcupineks/internal/PorcupineKSService.java
GiviMAD: That read is calling a native method so I think is always blocking so I maybe no need for the thread sleep, but I'm not totally sure.
Those line are almost an exactly copy of the porcupine java sample, I even left same comment. I can not see the binary implementation, I also think there is no change of read not returning the full buffer, as least on an infinite stream.

bundles/org.openhab.voice.porcupineks/src/main/java/org/openhab/voice/porcupineks/internal/PorcupineKSService.java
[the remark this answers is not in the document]
lolodomo: Please replace all var type, this will take you 2 minutes.

bundles/org.openhab.voice.porcupineks/src/main/java/org/openhab/voice/porcupineks/internal/PorcupineKSService.java
```diff
@@ -0,0 +1,283 @@
...
+@NonNullByDefault
+@Component(configurationPid = SERVICE_PID, property = Constants.SERVICE_PID + "=" + SERVICE_PID)
+@ConfigurableService(category = SERVICE_CATEGORY, label = SERVICE_NAME, description_uri = SERVICE_CATEGORY + ":"
+        + SERVICE_ID)
+public class PorcupineKSService implements KSService {
+    private final Logger logger = LoggerFactory.getLogger(PorcupineKSService.class);
+    private static final String PORCUPINE_FOLDER = Path.of(OpenHAB.getUserDataFolder(), ".porcupine").toString();
+    private static final String EXTRACTION_FOLDER = Path.of(OpenHAB.getUserDataFolder(), ".porcupine", "extracted")
+            .toString();
+    private final ScheduledExecutorService executor = ThreadPoolManager.getScheduledPool("audio");
```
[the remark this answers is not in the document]
GiviMAD: I did the same on both PRs in order to share the same one, is it better that I change it to "porcupine"?

bundles/org.openhab.voice.porcupineks/src/main/java/org/openhab/voice/porcupineks/internal/PorcupineKSService.java
```diff
@@ -0,0 +1,283 @@
...
+                logger.info("porcupine dir created {}", PORCUPINE_FOLDER);
+            }
+        }
+        File childDirectory = new File(EXTRACTION_FOLDER);
+        if (!childDirectory.exists()) {
+            if (childDirectory.mkdir()) {
+                logger.info("porcupine extraction file dir created {}", EXTRACTION_FOLDER);
+            }
+        }
+    }
```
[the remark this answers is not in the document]
GiviMAD: I just checked the android debug bridge binding where I did something similar for the keys folder and I did it this way. But I can change it to the constructor, let me know.

bundles/org.openhab.voice.porcupineks/src/main/java/org/openhab/voice/porcupineks/internal/PorcupineKSService.java
```diff
@@ -0,0 +1,283 @@
...
+        this.config = serviceConfig;
+        this.bundleContext = componentContext.getBundleContext();
+        if (serviceConfig.apiKey.isBlank()) {
+            logger.warn("Missing pico voice api key to use Porcupine Keyword Spotter");
+        }
+    }
+
+    private String prepareLib(BundleContext bundleContext, String path) throws IOException {
+        var relativePath = path.substring(path.indexOf("porcupine/"));
+        var porcupineResource = bundleContext.getBundle().getEntry(relativePath);
```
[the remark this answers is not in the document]
GiviMAD: This method extract the binary files from the bundle to the extracted folder and returns an absolute path to the extracted file, because the underlying binary need the file to exists in order to work.
As a summary I'm relying on the internal porcupine utils to get the correct library path, but it resolves paths that do not exists as files are inside the bundle, this method use those path to locale a file inside the bundle, extract it, and return an usable path.
Hope I explained it correctly, let me know if not.

bundles/org.openhab.voice.porcupineks/src/main/java/org/openhab/voice/porcupineks/internal/PorcupineKSService.java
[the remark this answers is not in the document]
lolodomo: In case you need to create a new thread pool, you have to respect a certain naming convention. I will provide more information later. 

Sharing the same thread pool between several bindings, I am not sure this is a good idea, except of course if the pool is a common pool from the core framework. We should check if a thread pool already exists for audio/voice.

bundles/org.openhab.voice.porcupineks/src/main/java/org/openhab/voice/porcupineks/internal/PorcupineKSService.java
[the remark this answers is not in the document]
lolodomo: Keep it as it is.

bundles/org.openhab.voice.porcupineks/src/main/java/org/openhab/voice/porcupineks/internal/PorcupineKSService.java
```diff
@@ -0,0 +1,283 @@
...
+    }
+
+    @Override
+    public String getLabel(@Nullable Locale locale) {
+        return SERVICE_NAME;
+    }
+
+    @Override
+    public Set<Locale> getSupportedLocales() {
+        return Set.of(Locale.ENGLISH, new Locale("es"), Locale.FRANCE, Locale.GERMAN);
```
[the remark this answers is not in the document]
GiviMAD: No, there is not default constant for ES :(

bundles/org.openhab.voice.porcupineks/src/main/java/org/openhab/voice/porcupineks/internal/PorcupineKSService.java
lolodomo: Ok, fine.

bundles/org.openhab.voice.porcupineks/src/main/java/org/openhab/voice/porcupineks/internal/PorcupineKSService.java
```diff
@@ -0,0 +1,283 @@
...
+            porcupine.delete();
+            this.porcupine = null;
+        };
+    }
+
+    private Porcupine initPorcupine(BundleContext bundleContext, Locale locale, String keyword)
+            throws IOException, PorcupineException {
+        // Suppress library logs
+        java.util.logging.Logger.getLogger(java.util.logging.Logger.GLOBAL_LOGGER_NAME)
+                .setLevel(java.util.logging.Level.OFF);
```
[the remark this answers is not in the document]
GiviMAD: I think no because we don't use the java logger, but I will reenable it after initialization that will be safer.

bundles/org.openhab.voice.porcupineks/src/main/java/org/openhab/voice/porcupineks/internal/PorcupineKSService.java
GiviMAD: Finally I have added a sleep before the continue clause. Is this enough?

GiviMAD:
> Is the underlying library compatible with Java 11 ?

Yes, it targets java 11

GiviMAD:
> Maybe you could explain (in documentation) if the underlying service is in the cloud or purely local. Are there audio data sent in the cloud. This is something most of users are very sensible.
> If I correctly understand, there are few files automatically downloaded from the WEB when initializing the service, but then are the audio data sent in the cloud to be analyzed ?

I will clarify this at the beginning of the read me. 
The processing is purely local and it doesn't download anything it just extract the binaries included within the porcupine java SDK which are the library binary (all architectures supported are included) the language model binary (only English one is included) and the default keywords (only for English). I don't know if it does any call to validate the pico voice api key but I hope not, I will try to run it with the network off.

bundles/org.openhab.voice.porcupineks/src/main/java/org/openhab/voice/porcupineks/internal/PorcupineKSService.java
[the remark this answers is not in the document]
lolodomo: Using the common thread pool would not be a good idea as it will consume one permanent thread for each started dialog.

Our thread naming convention was for bindings, not for services. It is `OH-binding-<thing UID>[-<differentiator>]`. So I suggesting the following name for your thread pool: `OH-voice-porcupineks`.

bundles/org.openhab.voice.porcupineks/src/main/java/org/openhab/voice/porcupineks/internal/PorcupineKSService.java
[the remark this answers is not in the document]
lolodomo: It could become clearer for me if you tell what are the values for `Porcupine.LIBRARY_PATH` and `Porcupine.MODEL_PATH`.

lolodomo:
I suppose the service will work well even if several dialogs are started in parallel (of course on different audio sources) ?

GiviMAD:
> I suppose the service will work well even if several dialogs are started in parallel (of course on different audio sources) ?

As far as I've tried it yes, seems to be thread save.

Another think I realize is that they track devices using the access key and limit those to three, so there should be a call to the service in some point to activate the device and validate the access key. I don't know why I can launch it without internet maybe the binary caches the activation in some place. I couldn't find information about it.

bundles/org.openhab.voice.porcupineks/src/main/java/org/openhab/voice/porcupineks/internal/PorcupineKSService.java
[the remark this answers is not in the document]
GiviMAD: I have reviewed other voice bindings and none of them seems to use an executor. What should I do with this one?

GiviMAD:
> Please also run `mvn i18:generate-default-translations` to produce the default translation properties.

This is giving me errors, I have tried from the root with `mvn i18:generate-default-translations -pl :org.openhab.voice.porcupineks` and from the project folder, both return same error.

```
[ERROR] No plugin found for prefix 'i18' in the current project and in the plugin groups [org.apache.maven.plugins, org.codehaus.mojo] available from the repositories [local (/Users/miguelgivi/.m2/repository), central (https://repo1.maven.org/maven2), openhab-release (https://openhab.jfrog.io/openhab/libs-release)] -> [Help 1]
[E
```

GiviMAD:
Here is some info from pico voice page.
```
We track the number of users via access keys that are used to activate our products. Once in a while, access keys are validated by our servers to stay active. (There is no voice data or anything other information is sent to our cloud. The voice data is processed offline. It’s your data, you own and control it.)
```

lolodomo:
I submitted a PR to your repo with the default translation properties.

lolodomo:
I will have a final look this evening. 
Can you please answer my question about the values of `Porcupine.LIBRARY_PATH` and `Porcupine.MODEL_PATH` ?

GiviMAD:
> I will have a final look this evening.
> Can you please answer my question about the values of `Porcupine.LIBRARY_PATH` and `Porcupine.MODEL_PATH` ?

Sorry, I missed that question, those are the path to the porcupine build-in files, model path is constant but library path is resolved by architecture. I though it throws an exception if is unable to resolve the library path but no, it returns null, so I've pr a small change.

lolodomo:
Ok but can you show me the value you get for the lib path for example.

GiviMAD:
> Ok but can you show me the value you get for the lib path for example.

Sure, I'll send it to you in the afternoon

GiviMAD:
I've made some changes to that method to clarify what is doing.

This is a sample of the resolved LIBRARY_PATH on my machine: /Users/miguelgivi/Downloads/openhab-3/userdata/cache/org.eclipse.osgi/239/26/bundleFile/porcupine/lib/java/mac/x86_64/libpv_porcupine_jni.dylib

bundles/org.openhab.voice.porcupineks/src/main/java/org/openhab/voice/porcupineks/internal/PorcupineKSService.java
```diff
@@ -0,0 +1,318 @@
...
+        logger.debug("Porcupine model path: {}", modelPath);
+        logger.debug("Porcupine keyword path: {}", keywordPath);
+        logger.debug("Porcupine sensitivity: {}", config.sensitivity);
+        try {
+            return new Porcupine(config.apiKey, libraryPath, modelPath, new String[] { keywordPath },
+                    new float[] { config.sensitivity });
+        } finally {
+            // restore log level
+            globalJavaLogger.setLevel(currentGlobalLogLevel);
+        }
```
lolodomo, line 217: Your idea is to restore the log level when the object will be destroyed ?
I have a big doubt here (I am not very familiar with this use case), maybe the code will be run just after the return ?

lolodomo (changes requested):
Final review

bundles/org.openhab.voice.porcupineks/src/main/java/org/openhab/voice/porcupineks/internal/PorcupineKSService.java
```diff
@@ -0,0 +1,318 @@
...
+            porcupine = initPorcupine(bundleContext, locale, keyword);
+        } catch (PorcupineException | IOException e) {
+            throw new KSException(e);
+        }
+        Future<?> scheduledTask = executor.submit(() -> processInBackground(porcupine, ksListener, audioStream));
+        return () -> {
+            loop = false;
+            scheduledTask.cancel(true);
+            porcupine.delete();
+        };
```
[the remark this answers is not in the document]
lolodomo: Wait a second, in fact, I think you could just replace:
```
        return () -> {
            loop = false;
            scheduledTask.cancel(true);
            porcupine.delete();
        };
```
by 
```
        return new KSServiceHandle() {
                @Override
                public void abort() {
                    loop = false;
                    scheduledTask.cancel(true);
                    porcupine.delete();
                }
        };
```
And that's all.

lolodomo:
@GiviMAD ; can you finalize this one?

bundles/org.openhab.voice.porcupineks/src/main/java/org/openhab/voice/porcupineks/internal/PorcupineKSService.java
GiviMAD: No, the finally clause get executed before the function returns the value, just after the call to the Porcupine constructor.
So the log is only disabled for the execution of this method.

lolodomo: Ok, I did not understand this was what er are trying to achieve.

bundles/org.openhab.voice.porcupineks/src/main/java/org/openhab/voice/porcupineks/internal/PorcupineKSService.java
```diff
@@ -0,0 +1,329 @@
...
+            throw new KSException(e);
+        }
+        Future<?> scheduledTask = executor.submit(() -> processInBackground(porcupine, ksListener, audioStream));
+        return new KSServiceHandle() {
+            @Override
+            public void abort() {
+                logger.debug("stopping service");
+                loop = false;
+                try {
+                    Thread.sleep(0);
```
lolodomo, line 189: Why addindg a sleep  especially with a 0 value?

lolodomo (commented):
Sorry, I have a last question.

bundles/org.openhab.voice.porcupineks/src/main/java/org/openhab/voice/porcupineks/internal/PorcupineKSService.java
lolodomo: Why not in this case simply do it like that (which looks clearer):
```
Porcupine p = new Porcupine(config.apiKey, libraryPath, modelPath, new String[] { keywordPath },
                    new float[] { config.sensitivity });
// restore log level
globalJavaLogger.setLevel(currentGlobalLogLevel);
return p;
```

jlaur: @lolodomo - if `Porcupine` constructor would throw an unchecked exception for any reason, log level would not be restored. Using `finally` is more robust as it will always be executed, even if an exception is thrown.

lolodomo: Ok, you're right.

bundles/org.openhab.voice.porcupineks/src/main/java/org/openhab/voice/porcupineks/internal/PorcupineKSService.java
GiviMAD: Sorry for the delay, I woke with fever yesterday and I was resting all day.
I want the background task to end by itself before calling close, in my tests sleeping by 0 is enough for it to complete. Would be better to add a small interval like 100ms?

lolodomo: Ok, I understand, that is not a bad idea.
I am not an expert with sleep(0) but just googling, it looks like it is a very special thing.
If you are sure that iit is an appropriate usage, tell me. If not, I would prefer a value > 0, like 100ms.

lolodomo (approved):
LGTM

lolodomo:
@GiviMAD : I realized that the documentation is not at the right level compared to all other voice services. You would be kind to create a PR to enhance it with a chapter "Obtaining credentials" and a chapter "Configuration". The best would be to do a copy/paste form the current Google TTS service documentation and then adjust it.

lolodomo:
> @GiviMAD : I realized that the documentation is not at the right level compared to all other voice services. You would be kind to create a PR to enhance it with a chapter "Obtaining credentials" and a chapter "Configuration". And explaining where to select the default keyword spotter (and define the default keyword), like in the chapter "Voice configuration" for the Google TTS service. The best would be to do a copy/paste form the current Google TTS service documentation and then adjust it.

I created PR #12116 for that.
What will be still missing is a chapter to explain how to get the API key.

lolodomo:
@GiviMAD : there is something I missed during my review: you should mention your dependency in the NOTICE file. Please open a PR to update this file.

lolodomo:
Found license for SDK: Apache License 2.0

Note that there is now a 2.1 version. You might check if it works well and then update the POM file to use it.

Issue #11483, opened 2021-10-31
[Tesla] Tesla add "set_charging_amps" from 2021.36


LuggyLuke:
Please add the "set_charging_amps" channel added in version 2021.36.
It enables the setting of the actual charging current via a command. 
I think this feature could be quite handy for all Solar users combining it with a Tesla, that way the chargin current can be adapted to the current Solar power, instead of starting/stopping the process based on thresholds. 
for further info i found a discussion on the teslamotorclub forum: 

https://teslamotorsclub.com/tmc/threads/model-s-rest-api.13410/page-148 

Thanks & greetings

ipweb2000:
+1. that would be great and an easier and more general way than via TWCManager

phuongpham:
I'm happy to test out this "feature"

DollC:
+1 I am also very interested to have this feature. If you need someone to test. I can do

DollC:
Thanks. I just installed the snapshot of 3.3.0 and it works. I can set chargingamps an see it directly in the app of my model 3

phuongpham:
Ok I managed to install the  binding 
321 │ Active │  80 │ 3.3.0.202201130406    │ openHAB Add-ons :: Bundles :: Tesla Binding

But I can't see the new channels.
I have tried to delete the thing and added again.
But same results.

<img width="682" alt="image" src="https://user-images.githubusercontent.com/2634344/149614698-694bf018-6175-4ee7-b4ef-10cf0caff315.png">

phuongpham:
Ok.
I managed to get this working. Seems like the previous binding was wrong version.
I installed follow version  3.3.0.202201150408 
I downloaded from https://openhab.jfrog.io/artifactory/libs-snapshot-local/org/openhab/addons/bundles/org.openhab.binding.tesla/3.3.0-SNAPSHOT/org.openhab.binding.tesla-3.3.0-20220115.032535-24.jar

ipweb2000:
Hey guys. I am having trouble to get the snapshot binding to run. Any suggestions? Thanks for your help!

> 12:36:27.715 [WARN ] [org.apache.felix.fileinstall         ] - Error while starting bundle: file:/usr/share/openhab/addons/org.openhab.binding.tesla-3.3.0-20220115.032535-24.jar
> org.osgi.framework.BundleException: Could not resolve module: org.openhab.binding.tesla [291]
[...]

PR #12029, opened 2022-01-11
[tesla] Add support for setting charging amps and sentry mode

kaikreuzer:
Implements https://github.com/openhab/openhab-addons/issues/11483


ipweb2000:
In the documentation, I guess it should read "Number:ElectricCurrent" and not "ElectricPotential" for chargingAmps
| chargingamps              | Number:ElectricPotential | Set Charging Amps             | Current (Ampere) to use for charging

kaikreuzer:
Thanks @ipweb2000, well spotted. I corrected it.

bundles/org.openhab.binding.tesla/src/main/java/org/openhab/binding/tesla/internal/handler/TeslaVehicleHandler.java
```diff
@@ -263,6 +265,18 @@ public void handleCommand(ChannelUID channelUID, Command command) {
                             }
                             break;
                         }
+                        case CHARGE_AMPS:
+                            if (command instanceof QuantityType) {
```
lolodomo, line 269: You could also accept Decimal type?

kaikreuzer: I guess so, yes. I've changed the code accordingly.

bundles/org.openhab.binding.tesla/src/main/java/org/openhab/binding/tesla/internal/handler/TeslaVehicleHandler.java
```diff
@@ -266,9 +266,14 @@ public void handleCommand(ChannelUID channelUID, Command command) {
                             break;
                         }
                         case CHARGE_AMPS:
-                            if (command instanceof QuantityType) {
-                                @SuppressWarnings({ "unchecked", "null" })
-                                int amps = ((QuantityType<ElectricCurrent>) command).toUnit(Units.AMPERE).intValue();
+                            if (command instanceof DecimalType) {
+                                int amps = ((DecimalType) command).intValue();
+                                if (command instanceof QuantityType<?>) {
```
lolodomo, line 271: Sorry but I believe it is not good. There is no hierarchy between DecimalType and QuanyityType. 
You have to separate the two cases.

lolodomo (changes requested):

bundles/org.openhab.binding.tesla/src/main/java/org/openhab/binding/tesla/internal/handler/TeslaVehicleHandler.java
kaikreuzer: Thanks for looking so closely! You're clearly right, I have fixed this. Hope this time I got it correct!

lolodomo (approved):

Issue #11611, opened 2021-11-19
[TESLA] center_display_state need to be number


adriananton:
## Expected Behavior
Central Display State must return the status (0 to 9).

0 - Off
2 - On, standby or Camp Mode
3 - On, charging screen
4 - On
5 - On, Big charging screen
6 - On, Ready to unlock
7 - Sentry Mode
8 - Dog Mode
9 - Media
See details:  https://tesla-api.timdorr.com/vehicle/state/vehiclestate#response

## Current Behavior
Only 0 - Off is reported, all the rest are generating errors.

## Possible Solution
Change centerdisplay from OnOffType.class to DecimalType:

"CENTER_DISPLAY("center_display_state", "centerdisplay", DecimalType.class, false),"

## Context
Because we don't receive any Tesla Push Notfication in case of sentry mode event, I setup openHAB to get the status of the display and generate a Push Notification using a simple rule. 

## Your Environment

* Version used: openHAB 3.1.0 and Tesla add-on versions: <4.0.0
* Environment : ALL
* Operating System : ALL

PR #12030, opened 2022-01-11
[tesla] Change center_display_state from switch to number

kaikreuzer:
Fixes https://github.com/openhab/openhab-addons/issues/11611


lolodomo (approved):

lolodomo:
@kaikreuzer : can you confirm the channel was already existing in OH 3.2 ? In this case, this is a breaking change to mention in the 3.3 release notes.

kaikreuzer:
@lolodomo Yes it already existed, but as described in #11611, it never worked, so I don't think anybody uses it. Not sure if it makes any sense to then mention it in the release notes - I would waive it in this case.

adriananton (approved):

PR #11853, opened 2021-12-24
[hdpowerview] Refactor dynamic channels

jlaur:
Signed-off-by: Jacob Laursen <jacob-github@vindvejr.dk>

Fixes #11852

Refactoring after:
- #11534
- #11637

### Changes
- Extracted creation of dynamic channels into separate classes.
- Renamed methods to be consistent with channel terminology (scene collections -> scene groups, scheduled events -> automations).
- Optimized **AutomationChannelBuilder** scene/scene collection lookups.
- Reduced log level from **error** to **warning** for scheduled events with unknown scene or scene collection.
- Added unit test coverage for **SceneChannelBuilder**, **SceneGroupChannelBuilder** and **AutomationChannelBuilder**.
- Fix small problem with empty properties during startup.

### Test
I have been running this version for a longer period in production without any issues. All my channels (scenes, scene groups and automations) are working exactly as before the changes. This is the intention as the PR is purely refactoring not changing any behavior.

PR #12032, opened 2022-01-11
[hdpowerview] Fix SAT warnings

### LINKED ISSUE #12027: [hdpowerview] Fix SAT missing null annotation warnings
Unfortunately #11944 and #11980 introduced four unnoticed SAT warnings:

```
[WARNING] Code Analysis Tool has found: 
 0 error(s)!
 4 warning(s)
 0 info(s)
[WARNING] org.openhab.binding.hdpowerview.internal.api.Firmware.java:[20]
Classes/Interfaces/Enums should be annotated with @NonNullByDefault
[WARNING] org.openhab.binding.hdpowerview.internal.api.responses.FirmwareVersion.java:[20]
Classes/Interfaces/Enums should be annotated with @NonNullByDefault
[WARNING] org.openhab.binding.hdpowerview.internal.api.responses.FirmwareVersions.java:[22]
Classes/Interfaces/Enums should be annotated with @NonNullByDefault
[WARNING] org.openhab.binding.hdpowerview.internal.api.responses.Survey.java:[25]
Classes/Interfaces/Enums should be annotated with @NonNullByDefault
```

The reason for omitting `@NonNullByDefault` was to avoid annotating all member variables with `@Nullable` since the classes are only used for JSON deserialization.

Ideally the null annotation could be used to define the contract (instead of working against us), so this class:

```
public class FirmwareVersions {
    public Firmware mainProcessor;
    public Firmware radio;
}
```

could be changed to:
```
@NonNullByDefault
public class FirmwareVersions {
    public Firmware mainProcessor;
    @Nullable
    public Firmware radio;
}
```

to indicate that **mainProcessor** should always be set, and otherwise `JsonParseException` should be thrown. Unfortunately this is not possible, since object is constructed with default constructor, leaving all member variables as null.

See these comments:
- https://github.com/openhab/openhab-addons/pull/12002#issuecomment-1009368319
- https://github.com/openhab/openhab-addons/pull/12002#issuecomment-1009989135

jlaur:

Fixes #12027 

Changes done:
1. Fix SAT warnings about missing `@NonNullByDefault` annotations on classes.

Next steps (in context of new PR):
1. Refactor error handling: Throw exception if contract is breached.
2. Consider separating deserialization classes so they are not exposed directly to handlers (with `@Nullable` when contract doesn't allow this). Or find other way to make sure handlers are not bloated with null handling which shouldn't be needed for valid responses.

### DIFF
--- a/bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/HDPowerViewWebTargets.java
+++ b/bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/HDPowerViewWebTargets.java
@@ -29,6 +29,7 @@ import org.openhab.binding.hdpowerview.internal.api.ShadePosition;
 import org.openhab.binding.hdpowerview.internal.api.requests.ShadeMove;
 import org.openhab.binding.hdpowerview.internal.api.requests.ShadeStop;
 import org.openhab.binding.hdpowerview.internal.api.responses.FirmwareVersion;
+import org.openhab.binding.hdpowerview.internal.api.responses.FirmwareVersions;
 import org.openhab.binding.hdpowerview.internal.api.responses.SceneCollections;
 import org.openhab.binding.hdpowerview.internal.api.responses.Scenes;
 import org.openhab.binding.hdpowerview.internal.api.responses.ScheduledEvents;
@@ -130,15 +131,23 @@ public class HDPowerViewWebTargets {
     /**
      * Fetches a JSON package with firmware information for the hub.
      *
-     * @return FirmwareVersion class instance
+     * @return FirmwareVersions class instance
      * @throws JsonParseException if there is a JSON parsing error
      * @throws HubProcessingException if there is any processing error
      * @throws HubMaintenanceException if the hub is down for maintenance
      */
-    public @Nullable FirmwareVersion getFirmwareVersion()
+    public FirmwareVersions getFirmwareVersions()
             throws JsonParseException, HubProcessingException, HubMaintenanceException {
         String json = invoke(HttpMethod.GET, firmwareVersion, null, null);
-        return gson.fromJson(json, FirmwareVersion.class);
+        FirmwareVersion firmwareVersion = gson.fromJson(json, FirmwareVersion.class);
+        if (firmwareVersion == null) {
+            throw new JsonParseException("Missing firmware response");
+        }
+        FirmwareVersions firmwareVersions = firmwareVersion.firmware;
+        if (firmwareVersions == null) {
+            throw new JsonParseException("Missing 'firmware' element");
+        }
+        return firmwareVersions;
     }
 
     /**
--- a/bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/api/Firmware.java
+++ b/bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/api/Firmware.java
@@ -12,12 +12,17 @@
  */
 package org.openhab.binding.hdpowerview.internal.api;
 
+import org.eclipse.jdt.annotation.NonNullByDefault;
+import org.eclipse.jdt.annotation.Nullable;
+
 /**
  * Firmware version information for HD PowerView components
  *
  * @author Jacob Laursen - Initial contribution
  */
+@NonNullByDefault
 public class Firmware {
+    @Nullable
     public String name;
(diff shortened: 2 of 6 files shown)


andrewfg:
I will look at you code tomorrow, but here is some quick feedback..

Personally I would not bother with your proposal 3. above, since what you call “bloat” is actually “good practice” (against the possibility of some device in future not honouring the contract).

In addition, while you are refactoring your classes, I would also refactor the Shade fields position0 and positionKind0 from int to @nullable Integer, and add the respective null checks on those, at the next higher level. There may be other similar cases to..

jlaur:
> I will look at you code tomorrow, but here is some quick feedback..

Thanks, no rush. This is so far just some initial steps. Will probably await #11853 and #12002 before doing anything major, since I'll be changing all methods in `HDPowerViewWebTargets` as well as all callers, so will collide with that work. Hopefully they can be merged soon. I was even just giving `getShadeSurvey()` same treatment as `getFirmwareVersions()`, but realized that I was on collision course on caller side with one of those PR's. Conflicts are part of life, but I would rather avoid having to do everything all over if I can.

> Personally I would not bother with your proposal 3. above, since what you call “bloat” is actually “good practice” (against the possibility of some device in future not honouring the contract).

One of the problems, as I see it, is that if something is actually **required** in a response for the response to make sense at all, this should be verified and caller shouldn't have to deal with this, but just provide generic error handling, i.e. catch some exception. 

**Example:** For the firmware response to make sense, **mainProcessor** is _required_. **radio**, on the other hand, is _optional_ as it was added in Hub v2, apparently.

So let's say we did this validation in getFirmwareVersions:

```
if (firmwareVersions.mainProcessor == null) {
    throw new JsonParseException("Missing 'mainProcessor' element");
}
```

Unfortunately this wouldn't help. Caller would still have to perform the exact same check, since everything is by definition `@Nullable`:

```
@Nullable
public Firmware mainProcessor;
```

That's why I was thinking about possibly introducing a class with stricter annotation adhering to a contract instead of just returning objects of `FirmwareVersions` where everything is nullable because of reflection/gson deserialization.

This would introduce some overhead with maintaining additional model classes, but on the other hand it would simplify the handlers/logic because they wouldn't have to check for nulls everywhere only because of undesired nullness annotation from deserialization classes. That's what I consider bloat, or at least cluttering. Logic is harder to read IMHO when entangled with checks needed because of nullness annotations.

Maybe there are some JSON deserialization alternatives that supports annotations and/or have runtime checks. I'll look into other bindings also as this must be a common challenge. :-)

> In addition, while you are refactoring your classes, I would also refactor the Shade fields position0 and positionKind0 from int to @nullable Integer, and add the respective null checks on those, at the next higher level. There may be other similar cases too..

Sure, plan would be to refactor all of them when having found a reasonable solution. The concrete example you are mentioning would have the same problem as described above. Since they would be nullable, we would have to provide duplicate checks to avoid warnings, I think. :-(

andrewfg:
The OH developers / maintainers decided on GSON as the standard JSON (de)serialization librarary. And this library does not (yet) support Required or Optional annotations (https://github.com/google/gson/issues/61) so it seems to me that this discussion abot replacing or overwriting the standard OH library should not be specific to this binding, but rather should be addressed to OH developers / maintainers for the whole OH eco-system.

lolodomo (approved):
LGTM

lolodomo:
Is it ready for a merge? In this case, please remove draft.

andrewfg:
> Is it ready for a merge? In this case, please remove draft/WIP?

@lolodomo this PR is still a draft, and not ready to merge; you are probably confusing it with #11853 which is indeed ready.

jlaur:
> Is it ready for a merge? In this case, please remove draft/WIP?

No, it's only getting started, but now a little bit stuck because of these other PR's that would need to go first to avoid major conflicts and rework:
- #11853 
- #12002

lolodomo:
No, I am not confusing it with another PR, I reviewed the code for this PR and it looks good to me.
Let me know when this is finished.

andrewfg:
> No, I am not confusing it with another PR, I reviewed the code for this PR and it looks good to me. Let me know when this is finished.

@lolodomo, IMHO this is far from finished, so perhaps your "LGTM" is premature (??) - please look at my comment a few posts above this https://github.com/openhab/openhab-addons/pull/12032#issuecomment-1011997519

lolodomo:
If you can keep small PR, that is really better for reviewing.

jlaur:
> If you can have several small PRs rather one very big PR, that is really better for reviewing.

Sure, I'll split it. This one can be merged, I'll update the title and description.

PR #12034, opened 2022-01-12
[icloud] Update README.md

gcatellani:
The fact that my iPhones weren't showing any information when configured via configuration files and the fact that it worked when configured via the OpenHab 3 Admin area, made me realize that the information provided in the readme is outdated. It misses any indication that for the device thing to work, a new "owner"-parameter needs to be provided. I updated the readme accordingly.


kaikreuzer:
Looking [at the code](https://github.com/openhab/openhab-addons/blob/main/bundles/org.openhab.binding.icloud/src/main/resources/OH-INF/thing/thing-types.xml#L49-L53), there is no such thing as a new "owner" parameter that would need to be filled. "owner" is simply a Thing property that the automatic discovery adds, but this is nothing required at all.

To verify it, I just tested it and my devices went ONLINE without any problem by just providing the deviceId.

Issue #11985, opened 2022-01-07
[knx] Dimmer Item assumes a brightness value of 100% when switched on


smartheit:
## Expected Behavior
When a ON command is sent the brightness value of a dimmer item should reflect the "real" brightness value not 100%.

## Current Behavior
When sending an "ON" command to the dimmer item the brightness state is always 100%. In my case the KNX DALI Gateway is configured to set the brightness value to the last known state.
This is the ETS bus monitor log:
![image](https://user-images.githubusercontent.com/88311932/148533121-6fd8446c-4afe-4c67-b170-7bb89dcb047e.png)

The On/Off commands were sent through OpenHAB. 

The dimmer item shows a brightness value of 100% not 30%:
![image](https://user-images.githubusercontent.com/88311932/148533524-a0a5f313-6aa9-419a-a2f2-fa2a23360cba.png)

The dimmer item is configured in the knx.things file:
`Type dimmer : Licht_EG_Wohnen_Wand [ switch="10/2/70+<11/2/70", position="13/2/70+<14/2/70", increaseDecrease="12/2/70" ]`

And the according item is:
`Dimmer Licht_EG_Wohnen_Wand        "EG Wohnen Wandlampe [%d %%]" <light> (gLicht, gLichterAn, gLichtAlarm) { channel="knx:device:bridge:lichter:Licht_EG_Wohnen_Wand" }`

When triggering a manual read of the GA containing the brightness readback value in the ETS the dimmer item updates with the correct value. In my test case reading from address 14/2/70 results in an updated item value (30% brightness)

## Context
The behavior is occurring in two OpenHAB installations, in my private Installation (JUNG Dali Gateway) and on the customers site (HAGER Dali Gateway) 

## Your Environment
Openhab Docker Container with OpenHAB 3.2.0
The behavior also occurred in previous versions including OH2.x

Rossko57:
Please check your `events.log` for the effects of autoupdate feature on your Item.

kaikreuzer:
As @Rossko57 says: The 100% event comes from the auto-update prediction. If you do not want this [you should disable auto-update for this item](https://www.openhab.org/docs/configuration/items.html#parameter-autoupdate), but in this case you should make sure that your KNX actuator sends a state update through a listening GA (`14/2/70`) to that channel.

kaikreuzer:
[...]

PR #12035, opened 2022-01-12
[knx] Update README.md

udo1toni:
Changed Example of Dimmer channel. Makes no sense to provide both Switch and Position Status in configuration, as an update of Switch Status will delete Position and vice versa. Position Status makes sense.


Confectrician (approved):
Looks good. I have followed the corresponding knx user forum discussion.

kaikreuzer (approved):
Thanks. Sounds pretty much like https://github.com/openhab/openhab-addons/issues/11985 as well.

PR #12049, opened 2022-01-15
[hdpowerview] Refactor exception handling

### LINKED ISSUE #12048: [hdpowerview] Consolidate error handling
In order to simplify exception handling in handlers, `JsonParseException` should be handled by `HDPowerViewWebTargets` and not exposed to the handlers. Instead it should be wrapped (chained) in a new custom exception which inherits from a common custom exception. This way callers can decide the granularity of the error handling and can choose to simply catch the base exception type.

Also, we are using model objects directly in the logic (handlers) and they have nullness annotations complying to gson, not to the web service contracts. As long as we have this, we can at least perform some basic checking and avoid returning null objects.

Previous step: #12027

jlaur:
Fixes #12048

Changes:
- `JsonParseException` is wrapped in custom exception and not exposed directly to handlers.
- `HDPowerViewWebTargets` methods no longer return null objects, instead will throw exceptions.
- Custom exceptions are now inherited from `HubException` so callers can decide the granularity of the error handling.

Continuation of #12032

### DIFF
--- a/bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/HDPowerViewWebTargets.java
+++ b/bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/HDPowerViewWebTargets.java
@@ -49,6 +49,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import com.google.gson.Gson;
+import com.google.gson.JsonElement;
 import com.google.gson.JsonObject;
 import com.google.gson.JsonParseException;
 import com.google.gson.JsonParser;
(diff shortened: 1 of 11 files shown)


bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/HDPowerViewWebTargets.java
```diff
@@ -268,34 +335,49 @@ public void activateSceneCollection(int sceneCollectionId) throws HubProcessingE
...
         String uri = scheduledEvents + "/" + scheduledEventId;
-        String json = invoke(HttpMethod.GET, uri, null, null);
-        JsonObject jsonObject = JsonParser.parseString(json).getAsJsonObject();
-        JsonObject scheduledEventObject = jsonObject.get("scheduledEvent").getAsJsonObject();
-        scheduledEventObject.addProperty("enabled", enable);
-        invoke(HttpMethod.PUT, uri, null, jsonObject.toString());
+        String jsonResponse = invoke(HttpMethod.GET, uri, null, null);
+        try {
+            JsonObject jsonObject = JsonParser.parseString(jsonResponse).getAsJsonObject();
+            JsonObject scheduledEventObject = jsonObject.get("scheduledEvent").getAsJsonObject();
```
lolodomo, line 375: `jsonObject.get("scheduledEvent")` could in theory return null

lolodomo (changes requested):

lolodomo:
@andrewfg for an additional review

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/HDPowerViewWebTargets.java
jlaur: You are right - fixed.

lolodomo (approved):
LGTM

PR #12055, opened 2022-01-15
[googlestt] initial contribution

GiviMAD:

This service allows to use Google Cloud Speech-to-Text service within openHAB.
I have tested it with the Keyword Spotter Porcupine from my other [PR](https://github.com/openhab/openhab-addons/pull/12028) using es-ES locale (setting the system microphone as default audio source), both seems to work fine, it correctly spots my wake word ("oye casa" in my case) and transcripts the audio quite fine.
Right now there is not easy way to test this, I had to create a service while developing to call the startDialog method on the voice manager. Will open an issue in the core asking if this can be exposed on the 3.3.0-snapshot so users can test them properly.
Hope the code is in an acceptable state, let me know for anything that can be improved.
Thanks in advance.


bundles/pom.xml
```diff
@@ -397,6 +397,7 @@
     <module>org.openhab.voice.picotts</module>
     <module>org.openhab.voice.pollytts</module>
     <module>org.openhab.voice.voicerss</module>
+    <module>org.openhab.voice.googlestt</module>
```
lolodomo, line 400: Please keep the alphabetic order

bundles/org.openhab.voice.googlestt/src/main/resources/OH-INF/config/config.xml
```diff
@@ -0,0 +1,67 @@
...
+		<parameter name="clientSecret" type="text" required="true" groupName="authentication">
+			<context>password</context>
+			<label>Client Secret</label>
+			<description>Google Cloud Platform OAuth 2.0-Client Secret.</description>
+		</parameter>
+		<parameter name="oauthCode" type="text" groupName="authentication">
+			<label>Authorization Code</label>
+			<description><![CDATA[The oauth code is a one-time code needed to retrieve the necessary access token from Google Cloud Platform. <b>Please go to your browser ...</b> https://accounts.google.com/o/oauth2/auth?client_id=\<YOUR_CLIENT_ID\>&redirect_uri=urn:ietf:wg:oauth:2.0:oob&scope=https://www.googleapis.com/auth/cloud-platform&response_type=code <b>... to generate an auth-code and paste it here</b>.]]></description>
+		</parameter>
+		<parameter name="singleUtteranceMode" type="boolean" required="true" groupName="stt">
```
lolodomo, line 34: I assume `required="true"` is wrong as you provide a default value

bundles/org.openhab.voice.googlestt/src/main/resources/OH-INF/config/config.xml
```diff
@@ -0,0 +1,67 @@
...
+			<label>Single Utterance Mode</label>
+			<description>When enabled Google Cloud Platform is responsible for detect when to stop listening after a single
+				utterance. (Recommended)</description>
+			<default>true</default>
+		</parameter>
+		<parameter name="maxTranscriptionSeconds" type="integer" required="true" groupName="stt">
```
lolodomo, line 40: I assume `required="true"` is wrong as you provide a default value.
But you add the unit, probably `unit="s"` (to be verified)

bundles/org.openhab.voice.googlestt/src/main/resources/OH-INF/config/config.xml
```diff
@@ -0,0 +1,67 @@
...
+			<label>Max Transcription Seconds</label>
+			<description>Max seconds to wait to force stop the transcription.</description>
+			<default>60</default>
+		</parameter>
+		<parameter name="maxSilenceSeconds" type="integer" required="true" groupName="stt">
```
lolodomo, line 45: I assume `required="true"` is wrong as you provide a default value
But you add the unit, probably `unit="s"` (to be verified)

bundles/org.openhab.voice.googlestt/src/main/resources/OH-INF/config/config.xml
```diff
@@ -0,0 +1,67 @@
...
+			<label>Max Silence Seconds</label>
+			<description>Only works when singleUtteranceMode is disabled, max seconds without get new transcriptions to stop
+				listening.</description>
+			<default>5</default>
+		</parameter>
+		<parameter name="refreshSupportedLocales" type="boolean" required="true" groupName="stt">
```
lolodomo, line 51: I assume `required="true"` is wrong as you provide a default value

bundles/org.openhab.voice.googlestt/src/main/java/org/openhab/voice/googlestt/internal/GoogleSTTBindingConstants.java
```diff
@@ -0,0 +1,43 @@
...
+ * used across the whole binding.
+ *
+ * @author Miguel Álvarez - Initial contribution
+ */
+@NonNullByDefault
+public class GoogleSTTBindingConstants {
+    /**
+     * Service name
+     */
+    static final String SERVICE_NAME = "Google Cloud Speech-to-Text";
```
lolodomo, line 28: Please add "public" on all your constants definition in this class.

bundles/org.openhab.voice.googlestt/pom.xml
```diff
@@ -0,0 +1,138 @@
...
+
+  <parent>
+    <groupId>org.openhab.addons.bundles</groupId>
+    <artifactId>org.openhab.addons.reactor.bundles</artifactId>
+    <version>3.3.0-SNAPSHOT</version>
+  </parent>
+
+  <artifactId>org.openhab.voice.googlestt</artifactId>
+
+  <name>openHAB Add-ons :: Bundles :: Google Cloud Speech to Text</name>
```
lolodomo, line 15: Rather <`name>openHAB Add-ons :: Bundles :: Voice :: Google Cloud Speech to Text</name>`

bundles/org.openhab.voice.googlestt/pom.xml
```diff
@@ -0,0 +1,138 @@
...
+    <bnd.importpackage>!*opencensus*,!org.bouncycastle*,!*jboss*,!javax.annotation.*,!net.jpountz.*,!lzma.sdk.*,org.eclipse.jetty.*;resolution:=optional,com.ning.*;resolution:=optional,com.jcraft.*;resolution:=optional,com.google.re2j.*;resolution:=optional,com.google.api.client.*;resolution:=optional,org.conscrypt.*;resolution:=optional,!io.grpc.census.*,com.sun.jndi.dns.*;resolution:=optional,org.apache.log.*;resolution:=optional,org.apache.http.*;resolution:=optional,sun.security.*;resolution:=optional,com.oracle.svm.core.annotate.*;resolution:=optional,*blockhound*;resolution:=optional,com.google.protobuf.nano.*;resolution:=optional,io.grpc.*;resolution:=optional,com.google.protobuf.*;resolution:=optional,io.perfmark.*;resolution:=optional</bnd.importpackage>
+  </properties>
+  <dependencies>
+    <dependency>
+      <groupId>com.google.cloud</groupId>
+      <artifactId>google-cloud-speech</artifactId>
+      <version>2.2.2</version>
+    </dependency>
+    <!--cloud-speech deps -->
+    <dependency>
```
lolodomo, line 26: To be checked by someone more expert than me, including if someone has to be added in your feature.xml file.

lolodomo (changes requested):
Review part 1 of 2

bundles/org.openhab.voice.googlestt/src/main/resources/OH-INF/config/config.xml
GiviMAD: Sorry I think I miss understood this. I want the user not to be able to save with this field without value. I think the required is ok for this as the default value will be used just on creation it will not prevent the user to unset this value. Am I wrong?

bundles/org.openhab.voice.googlestt/src/main/resources/OH-INF/config/config.xml
GiviMAD: Sorry I think I miss understood this. I though I should require those to avoid the user to clean the values from the interface and that the default values were only used the first time the service is registered. Same apply to others.

bundles/org.openhab.voice.googlestt/pom.xml
GiviMAD: Yes please, I'm also not an expert on this matter. This is the smallest set of dependencies I've been able use to made it work.

GiviMAD:
I was pushed some changes, the more relevant is the last one, after reviewing the dialog implementation I realize that I can not avoid emit the errors as I was doing, each execution need to emit a SpeechRecognitionEvent or SpeechRecognitionErrorEvent to notify it has ended.

bundles/org.openhab.voice.googlestt/pom.xml
lolodomo: Like in your other binding (spotter), I think it will better to add the "compile".

lolodomo:
Before I start again the second part of my review (the java code), can you please first apply all my comments from your other PR (var type, implementing STTServiceHandle, ...).

bundles/org.openhab.voice.googlestt/pom.xml
```diff
@@ -0,0 +1,138 @@
...
+    </dependency>
+    <dependency>
+      <groupId>com.google.api.grpc</groupId>
+      <artifactId>proto-google-cloud-speech-v1</artifactId>
+      <version>2.2.2</version>
+    </dependency>
+    <dependency>
+      <groupId>com.google.protobuf</groupId>
+      <artifactId>protobuf-java</artifactId>
+      <version>3.19.2</version>
```
jlaur, line 54: Just sharing information: #11990 upgrades protobuf-java to 3.16.1 for the chromecast binding. Maybe we can align to 3.19.3 if both are compatible. Or you can disregard, just remembered seeing this. :-)

bundles/org.openhab.voice.googlestt/src/main/java/org/openhab/voice/googlestt/internal/GoogleSTTService.java
```diff
@@ -0,0 +1,367 @@
...
+            final @Reference ConfigurationAdmin configAdmin) {
+        LoadBalancerRegistry.getDefaultRegistry().register(new PickFirstLoadBalancerProvider());
+        this.oAuthFactory = oAuthFactory;
+        this.configAdmin = configAdmin;
+    }
+
+    @Activate
+    protected void activate(Map<String, Object> config) {
+        var serviceConfig = new org.openhab.core.config.core.Configuration(config).as(GoogleSTTConfiguration.class);
+        this.config = serviceConfig;
```
lolodomo, line 101: You can avoid the variable serviceConfig and use directly this.config.
At the same time, it will remove one of your var type.

bundles/org.openhab.voice.googlestt/src/main/java/org/openhab/voice/googlestt/internal/GoogleSTTService.java
```diff
@@ -0,0 +1,367 @@
...
+import com.google.cloud.speech.v1.StreamingRecognizeRequest;
+import com.google.cloud.speech.v1.StreamingRecognizeResponse;
+import com.google.protobuf.ByteString;
+
+import io.grpc.LoadBalancerRegistry;
+import io.grpc.internal.PickFirstLoadBalancerProvider;
+
+/**
+ * The {@link GoogleSTTService} is responsible for handling commands, which are
+ * sent to one of the channels.
```
lolodomo, line 67: Description to be updated

bundles/org.openhab.voice.googlestt/src/main/java/org/openhab/voice/googlestt/internal/GoogleSTTService.java
```diff
@@ -0,0 +1,367 @@
...
+
+    private GoogleSTTConfiguration config = new GoogleSTTConfiguration();
+    private @Nullable OAuthClientService oAuthService;
+
+    @Activate
+    public GoogleSTTService(final @Reference OAuthFactory oAuthFactory,
+            final @Reference ConfigurationAdmin configAdmin) {
+        LoadBalancerRegistry.getDefaultRegistry().register(new PickFirstLoadBalancerProvider());
+        this.oAuthFactory = oAuthFactory;
+        this.configAdmin = configAdmin;
```
lolodomo, line 95: Without configAdmin, it is not possible to update the service configuration inside the service itself ?

bundles/org.openhab.voice.googlestt/src/main/java/org/openhab/voice/googlestt/internal/GoogleSTTService.java
```diff
@@ -0,0 +1,367 @@
...
+    }
+
+    @Activate
+    protected void activate(Map<String, Object> config) {
+        var serviceConfig = new org.openhab.core.config.core.Configuration(config).as(GoogleSTTConfiguration.class);
+        this.config = serviceConfig;
+        executor.submit(() -> GoogleSTTLocale.loadLocales(serviceConfig.refreshSupportedLocales));
+        String clientId = serviceConfig.clientId;
+        String clientSecret = serviceConfig.clientSecret;
+        if (!clientId.isEmpty() && !clientSecret.isEmpty()) {
```
lolodomo, line 105: You can even use isBlank.
In case one of both is not ok, you could log something.

bundles/org.openhab.voice.googlestt/src/main/java/org/openhab/voice/googlestt/internal/GoogleSTTService.java
```diff
@@ -0,0 +1,367 @@
...
+            var oAuthService = oAuthFactory.createOAuthClientService(SERVICE_PID, GCP_TOKEN_URI, GCP_AUTH_URI, clientId,
+                    clientSecret, GCP_SCOPE, false);
```
lolodomo, line 107: Please avoid var type in all the class.

bundles/org.openhab.voice.googlestt/src/main/java/org/openhab/voice/googlestt/internal/GoogleSTTService.java
```diff
@@ -0,0 +1,367 @@
...
+                new AudioFormat(AudioFormat.CONTAINER_OGG, "OPUS", null, null, null, 16000L),
+                new AudioFormat(AudioFormat.CONTAINER_OGG, "OPUS", null, null, null, 24000L),
+                new AudioFormat(AudioFormat.CONTAINER_OGG, "OPUS", null, null, null, 48000L));
+    }
+
+    @Override
+    public STTServiceHandle recognize(STTListener sttListener, AudioStream audioStream, Locale locale,
+            Set<String> set) {
+        var scheduledTask = backgroundRecognize(sttListener, audioStream, locale, set);
+        return () -> scheduledTask.cancel(true);
```
lolodomo, line 146: This is not false but to make it more understandable, please return new STTServiceHandle() with an override of the abort method.

bundles/org.openhab.voice.googlestt/src/main/java/org/openhab/voice/googlestt/internal/GoogleSTTService.java
```diff
@@ -0,0 +1,367 @@
...
+
+    private void getAccessToken(OAuthClientService oAuthService, String oauthCode) {
+        logger.debug("Trying to get access and refresh tokens.");
+        try {
+            oAuthService.getAccessTokenResponseByAuthorizationCode(oauthCode, GCP_REDIRECT_URI);
+        } catch (OAuthException | OAuthResponseException e) {
+            logger.debug("Error fetching access token: {}", e.getMessage(), e);
+            logger.error("Error fetching access token. Invalid oauth code? Please generate a new one.");
+        } catch (IOException e) {
+            logger.error("An unexpected IOException occurred: {}", e.getMessage());
```
lolodomo, line 157: WARN level is enough. Please avoid ERROR level in all the class.

Rather than 2 consecutive calls to logger, you could use an if then else with logger.isDebugEnabled
In case of IOException, please inject a context in the message, something like: An unexpected IOException occurred when fetching access token: {}

bundles/org.openhab.voice.googlestt/src/main/java/org/openhab/voice/googlestt/internal/GoogleSTTService.java
```diff
@@ -0,0 +1,367 @@
...
+        executor.submit(() -> GoogleSTTLocale.loadLocales(serviceConfig.refreshSupportedLocales));
+        String clientId = serviceConfig.clientId;
+        String clientSecret = serviceConfig.clientSecret;
+        if (!clientId.isEmpty() && !clientSecret.isEmpty()) {
+            var oAuthService = oAuthFactory.createOAuthClientService(SERVICE_PID, GCP_TOKEN_URI, GCP_AUTH_URI, clientId,
+                    clientSecret, GCP_SCOPE, false);
+            this.oAuthService = oAuthService;
+            if (!serviceConfig.oauthCode.isEmpty()) {
+                getAccessToken(oAuthService, serviceConfig.oauthCode);
+                deleteAuthCode();
```
lolodomo, line 111: You erase it from the configuration even in case of success ?

bundles/org.openhab.voice.googlestt/src/main/java/org/openhab/voice/googlestt/internal/GoogleSTTService.java
```diff
@@ -0,0 +1,367 @@
...
+            }
+        }
+    }
+
```
lolodomo, line 115: You do not need to handle an update of the configuration (method modified) ?

bundles/org.openhab.voice.googlestt/src/main/resources/OH-INF/config/config.xml
```diff
@@ -0,0 +1,67 @@
...
+			<description>Google Cloud Platform OAuth 2.0-Client Secret.</description>
+		</parameter>
+		<parameter name="oauthCode" type="text" groupName="authentication">
+			<label>Authorization Code</label>
+			<description><![CDATA[The oauth code is a one-time code needed to retrieve the necessary access token from Google Cloud Platform. <b>Please go to your browser ...</b> https://accounts.google.com/o/oauth2/auth?client_id=\<YOUR_CLIENT_ID\>&redirect_uri=urn:ietf:wg:oauth:2.0:oob&scope=https://www.googleapis.com/auth/cloud-platform&response_type=code <b>... to generate an auth-code and paste it here</b>.]]></description>
+		</parameter>
+		<parameter name="singleUtteranceMode" type="boolean" groupName="stt">
+			<label>Single Utterance Mode</label>
+			<description>When enabled Google Cloud Platform is responsible for detect when to stop listening after a single
+				utterance. (Recommended)</description>
```
lolodomo, line 37: ... is responsible for **detecting** when ...

bundles/org.openhab.voice.googlestt/src/main/resources/OH-INF/config/config.xml
```diff
@@ -0,0 +1,67 @@
...
+		</parameter>
+		<parameter name="maxTranscriptionSeconds" type="integer" unit="s" groupName="stt">
+			<label>Max Transcription Seconds</label>
+			<description>Max seconds to wait to force stop the transcription.</description>
+			<default>60</default>
+		</parameter>
+		<parameter name="maxSilenceSeconds" type="integer" unit="s" groupName="stt">
+			<label>Max Silence Seconds</label>
+			<description>Only works when singleUtteranceMode is disabled, max seconds without get new transcriptions to stop
+				listening.</description>
```
lolodomo, line 48: ... without **getting** new ...

bundles/org.openhab.voice.googlestt/src/main/resources/OH-INF/config/config.xml
```diff
@@ -0,0 +1,67 @@
...
+			<default>5</default>
+		</parameter>
+		<parameter name="refreshSupportedLocales" type="boolean" groupName="stt">
+			<label>Refresh Supported Locales</label>
+			<description>Try to load supported locales from the documentation page.</description>
```
lolodomo, line 53: Try **loading** ...

bundles/org.openhab.voice.googlestt/src/main/resources/OH-INF/config/config.xml
```diff
@@ -0,0 +1,67 @@
...
+			<default>false</default>
+		</parameter>
+		<parameter name="noResultsMessage" type="text" groupName="messages">
+			<label>No Results Message</label>
+			<description>Message to be told when no results. (Empty for disabled)</description>
+			<default>Sorry, I didn't understand you</default>
```
lolodomo, line 59: Having an English hardcoded default value is a problem.
I believe you can't use `@text` in a default tag.
I have no clear idea how to provide a translatable default value.

bundles/org.openhab.voice.googlestt/src/main/resources/OH-INF/config/config.xml
```diff
@@ -0,0 +1,67 @@
...
+		</parameter>
+		<parameter name="errorMessage" type="text" groupName="messages">
+			<label>Error Message</label>
+			<description>Message to be told when an error has happened. (Empty for disabled)</description>
+			<default>Sorry, something went wrong</default>
```
lolodomo, line 64: Same here

lolodomo (changes requested):
Review part 2 of 3

bundles/org.openhab.voice.googlestt/src/main/java/org/openhab/voice/googlestt/internal/GoogleSTTService.java
GiviMAD: Yes, to avoid that login is retried on a service restart because the token is managed and renewed by the oauthservice. 
Maybe you already notice this but I used the Google Text-to-Speech add-on as reference for that part so the code is mostly the same.

bundles/org.openhab.voice.googlestt/src/main/resources/OH-INF/config/config.xml
GiviMAD: Will be better if I remove the default value?

lolodomo: Let me think how to do that properly

bundles/org.openhab.voice.googlestt/src/main/java/org/openhab/voice/googlestt/internal/GoogleSTTService.java
GiviMAD: No, as I'm not using any base class for the service just implementing an interface I have no build-in method to update it's config. Or at least I couldn't find any. The Google Text-To-Speech service also use this method to update its configuration.

bundles/org.openhab.voice.googlestt/pom.xml
GiviMAD: I just tested it and not working with that version, I will continue with this one for now.

bundles/org.openhab.voice.googlestt/src/main/java/org/openhab/voice/googlestt/internal/GoogleSTTService.java
lolodomo: Ok

bundles/org.openhab.voice.googlestt/src/main/resources/OH-INF/config/config.xml
lolodomo: @cweitkamp / we need your advice here. This is the case we need a default parameter value that should be translated for each language. I believe we can't use `@text.xxx` as value because this is not managed by the core framework ?
What would be the proper solution in this case ?

lolodomo:
@GiviMAD : I submitted a PR to your branch regarding i18n and README.

bundles/org.openhab.voice.googlestt/pom.xml
lolodomo: @wborn : is it better to add compile tag also on all dependencies of the main dependency ?

bundles/org.openhab.voice.googlestt/src/main/java/org/openhab/voice/googlestt/internal/GoogleSTTLocale.java
```diff
@@ -0,0 +1,91 @@
...
+            HttpURLConnection con = (HttpURLConnection) url.openConnection();
+            con.setRequestMethod("GET");
+            con.setRequestProperty("Content-Type", "text/html");
+            int status = con.getResponseCode();
+            if (status != 200) {
+                logger.warn("Http error loading supported locales, code: {}", status);
+                loadLocalesFromLocal();
+                return;
+            }
+            var html = new String(con.getInputStream().readAllBytes());
```
lolodomo, line 66: var type everywhere in this class

bundles/org.openhab.voice.googlestt/src/main/java/org/openhab/voice/googlestt/internal/GoogleSTTLocale.java
```diff
@@ -0,0 +1,91 @@
...
+            while (matcher.find()) {
+                var locale = new Locale(matcher.group("lang"), matcher.group("country"));
+                if (lastLocale == null || !lastLocale.equals(locale)) {
+                    lastLocale = locale;
+                    SUPPORTED_LOCALES.add(locale);
+                    logger.debug("Locale added {}", locale.toLanguageTag());
+                }
+            }
+        } catch (IOException e) {
+            logger.error("Error  loading supported locales: {}", e.getMessage());
```
lolodomo, line 80: warn

bundles/org.openhab.voice.googlestt/src/main/java/org/openhab/voice/googlestt/internal/GoogleSTTService.java
```diff
@@ -0,0 +1,381 @@
...
+            final @Reference ConfigurationAdmin configAdmin) {
+        LoadBalancerRegistry.getDefaultRegistry().register(new PickFirstLoadBalancerProvider());
+        this.oAuthFactory = oAuthFactory;
+        this.configAdmin = configAdmin;
+    }
+
+    @Activate
+    protected void activate(Map<String, Object> config) {
+        this.config = new Configuration(config).as(GoogleSTTConfiguration.class);
+        executor.submit(() -> GoogleSTTLocale.loadLocales(this.config.refreshSupportedLocales));
```
lolodomo, line 89: Just a thought: is it not necessary to be sure that all locales are loaded before going further ?

lolodomo (changes requested):
Review part 2,5 from 3

bundles/org.openhab.voice.googlestt/pom.xml
wborn: It's subjective if it is better or not to add "compile" as scope because it is the default scope anyways. With openHAB I usually add it because it is more common to do that here but on other projects I don't.

bundles/org.openhab.voice.googlestt/src/main/resources/OH-INF/config/config.xml
lolodomo: Or we can simply keep the default as it is and users wanting other language than English will update this setting.

bundles/org.openhab.voice.googlestt/src/main/java/org/openhab/voice/googlestt/internal/GoogleSTTService.java
lolodomo: What will happen if recognize is called iwhile they are not yet loaded?

bundles/org.openhab.voice.googlestt/src/main/java/org/openhab/voice/googlestt/internal/GoogleSTTLocale.java
lolodomo: For your information, I am trying to clarify the status with var type with maintainers.

lolodomo: Ok, it looks like there is no reason to "reject" var type. So sorry for my previous requests to replace var type with real types.
If you like to use var type, you can.

bundles/org.openhab.voice.googlestt/src/main/resources/OH-INF/config/config.xml
cweitkamp: I think we are okay with the English text as default value here as the user is able to define their own value anyways.

lolodomo: Yes, ok for me

bundles/org.openhab.voice.googlestt/src/main/java/org/openhab/voice/googlestt/internal/GoogleSTTService.java
GiviMAD: I think is that case the framework will tell you your language is not supported, I'm not really using those values, they are just returned in the supported languages call, is done in a similar way by the Google TTS that takes the supported languages from the voices which are also loaded asynchronously.

bundles/org.openhab.voice.googlestt/src/main/java/org/openhab/voice/googlestt/internal/GoogleSTTService.java
```diff
@@ -0,0 +1,381 @@
...
+            }
+        } catch (IOException e) {
+            logger.warn("Failed to delete current oauth code, please delete it manually.");
+        }
+    }
+
+    private Future<?> backgroundRecognize(STTListener sttListener, AudioStream audioStream, AtomicBoolean keepStreaming,
+            Locale locale, Set<String> set) {
+        Credentials credentials = getCredentials();
+        return executor.submit(() -> {
```
lolodomo, line 190: Is two tasks needed ?
Why not doing like in your procupineks service ?

bundles/org.openhab.voice.googlestt/src/main/java/org/openhab/voice/googlestt/internal/GoogleSTTService.java
```diff
@@ -0,0 +1,381 @@
...
+        long maxSilenceMillis = (config.maxSilenceSeconds * 1000L);
+        while (keepStreaming.get()) {
+            byte[] data = new byte[6400];
+            int dataN = audioStream.read(data);
+            if (!keepStreaming.get() || isExpiredInterval(maxTranscriptionMillis, startTime)) { // 60 seconds
+                logger.debug("Stops listening, max transcription time reached");
+                break;
+            }
+            if (!config.singleUtteranceMode
+                    && isExpiredInterval(maxSilenceMillis, responseObserver.getLastInputTime())) { // 60 seconds
```
lolodomo, line 247: Comment is wrong as it is 5 seconds by default

bundles/org.openhab.voice.googlestt/src/main/java/org/openhab/voice/googlestt/internal/GoogleSTTService.java
```diff
@@ -0,0 +1,381 @@
...
+        int frequency = Math.toIntExact(longFrequency);
+        // First thing we need to send the stream config
+        sendStreamConfig(clientStream, streamEncoding, frequency, channels, locale);
+        // Loop sending audio data
+        long startTime = System.currentTimeMillis();
+        long maxTranscriptionMillis = (config.maxTranscriptionSeconds * 1000L);
+        long maxSilenceMillis = (config.maxSilenceSeconds * 1000L);
+        while (keepStreaming.get()) {
+            byte[] data = new byte[6400];
+            int dataN = audioStream.read(data);
```
lolodomo, line 241: No need to check dataN ? In case of error...

bundles/org.openhab.voice.googlestt/src/main/java/org/openhab/voice/googlestt/internal/GoogleSTTService.java
```diff
@@ -0,0 +1,381 @@
...
+            logger.debug("Got {} results", response.getResultsList().size());
+            if (results.isEmpty()) {
+                logger.debug("No results");
+                return;
+            }
+            results.forEach(result -> {
+                List<SpeechRecognitionAlternative> alternatives = result.getAlternativesList();
+                logger.debug("Got {} alternatives", alternatives.size());
+                SpeechRecognitionAlternative alternative = alternatives.stream()
+                        .min(Comparator.comparing(SpeechRecognitionAlternative::getConfidence)).orElse(null);
```
lolodomo, line 330: You take the one with the lower confidence ?

bundles/org.openhab.voice.googlestt/src/main/java/org/openhab/voice/googlestt/internal/GoogleSTTService.java
```diff
@@ -0,0 +1,381 @@
...
+            sttListener.sttEventReceived(new SpeechStopEvent());
+            float averageConfidence = confidenceSum / (float) responseCount;
+            String transcript = transcriptBuilder.toString();
+            if (!transcript.isBlank()) {
+                sttListener.sttEventReceived(new SpeechRecognitionEvent(transcript, averageConfidence));
+            } else {
+                if (!config.noResultsMessage.isBlank()) {
+                    sttListener.sttEventReceived(new SpeechRecognitionErrorEvent(config.noResultsMessage));
+                } else {
+                    sttListener.sttEventReceived(new SpeechRecognitionErrorEvent("No results"));
```
lolodomo, line 359: You have to put this text in properties to make it translatable in other languages.
If I am not wrong, this message will be said with the audio sink.

bundles/org.openhab.voice.googlestt/src/main/java/org/openhab/voice/googlestt/internal/GoogleSTTService.java
```diff
@@ -0,0 +1,381 @@
...
+        public void onError(@Nullable Throwable t) {
+            logger.warn("Recognition error: ", t);
+            completeListener.accept(t);
+            sttListener.sttEventReceived(new SpeechStopEvent());
+            if (!config.errorMessage.isBlank()) {
+                sttListener.sttEventReceived(new SpeechRecognitionErrorEvent(config.errorMessage));
+            } else {
+                String errorMessage = t.getMessage();
+                sttListener.sttEventReceived(
+                        new SpeechRecognitionErrorEvent(errorMessage != null ? errorMessage : "Unknown error"));
```
lolodomo, line 373: Same here

bundles/org.openhab.voice.googlestt/README.md
```diff
@@ -0,0 +1,62 @@
...
+```
+org.openhab.voice.googlestt:clientId=ID
+org.openhab.voice.googlestt:clientSecret=SECRET
+org.openhab.voice.googlestt:authcode=XXXXX
+org.openhab.voice.googlestt:singleUtteranceMode=true
+org.openhab.voice.googlestt:maxTranscriptionSeconds=60
+org.openhab.voice.googlestt:maxSilenceSeconds=5
+org.openhab.voice.googlestt:refreshSupportedLocales=false
+org.openhab.voice.googlestt:noResultsMessage="Sorry, I didn't understand you"
+org.openhab.voice.googlestt:errorMessage="Sorry, something went wrong"
```
lolodomo, line 61: I have a doubt if `"` is expected r not. Can you check that please ?

lolodomo (changes requested):
Review part 3 of 3

bundles/org.openhab.voice.googlestt/src/main/java/org/openhab/voice/googlestt/internal/GoogleSTTService.java
GiviMAD: The user can overwrite this message. Is it not enough?
If not, can you point me to an example? I have no idea how to do that.

bundles/org.openhab.voice.googlestt/README.md
GiviMAD: I never have never configured anything using text files. But this is the content of my generated config file:
```
Miguels-MacBook-Pro:openhab-3 miguelgivi$ cat userdata/config/org/openhab/voice/googlestt.config
:org.apache.felix.configadmin.revision:=L"151"
authcode=""
clientId="******"
clientSecret="*******"
errorMessage="Ha\ habido\ un\ error,\ revisa\ los\ logs\ del\ sistema"
maxSilenceSeconds="10"
maxTranscriptionSeconds="15"
noResultsMessage="Perdona,\ no\ te\ he\ entendido"
oauthCode=""
service.pid="org.openhab.voice.googlestt"
singleUtteranceMode=B"true"
startDialog=B"false"
```

bundles/org.openhab.voice.googlestt/src/main/java/org/openhab/voice/googlestt/internal/GoogleSTTService.java
lolodomo: You're right. 
Keep it like that.

bundles/org.openhab.voice.googlestt/README.md
lolodomo: Ok, so this is certainly correct.

bundles/org.openhab.voice.googlestt/src/main/java/org/openhab/voice/googlestt/internal/GoogleSTTService.java
GiviMAD: That text should never be told as the exceptions should always have a message which we can not translate. But I have to put something there because of nullability. Could we left it like that? User can also customize this.

bundles/org.openhab.voice.googlestt/src/main/java/org/openhab/voice/googlestt/internal/GoogleSTTService.java
GiviMAD: I do not understand this comment. I'm just using one task. Did you mean why I call executor submit inside the function, instead of wrap the method call? I can change it to be a void and call the executor on the recognize method if so.

bundles/org.openhab.voice.googlestt/src/main/java/org/openhab/voice/googlestt/internal/GoogleSTTService.java
lolodomo: Ok, understood. Keep it as it is.

bundles/org.openhab.voice.googlestt/src/main/java/org/openhab/voice/googlestt/internal/GoogleSTTService.java
lolodomo: Oups, this is just a wrong comment, a bad reading of the code.
This looks ok as you did.

lolodomo (approved):
LGTM

lolodomo:
Thank you for your new contribution, the first Speech-To-Text service integrated into openHAB.

Now, you could add your addon's logo to the openHAB website. See https://www.openhab.org/docs/developer/bindings/#add-your-binding-s-logo-to-the-openhab-website

lolodomo:
@GiviMAD : there is something I missed during my review: you should mention your dependency (google-cloud-speech) in the NOTICE file. Please open a PR to update this file.

PR #12056, opened 2022-01-15
[mielecloud] Fix tests by adding missing email configuration parameter

wborn:
The validation did not fail previously because it was broken and fixed in https://github.com/openhab/openhab-core/pull/2683


jlaur (approved):
LGTM

PR #12058, opened 2022-01-15
Upgrade to Karaf 4.3.6

wborn:
* Syncs the karaf.version so the new Maven plugin is used
* Resolves itest runbundles for the new runtime dependencies

---

Related to openhab/openhab-distro#1363

### DIFF
--- a/itests/org.openhab.binding.astro.tests/itest.bndrun
+++ b/itests/org.openhab.binding.astro.tests/itest.bndrun
@@ -40,7 +40,6 @@ Fragment-Host: org.openhab.binding.astro
 	org.objenesis;version='[3.2.0,3.2.1)',\
 	biz.aQute.tester.junit-platform;version='[6.1.0,6.1.1)',\
 	org.apache.felix.scr;version='[2.1.30,2.1.31)',\
-	org.ops4j.pax.logging.pax-logging-api;version='[2.0.12,2.0.13)',\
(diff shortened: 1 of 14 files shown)


wborn:
Hi @jlaur! I've removed the "enhancement" label again because otherwise it will show up twice in the release notes. That's because https://github.com/openhab/openhab-distro/pull/1363 also has that label.

jlaur:
> I've removed the "enhancement" label again because otherwise it will show up twice in the release notes. That's because [openhab/openhab-distro#1363](https://github.com/openhab/openhab-distro/pull/1363) also has that label.

Makes sense, thanks. Do you have a proposal for anyone who can assess this upgrade, as I don't think I can. I checked previous upgrade, #11750, but this one is slightly different. It was merged by @kaikreuzer. If you believe it's safe to merge, I can do that. Still not into all routines.

wborn:
I always try to align the runtime/test dependencies with the new Karaf release in openhab-core as much as possible (so it is easier to reproduce issues in tests/Eclipse). Then in openhab-addons only the itest runbundles need to be resolved and I update the `karaf.version` so it will use the karaf-maven-plugin. So usually you only need to check that the build still succeeds with the updated dependencies and no unrelated changes got commited. :wink:

jlaur (approved):
LGTM

Issue #11852, opened 2021-12-24
[hdpowerview] Bridge handler dynamic channel refactoring


jlaur:
The bridge handler class `HDPowerViewHubHandler` has grown after:
- #11534
- #11637

Especially the logic for creating automation channels is quite large (> 100 lines) and should be extracted to a separate class. This will improve the readability of the class by removing some responsibilities also make it easier to cover by unit tests.

PR #12061, opened 2022-01-16
[hdpowerview] Add support for repeaters

### LINKED ISSUE #12060: [hdpowerview] Repeater support
Repeaters can be added to the PowerView system to improve signal reception for shades too far away from the PowerView Hub (shades with power supplies unfortunately do not serve as repeaters). Repeaters don't do much except for repeating signals, but they are known by the Hub and they can be configured with blinking enabled/disabled as well as identied by flashing the LED. Furthermore, when performing a shade survey for signal strength, the ID of any used repeater is returned, which makes it possible to verify if any given shade is in fact using the repeater.

Supporting the Repeater as a thing would make it possible to:
- Enable/disable blinking during commands. For example this could be used to disable at night when repeater is in a bedroom.
- Flash repeater LED's on request. For example this could be used in combination with a motion sensor to provide position lighting at night.
- See firmware version information.

# API information

## Get all repeaters
**Request URL**: `GET /api/repeaters/`
**Response**:
```
{
	"repeaterIds": [
		17805
	],
	"repeaterData": [
		{
			"id": 17805,
			"blinkEnabled": false,
			"roomId": 23331,
			"groupId": 49464,
			"firmware": {
				"revision": 2,
				"subRevision": 0,
				"build": 2928,
				"index": 33
			},
			"name": "UmVwZWF0ZXIgc292ZXbDpnJlbHNl"
		}
	]
}
```

## Enable blinking
**Request URL**: `PUT /api/repeaters/17805`
**Request body**: `{"repeater":{"blinkEnabled":true,"id":17805}}`
**Response**:
```
{
	"repeater": {
		"id": 17805,
		"blinkEnabled": true,
		"roomId": 23331,
		"groupId": 49464,
		"firmware": {
			"revision": 2,
			"subRevision": 0,
			"build": 2928,
			"index": 33
		},
		"name": "UmVwZWF0ZXIgc292ZXbDpnJlbHNl"
	}
}
```

## Identify
**Request URL**: `GET /api/repeaters/17805?identify=true`
**Response**:
```
{
	"repeater": {
		"id": 17805,
		"blinkEnabled": false,
		"roomId": 23331,
		"groupId": 49464,
		"firmware": {
			"revision": 2,
			"subRevision": 0,
			"build": 2928,
			"index": 33
		},
		"name": "UmVwZWF0ZXIgc292ZXbDpnJlbHNl"
	}
}
```

jlaur:
Fixes #12060

Add support for Hunter Douglas PowerView Repeaters.

Supporting the Repeater as a thing makes it possible to:
- Enable/disable blinking during commands. For example this can be used to disable at night when repeater is in a bedroom.
- Flash repeater LED's on request. For example this could be used in combination with a motion sensor to provide position lighting at night.
- See firmware version information.

### Technical approach
- New thing type **repeater** created (with discovery support).
- Repeaters are polled every 5 minutes to synchronize "blink enabled" state when done from PowerView app or other integrations.
- When receiving repeater information, firmware information is also obtained and set as property.
- When **identify** channel is triggered, response received will be used for immediate update (in between 5 minute polls).
- Each repeater thing will have its own scheduled job. This could be slightly optimized by letting the hub fetch data for all repeaters in same job. On the other hand it would mean more direct dependencies between bridge handler and repeater thing handler. Usually I would expect no more than one or two repeaters in a relatively big house.

#### Thing
image

#### Channels
image

### Additional refactoring

After testing with frequent binding restarts and configuration changes, I noticed a few warnings on follow-up log inspection:

```
2022-01-18 20:01:04.197 [TRACE] [rview.internal.HDPowerViewWebTargets] - API command GET http://192.168.0.239/api/shades/0?refresh=true
2022-01-18 20:01:04.240 [WARN ] [rview.internal.HDPowerViewWebTargets] - Hub returned HTTP 403 Forbidden
2022-01-18 20:01:04.245 [WARN ] [rnal.handler.HDPowerViewShadeHandler] - Unexpected error: HTTP 403 error
```

I have not yet found the root cause, and I do not suspect the changes in this PR to be causing this behavior. But nevertheless I would like to get to the bottom of this, so as first step I simplified the configuration **ID** handling to be more robust by changing it from string to integer and adding validation. Doing this for the new repeater thing was fully within the scope of this PR, but for consistency I decided to perform the same refactoring for the shade thing. I have tested backwards compatibility for the change by creating a thing with version 3.2 of the binding from UI, then migrated to the PR version and verified that thing still works correctly.

### DIFF
--- a/bundles/org.openhab.binding.hdpowerview/README.md
+++ b/bundles/org.openhab.binding.hdpowerview/README.md
@@ -102,7 +102,7 @@ All of these channels appear in the binding, but only those which have a physica
 
 | Channel         | Item Type | Description                   |
 |-----------------|-----------|-------------------------------|
-| identify        | Switch    | Flash repeater to identify. Setting this to ON will identify the repeater by blinking. Note: include `{autoupdate="false"}` in the item configuration to avoid having to reset it to off after use. |
+| identify        | String    | Flash repeater to identify. Valid values are: IDENTIFY |
 | blinkingEnabled | Switch    | Blink during commands.        |
 
 ### Roller Shutter Up/Down Position vs. Open/Close State
@@ -227,7 +227,7 @@ Switch Living_Room_Shade_Calibrate "Living Room Shade Calibrate" {channel="hdpow
 Repeater items:
 
 ```
-Switch Bedroom_Repeater_Identify "Bedroom Repeater Identify" {channel="hdpowerview:repeater:g24:r16384:identify", autoupdate="false"}
+String Bedroom_Repeater_Identify "Bedroom Repeater Identify" {channel="hdpowerview:repeater:g24:r16384:identify"}
 Switch Bedroom_Repeater_BlinkingEnabled "Bedroom Repeater Blinking Enabled [%s]" {channel="hdpowerview:repeater:g24:r16384:blinkingEnabled"}
 ```
 
@@ -240,8 +240,12 @@ Switch Living_Room_Shades_Scene_Heart "Living Room Shades Scene Heart" <blinds> 
 ### `demo.sitemap` File
 
 ```
-Frame label="Living Room Shades" {
-  Switch item=Living_Room_Shades_Scene_Open
-  Slider item=Living_Room_Shade_1_Position 
+Frame label="Living Room Shades" {
+    Switch item=Living_Room_Shades_Scene_Open
+    Slider item=Living_Room_Shade_1_Position
+}
+Frame label="Bedroom" {
+    Switch item=Bedroom_Repeater_Identify mappings=[IDENTIFY="Identify"]
+    Switch item=Bedroom_Repeater_BlinkingEnabled
 }
 ```
--- a/bundles/org.openhab.binding.hdpowerview/README.md
+++ b/bundles/org.openhab.binding.hdpowerview/README.md
@@ -13,10 +13,11 @@ By using a scene to control multiple shades at once, the shades will all begin m
 
 ## Supported Things
 
-| Thing | Thing Type | Description |
-|-------|------------|-------------|
-| hub   | Bridge     | The PowerView hub provides the interface between your network and the shade's radio network. It also contains channels used to interact with scenes. |
-| shade | Thing      | A motorized shade. |
+| Thing    | Thing Type | Description |
+|----------|------------|-------------|
+| hub      | Bridge     | The PowerView hub provides the interface between your network and the shade's radio network. It also contains channels used to interact with scenes. |
+| shade    | Thing      | A motorized shade. |
+| repeater | Thing      | A PowerView signal repeater. |
 
 ## Discovery
 
@@ -46,15 +47,24 @@ If in the future, you add additional shades or scenes to your system, the bindin
 | hardRefresh             | The number of minutes between hard refreshes of the PowerView hub's shade state (default 180 three hours). See [Refreshing the PowerView Hub Cache](#Refreshing-the-PowerView-Hub-Cache). |
 | hardRefreshBatteryLevel | The number of hours between hard refreshes of battery levels from the PowerView Hub (or 0 to disable, defaulting to weekly). See [Refreshing the PowerView Hub Cache](#Refreshing-the-PowerView-Hub-Cache). |
 
-### Thing Configuration for PowerView Shades
+### Thing Configuration for PowerView Shades and Accessories
 
-PowerView shades should preferably be configured via the automatic discovery process.
-It is quite difficult to configure manually as the `id` of the shade is not exposed in the PowerView app.
-However, the configuration parameters are described below:
+PowerView shades and repeaters should preferably be configured via the automatic discovery process.
+It is quite difficult to configure manually as the `id` of the shade or repeater is not exposed in the
+PowerView app. However, the configuration parameters are described below.
+
+#### Thing Configuration for PowerView Shades
 
 | Configuration Parameter | Description |
 |-------------------------|-------------|
 | id                      | The ID of the PowerView shade in the app. Must be an integer. |
+
+#### Thing Configuration for PowerView Repeaters
+
+
+| Configuration Parameter | Description |
+|-------------------------|-------------|
+| id                      | The ID of the PowerView repeater in the app. Must be an integer. |
 
 ## Channels
 
@@ -87,6 +97,13 @@ All of these channels appear in the binding, but only those which have a physica
 | batteryLevel   | Number                   | Battery level (10% = low, 50% = medium, 100% = high)
 | batteryVoltage | Number:ElectricPotential | Battery voltage reported by the shade. |
 | signalStrength | Number                   | Signal strength (0 for no or unknown signal, 1 for weak, 2 for average, 3 for good or 4 for excellent) |
+
+### Channels for Repeaters (Thing type `repeater`)
+
+| Channel         | Item Type | Description                   |
+|-----------------|-----------|-------------------------------|
+| identify        | Switch    | Flash repeater to identify. Setting this to ON will identify the repeater by blinking. Note: include `{autoupdate="false"}` in the item configuration to avoid having to reset it to off after use. |
+| blinkingEnabled | Switch    | Blink during commands.        |
 
 ### Roller Shutter Up/Down Position vs. Open/Close State
 
@@ -191,6 +208,7 @@ For single shades the refresh takes the item's channel into consideration:
 ```
 Bridge hdpowerview:hub:g24 "Luxaflex Hub" @ "Living Room" [host="192.168.1.123"] {
     Thing shade s50150 "Living Room Shade" @ "Living Room" [id="50150"]
+    Thing repeater r16384 "Bedroom Repeater" @ "Bedroom" [id="16384"]
 }
 ```
 
@@ -200,14 +218,17 @@ Shade items:
 
 ```
 Rollershutter Living_Room_Shade_Position "Living Room Shade Position [%.0f %%]" {channel="hdpowerview:shade:g24:s50150:position"}
-
 Rollershutter Living_Room_Shade_Secondary "Living Room Shade Secondary Position [%.0f %%]" {channel="hdpowerview:shade:g24:s50150:secondary"}
-
 Dimmer Living_Room_Shade_Vane "Living Room Shade Vane [%.0f %%]" {channel="hdpowerview:shade:g24:s50150:vane"}
-
 Switch Living_Room_Shade_Battery_Low_Alarm "Living Room Shade Battery Low Alarm [%s]" {channel="hdpowerview:shade:g24:s50150:lowBattery"}
-
 Switch Living_Room_Shade_Calibrate "Living Room Shade Calibrate" {channel="hdpowerview:shade:g24:s50150:calibrate", autoupdate="false"}
+```
+
+Repeater items:
+
+```
+Switch Bedroom_Repeater_Identify "Bedroom Repeater Identify" {channel="hdpowerview:repeater:g24:r16384:identify", autoupdate="false"}
+Switch Bedroom_Repeater_BlinkingEnabled "Bedroom Repeater Blinking Enabled [%s]" {channel="hdpowerview:repeater:g24:r16384:blinkingEnabled"}
 ```
 
 Scene items:
--- a/bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/handler/HDPowerViewRepeaterHandler.java
+++ b/bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/handler/HDPowerViewRepeaterHandler.java
@@ -57,10 +57,9 @@ public class HDPowerViewRepeaterHandler extends AbstractHubbedThingHandler {
 
     @Override
     public void initialize() {
-        logger.debug("Initializing repeater handler");
-        try {
-            repeaterId = getRepeaterId();
-        } catch (NumberFormatException e) {
+        repeaterId = getConfigAs(HDPowerViewRepeaterConfiguration.class).id;
+        logger.debug("Initializing repeater handler for repeater {}", repeaterId);
+        if (repeaterId <= 0) {
             updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.CONFIGURATION_ERROR,
                     "@text/offline.conf-error.invalid-id");
             return;
@@ -88,14 +87,6 @@ public class HDPowerViewRepeaterHandler extends AbstractHubbedThingHandler {
     public void dispose() {
         logger.debug("Disposing repeater handler for repeater {}", repeaterId);
         cancelJob();
-    }
-
-    private int getRepeaterId() throws NumberFormatException {
-        String str = getConfigAs(HDPowerViewRepeaterConfiguration.class).id;
-        if (str == null) {
-            throw new NumberFormatException("null input string");
-        }
-        return Integer.parseInt(str);
     }
 
     @Override
--- a/bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/handler/HDPowerViewRepeaterHandler.java
+++ b/bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/handler/HDPowerViewRepeaterHandler.java
@@ -192,7 +192,6 @@ public class HDPowerViewRepeaterHandler extends AbstractHubbedThingHandler {
         } catch (HubMaintenanceException e) {
             // exceptions are logged in HDPowerViewWebTargets
         } catch (HubException e) {
-            logger.warn("Error connecting to bridge: {}", e.getMessage());
             updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.BRIDGE_OFFLINE, e.getMessage());
         }
     }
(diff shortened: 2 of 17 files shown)


andrewfg:
@jlaur unfortunately I don't have any repeaters so I cannot test the new functionality, but I will at least check that existing functionality (normal shades) has not been broken :)

jlaur:
> unfortunately I don't have any repeaters so I cannot test the new functionality

I would say fortunately. Not needing them is much desired. :-)

> I will at least check that existing functionality (normal shades) has not been broken :)

Thanks. I have it running in production now, but would like to run it for some more time before marking as ready. Since it's a new thing type everything is quite isolated and existing functionality should not be affected much. But I already found an assumption in `HDPowerViewHubHandler` about all things being of type **shade**, so you never know.

andrewfg:
A trivial issue, but one of your JUnit tests is issuing an unexpected logger message..

image

jlaur:
> A trivial issue, but one of your JUnit tests is issuing an unexpected logger message..

I know, but didn't think it would be a problem since it's not considered a warning? I'll have another look at this and see if I can mock logger somehow. Please don't let it block this PR as it was introduced by #11853 which is already merged. I'll provide a new PR for this issue and link it to #11852 once I have found a solution.

andrewfg:
> see if I can mock logger somehow

Perhaps one can set a lower logger level when running JUnit tests (e.g. `log:set ERROR` or even `log:set OFF`), or add an argument to the method being called (e.g. `method(..., boolean loggingEnabled)` -- although this latter is a bit clunky..)

> don't let it block this PR

Ok with me.

jlaur:
> > see if I can mock logger somehow
> 
> Perhaps one can set a lower logger level when running JUnit tests (e.g. `log:set ERROR` or even `log:set OFF`), or add an argument to the method being called (e.g. `method(..., boolean loggingEnabled)` -- although this latter is a bit clunky..)

I'm definitely against the latter. As for log level manipulration, I could imagine setting log level this way could introduce problems with parallel test runs, but not sure. Also it wouldn't be able to test if something was indeed logged.

Initially I tried to create some mocks based on `LoggerFactory.getILoggerFactory()` but failed. I looked into a few other approaches, and ended up with this static scoped one as being the cleanest and simplest for this scenario:

```
    @Test
    public void emptyListWhenNoSceneForScheduledEvent() {
        try (MockedStatic<LoggerFactory> staticMock = mockStatic(LoggerFactory.class)) {
            @Nullable
            final Logger logger = mock(Logger.class);
            staticMock.when(() -> LoggerFactory.getLogger(any(Class.class))).thenReturn(logger);

            builder = AutomationChannelBuilder.create(translationProvider, CHANNEL_GROUP_UID);
            ScheduledEvent scheduledEvent = createScheduledEventWithSceneCollection(
                    ScheduledEvents.SCHEDULED_EVENT_TYPE_SUNRISE);
            List<ScheduledEvent> scheduledEvents = new ArrayList<>(List.of(scheduledEvent));
            List<Channel> channels = builder.withScenes(scenes).withScheduledEvents(scheduledEvents).build();

            assertEquals(0, channels.size());
            Logger verifyLogger = verify(logger);
            assertNotNull(verifyLogger);
            if (verifyLogger != null) {
                verifyLogger.warn(any(String.class), any(), any());
            }
        }
    }
```

`verify()` can of course be adapted to needs, for this PoC it just verifies that any warning with two additional parameters is logged.

I was able to run it after modifying **pom.xml** for the binding:
image

Now wondering if **mockito-inline** could/should be added to:
https://github.com/openhab/openhab-core/blob/main/bom/test/pom.xml

which is referenced from here:
https://github.com/openhab/openhab-addons/blob/main/bom/test-index/pom.xml

@wborn - any opinion/advice here, i.e. if it would be okay to include **mockito-inline**, and if yes, in which POM?

andrewfg:
IMHO you should certainly not add a mockito dependency to the binding pom..

Perhaps another solution is to have JUnit call an inner class for the test, that does not do any logging, but throws an exception instead; and then let the binding itself call an outer wrapper class that catches that exception and logs the warning.

PS notwithstanding the above, I can tell you that the jar for this build has been running on my operative system for 6 hours or so, and seems to be working as it should. (Although, actually I personally find it rather uncanny that when one moves the shades, the OH UI updates to the new position immediately — even though the shades have not actually achieved that position yet..)

jlaur:
> IMHO you should certainly not add a mockito dependency to the binding pom..

We fully agree. It was a Proof-of-Concept, suggestion was to add it to the linked POM:
https://github.com/openhab/openhab-core/blob/main/bom/test/pom.xml

> Perhaps another solution is to have JUnit call an inner class for the test, that does not do any logging, but throws an exception instead; and then let the binding itself call an outer wrapper class that catches that exception and logs the warning.

I'm not sure I understand what you mean by inner/outer classes, do you mean changing `AutomationChannelBuilder` itself? I rather think it should stay as a single class without any need to extract that private method to a new class just for sake of testing. Also would not throw/catch exceptions within a loop to handle such simple internal cases. Instead this could be changed to some try* pattern, though. The public method should be enough for testing that functionality within the private method also, we just need to mock the logger, and as I've shown, this is certainly possible.

> PS notwithstanding the above, I can tell you that the jar for this build has been running on my operative system for 6 hours or so, and seems to be working as it should.

Thanks for giving it a shot!

> (Although, actually I personally find it rather uncanny that when one moves the shades, the OH UI updates to the new position immediately — even though the shades have not actually achieved that position yet..)

Hmm, I think you introduced that behavior in #11698? It was slightly changed in #12002 by using the response directly instead of requesting positions again by sending a new request. I'm open for discussing this, it could even easily be reverted/changed/refined. However, please consider:
- We have no way of knowing how long it takes to move the shade, i.e. when it will reach its target position.
- The position will be updated in the hub immediately.

So with the new approach we introduced, we'll synchronize with the hub immediately. Previously (before your change) it would be synchronized at a random moment after moving the shade, i.e. upon next poll. So if next poll would be one second after moving the shade, it would be updated then. But it could also have to wait up to one minute.

andrewfg:
> not sure I understand what you mean by inner/outer classes

Oops. I meant to say “inner/outer method”. Sorry for the confusion.

> introduced that behavior in #11698?

Yes indeed. But it was not my request, and came rather from the guy who opened the issue. As said, I personally find it uncanny … but I can live with it..

jlaur:
@andrewfg - I did a new commit, 271d0271c23c1c4a47685e05d0649710f6f6cbf7, today after some suspicious warnings in my logs. I still don't know what caused that, but I'm keeping an eye on it. Have you (ever) seen anything like that? Unfortunately I'm not keeping old logs, so can't check first occurrence.

wborn:
> any opinion/advice here, i.e. if it would be okay to include mockito-inline instead of mockito-core

I always stay away from expirimental or incubating features because they can break with any upgrade. We've seen incubating mockito/junit functionality break before.

jlaur:
@andrewfg - would you be able to review this PR before assigning it to maintainers for review? Asking because I now prepared a fix for #12071 but it will conflict, so will hold that back for a while.

andrewfg (approved):
I don't have a repeater so could not test that Thing type; however I can confirm that the original Shade Things are (still) working.

jlaur:
@lolodomo - looking at recent hdpowerview PR review history, would this be something for you to have a look at? :-)

jlaur:
@openhab/add-ons-maintainers - anyone willing to review this PR? I have some additional work planned for another PR, but need this one merged first.

bundles/org.openhab.binding.hdpowerview/README.md
```diff
@@ -88,6 +98,13 @@ All of these channels appear in the binding, but only those which have a physica
 | batteryVoltage | Number:ElectricPotential | Battery voltage reported by the shade. |
 | signalStrength | Number                   | Signal strength (0 for no or unknown signal, 1 for weak, 2 for average, 3 for good or 4 for excellent) |
 
+### Channels for Repeaters (Thing type `repeater`)
+
+| Channel         | Item Type | Description                   |
+|-----------------|-----------|-------------------------------|
+| identify        | Switch    | Flash repeater to identify. Setting this to ON will identify the repeater by blinking. Note: include `{autoupdate="false"}` in the item configuration to avoid having to reset it to off after use. |
```
fwolter, line 105: You could avoid the autoupdate quirk by making this a command channel. See https://www.openhab.org/docs/developer/bindings/thing-xml.html#command-description

fwolter (changes requested):

fwolter:
> @openhab/add-ons-maintainers - anyone willing to review this PR? I have some additional work planned for another PR, but need this one merged first.

You could base your work on this PR's branch and add the "awaiting other PR" label for your new PR.

bundles/org.openhab.binding.hdpowerview/README.md
jlaur: Thanks for this hint. I agree this is more elegant. Changed and tested. I created only one command, **IDENTIFY**, and set the state back to **UNDEF** after three seconds, which seems to match the period where the repeater blinks. I considered **NONE** as command, but IMHO it doesn't make sense, since the blinking can't be stopped anyway. How does it look for you?

jlaur:
> You could base your work on this PR's branch and add the "awaiting other PR" label for your new PR.

True. I'm just not a fan of such dependency chains as there could also be other things merged in the meantime. Or this one might not be merged at all. So when preparing small changes with big risk of conflicts, I tend to prefer waiting. But thanks for the hint!

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/handler/HDPowerViewRepeaterHandler.java
```diff
@@ -0,0 +1,218 @@
...
+            updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.BRIDGE_UNINITIALIZED,
+                    "@text/offline.conf-error.invalid-bridge-handler");
+            return;
+        }
+        ThingStatus bridgeStatus = bridge.getStatus();
+        if (bridgeStatus == ThingStatus.ONLINE) {
+            updateStatus(ThingStatus.UNKNOWN);
+        } else {
+            updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.BRIDGE_OFFLINE);
+        }
```
fwolter, line 87: The statuses of the child Things are automatically set by the framework according their bridge's status. This works if you set the Thing status in `initialize()` to at least UNKNOWN, i.e. the Thing status must not be INITIALIZING when leaving `initialize()`.

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/handler/HDPowerViewRepeaterHandler.java
```diff
@@ -0,0 +1,218 @@
...
+        scheduleRefreshJob();
+    }
+
+    @Override
+    public void dispose() {
+        logger.debug("Disposing repeater handler for repeater {}", repeaterId);
```
fwolter, line 93: This is already logged by the framework and can therefore be removed.

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/handler/HDPowerViewRepeaterHandler.java
```diff
@@ -0,0 +1,218 @@
...
+            if (cause == null) {
+                logger.warn("Bridge returned a bad JSON response: {}", e.getMessage());
+            } else {
+                logger.warn("Bridge returned a bad JSON response: {} -> {}", e.getMessage(), cause.getMessage());
+            }
+        } catch (HubMaintenanceException e) {
+            // exceptions are logged in HDPowerViewWebTargets
+        } catch (HubException e) {
+            logger.warn("Error connecting to bridge: {}", e.getMessage());
+            updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.BRIDGE_OFFLINE, e.getMessage());
```
fwolter, line 201: The log message could be removed, as the status update is already logged by the framework.

The state change originated by updateStatus() is logged to events.log. Including the status detail message.

bundles/org.openhab.binding.hdpowerview/src/main/resources/OH-INF/thing/thing-types.xml
```diff
@@ -75,13 +75,40 @@
 		<representation-property>id</representation-property>
 
 		<config-description>
-			<parameter name="id" type="text" required="true">
+			<parameter name="id" type="integer" min="1" required="true">
```
fwolter, line 78: Is this backward compatible?

bundles/org.openhab.binding.hdpowerview/src/main/resources/OH-INF/thing/thing-types.xml
```diff
@@ -122,6 +149,23 @@
...
+		<command>
+			<options>
+				<option value="IDENTIFY">Identify</option>
+			</options>
+		</command>
+	</channel-type>
+
+	<channel-type id="repeater-blinking-enabled">
+		<item-type>Switch</item-type>
+		<label>Blinking enabled</label>
```
fwolter, line 165:
```suggestion
		<label>Blinking Enabled</label>
```

fwolter (changes requested):
The command looks good in general. Did you try to set the auto update policy to `veto` to prevent the necessity to reset the Channel value manually? See https://www.openhab.org/docs/developer/bindings/thing-xml.html#auto-update-policies

jlaur:
> The command looks good in general. Did you try to set the auto update policy to `veto` to prevent the necessity to reset the Channel value manually? See https://www.openhab.org/docs/developer/bindings/thing-xml.html#auto-update-policies

No, was unaware - thanks. Now after reading it, I think it will work better with current approach, since state will be kept for the duration of the blinking period. This looks kind of cool in the UI. :-) I don't think it's possible to set a timer using the policies?

bundles/org.openhab.binding.hdpowerview/src/main/resources/OH-INF/thing/thing-types.xml
jlaur: Yes, but hidden as last sentence in the PR description: 
> I have tested backwards compatibility for the change by creating a thing with version 3.2 of the binding from UI, then migrated to the PR version and verified that thing still works correctly.

fwolter:
There's no timer for update policies I'm aware of. Your approach is good to go.

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/handler/HDPowerViewRepeaterHandler.java
jlaur: Thanks, fixed.

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/handler/HDPowerViewRepeaterHandler.java
jlaur: Good point, removed.

bundles/org.openhab.binding.hdpowerview/src/main/resources/OH-INF/thing/thing-types.xml
jlaur: Fixed.

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/handler/HDPowerViewRepeaterHandler.java
jlaur: I can not find this being logged? At least not anything including the ID.

jlaur: Do you indirectly like this?
```
2022-01-24 21:45:32.507 [DEBUG] [l.handler.HDPowerViewRepeaterHandler] - Disposing repeater handler for repeater 17805
2022-01-24 21:45:32.520 [INFO ] [ab.event.ThingStatusInfoChangedEvent] - Thing 'hdpowerview:repeater:hub:repeater' changed from UNINITIALIZED to UNINITIALIZED (HANDLER_MISSING_ERROR)
```

Maybe it would still be useful with the distinct dispose() logging including ID as long as it's only with debug level?

fwolter: OK

fwolter (approved):
LGTM

jlaur:
> Is it ok to simply override all Crowdin translations by those in Github?

Yes, at least the Danish ("_da") ones which seems missing in Crowdin. I'm not sure if there are any other translations in Crowdin for hdpowerview not yet in the repository, but probably not? Can't figure out how to see all translations for a single bundle.

lolodomo:
Translations uploaded in Crowdin. Only 29% of translations are completed.

jlaur:
> Translations uploaded in Crowdin.

Thanks!

Do we have any guidelines or documentation for the processes involving Crowdin synchronization/workflows? By now I understand that default (English) properties are one-way synced from repository to Crowdin. Then translations are synced back to repository, and it creates some kind of trouble if they are going the other way - perhaps simply because there is no automation for that? But some details would be nice for the understanding. Also what it takes to assess if a Crowin PR is safe to merge. I can help write it if it doesn't exist, but first I would need some more knowledge. :)

lolodomo:
Default (English) properties are one-way automatically synced from repository to Crowdin.
Non default properties are one-way automatically synced from Crowdin to repository when translations are fully completed in Crowdin.
Problems are introduced if non default translations are merged directly in repository (not coming from Crowdin). That is the reason why we have to reject PRs including such translations.
In Crowdin, like I did for the Danish translations, you have a command to upload translations from a file.
@wborn and @cweitkamp are the ones who have the best knowledge of all of that for the project.

Issue #12064, opened 2022-01-17
[oppo] Play Mode and Disc Type channels update inconsistently


mlobstein:
When the player changes states the Play Mode and Disc Type channels do not always update correctly. The states of these channels will remain on the previous value. The issue occurs on a BDP-103 using a direct serial port connection. A pull request to fix the issue has been opened.

lolodomo:
Fixed by #12066

PR #12066, opened 2022-01-17
[oppo] Fix Play Mode and Disc Type updates

mlobstein:
fixes issue #12064 
In some configurations, the player sends unsolicited updates but the update message is an abbreviated code when compared to the longer response provided by the query command that must used in other configurations. The previous strategy to immediately call the query command upon receiving an unsolicited update was flawed in that the query command sometimes provided an outdated value. This PR fixes the issue for Play Mode and Disc Type by mapping the abbreviated unsolicited update code to the values provided by the query command. For Audio Type and Subtitle Type, a slight delay was added before calling the query command.

### DIFF
--- a/bundles/org.openhab.binding.oppo/src/main/java/org/openhab/binding/oppo/internal/communication/OppoStatusCodes.java
+++ b/bundles/org.openhab.binding.oppo/src/main/java/org/openhab/binding/oppo/internal/communication/OppoStatusCodes.java
@@ -12,6 +12,8 @@
  */
 package org.openhab.binding.oppo.internal.communication;
 
+import static org.openhab.binding.oppo.internal.OppoBindingConstants.*;
+
 import java.util.HashMap;
 import java.util.Map;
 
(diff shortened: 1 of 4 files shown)


bundles/org.openhab.binding.oppo/src/main/java/org/openhab/binding/oppo/internal/handler/OppoHandler.java
```diff
@@ -506,16 +514,17 @@ public void onNewMessageEvent(OppoMessageEvent evt) {
...
-                        // we got the disc type status update, throw it away
-                        // and call the query because the text output is better
-                        connector.sendCommand(OppoCommand.QUERY_DISC_TYPE);
                     case QDT:
-                        currentDiscType = updateData;
-                        updateChannelState(CHANNEL_DISC_TYPE, updateData);
+                        // try to normalize the slightly different responses between UDT and QDT
+                        final String discType = OppoStatusCodes.DISC_TYPE.get(updateData);
+                        currentDiscType = (discType != null ? discType : updateData);
+                        updateChannelState(CHANNEL_DISC_TYPE, (discType != null ? discType : updateData));
```
lolodomo, line 521: Rather use currentDiscType.

bundles/org.openhab.binding.oppo/src/main/java/org/openhab/binding/oppo/internal/communication/OppoStatusCodes.java
```diff
@@ -54,4 +56,36 @@
         ZOOM_MODE.put("11", "1/3");
         ZOOM_MODE.put("12", "1/4");
     }
+
+    // map to lookup disc type
+    public static final Map<String, String> DISC_TYPE = new HashMap<>();
+    static {
```
lolodomo, line 62: No chance to use Map.of? 
Maybe too many entries...

lolodomo (changes requested):

bundles/org.openhab.binding.oppo/src/main/java/org/openhab/binding/oppo/internal/handler/OppoHandler.java
mlobstein: yes, much simpler. fixed.

bundles/org.openhab.binding.oppo/src/main/java/org/openhab/binding/oppo/internal/communication/OppoStatusCodes.java
mlobstein: Yes this constructor allows a max of 10 entries. If more than that, Map.ofEntries would have to be used.

lolodomo: So I let you change the code.

mlobstein: done

lolodomo (approved):
LGTM

PR #12070, opened 2022-01-17
[tesla] Fix NPE when Thing configuration changes

### LINKED ISSUE #10697: [tesla] NPE in TeslaVehicleHandler
I found this in my log file (openHAB 3.1.0.M4):
```
        at org.openhab.binding.tesla.internal.handler.TeslaVehicleHandler.getVehicleId(TeslaVehicleHandler.java:226) ~[?:?]
        at org.openhab.binding.tesla.internal.handler.TeslaAccountHandler$Request.run(TeslaAccountHandler.java:465) ~[?:?]
        at org.openhab.binding.tesla.internal.throttler.QueueChannelThrottler.lambda$0(QueueChannelThrottler.java:45) ~[?:?]
        at java.lang.Thread.run(Thread.java:834) [?:?]```

### PR DESCRIPTION
**Pull request by kaikreuzer**
Fixes #10697

### DIFF
--- a/bundles/org.openhab.binding.tesla/src/main/java/org/openhab/binding/tesla/internal/handler/TeslaVehicleHandler.java
+++ b/bundles/org.openhab.binding.tesla/src/main/java/org/openhab/binding/tesla/internal/handler/TeslaVehicleHandler.java
@@ -223,7 +223,11 @@ public class TeslaVehicleHandler extends BaseThingHandler {
      * @return the vehicle id
      */
     public String getVehicleId() {
-        return vehicle.id;
+        if (vehicle != null) {
+            return vehicle.id;
+        } else {
+            return null;
+        }
     }
 
     @Override

### CODE REVIEW

File: bundles/org.openhab.binding.tesla/src/main/java/org/openhab/binding/tesla/internal/handler/TeslaVehicleHandler.java [hunk 1]
```diff
@@ -223,7 +223,11 @@ public void dispose() {
      * @return the vehicle id
      */
     public String getVehicleId() {
-        return vehicle.id;
+        if (vehicle != null) {
+            return vehicle.id;
+        } else {
+            return null;
```
jlaur, line 229: It looks valid as a bugfix. Just consider:
- If null is returned here, this is directly passed to `invokeAndParse()` as first parameter.
- `invokeAndParse()` handles this null parameter and returns null.
- When null it returned by `invokeAndParse()` this is also handled.

It works, but... Calling method will a null parameter just for that method to return null doesn't seem clean. At some point it must be considered an error and handled somehow, or new NPE's will likely be introduced. Not a request for change in context of this PR, just sharing thoughts.

jlaur (approved):
LGTM

bundles/org.openhab.binding.tesla/src/main/java/org/openhab/binding/tesla/internal/handler/TeslaVehicleHandler.java
kaikreuzer: I agree it is not ideal, but the fact that this method accesses a field for returning its result is in general a bad design - as it has no influence over whether this is set or not. So returning `null` here in case it isn't (yet) initialized is the most logical thing to do witthout going into a more complex refactoring.

jlaur: I agree, more refactoring would be needed to improve this, so fully accepted as a bugfix. :)

PR #11720, opened 2021-12-06
[homekit] restart HomeKit bridge on network changes

yfre:
this PR:
- add listener to network changes and restart HomeKit bridge on changes. This will recover HomeKit bridge after network connection losses and IP address changes. this will fix the issue reported here https://community.openhab.org/t/oh3-homekit-bug-fails-after-network-down-up/
- make logger messages consistent with small letter.

Signed-off-by: Eugen Freiter <freiter@gmx.de>

PR #12072, opened 2022-01-18
[homekit] quick restart on network changes

yfre:
homekit binding makes currently a full restart on network changes like IP address changes.
the full restart includes removing all openhab items from the list of homekit accessories and adding them again. basically, the start/stop methods for binding start/stop were re-used

this could lead in some case to config reset in home all, like reported here
https://community.openhab.org/t/homekit-no-response-error-on-all-openhab-items/118200/172

this PR reduces the stop logic to minimum and especially skip the removal of homekit accessories on stop.

the changes are following:
- slight rework on "start" function in order to remember the address the HomeKit was bound to
- replace call of the standard "stop" function copy the few needed step to onChange function
- add check to onChange function whether the removed address is the address HomeKit was bound to

### DIFF
--- a/bundles/org.openhab.io.homekit/src/main/java/org/openhab/io/homekit/internal/HomekitImpl.java
+++ b/bundles/org.openhab.io.homekit/src/main/java/org/openhab/io/homekit/internal/HomekitImpl.java
@@ -14,6 +14,7 @@ package org.openhab.io.homekit.internal;
 
 import java.io.IOException;
 import java.net.InetAddress;
+import java.net.UnknownHostException;
 import java.security.InvalidAlgorithmParameterException;
 import java.util.ArrayList;
 import java.util.Dictionary;
@@ -116,10 +117,7 @@ public class HomekitImpl implements Homekit, NetworkAddressChangeListener {
         if (props == null) { // if null, the configuration is new
             props = new Hashtable<>();
         }
-        if (settings.networkInterface == null) {
-            settings.networkInterface = networkAddressService.getPrimaryIpv4HostAddress();
-            props.put("networkInterface", settings.networkInterface);
-        }
+
         if (settings.setupId == null) { // generate setupId very first time
             settings.setupId = HAPSetupCodeUtils.generateSetupId();
             props.put("setupId", settings.setupId);
@@ -210,13 +208,12 @@ public class HomekitImpl implements Homekit, NetworkAddressChangeListener {
     private void startHomekitServer() throws IOException {
         logger.trace("start HomeKit bridge");
         if (homekitServer == null) {
-            if (settings.useOHmDNS) {
-                if ((settings.networkInterface == null) || (settings.networkInterface.isEmpty())) {
-                    logger.trace(
-                            "no IP address configured in HomeKit settings. HomeKit will use the first configured address of openHAB");
-                    homekitServer = new HomekitServer(mdnsClient.getClientInstances().iterator().next(), settings.port);
-                } else {
-                    networkInterface = InetAddress.getByName(settings.networkInterface);
+            try {
+                networkInterface = InetAddress
+                        .getByName(((settings.networkInterface != null) && (!settings.networkInterface.isEmpty()))
+                                ? settings.networkInterface
+                                : networkAddressService.getPrimaryIpv4HostAddress());
+                if (settings.useOHmDNS) {
                     for (JmDNS mdns : mdnsClient.getClientInstances()) {
                         if (mdns.getInetAddress().equals(networkInterface)) {
                             logger.trace("suitable mDNS client for IP {} found and will be used for HomeKit",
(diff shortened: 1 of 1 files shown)


jlaur:
Regression of #11720, looking at this might ease review.

Sail-Away:
Some additional hints: in my logs I see that the reaction to short network interruptions is somewhat random. One event either causes no restart, one restart or two consecutive restarts. Looks like, depending on the actual timing, the network change listener either doesn't catch the event at all, or just one added or removed interface event, or both the removed and added events.

So would it make sense to add some debouncing?

Also, what I don't understand: why would we restart the bridge at all if the interface has just been removed, or if a different, unused network interface changes? Shouldn't a restart only be necessary if the used interface actually _changes_ to a different address? However, this would make it necessary to evaluate the reoved and added, so it's not as straight forward as the current solution.

yfre:
> Some additional hints: in my logs I see that the reaction to short network interruptions is somewhat random. One event either causes no restart, one restart or two consecutive restarts. Looks like, depending on the actual timing, the network change listener either doesn't catch the event at all, or just one added or removed interface event, or both the removed and added events.
>
> So would it make sense to add some debouncing?
not sure about this behaviour. my idea was to keep it the logic small and simple so that we can quickly restart on every event. 



> Also, what I don't understand: why would we restart the bridge at all if the interface has just been removed, or if a different, unused network interface changes? Shouldn't a restart only be necessary if the used interface actually _changes_ to a different address? However, this would make it necessary to evaluate the reoved and added, so it's not as straight forward as the current solution.

good idea. it just one check which does not make it too complex but maybe reduce number of restart. added to PR

bundles/org.openhab.io.homekit/src/main/java/org/openhab/io/homekit/internal/HomekitImpl.java
```diff
@@ -291,13 +291,24 @@ public void clearHomekitPairings() {
...
-            startHomekitServer();
-        } catch (IOException e) {
-            logger.warn("could not initialize HomeKit bridge: {}", e.getMessage());
+    public void onChanged(final List<CidrAddress> added, final List<CidrAddress> removed) {
+        logger.trace("restarting HomeKit bridge on network interface changes.");
+        removed.forEach(i -> {
+            logger.trace("removed interface {}", i.getAddress().toString());
+            if (i.getAddress().equals(networkInterface)) {
+                this.bridge.stop();
+                this.homekitServer.stop();
```
lolodomo, line 300: No need to check that this.bridge and this.homekitServer are not null ?

lolodomo (changes requested):

bundles/org.openhab.io.homekit/src/main/java/org/openhab/io/homekit/internal/HomekitImpl.java
yfre: valid point. fixed

bundles/org.openhab.io.homekit/src/main/java/org/openhab/io/homekit/internal/HomekitImpl.java
```diff
@@ -296,10 +296,16 @@ public void onChanged(final List<CidrAddress> added, final List<CidrAddress> rem
...
+                final @Nullable HomekitRoot bridge = this.bridge;
+                if (this.bridge != null) {
+                    this.bridge.stop();
+                    this.bridge = null;
+                }
+                final @Nullable HomekitServer homekitServer = this.homekitServer;
+                if (homekitServer != null) {
+                    this.homekitServer.stop();
+                    this.homekitServer = null;
+                }
```
lolodomo, line 308: The proper way to avoid compilation warnings is:
```
                final @Nullable HomekitRoot bridge = this.bridge;
                if (bridge != null) {
                    bridge.stop();
                    this.bridge = null;
                }
                final @Nullable HomekitServer homekitServer = this.homekitServer;
                if (homekitServer != null) {
                    homekitServer.stop();
                    this.homekitServer = null;
                }
```

lolodomo (changes requested):

bundles/org.openhab.io.homekit/src/main/java/org/openhab/io/homekit/internal/HomekitImpl.java
yfre: right, make sense. this.bridge could have become not null via other thread. but local bridge is final. 
the best - i copied the correct version from another place and "improved" it. 
thank you for pointing out. fixed.

lolodomo (approved):
LGTM

bundles/org.openhab.io.homekit/src/main/java/org/openhab/io/homekit/internal/HomekitImpl.java
lolodomo: I would have preferred i`f (bridge != null) {` rather than `if (this.bridge != null) {` but I hope it is ok like it is.

yfre: @lolodomo 
argh.. i checked 3 times and asked myself what else i could have missed :(  completely blind. 
3 error in 4 lines :(
need to learn to pay more attention to details. 
thank for your patience 
will take with the next feature-PR.

Sail-Away:
Here's my first feedback after serveral days of testing (with a slightly modified code, as I also wanted to catch the added interfaces in the log; code can be found [here](https://github.com/Sail-Away/openhab-addons/commit/4b8260286969b90dabe104de80df291a7530d0f5#diff-a0ee6722d485ec547013afd0fa1efeeb24b5bc7803781f7d212420ea94547e04)). Really seems to be an improvement enhancing stability, had no issues so far.

Stopping and starting of the bridge now is skipped when an (anyway unused IPv6) interface gets removed or added, which happens quite often. Number of actual restarts thus is considerably reduced.

Actual restarts in case of a network interruption seem to behave well now, though accessories still get removed from the accessory list. This is due to the fact that on removing the interface all connections to subscribing devices get terminated, and in this case the SubscriptionManager unsubscribes all characteristics. However, as the restart of the bridge only is triggered by a second event when the network interface is added again, and this typically happens in the order of several seconds later, nothing gets messed up. To illustrate this, I copied a few lines  from the log:
[homekit_excerpt.log](https://github.com/openhab/openhab-addons/files/7996756/homekit_excerpt.log)

Still I think that onChanged at least theoretically could be entered a second time before finishing the first time, as the network changes appear as separate events. So I wonder if it would make sense to lock onChanged (define it as synchronized).

One other small hint. I would change the log message 
`logger.trace("restarting HomeKit bridge on network interface changes.");`
into something like
`logger.trace("HomeKit bridge reacting on network interface changes.");`,
as a call of onChanged now not necessarily leads to a restart.

yfre:
@Sail-Away  thank you a lot for testing and detailed feedback. it good to hear PR improves behaviour. 

valid point with overlapping restarts and synchronized.  
i will open new pull request for this.

yfre:
@Sail-Away 
here we go
https://github.com/openhab/openhab-addons/pull/12205

Issue #12040, opened 2022-01-13
[knx] GroupRead / SendToKNX fails with NullPointerException


perrin-1:
I recently upgraded from OH 2.5. to OH 3.2.0
In the logs I see some exceptions that seem to be related to KNX:

`2022-01-13 00:00:03.747 [WARN ] [mmon.WrappedScheduledExecutorService] - Scheduled runnable ended with an exception:
java.lang.NullPointerException: null
        at tuwien.auto.calimero.device.ProcessCommunicationResponder.send(ProcessCommunicationResponder.java:256) ~[?:?]
        at tuwien.auto.calimero.device.ProcessCommunicationResponder.write(ProcessCommunicationResponder.java:249) ~[?:?]
        at tuwien.auto.calimero.device.ProcessCommunicationResponder.write(ProcessCommunicationResponder.java:226) ~[?:?]
        at org.openhab.binding.knx.internal.client.AbstractKNXClient.sendToKNX(AbstractKNXClient.java:460) ~[?:?]
        at org.openhab.binding.knx.internal.client.AbstractKNXClient.respondToKNX(AbstractKNXClient.java:439) ~[?:?]
        at org.openhab.binding.knx.internal.handler.DeviceThingHandler.lambda$8(DeviceThingHandler.java:297) ~[?:?]
        at org.openhab.binding.knx.internal.handler.DeviceThingHandler.withKNXType(DeviceThingHandler.java:148) ~[?:?]
        at org.openhab.binding.knx.internal.handler.DeviceThingHandler.sendGroupValueResponse(DeviceThingHandler.java:286) ~[?:?]
        at org.openhab.binding.knx.internal.handler.DeviceThingHandler.lambda$10(DeviceThingHandler.java:318) ~[?:?]
        at org.openhab.binding.knx.internal.handler.DeviceThingHandler.withKNXType(DeviceThingHandler.java:148) ~[?:?]
        at org.openhab.binding.knx.internal.handler.DeviceThingHandler.onGroupRead(DeviceThingHandler.java:312) ~[?:?]
        at org.openhab.binding.knx.internal.client.AbstractKNXClient$1.lambda$1(AbstractKNXClient.java:115) ~[?:?]
        at org.openhab.binding.knx.internal.client.AbstractKNXClient.lambda$8(AbstractKNXClient.java:258) ~[?:?]
        at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:515) ~[?:?]
[...]

PR #11759, opened 2021-12-11
[knx] Upgrade Calimero library to release 2.5.

holgerfriedrich:
- Upstream update of base library for KNX access from v2.4 to v2.5.
- Adapt AbstractKNXClient to new interface and replace calls of deprecated
  methods.
Fixes #6849.

Signed-off-by: Holger Friedrich <mail@holger-friedrich.de>

Issue #8872, opened 2020-10-26
[knx] support KNX data secure for OH3


holgerfriedrich:
Support for KNX data secure packets shall be added to OpenHAB.
Reading and writing authenticated KNX packets should be possible.

Calimero library seems to have all the necessary prerequisites at hand in latest Github version (not released).
The process includes exporting an encrypted keyring from ETS software, decrypt it using the Calimero library functions in the OH knx binding, and use the corresponding group address keys and sequence numbers from the keyring file.

## Prerequesites
* Calimero -> Java 11
* Calimero -> next Calimero version >2.4, released versions to not include security feature (also requested in #6849)

## Suggested Approach
* switch to calimero 2.5-SNAPSHOT (at least until 2.5 is released)
* add configuration items for keyring file(s) and password(s)
* add reading of keyring, and use decryption for packets with matching group addresses from the keyring
* implement encryption function
* handling of sequence numbers
* deal with "IP secure" topics later

holgerfriedrich:
I have already been playing around with this and willing to implement. If anyone is already working on this topic or wants to give some advice, it would be appreciated.

J-N-K:
If there is no immediate need (i.e. bugs or security issues), we usually don't use snapshots of dependencies. Do you happen to know when Calimero 2.5 will be released?

holgerfriedrich:
@J-N-K thanks for letting me know. My PR is still WIP and far from finished. Decrypting secure PDUs using the exported keyring from ETS works fine with the Calimero snapshot.
The more complicated thing is to setup a secure device needed for transmitting secure messages from openhab to other devices on the bus. I was not able to get this implemented properly yet.
Once it is running, I will contact the author.

J-N-K:
If you need help regarding the openHAB side, please ping me.

scholzi100:
Looks like Calimero got a milestone release ready. Following a release candidate in autumn. ([Ref](https://github.com/calimero-project/calimero-core/issues/97#issuecomment-765497671))

holgerfriedrich:
I have put together a version which actually supports *IP secure*, i.e. secure tunneling (tested) and secure routing (untested).
*Data secure* is working only passive (i.e. it can decrypt any received message, but not send successfully).
Documentation can be found here:
https://github.com/holgerfriedrich/openhab-addons/tree/pr-knx-data-secure/bundles/org.openhab.binding.knx#knx-secure
[...]

PR #12074, opened 2022-01-19
[knx] GroupRead / SendToKNX fails with NullPointerException #12040

FSeidinger:
# [knx] Fix for the NullPointerException as described in issue #12040 

# Description

This PR fixes the `NullPointerException` that arises from *-control items when a `GroupValueRead` request is issued by the KNX side and openHAB tries to answer with the state known to the bound item.

The cause is the introduction of the secure layer in the calimero library version 2.5. The constructor of the `ProcessCommunicationResponder` now takes two arguments. A `KNXNetworkLink` instance and a `SecureApplicationLayer` instance. The security application layer must not be null, although this precondition is not enforced by the constructor of the `ProcessCommunicationResponder`.

# Testing

This is the URL to the impacted build artefact:
https://openhab.jfrog.io/ui/native/libs-pullrequest-local/org/openhab/addons/bundles/org.openhab.binding.knx/3.3.0-SNAPSHOT/org.openhab.binding.knx-3.3.0-SNAPSHOT.jar


holgerfriedrich (approved):
LGTM

Sorry I missed this when upgrading Calimero #11759. Same change proposed for #8872.

Does it fully fix #12040?

kaikreuzer (approved):
Thanks for the analysis and the fix!

PR #12081, opened 2022-01-21
[hdpowerview] Update shade database

andrewfg:
Add Shade type 65 to database.


lolodomo (approved):
LGTM

jlaur:
@andrewfg - consider if PR title should be changed to "[hdpowerview] Add full support for Vignette Duolite" or something more specific for the release notes? As I understand it, adding correct capabilities to the database will ensure that position calculations are correct, so without it defaults are used, which can lead to some undesired behavior?

andrewfg:
@jlaur I chose the title 'Update Shade Database' because there will be several such generic updates, and this is just one of many. The PR description (will) say(s) what was actually updated in each PR.

Note: this new shade type has an existing 'capabilities' value, so its functionality has already been coded, and therefore no functional changes are required now..

lolodomo:
I agree that a more precise title would be better.

PR #12083, opened 2022-01-21
[homekit] use quantity type conversions for temperature characteristics

ccutrer:
if an Item associated with a temperature characteristic has a QuantityType
(of dimension Temperature) as its state, regardless of current unit,
use that to convert to celsius instead of any other configuration.

Note that this is only for supply values to HomeKit; commands coming from
HomeKit will still send a DecimalType with units according to the HomeKit-wide
useFahrenheit configuration.


andylintner (approved):
LGTM

yfre:
LGTM
thank you @ccutrer !

yfre (approved):

lolodomo (approved):
LGTM

PR #12093, opened 2022-01-22
[teleinfo] improve documentation for Linky Standard TIC mode

olivierkeke:
This PR improves documentation for Linky Standard TIC mode support (added in OH3.2).


lolodomo (approved):
I have not checked in details but it looks globally good.

PR #12095, opened 2022-01-23
[renault] Renault more channels and HVAC ON / toggle charge mode

dougculnane:
The initial Renault binding added in openhab 3.2 had a few basic channels.  This update adds more channels and the ability to interact with the car.

Turning the HVAC ON allows users to pre-heat their cars.

Switching charge mode allows users to control the charge limit of their cars without a smart charger.

There is much discussion, testing and feedback to these changes in the forum starting here: https://community.openhab.org/t/betatest-renault-ze-services-binding/72226/127

@lolodomo kindly reviewed inital contribution.  I have tried to continue the style of his feedback with the additional changes.  

The HVAC ON command to the car and wait for a response from the server is a problem i have tried to solve in a not too confusing way. I hope this is OK.  The status is only polled every 10 minutes so feedback that the HVAC is really ON needs a delayed response to give the car and MyRenault server time to react....


lolodomo:
I will review this PR.

dougculnane:
@lolodomo Thank you for reviewing this update to the binding.  

The HVAC ON command to the car and wait for a response from the server might not be the most openHAB way to do this. If you know of a better technique please let me know.

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/api/MyRenaultHttpSession.java
```diff
@@ -224,6 +229,67 @@ public void getLocation(Car car)
...
+                    throw new RenaultForbiddenException(
+                            "Kamereon Response Forbidden! Ensure the car is paired in your MyRenault App.");
+                } else if (HttpStatus.NOT_IMPLEMENTED_501 == response.getStatus()) {
+                    throw new RenaultNotImplementedException(
+                            "Kamereon Service Not Implemented: [" + response.getStatus() + "] " + response.getReason());
+                }
+            }
+        } catch (JsonParseException | InterruptedException | TimeoutException | ExecutionException e) {
+            logger.warn("Kamereon Request: {} threw exception: {} ", request.getURI().toString(), e.getMessage());
+        }
```
lolodomo, line 261: Please separate the case for InterruptedException in order to add `Thread.currentThread().interrupt();`
```
         } catch (InterruptedException e) {
            logger.warn(...
        }
```
Please also fix the existing method getKamereonResponse

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/api/MyRenaultHttpSession.java
```diff
@@ -224,6 +229,67 @@ public void getLocation(Car car)
...
+            if (HttpStatus.OK_200 != response.getStatus()) {
+                logger.warn("Kamereon Response: [{}] {} {}", response.getStatus(), response.getReason(),
+                        response.getContentAsString());
+                if (HttpStatus.FORBIDDEN_403 == response.getStatus()) {
+                    throw new RenaultForbiddenException(
+                            "Kamereon Response Forbidden! Ensure the car is paired in your MyRenault App.");
+                } else if (HttpStatus.NOT_IMPLEMENTED_501 == response.getStatus()) {
+                    throw new RenaultNotImplementedException(
+                            "Kamereon Service Not Implemented: [" + response.getStatus() + "] " + response.getReason());
+                }
```
lolodomo, line 257: Nothing if other status ?

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/api/MyRenaultHttpSession.java
```diff
@@ -224,6 +229,67 @@ public void getLocation(Car car)
...
```
lolodomo, line 286: Nothing if other status ?

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/api/MyRenaultHttpSession.java
```diff
@@ -224,6 +229,67 @@ public void getLocation(Car car)
...
+            }
+        } catch (JsonParseException | InterruptedException | TimeoutException | ExecutionException e) {
```
lolodomo, line 288: See my previous message.

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/api/MyRenaultHttpSession.java
```diff
@@ -224,6 +229,67 @@ public void getLocation(Car car)
...
+                        + "/kamereon/kca/car-adapter/v1/cars/" + config.vin + "/actions/charge-mode?country="
+                        + getCountry(config))
+                .method(HttpMethod.POST).header("Content-type", "application/vnd.api+json")
+                .header("apikey", this.constants.getKamereonApiKey())
+                .header("x-kamereon-authorization", "Bearer " + kamereonToken).header("x-gigya-id_token", jwt);
+        request.content(new StringContentProvider(
+                "{\"data\":{\"type\":\"ChargeMode\",\"attributes\":{\"action\":\"" + mode + "\"}}}", "utf-8"));
+        try {
+            ContentResponse response = request.send();
+            logger.debug("Kamereon Response HVAC ON: {}", response.getContentAsString());
```
lolodomo, line 276: Charging mode

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/handler/RenaultHandler.java
```diff
@@ -104,13 +105,70 @@ public void initialize() {
         }
         updateStatus(ThingStatus.UNKNOWN);
 
+        updateState(CHANNEL_HVAC_TARGET_TEMPERATURE,
+                new QuantityType<Temperature>(car.getHvacTargetTemperature().doubleValue(), SIUnits.CELSIUS));
```
lolodomo, line 109: The idea is to force a value 20°C at init ?

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/handler/RenaultHandler.java
```diff
@@ -104,13 +105,70 @@ public void initialize() {
...
 
+    @Override
+    public void handleCommand(ChannelUID channelUID, Command command) {
+        switch (channelUID.getId()) {
+            case RenaultBindingConstants.CHANNEL_HVAC_TARGET_TEMPERATURE:
+                if (command instanceof RefreshType && !car.isDisableHvac()) {
+                    updateState(CHANNEL_HVAC_TARGET_TEMPERATURE, new QuantityType<Temperature>(
+                            car.getHvacTargetTemperature().doubleValue(), SIUnits.CELSIUS));
+                } else if (command instanceof QuantityType) {
+                    car.setHvacTargetTemperature(((QuantityType<?>) command).doubleValue());
```
lolodomo, line 126: This is not good as you could receive a value in Fahrenheit. You have to convert your quantity in Celsius first.
You can find example in other bindings (managing a target temperature).

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/handler/RenaultHandler.java
```diff
@@ -104,13 +105,70 @@ public void initialize() {
...
+                    updateState(CHANNEL_HVAC_TARGET_TEMPERATURE, new QuantityType<Temperature>(
+                            car.getHvacTargetTemperature().doubleValue(), SIUnits.CELSIUS));
+                }
```
lolodomo, line 129: You should also accept DecimalType and in this case you consider that the provided value is in Celsius.

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/handler/RenaultHandler.java
```diff
@@ -104,13 +105,70 @@ public void initialize() {
...
+                    final MyRenaultHttpSession httpSession = new MyRenaultHttpSession(this.config, httpClient);
+                    try {
+                        updateState(CHANNEL_HVAC_STATUS, new StringType("PENDING"));
+                        car.resetHVACStatus();
+                        httpSession.initSesssion(car);
+                        httpSession.actionHvacOn(car.getHvacTargetTemperature());
+                        scheduler.schedule(() -> {
+                            updateHvacStatus(httpSession);
+                        }, config.updateDelay, TimeUnit.SECONDS);
+                    } catch (Exception e) {
```
lolodomo, line 144: Please avoid this generic `Exception`

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/handler/RenaultHandler.java
```diff
@@ -104,13 +105,70 @@ public void initialize() {
...
+                if (command instanceof StringType) {
+                    car.setChargeMode(command.toString());
+                    String mode = car.getChargingMode();
+                    if (mode != null) {
+                        MyRenaultHttpSession httpSession = new MyRenaultHttpSession(this.config, httpClient);
+                        try {
+                            httpSession.initSesssion(car);
+                            httpSession.actionChargeMode(command.toString());
+                            updateState(CHANNEL_CHARGING_MODE, new StringType(mode));
+                        } catch (Exception e) {
```
lolodomo, line 160: Please avoid this generic `Exception`

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/handler/RenaultHandler.java
```diff
@@ -104,13 +105,70 @@ public void initialize() {
...
+                        && !car.isDisableHvac()) {
+                    final MyRenaultHttpSession httpSession = new MyRenaultHttpSession(this.config, httpClient);
+                    try {
+                        updateState(CHANNEL_HVAC_STATUS, new StringType("PENDING"));
+                        car.resetHVACStatus();
+                        httpSession.initSesssion(car);
+                        httpSession.actionHvacOn(car.getHvacTargetTemperature());
+                        scheduler.schedule(() -> {
+                            updateHvacStatus(httpSession);
+                        }, config.updateDelay, TimeUnit.SECONDS);
```
lolodomo, line 143: This would have been better to save the job and try to cancel it when the thing handler is disposed.
Just to avoid avoid this job executed while your thing handler is already disposed.

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/handler/RenaultHandler.java
```diff
@@ -104,13 +105,70 @@ public void initialize() {
...
+                    } catch (Exception e) {
+                        logger.warn("Error My Renault Http Session.", e);
+                        updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.COMMUNICATION_ERROR, e.getMessage());
+                    }
+                }
+                break;
+            case RenaultBindingConstants.CHANNEL_CHARGING_MODE:
+                if (command instanceof StringType) {
+                    car.setChargeMode(command.toString());
+                    String mode = car.getChargingMode();
```
lolodomo, line 153: Maybe it would have been better to check the validity of the value here and log something if the value is not good, rather than in car.setChargeMode. Then car.setChargeMode can be a usual setter.

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/handler/RenaultHandler.java
```diff
@@ -104,13 +105,70 @@ public void initialize() {
...
+                    if (mode != null) {
+                        MyRenaultHttpSession httpSession = new MyRenaultHttpSession(this.config, httpClient);
+                        try {
+                            httpSession.initSesssion(car);
+                            httpSession.actionChargeMode(command.toString());
```
lolodomo, line 158: I think you want to use `mode` here.

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/handler/RenaultHandler.java
```diff
@@ -104,13 +105,70 @@ public void initialize() {
...
+                            updateState(CHANNEL_CHARGING_MODE, new StringType(mode));
+                        } catch (Exception e) {
+                            httpSession = null;
```
lolodomo, line 161: Note that it is not done in the other case (channel CHANNEL_HVAC_STATUS).
I let you think if this is normal or not.

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/handler/RenaultHandler.java
```diff
@@ -198,6 +269,29 @@ private void updateBattery(MyRenaultHttpSession httpSession) {
...
+                    updateState(CHANNEL_PLUG_STATUS, new StringType(plugStatus));
+                }
+                String chargingStatus = car.getChargingStatus();
+                if (chargingStatus != null) {
+                    updateState(CHANNEL_CHARGING_STATUS, new StringType(chargingStatus));
+                }
+                Double batteryAvailableEnergy = car.getBatteryAvailableEnergy();
+                if (batteryAvailableEnergy != null) {
+                    updateState(CHANNEL_BATTERY_AVAILABLE_ENERGY,
+                            new DecimalType(batteryAvailableEnergy.doubleValue()));
```
lolodomo, line 288: Please use `QuantityType<Energy>`

bundles/org.openhab.binding.renault/src/main/resources/OH-INF/thing/thing-types.xml
```diff
@@ -69,26 +78,88 @@
...
 	</thing-type>
 
-	<!-- Sample Channel Type -->
-	<channel-type id="hvacstatus">
-		<item-type>Switch</item-type>
-		<label>HVAC Status</label>
+	<channel-type id="batteryavailableenergy">
+		<item-type>Number:Energy</item-type>
+		<label>Battery Energy Available</label>
+		<state pattern="%.1f kWh" readOnly="true"></state>
```
lolodomo, line 92: Use `%unit%` in pattern rather than kWh

bundles/org.openhab.binding.renault/src/main/resources/OH-INF/thing/thing-types.xml
```diff
@@ -69,26 +78,88 @@
...
+	</channel-type>
+	<channel-type id="plugstatus">
+		<item-type>String</item-type>
+		<label>Plug Status</label>
+		<description>Status of charging plug.</description>
 		<state readOnly="true"></state>
```
lolodomo, line 98: Please add state options to map your possible values to a proper label for UI, a label that could even be translated later in different languages.

bundles/org.openhab.binding.renault/src/main/resources/OH-INF/thing/thing-types.xml
```diff
@@ -69,26 +78,88 @@
...
 	</channel-type>
-	<channel-type id="image">
+	<channel-type id="chargingstatus">
 		<item-type>String</item-type>
-		<label>Image URL</label>
-		<description>Image URL of MyRenault</description>
+		<label>Charging Status</label>
 		<state readOnly="true"></state>
```
lolodomo, line 103: Please add state options to map your possible values to a proper label for UI, a label that could even be translated later in different languages.

bundles/org.openhab.binding.renault/src/main/resources/OH-INF/thing/thing-types.xml
```diff
@@ -69,26 +78,88 @@
...
 	</channel-type>
+	<channel-type id="chargingmode">
+		<item-type>String</item-type>
+		<label>Charging Mode</label>
+	</channel-type>
```
lolodomo, line 108: Please add state options to map your possible values to a proper label for UI, a label that could even be translated later in different languages.

bundles/org.openhab.binding.renault/src/main/resources/OH-INF/thing/thing-types.xml
```diff
@@ -69,26 +78,88 @@
...
 		<state pattern="%.1f %unit%" readOnly="true"></state>
 	</channel-type>
+	<channel-type id="hvacstatus">
+		<item-type>String</item-type>
+		<label>HVAC Status (ON | OFF | PENDING)</label>
+		<state readOnly="false">
+			<options>
+				<option value="ON">ON</option>
+				<option value="PENDING">PENDING</option>
+				<option value="OFF">OFF</option>
```
lolodomo, line 134: Do you really want labels fully in uppercase ?

bundles/org.openhab.binding.renault/src/main/resources/OH-INF/thing/thing-types.xml
```diff
@@ -11,10 +11,19 @@
...
+			<channel id="batteryavailableenergy" typeId="batteryavailableenergy"/>
+			<channel id="plugstatus" typeId="plugstatus"/>
+			<channel id="chargingstatus" typeId="chargingstatus"/>
+			<channel id="chargingmode" typeId="chargingmode"/>
+			<channel id="chargingremainingtime" typeId="chargingremainingtime"/>
+			<channel id="estimatedrange" typeId="estimatedrange"/>
+			<channel id="odometer" typeId="odometer"/>
 			<channel id="hvacstatus" typeId="hvacstatus"/>
+			<channel id="hvactargettemperature" typeId="hvactargettemperature"/>
+			<channel id="externaltemperature" typeId="externaltemperature"/>
```
lolodomo, line 23: I think the existing system channel type is appropriate here : system.outdoor-temperature

bundles/org.openhab.binding.renault/src/main/resources/OH-INF/thing/thing-types.xml
```diff
@@ -69,26 +78,88 @@
...
+		<category>Temperature</category>
+		<state pattern="%.1f %unit%"></state>
+	</channel-type>
+	<channel-type id="externaltemperature" advanced="true">
+		<item-type>Number:Temperature</item-type>
+		<label>External Temperature</label>
+		<description>Temperature outside of the car</description>
+		<category>Temperature</category>
+		<state pattern="%.1f %unit%" readOnly="true"></state>
+	</channel-type>
```
lolodomo, line 151: To be suppressed if the system channel type is used.

lolodomo (changes requested):

lolodomo:
Additional comment: you will have to update your file `renault.properties`. This can be done automatically for you by running the command: `mvn i18n:generate-default-translations`

bundles/org.openhab.binding.renault/README.md
```diff
@@ -1,43 +1,126 @@
...
+| hvactargettemperature  | Number:Temperature | HVAC target temperature (19 to 21)              | No        |
+| externaltemperature    | Number:Temperature | Temperature outside of the car                  | Yes       |
+| image                  | String             | Image URL of MyRenault                          | Yes       |
+| location               | Location           | The GPS position of the vehicle                 | Yes       |
+| locationupdated        | DateTime           | Timestamp of the last location update           | Yes       |
+
+
+## Example
+
+renaultcar.sitemap:
```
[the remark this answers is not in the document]
dougculnane: The example is based on the default add thing -> configure car -> add all channels creating items.  So I do not have a .thing and .item file.  I am not sure how to export this.  I can reverse engineer it but not without a lot of work and testing?

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/api/Car.java
```diff
@@ -38,28 +44,63 @@
     private boolean disableHvac = false;
 
     private @Nullable Double batteryLevel;
+    private @Nullable Double batteryAvailableEnergy;
+    private @Nullable Integer chargingRemainingTime;
+    private @Nullable String chargingStatus;
+    private @Nullable String chargingMode;
     private @Nullable Boolean hvacstatus;
+    private Double hvacTargetTemperature = 20.0;
```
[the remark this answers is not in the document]
dougculnane: As far as I understand this value can not be got from the car it is just a pre-heat target temperature that can have the value 19, 20 or 21 on my car.  This seams to also be the feedback from users and their research in the python library forums.

I did discuss scrapping it and hard coding 20.  But it might work on some cars and in the future.  So we decided to include it and set it to be "Advanced".  The sitemap restricts it to 19, 20, 21 values.  So it is there but should not be messed with and set to 20.  This hopefully explains it in the code.

double !Double is a good idea.

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/api/Car.java
```diff
@@ -38,28 +44,63 @@
...
     private @Nullable Double odometer;
+    private @Nullable Double estimatedRange;
     private @Nullable String imageURL;
+    private @Nullable String locationUpdated;
```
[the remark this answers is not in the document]
dougculnane: I can do but the code parsing the string in the DateTimeType will be better than my code.
```
 updateState(CHANNEL_LOCATION_UPDATED, new DateTimeType(locationUpdated));
```

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/api/Car.java
```diff
@@ -128,80 +172,109 @@ public boolean isDisableLocation() {
...
+        switch (mode) {
+            case CHARGING_MODE_SCHEDULE:
+                chargingMode = CHARGING_MODE_SCHEDULE;
+                break;
+            case CHARGING_MODE_ALWAYS:
+                chargingMode = CHARGING_MODE_ALWAYS;
+                break;
+            default:
+                break;
+        }
```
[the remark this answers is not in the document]
dougculnane: I guess I should change the mode to an ChargingMode enum.  I was not sure what modes exist, but using an enum would be more robust.

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/api/MyRenaultHttpSession.java
dougculnane: I always log.warn if not HTTP 200.  I only throw errors for known situations like not implemented or forbidden.

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/handler/RenaultHandler.java
dougculnane: Yes. 20 is one of a few values that does not cause problems... I think it only affects pre-heat.  (This is at least true from my testing)

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/handler/RenaultHandler.java
dougculnane: I return update channel in Celsius or the supplied unit?

bundles/org.openhab.binding.renault/src/main/resources/OH-INF/thing/thing-types.xml
dougculnane: Yes because then the switch (ON OFF) habpannel widgets work...

bundles/org.openhab.binding.renault/src/main/resources/OH-INF/thing/thing-types.xml
dougculnane: Great idea but...  I want to hide this because it does not work on most cars so I want it to be "Advanced"  (Advanced means it probably will not work great....)

bundles/org.openhab.binding.renault/src/main/resources/OH-INF/thing/thing-types.xml
dougculnane: Done and set the units in the channel.
 ```
         updateState(CHANNEL_BATTERY_AVAILABLE_ENERGY,
                            new QuantityType<>(batteryAvailableEnergy.doubleValue(), org.openhab.core.library.unit.Units.KILOWATT_HOUR));
 ```

dougculnane:
Thanks @lolodomo 
I pushed some not tested changes I will do some more when I have more time and then I need to do some testing...

bundles/org.openhab.binding.renault/README.md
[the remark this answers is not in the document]
lolodomo: This is easy but ok forget, I should have requested during the initial binding review.

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/api/Car.java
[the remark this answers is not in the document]
lolodomo: But it will work only if the string has exactly the expected format. It probably has if you got a good date & time.
So keep it like that.

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/api/MyRenaultHttpSession.java
lolodomo: OK

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/handler/RenaultHandler.java
lolodomo: Why setting a value at init if you do not what is the real value. This can show a wrong information to users.
This channel is more a "command" channel to set a target temperature. It can be NULL at init and the user will set it to a value when he wants to set a target temperature.

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/handler/RenaultHandler.java
lolodomo: You are storing a decimal value and this value is considered in Celsius unit. So you have to convert the received QuantityType in unit Celsis.
Something like that:
```
                        QuantityType<Temperature> tempQtty = ((QuantityType<Temperature>) command)
                                .toUnit(SIUnits.CELSIUS);
                        if (tempQtty != null) {
                              car.setHvacTargetTemperature(tempQtty .doubleValue());
                       }
```

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/handler/RenaultHandler.java
lolodomo: Maybe you could just restart the existing refresh job considering an intial delay.
That way, no need to have a new job.

bundles/org.openhab.binding.renault/src/main/resources/OH-INF/thing/thing-types.xml
lolodomo: You tell me that you are labelling your options as a workaround for habpannel ?! I hope this is not the case. Note that these labels will be translated with different labels in other languages.

bundles/org.openhab.binding.renault/src/main/resources/OH-INF/thing/thing-types.xml
lolodomo: Advanced is not for channels that do not work !
Advanced is for channels that are working but only a part of users will need them.
But I agree that the system channel type cannot be advanced, so keep your specific channel type and mention in the channel description that this channel is working only with few car models. To be mentioned in documentation too.

bundles/org.openhab.binding.renault/src/main/resources/OH-INF/thing/thing-types.xml
dougculnane: Ok I think I have lowercase labels and uppercase values and enums for the options.  I think this is correct now but will need to test it.

bundles/org.openhab.binding.renault/src/main/resources/OH-INF/thing/thing-types.xml
dougculnane: The channel works but the api only delivered a value on 1 of the 3 test cars.  I have this sentence in the README.md

"The "Advanced" channels are especially likely to cause confusion and problems."

I want to avoid describing specific problems because I do not know what works where and can not support individual car models...  So I am purposefully vague.  

This is not great but the alternative is to remove these channels from the binding.

What would you advice about the best approach to theses API quirks...

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/handler/RenaultHandler.java
dougculnane: I want them to keep the default so I try to force it.

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/handler/RenaultHandler.java
dougculnane: Love to avoid it but I have no idea what could go wrong.

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/handler/RenaultHandler.java
dougculnane: We chatted about this last time.  This runs inside the pooling jobs and anything could happens so I catch all.  I understand it is a bit basic but not sure of all the possible runtime errors.

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/handler/RenaultHandler.java
dougculnane: That is an idea.  Not very efficient because i refreshes all and that means multiple API calls but it is nice and robust.  I will do it.

bundles/org.openhab.binding.renault/README.md
```diff
@@ -1,43 +1,126 @@
...
+}
+```
+
+![Sitemap](doc/sitemap.png)
+
+If you do not have a smart charger and want to limit the charge of the battery you can 
+set up an active 15 minute charge schedule in the MyRenault App. Then create 
+a Dimmer item "RenaultCar_ChargeLimit" and set it to 80% for example. This rule 
+will change the RenaultCar_ChargingMode to schedule_mode when the limit is reached.
+This stops charging after the battery level reaches the charge limit.
```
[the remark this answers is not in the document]
dougculnane: DONE.

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/api/Car.java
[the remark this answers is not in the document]
dougculnane: I have created enums for these state options.

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/api/Car.java
```diff
@@ -211,4 +284,44 @@ public void setGpsLongitude(Double gpsLongitude) {
...
+            case "0":
+                return "UNPLUGGED";
+            case "1":
+                return "PLUGGED";
+            case "-1":
+                return "PLUG_ERROR";
+            case "-2147483648":
+                return "PLUG_UNKNOWN";
+            default:
+                return "UNKNOWEN";
```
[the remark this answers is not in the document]
dougculnane: DONE

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/api/Car.java
```diff
@@ -211,4 +284,44 @@ public void setGpsLongitude(Double gpsLongitude) {
...
+            case "0.4":
+                return "ENERGY_FLAP_OPENED";
+            case "1.0":
+                return "CHARGE_IN_PROGRESS";
+            case "-1.0":
+                return "CHARGE_ERROR";
+            case "-1.1":
+                return "UNAVAILABLE";
+            default:
+                return "UNKNOWEN";
```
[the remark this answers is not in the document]
dougculnane: DONE

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/api/MyRenaultHttpSession.java
dougculnane: Done

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/api/MyRenaultHttpSession.java
dougculnane: Done

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/api/MyRenaultHttpSession.java
dougculnane: Fixed.

bundles/org.openhab.binding.renault/src/main/resources/OH-INF/thing/thing-types.xml
dougculnane: Done now with enums will run translation later once code stable.

bundles/org.openhab.binding.renault/src/main/resources/OH-INF/thing/thing-types.xml
dougculnane: Done now with enums will run translation later once code stable.

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/api/MyRenaultHttpSession.java
dougculnane: fixed.

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/handler/RenaultHandler.java
dougculnane: That is nice I have used this idea.

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/handler/RenaultHandler.java
dougculnane: Refactored with enums.

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/handler/RenaultHandler.java
dougculnane: refactored and improved

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/handler/RenaultHandler.java
dougculnane: I need to think about if I can reuse this httpSession for the command update status update.  It is valid for a few minutes (15 from memory) so reusing it might save the re-authentication flow which is quite expensive.  I would need a timeout flag to check and reuse or destroy and recreate...

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/handler/RenaultHandler.java
dougculnane: Done

bundles/org.openhab.binding.renault/src/main/resources/OH-INF/thing/thing-types.xml
dougculnane: done with enums.

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/api/Car.java
```diff
@@ -30,36 +30,101 @@
...
     private @Nullable String imageURL;
+    private @Nullable String locationUpdated;
     private @Nullable Double gpsLatitude;
     private @Nullable Double gpsLongitude;
+    private @Nullable Double externalTemperature;
+
+    public enum ChargingMode {
+        UNKNOWN,
+        SCHEDULE_MODE,
+        ALWAYS_CHARGING()
```
lolodomo, line 63: What is this ending `()` for?

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/api/Car.java
```diff
@@ -128,80 +196,114 @@ public boolean isDisableLocation() {
...
+     * @param mode
+     */
+    public void setChargeMode(ChargingMode mode) {
+        switch (mode) {
+            case SCHEDULE_MODE:
+                chargingMode = mode;
+                break;
+            case ALWAYS_CHARGING:
+                chargingMode = mode;
+                break;
```
lolodomo, line 303: Can be simplified by:
```
           case SCHEDULE_MODE:
            case ALWAYS_CHARGING:
                chargingMode = mode;
                break;
```

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/api/Car.java
```diff
@@ -128,80 +196,114 @@ public boolean isDisableLocation() {
...
+
+    public @Nullable Integer getChargingRemainingTime() {
+        return chargingRemainingTime;
+    }
+
+    public @Nullable Double getBatteryAvailableEnergy() {
+        return batteryAvailableEnergy;
+    }
+
+    public Double getHvacTargetTemperature() {
```
lolodomo, line 267: You can return simply a `double`. Make sense as the class member is now a `double`.

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/api/Car.java
```diff
@@ -128,80 +196,114 @@ public boolean isDisableLocation() {
...
+        return hvacTargetTemperature;
+    }
+
+    public void setHvacTargetTemperature(Double hvacTargetTemperature) {
```
lolodomo, line 271: Parameter could be a simple `double`. Make sense as the class member is now a `double`.

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/api/MyRenaultHttpSession.java
```diff
@@ -224,6 +233,75 @@ public void getLocation(Car car)
         }
     }
 
+    public void actionHvacOn(Double hvacTargetTemperature)
```
lolodomo, line 236: Parameter should be a simple `double`.

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/handler/RenaultHandler.java
```diff
@@ -104,13 +108,92 @@ public void initialize() {
...
+                    if (command instanceof RefreshType && !car.isDisableHvac()) {
+                        updateState(CHANNEL_HVAC_TARGET_TEMPERATURE, new QuantityType<Temperature>(
+                                car.getHvacTargetTemperature().doubleValue(), SIUnits.CELSIUS));
+                    } else if (command instanceof DecimalType) {
+                        car.setHvacTargetTemperature(((DecimalType) command).doubleValue());
+                        updateState(CHANNEL_HVAC_TARGET_TEMPERATURE, new QuantityType<Temperature>(
+                                car.getHvacTargetTemperature().doubleValue(), SIUnits.CELSIUS));
+                    } else if (command instanceof QuantityType && !car.isDisableHvac()) {
+                        double celsius = ((QuantityType<Temperature>) command).toUnit(SIUnits.CELSIUS).doubleValue();
+                        car.setHvacTargetTemperature(celsius);
```
lolodomo, line 135: Talke care, `toUnit` could return null.

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/handler/RenaultHandler.java
```diff
@@ -104,13 +108,92 @@ public void initialize() {
...
             pollingJob = scheduler.scheduleWithFixedDelay(this::getStatus, 0, config.refreshInterval, TimeUnit.MINUTES);
         }
     }
 
+    @Override
+    public void handleCommand(ChannelUID channelUID, Command command) {
+        switch (channelUID.getId()) {
+            case RenaultBindingConstants.CHANNEL_HVAC_TARGET_TEMPERATURE:
+                if (!car.isDisableHvac()) {
+                    if (command instanceof RefreshType && !car.isDisableHvac()) {
```
lolodomo, line 126: Condition `!car.isDisableHvac()` no more needed here

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/handler/RenaultHandler.java
```diff
@@ -104,13 +108,92 @@ public void initialize() {
...
+                        updateState(CHANNEL_HVAC_TARGET_TEMPERATURE, new QuantityType<Temperature>(
+                                car.getHvacTargetTemperature().doubleValue(), SIUnits.CELSIUS));
+                    } else if (command instanceof DecimalType) {
+                        car.setHvacTargetTemperature(((DecimalType) command).doubleValue());
+                        updateState(CHANNEL_HVAC_TARGET_TEMPERATURE, new QuantityType<Temperature>(
+                                car.getHvacTargetTemperature().doubleValue(), SIUnits.CELSIUS));
+                    } else if (command instanceof QuantityType && !car.isDisableHvac()) {
```
lolodomo, line 133: Condition `!car.isDisableHvac()` no more needed here

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/api/MyRenaultHttpSession.java
```diff
@@ -224,6 +233,75 @@ public void getLocation(Car car)
...
+                    throw new RenaultForbiddenException(
+                            "Kamereon Response Forbidden! Ensure the car is paired in your MyRenault App.");
+                } else if (HttpStatus.NOT_IMPLEMENTED_501 == response.getStatus()) {
+                    throw new RenaultNotImplementedException(
+                            "Kamereon Service Not Implemented: [" + response.getStatus() + "] " + response.getReason());
+                }
+            }
+        } catch (InterruptedException e) {
+            Thread.currentThread().interrupt();
+            logger.warn("Kamereon Request: {} threw exception: {} ", request.getURI().toString(), e.getMessage());
```
lolodomo, line 265: You should probably first log and then call interrupt.
Please update everywhere.

lolodomo (changes requested):

bundles/org.openhab.binding.renault/src/main/resources/OH-INF/thing/thing-types.xml
lolodomo: A good practice is to add a chapter "Limitations" at the end of the README. You could then remove your sentence from the channels chapter and in this new chapter you explain clearly which channels are working only for some of the cars. I see only two advanced channels, the outside temperature and the target heat temperature.

lolodomo:
There is now no more too much effort to have something we can merge.

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/handler/RenaultHandler.java
dougculnane: I investigated the default session token and it is valid for only 5 mins.  So I will not try to reuse the session and discard it after each use.

dougculnane:
@lolodomo I have pushed some fixes.  I need to do some testing because the binding broke with the enum changes.  I think I have fixed it but it will do no harm to live with it and test it for a few days. 

I still have to define unknown Exceptions.... I do not know how.

I also have to do the translations.

bundles/org.openhab.binding.renault/src/main/resources/OH-INF/thing/thing-types.xml
dougculnane: Added text to README.

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/handler/RenaultHandler.java
dougculnane: Removed the httpSession = null

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/api/Car.java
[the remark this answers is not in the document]
dougculnane: Done.

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/api/Car.java
dougculnane: mistake has been removed.

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/api/Car.java
dougculnane: done

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/api/MyRenaultHttpSession.java
dougculnane: done

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/handler/RenaultHandler.java
dougculnane: added a null handeler

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/handler/RenaultHandler.java
dougculnane: removed.

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/handler/RenaultHandler.java
dougculnane: removed

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/api/MyRenaultHttpSession.java
dougculnane: done.

lolodomo:
@dougculnane ; are your tests finally OK? 
I will have a quick final look.

dougculnane:
@lolodomo The binding is working well (again).  However there are some strange Renault API behaviours that I do not fully understand.

I will do the mvn i18n:generate-default-translations command and get the German and English translations done soon.

I am still not sure how to avoid generic Exceptions without a massive catch list or a dead job task...

lolodomo:
> I will do the mvn i18n:generate-default-translations command and get the German and English translations done soon.

The German translation must not be part of this PR. It should be done in Crowdin.

> I am still not sure how to avoid generic Exceptions without a massive catch list or a dead job task...

Let me check, if you are in the case of scheduled task, this may be acceptable.

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/handler/RenaultHandler.java
lolodomo: This is not inside a regular scheduled job, this is just run in the `handleCommand`. I see no valid reason to not catch only the exceptions that could be thrown by your code in the try part.

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/handler/RenaultHandler.java
lolodomo: This is not inside a regular scheduled job, this is just run in the `handleCommand`. I see no valid reason to not catch only the exceptions that could be thrown by your code in the try part.

lolodomo:
> Let me check again, if you are in the case of scheduled task, this may be acceptable.

This is not code run in a regular scheduled task so I see no reason to not catch so largely.

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/handler/RenaultHandler.java
dougculnane: Ok if the handleCommand throws a runtime exception is this an issue?

dougculnane:
@lolodomo I have pushed a list of errors that could be thrown.  I am not sure if this is what you wanted me to do.  I am concerned that a runtime error (like "SSL cert no long valid", or "Unknown host exception".... ) will throw a nasty error up the openHAB stack to be handled elsewhere.  I am not sure I can for-see all the unforeseen situations.... if I try to list them.  Maybe I misunderstand this or maybe openHAB handles these situations better.  Please let me know what you are after if my push is not it.

dougculnane:
@lolodomo Thanks for your response. I can reproduce the bug I think and I think it is an "analysis screen" issue.

https://community.openhab.org/t/betatest-renault-ze-services-binding/72226/180?u=doug_culnane

Therefore I will not try to patch the binding at this stage.

dougculnane:
> What is important for you is the state value of the linked item.

This is good I have check the display value and it is correct and the mysql stored value which is also correct (in meters)

```
| 2022-02-10 13:17:45 | 7642230 |
| 2022-02-10 16:28:39 | 7642540 |
| 2022-02-10 16:58:49 | 7647060 |
| 2022-02-10 17:08:53 | 7650330 |
| 2022-02-10 17:18:57 | 7653790 |
+---------------------+---------+
258 rows in set (0.001 sec)

MariaDB [openhab]> select * from item0989;
```

lolodomo:
> This is good I have check the display value and it is correct and the mysql stored value which is also correct (in meters)

Why in meters ? Your code assumes that the API value is in kilometers if you provide unit KILO(METER) when building the QuantityType.
Of course if you provide a pattern requesting the display in meters, mainUI will show meters.
Regarding persistence, as UoM is not managed by persistence, my feeling is that it should store the value you provide in the updateState, so a value in kilimeters.

lolodomo:
The first step is to check the unit of the value returned by the API. Any other discussion is probably just loss of time ;)

lolodomo:
> The totalMileage is km. I use this value with the KM unit. But I guess... the value is converted to a standard meters value as a Number:Length type...???

There is no standard meters value. 
Maybe a conversion specific to the mySQL persistence service?

dougculnane:
@lolodomo OK there may be issues but they are outside the scope of this binding and pull request I guess.

Did you get a chance to look at the exception handling?  Do you want to chat about this?  We can do a quick meeting now if that would save time.  I am not sure how i can resolve this.

dougculnane:
Tesla uses MILEs...
```
 ODOMETER("odometer", "odometer", DecimalType.class, false) {
            @Override
            public State getState(String s, TeslaChannelSelectorProxy proxy, Map<String, String> properties) {
                State someState = super.getState(s);
                BigDecimal value = ((DecimalType) someState).toBigDecimal();
                return new QuantityType<>(value, ImperialUnits.MILE);
            }
        },
 ```
 
 Volvo use  KILO(METRE)

 ```
                 return status.odometer != UNDEFINED ? new QuantityType<>((double) status.odometer / 1000, KILO(METRE))
                        : UnDefType.UNDEF;
  
```
https://github.com/openhab/openhab-addons/blob/888cf00245f18e3b04535325327c9977a053086f/bundles/org.openhab.binding.volvooncall/src/main/java/org/openhab/binding/volvooncall/internal/handler/VehicleHandler.java#L503

lolodomo:
So your code is correct IMHO.

Maybe there is a bug in MainUI.

lolodomo (approved):
LGTM

lolodomo:
> I am concerned that a runtime error (like "SSL cert no long valid", or "Unknown host exception".... ) will throw a nasty error up the openHAB stack to be handled elsewhere.

You are using Jetty HTTP client in your code and you are handling the exceptions that can be thrown by Jetty calls. So I don't understand what exception you could loose.

PR #12113, opened 2022-01-24
[hdpowerview] Optimize initialization of shades

### LINKED ISSUE #12071: [hdpowerview] Timing dependency during initialization of shades
#11707 fixed an issue with shades going **ONLINE** initially, then into **OFFLINE** with communication error for a minute. This fix was correct, but not enough.

#12002 changed this behavior to initialize to **UNKNOWN** instead of **ONLINE** - c8647e81bd6b42defff3e09aaded97a054b9b872. This makes sense since we don't know if it's online until we have received data. It could even be unknown to the hub if configured wrongly.

However, this last change revealed a problem that was luring since the beginning. Some shades will go **ONLINE** before switching state to **UNKNOWN**, then only back to **ONLINE** after a minute.

After analyzing this again I found the real issue. Look at this example (id 36321 is blind11):

```
2022-01-18 16:59:45.298 [DEBUG] [ternal.handler.HDPowerViewHubHandler] - Initializing hub
2022-01-18 16:59:45.311 [DEBUG] [ternal.handler.HDPowerViewHubHandler] - Scheduling poll for 5000 ms out, then every 60000 ms
2022-01-18 16:59:50.313 [DEBUG] [ternal.handler.HDPowerViewHubHandler] - Polling for state
2022-01-18 16:59:50.445 [DEBUG] [ternal.handler.HDPowerViewHubHandler] - Received data for 11 shades
2022-01-18 16:59:50.448 [DEBUG] [ternal.handler.HDPowerViewHubHandler] - Shade '24927' handler not initialized
[...]
2022-01-18 16:59:50.465 [DEBUG] [ternal.handler.HDPowerViewHubHandler] - Updating shade '36321'
2022-01-18 16:59:50.555 [INFO ] [ab.event.ThingStatusInfoChangedEvent] - Thing 'hdpowerview:shade:hub:blind11' changed from UNINITIALIZED (BRIDGE_UNINITIALIZED) to ONLINE
2022-01-18 16:59:50.562 [INFO ] [ab.event.ThingStatusInfoChangedEvent] - Thing 'hdpowerview:shade:hub:blind11' changed from ONLINE to INITIALIZING
2022-01-18 16:59:50.565 [DEBUG] [rnal.handler.HDPowerViewShadeHandler] - Initializing shade handler
2022-01-18 16:59:54.368 [INFO ] [ab.event.ThingStatusInfoChangedEvent] - Thing 'hdpowerview:shade:hub:blind11' changed from INITIALIZING to UNKNOWN
2022-01-18 17:01:25.971 [INFO ] [ab.event.ThingStatusInfoChangedEvent] - Thing 'hdpowerview:shade:hub:blind11' changed from UNKNOWN to ONLINE
```

Timeline of events:
- Hub is initialized and poll job is scheduled after 5 seconds.
- Shade data is fetched by hub handler, but shade handlers are not yet initialized.
- Some shade handlers may exist even if they are not initialized.
- Those that exists will receive data and marked as **ONLINE**.
- During initialization they will then be reset to **UNKNOWN**.

It's possible that the 5 second delay in starting the job was some kind of work-around for this issue. However, with a slow system (Raspberry Pi 3) and many shades (I have 11) this assumption about when handlers would be initialized is false.

Problem is here, in **HDPowerViewHubHandler**:

```
    private void updateShadeThing(String shadeId, Thing thing, @Nullable ShadeData shadeData) {
        HDPowerViewShadeHandler thingHandler = ((HDPowerViewShadeHandler) thing.getHandler());
        if (thingHandler == null) {
            logger.debug("Shade '{}' handler not initialized", shadeId);
            return;
        }
        if (shadeData == null) {
            logger.debug("Shade '{}' has no data in hub", shadeId);
        } else {
            logger.debug("Updating shade '{}'", shadeId);
        }
        thingHandler.onReceiveUpdate(shadeData);
    }
```

This does not consider the status of the shade things, only if it's possible to get the handler.

Solution requirements:
- Do not introduce a delay in initializing shades, i.e. wait up to one minute for polling.
- Get rid of the 5 second initial delay.

Proposal:
- `HDPowerViewHubHandler` should fetch data immediately and update shade handlers which are fully initialized.
- For shades not yet ready, keep the data.
- Implement `childHandlerInitialized` to initialize shade handlers when they are ready. See https://www.openhab.org/docs/developer/bindings/#handler-initialization-notification.
- Get rid of data in hub handler as shades are updated.

jlaur:
Fixes #12071 - please see this issue for further details.

This pull request fixes a concurrency issue and optimizes the status transitions for shades to come online as fast as possible:
- A five second initial delay in requesting shade data has been removed. It probably existed because of next problem:
- Bridge could receive data before shade things were initialized. This would cause them to come **ONLINE** prematurely, thus go back to **UNKNOWN** when being initialized. After this there would be a one minute delay in coming **ONLINE** again after next data poll.

Example of transitions for a single shade after the changes:
```
2022-01-24 23:28:01.835 [INFO ] [ab.event.ThingStatusInfoChangedEvent] - Thing 'hdpowerview:shade:hub:blind11' changed from UNINITIALIZED (BRIDGE_UNINITIALIZED) to INITIALIZING
2022-01-24 23:28:01.847 [DEBUG] [rnal.handler.HDPowerViewShadeHandler] - Initializing shade handler for shade 36321
2022-01-24 23:28:01.873 [INFO ] [ab.event.ThingStatusInfoChangedEvent] - Thing 'hdpowerview:shade:hub:blind11' changed from INITIALIZING to UNKNOWN
2022-01-24 23:28:02.010 [DEBUG] [ternal.handler.HDPowerViewHubHandler] - Child handler initialized: hdpowerview:shade:hub:blind11
2022-01-24 23:28:02.026 [DEBUG] [ternal.handler.HDPowerViewHubHandler] - Updating shade '36321'
2022-01-24 23:28:02.029 [DEBUG] [rnal.handler.HDPowerViewShadeHandler] - Caching capabilities 7 for shade 36321
2022-01-24 23:28:02.206 [INFO ] [ab.event.ThingStatusInfoChangedEvent] - Thing 'hdpowerview:shade:hub:blind11' changed from UNKNOWN to ONLINE
```

Planned for next PR in this area:
- Implement handling of shade removal, i.e. deletion of a shade in the PowerView Hub.

### DIFF
--- a/bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/handler/HDPowerViewHubHandler.java
+++ b/bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/handler/HDPowerViewHubHandler.java
@@ -17,6 +17,7 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.CopyOnWriteArrayList;
 import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.TimeUnit;
@@ -57,6 +58,7 @@ import org.openhab.core.thing.ChannelUID;
 import org.openhab.core.thing.Thing;
 import org.openhab.core.thing.ThingStatus;
 import org.openhab.core.thing.ThingStatusDetail;
+import org.openhab.core.thing.ThingUID;
 import org.openhab.core.thing.binding.BaseBridgeHandler;
 import org.openhab.core.thing.binding.ThingHandler;
 import org.openhab.core.thing.binding.builder.ChannelBuilder;
@@ -77,8 +79,6 @@ import org.slf4j.LoggerFactory;
 @NonNullByDefault
 public class HDPowerViewHubHandler extends BaseBridgeHandler {
 
-    private static final long INITIAL_SOFT_POLL_DELAY_MS = 5_000;
-
     private final Logger logger = LoggerFactory.getLogger(HDPowerViewHubHandler.class);
     private final HttpClient httpClient;
     private final HDPowerViewTranslationProvider translationProvider;
@@ -92,6 +92,7 @@ public class HDPowerViewHubHandler extends BaseBridgeHandler {
     private @Nullable ScheduledFuture<?> hardRefreshPositionFuture;
     private @Nullable ScheduledFuture<?> hardRefreshBatteryLevelFuture;
 
+    private ConcurrentHashMap<ThingUID, ShadeData> pendingShadeInitializations = new ConcurrentHashMap<>();
     private List<Scene> sceneCache = new CopyOnWriteArrayList<>();
     private List<SceneCollection> sceneCollectionCache = new CopyOnWriteArrayList<>();
     private List<ScheduledEvent> scheduledEventCache = new CopyOnWriteArrayList<>();
@@ -135,11 +136,11 @@ public class HDPowerViewHubHandler extends BaseBridgeHandler {
             if (sceneChannelTypeUID.equals(channel.getChannelTypeUID()) && OnOffType.ON == command) {
                 webTargets.activateScene(id);
                 // Reschedule soft poll for immediate shade position update.
-                scheduleSoftPoll(0);
+                scheduleSoftPoll();
             } else if (sceneGroupChannelTypeUID.equals(channel.getChannelTypeUID()) && OnOffType.ON == command) {
                 webTargets.activateSceneCollection(id);
                 // Reschedule soft poll for immediate shade position update.
-                scheduleSoftPoll(0);
+                scheduleSoftPoll();
             } else if (automationChannelTypeUID.equals(channel.getChannelTypeUID())) {
                 webTargets.enableScheduledEvent(id, OnOffType.ON == command);
             }
@@ -162,6 +163,7 @@ public class HDPowerViewHubHandler extends BaseBridgeHandler {
             return;
         }
 
+        pendingShadeInitializations = new ConcurrentHashMap<>();
         webTargets = new HDPowerViewWebTargets(httpClient, host);
         refreshInterval = config.refresh;
         hardRefreshPositionInterval = config.hardRefresh;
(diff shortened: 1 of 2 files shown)


bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/handler/HDPowerViewHubHandler.java
```diff
@@ -92,6 +92,7 @@
     private @Nullable ScheduledFuture<?> hardRefreshPositionFuture;
     private @Nullable ScheduledFuture<?> hardRefreshBatteryLevelFuture;
 
+    private ConcurrentHashMap<ThingUID, ShadeData> pendingShadeInitializations = new ConcurrentHashMap<>();
```
andrewfg, line 95: You can declare and initialize this as `private final` ..

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/handler/HDPowerViewHubHandler.java
```diff
@@ -162,6 +163,7 @@ public void initialize() {
             return;
         }
 
+        pendingShadeInitializations = new ConcurrentHashMap<>();
```
andrewfg, line 166: If the HashMap is initialized as `final` then it should not be initialized again..

andrewfg (changes requested):

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/handler/HDPowerViewHubHandler.java
jlaur: See next comment.

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/handler/HDPowerViewHubHandler.java
jlaur: That would mean it would have to be `clear()`'ed here. I think it will be faster to create new object and let old one be garbage collected. Haven't tested though, but I don't see any clear advantage of this approach compared to the other?

jlaur: Hmm, reading it again, this is `initialize()`. So it would even have to clear the hash map even upon first initialization. So I really don't see any advantage in having it `final`. It's similar to other instance variables initialized from `initialize()`, like webTargets and futures.

jlaur: Okay, I concur. It's a pickle. Really don't want to make it `@Nullable` just for the sake of this. Clearing an empty hash map should be quite fast anyway. Fixed.

andrewfg (approved):

jlaur:
@fwolter - sorry to exploit you a little, but https://github.com/openhab/openhab-addons/pull/12061#discussion_r791120507 seems to suggest that these channel state changes is your ball game, so maybe you could also have a look at this one now that we're at it? :-)

fwolter (approved):

PR #12123, opened 2022-01-26
[jrubyscripting] Apply RUBYLIB configuration to $LOAD_PATH

jimtng:
The `rubylib` addon configuration is supposed to alter the $LOAD_PATH but currently it doesn't. As a result, personal libraries located in RUBYLIB path won't be found. Currently a workaround is applied in the helper library to parse RUBYLIB and add the paths into $LOAD_PATH. 

However, GUI users who opt not to use the helper library will not be able to load personal libraries based on the `rubylib` search path.

This PR remedies this situation by updating `$LOAD_PATH` within the addon.

### DIFF
--- a/bundles/org.openhab.automation.jrubyscripting/src/main/java/org/openhab/automation/jrubyscripting/internal/JRubyScriptEngineConfiguration.java
+++ b/bundles/org.openhab.automation.jrubyscripting/src/main/java/org/openhab/automation/jrubyscripting/internal/JRubyScriptEngineConfiguration.java
@@ -234,7 +234,7 @@ public class JRubyScriptEngineConfiguration {
             try {
                 engine.eval(code);
             } catch (ScriptException exception) {
-                logger.error("Error setting $LOAD_PATH from RUBYLIB='{}': {}", rubyLib.get(), exception.getMessage());
+                logger.warn("Error setting $LOAD_PATH from RUBYLIB='{}': {}", rubyLib.get(), exception.getMessage());
             }
         }
     }
--- a/bundles/org.openhab.automation.jrubyscripting/src/main/java/org/openhab/automation/jrubyscripting/internal/JRubyScriptEngineConfiguration.java
+++ b/bundles/org.openhab.automation.jrubyscripting/src/main/java/org/openhab/automation/jrubyscripting/internal/JRubyScriptEngineConfiguration.java
@@ -47,6 +47,7 @@ public class JRubyScriptEngineConfiguration {
     private static final Path DEFAULT_RUBYLIB = Paths.get(OpenHAB.getConfigFolder(), "automation", "lib", "ruby");
 
     private static final String GEM_HOME = "gem_home";
+    private static final String RUBYLIB = "rubylib";
 
     // Map of configuration parameters
     private static final Map<String, OptionalConfigurationElement> CONFIGURATION_PARAMETERS = Map.ofEntries(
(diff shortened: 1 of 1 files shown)


bundles/org.openhab.automation.jrubyscripting/src/main/java/org/openhab/automation/jrubyscripting/internal/JRubyScriptEngineConfiguration.java
```diff
@@ -209,6 +210,33 @@ private void configureRubyEnvironment(List<OptionalConfigurationElement> optiona
...
+
+        Optional<String> rubyLib = rubyLibConfigElement.getValue();
+        if (rubyLib.isPresent() && !rubyLib.get().trim().isEmpty()) {
+            final String code = "$LOAD_PATH.unshift *ENV['RUBYLIB']&.split(File::PATH_SEPARATOR)" + //
+                    "&.reject(&:empty?)" + //
+                    "&.reject { |path| $LOAD_PATH.include?(path) }"; //
+            try {
+                engine.eval(code);
+            } catch (ScriptException exception) {
+                logger.error("Error setting $LOAD_PATH from RUBYLIB='{}': {}", rubyLib.get(), exception.getMessage());
```
lolodomo, line 237: warn looks more compliant with guidelines

lolodomo (changes requested):

jimtng:
@lolodomo I've changed it to `warn`. There are two other existing logger calls along the same line that use error instead of warn. Should I change them too?

lolodomo:
Code guideline: https://www.openhab.org/docs/developer/guidelines.html#f-logging
Extract from doc;

> error logging should only be used
> 
> - to inform the user that something is tremendously wrong in the setup, the system cannot function normally anymore, and there is a need for immediate action.
> - in case some code fails irrecoverably and the user should report it as a severe bug.

So error should rarely be used. 

If you see other inappropriate usages of error, yes please fix them too.

jimtng:
@lolodomo I've given this further thought, but I still can't determine with 100% certainty whether `error` or `warn` should be used here. I am leaning towards using `error` as in the first commit.

Essentially the code tries to add to the $LOAD_PATH global variable, and catches the exception in case of error. In such situation, the intended effect would not have been achieved, so user scripts that want to load additional libraries from the LOAD_PATH would fail.

Wouldn't this be considered that something is "tremendously wrong and cannot function normally"? Furthermore, actions are required from the user to remedy the situation (by fixing the config that caused the error).

lolodomo:
As I understand, you are adding a new feature, not a mandatory feature, the system can work well without it. Not a critical error.

> warn logging should only be used
> 
>     to inform the user that something seems to be wrong in the overall setup, but the system can nonetheless function as normal,
>     in recoverable situations when a section of code that is not accessed under normal operating conditions is reached.

jimtng:
This isn't actually a new feature, more a bug fix. Originally the addon allows to set `rubylib` in the config. The addon then sets the corresponding RUBYLIB environment variable, which is supposed to search for user's scripts in the given paths.

```
require 'mylibrary`
```

However this isn't currently working.

So this PR adds RUBYLIB into LOAD_PATH post scriptEngine instantiation to make it work. The same solution is currently implemented within the jruby helper library, but since this helper library isn't a part of the addon, the best solution is to incorporate this patch into the addon.

jimtng:
The RUBYLIB setting will not work for those who aren't using the helper library (eg in a GUI rule). 

A decision was made to implement the fix in the helper library first because it was the easiest to do at the time. This PR now includes the fix in the addon, so it would work with or without the helper library, e.g. in a GUI rule. See the discussion here https://github.com/boc-tothefuture/openhab-jruby/pull/267#issuecomment-920429079

jimtng:
@lolodomo if you believe the proper level is indeed warn, then this PR can be merged now.

lolodomo (approved):
LGTM

PR #12138, opened 2022-01-27
[hdpowerview] Introduce command channel for shades

jlaur:

Refactor #12002 by changing the calibrate channel of type **Switch** to a command channel with option **CALIBRATE**.

I learned from https://github.com/openhab/openhab-addons/pull/12061#discussion_r790944826 about this possibility and it seems more appropriate here.

One more command is already in the pipeline: **IDENTIFY**.


jlaur:
The previous pull request introducing the calibrate channel was merged into 3.3, so this is not a breaking change. It even might have a chance to get into same milestone release? :-)

fwolter (approved):

PR #12143, opened 2022-01-28
[jrubyscripting] Upgrade jruby version to 9.3.3.0

jimtng:
Upgrades jruby version to 9.3.3.0, which was released on the 19th of January 2022

https://www.jruby.org/2022/01/19/jruby-9-3-3-0.html


lolodomo (approved):
LGTM

PR #12145, opened 2022-01-28
[radiothermostat] Workaround for incorrectly reported fan state

### LINKED ISSUE #12144: [radiothermostat] Some thermostats do not correctly report fan state
Some models and/or firmware revisions of the thermostats supported by this binding do not correctly report the fan status as on when the thermostat is calling for heat or cool. On the thermostats that behave incorrectly, the 'fstate' json variable is reported as 0 when it should be 1.

Discussed in this community thread: https://community.openhab.org/t/hvac-fan-runtime-based-on-min-percentage-rule/132302/9

mlobstein:
Apply logic to the fan_status channel for some thermostats that don't properly report that the fan is on during heating or cooling. Any time the thermostat status is heating (1) or cooling (2), the fan status will be reported as 1. When the thermostat status is off (0), the fan_status will be reported from the json data.

fixes #12144


lolodomo (approved):
LGTM

PR #12101, opened 2022-01-23
[wemo] Refactor duplicated code

jlaur:
Signed-off-by: Jacob Laursen <jacob-github@vindvejr.dk>

Fixes #12100 

Changes:
- `getHost()`, `getUDN()` and `isUpnpDeviceRegistered()` have been extracted to a common base class for all handlers.
- `AbstractWemoHandler` has been deleted as it served no purpose other than containing a `WemoHttpCall` instance, but actually each handler held their own instances. This instance has been moved to `WemoBaseThingHandler` and handlers have been refactored to use it.

PR #12120, opened 2022-01-25
[wemo] Separate handlers for Insight, Motion and Socket/Light Switch

jlaur:
Fixes #12105

Signed-off-by: Jacob Laursen <jacob-github@vindvejr.dk>

Review notes:
- First commit contains **no logic/code changes**, just extraction/separation. Method `onValueReceived()` was copied into each handler and the parts not concerning that handler were removed.
- This is a work-in-progress, but code is ready to be merged at any time to reduce potential conflicts with other ongoing work and to keep review process manageable.
- Next steps will be some actual refactoring of `onValueReceived()`, see linked issue for details.

PR #12148, opened 2022-01-28
[wemo] Improve GENA subscription reliability and error handling

jlaur:

Continuation of #12101:
- Service subscription consolidated in base class (removing a lot of duplicate code).
- Fix status wrongly set to **ONLINE** when exception is thrown.
- Fix exception thrown from `handleCommand()` - set status **OFFLINE** instead.
- Rework error handling for `WemoHttpCall`: Duplicated logging moved inside and throws exception rather than returning null.
- Decrease log level from **error** to **warning** for communication errors and avoid warning every minute while a device is offline.
- Improve reliability of GENA events by renewing subscriptions automatically.

Review notes:
- Reading commit by commit might be helpful to understand the process/distinct steps.

### DIFF
--- a/bundles/org.openhab.binding.wemo/src/main/java/org/openhab/binding/wemo/internal/discovery/WemoLinkDiscoveryService.java
+++ b/bundles/org.openhab.binding.wemo/src/main/java/org/openhab/binding/wemo/internal/discovery/WemoLinkDiscoveryService.java
@@ -133,100 +133,97 @@ public class WemoLinkDiscoveryService extends AbstractDiscoveryService implement
 
                 String endDeviceRequest = wemoHttpCaller.executeCall(wemoURL, soapHeader, content);
 
-                if (endDeviceRequest != null) {
-                    logger.trace("endDeviceRequest answered '{}'", endDeviceRequest);
+                logger.trace("endDeviceRequest answered '{}'", endDeviceRequest);
 
-                    try {
-                        String stringParser = substringBetween(endDeviceRequest, "<DeviceLists>", "</DeviceLists>");
+                try {
+                    String stringParser = substringBetween(endDeviceRequest, "<DeviceLists>", "</DeviceLists>");
 
-                        stringParser = unescapeXml(stringParser);
+                    stringParser = unescapeXml(stringParser);
 
-                        // check if there are already paired devices with WeMo Link
-                        if ("0".equals(stringParser)) {
-                            logger.debug("There are no devices connected with WeMo Link. Exit discovery");
-                            return;
-                        }
-
-                        // Build parser for received <DeviceList>
-                        DocumentBuilderFactory dbf = DocumentBuilderFactory.newInstance();
-                        // see
-                        // https://cheatsheetseries.owasp.org/cheatsheets/XML_External_Entity_Prevention_Cheat_Sheet.html
-                        dbf.setFeature("http://xml.org/sax/features/external-general-entities", false);
-                        dbf.setFeature("http://xml.org/sax/features/external-parameter-entities", false);
-                        dbf.setFeature("http://apache.org/xml/features/nonvalidating/load-external-dtd", false);
-                        dbf.setXIncludeAware(false);
-                        dbf.setExpandEntityReferences(false);
-                        DocumentBuilder db = dbf.newDocumentBuilder();
-                        InputSource is = new InputSource();
-                        is.setCharacterStream(new StringReader(stringParser));
-
-                        Document doc = db.parse(is);
-                        NodeList nodes = doc.getElementsByTagName("DeviceInfo");
-
-                        // iterate the devices
-                        for (int i = 0; i < nodes.getLength(); i++) {
-                            Element element = (Element) nodes.item(i);
-
-                            NodeList deviceIndex = element.getElementsByTagName("DeviceIndex");
-                            Element line = (Element) deviceIndex.item(0);
-                            logger.trace("DeviceIndex: {}", getCharacterDataFromElement(line));
-
-                            NodeList deviceID = element.getElementsByTagName("DeviceID");
-                            line = (Element) deviceID.item(0);
-                            String endDeviceID = getCharacterDataFromElement(line);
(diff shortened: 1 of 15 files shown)


hmerk:
Thanks @jlaur, this looks really good at first quick sight.
Only one thing I don‘t agree, commit 7d073f6 - removing subscription cache.
Subscroption duration is 600s per default, refresh cycle is 60s. With your proposed change, we try to subscribe to events ever 60s, even there are registered subscriptions.
I don‘t know how the Wemo devices will handle this on the long run, but I would not want to try…

jlaur:
> Subscroption duration is 600s per default, refresh cycle is 60s. With your proposed change, we try to subscribe to events ever 60s, even there are registered subscriptions.

Thanks for this comment, I missed this logic. I'll rework it! I don't quite follow the original implementation, though. It seems to me that there is no resubscription performed? `addSubscription()` will be called every minute, but if subscription is already cached, it will not subscribe again.

hmerk:
> It seems to me that there is no resubscription performed?

Indeed, unfortunately, resubsciption is not implemented in org.openhab.core.io.transport.upnp.

jlaur:
> Indeed, unfortunately, resubsciption is not implemented in org.openhab.core.io.transport.upnp.

So summing up:
- Original implementation: Cache prevents any kind of resubscription and caches before getting confirmation about subscription. Do you know if this in practice means that we stop receiving events 10 minutes after binding starts? I'll test this to verify.
- Current PR implementation: Resubscription is too frequent (every minute).

After some tests I'll implement a proper subscription renewal mechanism.

hmerk:
No, cache prevents adding a new subscription as long as there is a valid subscription. So we would loose events just for about a minute. It is the same implementation as in Sonos binding.

jlaur:
> No, cache prevents adding a new subscription as long as there is a valid subscription. So we would loose events just for about a minute. It is the same implementation as in Sonos binding.

But cache doesn't know if subscription is still valid, it just caches the service name once and keeps it? At least if participant is still registered.

hmerk:
You are right, so there must be some logic within JUPnP to resubscribe, as I did not see loosing subscriptions after 10 minutes.
What about adding a cache which gets invalid after DEFAULT-SUBSCRIPTION-DURATION and then add a new subscription ?

jlaur:
> What about adding a cache which gets invalid after DEFAULT-SUBSCRIPTION-DURATION and then add a new subscription ?

That's similar to what I had in mind with:

> After some tests I'll implement a proper subscription renewal mechanism.

I have pushed a draft implementation of this and is currently testing it. I was not able to reproduce any loss of events (with original code) after subscription supposedly expires:

```
    /**
     * Subscribe to a GENA subscription
     * 
     * @param participant the participant to the subscription is for
     * @param serviceID the UPNP service we want to subscribe to
     * @param duration the duration of the subscription
     */
    public void addSubscription(UpnpIOParticipant participant, String serviceID, int duration);
```

I also double-checked that duration is in seconds since not documented here. So my implementation might be a theoretic one as duration might have some other meaning, or if there is renewal implemented elsewhere. I never missed events from my WeMo Insight, but it might have gone unnoticed since there is also polling of values each minute.

hmerk (approved):
LGTM

jlaur:
JAR here: https://drive.google.com/file/d/12hX52zTN53LEUra6rCiwwI-NNy4Gw6ki/view?usp=sharing

@jaywiseman1971 - if you would still have some nerve for this, this one would be nice to have tested with a broader equipment selection.

jlaur:
@hmerk - I pushed two new commits after testing handling of Insight device being turned off and on again as well as being turned off initially when binding is started, then turned on. I found some issues which are now also addressed.

jaywiseman1971:
> JAR here: https://drive.google.com/file/d/12hX52zTN53LEUra6rCiwwI-NNy4Gw6ki/view?usp=sharing
> 
> @jaywiseman1971 - if you would still have some nerve for this, this one would be nice to have tested with a broader equipment selection.

Once it becomes more unchanged for a while, I will test and provide feedback.  I get all the emails on this thread daily so I'm aware of all the massive changes.  I just can't be the first one to test all these changes in my PRO system.

jlaur:
> Once it becomes more unchanged for a while, I will test and provide feedback. I get all the emails on this thread daily so I'm aware of all the massive changes. I just can't be the first one to test all these changes in my PRO system.

That's fine, I'll not bug you anymore by tagging you, I think I understand your standpoint by now. :-) Everything is already tested with Insight, for other devices I hope that no regressions are introduced by mistake, since I can't test.

hmerk (approved):

hmerk:
@jlaur Could you please change Line 495 in WemoDimmerHandler from
logger.trace("New attribute brewed '{}' received", dateTimeState);
to
logger.trace("New dateTime attribute '{}' received", dateTimeState);

hmerk:
@jlaur There seems to be an issue with the new WemoInsightHandler, as my logs are being flooded with warnings „Handler already disposed“ while trying to do state updates….

jlaur:
> There seems to be an issue with the new WemoInsightHandler, as my logs are being flooded with warnings „Handler already disposed“ while trying to do state updates….

Are you running a build from this PR, or #12120 (now main)? Can you reproduce? Can you post logs from when it started, i.e. what happened just before? I have it running in production now with two WeMo Insights (one online 24/7, one being turned on/off every now and then) for a week without any issues.

hmerk:
@jlaur running from this PR. I will check the logs tomorrow to see when it started and what happened before.
Will try to reproduce with stopping and restarting the binding.
For completeness, i am on 3.3 M1..

hmerk:
@jlaur Seems to have been a coincedence with reboot after upgrade, as I have seen the same error on another binding, just not as frequently.
I will do a clean restart over the weekend and keep an eye on this.

jlaur:
@openhab/add-ons-maintainers, anyone willing to have a look at this one before next round? :-)

kaikreuzer (approved):
Thanks @jlaur, your code lgtm and as @hmerk has tested it, it should all be good to merge!

PR #12151, opened 2022-01-29
[ipobserver] Add support for WiFi version and push based method.

Skinah:
A newer WiFi based ipObserver has been released that physically looks the same, but does not have the RJ45 connector and has the livedata page removed from its firmware that the binding uses to scrape the data from. This PR adds support for the WiFi unit by capturing the WeatherUnderground packets and will parse them into the same channels. The updated method works for both RJ45 and WiFi models and there are a number of users already using this for many months and reporting it works well for them.


MikeTheTux:
> | rainToday | Number:Length | Amount of rain since 12:00am. |
Shouldn't this be 00:00am? Midnight an not noon?

MikeTheTux:
> | responseTime | Number:Time | How long it took the weather station to reply to a request for the live data. |
What is this channel reporting in the push mode?
It could report the update period between two push requests. In this case channel name and description should differ between the two modes

Skinah:
00:00 is for 24 hour clocks, so if it is mentioning AM it is not 24 hour format.
12:00am is correct as it is 11:59PM and then it changes at midnight to 12:00am
00:00 with the `am` missing is correct

It is confusing and this is why army/navy/airforce use the 24 hour to save the confusion. Happy to change it to 'midnight' if that is a clearer way to write the text.

Skinah:
Thank you for the suggestions and taking a look.... Regarding the responseTime channel currently the channel just stays as NULL and it is marked as an Advanced channel and not shown by default. I'm not sure how useful it would be unless a thread was started that checks every minute and updates the channel, otherwise it would not get updated if the push events stop unless it was done in a new thread with 1 minute delay between runs. It would be useful to know if the time between pushes was above X time so a SMS/email/telegram/push alert could be sent.

MikeTheTux:
> 00:00 is for 24 hour clocks, so if it is mentioning AM it is not 24 hour format.
> 12:00am is correct as it is 11:59PM and then it changes at midnight to 12:00am
You are right. My fault.

> Happy to change it to 'midnight' if that is a clearer way to write the text.
Your decission. I guess it would be easier to read for all non-native speakers.

MikeTheTux:
> The other binding worth mentioning is the weather underground binding that allows the data to be intercepted on its way to WU

I guess you mean the not yet merged "Wunderground Update Receiver Binding"? In this case the name should be adapted and maybe even a hint about the open PR given.
Is the reference anyhow still required, as this binding has now a similar functionality?

MikeTheTux:
> I'm not sure how useful it would be unless a thread was started that checks every minute and updates the channel, otherwise it would not get updated if the push events stop unless it was done in a new thread with 1 minute delay between runs. It would be useful to know if the time between pushes was above X time so a SMS/email/telegram/push alert could be sent.

In my current weatherstation integration, I'm calculating the period of the last update and it is typically 16s. Somewhen the time starts to increase, which is caused by the unstable weather station firmware (memory leak). Then it is time to reset the station ...
Therefore I've implemented a rule that supervises the AVG period on a regular base (once a hour). If it is getting too big over a too long time, I trigger a reset of the station.

Trigger a reset of the weather station could be a feature request for this binding ;-)

Skinah:
Yes I do mean the not yet merged binding which is why I was removing the mention of it as I expected it to be merged by now. Yes there is similar functionality between the two and the reference is no longer needed. I would not be keen on increasing the scope of this PR until it is clear that having two bindings with similar/crossing functionality is not an issue. So I would prefer to get this merged and only then consider feature requests.

Regarding triggering a reset of the station yes it can be done for the RJ45 model (already implemented and works for the scraping method), but the WiFi model does not support a reset by calling a URL... In addition both of the models have a setting in their firmware that if they lose connection to the server for >20 minutes they can auto restart. This could be implemented pretty easily, but if stations from all brands that are weather underground capable have this ability, it may be better to contribute that functionality to the other binding.

Regarding use of midnight VS a number, I feel the number is easier for non English people to recognize even if they are confused. It is not hard to connect what is intended in the context especially when they are used to seeing the rainfall reset at midnight on the weather stations screen. I really don't feel strongly either way on it.

bundles/org.openhab.binding.ipobserver/README.md
```diff
@@ -15,15 +19,20 @@ There is only one thing that can be added and is called `weatherstation`.
 
 ## Discovery
 
-Auto discovery is supported and may take a while to complete as it scans all IP addresses on your network one by one.
+Auto discovery is supported for the RJ45 models, and the wifi ipobservers will need to be manually added. Discovery may take a while to complete as it scans all IP addresses on your network one by one.
```
lolodomo, line 22: Please go to a new line after each `.`
I am certainly not fluent in English but isn't "while" more appropriate in this case ? Auto discovery is supported for the RJ45 models, **while** the wifi ipobservers will need to be manually added.

bundles/org.openhab.binding.ipobserver/README.md
```diff
@@ -15,15 +19,20 @@ There is only one thing that can be added and is called `weatherstation`.
...
 
 ## Thing Configuration
 
+If the id and password are supplied, the binding will work via the second WU push method, if they are left blank, the binding will work in the scraping mode (RJ45 model only).
```
lolodomo, line 26:
```
... the second WU push method.
If they are left blank, ...
```

bundles/org.openhab.binding.ipobserver/README.md
```diff
@@ -15,15 +19,20 @@ There is only one thing that can be added and is called `weatherstation`.
...
+When the id and password is supplied you need to set the custom WU path to be `/weatherstation/updateweatherstation.php` and the port to be the same as openHAB which is port 8080 by default.
```
lolodomo, line 27: When the id and password **are** supplied**,** you need to set the custom WU path to `/weatherstation/updateweatherstation.php` and ...
I think "set the xxx to yyy" rather than "set the xxx to be yyy" ?

bundles/org.openhab.binding.ipobserver/src/main/java/org/openhab/binding/ipobserver/internal/IpObserverHandler.java
```diff
@@ -135,9 +138,47 @@ public void processValue(String sensorValue) {
...
+     * into keys and values made up from the weather stations readings.
+     *
+     * @param update
+     */
+    public void processServerQuery(String update) {
+        if (update.startsWith(idPass)) {
+            String matchedUpdate = update.substring(idPass.length() + 1, update.length());
+            logger.trace("Update received:{}", matchedUpdate);
+            updateState(LAST_UPDATED_TIME, new DateTimeType(ZonedDateTime.now()));
+            Map<String, String> mappedQuery = new HashMap<String, String>();
```
lolodomo, line 158: `Map<String, String> mappedQuery = new HashMap<>();`

bundles/org.openhab.binding.ipobserver/src/main/java/org/openhab/binding/ipobserver/internal/IpObserverHandler.java
```diff
@@ -244,6 +285,27 @@ private void createChannelHandler(String chanName, Class<? extends State> type,
...
+        createChannelHandler(TEMP_INDOOR, QuantityType.class, ImperialUnits.FAHRENHEIT, "indoortempf");
+        createChannelHandler(TEMP_OUTDOOR, QuantityType.class, ImperialUnits.FAHRENHEIT, "tempf");
+        createChannelHandler(TEMP_WIND_CHILL, QuantityType.class, ImperialUnits.FAHRENHEIT, "windchillf");
+        createChannelHandler(TEMP_DEW_POINT, QuantityType.class, ImperialUnits.FAHRENHEIT, "dewptf");
+        createChannelHandler(HOURLY_RAIN_RATE, QuantityType.class, ImperialUnits.INCH, "rainin");
+        createChannelHandler(DAILY_RAIN, QuantityType.class, ImperialUnits.INCH, "dailyrainin");
+        createChannelHandler(WEEKLY_RAIN, QuantityType.class, ImperialUnits.INCH, "weeklyrainin");
+        createChannelHandler(MONTHLY_RAIN, QuantityType.class, ImperialUnits.INCH, "monthlyrainin");
+        createChannelHandler(YEARLY_RAIN, QuantityType.class, ImperialUnits.INCH, "yearlyrainin");
+        createChannelHandler(UV_INDEX, DecimalType.class, SIUnits.CELSIUS, "UV");
```
lolodomo, line 301: The unit is SIUnits.CELSIUS just because you need to define a unit ?

bundles/org.openhab.binding.ipobserver/src/main/java/org/openhab/binding/ipobserver/internal/IpObserverHandler.java
```diff
@@ -333,11 +395,19 @@ private void setupChannels() {
     public void initialize() {
         config = getConfigAs(IpObserverConfiguration.class);
         updateStatus(ThingStatus.UNKNOWN);
-        pollingFuture = scheduler.scheduleWithFixedDelay(this::pollStation, 1, config.pollTime, TimeUnit.SECONDS);
+        if (!config.id.isEmpty() && !config.password.isEmpty()) {
```
lolodomo, line 398: You could even use `isBlank`

bundles/org.openhab.binding.ipobserver/src/main/java/org/openhab/binding/ipobserver/internal/IpObserverHandler.java
```diff
@@ -333,11 +395,19 @@ private void setupChannels() {
     public void initialize() {
         config = getConfigAs(IpObserverConfiguration.class);
         updateStatus(ThingStatus.UNKNOWN);
```
lolodomo, line 397: Move this updateStatus in your else branch, just before calling your scheduled task.

bundles/org.openhab.binding.ipobserver/src/main/java/org/openhab/binding/ipobserver/internal/IpObserverUpdateReceiver.java
```diff
@@ -0,0 +1,74 @@
...
+/**
+ * The {@link IpObserverUpdateReceiver} captures any updates sent to the openHAB Jetty server if the weather station is
+ * setup to direct the weather updates to the HTTP server of openHAB which is normally port 8080.
+ *
+ * @author Matthew Skinner - Initial contribution
+ */
+@NonNullByDefault
+public class IpObserverUpdateReceiver extends HttpServlet {
+    private final Logger logger = LoggerFactory.getLogger(this.getClass());
+    private static final long serialVersionUID = -234658674L;
```
lolodomo, line 41: Please move up the static before the non static

bundles/org.openhab.binding.ipobserver/src/main/java/org/openhab/binding/ipobserver/internal/IpObserverUpdateReceiver.java
```diff
@@ -0,0 +1,74 @@
...
+    private ArrayList<IpObserverHandler> listOfHandlers = new ArrayList<>(1);
```
lolodomo, line 42: `private List<IpObserverHandler> listOfHandlers = new ArrayList<>(1);`

bundles/org.openhab.binding.ipobserver/src/main/resources/OH-INF/thing/thing-types.xml
```diff
@@ -10,6 +10,8 @@
 		<channels>
 			<channel id="temperatureIndoor" typeId="temperatureIndoor"/>
 			<channel id="temperatureOutdoor" typeId="system.outdoor-temperature"/>
+			<channel id="temperatureWindChill" typeId="temperatureWindChill"/>
```
lolodomo, line 13: You need to update the default translation files (properties).

lolodomo (changes requested):

bundles/org.openhab.binding.ipobserver/src/main/java/org/openhab/binding/ipobserver/internal/IpObserverUpdateReceiver.java
Skinah: DONE

bundles/org.openhab.binding.ipobserver/src/main/java/org/openhab/binding/ipobserver/internal/IpObserverHandler.java
Skinah: Yes that is correct. It is not a typo or bug and it works fine.

bundles/org.openhab.binding.ipobserver/README.md
Skinah: Yes `while` sounds much better, thank you.

bundles/org.openhab.binding.ipobserver/README.md
Skinah: I have reworked to be what I think you were suggesting.

lolodomo:
Code is now OK for me except maybe semantic tags for which I am not sure that is what we expect.
@kaikreuzer : is it ok for you to add tags measurement + temperature on "Wind Chill Temperature Outdoors" and "Dew Point Temperature Outdoors" ?
More generally, as tags were assigned almost on all channels of this binding can you please check if you find something wrong with semantic tags in this file:
https://github.com/openhab/openhab-addons/blob/main/bundles/org.openhab.binding.ipobserver/src/main/resources/OH-INF/thing/thing-types.xml

Skinah:
IMHO we should not tag windchill. The semantic model will put these on the UI and they will confuse users as windchill is a different temperature to the `real` temperature. Realtime and correct readings only should be tagged, so I removed some of the tags. Max Wind Gust for the entire day is one example. If an icon showing that value is shown in the UI it can be confusing as it does not reflect a realtime measurement.

Thanks for looking over the binding and the review.

lolodomo (approved):
LGTM

lolodomo:
As the new channels have now no semantic tags, I see nothing blocking the merge.
By the way, I am going to open a new issue to share the current guidelines about semantic tags and to check all bindings that defined semantic tags on channels. You can discuss again about this binding in this new issue.

PR #12152, opened 2022-01-29
[mqtt] Add birth and shutdown message

### LINKED ISSUE #12104: [mqtt] add a correct LWT mechanism
As stated in https://community.openhab.org/t/openhab-sending-out-its-lwt/132313/5 there is a missing feature when it comes to LWT for openHAB itself.

The addon should provide full LWT functionality, this is:

- If openHAB connection to the broker is lost, send a message -> this part is working as expected
- If openHAB is shut down properly, send a message before closing connection -> seems not to work
- If openHAB is started send another message -> not implemented yet, one has to do it manually via rules.

jimtng:
Support birth and shutdown message for OpenHAB MQTT to complete the LWT feature. Resolve #12104

### DIFF
--- a/bundles/org.openhab.binding.mqtt/src/main/resources/OH-INF/thing/thing-types.xml
+++ b/bundles/org.openhab.binding.mqtt/src/main/resources/OH-INF/thing/thing-types.xml
@@ -94,6 +94,44 @@
 				<advanced>true</advanced>
 			</parameter>
 
+			<parameter name="birthMessage" type="text">
+				<label>Birth Message</label>
+				<description>The message to send to the broker when a connection is established.</description>
+				<advanced>true</advanced>
+			</parameter>
+
+			<parameter name="birthTopic" type="text">
+				<label>Birth Topic</label>
+				<description>Defaults to empty and therefore disables the birth message.</description>
+				<advanced>true</advanced>
+			</parameter>
+
+			<parameter name="birthRetain" type="boolean">
+				<label>Birth Message Retain</label>
+				<description>True if the birth message should be retained (defaults to true)</description>
+				<default>true</default>
+				<advanced>true</advanced>
+			</parameter>
+
+			<parameter name="shutdownMessage" type="text">
+				<label>Shutdown Message</label>
+				<description>The message to send to the broker before the connection terminates.</description>
+				<advanced>true</advanced>
+			</parameter>
+
+			<parameter name="shutdownTopic" type="text">
+				<label>Shutdown Topic</label>
+				<description>Defaults to empty and therefore disables the shutdown message.</description>
+				<advanced>true</advanced>
+			</parameter>
+
+			<parameter name="shutdownRetain" type="boolean">
+				<label>Shutdown Message Retain</label>
+				<description>True if the shutdown message should be retained (defaults to true)</description>
+				<default>false</default>
+				<advanced>true</advanced>
+			</parameter>
+
 			<parameter name="username" type="text">
 				<label>Username</label>
 				<description>The MQTT username</description>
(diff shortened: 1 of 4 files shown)


bundles/org.openhab.binding.mqtt/src/main/resources/OH-INF/thing/thing-types.xml
```diff
@@ -94,6 +94,44 @@
...
+				<description>Defaults to empty and therefore disables the shutdown message.</description>
+				<advanced>true</advanced>
+			</parameter>
+
+			<parameter name="shutdownRetain" type="boolean">
+				<label>Shutdown Message Retain</label>
+				<description>True if the shutdown message should be retained (defaults to true)</description>
+				<default>false</default>
+				<advanced>true</advanced>
+			</parameter>
```
jlaur, line 133: Please add these texts resources/OH-INF/i18n/mqttbroker.properties also - either manually or by running `mvn i18n:generate-default-translations`.

jlaur (changes requested):
Looks good, only i18n texts missing.

bundles/org.openhab.binding.mqtt/src/main/resources/OH-INF/thing/thing-types.xml
jimtng: `mvn i18n:generate-default-translations` doesn't work for me.

[ERROR] No plugin found for prefix 'i18n' in the current project and in the plugin groups [org.apache.maven.plugins, org.codehaus.mojo] available from the repositories [local (~/.m2/repository), central (https://repo1.maven.org/maven2), openhab-release (https://openhab.jfrog.io/openhab/libs-release)] -> [Help 1]

jimtng: This command worked for me:

```
mvn org.openhab.core.tools:i18n-maven-plugin:3.2.0:generate-default-translations
```

jlaur: Hmm, there's already a mqtt**broker**.properties file. Can you check what's up with that? I just checked that previous parameter strings (compared to your changes) existed in that file, but don't know why it's named like that and if it is actually correctly found and used. At least there seems to be a conflict.

jimtng: Good catch! I just ran the command above. Either the command I used is outdated, or the other files used an outdated naming scheme? Either way, we'd need to pick one or the other. What would you suggest?

jimtng: I think I understand this now that I've had a proper look into it. The name should be mqtt.properties as per what was just generated by the i18n tool. I'll push a fix shortly to include the other translations.

jlaur: So the other file never had any effect? Could be tested by simply adding two different strings to one of the strings in each file so it would be visible which file is actually used.

jimtng: @jlaur, I will push the properties fix in a separate PR so that it is noted in the change log. The existing translations aren't taking effect currently.

jimtng: The files are renamed in #12193

jlaur (approved):
LGTM - thanks!

PR #12155, opened 2022-01-29
[voicerss] Fix and enhance the external tool to prefill the cache

lolodomo:
Make the tool independent of core openHAB

Fix the parsing of the command arguments.
Allow to provide optional audio codec and format.

Null annotations added on most of the classes

### DIFF
--- a/bundles/org.openhab.voice.voicerss/src/main/java/org/openhab/voice/voicerss/internal/VoiceRSSTTSService.java
+++ b/bundles/org.openhab.voice.voicerss/src/main/java/org/openhab/voice/voicerss/internal/VoiceRSSTTSService.java
@@ -176,9 +176,11 @@ public class VoiceRSSTTSService implements TTSService {
                 case "MP3":
                     audioFormats.add(new AudioFormat(AudioFormat.CONTAINER_NONE, AudioFormat.CODEC_MP3, null, 16, 64000,
                             44_100L));
+                    break;
                 case "OGG":
                     audioFormats.add(new AudioFormat(AudioFormat.CONTAINER_OGG, AudioFormat.CODEC_VORBIS, null, 16,
                             null, 44_100L));
+                    break;
                 case "AAC":
                     audioFormats.add(new AudioFormat(AudioFormat.CONTAINER_NONE, AudioFormat.CODEC_AAC, null, 16, null,
                             44_100L));
--- a/bundles/org.openhab.voice.voicerss/src/main/java/org/openhab/voice/voicerss/internal/VoiceRSSTTSService.java
+++ b/bundles/org.openhab.voice.voicerss/src/main/java/org/openhab/voice/voicerss/internal/VoiceRSSTTSService.java
@@ -170,7 +170,80 @@ public class VoiceRSSTTSService implements TTSService {
      * @return The audio formats of this instance
      */
     private Set<AudioFormat> initAudioFormats() {
-        return voiceRssImpl.getAvailableAudioFormats();
+        Set<AudioFormat> audioFormats = new HashSet<>();
+        for (String codec : voiceRssImpl.getAvailableAudioCodecs()) {
+            switch (codec) {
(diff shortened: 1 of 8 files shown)


lolodomo:
To run this tool in a console command, you need to specify the classpath of 1 the voicerss jar and 2 the core.audio jar. This is even not enough as sl4j jar is still missing. I don't find this sl4j jar in a standard user installation.
The only solution I found was to comment all usage of Logger in voicerss code (this is not part of the current PR) and then I am able to run successfully the tool in a console command.

Should I update the POM file to build a specific additional jar (for this tool) including all the necessary dependencies ?
Advice would be appreciated.

lolodomo:
I just realized the dependency of the cache tool with core openHAB was introduced by recently merged PR #11916 .
I will have to partially revert this PR to have the cache tool non depending on openHAB core.
But this will not solve the dependency with sl4j.

lolodomo:
The 2 problems are solved:

1. Dependency of the cache tool with openHAB core is removed
2. Requirement of slf4j to use the cache tool is removed

The PR is ready for a review.

bundles/org.openhab.voice.voicerss/src/main/java/org/openhab/voice/voicerss/internal/VoiceRSSTTSService.java
```diff
@@ -170,7 +170,80 @@ public AudioStream synthesize(String text, Voice voice, AudioFormat requestedFor
...
      */
     private Set<AudioFormat> initAudioFormats() {
-        return voiceRssImpl.getAvailableAudioFormats();
+        Set<AudioFormat> audioFormats = new HashSet<>();
+        for (String codec : voiceRssImpl.getAvailableAudioCodecs()) {
+            switch (codec) {
+                case "MP3":
+                    audioFormats.add(new AudioFormat(AudioFormat.CONTAINER_NONE, AudioFormat.CODEC_MP3, null, 16, 64000,
+                            44_100L));
+                case "OGG":
```
jlaur, line 179: Fall-through on purpose or missing `break`? Consider adding a comment for this in intended.

bundles/org.openhab.voice.voicerss/src/main/java/org/openhab/voice/voicerss/internal/VoiceRSSTTSService.java
```diff
@@ -170,7 +170,80 @@ public AudioStream synthesize(String text, Voice voice, AudioFormat requestedFor
...
+                    audioFormats.add(new AudioFormat(AudioFormat.CONTAINER_OGG, AudioFormat.CODEC_VORBIS, null, 16,
+                            null, 44_100L));
+                case "AAC":
```
jlaur, line 182: Fall-through on purpose or missing `break`? Consider adding a comment for this in intended.

jlaur (changes requested):
Not a full review for now, just spotted something that might be worth checking.

bundles/org.openhab.voice.voicerss/src/main/java/org/openhab/voice/voicerss/internal/VoiceRSSTTSService.java
lolodomo: Break is missing! To be fired

lolodomo: Fixed

bundles/org.openhab.voice.voicerss/src/main/java/org/openhab/voice/voicerss/internal/VoiceRSSTTSService.java
lolodomo: Fixed

jlaur (approved):
LGTM

PR #11375, opened 2021-10-12
[teleinfo] Add support for Standard tic mode

olivierkeke:
Resolves #10527

This PR adds support for new Standard tic mode introduced by Linky Telemeters.
This update is fully backward compatible with old configs.
Early versions have been tested by @Jacques83300.
Binaries can be found in [this community thread](https://community.openhab.org/t/teleinfo-binding-standard-tic-mode-support-for-linky-electricitymeter/121835).

PR #12156, opened 2022-01-29
[teleinfo] Update translation files

olivierkeke:
Signed-off-by: Olivier Marceau <hollysaiqs@marceau.ovh>

It seems that descriptions for channels added in the PR #11375 don't appear on crowdin, so this PR updates translation files.
This applies changes made by `mvn org.openhab.core.tools:i18n-maven-plugin:3.2.0:generate-default-translations -pl :org.openhab.binding.teleinfo` command.


bundles/org.openhab.binding.teleinfo/src/main/resources/OH-INF/i18n/teleinfo.properties
```diff
@@ -70,10 +283,22 @@ channel-type.teleinfo.couleurDemainType.description = Following day color
 channel-type.teleinfo.couleurDemainType.state.option.Bleu = Bleu
 channel-type.teleinfo.couleurDemainType.state.option.Blanc = Blanc
 channel-type.teleinfo.couleurDemainType.state.option.Rouge = Rouge
+channel-type.teleinfo.currentType.label = Teleinfo Current Type
+channel-type.teleinfo.currentType.label = Teleinfo Current Type
+channel-type.teleinfo.currentType.label = Teleinfo Current Type
```
lolodomo, line 288: It looks like you have several duplicated entries in the "channel types" section.
You have 7 channel types appearing 3 times.
Please check what would be really expected.
Would it be a bug in the i18n tool ?

lolodomo (changes requested):

bundles/org.openhab.binding.teleinfo/src/main/resources/OH-INF/i18n/teleinfo.properties
olivierkeke: I've removed the duplicated lines. 
I think the i18n tool generates these entries because identical channel types are defined in several xml files.

lolodomo (approved):
LGTM

bundles/org.openhab.binding.teleinfo/src/main/resources/OH-INF/i18n/teleinfo.properties
lolodomo: You should not have a channel type defined several times. That is certainly the root cause and what should be fixed. The common solution is to put channel types in a film named channels.xml.

olivierkeke: Yes I probably should do some clean up in the xml config files.

I wanted to have a separate xml file for each `channel-group-type` and it seems that `channel-group-type` cannot use a `channel-type` from another file so I end up duplicating `channel-type` in each xml file.

What you suggest is to aggregate every `channel-group-type` and `channel-type` in the same xml file? Isn't there a way to keep separate files?

lolodomo: I am not aware of this limit. 
I have not played with groups of channels very often. 
I could be wrong but I think I already saw channels groups and channel types defined in separate files. Let me search an example.

lolodomo: You can have a look at the NibeUplink Binding: `channel-group-type` and `channel-type` are defined in different XML files.
Apparently there is no constraint to have both in the same XML file.

olivierkeke: Thank you! The problem should have been located between chair and keyboard, as usual. I've submitted a new PR: #12204.

PR #12170, opened 2022-01-30
[mqtt] publishMQTT Action accepts a bytearray payload

### LINKED ISSUE #10762: [mqtt] add bytearray payload support in MQTTActions.java
Strings payloads have to convert to an bytearray anyway.
So it would be great if it can be addressed directly.

[openhab-addons/bundles/org.openhab.binding.mqtt/src/main/java/org/openhab/binding/mqtt/internal/action/MQTTActions.java](https://github.com/openhab/openhab-addons/blob/19029b73589fecbc1e4eb0d1979b880ef4243edc/bundles/org.openhab.binding.mqtt/src/main/java/org/openhab/binding/mqtt/internal/action/MQTTActions.java#L79)

jimtng:
Resolve #10762

This PR adds variants of publishMQTT that accept byte array instead of String.

### DIFF
--- a/bundles/org.openhab.binding.mqtt.generic/README.md
+++ b/bundles/org.openhab.binding.mqtt.generic/README.md
@@ -220,6 +220,7 @@ Once this action instance is retrieved, you can invoke the `publishMQTT(String t
 ```
 mqttActions.publishMQTT("mytopic","myvalue", true)
 ```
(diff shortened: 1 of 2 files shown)


lolodomo (commented):
Would it make sense to add the correspinding static methods? They are required for DSL rules I believe.

jimtng:
I've added it to the static methods. Also updated the README.

lolodomo (approved):
LGTM

PR #12175, opened 2022-01-30
[hdpowerview] Add shade identify command

### LINKED ISSUE #12174: [hdpowerview] Add command for identifying a shade
After #12138 and #12061 it would make sense to add another shade command, **IDENTIFY**, like the similar repeater command. This command should jog the shade same way as shown here in the PowerView app:

image

jlaur:
Fixes #12174

Add a command for identifying a shade by jogging it.

Extends #12138 to have more than one command and provides same functionality for shades as introduced for repeaters in #12061.

### DIFF
--- a/bundles/org.openhab.binding.hdpowerview/src/main/resources/OH-INF/i18n/hdpowerview_da.properties
+++ b/bundles/org.openhab.binding.hdpowerview/src/main/resources/OH-INF/i18n/hdpowerview_da.properties
@@ -12,7 +12,6 @@ channel-type.hdpowerview.repeater-identify.description = Identificer ved at lade
 channel-type.hdpowerview.repeater-identify.command.option.IDENTIFY = Identificer
 channel-type.hdpowerview.shade-command.label = Kommando
 channel-type.hdpowerview.shade-command.description = Send en kommando til gardinet
-channel-type.hdpowerview.shade-command.command.option.IDENTIFY = Identificer
 channel-type.hdpowerview.shade-command.command.option.CALIBRATE = Kalibrer
 
 # dynamic channels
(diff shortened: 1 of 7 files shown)


bundles/org.openhab.binding.hdpowerview/src/main/resources/OH-INF/i18n/hdpowerview_da.properties
```diff
@@ -12,6 +12,7 @@ channel-type.hdpowerview.repeater-identify.description = Identificer ved at lade
 channel-type.hdpowerview.repeater-identify.command.option.IDENTIFY = Identificer
 channel-type.hdpowerview.shade-command.label = Kommando
 channel-type.hdpowerview.shade-command.description = Send en kommando til gardinet
+channel-type.hdpowerview.shade-command.command.option.IDENTIFY = Identificer
```
lolodomo, line 15: Please do not update translation files other than the default one. It would I make translation in Crowdin out of sync.. 
This German translation should be done in Crowdin once the PR is merged.

lolodomo (changes requested):

bundles/org.openhab.binding.hdpowerview/src/main/resources/OH-INF/i18n/hdpowerview_da.properties
jlaur: This one as Danish and since it's a new string added in same PR it can't conflict with Crowdin?

lolodomo: It will not be loaded on Crowdin automatically. 
Good practice is to not provide translations in PRe except of course the default translation. 

A workaround exists but it requires a manual action in Crowdin.

jlaur: Okay, sure, I'll stop this approach. I can remove the string from the PR tonight.

lolodomo: Maybe a possible exception for a fully new binding, if translations are provided in the PR, it should be merged only if the one who merged take the action to load the files in Crowdin.

jlaur: @lolodomo - do we have any written guidelines? Also wondering about best practice when a key is renamed or removed. Should translations be left with keys that no longer exists? This would also require manual retranslations as opposed to developer renaming the keys in translations to be in sync with default strings?

jlaur: Removed.

lolodomo (approved):
LGTM

PR #12178, opened 2022-01-31
[rfxcom] Better support for Legrand CAD protocol

jpenninkhof:
# Title

Better support for Legrand CAD protocol

# Description

Improvement: Le Grand CAD is a relatively old protocol, supported by the RFXComXL device. The protocol as applies by LeGrand doesn't use certain fields such as Unitcode and a Command. Whenever a CAD message is sent, it is always of type TOGGLE (On or Off doesn't exist). This does cause a couple of challenges within OH3, but with this patch OH at least doesn't throw exceptions anymore when a CAD command is received.

Signed-off-by: Jan Penninkhof <jan@penninkhof.com>

### DIFF
--- a/bundles/org.openhab.binding.rfxcom/src/main/java/org/openhab/binding/rfxcom/internal/messages/RFXComLighting5Message.java
+++ b/bundles/org.openhab.binding.rfxcom/src/main/java/org/openhab/binding/rfxcom/internal/messages/RFXComLighting5Message.java
@@ -59,6 +59,7 @@ public class RFXComLighting5Message extends RFXComDeviceMessageImpl<RFXComLighti
         EURODOMEST(9),
         LIVOLO_APPLIANCE(10),
         MDREMOTE_107(12),
+        LEGRAND_CAD(13),
         AVANTEK(14),
         IT(15),
         MDREMOTE_108(16),
@@ -118,7 +119,8 @@ public class RFXComLighting5Message extends RFXComDeviceMessageImpl<RFXComLighti
         TOGGLE_6(0x0B, LIVOLO_APPLIANCE),
         TOGGLE_7(0x06, LIVOLO_APPLIANCE),
         TOGGLE_8(0x0A, LIVOLO_APPLIANCE),
-        TOGGLE_9(0x05, LIVOLO_APPLIANCE);
+        TOGGLE_9(0x05, LIVOLO_APPLIANCE),
+        TOGGLE(0x00, LEGRAND_CAD);
 
         private final int command;
         private final List<SubType> supportedBySubTypes;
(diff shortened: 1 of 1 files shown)


lolodomo:
PR not properly signed

lolodomo:
@martinvw : is it ok for you?

martinvw:
> @martinvw : is it ok for you? 

Sorry, completely missed this one, I'll take a look now

bundles/org.openhab.binding.rfxcom/src/main/java/org/openhab/binding/rfxcom/internal/messages/RFXComLighting5Message.java
```diff
@@ -178,9 +180,16 @@ public void encodeMessage(byte[] data) throws RFXComException {
         subType = fromByte(SubType.class, super.subType);
 
         sensorId = (data[4] & 0xFF) << 16 | (data[5] & 0xFF) << 8 | (data[6] & 0xFF);
-        unitCode = data[7];
 
-        command = fromByte(Commands.class, data[8], subType);
+        switch (subType) {
```
martinvw, line 184: For a single case this repo normally just uses an if

bundles/org.openhab.binding.rfxcom/src/main/java/org/openhab/binding/rfxcom/internal/messages/RFXComLighting5Message.java
```diff
@@ -178,9 +180,16 @@ public void encodeMessage(byte[] data) throws RFXComException {
...
+            case LEGRAND_CAD:
+                unitCode = 0x01;
```
martinvw, line 186: Is this really needed, reading the SDK unit-codes are not used with a lot the sub-types so this distinction is not perse correct

bundles/org.openhab.binding.rfxcom/src/main/java/org/openhab/binding/rfxcom/internal/messages/RFXComLighting5Message.java
```diff
@@ -178,9 +180,16 @@ public void encodeMessage(byte[] data) throws RFXComException {
...
+                command = Commands.TOGGLE;
+                break;
+            default:
+                unitCode = data[7];
+                command = fromByte(Commands.class, data[8], subType);
```
martinvw, line 191: I would expect hope this call could also be used for the newly added command

martinvw (changes requested):
My proposal would be to just add the two enum values, I would not expect the other changes to be needed.

lolodomo:
@jpenninkhof : did you see the review comments form @martinvw ?

lolodomo:
Closing this PR as there was no activity/feedback since 9 months.
@penninkhof : reopen it in case you would like to finish that PR.

PR #12179, opened 2022-01-31
[lifx] Add support for new LIFX products

wborn:
Allows for discovering all the new LIFX products and their features.

See: https://github.com/LIFX/products/blob/master/products.json


lolodomo (approved):
LGTM

PR #12185, opened 2022-02-01
[openthermgateway] Rationalize exception handling & decouple tasks on different threads.

andrewfg:
**Background**

There are several issues (bugs) concerning the communications between the binding and the OTGW ..

1.  The `OpenThermGatewaySocket` connector was not catching some IOExceptions properly. In particular it was not catching `SocketReadTimeout` exceptions when reading from the OTGW, so the Thing could go permanently offline (even if `connectionRetryInterval` had a non zero value); thus requiring the user to manually toggle the Thing Offline and Online again.
2. The `OpenThermGatewaySocket` connector was not catching and handling write errors.
3. The `OpenThermGatewaySocket` polling loop runs on a different thread than the `OpenThermGatewayHandler` (inside the framework), so cross thread callbacks had the potential to block on each other.
4. The `OpenThermGatewaySocket` `sendCommand()` method could be called in parallel from two different threads without proper synchronization.

**Solution**

This PR fixes all of the above issues.

### DIFF
--- a/bundles/org.openhab.binding.openthermgateway/src/main/java/org/openhab/binding/openthermgateway/handler/OpenThermGatewayHandler.java
+++ b/bundles/org.openhab.binding.openthermgateway/src/main/java/org/openhab/binding/openthermgateway/handler/OpenThermGatewayHandler.java
@@ -70,11 +70,7 @@ public class OpenThermGatewayHandler extends BaseThingHandler implements OpenThe
     @Override
     public void initialize() {
         logger.debug("Initializing OpenTherm Gateway handler for uid '{}'", getThing().getUID());
-
-        updateStatus(ThingStatus.UNKNOWN, ThingStatusDetail.NONE, "Initializing");
-
         config = getConfigAs(OpenThermGatewayConfiguration.class);
-
         connect();
     }
 
@@ -126,32 +122,47 @@ public class OpenThermGatewayHandler extends BaseThingHandler implements OpenThe
                     updateState(OpenThermGatewayBindingConstants.CHANNEL_OVERRIDE_CENTRAL_HEATING2_ENABLED,
                             OnOffType.from(!gatewayCommand.getMessage().equals("0.0")));
                 }
-            } else {
-                connect();
             }
         }
     }
 
     @Override
     public void connecting() {
+        // decouple threads via a scheduler task
+        scheduler.submit(() -> connectingTask());
+    }
+
+    private void connectingTask() {
         connecting = true;
         updateStatus(ThingStatus.UNKNOWN, ThingStatusDetail.NONE, "Connecting");
     }
 
     @Override
     public void connected() {
+        // decouple threads via a scheduler task
+        scheduler.submit(() -> connectedTask());
+    }
+
+    private void connectedTask() {
         connecting = false;
         updateStatus(ThingStatus.ONLINE);
     }
 
     @Override
     public void disconnected() {
+        if (!explicitDisconnect) {
+            // decouple threads via a scheduler task
+            scheduler.submit(() -> disconnectedTask());
+        }
+    }
+
+    private void disconnectedTask() {
         @Nullable
         OpenThermGatewayConfiguration conf = config;
 
         connecting = false;
 
-        updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.NONE, "Disconnected");
+        updateStatus(ThingStatus.OFFLINE);
 
         // retry connection if disconnect is not explicitly requested
         if (!explicitDisconnect && conf != null && conf.connectionRetryInterval > 0) {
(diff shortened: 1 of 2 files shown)


ArjenKorevaar:
As mentioned in the thread, this issue was already resolved. Either remove this PR or rebase onto https://github.com/openhab/openhab-addons/pull/11203 please as that PR still awaits reviewing.

andrewfg:
> Either remove this PR or rebase onto #11203 please

Actually I think you should rebase on my PR (not the other way round). Mine is a bug fix that can be merged immediately, whereas I think that yours may require longer discussions.  So in the meantime with my PR at least users will have a system that works..

lolodomo:
The other PR is mentioning no bug fix.

andrewfg:
> is better to separate bug fixes and enhancements in separate PRs

That is my opinion too. So I hope it’s not impertinent to ask if you could review this one?? :)

lolodomo:
Very impertinent :)

bundles/org.openhab.binding.openthermgateway/src/main/java/org/openhab/binding/openthermgateway/handler/OpenThermGatewayHandler.java
```diff
@@ -126,32 +122,47 @@ public void handleCommand(ChannelUID channelUID, Command command) {
...
-            } else {
-                connect();
             }
         }
     }
 
     @Override
     public void connecting() {
+        // decouple threads via a scheduler task
+        scheduler.submit(() -> connectingTask());
```
jlaur, line 132: Why does this need to be done in separate thread?

ArjenKorevaar:
In retrospect, sure, but I have been pretty much the only developer on this binding since the start, at the end of refactoring large parts and fixed the bug along the way. The PR review process takes way too much time as it is already, so I don't need the extra unnecessary work of merging a PR that fixes something that is already fixed in the existing PR. Something that could have easily been prevented by looking at the discussion on the community or with a simple message. Most users are already using the new unofficial version anyway, because of the support for new openterm devices, and have also confirmed the bug fixed as tested on different OTGW's and serial to ethernet converters. Something that cannot be said from this PR? It's not up to me to review or accept this PR, but I won't spend the extra time on merging with the other PR.

bundles/org.openhab.binding.openthermgateway/src/main/java/org/openhab/binding/openthermgateway/handler/OpenThermGatewayHandler.java
andrewfg: I found that the callback tasks, which access the framework, sometimes block; which means that the polling thread gets delayed. And by firing a scheduler.submit task, it can completely be decoupled.

andrewfg:
> I have been pretty much the only developer on this binding since the start

@ArjenKorevaar I do apologise if you feel upset; I am not sure if you want to guard your ‘baby’ to yourself; or if you would like to encourage other developers (like me) to become involved? Certainly, if the latter, then I would also be happy to review your own PR too, and help it to become adopted.

> The PR review process takes way too much time as it is already

I do agree. However in my experience, if one makes a series of small ‘atomic’ PR changes, it is easier for the maintainers to overview them, and approve them faster. I think your PR is quite a big change affecting many files and structures, so that could explain why it has taken so long. ??

For example on the HDPowerView binding, @jlaur and @lolodomo have together made and approved about 10 PR changes in about 1 month.

... there is a riddle: Q: “how do you eat an elephant” / A: “slice by slice”..

andrewfg:
^
Ok @ArjenKorevaar let us try not to get into a fight over this. I will look over you PR later today or tomorrow, and a) look for any potential merge conflicts, b) do a general code review, and c) test it on my OTGW as well.

bundles/org.openhab.binding.openthermgateway/src/main/java/org/openhab/binding/openthermgateway/handler/OpenThermGatewayHandler.java
jlaur: I missed that it was a callback, thanks.

jlaur:
@andrewfg - if you want to prevent this from getting merged, you can mark is as draft. It seems you contributed to #11203 (instead?) but this may not be completely obvious to everyone.

andrewfg:
^
@jlaur I am riding two horses :) 

1. #11203 is probably the final answer. It includes the same fixes as this PR does (albeit refactored to match the code in the new PR), but a) it has changes and additions in over thirty files, b) it has been open already for 4 months, c) the author admits that he is quite busy with other things, and d) it is a breaking change. So it may take some time to merge it. Although I am optimistic that my contributions could help to speed things up.
2. Whereas this PR just fixes the communication read time out and thread synchronization bugs.

So certainly if 1. would be merged in time to be in the next milestone (say) then 2. can be skipped. Whereas if it looks like 1. would miss the next milestone, then I think we should at least try to get 2. into that milestone.

=> What do you think?

jlaur:
> What do you think?

Sure, let's keep both options on the table and see how far the other PR will progress before next milestone is approaching. In general I agree that it's preferable to merge small bugfix PR's continuously to not let major refactoring PR's block straight-forward bugfixes. Will you update this PR when a decision needs to be made?

andrewfg:
> Will you update this PR when a decision needs to be made?

Yes.

PR #12190, opened 2022-02-02
[voicerss] Null annotations added on the main service class

lolodomo:

### DIFF
--- a/bundles/org.openhab.voice.voicerss/src/main/java/org/openhab/voice/voicerss/tool/CreateTTSCache.java
+++ b/bundles/org.openhab.voice.voicerss/src/main/java/org/openhab/voice/voicerss/tool/CreateTTSCache.java
@@ -16,6 +16,7 @@ import java.io.BufferedReader;
 import java.io.File;
 import java.io.FileReader;
 import java.io.IOException;
+import java.io.PrintStream;
 
 import org.eclipse.jdt.annotation.NonNullByDefault;
 import org.openhab.voice.voicerss.internal.cloudapi.CachedVoiceRSSCloudImpl;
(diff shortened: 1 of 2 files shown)


bundles/org.openhab.voice.voicerss/src/main/java/org/openhab/voice/voicerss/tool/CreateTTSCache.java
```diff
@@ -119,17 +118,29 @@ private void generateCacheForMessage(String apiKey, String cacheDir, String loca
...
+        System.out.println();
+    }
+
+    private void systemOutLog(String text) {
+        System.out.println(text);
+    }
+
+    private void systemErrLog(String text) {
+        System.err.println(text);
+    }
```
jlaur, line 145: Are these wrappers really needed, they don't seem to add anything, except adding a layer to something quite standard and known? Second question, any reason to call `System.out.println()` for each line - perhaps it would be simpler (and more performant) to call `System.out.print()` with a formatted string containing all the lines to be output?

lolodomo: The only reason was to reduce the number of warnings when compiling. Should I revert ?

Yes, it should be possible, I think, to have only one unique big stirrng with `\n`.

jlaur: Ahh, that's why... nullness annotations. So the fix would be something like this:

```
    private void usage() {
        PrintStream printStream = System.out;
        if (printStream == null) {
            return;
        }
        printStream.println("Usage: java org.openhab.voice.voicerss.tool.CreateTTSCache <args>");
```

Probably not worth to output as a single string.

jlaur (changes requested):
Please see open comment.

bundles/org.openhab.voice.voicerss/src/main/java/org/openhab/voice/voicerss/tool/CreateTTSCache.java
lolodomo: Fixed.
I don't know why I did not think about that !

jlaur (approved):
LGTM

PR #12193, opened 2022-02-03
[mqtt] Fix translation file names

jimtng:
It was discovered in #12152 that the i18n properties file names were wrong. As a result, it hasn't benefited from the translated files. This PR fixes the file name and also include the new entries that were added in #12152

### DIFF
--- a/bundles/org.openhab.binding.mqtt/src/main/resources/OH-INF/i18n/mqtt.properties
+++ b/bundles/org.openhab.binding.mqtt/src/main/resources/OH-INF/i18n/mqtt.properties
@@ -10,6 +10,12 @@ thing-type.mqtt.broker.description = A connection to a MQTT broker
 
 # thing types config
 
+thing-type.config.mqtt.broker.birthMessage.label = Birth Message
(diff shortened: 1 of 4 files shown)


jlaur:
@lolodomo - can you assess if the removed strings in translation files will be a problem for synchronizing with Crowdin, i.e. if they must be deleted from upstream (Crowdin)?

lolodomo:
As files are renamed, I have no idea what will happen exactly but I guess we will have to load them manually in Crowdin (of course not the default English translation).
So I would suggest you merge and this evening I will load the 3 translation files in Crowdin.

jlaur (approved):
Thanks! LGTM.

lolodomo:
The Hungarian and Italian translations are apparently ok in Crowdin. There are few items still to be translated, certainly due to another merged PORs.
The German translations are even more advanced in Crowdin than in Git. It has to be finished in Crowdin and then it will be pushed in Git.
So I see nothing to do. When removing entries, I guess it is probably not a problem to update all the properties file in Git.

PR #12200, opened 2022-02-04
[mqtt] Remove redundant call to connectionStateChanged upon connection

jimtng:
BrokerHandler::connectionStateChanged() was being called twice upon each connection, resulting in updateState() and channel triggers being called twice. Furthermore, the birthMessage added in #12152 got published twice when a clientID is configured.

This is because connectionStateChanged is called by `MqttBrokerConnection` as its connection observer, as registered in https://github.com/openhab/openhab-addons/blob/97db73938d37d60a31755681e84d979e2e023777/bundles/org.openhab.binding.mqtt/src/main/java/org/openhab/binding/mqtt/handler/AbstractBrokerHandler.java#L109

So the additional call to connectionStateChanged in here is redundant: https://github.com/openhab/openhab-addons/blob/97db73938d37d60a31755681e84d979e2e023777/bundles/org.openhab.binding.mqtt/src/main/java/org/openhab/binding/mqtt/handler/AbstractBrokerHandler.java#L118 

This PR removed the redundant call to connectionStateChanged and fixed the birthMessage so it would be sent out just once.


jlaur (approved):
Looks good, thanks!

PR #12204, opened 2022-02-04
[teleinfo] Refactor channel config files for Linky standard mode

olivierkeke:
Signed-off-by: Olivier Marceau <hollysaiqs@marceau.ovh>

This PR removes some channel-type xml definitions duplicated in several config files.


lolodomo (approved):
LGTM

lolodomo:
Did you run the i18n tool? Normally it should generate no changes on the properties file.

olivierkeke:
> Did you run the i18n tool? Normally it should generate no changes in the properties file.

It still generates one duplicate but this one comes from an old part of the binding on which I'm less familiar. I'm investigating on it.

lolodomo:
Last question: don't you prefer to remove the file common-cbemm-evo-icc-channel-types.xml which contains only this channel?

lolodomo (approved):
Thank you

PR #12205, opened 2022-02-04
[homekit] make network onChange method synchronized

yfre:
as discussed https://github.com/openhab/openhab-addons/pull/12072#issuecomment-1029225393

it could happen that on very fast network changes the onChange method get called before the previous "onChange" method has finished the homekit bridge restart.  this could lead to reset of homekit configuration in home app

this PR adds synchronized to ensure that only only one onChange is executed at the same time. The notification is done in openhab core via SafeCaller thread so that synchronized will not block openhab core


lolodomo (approved):
LGTM

PR #12211, opened 2022-02-05
    [ojelectronics]    Fixing Issue if server returns not 200 while signing in

EvilPingu:
This fixes error #11531 if the server returns an error while signing in. In this case it is a issue with the connection or the server and not with the username or password.


lolodomo (approved):
LGTM

EvilPingu:
@lolodomo Could you please add issue #11531 to this pull request because the issue is still open

PR #12214, opened 2022-02-05
[hdpowerview] Eliminate risk of bad firmware response breaking shade/scene updates

jlaur:

This is small improvement for:
- Eliminating risk of breaking shade and scene/scene group/automation updates in case of bad firmware response.
- Updating firmware versions when binding is restarted or reconfigured.


lolodomo (approved):
LGTM

PR #12221, opened 2022-02-06
adjustments after changing null-annotations for updating properties

J-N-K:
Related to https://github.com/openhab/openhab-core/pull/2735

### DIFF
--- a/bundles/org.openhab.binding.hpprinter/src/main/java/org/openhab/binding/hpprinter/internal/HPPrinterHandler.java
+++ b/bundles/org.openhab.binding.hpprinter/src/main/java/org/openhab/binding/hpprinter/internal/HPPrinterHandler.java
@@ -155,7 +155,7 @@ public class HPPrinterHandler extends BaseThingHandler {
     }
 
     @Override
-    protected void updateProperties(final Map<String, String> properties) {
+    protected void updateProperties(final @Nullable Map<String, String> properties) {
         super.updateProperties(properties);
     }
 
--- a/bundles/org.openhab.binding.exec/src/main/java/org/openhab/binding/exec/internal/ExecWhitelistWatchService.java
+++ b/bundles/org.openhab.binding.exec/src/main/java/org/openhab/binding/exec/internal/ExecWhitelistWatchService.java
@@ -59,7 +59,7 @@ public class ExecWhitelistWatchService extends AbstractWatchService {
     }
 
     @Override
-    protected Kind<?>[] getWatchEventKinds(@Nullable Path directory) {
+    protected Kind<?> @Nullable [] getWatchEventKinds(@Nullable Path directory) {
         return new Kind<?>[] { ENTRY_CREATE, ENTRY_DELETE, ENTRY_MODIFY };
     }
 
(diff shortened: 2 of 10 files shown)


bundles/org.openhab.binding.hpprinter/src/main/java/org/openhab/binding/hpprinter/internal/HPPrinterHandler.java
```diff
@@ -155,7 +155,7 @@ protected ThingBuilder editThing() {
     }
 
     @Override
-    protected void updateProperties(final Map<String, String> properties) {
+    protected void updateProperties(final @Nullable Map<String, String> properties) {
         super.updateProperties(properties);
```
cweitkamp, line 159: In case the overridden methods just coll the super method without any other individual logic I tend to remove them completely. Wdyt?

J-N-K: That‘s not possible. `protected` is visible to sub-classes and other classes in the same package, but not to classes in the same package of the sub-class. I’m this case another class in the package of the sub-class needs access and therefore it is necessary to override it. I didn‘t think about that, too.

cweitkamp: I expected that answer and I have the same feeling that it will be clearer if the overriding method uses `public` modifier. But there is no guideline how to do it.

J-N-K:
In general I think it would be better to make these methods `protected final`, it feels wrong to expose them outside the handler because they change the thing, but that would require bigger refactoring. I can change it to `public` here if you prefer that.

lolodomo:
Few weeks ago, I reviewed a PR with a strange code to remove a property due to the fact that it was not possible to pass a null value. We should certainly find all of these workarounds in bindings code and fix them.

J-N-K:
I believe this is beyond scope here, so in general I agree.

lolodomo:
> I believe this is beyond scope here, although in general I agree.

Yes but I think it was interesting to mention for a follow-up PR.

cweitkamp:
@lolodomo Do you remember which one? The on mentioned in OHC issue (https://github.com/openhab/openhab-core/issues/2623): https://github.com/openhab/openhab-addons/pull/11692?

lolodomo:
> @lolodomo Do you remember which one? The on mentioned in OHC issue ([openhab/openhab-core#2623](https://github.com/openhab/openhab-core/issues/2623)): #11692?

Yes this one but we have to search this pattern in other bindings too.

Cossey:
I don't really understand the nullable stuff in java too well so 👍🏼  when its all working.

jaywiseman1971:
Could this be the issue that is causing the HP Thing to jump off/on?

https://community.openhab.org/t/oh-3-x-bug-things-going-offline-online-within-milliseconds/131356

Best, Jay

cweitkamp (approved):
LGTM. Thanks.

J-N-K:
I'll add the missing null annotations.

marcelrv (approved):
LGTM for miio

wborn (approved):
Thanks!

PR #12225, opened 2022-02-06
[jdbc] Fixed ClassCastException for ColorItem / HSBType

### LINKED ISSUE #12218: [jdbc] Storing and retrieving colors causes invalid cast exception
Currently if you utilize JDBC and persist a color state (i.e a HSBType) to the persistent store the data gets stored fine, however when you try and recall it an error is generated as the JDBC plugin tries to cast to an integer type. 

Examples of people reporting issue - 
https://community.openhab.org/t/color-historic-state/124209
https://community.openhab.org/t/jdbc-type-cast-error-for-color-item/126758

## Expected Behavior
When saving a Color state to the persistence store, recalling it should give the data back

## Current Behavior
Script fails with following message - 
`Script execution of rule with UID 'scratchpad' failed: class java.lang.String cannot be cast to class java.lang.Integer (java.lang.String and java.lang.Integer are in module java.base of loader 'bootstrap')`

Define a color item with persistence using JDBC 

`Color ExampleColoredLight "Test"          (gBedroom)  { channel="hue:0210:00xxxxxxxxxx:lbExampleColoredLight" }`

Then in either a scratchpad or in a rule try and change state/recall 

```
ExampleColoredLight.sendCommand("100,100,100") 
// Wait a period of time
ExampleColoredLight.previousState()
```

## Context
Unable to save the current state and recall it - meaning scripts that use the light bulb as a notification system (flash a color and revert to the previous state) doesn't work. 

## Your Environment
openHAB 3.2.0
MySQL

cweitkamp:
- Fixed ClassCastException for `ColorItem` / `HSBType`

Fixes #12218 

### DIFF
--- a/bundles/org.openhab.persistence.jdbc/src/test/java/org/openhab/persistence/jdbc/db/JdbcBaseDAOTest.java
+++ b/bundles/org.openhab.persistence.jdbc/src/test/java/org/openhab/persistence/jdbc/db/JdbcBaseDAOTest.java
@@ -76,7 +76,7 @@ public class JdbcBaseDAOTest {
     }
 
     @Test
-    public void testObjectAsStateReturnValidState() {
+    public void testObjectAsStateReturnsValidState() {
         State decimalType = jdbcBaseDAO.objectAsState(new NumberItem("testNumberItem"), null, 7.3);
         assertInstanceOf(DecimalType.class, decimalType);
         assertThat(decimalType, is(DecimalType.valueOf("7.3")));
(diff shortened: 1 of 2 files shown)


bundles/org.openhab.persistence.jdbc/src/test/java/org/openhab/persistence/jdbc/db/JdbcBaseDAOTest.java
```diff
@@ -47,6 +75,65 @@ public void setup() {
         filter = new FilterCriteria();
     }
 
+    @Test
+    public void testObjectAsStateReturnValidState() {
```
cweitkamp, line 79:
```suggestion
    public void testObjectAsStateReturnsValidState() {
```

jlaur (approved):
LGTM - thanks for the fix!

PR #12227, opened 2022-02-06
[Wemo] add @jlaur as code-owner

hmerk:


jlaur (approved):
Thanks.

hmerk:
@jlaur Don‘ think such a change needs to be added to a milestone

jlaur:
@hmerk - you are probably right. When not having any labels it won't be mentioned in the change log anyway, but probably safer/more correct not to add milestone.

kaikreuzer:
@jlaur No, as we had discussed, any PR should be tagged with a milestone.

jlaur:
> @jlaur No, as we had discussed, any PR should be tagged with a milestone.

Thanks for clarifying.

PR #12157, opened 2022-01-29
[mqtt] Remove MQTT System Broker

jimtng:
Resolve #11848 

It seems to work fine on my system.

Considering that MQTT is a key part in OpenHAB, it would be a great idea to test this. It can be compiled from my branch, or if you trust a jar: https://github.com/jimtng/openhab-addons/releases/download/v0.0.1/org.openhab.binding.mqtt-3.3.0-SNAPSHOT.jar

PR #12230, opened 2022-02-06
[mqtt] Add i18n translations file

### LINKED ISSUE #12220: [mqtt.generic] There is not yet any default translation properties file.
There is not yet any default translation properties file.

_Originally posted by @lolodomo in https://github.com/openhab/openhab-addons/issues/12167#issuecomment-1030785190_

I tried creating the translations file for `org.openhab.binding.mqtt.generic` using this command:
```
$ mvn org.openhab.core.tools:i18n-maven-plugin:3.2.0:generate-default-translations -pl :org.openhab.binding.mqtt.generic
```

It created a file called `mqtt.properties`

Questions:
1. Should the existing mqttbroker_de.properties be deleted? I believe this file is an orphaned file from an older binding structure
2. The `mqtt.properties` file that was generated from the command above, don't include the texts from `thing/channels.xml` or `thing/generic.thing.xml` -  @wborn is this perhaps a bug in the `generate-default-translations` maven plugin?

jimtng:
Resolve #12220 

Also removed the "System Broker" bridge type from xml files and README.md see #12157


bundles/org.openhab.binding.mqtt.homeassistant/src/main/resources/OH-INF/binding/binding.xml
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<binding:binding id="mqtt" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
```
lolodomo, line 2: Is it expected to use id mqrt and not an id about homeassistant?

bundles/org.openhab.binding.mqtt.homie/src/main/resources/OH-INF/binding/binding.xml
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<binding:binding id="mqtt" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
```
lolodomo, line 2: Is it expected to use id mqrt and not an id about homie?

lolodomo (commented):
No need to also do it for espmilighthub?

bundles/org.openhab.binding.mqtt.homeassistant/src/main/resources/OH-INF/binding/binding.xml
jimtng: You're correct, the id is wrong. This would override the main org.openhab.binding.mqtt

jimtng:
@lolodomo Thanks for looking into this PR. I am going to close it due to the id conflicts that you discovered. I will explain the problems further in #12220.

PR #12231, opened 2022-02-07
[ecovacs] Initial contribution

maniac103:
Add initial version of a binding for vacuum cleaners made by Ecovacs. It should work for all devices made in the last 3 years, but 
only a limited subset of those have been tested actively (the README spells out the respective devices).

Release for testing in the marketplace: [here](https://community.openhab.org/t/ecovacs-vacuum-cleaners-binding/132989)


maniac103:
@Flowdalic l would appreciate some help regarding the integration of Smack here. My problem is the following:
- I only need smack-core for this binding, as I don't need any actual chat features, and Ecovacs' XMPP server doesn't support them anyway
- The XMPP client binding, on the other side, *does* need IM features
- Since smack-core is loaded as OSGi bundle, there's only one instance of both code and static variables for both
- Smack's initializer code will try to load any relevant classes in the classpath, which will automagically enable any feature in the class path for all connections by a callback system in static variables
- This means that (at least) when the XMPP binding is loaded, the IM features (e.g. roster) are active for my connections as well
- Ecovacs' server does not support roster, so each connection attempt yields error messages and stack traces in the log
- Unfortunately, the XMPP connection to the vacuum seems a bit unstable, so I have to reconnect rather often, which leads to a lot of clutter in the log

I now need some suggestions how to proceed from here. Could I somehow disable the extended features *just for a single connection* (I tried, but didn't find a way)? Can the error message on roster reload failure be toned down in case the server doesn't support it? (I would have submitted a PR for that, but I seemingly can't build Smack on my machine; probably due to wrong Gradle version?) Is there another way to split the use cases of the XMPP and Ecovacs bindings?
I'd appreciate any pointers. Thanks!

maniac103:
The `WIP` tag is only for the XMPP issue described above, everything else in the binding should be ready for review.

Flowdalic:
> * This means that (at least) when the XMPP binding is loaded, the IM features (e.g. roster) are active for my connections as well

Have you tried `Roster.setLoadedOnLogin(false)`?

> Can the error message on roster reload failure be toned down in case the server doesn't support it?

IIRC the XMPP roster is a non discoverable feature. But the IQ request that yield an IQ error response that indicates that the service does not support the roster, which we may could use to change the logging behavior.

> I would have submitted a PR for that, but I seemingly can't build Smack on my machine; probably due to wrong Gradle version?

Smack currently needs Gradle 6.8.3 to build.

> Is there another way to split the use cases of the XMPP and Ecovacs bindings?

Ideally, and if I understood your issue(s) correctly, we wouldn't need that.

bundles/org.openhab.binding.ecovacs/pom.xml
```diff
@@ -0,0 +1,45 @@
...
+      <scope>provided</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.igniterealtime.smack</groupId>
+      <artifactId>smack-resolver-javax</artifactId>
+      <version>${smack.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>org.minidns</groupId>
+      <artifactId>minidns-core</artifactId>
```
Flowdalic, line 40: I could imagine that you potentially do not need MiniDNS to reach the Ecovacs XMPP endpoint.

bundles/org.openhab.binding.ecovacs/src/main/java/org/openhab/binding/ecovacs/internal/api/impl/EcovacsXmppDevice.java
```diff
@@ -0,0 +1,463 @@
...
+
+        private String createRequestId() {
+            // Ecovacs' app uses numbers for request IDs, so better constrain ourselves to that as well
+            int random8DigitNumber = 10000000 + new Random().nextInt(90000000);
+            return Integer.toString(random8DigitNumber);
+        }
+    }
+
+    private static class PingIQ extends SimpleIQ {
+        public PingIQ(Jid from, Jid to) {
```
Flowdalic, line 421: Why not use org.jivesoftware.smackx.ping.packet.Ping?

bundles/org.openhab.binding.ecovacs/src/main/java/org/openhab/binding/ecovacs/internal/api/impl/EcovacsXmppDevice.java
```diff
@@ -0,0 +1,463 @@
...
+        }
+    }
+
+    private static class CommandIQProvider extends IQProvider<@Nullable DeviceCommandIQ> {
+        @Override
+        public @Nullable DeviceCommandIQ parse(@Nullable XmlPullParser parser, int initialDepth) throws Exception {
+            @Nullable
+            DeviceCommandIQ packet = null;
+
+            if (parser == null) {
```
Flowdalic, line 435: Null checking the parser should be unnecessary.

bundles/org.openhab.binding.ecovacs/src/main/java/org/openhab/binding/ecovacs/internal/api/impl/EcovacsXmppDevice.java
```diff
@@ -0,0 +1,463 @@
...
+                CommandResponseHolder responseHolder = new CommandResponseHolder();
+                handler.registerPendingCommand(request.id, responseHolder);
+
+                logger.trace("{}: sending command {}, retry {}", getSerialNumber(),
+                        command.getName(ProtocolVersion.XML), retry);
+                synchronized (responseHolder) {
+                    conn.sendIqRequestAsync(request);
+                    responseHolder.wait(1500);
+                }
+                handler.unregisterPendingCommand(request.id);
```
Flowdalic, line 130: That smells like something you would want to do in a `finally` block to prevent leaks.

bundles/org.openhab.binding.ecovacs/src/main/java/org/openhab/binding/ecovacs/internal/api/impl/EcovacsXmppDevice.java
```diff
@@ -0,0 +1,463 @@
...
+                } catch (Exception e) {
+                    listener.onEventStreamFailure(EcovacsXmppDevice.this, e);
+                }
+            } else if (iqRequest instanceof ErrorIQ) {
+                StanzaError error = ((ErrorIQ) iqRequest).getError();
+                logger.trace("{}: Got error response {}", getSerialNumber(), error);
+                listener.onEventStreamFailure(EcovacsXmppDevice.this,
+                        new XMPPException.XMPPErrorException(iqRequest, error));
+            }
+            return null;
```
Flowdalic, line 369: Maybe I am missing something, or it is just not expected by the communication partner, but *not* sending an IQ response would be an violation of the XMPP.

bundles/org.openhab.binding.ecovacs/src/main/java/org/openhab/binding/ecovacs/internal/api/impl/EcovacsXmppDevice.java
```diff
@@ -0,0 +1,463 @@
...
+                            logger.debug("{}: Got unexpected XML payload {}", getSerialNumber(), iq.payload);
+                        }
+                    }
+                } catch (Exception e) {
+                    listener.onEventStreamFailure(EcovacsXmppDevice.this, e);
+                }
+            } else if (iqRequest instanceof ErrorIQ) {
+                StanzaError error = ((ErrorIQ) iqRequest).getError();
+                logger.trace("{}: Got error response {}", getSerialNumber(), error);
+                listener.onEventStreamFailure(EcovacsXmppDevice.this,
```
Flowdalic, line 366: Hmm, do you handle IQ responses in an IQ *request* handler? If so, then you may want to reconsider the implementation. Smack does provide high level API for sending IQ requests and gathering responses, including handling response timeouts.

Flowdalic (commented):
I had a quick look at your implementation and left some comments on things that looked strange to me. I hope you do not mind, especially since I may not fully understood what's done here.

bundles/org.openhab.binding.ecovacs/pom.xml
maniac103: Right, it's only a runtime dependency. Will remove this.

bundles/org.openhab.binding.ecovacs/src/main/java/org/openhab/binding/ecovacs/internal/api/impl/EcovacsXmppDevice.java
maniac103: Because smack-extensions depends on smack-im, which I didn't want to pull in due to the issues mentioned above. I couldn't use `PingManager`s timer because I want to ping the device (a particular JID) and not the server [1].
Therefore I thought duplicating that simple class should be fine.

[1] Ecovacs implements an XMPP proxy server that itself talks to the actual device, and while the connection to their server is stable, the connection between their server and the device looks like it isn't.

bundles/org.openhab.binding.ecovacs/src/main/java/org/openhab/binding/ecovacs/internal/api/impl/EcovacsXmppDevice.java
maniac103: I had to do that to keep the null checks happy. Since Smack's code does not have null annotations, it's assumed to be nullable, and the null checker then requires to check for null.

bundles/org.openhab.binding.ecovacs/src/main/java/org/openhab/binding/ecovacs/internal/api/impl/EcovacsXmppDevice.java
maniac103: Good point, will fix this.

bundles/org.openhab.binding.ecovacs/src/main/java/org/openhab/binding/ecovacs/internal/api/impl/EcovacsXmppDevice.java
maniac103: But this request handler actually is the response handler ;-) More on that below...

bundles/org.openhab.binding.ecovacs/src/main/java/org/openhab/binding/ecovacs/internal/api/impl/EcovacsXmppDevice.java
maniac103: I know, but I can't use it due to Ecovacs' braindead implementation. A sample request-response pair looks like this:
```
2022-02-03 07:43:51.737 [TRACE] [.internal.api.impl.EcovacsXmppDevice] - XXX: SENT (0): <iq to='XXX@115.ecorobot.net/atom' id='GWpLy-18' type='set'><query xmlns='com:ctl'><ctl id="55792281" name="FW" td="GetVersion"/></query></iq>
2022-02-03 07:43:51.812 [TRACE] [.internal.api.impl.EcovacsXmppDevice] - XXX: RECV (0): <iq to="XXX@ecouser.net/e7a4f07a" type="set" id="099825546960" from="XXX@115.ecorobot.net/atom"><query xmlns="com:ctl"><ctl id="55792281" ret="ok">
```
You can see that they do not use the IQ ID field the way it's intended and return a different ID (with whatever source/meaning). Instead, the client (here: the binding) is supposed to add an ID field to the payload, and that one will be echoed to be able to match request and response. I tried to use `AbstractXMPPConnection.sendIqRequestAndWaitForResponse()` first ;-)
Additionally their response is a `set` request instead of a `result`. Out-of-bounds messages are sent as `set` requests as well.

bundles/org.openhab.binding.ecovacs/src/main/java/org/openhab/binding/ecovacs/internal/api/impl/EcovacsXmppDevice.java
maniac103: Tried removing the `@Nullable` annotation for `parser` (which allows removing this check), but:

```
[ERROR] openhab-addons/bundles/org.openhab.binding.ecovacs/src/main/java/org/openhab/binding/ecovacs/internal/api/impl/EcovacsXmppDevice.java:[435,48] Illegal redefinition of parameter parser, inherited method from org.jivesoftware.smack.provider.Provider<org.openhab.binding.ecovacs.internal.api.impl.EcovacsXmppDevice.DeviceCommandIQ> does not constrain this parameter
```

... so I guess I'll have to keep it.

bundles/org.openhab.binding.ecovacs/src/main/java/org/openhab/binding/ecovacs/internal/api/impl/EcovacsXmppDevice.java
Flowdalic: A little piece of me just died.

Flowdalic:
> > Have you tried Roster.setLoadedOnLogin(false)?
> 
> I can not find that method in Smack's sources?

Sorry it is `Roster.setRosterLoadedAtLogin()`.

maniac103:
> Sorry it is Roster.setRosterLoadedAtLogin().

That one works and fixes the log spam, thanks!

bundles/org.openhab.binding.ecovacs/src/main/java/org/openhab/binding/ecovacs/internal/api/impl/EcovacsXmppDevice.java
maniac103: Since I now need `smack-im` for disabling the roster loading, I also switched to using `PingManager`.

bundles/org.openhab.binding.ecovacs/src/main/java/org/openhab/binding/ecovacs/internal/api/impl/EcovacsXmppDevice.java
maniac103: Heh, I had precisely the same thought when seeing the first log with that communication. Nothing we can change here I guess, though...

bundles/org.openhab.binding.ecovacs/src/main/java/org/openhab/binding/ecovacs/internal/handler/EcovacsVacuumHandler.java
```diff
@@ -0,0 +1,652 @@
...
+    }
+
+    private @Nullable String determineCommandChannelValue(boolean charging, CleanMode cleanMode) {
+        if (charging) {
+            return CMD_CHARGE;
+        }
+        switch (cleanMode) {
+            case AUTO:
+                return CMD_AUTO_CLEAN;
+            case PAUSE:
```
maniac103, line 570: Need to add spot area handling here.

bundles/org.openhab.binding.ecovacs/src/main/feature/feature.xml
```diff
@@ -0,0 +1,22 @@
...
+		<feature dependency="true">openhab.tp-hivemqclient</feature>
+		<bundle dependency="true">mvn:org.igniterealtime.smack/smack-tcp/4.3.3</bundle>
+		<bundle dependency="true">mvn:org.jxmpp/jxmpp-core/0.6.3</bundle>
+		<bundle dependency="true">mvn:org.jxmpp/jxmpp-jid/0.6.3</bundle>
+		<bundle dependency="true">mvn:org.jxmpp/jxmpp-util-cache/0.6.3</bundle>
+		<bundle dependency="true">mvn:org.minidns/minidns-core/0.3.3</bundle>
+		<bundle dependency="true">mvn:org.igniterealtime.smack/smack-core/4.3.3</bundle>
+		<bundle dependency="true">mvn:org.igniterealtime.smack/smack-im/4.3.3</bundle>
+		<bundle dependency="true">mvn:org.igniterealtime.smack/smack-extensions/4.3.3</bundle>
+		<bundle dependency="true">mvn:org.igniterealtime.smack/smack-sasl-javax/4.3.3</bundle>
```
maniac103, line 16: Despite listing this dependency, it looks like it is not installed automatically when installing a kar file built from these sources (see [here](https://community.openhab.org/t/ecovacs-vacuum-cleaners-binding/132989/18)). I'd appreciate any guidance from somebody more experienced with OSGi than me about what I'm missing here.
(@wborn or @cweitkamp perhaps? Sorry for the somewhat random tagging, I just selected you because I found you being involved in previous discussions about related topics. If I mistagged you, it would be great if you could hint me at who the right person to ask is ... thanks!)

maniac103:
@openhab/add-ons-maintainers May I ask for a review and/or help with the OSGi question, please? Thanks.

bundles/org.openhab.binding.ecovacs/src/main/feature/feature.xml
fwolter: At first sight, I see no reason why they wouldn't get downloaded automatically. What makes you say they aren't?

maniac103: The forum feedback suggests it isn't installed automatically:
- https://community.openhab.org/t/ecovacs-vacuum-cleaners-binding/132989/18 (this is the one I linked to above)
- https://community.openhab.org/t/ecovacs-vacuum-cleaners-binding/132989/33 (second user with the same symptoms)

bundles/org.openhab.binding.ecovacs/README.md
```diff
@@ -0,0 +1,126 @@
...
+## Channels
+
+The list below lists all channels supported by the binding. In case a particular channel is not supported by a given device (see remarks),
+it is automatically removed from the given thing.
+
+| Channel                          | Type                 | Description                                               | Read Only | Updated By | Remarks  |
+|----------------------------------|----------------------|-----------------------------------------------------------|-----------|------------|----------|
+| actions#command                  | String               | Command to execute                                        | No        | Event      | [1]      |
+| status#state                     | String               | Current operational state                                 | Yes       | Event      | [2]      |
+| status#current-cleaning-mode     | Number:Time          | Mode used in current cleaning run                         | Yes       | Event      | [3], [4] |
```
mueller-ma, line 64: Is this really a `Number:Time`?

maniac103: No, this was a copy'n'paste mistake. Thanks, fixed.

bundles/org.openhab.binding.ecovacs/pom.xml
```diff
@@ -0,0 +1,51 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns="http://maven.apache.org/POM/4.0.0"
+  xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd">
+
+  <modelVersion>4.0.0</modelVersion>
+
+  <parent>
+    <groupId>org.openhab.addons.bundles</groupId>
+    <artifactId>org.openhab.addons.reactor.bundles</artifactId>
+    <version>3.3.0-SNAPSHOT</version>
```
wborn, line 10:
```suggestion
    <version>3.4.0-SNAPSHOT</version>
```

maniac103: Done. I've also rebased the branch to `main`.

maniac103:
@openhab/add-ons-maintainers Any chance of getting this reviewed/merged for 3.4?
From my side it should be ready, only open question is the XMPP SASL jar seemingly not being auto-installed for all users.

lolodomo:
It is apparently no more "work in progress" so I remove the label.

bundles/org.openhab.binding.ecovacs/pom.xml
```diff
@@ -0,0 +1,51 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns="http://maven.apache.org/POM/4.0.0"
+  xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd">
+
+  <modelVersion>4.0.0</modelVersion>
+
+  <parent>
+    <groupId>org.openhab.addons.bundles</groupId>
+    <artifactId>org.openhab.addons.reactor.bundles</artifactId>
+    <version>3.4.0-SNAPSHOT</version>
```
wborn, line 10:
```suggestion
    <version>4.0.0-SNAPSHOT</version>
```

maniac103:
Any chance of getting review on this? It's sitting since almost 1 year now. Since the binding has been in Marketplace testing during all that time, I consider it stable/ready for review; I lately just added some more features. Only open question is the XMPP install thing mentioned [here](https://github.com/openhab/openhab-addons/pull/12231#discussion_r816542412)

bundles/org.openhab.binding.ecovacs/README.md
```diff
@@ -0,0 +1,162 @@
...
+- Deebot OZMO Slim 10/11
+- Deebot N8 series
+
+The following devices will likely work because they are using similar protocols as the above ones:
+
+- Deebot 600/601/605
+- Deebot OZMO 610
+- Deebot 710/711/711s
+- Deebot 900/901
+- Deebot OZMO 900/905
```
mueller-ma, line 25: I can confirm that the binding works with Deebot OZMO 900.

bundles/org.openhab.binding.ecovacs/src/main/resources/OH-INF/binding/binding.xml
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<binding:binding id="ecovacs" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+	xmlns:binding="https://openhab.org/schemas/binding/v1.0.0"
+	xsi:schemaLocation="https://openhab.org/schemas/binding/v1.0.0 https://openhab.org/schemas/binding-1.0.0.xsd">
+
+	<name>Ecovacs Binding</name>
+	<description>This is the binding for Ecovacs Deebot vacuum cleaners.</description>
+
+</binding:binding>
```
J-N-K, line 9: Please convert to the new `addon.xml`.

stefan-hoehn:
I can confirm that the binding works very well with the DEEBOT OZMO T8 AIVI for quite a number of months. Thanks @maniac103 for the contribution. I hope it will soon make it into 4.0

jlaur:
@maniac103 - removing "help wanted" as I assume you got the OSGi question sorted out?

bundles/org.openhab.binding.ecovacs/src/main/java/org/openhab/binding/ecovacs/internal/EcovacsBindingConstants.java
```diff
@@ -0,0 +1,158 @@
...
+
+        @SafeVarargs
+        public static <T extends Enum<T>> StateOptionMapping<T> of(StateOptionEntry<T>... entries) {
+            StateOptionMapping<T> map = new StateOptionMapping<>();
+            for (StateOptionEntry<T> entry : entries) {
+                map.put(entry.enumValue, entry);
+            }
+            return map;
+        }
+    }
```
jlaur, line 124: Consider moving to another file to not clutter this class intended for constants.

bundles/org.openhab.binding.ecovacs/src/main/resources/OH-INF/thing/thing-types.xml
```diff
@@ -0,0 +1,351 @@
...
+			</parameter>
+		</config-description>
+	</thing-type>
+
+	<channel-group-type id="status">
+		<label>Status</label>
+		<channels>
+			<channel id="state" typeId="state"/>
+			<channel id="battery" typeId="system.battery-level"/>
+			<channel id="current-cleaning-mode" typeId="current-cleaning-mode"/>
```
jlaur, line 43: The current naming convention for channel ids is camelCase:
https://www.openhab.org/docs/developer/guidelines.html#java-coding-style

I believe the same convention is preferred for channel type ids as well:
```suggestion
			<channel id="currentCleaningMode" typeId="currentCleaningMode"/>
```

this applies to all channel ids and channel type ids.

bundles/org.openhab.binding.ecovacs/src/main/resources/OH-INF/thing/thing-types.xml
```diff
@@ -0,0 +1,351 @@
...
+				<option value="medium">Medium</option>
+				<option value="high">High</option>
+				<option value="veryhigh">Very high</option>
+			</options>
+		</state>
+	</channel-type>
+
+	<channel-type id="wifi-rssi" advanced="true">
+		<item-type>Number:Power</item-type>
+		<label>Wi-Fi Signal Strength</label>
```
jlaur, line 298: Description missing. Consider also:
```suggestion
		<label>Wi-Fi Signal Strength</label>
		<category>QualityOfService</category>
```

bundles/org.openhab.binding.ecovacs/src/main/resources/OH-INF/thing/thing-types.xml
```diff
@@ -0,0 +1,351 @@
...
+				<option value="normal">Normal</option>
+				<option value="high">Maximum</option>
+				<option value="higher">Maximum+</option>
+			</options>
+		</state>
+	</channel-type>
+
+	<channel-type id="true-detect-3d" advanced="true">
+		<item-type>Switch</item-type>
+		<label>True Detect 3D</label>
```
jlaur, line 273: Description missing.

bundles/org.openhab.binding.ecovacs/src/main/resources/OH-INF/thing/thing-types.xml
```diff
@@ -0,0 +1,351 @@
...
+
+	<channel-type id="continuous-cleaning" advanced="true">
+		<item-type>Switch</item-type>
+		<label>Continuous Cleaning</label>
+		<description>Automatically resume unfinished cleaning after charging</description>
+	</channel-type>
+
+	<channel-type id="suction-power">
+		<item-type>String</item-type>
+		<label>Cleaning Power Level</label>
```
jlaur, line 260: Description missing.

bundles/org.openhab.binding.ecovacs/README.md
```diff
@@ -0,0 +1,163 @@
...
+
+## Thing Configuration
+
+For the cloud API thing, the following parameters must be configured:
+
+| Config    | Description                                                                                                                   |
+|-----------|-------------------------------------------------------------------------------------------------------------------------------|
+| email     | The email address you used when registering the Ecovacs cloud account                                                         |
+| password  | The cloud account password                                                                                                    |
+| continent | The continent you are residing on, or 'World' if none matches. This is used to select the correct cloud server to connect to. |
```
jlaur, line 47: Can the four possible values be mentioned in order to ease textual configuration?

bundles/org.openhab.binding.ecovacs/src/main/resources/OH-INF/addon/addon.xml
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<addon:addon id="ecovacs" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+	xmlns:addon="https://openhab.org/schemas/addon/v1.0.0"
+	xsi:schemaLocation="https://openhab.org/schemas/addon/v1.0.0 https://openhab.org/schemas/addon-1.0.0.xsd">
+
+	<type>binding</type>
+	<name>Ecovacs Binding</name>
+	<description>This is the binding for Ecovacs Deebot vacuum cleaners.</description>
```
jlaur, line 8: This could also be added:
```suggestion
	<description>This is the binding for Ecovacs Deebot vacuum cleaners.</description>
	<connection>cloud</connection>
```

bundles/org.openhab.binding.ecovacs/src/main/java/org/openhab/binding/ecovacs/internal/handler/EcovacsApiHandler.java
```diff
@@ -0,0 +1,173 @@
...
+    }
+
+    private EcovacsApi createApi(String deviceIdSuffix, String country) {
+        EcovacsApiConfiguration config = getConfigAs(EcovacsApiConfiguration.class);
+        String deviceId = config.installId + deviceIdSuffix;
+        org.openhab.binding.ecovacs.internal.api.EcovacsApiConfiguration apiConfig = new org.openhab.binding.ecovacs.internal.api.EcovacsApiConfiguration(
+                deviceId, config.email, config.password, config.continent, country, "EN", ClientKeys.CLIENT_KEY,
+                ClientKeys.CLIENT_SECRET, ClientKeys.AUTH_CLIENT_KEY, ClientKeys.AUTH_CLIENT_SECRET);
+
+        return EcovacsApi.create(httpClientFactory.getCommonHttpClient(), apiConfig);
```
jlaur, line 143: If only common client is used, you could consider refactoring the constructor to simply take `HttpClient` rather than `HttpClientFactory`.

bundles/org.openhab.binding.ecovacs/src/main/java/org/openhab/binding/ecovacs/internal/config/EcovacsApiConfiguration.java
```diff
@@ -0,0 +1,28 @@
...
+/**
+ * The {@link EcovacsApiConfiguration} class contains fields mapping thing configuration parameters.
+ *
+ * @author Danny Baumann - Initial contribution
+ */
+@NonNullByDefault
+public class EcovacsApiConfiguration {
+    public String email = "";
+    public String password = "";
+    public String continent = "";
```
jlaur, line 26:
```suggestion
    public String continent = "ww";
```

bundles/org.openhab.binding.ecovacs/src/main/resources/OH-INF/thing/bridge.xml
```diff
@@ -0,0 +1,39 @@
...
+				<label>Password</label>
+				<context>password</context>
+				<description>Password for logging in to Ecovacs server</description>
+				<required>true</required>
+			</parameter>
+			<parameter name="continent" type="text">
+				<label>Continent</label>
+				<description>Continent the account was registered on. Choose the one you are located in, or "World" if none matches.</description>
+				<default>ww</default>
+				<required>true</required>
```
jlaur, line 29: Not required since it has default:
```suggestion
```

jlaur (changes requested):
Thanks for your contribution and sorry for the long delay. I have started a review, and the code already looks quite good. I'm half-way through and only found a few things to comment so far.

jlaur:
> > removing "help wanted" as I assume you got the OSGi question sorted out?
> 
> @jlaur Unfortunately not, that's why the thread above is still open. Whenever people stumble upon this, I tell them to manually install the mentioned bundle (`smack-sasl-javax`) or to alternatively install the xmppclient binding.

Okay, I have re-added the label. Perhaps the PR should be marked as draft to prevent it from being merged while this issue persists?

maniac103:
> Perhaps the PR should be marked as draft to prevent it from being merged while this issue persists?

From my side it *is* ready though (that is, I do not have any additional improvement ideas at this point). The open thread should block it from being merged ... but I would really like insights on what I'm doing wrong there.

bundles/org.openhab.binding.ecovacs/src/main/java/org/openhab/binding/ecovacs/internal/api/EcovacsApiConfiguration.java
```diff
@@ -0,0 +1,140 @@
...
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package org.openhab.binding.ecovacs.internal.api;
+
+import org.eclipse.jdt.annotation.NonNullByDefault;
+import org.openhab.binding.ecovacs.internal.api.util.MD5Util;
+
+/**
+ * @author Johannes Ptaszyk - Initial contribution
```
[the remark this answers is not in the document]
maniac103: Yes. We (as in: Johannes and me) started with [a PoC application](https://github.com/JohannesPtaszyk/EcovacsDeebotJava) which I then later converted to a binding. I kept the original author tag where I copied code from the PoC more or less unchanged.

bundles/org.openhab.binding.ecovacs/src/main/java/org/openhab/binding/ecovacs/internal/api/ClientKeys.java
```diff
@@ -0,0 +1,33 @@
...
+ * @author Johannes Ptaszyk - Initial contribution
+ */
+@NonNullByDefault
+public class ClientKeys {
+
+    private ClientKeys() {
+        // Prevent instantiation
+    }
+
+    // TODO Find out where to get these keys and secret from to provide further information for OpenHab users.
```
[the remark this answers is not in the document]
maniac103: Well, without client ID/secret communication is impossible :shrug: They're extracted from Ecovacs' Android app, and - unlike services like Netatmo - there's no portal for users to register an application for the API.
I've found at least two predecents in the openhab-addons repo for hardcoded client ID/secret (the evohome and myq bindings).

I think @JohannesPtaszyk originally intended to make those configurable in case Ecovacs at some point blocks those specific credentials, but given the hard way of obtaining them, I'm not sure whether that's useful?

I'm personally inclined to remove this class and hardcode them in `EcovacsApiConfiguration`.

jlaur:
> From my side it _is_ ready though (that is, I do not have any additional improvement ideas at this point). The open thread should block it from being merged ... but I would really appreciate any insights on what I'm doing wrong there.

Unfortunately I'm not able to help with that. Open comments do not prevent a merge and can easily be missed as the comment history grows. The most effective way to block it from being merged is by marking it as draft, but otherwise I would suggest to add the "Work in progress" label.

maniac103:
> The most effective way to block it from being merged is by marking it as draft, but otherwise I would suggest to add the "Work in progress" label.

Won't either ultimately drive reviewers away from reviewing this? `The author isn't done yet, so it's pointless to review at this point.`? The "Help wanted" label indeed seems like the most appropriate to me.

bundles/org.openhab.binding.ecovacs/src/main/java/org/openhab/binding/ecovacs/internal/api/ClientKeys.java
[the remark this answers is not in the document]
jlaur: I know they are needed, so it's kind of a hard question. I have just been in contact with Velux about the same issue - to get their opinion, from a legal perspective, how they will would react if a public integration would be made based on reverse engineering of their app and using the key obtained from the app. They highly recommended not doing that, but didn't directly threaten me with legal consequences, so I have sent some additional questions about the matter.

If you want the keys to be configurable it would make sense to do what you suggested, and otherwise I guess they could be moved to `EcovacsBindingConstants`. If they would change the keys, it would probably be to fight integrations, in which case they will probably also try to obscure other things at the same time.

jlaur:
> Won't either ultimately drive reviewers away from reviewing this? `The author isn't done yet, so it's pointless to review at this point.`?

Potentially yes, but with some sense, since from a priority perspective, time is better spent reviewing PR's that can actually be merged soon. However, in this case I already started a review, and plan to continue although not entirely ready.

> The "Help wanted" label indeed seems like the most appropriate to me.

Indeed that label is appropriate.

maniac103:
> since from a priority perspective, time is better spent reviewing PR's that can actually be merged soon.

Just a remark regarding that prioritization from the perspective of a first time contributor: that mentioned approach, while I completely understand the rationale, potentially blocks contributions with minor/few, but hard to resolve, issues ad infinitum. As I have no experience at all with OSGi, I have no idea on how to resolve this issue (assuming there actually is an issue in my code), so there's absolutely no chance of ever getting rid of the 'not finished yet' label without external help. It would be great if there would be a way to flag 'I need help from experts/maintainers with a specific question'. (One unfortunately can't mention a team without being member of the OH org - which in this particular case I am only since a few days for being one of the Android app maintainers)

> However, in this case I already started a review, and plan to continue although not entirely ready.

Thanks :+1:

bundles/org.openhab.binding.ecovacs/src/main/feature/feature.xml
J-N-K: I guess the issue is that classes from this bundle are loaded dynamically, with no import statement in the manifest of the loading bundle. `dependency="true"` essentially means "install if it is required by some other bundle" and this requirement is determined by the other bundle's import statements. You can try do remove the dependency attribute (like the other two bundles at the end of the feature).

bundles/org.openhab.binding.ecovacs/src/main/java/org/openhab/binding/ecovacs/internal/api/ClientKeys.java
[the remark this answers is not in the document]
maniac103: Moved them to `EcovacsBindingConstants`. Other integration libraries [are doing the same thing](https://github.com/mrbungle64/ecovacs-deebot.js/blob/master/library/constants.js#L23-L42) ... I doubt Ecovacs really cares, TBH.

bundles/org.openhab.binding.ecovacs/src/main/feature/feature.xml
maniac103: Thanks, I'll give that a try :+1: 
I assume I should do the same to the other smack bundles as well them, right?
(FWIW, I copied those lines from the [xmppclient binding](https://github.com/openhab/openhab-addons/blob/main/bundles/org.openhab.binding.xmppclient/src/main/feature/feature.xml#L8-L23) - I assume it has the same basic issue, but it went unnoticed because the servers people use that binding with don't use SASL?)

J-N-K: No, just for the bundles that are not installed automatically. This is a work-around to force installation, in general it is better to leave the dependency installation to the framework. Maybe the binding itself uses a class from that bundle.

jlaur:
> > since from a priority perspective, time is better spent reviewing PR's that can actually be merged soon.
> 
> Just a remark regarding that prioritization from the perspective of a first time contributor: that mentioned approach, while I completely understand the rationale, potentially blocks contributions with minor/few, but hard to resolve, issues ad infinitum. As I have no experience at all with OSGi, I have no idea on how to resolve this issue (assuming there actually is an issue in my code), so there's absolutely no chance of ever getting rid of the 'not finished yet' label without external help. It would be great if there would be a way to flag 'I need help from experts/maintainers with a specific question'. (One unfortunately can't mention a team without being member of the OH org - which in this particular case I am only since a few days for being one of the Android app maintainers)

I fully understand this frustration, but I think we are discussing two different things. You need:
- Help from anyone able to help (e.g. other contributors).
- Review from maintainers.

Perhaps help could be requested from the community forum in the **Development** category? I see @J-N-K just gave some advise, so suggestion is general.

For the review part, labels and PR status are means for maintainers to ease decision making, e.g. for filtering - and yes, possibly prioritizing ready PR's higher than WIP, because of current capacity issues.

Draft status will effectively prevent the merge button from being accessible, while a lot of text conversations can be harder to interpret and ultimately could lead to a wrong decision being made. It should not happen, but we are only human. I just wanted to add this reasoning for my suggestion, although in this case I think other maintainers will now see these new comments and decide not to merge just yet. 😉

bundles/org.openhab.binding.ecovacs/src/main/resources/OH-INF/thing/thing-types.xml
maniac103: Is this a hard requirement? Grepping through the repo, I see all kinds of styles (`alllowercase`, `camelCase`, `using-hyphens`, `snake_case`), some use one style for channel IDs and another for channel type IDs (bluetooth.am43, hdpowerview). This also applies to rather new bindings like mercedesme, which was merged last year and whose code is actually *newer* than this PR (#13044).

The reason for my question is that I think it's a good idea to keep the current style otherwise, to make sure the people who started using the binding from the marketplace during the last year don't have to reassign all of their items, which is (when done via UI) a rather tedious experience.

bundles/org.openhab.binding.ecovacs/src/main/resources/OH-INF/thing/thing-types.xml
maniac103: :+1: for category, but is a description mandatory? Thinking of how a description could be worded that is not just a copy of the label.

bundles/org.openhab.binding.ecovacs/src/main/resources/OH-INF/thing/thing-types.xml
jlaur: Not enforcing this rule exactly leads to the problem you mention where we end up with a mix of different naming conventions.

I don't know when this camelCase convention was introduced, but bindings created before then still suffers from this problem because channels cannot be renamed without breaking, and consistency within the binding takes precedence over cross-binding consistency. Meaning that if an old binding has all channels in snake_case, then new channels should keep this convention as well.

Today I would guess that camelCase is the mostly adopted convention with using-hyphens as a close second position. Probably due to the fact that this convention is used for system channel types.

You caught me with #13044. I don't remember if I missed it or if there was some argument for doing that.

You do have a valid argument with the existing community userbase, but on the other hand **only** those users will be affected, whereas **all** users would be affected if the renaming would happen after merging into the official distribution. So it's a bad moment, but there will never be a better moment. All new bindings could in reality be pre-released for the community without following the naming convention, and this would then have to be preserved because of existing users. This way it would be impossible to ever reach a common convention.

Back to your question, is it a "hard" requirement? For sure it's a documented official requirement, but I guess even maintainers can be soft. 😉 I would like to hear other opinions though from @openhab/add-ons-maintainers.

bundles/org.openhab.binding.ecovacs/src/main/java/org/openhab/binding/ecovacs/internal/handler/EcovacsApiHandler.java
```diff
@@ -0,0 +1,174 @@
...
+
+        return EcovacsApi.create(httpClient, apiConfig);
+    }
+
+    private synchronized void loginToApi(final EcovacsApi api) {
+        Future<?> loginFuture = this.loginFuture;
+        if (loginFuture != null && !loginFuture.isDone()) {
+            return;
+        }
+        loginFuture = scheduler.submit(() -> {
```
jlaur, line 152: Before scheduling this task, status should be set to **UNKNOWN**, so that status is set when leaving `initialize()`.

bundles/org.openhab.binding.ecovacs/src/main/java/org/openhab/binding/ecovacs/internal/handler/EcovacsApiHandler.java
```diff
@@ -0,0 +1,174 @@
...
+                    discoveryService.startScan();
+                }
+            } catch (InterruptedException e) {
+                Thread.currentThread().interrupt();
+                updateStatus(ThingStatus.OFFLINE);
+                this.api = null;
+            } catch (EcovacsApiException e) {
+                logger.debug("Ecovacs API login failed", e);
+                this.api = null;
+                updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.COMMUNICATION_ERROR, e.getMessage());
```
jlaur, line 170: Can this happen because of network error, for instance? Will the binding be able to recover from this, i.e. automatically retry?

jlaur (changes requested):
Next round. A few classes left, and a few to look a bit closer into.

bundles/org.openhab.binding.ecovacs/src/main/java/org/openhab/binding/ecovacs/internal/api/impl/JsonReportParser.java
```diff
@@ -0,0 +1,159 @@
...
+                break;
+            }
+            case "lifespan": {
+                // ComponentLifeSpanReport report = payloadAs(response, ComponentLifeSpanReport.class);
+                break;
+            }
+            case "speed": {
+                // SpeedReport report = payloadAs(response, SpeedReport.class);
+                // SuctionPower power = SuctionPower.fromJsonValue(report.speedLevel);
+                // TODO: report change
```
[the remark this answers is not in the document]
maniac103: Removed.

bundles/org.openhab.binding.ecovacs/src/main/java/org/openhab/binding/ecovacs/internal/api/impl/EcovacsIotMqDevice.java
```diff
@@ -0,0 +1,212 @@
...
+        if (loginData == null) {
+            throw new EcovacsApiException("Can not connect when not logged in");
+        }
+
+        // XML message handler does not receive firmware version information with events, so fetch in advance
+        if (desc.protoVersion == ProtocolVersion.XML) {
+            listener.onFirmwareVersionChanged(this, sendCommand(new GetFirmwareVersionCommand()));
+        }
+
+        // TOOD: use realm from config
```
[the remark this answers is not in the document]
maniac103: Actually implemented now, but probably an academic exercise at this point.

maniac103:
> maniac103 force-pushed the ecovacs-rebase branch from 59a58e1 to 30d9487

I unfortunately had to rebase the PR branch to `main` as otherwise the binding didn't build for me:
```
[ERROR] Failed to execute goal org.apache.karaf.tooling:karaf-maven-plugin:4.3.7:verify (karaf-feature-verification) on project org.openhab.binding.ecovacs: Feature resolution failed for [openhab-binding-ecovacs/4.0.0.SNAPSHOT]
[ERROR] Message: Unable to resolve root: missing requirement [root] osgi.identity; osgi.identity=openhab-binding-ecovacs; type=karaf.feature; version=4.0.0.SNAPSHOT; filter:="(&(osgi.identity=openhab-binding-ecovacs)(type=karaf.feature)(version>=4.0.0.SNAPSHOT))" [caused by: Unable to resolve openhab-binding-ecovacs/4.0.0.SNAPSHOT: missing requirement [openhab-binding-ecovacs/4.0.0.SNAPSHOT] osgi.identity; osgi.identity=openhab-runtime-base; type=karaf.feature [caused by: Unable to resolve openhab-runtime-base/4.0.0.SNAPSHOT: missing requirement [openhab-runtime-base/4.0.0.SNAPSHOT] osgi.identity; osgi.identity=openhab-core-base; type=karaf.feature [caused by: Unable to resolve openhab-core-base/4.0.0.SNAPSHOT: missing requirement [openhab-core-base/4.0.0.SNAPSHOT] osgi.identity; osgi.identity=org.openhab.core; type=osgi.bundle; version="[4.0.0.202303090301,4.0.0.202303090301]"; resolution:=mandatory [caused by: Unable to resolve org.openhab.core/4.0.0.202303090301: missing requirement [org.openhab.core/4.0.0.202303090301] osgi.wiring.package; filter:="(&(osgi.wiring.package=org.osgi.framework)(version>=1.10.0)(!(version>=2.0.0)))"]]]]
```

It contains only new changes from the `main` branch, I did not do any changes to the binding code during the rebase.

bundles/org.openhab.binding.ecovacs/src/main/java/org/openhab/binding/ecovacs/internal/handler/EcovacsApiHandler.java
maniac103: Added retry. I'm not entirely sure yet it's a good idea though: One possible occasion of failure is wrong username/password. This is mentioned in the detail status message in that case, but with a retry mechanism, if left unattended, it could lead to the account being (temporarily) blocked from further login attempts. I unfortunately can't really distinguish login failures from other failures because I don't get a clear indication back (such as HTTP 401), but just a JSON object that looks like this:

```json
"{"code":"10101","msg":"Wrong password. Unable to log in. Please try again after 15 minutes.","time":1678369207194,"data":null,"success":false}"
```

The possible codes are of course undocumented :-/

It's probably not a big problem in practice, but what are your thoughts on this? Retry in any case? Retry on timeouts only (but temporary API glitches might cause permanent failure then as well)?

bundles/org.openhab.binding.ecovacs/src/main/feature/feature.xml
maniac103: That worked, thanks :+1:

bundles/org.openhab.binding.ecovacs/src/main/java/org/openhab/binding/ecovacs/internal/api/impl/EcovacsXmppDevice.java
```diff
@@ -0,0 +1,468 @@
...
+            this.pingHandler = pingHandler;
+
+            conn.login();
+            conn.setReplyTimeout(1000);
+
+            logger.trace("{}: XMPP connection established", getSerialNumber());
+
+            listener.onFirmwareVersionChanged(this, sendCommand(new GetFirmwareVersionCommand()));
+            pingHandler.start();
+        } catch (EcovacsApiException e) {
```
jlaur, line 223: Only a small remark for consideration: I didn't check if you actually make use of the fact that `EcovacsApiException` is derived from `IOException`. In this case it works against you, because you need this additional catch/throw to avoid wrapping it into `EcovacsApiException` when already this type. If you mostly benefit from handling `EcovacsApiException` as `IOException`, this is fine, but otherwise you could consider making `EcovacsApiException` derive straight from `Exception` and remove this rethrow.

bundles/org.openhab.binding.ecovacs/src/main/resources/OH-INF/thing/thing-types.xml
jlaur: Gentle ping @openhab/add-ons-maintainers

bundles/org.openhab.binding.ecovacs/src/main/java/org/openhab/binding/ecovacs/internal/api/impl/EcovacsXmppDevice.java
maniac103: The reason for deriving from IOException is that I wanted EcovacsApiException to be a checked exception, so that I don't / can't forget to handle it.

bundles/org.openhab.binding.ecovacs/src/main/java/org/openhab/binding/ecovacs/internal/handler/EcovacsVacuumHandler.java
```diff
@@ -0,0 +1,830 @@
...
+            logger.debug("{}: Handling device command {} failed", getDeviceSerial(), command, e);
+        }
+    }
+
+    @Override
+    public void initialize() {
+        final String serial = getConfigAs(EcovacsVacuumConfiguration.class).serialNumber;
+        if (serial.isEmpty()) {
+            logger.info("Thing {} is missing serial number information", getThing().getUID());
+            updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.CONFIGURATION_ERROR);
```
jlaur, line 224:
```suggestion
            updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.CONFIGURATION_ERROR, "@text/offline-conf-error-missing-serial-number");
```

(where `offline-conf-error-missing-serial-number` refers to an I18N string)

bundles/org.openhab.binding.ecovacs/src/main/resources/OH-INF/i18n/ecovacs.properties
```diff
@@ -0,0 +1,173 @@
+# binding
+binding.ecovacs.name = Ecovacs Binding
+binding.ecovacs.description = This is the binding for Ecovacs Deebot vacuum cleaners.
```
jlaur, line 3: This needs to be regenerated after migrating to addon.xml.

bundles/org.openhab.binding.ecovacs/src/main/java/org/openhab/binding/ecovacs/internal/handler/EcovacsVacuumHandler.java
```diff
@@ -0,0 +1,830 @@
...
+                removeUnsupportedChannels(device);
+                connectToDevice();
+            } else {
+                logger.info("{}: Device not found in device list, setting offline", serial);
+                updateStatus(ThingStatus.OFFLINE);
+            }
+        } catch (InterruptedException e) {
+            Thread.currentThread().interrupt();
+        } catch (ConfigurationException e) {
+            updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.CONFIGURATION_ERROR);
```
jlaur, line 535: Does `ConfigurationException` contain a reason that could be used for the status update?

bundles/org.openhab.binding.ecovacs/src/main/java/org/openhab/binding/ecovacs/internal/handler/EcovacsVacuumHandler.java
```diff
@@ -0,0 +1,830 @@
...
+        } catch (EcovacsApiException e) {
+            updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.COMMUNICATION_ERROR);
```
jlaur, line 537: Ditto:
```suggestion
            updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.COMMUNICATION_ERROR, e.getMessage());
```

?

bundles/org.openhab.binding.ecovacs/src/main/java/org/openhab/binding/ecovacs/internal/handler/EcovacsVacuumHandler.java
```diff
@@ -0,0 +1,830 @@
...
+        if (device == null) {
+            return;
+        }
+        try {
+            action.run(device);
+        } catch (InterruptedException e) {
+            Thread.currentThread().interrupt();
+        } catch (EcovacsApiException e) {
+            logger.debug("{}: Failed communicating to device, reconnecting", getDeviceSerial(), e);
+            updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.COMMUNICATION_ERROR);
```
jlaur, line 806: I would suggest to add a thing status detail here.

bundles/org.openhab.binding.ecovacs/src/main/java/org/openhab/binding/ecovacs/internal/handler/EcovacsVacuumHandler.java
```diff
@@ -0,0 +1,830 @@
...
+                device.disconnect(scheduler);
+                this.device = null;
+            }
+            teardownAndScheduleReconnection();
+        }
+    }
+
+    private String getDeviceSerial() {
+        final String serialProp = getThing().getProperties().getOrDefault(Thing.PROPERTY_SERIAL_NUMBER, "");
+        return serialProp.isEmpty() ? getConfigAs(EcovacsVacuumConfiguration.class).serialNumber : serialProp;
```
jlaur, line 823: Serial number seems required in configuration, is there any reason to look in properties? In which case will there be a valid serial number property, but not any configured one for the thing?

jlaur (changes requested):
Almost done, a few minor additional comments.

bundles/org.openhab.binding.ecovacs/src/main/java/org/openhab/binding/ecovacs/internal/handler/EcovacsVacuumHandler.java
maniac103: The thought was that on automatic discovery, properties are filled from the `DiscoveryResult`, but configuration might not be. After re-reading the docs, one can read it as if configuration is always pre-filled from the `DiscoveryResult`. Is this guaranteed to be the case?

bundles/org.openhab.binding.ecovacs/src/main/java/org/openhab/binding/ecovacs/internal/handler/EcovacsVacuumHandler.java
maniac103: Do you mean a description? (A thing status detail is already present)

bundles/org.openhab.binding.ecovacs/src/main/resources/OH-INF/i18n/ecovacs.properties
maniac103: Done ... strangely enough, `mvn i18n:generate-default-translations` still adds the 'binding...' texts, though.

bundles/org.openhab.binding.ecovacs/src/main/java/org/openhab/binding/ecovacs/internal/handler/EcovacsVacuumHandler.java
maniac103: Added the description.

bundles/org.openhab.binding.ecovacs/src/main/resources/OH-INF/i18n/ecovacs.properties
jlaur: Do you still have `binding.xml` locally?

jlaur: Do you mean it adds them or that it doesn't remove them?

maniac103: > Do you still have binding.xml locally?

Nope:

```
find src/main/resources/OH-INF
src/main/resources/OH-INF
src/main/resources/OH-INF/i18n
src/main/resources/OH-INF/i18n/ecovacs.properties
src/main/resources/OH-INF/thing
src/main/resources/OH-INF/thing/bridge.xml
src/main/resources/OH-INF/thing/thing-types.xml
src/main/resources/OH-INF/addon
src/main/resources/OH-INF/addon/addon.xml
```

> Do you mean it adds them or that it doesn't remove them?

The former: I first manually converted lines 2 and 3 to `addon.ecovacs...`. After running `mvn i18n:generate-default-translations`, I suddenly had both.

maniac103: I now manually deleted them, and the tool keeps them deleted. Somewhat strange, but whatever...

bundles/org.openhab.binding.ecovacs/src/main/java/org/openhab/binding/ecovacs/internal/api/impl/EcovacsXmppDevice.java
maniac103: Thanks, I wasn't aware of that :+1: Rethrow removed.

bundles/org.openhab.binding.ecovacs/src/main/java/org/openhab/binding/ecovacs/internal/handler/EcovacsVacuumHandler.java
jlaur: > After re-reading the docs, one can read it as if configuration is always pre-filled from the `DiscoveryResult`. Is this guaranteed to be the case?

Correct, properties set by discovery will populate initial configuration, so you shouldn't need to use the properties in your logic.

> - for discovered things, what's used is always the discovered serial number
> - for manually added things, what's used is always the configured serial number (as no serial property is set in that case)

Third option: Manually configured things in files: no serial number.

Does it matter that serial number might be included in the Thing UID? The same is true for other bindings, for example the Miele binding:

https://github.com/openhab/openhab-addons/blob/dae33c405ee482d1cb226dfe4de37a57ba775ecf/bundles/org.openhab.binding.miele/src/main/java/org/openhab/binding/miele/internal/discovery/MieleApplianceDiscoveryService.java#L155

And the configuration is also populated with appliance id (which is serial number):

https://github.com/openhab/openhab-addons/blob/dae33c405ee482d1cb226dfe4de37a57ba775ecf/bundles/org.openhab.binding.miele/src/main/java/org/openhab/binding/miele/internal/discovery/MieleApplianceDiscoveryService.java#L113

And used:

https://github.com/openhab/openhab-addons/blob/dae33c405ee482d1cb226dfe4de37a57ba775ecf/bundles/org.openhab.binding.miele/src/main/java/org/openhab/binding/miele/internal/handler/MieleApplianceHandler.java#L147

jlaur:
> Draft status will effectively prevent the merge button from being accessible, while a lot of text conversations can be harder to interpret and ultimately could lead to a wrong decision being made.

I lost track. I still see the "help wanted" label. Besides the comments from my side, is this PR still blocked because of dependency issues?

bundles/org.openhab.binding.ecovacs/src/main/java/org/openhab/binding/ecovacs/internal/handler/EcovacsApiHandler.java
jlaur: Can you detect this known code (10101) and avoid retries for 15 minutes in this case? The issue could either be resolved by changing the password to match the configuration, in which case the retry itself could bring the thing online. Or, perhaps more commonly, the user will adapt the configuration, and the thing will be initialized again when saving the configuration.

Other than that, in general I think retrying will make the binding more reliable. Of course the level of sophistication can be discussed, i.e. retry with exponential backoff and jitter might be a good choice, but could also be overkill. I didn't check your retry algorithm yet, but even simple linear retry is better than no retrying.

maniac103: > Can you detect this known code (10101) and avoid retries for 15 minutes in this case?

The 15 minutes aren't a fixed value, but rather a countdown (trying 3 minutes later will yield 'Please try again after 12 minutes.'. I really don't want to parse that countdown value out of the error message ;-)
I'm also a bit reluctant to hardcode that value here, since it's just the value I happened to get when trying this once. With the general quality of their API, I'm really unsure whether it's the only possible error code for such a case.

Further attempts in the wait period don't seem to prolong it though (and don't seem to lead to blocking permanently), so it's probably not a big problem (if at all) in practice. I guess most users will check initialization right after creating a new thing (here: API bridge), so they'll likely notice the password mismatch.

bundles/org.openhab.binding.ecovacs/src/main/java/org/openhab/binding/ecovacs/internal/handler/EcovacsVacuumHandler.java
```diff
@@ -0,0 +1,824 @@
...
+        } catch (EcovacsApiException e) {
+            logger.debug("{}: Handling device command {} failed", getDeviceSerial(), command, e);
+        }
+    }
+
+    @Override
+    public void initialize() {
+        logger.debug("{}: Initializing handler", getDeviceSerial());
+        updateStatus(ThingStatus.UNKNOWN);
+        initTask.submit();
```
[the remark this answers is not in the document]
maniac103: Good point I did not think about. :+1:  (I assumed marking the serial number as required in XML was sufficient to not get to that case) ... restored.

bundles/org.openhab.binding.ecovacs/src/main/java/org/openhab/binding/ecovacs/internal/handler/EcovacsVacuumHandler.java
maniac103: > Does it matter that serial number might be included in the Thing UID?

That's the question ;-) I guess if a user is to do that edit, they hopefully know what they're doing. As I see the Miele binding isn't doing dynamic updating (as in: `handleConfigurationUpdate` handling) for the serial number either, I guess this should be OK in my case as well. I updated the handling accordingly.

maniac103:
AFAICT only open question that's remaining should be the [channel naming discussion](https://github.com/openhab/openhab-addons/pull/12231#discussion_r1128602473) now.

bundles/org.openhab.binding.ecovacs/src/main/java/org/openhab/binding/ecovacs/internal/handler/EcovacsVacuumHandler.java
```diff
@@ -0,0 +1,819 @@
...
+    private final String serialNumber;
+
+    public EcovacsVacuumHandler(Thing thing, TranslationProvider i18Provider, LocaleProvider localeProvider,
+            EcovacsDynamicStateDescriptionProvider stateDescriptionProvider) {
+        super(thing);
+        this.i18Provider = i18Provider;
+        this.localeProvider = localeProvider;
+        this.stateDescriptionProvider = stateDescriptionProvider;
+        bundle = FrameworkUtil.getBundle(getClass());
+        serialNumber = getConfigAs(EcovacsVacuumConfiguration.class).serialNumber;
```
jlaur, line 147: The correct place to initialize serialNumber is in `initialize()`. Otherwise it will not be picked up when changed. In the constructor you could even just set it to an empty string to satisfy null annotations.

jlaur (changes requested):
Another serialNumber round. 🙂

bundles/org.openhab.binding.ecovacs/src/main/java/org/openhab/binding/ecovacs/internal/handler/EcovacsVacuumHandler.java
jlaur: Exactly.

bundles/org.openhab.binding.ecovacs/README.md
```diff
@@ -0,0 +1,175 @@
...
+
+For the vacuum things, there is no required configuration (when using discovery). The following parameters exist:
+
+| Config       | Description                                                                                                                   |
+|--------------|-------------------------------------------------------------------------------------------------------------------------------|
+| serialNumber | Required: The device's serial number as printed on the barcode below the dust bin. Filled automatically when using discovery. |
+| refresh | Refresh interval for polled data (see below), in minutes. By default set to 5 minutes.                                             |
```
jlaur, line 55:
```suggestion
| refresh      | Refresh interval for polled data (see below), in minutes. By default set to 5 minutes.                                        |
```

maniac103: Thanks, stupid me :facepalm:

jlaur: Can we align it fully? 🙂

maniac103: Yes ... looking at https://github.com/openhab/openhab-addons/pull/12231/commits/ba11c9766bf44b72ca7caea71d49da6571484115 it's precisely what I did ;-)

jlaur: > Yes ... looking at https://github.com/openhab/openhab-addons/pull/12231/commits/ba11c9766bf44b72ca7caea71d49da6571484115 it's precisely what I did ;-)

Ah, missed that, sorry.

jlaur:
@maniac103 - there are some checkstyle warnings left. You could take a look at target/code-analysis/report.html.

bundles/org.openhab.binding.ecovacs/src/main/resources/OH-INF/thing/thing-types.xml
jlaur: @maniac103 - our naming convention seems not as graved in stone as I thought, and needs some further discussions and decisions. Your choice is consistent with the system channel ids and also by the convention used by many other bindings. Therefore we can merge as-is, so I'll resolve this comment.

bundles/org.openhab.binding.ecovacs/src/main/java/org/openhab/binding/ecovacs/internal/handler/EcovacsApiHandler.java
jlaur: Agreed.

maniac103:
> there are some checkstyle warnings left. You could take a look at target/code-analysis/report.html.

Fixed, thanks for the heads-up.

jlaur (approved):
Thanks for your contribution!

JohannesPtaszyk:
Happy to see how much you invested into this after the initial poc we built together 😍
Well done! Keep the great work and thank you so much for this contribution!

PR #12234, opened 2022-02-07
[porcupineks] fix file bundle extraction on Windows and null pointer exception 

GiviMAD:
Fixes file bundle extraction on Windows and null pointer exception


lolodomo (approved):
LGTM

PR #12238, opened 2022-02-07
[mqtt.generic] separate command parsing from cached value updating

### LINKED ISSUE #12150: [mqtt] REFRESH command doesn't restore item to current value in MQTT
I have an MQTT channel with separate command and state topics. The Item is set to autoupdate=true (the default). The device at the other end may take a long time updating the state topic with the new value after receiving a command (or not update at all, if it's rejected!)

## Expected Behavior
When the device rejects the command, I should be able to send a REFRESH to the item, and it will restore it to the current state in MQTT. But it does not.

## Current Behavior
It just stays forever, making OpenHAB out-of-sync with the device, with no way to fix it besides restarting the MQTT binding (or maybe just the thing?)

## Possible Solution
Make https://github.com/openhab/openhab-addons/blob/78e214651fa5ea715a65d8464df1108f1632290c/bundles/org.openhab.binding.mqtt.generic/src/main/java/org/openhab/binding/mqtt/generic/ChannelState.java#L334 only execute if you have a state topic and a command topic that aren't the same. If they're the same, that line makes sense to immediately update the cache, because the command you sent _is_ the new state of the item in MQTT. But if a device has separate topics, it might ignore the command, and the state should be able to be restored.

1. Set up an MQTT channel with differing (completely unrelated!) command and state topics. Have the state topic start with a retained value of say "77"
1. Link an item to the channel
1. Send a command, such as "78" to the channel
1. Note that in MQTT, the state topic still has a value of "77"
1. Send a REFRESH to the item.
1. It should now restore to 77, but it doesn't change.

## Context
I have a device that's very slow to acknowledge commands, and occasionally drops them, but I still want to use autoupdate to make my UI easy to interact with. I have written a rule so that when a command is sent to the item, after 30s it will send a REFRESH so that OpenHAB will show the actual state of the world. I would expect it to work, but it's not.

## Your Environment
* OpenHAB Version used: 3.2.0
* Operating System: `Linux 5.4.0-89-generic #100-Ubuntu SMP Fri Sep 24 14:50:10 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux` (Ubuntu)

ccutrer:
fixes #12150

Previously, Value.update would parse the command, _and_ update the cached
value with that command. Which means that when sending a command towards
MQTT (instead of processing an update from MQTT), the cached value was
unintentionally updated. This prevented the REFRESH command from returning
the most recent value received from the device.

Separating the two concerns also makes the test more obvious what they are
testing, and vastly simplified a kludgy workaround that RollershutterValue
was using to be able to process Commands that aren't States.

I've tested this in my local openhab installation to confirm that an MQTT channel with separate state and command topics, where the state topic does not change after sending a command, the item is set to autoupdate, I can now use REFRESH against the item and the prior state is restored.

### DIFF
--- a/bundles/org.openhab.binding.mqtt.generic/src/main/java/org/openhab/binding/mqtt/generic/ChannelState.java
+++ b/bundles/org.openhab.binding.mqtt.generic/src/main/java/org/openhab/binding/mqtt/generic/ChannelState.java
@@ -201,7 +201,7 @@ public class ChannelState implements MqttMessageSubscriber {
         Command parsedCommand;
         // Map the string to a command, update the cached value and post the command to the framework
         try {
-            parsedCommand = cachedValue.parseCommand(command);
+            parsedCommand = cachedValue.parseMessage(command);
         } catch (IllegalArgumentException | IllegalStateException e) {
             logger.warn("Command '{}' from channel '{}' not supported by type '{}': {}", strValue, channelUID,
                     cachedValue.getClass().getSimpleName(), e.getMessage());
--- a/bundles/org.openhab.binding.mqtt.generic/src/main/java/org/openhab/binding/mqtt/generic/ChannelState.java
+++ b/bundles/org.openhab.binding.mqtt.generic/src/main/java/org/openhab/binding/mqtt/generic/ChannelState.java
@@ -31,6 +31,7 @@ import org.openhab.core.io.transport.mqtt.MqttMessageSubscriber;
 import org.openhab.core.library.types.StringType;
 import org.openhab.core.thing.ChannelUID;
 import org.openhab.core.types.Command;
+import org.openhab.core.types.State;
 import org.openhab.core.types.TypeParser;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -197,22 +198,24 @@ public class ChannelState implements MqttMessageSubscriber {
             return;
         }
 
-        Command postOnlyCommand = cachedValue.isPostOnly(command);
-        if (postOnlyCommand != null) {
-            channelStateUpdateListener.postChannelCommand(channelUID, postOnlyCommand);
-            receivedOrTimeout();
-            return;
-        }
-
+        Command parsedCommand;
         // Map the string to a command, update the cached value and post the command to the framework
         try {
-            cachedValue.update(command);
+            parsedCommand = cachedValue.parseCommand(command);
         } catch (IllegalArgumentException | IllegalStateException e) {
             logger.warn("Command '{}' from channel '{}' not supported by type '{}': {}", strValue, channelUID,
                     cachedValue.getClass().getSimpleName(), e.getMessage());
             receivedOrTimeout();
             return;
         }
+
+        // things that are only Commands _must_ be posted as a command (like STOP)
+        if (!(parsedCommand instanceof State)) {
+            channelStateUpdateListener.postChannelCommand(channelUID, parsedCommand);
+            receivedOrTimeout();
+            return;
+        }
+        cachedValue.update((State) parsedCommand);
 
         if (config.postCommand) {
             channelStateUpdateListener.postChannelCommand(channelUID, (Command) cachedValue.getChannelState());
@@ -348,10 +351,6 @@ public class ChannelState implements MqttMessageSubscriber {
      *         and exceptionally otherwise.
      */
     public CompletableFuture<Boolean> publishValue(Command command) {
-        cachedValue.update(command);
-
-        Value mqttCommandValue = cachedValue;
-
         final MqttBrokerConnection connection = this.connection;
 
         if (connection == null) {
@@ -360,6 +359,9 @@ public class ChannelState implements MqttMessageSubscriber {
                     "The connection object has not been set. start() should have been called!"));
             return f;
         }
+
+        Command mqttCommandValue = cachedValue.parseCommand(command);
+        Value mqttFormatter = cachedValue;
 
         if (readOnly) {
             logger.debug(
@@ -370,12 +372,11 @@ public class ChannelState implements MqttMessageSubscriber {
 
         // Outgoing transformations
         for (ChannelStateTransformation t : transformationsOut) {
-            String commandString = mqttCommandValue.getMQTTpublishValue(null);
+            String commandString = mqttFormatter.getMQTTpublishValue(mqttCommandValue, null);
             String transformedValue = t.processValue(commandString);
             if (transformedValue != null) {
-                Value textValue = new TextValue();
-                textValue.update(new StringType(transformedValue));
-                mqttCommandValue = textValue;
+                mqttFormatter = new TextValue();
+                mqttCommandValue = new StringType(transformedValue);
             } else {
                 logger.debug("Transformation '{}' returned null on '{}', discarding message", mqttCommandValue,
                         t.serviceName);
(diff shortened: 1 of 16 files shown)


jochen314:
Hi, I really like where this is going :-)
I just have one suggestion: Currently the same `parseCommand` is used for messages received from MQTT and for values to be published to MQTT. Can we separate them? Could we have 2 different parse functions, where one calls the other by default?

I remember, that we do have some problems with the percentage value. During command parsing it is not clear, if the devices or OHs value range should be used. If we would have separate functions it would be clear.

ccutrer:
I totally agree on that point - I was super surprised when I realized _incoming_ values from MQTT went through the same parsing function. Sure, I can add a separate name and default to current implementation relatively easily. I think any change of implementation for particular data types probably belong in their own PR.

ccutrer:
@jochen314 : I've updated the PR with your suggestion. do you have any other feedback?

@lolodomo : can you look at this as well?

bundles/org.openhab.binding.mqtt.generic/src/main/java/org/openhab/binding/mqtt/generic/ChannelState.java
```diff
@@ -197,23 +199,25 @@ public void processMessage(String topic, byte[] payload) {
...
-            channelStateUpdateListener.postChannelCommand(channelUID, postOnlyCommand);
-            receivedOrTimeout();
-            return;
-        }
-
+        Type parsedCommand;
         // Map the string to a command, update the cached value and post the command to the framework
         try {
-            cachedValue.update(command);
+            parsedCommand = cachedValue.parseMessage(command);
```
jochen314, line 205: The variable name does not match any more. 
It is now a parsedMessage, right?

ccutrer: Fixed with latest commit

jochen314 (approved):
LGTM

ccutrer:
@lolodomo can this be merged now?

bundles/org.openhab.binding.mqtt.generic/src/main/java/org/openhab/binding/mqtt/generic/values/OpenCloseValue.java
```diff
@@ -54,28 +55,28 @@ public OpenCloseValue(@Nullable String openValue, @Nullable String closeValue) {
...
     @Override
-    public String getMQTTpublishValue(@Nullable String pattern) {
+    public String getMQTTpublishValue(Type type, @Nullable String pattern) {
         String formatPattern = pattern;
         if (formatPattern == null) {
             formatPattern = "%s";
         }
 
-        return String.format(formatPattern, state == OpenClosedType.OPEN ? openString : closeString);
+        return String.format(formatPattern, type == OpenClosedType.OPEN ? openString : closeString);
```
lolodomo, line 80: No need here to test if type is UNDEF?

bundles/org.openhab.binding.mqtt.generic/src/main/java/org/openhab/binding/mqtt/generic/values/RollershutterValue.java
```diff
@@ -59,73 +59,72 @@ public RollershutterValue(@Nullable String upString, @Nullable String downString
...
+    public String getMQTTpublishValue(Type type, @Nullable String pattern) {
         final String upString = this.upString;
         final String downString = this.downString;
-        if (this.nextIsStop) {
-            this.nextIsStop = false;
-            return stopString;
-        } else if (state instanceof PercentType) {
-            if (state.equals(PercentType.HUNDRED) && downString != null) {
+        final String stopString = this.stopString;
+        if (type == UpDownType.UP) {
```
lolodomo, line 103: No need here to test if type is UNDEF?

bundles/org.openhab.binding.mqtt.generic/src/main/java/org/openhab/binding/mqtt/generic/values/Value.java
```diff
@@ -94,11 +95,11 @@ public final State getChannelState() {
         return state;
     }
 
-    public String getMQTTpublishValue(@Nullable String pattern) {
+    public String getMQTTpublishValue(Type type, @Nullable String pattern) {
         if (pattern == null) {
-            return state.format("%s");
+            return type.format("%s");
```
lolodomo, line 100: No need here to test if type is UNDEF?

lolodomo (commented):
Just questions to be sure

lolodomo:
@ccutrer : any feedback on my comments ?

Note that there are now conflicts on few files.

ccutrer:
> @ccutrer : any feedback on my comments ?
> 
> 
> 
> Note that there are now conflicts on few files.

I think there's a good chance your comments need addressed, but I've been too busy to look at them. I'm
hoping to address them in the next couple days. And I'll also address any merge conflicts.

lolodomo:
@ccutrer ; any chance that you finished?

ccutrer:
@lolodomo: I finally got around to this. and no, there is no chance that `type` in those locations could be `UNDEF` (or `NULL`, or `null`). I've updated the prototypes to be `Command` instead of `Type` to ensure this as well. Those methods are called either from `ChannelState.publishValue`, in which case it's already a Command, or `ChannelState.processMessage`, which used `TypeParser.parseCommand` first, taking the declared types each Value allows, and returning if it's not one of those types.

ccutrer:
@jochen314 @lolodomo any more feedback on this?

lolodomo (changes requested):
The changes look now more impacting than in the original PR. I hope you are really sure of these changes as it looks to me as a significative change of behaviour of the MQTT generic binding.

lolodomo:
I will prefer a second review by @openhab/add-ons-maintainers before merging this PR, I have some difficulties to evaluate the risks caused by these changes on all MQTT bindings in openHAB.

bundles/org.openhab.binding.mqtt.generic/src/main/java/org/openhab/binding/mqtt/generic/ChannelState.java
```diff
@@ -348,10 +351,6 @@ public boolean hasSubscribed() {
      *         and exceptionally otherwise.
      */
     public CompletableFuture<Boolean> publishValue(Command command) {
-        cachedValue.update(command);
```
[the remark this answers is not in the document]
ccutrer: Correct, and was the entire impetus for this PR. The disambiguation between the direction of values was a happy side effect.

ccutrer:
+1 additional eyes. It is a significant change. But I have been running it at my home for several months, and I use all of the MQTT sub-bindings (manually configured things/channels; homie; home assistant; ESP Milight Hub). I'm very confident that the only end-user visible change is that if you send a command to an MQTT channel, and the device does not update a corresponding state topic, and then you send REFRESH to the channel, the cached value will not have changed from the command. If you never send REFRESH to the channel, you'll never notice that the cached value wasn't changed. If you're concerned about this, I could see having an additional configuration option to say "assume the published values are accepted, and update internal cache". However, this seems contradictory to me with the auto update framework and state predictions. Maybe a better approach would be a way to say "for this MQTT device, make the AutoUpdatePolicy RECOMMEND rather than DEFAULT." If either of those get recommended, I'd rather implement them as a separate PR though - this one is already large enough.

lsiepel:
Would be nice if this can be merged for the next milestone so we can get some testing milage. I did some basic tests (not using all mqtt sub bindings though) and did not have any issue to report.

antroids:
> @antroids : any chance you'd be willing to look at this?

LGTM since it was tested and didn't broke something.

As for me, I would like to move State out of Value, rename Values to ValueHandlers or MqttTypes and make them immutable. These parse methods could be something like reducers: Command reduceIncoming(Outgoing)MqttMessage(State previousState, Command message). 
This could be another refactoring PR, for sure.

lolodomo:
Are you are several to be confident with these changes, I will merge them.
I have to take a look to additional changes applied after my last review.

antroids:
> I've tested this in my local openhab installation to confirm that an MQTT channel with separate state and command topics, where the state topic does not change after sending a command, the item is set to autoupdate, I can now use REFRESH against the item and the prior state is restored.

Trying to figure out all possible flows.
What if we have Percentage (or any depending on the internal state) channel with the only command topic? 
I don't think that it will work properly after the change.
Maybe we should consider command only channels as optimistic and update the value for them.

ccutrer:
He's referring to IncreaseDecreaseType essentially not working for a channel without a state topic, since we no longer keep track of the last command sent to it. I would argue that it never _should_ have worked. What does it mean to ask a channel to increase if it doesn't keep track of its value? Alternately, we should enable (optionally) passing such commands directly through to MQTT, which is one of the follow up PRs I've been waiting for this one to be merged before working on.

antroids:
> > What if we have Percentage (or any depending on the internal state) channel with the only command topic?
> > I don't think that it will work properly after the change.
> > Maybe we should consider command only channels as optimistic and update the value for them.
> 
> Just before I push the merge button, I read that. Does it mean that I should not merge ?

I think It can be merged. 
It rather addon design issue than error in this particular PR.

PR #12242, opened 2022-02-08
[mqtt.homie] remove unused constants in tests

ccutrer:
I plan on doing some more work in this binding, but want to clean up
the tests a bit first so I have less distraction while writing new tests.
These constants seem to be copy/pasted from the mqtt.generic binding
tests, and then never used.


jlaur (approved):
LGTM - thanks!

PR #12243, opened 2022-02-08
[insteon] Prevent null values from being inferred when printing to console

wborn:
This prevents the null analysis from inferring null values when printing to a Console.

Related to openhab/openhab-core#2742


robnielsen (approved):
lgtm

jlaur (approved):
Thanks!

jlaur:
@wborn - do I understand correctly that it's related to openhab/openhab-core#2742. but without any dependency, so ready for being merged?

wborn:
Yes it can already be merged. It prevents a compilation issue when https://github.com/openhab/openhab-core/pull/2742 is merged eventually. I just rewrote the code to help the null analysis with inferring that the printed values are never null. If you operate on the entrySet of a map you'll never run into non-existing keys.

PR #12245, opened 2022-02-08
[deconz] Fix unreachable lights and sensors being marked as GONE

### LINKED ISSUE #12244: [deconz] Unreachable devices are marked as GONE
Devices which are not reachable are marked as **OFFLINE** with status details **GONE**.

## Expected Behavior
Unreachable devices should not be treated as an error as this is a common temporary state, for example when a bulb is turned off. According to the [documentation](https://www.openhab.org/docs/concepts/things.html#status-details) **GONE** means: _"The Thing has been removed from the bridge or the network to which it belongs and is no longer available for use. The user can now remove the Thing from the system."_

## Current Behavior
When a device becomes unreachable:

        "state": {
            "reachable": false
        },

it is marked as **OFFLINE** and **GONE**.

## Possible Solution
In the Hue binding this is handled correctly by setting status details to **NONE**:
https://github.com/openhab/openhab-addons/blob/99cfb65abaf5fbdc65ebf87b7c83222440f7560b/bundles/org.openhab.binding.hue/src/main/java/org/openhab/binding/hue/internal/handler/HueLightHandler.java#L504-L510

## Steps to Reproduce
Remove power for a device and wait for it to be detected unreachable.

jlaur:
Fixes #12244

Please see linked issue for details.


lolodomo (approved):
LGTM

PR #12249, opened 2022-02-08
[voskstt] initial contribution

GiviMAD:

This STT service allows offline "Speech-to-Text" on openHAB.


bundles/org.openhab.voice.voskstt/src/main/java/org/openhab/voice/voskstt/internal/VoskSTTService.java
```diff
@@ -0,0 +1,226 @@
...
+                        sttListener.sttEventReceived(new SpeechRecognitionErrorEvent(config.noResultsMessage));
+                    } else {
+                        sttListener.sttEventReceived(new SpeechRecognitionErrorEvent("No results"));
+                    }
+                }
+            } catch (IOException e) {
+                logger.warn("Error running speech to text: {}", e.getMessage());
+                if (config.errorMessage.isBlank()) {
+                    sttListener.sttEventReceived(new SpeechRecognitionErrorEvent("Error"));
+                } else if (!config.errorMessage.isBlank()) {
```
lolodomo, line 204: No need of the if as it will always be true

bundles/org.openhab.voice.voskstt/src/main/java/org/openhab/voice/voskstt/internal/VoskSTTService.java
```diff
@@ -0,0 +1,226 @@
...
+                        trySleep(100);
+                        continue;
+                    }
+                    if (recognizer.acceptWaveForm(b, nbytes)) {
+                        lastInputTime = System.currentTimeMillis();
+                        var result = recognizer.getResult();
+                        logger.debug("Result: {}", result);
+                        ObjectMapper mapper = new ObjectMapper();
+                        var json = mapper.readTree(result);
+                        transcriptBuilder.append(json.get("text").asText()).append(" ");
```
lolodomo, line 180: No risk of exceptions other than IOException in these 2 lines ?

bundles/org.openhab.voice.voskstt/src/main/java/org/openhab/voice/voskstt/internal/VoskSTTService.java
```diff
@@ -0,0 +1,226 @@
...
+                    sttListener.sttEventReceived(new SpeechRecognitionErrorEvent("Error"));
+                } else if (!config.errorMessage.isBlank()) {
+                    sttListener.sttEventReceived(new SpeechRecognitionErrorEvent(config.errorMessage));
+                }
+            }
+            try {
+                audioStream.close();
+            } catch (IOException e) {
+                logger.warn("IOException on close: {}", e.getMessage());
+            }
```
lolodomo, line 212: I have a doubt: this is the STT service that closes the source ?
You did like that in the two other STT services ?

lolodomo (changes requested):

bundles/org.openhab.voice.voskstt/src/main/java/org/openhab/voice/voskstt/internal/VoskSTTService.java
```diff
@@ -0,0 +1,226 @@
...
+    }
+
+    @Override
+    public String getLabel(@Nullable Locale locale) {
+        return SERVICE_NAME;
+    }
+
+    @Override
+    public Set<Locale> getSupportedLocales() {
+        return Set.of(localeService.getLocale(null));
```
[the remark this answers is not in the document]
GiviMAD: I make that assumption because you language support depends on vosk, but you are right it makes no sense to return languages not supported by openHAB, how can I get a list of all supported locales to return here?

bundles/org.openhab.voice.voskstt/src/main/java/org/openhab/voice/voskstt/internal/VoskSTTService.java
GiviMAD: No, but the Watson STT already does this internally, and for the Google one I will add a PR to do it. The current dialog processor implementation only ensure the stream is closed before open a new one which I think is fine, but due to that I think the STT service should be on charge of closing the stream when it will no longer be used to free the resources as soon as posible.

bundles/org.openhab.voice.voskstt/README.md
```diff
@@ -0,0 +1,59 @@
+# Vosk Speech-to-Text
+
+Vosk STT Service uses [vosk-api](https://github.com/alphacep/vosk-api) to perform offline speech-to-text on OpenHAB.
+
+[Vosk](https://alphacephei.com/vosk/) is an offline open source speech recognition toolkit.
+It enables speech recognition for 20+ languages and dialects - English, Indian English, German, French, Spanish, Portuguese, Chinese, Russian, Turkish, Vietnamese, Italian, Dutch, Catalan, Arabic, Greek, Farsi, Filipino, Ukrainian, Kazakh, Swedish, Japanese, Esperanto.
+More to come.
```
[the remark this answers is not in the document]
GiviMAD: This part is a copy/paste from the vosk website. As is like their claim, I will prefer to let it as it is.

bundles/org.openhab.voice.voskstt/src/main/java/org/openhab/voice/voskstt/internal/VoskSTTService.java
GiviMAD: Does it makes sense to you?

GiviMAD:
@lolodomo I have added a new option to keep the language model loaded, I think is a nice one to have.
Sorry for the late changes, they are is a separate commit.

bundles/org.openhab.voice.voskstt/src/main/java/org/openhab/voice/voskstt/internal/VoskSTTService.java
lolodomo: No

bundles/org.openhab.voice.voskstt/src/main/java/org/openhab/voice/voskstt/internal/VoskSTTService.java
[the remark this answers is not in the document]
lolodomo: The purpose here is to list the locales supported by Vosk, not openHAB. Is this list unknown? You are listing them in README.

bundles/org.openhab.voice.voskstt/src/main/java/org/openhab/voice/voskstt/internal/VoskSTTService.java
lolodomo: You're tight. 
I am just asking myself if this is not something to fix rather in the dialog processor? I am not sure. 
If you really think it is more the job of each STT service to close the audio source, that's ok for me but it should be documented (probably in the interface description). 
WDYT?

lolodomo:
Please add the missing i18n properties.

bundles/org.openhab.voice.voskstt/src/main/java/org/openhab/voice/voskstt/internal/VoskSTTService.java
GiviMAD: In my opinion the current implementation of the dialog processor is ok, as the stream will be always closed at the end. I think is something that doesn't need an interface change, maybe a comment on the interface method to tell that is ok to close the stream when finished is enough.

Related to this, there is something I want to change on all the voice add-ons (the ones I uploaded recently), I want stop closing the scheduled task inside the abort call, and just let it finish ensuring it does not produce more events. That close call was causing this error https://github.com/openhab/openhab-core/pull/2761.
Is it ok to you?

bundles/org.openhab.voice.voskstt/src/main/java/org/openhab/voice/voskstt/internal/VoskSTTService.java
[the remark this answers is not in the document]
GiviMAD: Then I think is better like this, as the language support depends on the availability of language models this could change without need to change the vosk version, so I think we will be adding a limitation that does not exists.
But I can change it, let me know.

bundles/org.openhab.voice.voskstt/src/main/java/org/openhab/voice/voskstt/internal/VoskSTTService.java
lolodomo: Ok

bundles/org.openhab.voice.voskstt/src/main/resources/OH-INF/i18n/voskstt.properties
```diff
@@ -0,0 +1,16 @@
...
+voice.config.voskstt.maxSilenceSeconds.label = Max Silence Seconds
+voice.config.voskstt.maxSilenceSeconds.description = Only works when singleUtteranceMode is disabled, max seconds without getting new transcriptions to stop listening.
+voice.config.voskstt.maxTranscriptionSeconds.label = Max Transcription Seconds
+voice.config.voskstt.maxTranscriptionSeconds.description = Max seconds to wait to force stop the transcription.
+voice.config.voskstt.noResultsMessage.label = No Results Message
+voice.config.voskstt.noResultsMessage.description = Message to be told when no results.
+voice.config.voskstt.preloadModel.label = Preload Model
+voice.config.voskstt.preloadModel.description = Keep language model loaded. If true model is just reload the model on config updates, if not model will be loaded and offloaded on each execution. It will fallback to try to load the model when executed if it was not able to load it before.
+voice.config.voskstt.singleUtteranceMode.label = Single Utterance Mode
+voice.config.voskstt.singleUtteranceMode.description = When enabled recognition stops listening after a single utterance.
```
lolodomo, line 16: Please add the missing service label like in your STT other services

lolodomo (changes requested):

bundles/org.openhab.voice.voskstt/src/main/java/org/openhab/voice/voskstt/internal/VoskSTTService.java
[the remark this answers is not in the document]
lolodomo: By doing that, you are probably limiting the service to one language, the default language that is setup. 
I have to check how is used this method by the dialog processor.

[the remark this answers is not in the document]
lolodomo: But I have the feeling that the service can only work with one unique model and so one unique locale? 
Isn't it possible to find the locale from the model directory?

GiviMAD:
> Please add the missing i18n properties.

Done.
I think everything is in place unless the getSupportedLocales problem. Let me know if I miss something else.

bundles/org.openhab.voice.voskstt/src/main/java/org/openhab/voice/voskstt/internal/VoskSTTService.java
```diff
@@ -0,0 +1,291 @@
...
+    @Override
+    public Set<AudioFormat> getSupportedFormats() {
+        return Set.of(
+                new AudioFormat(AudioFormat.CONTAINER_WAVE, AudioFormat.CODEC_PCM_SIGNED, false, null, null, 16000L));
+    }
+
+    @Override
+    public STTServiceHandle recognize(STTListener sttListener, AudioStream audioStream, Locale locale, Set<String> set)
+            throws STTException {
+        Future<?> task;
```
lolodomo, line 150: Please suppress this variable which is now unused.
And take a look on SAT checks to be sure that this is clean.

lolodomo (changes requested):

bundles/org.openhab.voice.voskstt/src/main/java/org/openhab/voice/voskstt/internal/VoskSTTService.java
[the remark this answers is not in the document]
lolodomo: Please add a comment of this kind in the method: as it is not possible to determine the language of the model that was downloaded and setup by the user, it is assumed the language of the model is matching the locale of the openHAB server.

And there is something to fix in startDialog (openHAB core): it should be checked that the provided locale is supported by each service. I will propose a fix. getSupportedLocales is currently never used.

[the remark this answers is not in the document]
GiviMAD: > And there is something to fix in startDialog (openHAB core): it should check that the provided locale is supported by each service. I will propose a fix. getSupportedLocales is currently never used.
As getSupported locales is not used right now I will suggest to just warn the user for incompatibility.

Ok I will add the comment, thanks for the help.

bundles/org.openhab.voice.voskstt/src/main/java/org/openhab/voice/voskstt/internal/VoskSTTService.java
GiviMAD: Done.
This is reported for all this kind of services 'Unused configuration reference with uri - voice:voskstt'. As it was already happening on others previous add-ons like the GoogleTSS I assumed it was a known error. Just in case you were not aware of it.

lolodomo (approved):
LGTM

Issue #8216, opened 2020-07-28
Normalize thread naming of threads in add-ons


Hilbrand:
A number of add-ons start there own threads. To track threads better I propose these threads should all name their thread using a specific name convention. My proposal is to use:
* If there is 1 thread::`OH-binding-<thingUID>`,
* If there are multiple threads: `OH-binding-<thingUID>-<custom>`. Where `custom` is a developer defined word that suggest which specific thread it is.

Some bindings use `Executors`. My suggested would be to either standardize to all use the `org.eclipse.smarthome.core.common. ThreadPoolManager` or another option to allow a generic `Executors`, but initialize with the openHAB `NamedThreadFactory`. The named thread factory/pool manager should than be initialized with `binding-<thingUID>`. Note that using the ThreadPoolManager` will add an incremental number to the thread name for each thread started.

fwolter:
Many bindings append hostname and port number to the thread name. I think this is more meaningful than a number. I suggest to leave the part after the binding ID up to the binding developer.

Hilbrand:
Yes that might be a better idea, to leave it to the developer. Maybe the ThingID and some additional info so one can see which thing has started the thread. Or just instead of bindingId use the thing UID. I'll update the text.

kaikreuzer:
I'd like to mention that the general advice is to NOT use any threads at all within bindings. Most things should be possible to solve with the existing `scheduler` that is provided within the handlers. Requiring a dedicated thread should thus be an exception for the situations where it is really needed, such as continuously listening on TCP ports for events.

Please keep that in mind for all binding reviews!

Hilbrand:
@kaikreuzer my idea is to go through all bindings and see how threads are used and see if we can make them consistent. For example some use thread.start, while others use a single thread executor. Some use our thread executor to create a thread pool, others use the generic executor. I don't know what is preferred. But would like if we can make a general guideline. (Besides don't use a thead when not needed) Do you have an opinion about this?

kaikreuzer:
@Hilbrand Yes, I think a general guidelines makes sense.
Just like we offer thread pools through `ThreadPoolManager`, we should have a way to deal with single/permanent threads.
I don't have a strong opinion, how this guideline should look like, so feel free to suggest one. Aligning thread naming across all binings as a first step definitely makes sense as well 👍 .

lsiepel:
How can we determine if all bindings have the thread name set correct?

PR #12250, opened 2022-02-09
[smsmodem] Initial contribution

dalgwen:
This binding connects to a USB serial GSM modem (or a network exposed one, a.k.a ser2net) and allows openHAB to send and receive SMS through it.
With this binding, there is no need for complex configuration with external tools, such as gammu.

[Here is the community discussion](https://community.openhab.org/t/smsmodem-binding-for-sending-and-receiving-sms-on-a-gsm-serial-modem/128412), where you can find a binary snapshot.

This binding uses smslibv4, directly included (org.openhab.binding.smsmodem.internal.smslib packages and sub packages) and heavily refactored to pass openHAB code standard. I [asked on the forum if it was OK](https://community.openhab.org/t/inclusion-of-a-very-slightly-modified-library-in-the-source-code-of-a-binding-and-subsequent-issue/127597/2) and decided to go on with the inclusion. Hope this is OK. Honestly, I don't understand everything in this library : SMS is a complex standard and I didn't take the time to grab it as the smslibv4 was here for me to hide the complexity.

The code still has a "deviation" from the openHAB standard : in the internal smslib, it uses a manually created Thread to continuously read from the serial port (the PollReader class). As the thread is continuously running, I thought it could be a valid use case (to not take forever a thread from the openHAB pool). If you don't think so, please tell me and I will refactor it.


Hilbrand:
There are other add-ons that include third party code. But because it's under a different license this needs to be added differently. You need to put the sources in a different map: `src/3thparty/java/` and than you can keep the original package and such, and add the directory to the `pom.xml`. See for example the smartmeter binding: https://github.com/openhab/openhab-addons/tree/main/bundles/org.openhab.binding.smartmeter. Don't

The way you have added it now seems to me illegal as you have changed the license on the 3rd party source code. Which you only can do if you are the copyright owner of the source code.

Regarding threads. It's ok to have an own thread. In fact if you have long running threads you must create them as only short running threads are allowed on the scheduled executor. See also #8216 about naming the thread and do `setDeamon(true)`.

dalgwen:
Hello @Hilbrand,

Thank you very much for these informations.
Indeed, the way I included the code from smslib was problematic and I'm glad you directed me to a valid solution.
So, I did as you said, and refactored the inclusion. For information :
- I put back the smslib code in the org.smslib package
- I put this package and subpackages in the src/3rdparty/java folder
- I deleted the openHAB copyright information I had improperly added on top of every org.smslib file (though I let the mention "extracted from smslib" that I had put before).
- I completed the NOTICE file and the pom.xml with relevant information (I took the smartmeter binding as a model)

I also put a name and the daemon property on the always living thread polling the device.

bundles/org.openhab.binding.smsmodem/pom.xml
```diff
@@ -0,0 +1,39 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns="http://maven.apache.org/POM/4.0.0"
+  xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+
+  <modelVersion>4.0.0</modelVersion>
+
+  <parent>
+    <groupId>org.openhab.addons.bundles</groupId>
+    <artifactId>org.openhab.addons.reactor.bundles</artifactId>
+    <version>3.3.0-SNAPSHOT</version>
```
wborn, line 10:
```suggestion
    <version>3.4.0-SNAPSHOT</version>
```

fwolter:
@dalgwen It's been a long time you submitted your new addon. Are you still around to implement upcoming review feedback?

dalgwen:
Yes, I would be happy to do so.
I use this binding every day and also provide support on the forum for the (very) few interested users.

But I'm aware that it doesn't attract many people and is low on the priority list... so no problem if it takes time !

Thanks,

bundles/org.openhab.binding.smsmodem/README.md
```diff
@@ -0,0 +1,129 @@
+# SMSModem Binding
+
+This binding connects to a USB serial GSM modem (or a network exposed one, see ser2net) and allows openHAB to send and receive SMS through it.
+
+Serial modem should all use the same communication protocol (AT message) and therefore this binding _should_ be compatible with every dongle. However, there is a gap between theory and reality and success may vary. The protocol stack is based on the no longer supported smslib project (more precisely a v4 fork), and all modems supported by this library should be OK. The following devices have been reported functional :
```
fwolter, line 5: If you put each sentence in a separate line, reading the diffs would be easier.

bundles/org.openhab.binding.smsmodem/README.md
```diff
@@ -0,0 +1,129 @@
...
+
+## Thing Configuration
+
+The *smsmodembridge* thing requires at least two parameters to work properly (serialPortOrIP, baudOrNetworkPort).
+Depending on the nature of the connection (direct serial modem, or serial over network), this two fields will be used differently :
+
+| field | direct serial modem   | serial over network                  |
+|-------|--------------------------|-----------------------------------|
+|serialPortOrIP| The serial port to access (eg. /dev/tty/USBx) | IP address of the computer hosting the ser2net service|
+|baudOrNetworkPort| Baud rate                                 | The network port of the ser2net service |
```
fwolter, line 29: Did you consider creating two bridge types? One for direct connection and another for ser2net?

bundles/org.openhab.binding.smsmodem/README.md
```diff
@@ -0,0 +1,129 @@
...
+
+The other parameters are optional :
+
+| field | description     |
+|-------|--------------------------------------|
+|simPin          |If your sim card is protected, fill this field with the PIN code|
+|pollingInterval| Delay between two checks for new message|
+|delayBetweenSend|Delay to wait between two messages post (could be necessary for slow modem)|
```
fwolter, line 37: You could specify the type of the parameter (String, Decimal, ...) and the unit (seconds, ...).

bundles/org.openhab.binding.smsmodem/README.md
```diff
@@ -0,0 +1,129 @@
...
+
+```
+Bridge smsmodem:smsmodembridge:adonglename [ serialPortOrIP="/dev/ttyUSB0", baudOrNetworkPort="19200", enableAutoDiscovery="true" ]
+```
```
fwolter, line 41: This should go in the Full Example section.

bundles/org.openhab.binding.smsmodem/README.md
```diff
@@ -0,0 +1,129 @@
...
+
+The *smsconversation* thing is just a shortcut to address/receive messages with a specific msisdn. It is not mandatory to use the binding, as you can use action and trigger channel to send/receive a message once the smsmodem bridge is configured.
+
+| field | description               |
```
fwolter, line 45:
```suggestion
| Parameter Name | description               |
```

bundles/org.openhab.binding.smsmodem/README.md
```diff
@@ -0,0 +1,129 @@
...
+|-------|--------------------------|
+| recipient | The msisdn of the phone you want to discuss with.|
+| deliveryReport | If enabled, ask the network for a delivery report (default false)|
+| encoding | The encoding to use when sending the message (either Enc7, Enc8, EncUcs2, EncCustom, default is Enc7). EncUcs2 is good for non latin character, but SMS character size limit is then reduced|
```
fwolter, line 49: Please specify the types of the parameters.

bundles/org.openhab.binding.smsmodem/README.md
```diff
@@ -0,0 +1,129 @@
...
+| receive | String| The last message received |
+| send | String| A message to send |
+|deliverystatus| String| Delivery status (either UNKNOWN, QUEUED, SENT, PENDING, DELIVERED, EXPIRED, or FAILED). Several status are only possible if the delivery report parameter is enabled|
+
+## Trigger channels
+
+The *smsmodembridge* has the following trigger channel :
+| Channel ID          | event                      |  
+|---------------------|----------------------------|
+|receivetrigger| The msisdn and message received (concatened with the '\|' character as a separator)|
```
fwolter, line 70: "trigger" is redundant and can therefore be removed.
```suggestion
|receive| The msisdn and message received (concatened with the '\|' character as a separator)|
```

bundles/org.openhab.binding.smsmodem/README.md
```diff
@@ -0,0 +1,129 @@
...
+
+```
+smsAction.sendSMS("1234567890", "Hello world!", "EncUcs2")
+```
+
+## Full Example
+
+### Send SMS
+
+`sms.rules` for DSL :
```
fwolter, line 105:
```suggestion
`sms.rules` for DSL:
```

fwolter (changes requested):
I reviewed the readme so far.

There are some checkstyle warnings left. You could take a look at `target/code-analysis/report.html` after building it with `mvn clean install`.

bundles/org.openhab.binding.smsmodem/pom.xml
dalgwen: Done, by rebasing

bundles/org.openhab.binding.smsmodem/README.md
dalgwen: Done

bundles/org.openhab.binding.smsmodem/README.md
dalgwen: Indeed ! Done

bundles/org.openhab.binding.smsmodem/README.md
dalgwen: Done

bundles/org.openhab.binding.smsmodem/README.md
dalgwen: Done

bundles/org.openhab.binding.smsmodem/README.md
dalgwen: There is already a "receive" channel. So you are saying, there won't be a naming issue ?

bundles/org.openhab.binding.smsmodem/README.md
dalgwen: Done. Why french and english can't have the same rules ??? 😅

dalgwen:
Thank you @fwolter !
I checked the code-analysis but the report says there is no issue now.
Do you still have some ?

bundles/org.openhab.binding.smsmodem/README.md
fwolter: I think it would be better from the usability perspective, but also from the modelling. But I won't insist on changing it.

bundles/org.openhab.binding.smsmodem/README.md
fwolter: Ah, seems like I overlooked this. Then, everything is fine.

bundles/org.openhab.binding.smsmodem/README.md
fwolter: Interesting. Wasn't aware of that fine difference.

fwolter:
> I checked the code-analysis but the report says there is no issue now.

Looks good! Just give me a note if you want to change the modelling of the bridge or whether I shall continue reviewing.

dalgwen:
Thank you @fwolter 
You know better than me the good usage 👍, I will change the modelling of the bridge

dalgwen:
As proposed, I split the bridge thing-type in two different bridge thing-types.
Binding is now ready for review.

bundles/org.openhab.binding.smsmodem/src/main/java/org/openhab/binding/smsmodem/internal/handler/SMSConversationHandler.java
```diff
@@ -0,0 +1,152 @@
...
+                case UNINITIALIZED:
+                    updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.BRIDGE_OFFLINE);
+                    break;
+                case UNKNOWN:
+                    updateStatus(ThingStatus.UNKNOWN, ThingStatusDetail.BRIDGE_OFFLINE);
+                    break;
+            }
+        } else {
+            updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.BRIDGE_OFFLINE);
+        }
```
fwolter, line 150: The Thing status is set by the framework automatically according the bridge status, if you set the Thing status to UNKNOWN in `initialize()`. Then, this could be removed.

bundles/org.openhab.binding.smsmodem/src/main/java/org/openhab/binding/smsmodem/internal/handler/SMSModemBridgeHandler.java
```diff
@@ -0,0 +1,473 @@
...
+
+    /**
+     * The smslib object responsible for the serial communication with the modem
+     */
+    private @NonNullByDefault({}) Modem modem;
+
+    /**
+     * A scheduled watchdog check
+     */
+    private @NonNullByDefault({}) ScheduledFuture<?> checkScheduled;
```
fwolter, line 92: The null check annotations should only be disabled in certain cases like injection. These `@NonNullByDefault({})` should be removed and the attributes be checked for null values before they are accessed.

bundles/org.openhab.binding.smsmodem/src/main/java/org/openhab/binding/smsmodem/internal/handler/SMSModemBridgeHandler.java
```diff
@@ -0,0 +1,473 @@
...
+        shouldRun = false;
+        checkScheduled.cancel(true);
+        scheduler.execute(modem::stop);
+        modem.registerStatusListener(null);
+        modem.registerMessageListener(null);
+        modem.registerInformationListener(null);
+        modem = null;
+    }
+
+    public SMSModemBridgeHandler(Bridge bridge, SerialPortManager serialPortManager) {
```
fwolter, line 111: The constructur should be placed after the field declerations.

bundles/org.openhab.binding.smsmodem/src/main/java/org/openhab/binding/smsmodem/internal/handler/SMSModemBridgeHandler.java
```diff
@@ -0,0 +1,473 @@
...
+                logger.debug("Now trying to start SMSModem {}", logName);
+                modem.registerStatusListener(this);
+                modem.registerMessageListener(this);
+                modem.registerInformationListener(this);
+                modem.start();
+                logger.debug("SMSModem {} started", logName);
+            }
+        } catch (ModemConfigurationException e) {
+            String message = e.getMessage();
+            logger.error(message, e);
```
fwolter, line 170: The log message could be removed, as the status update is already logged by the framework.

The state change originated by updateStatus() is logged to events.log. Including the status detail message.

bundles/org.openhab.binding.smsmodem/src/main/java/org/openhab/binding/smsmodem/internal/handler/SMSModemBridgeHandler.java
```diff
@@ -0,0 +1,473 @@
...
+            updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.COMMUNICATION_ERROR, message);
```
fwolter, line 171: Is this rather a CONFIGURATION_ERROR?

bundles/org.openhab.binding.smsmodem/src/main/java/org/openhab/binding/smsmodem/internal/handler/SMSModemBridgeHandler.java
```diff
@@ -0,0 +1,473 @@
...
+    public void messageReceived(InboundMessage message) {
+        String sender = message.getOriginatorAddress().getAddress();
+        Payload payload = message.getPayload();
+        String messageText;
+        if (payload.getType().equals(Type.Text)) {
+            String text = payload.getText();
+            if (text != null) {
+                messageText = text;
+            } else {
+                logger.error("Message has no payload !");
```
fwolter, line 247: See above. Logging to error. Please check all.

bundles/org.openhab.binding.smsmodem/src/main/java/org/openhab/binding/smsmodem/internal/handler/SMSModemBridgeHandler.java
```diff
@@ -0,0 +1,473 @@
...
+    }
+
+    @Override
+    public void childHandlerInitialized(ThingHandler childHandler, Thing childThing) {
+        if (childHandler instanceof SMSConversationHandler) {
+            childHandlers.add((SMSConversationHandler) childHandler);
+        } else {
+            logger.error("The SMSModemBridgeHandler can only handle SMSConversation as child");
+        }
+    }
```
fwolter, line 226: Are you aware you can access the child Things via `getBridge().getThings()`?

bundles/org.openhab.binding.smsmodem/src/main/java/org/openhab/binding/smsmodem/internal/handler/SMSModemBridgeHandler.java
```diff
@@ -0,0 +1,473 @@
...
+    public Collection<Class<? extends ThingHandlerService>> getServices() {
+        return Set.of(SMSModemActions.class, SMSConversationDiscoveryService.class);
+    }
+
+    @Override
+    public boolean processStatusCallback(Modem.Status oldStatus, Modem.Status newStatus) {
+        switch (newStatus) {
+            case Error:
+                logger.error("SMSLib reported an error on the underlying modem {}", modem.getDescription());
+                updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.COMMUNICATION_ERROR);
```
fwolter, line 325: You could add the error message as a third argument to `updateStatus()`. Then, the error message is displayed in the UI and the logging could be removed.

bundles/org.openhab.binding.smsmodem/src/main/resources/OH-INF/binding/binding.xml
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<binding:binding id="smsmodem" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+	xmlns:binding="https://openhab.org/schemas/binding/v1.0.0"
+	xsi:schemaLocation="https://openhab.org/schemas/binding/v1.0.0 https://openhab.org/schemas/binding-1.0.0.xsd">
+
+	<name>SMSModem Binding</name>
+	<description>This binding handle a GSM modem connected to the openHAB server (Serial), or exposed on the network. It
```
fwolter, line 7:
```suggestion
	<description>This binding handles a GSM modem connected to the openHAB server (Serial), or exposed on the network. It
```

bundles/org.openhab.binding.smsmodem/src/main/resources/OH-INF/thing/smsmodem.xml
```diff
@@ -0,0 +1,91 @@
...
+		</config-description>
+	</bridge-type>
+
+	<channel-type id="smsmodemreceivetrigger">
+		<kind>trigger</kind>
+		<label>Message Received</label>
+		<description>Triggered when a message is received, in the form "&lt;msisdn_sender&gt;|&lt;text&gt;"</description>
+		<event>
+		</event>
+	</channel-type>
```
fwolter, line 89: Can you apply the correct indentation?

fwolter (changes requested):

bundles/org.openhab.binding.smsmodem/src/main/java/org/openhab/binding/smsmodem/internal/SMSModemHandlerFactory.java
```diff
@@ -0,0 +1,105 @@
...
+            } else {
+                throw new IllegalArgumentException("Cannot create a SMSConversation without a SMSModem bridge");
+            }
+
+        }
+        throw new IllegalArgumentException("The thing type " + thingTypeUID + " is not supported by the binding.");
+    }
+
+    public static ThingUID getSMSConversationUID(ThingTypeUID thingTypeUID, String recipient, ThingUID bridgeUID) {
+        return new ThingUID(thingTypeUID, recipient, bridgeUID.getId());
```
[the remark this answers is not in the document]
dalgwen: The "getSMSConversationUID" method is called by the autodiscovery method, or from the createThing method if and only if the "thingUID" is null.
With this method, I want to give a default UID, IF NOT set by the user. If the user set the thing UID, it will use the user's one.
Your comment seems to mean that this approach is invalid, or do you mean it's badly implemented ?

bundles/org.openhab.binding.smsmodem/src/main/java/org/openhab/binding/smsmodem/internal/SMSModemBridgeConfiguration.java
```diff
@@ -0,0 +1,30 @@
...
+/**
+ * The {@link SMSModemBridgeConfiguration} class contains fields mapping bridge configuration parameters.
+ *
+ * @author Gwendal ROULLEAU - Initial contribution
+ */
+@NonNullByDefault
+public class SMSModemBridgeConfiguration {
+
+    public String serialPort = "";
+    public Integer baud = 9800;
```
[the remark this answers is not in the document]
dalgwen: Yes, indeed, 9600 seems more valid than 9800.
(Mine is 19200, I don't know what is the most used value.)

Done

bundles/org.openhab.binding.smsmodem/src/main/java/org/openhab/binding/smsmodem/internal/actions/SMSModemActions.java
```diff
@@ -0,0 +1,81 @@
...
+
+    @RuleAction(label = "Send Message With Special Encoding", description = "Send a message and specify encoding")
+    public void sendSMS(
+            @ActionInput(name = "recipient", label = "recipient", description = "Recipient of the message") @Nullable String recipient,
+            @ActionInput(name = "message", label = "message", description = "Message to send") @Nullable String message,
+            @ActionInput(name = "encoding", label = "encoding", description = "Encoding") @Nullable String encoding) {
+        if (recipient != null && !recipient.isEmpty() && message != null) {
+            handler.send(recipient, message, false, encoding);
+        } else {
+            logger.error("SMSModem cannot send a message with no recipient or text");
```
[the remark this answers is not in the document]
dalgwen: Thanks. (and sorry, I keep making the same mistake despite being told each time 😂... )

Done

bundles/org.openhab.binding.smsmodem/src/main/java/org/openhab/binding/smsmodem/internal/handler/SMSConversationHandler.java
```diff
@@ -0,0 +1,152 @@
...
+    public String getRecipient() {
+        return config.recipient.trim();
+    }
+
+    private synchronized @Nullable SMSModemBridgeHandler getBridgeHandler() {
+        if (this.bridgeHandler == null) {
+            Bridge bridge = getBridge();
+            if (bridge == null) {
+                logger.error("Required bridge not defined for SMSconversation {} with {}.", thing.getUID(),
+                        getRecipient());
```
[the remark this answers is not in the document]
dalgwen: Done

bundles/org.openhab.binding.smsmodem/src/main/java/org/openhab/binding/smsmodem/internal/handler/SMSConversationHandler.java
dalgwen: I made a test and if I set the child handler status to "UNKNOWN" and leave if for the Bridge to update (deleting setStatusByBridgeStatus() method), it stays "UNKNOWN", even if the bridge status switch to "online". 
Or maybe I didn't understand your proposition ?

By reading code, the bridgeStatusChanged() method of BaseThingHandler effectively change the child handler status to "ONLINE" when the bridge get online, but only if the last status details was BRIDGE_OFFLINE

bundles/org.openhab.binding.smsmodem/src/main/java/org/openhab/binding/smsmodem/internal/handler/SMSModemBridgeHandler.java
dalgwen: Done

bundles/org.openhab.binding.smsmodem/src/main/java/org/openhab/binding/smsmodem/internal/handler/SMSModemBridgeHandler.java
dalgwen: Thank you, modification done.

bundles/org.openhab.binding.smsmodem/src/main/java/org/openhab/binding/smsmodem/internal/handler/SMSModemBridgeHandler.java
dalgwen: Thanks, done

bundles/org.openhab.binding.smsmodem/src/main/resources/OH-INF/binding/binding.xml
dalgwen: Done

bundles/org.openhab.binding.smsmodem/src/main/resources/OH-INF/thing/smsmodem.xml
dalgwen: Sorry, I don't see what is the issue here. Is it for the "event" element ? if so, I changed it to <event/>

bundles/org.openhab.binding.smsmodem/src/main/java/org/openhab/binding/smsmodem/internal/SMSModemHandlerFactory.java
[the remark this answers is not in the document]
fwolter: In `SMSConversationDiscoveryService` you are calling `getSMSConversationUID()` directly, when creating the discovered Thing (which is the right approach). So, the default-ThingUID-use-case is handled there.

If the user adds the Thing manually, the ThingUID can't be left empty/null (neither via UI nor text based configuration).

What use case do you intend to handle by this code?

bundles/org.openhab.binding.smsmodem/src/main/java/org/openhab/binding/smsmodem/internal/handler/SMSConversationHandler.java
fwolter: You still need to set the child Thing ONLINE, if you think it is online. In your case this might be always the case, if the bridge is online. So, you can set it ONLINE in `initialize()` directly. Then, it will be online if the bridge is online and offline if the bridge is offline.

You only need to set it to UNKNOWN (and later to ONLINE) if you don't know the state in `initialize()`, yet.

dalgwen: Ah, I get it now !
Indeed, I can set it to ONLINE directly.
Done, thanks

By the way, there is one drawback : the bridge start as "unknown" because it has to establish connection. But the smsconversation thing is immediately online at this time (and the status doesn't reflect the reality : it is not fully functionnal because of the bridge).

bundles/org.openhab.binding.smsmodem/src/main/java/org/openhab/binding/smsmodem/internal/SMSModemHandlerFactory.java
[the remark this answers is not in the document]
dalgwen: No particular use case here. I was just trying to implement the createThing method from the parent class and respecting the annotation : in this method, the parameter thingUID is nullable.
If you thing it is best, I can remove the call to the getSMSConversationUID in createThing, but the compiler will log a warning I guess ?
Or I could fail fast with an exception if thingUID is null.
What do you think ?

bundles/org.openhab.binding.smsmodem/src/main/java/org/openhab/binding/smsmodem/internal/handler/SMSModemBridgeHandler.java
dalgwen: Done

bundles/org.openhab.binding.smsmodem/src/main/java/org/openhab/binding/smsmodem/internal/handler/SMSModemBridgeHandler.java
dalgwen: Done

bundles/org.openhab.binding.smsmodem/src/main/java/org/openhab/binding/smsmodem/internal/handler/SMSModemBridgeHandler.java
dalgwen: Done

bundles/org.openhab.binding.smsmodem/src/main/java/org/openhab/binding/smsmodem/internal/handler/SMSModemBridgeHandler.java
dalgwen: Both !
A misconfiguration (bad address) or network/serial error prevents from communicating with the modem, which is not joinable at the address set.
So, what should I use ?

bundles/org.openhab.binding.smsmodem/src/main/resources/OH-INF/thing/smsmodem.xml
dalgwen: My answer was truncated.
Sorry, I don't see what is the issue here. Is it for the "event" element ? if so, I changed it to <event/>

bundles/org.openhab.binding.smsmodem/src/main/java/org/openhab/binding/smsmodem/internal/handler/SMSModemBridgeHandler.java
fwolter: This is what I saw. That sounds more like a configuration error. I leave it up to you what makes more sense.
grafik

bundles/org.openhab.binding.smsmodem/src/main/java/org/openhab/binding/smsmodem/internal/handler/SMSConversationHandler.java
fwolter: Just tested your code and a reason why it doesn't work as expected is that you don't set the ThingStatus in the bridge's `initialize()` to UNKNOWN. It works for me if I do that.

bundles/org.openhab.binding.smsmodem/src/main/java/org/openhab/binding/smsmodem/internal/handler/SMSModemBridgeHandler.java
```diff
@@ -0,0 +1,487 @@
...
+            if (finalModem != null) {
+                finalModem.stop();
+            }
+            checkAndStartModemIfNeeded();
+        });
+    }
+
+    @Override
+    public void initialize() {
+        shouldRun = true;
```
fwolter, line 135: This is necessary to make the thing status automatism work.
```suggestion
        updateStatus(ThingStatus.UNKNOWN);
        shouldRun = true;
```

fwolter (changes requested):

bundles/org.openhab.binding.smsmodem/src/main/java/org/openhab/binding/smsmodem/internal/SMSModemHandlerFactory.java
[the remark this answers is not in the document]
dalgwen: Oh my ! I wouldn't have wasted your time if I knew that before... Sorry 😨
Deleted !

bundles/org.openhab.binding.smsmodem/src/main/java/org/openhab/binding/smsmodem/internal/handler/SMSConversationHandler.java
dalgwen: image

Strange, it doesn't work for me. When the bridge is initializing (unknown), the child conversation thing is online. Maybe an order initialization issue ?
But it is no big deal, I'm OK with that, so I let you decide if it needs further investigation.

bundles/org.openhab.binding.smsmodem/src/main/java/org/openhab/binding/smsmodem/internal/handler/SMSModemBridgeHandler.java
dalgwen: You are right.
I splitted it in two, depending on the root cause (IOException = communication error, other = configuration error).
Done

bundles/org.openhab.binding.smsmodem/src/main/java/org/openhab/binding/smsmodem/internal/handler/SMSConversationHandler.java
fwolter: I just tested it again and you're right, as long as the bridge is in UNKNOWN state, the child Things are ONLINE. Only if the bridge is set to OFFLINE, the child Things are set to OFFLINE, too. Normally you wouldn't notice, because the bridge stays in UNKNOWN state only for very short time. In this case the timeout is quite long.

I would expect the SMS modem to respond within subseconds and would chose a shorter timeout, anyway. But maybe there are things I didn't consider.

I merge it for now. If you want to improve the behavior, you could file a follow-up PR.

fwolter (approved):
LGTM

dalgwen:
Great, thank you again for your time !

lolodomo:
As the French translations were included on the PR, they have to be uploaded manually in Crowdin.

lolodomo:
@dalgwen : you forgot adding an entry in CODEOWNERS file, can you please submit a small PR for that?

dalgwen:
> As the French translations were included on the PR, they have to be uploaded manually in Crowdin.

First time in crowdin for me. I think I did what you suggest. With some fix after viewing warning from crowdin.

> you forgot adding an entry in CODEOWNERS file, can you please submit a small PR for that?

Oops, sorry. (Done here)[https://github.com/openhab/openhab-addons/pull/13834].

lolodomo:
> First time in crowdin for me. I think I did what you suggest. With some fixes after viewing warnings from crowdin.

I have validated with few changes.
Syncing is in progress, the PR should be created in few minutes.

PR #12252, opened 2022-02-09
[souliss] fix t31 fire and system statuses

lucacalcaterra:
Signed-off-by: Luca Calcaterra calcaterra.luca@gmail.com


lolodomo:
Changing the kind of an existing channel (from trigger to state) is a breaking change for users already using these channels. That's the reason why I added the corresponding label. Something will have to be mentioned in 3.3 release notes if this PR is merged.

lolodomo (approved):
LGTM

lolodomo:
I have the feeling that channels were wrongly marked as trigger channels but already handled as state channels !
So maybe there is no breaking changes here for users. But users could probably not linked these channels to items.

PR #12254, opened 2022-02-09
[mqtt.homie] handle exceptions parsing attributes

### LINKED ISSUE #10711: [mqtt.homie] Homie Thing never comes online
## Expected Behavior
Homie device conforming to specification comes online. If it can't a more detailed message (what topic did it not receive?) would be useful.

## Current Behavior
The thing initially shows `ONLINE`, but after a while it switches to `OFFLINE (COMMUNICATION_ERROR): Did not receive all required topics`. AFAICT all topics are present, and they're sent to the broker QoS 1 and retained: true. I have another device running the exact same code, but has fewer constituent parts that is working fine. I initially thought it was because it was just such a large device (851 topics), but I have another Homie device with 10,973 topics that works just fine. It behaves pretty much the same regardless if disable/enable the Thing, restart the broker or homie bindings, or restart all of OpenHAB. The key log lines:

```
2021-05-17 12:59:37.037 [INFO ] [ab.event.ThingStatusInfoChangedEvent] - Thing 'mqtt:homie300:mosquitto:monoprice10761a' changed from UNINITIALIZED (DISABLED) to INITIALIZING
2021-05-17 12:59:37.041 [DEBUG] [e.internal.handler.HomieThingHandler] - About to initialize Homie device null
2021-05-17 12:59:37.041 [DEBUG] [e.internal.handler.HomieThingHandler] - About to start Homie device null
2021-05-17 12:59:37.047 [INFO ] [ab.event.ThingStatusInfoChangedEvent] - Thing 'mqtt:homie300:mosquitto:monoprice10761a' changed from INITIALIZING to ONLINE
2021-05-17 12:59:42.042 [WARN ] [core.thing.internal.ThingManagerImpl] - Initializing handler for thing 'mqtt:homie300:mosquitto:monoprice10761a' takes more than 5000ms.
2021-05-17 13:00:07.044 [INFO ] [ab.event.ThingStatusInfoChangedEvent] - Thing 'mqtt:homie300:mosquitto:monoprice10761a' changed from ONLINE to OFFLINE (COMMUNICATION_ERROR): Did not receive all required topics
```

I'll attach a full log that shows all of the subscription logs, but basically all the subscribes happen within 0.1s after the thing goes ONLINE. I've also confirmed that every subscribeRaw line in the log has a corresponding Successfully subscribed line, and that for every one of those lines, a retained topic _does_ exist in the broker (with the exception of optional attributes $format, $unit, $retained, and $settable), so I'm not missing any required topics. And vice versa - every topic on the broker was subscribed to, with the exception of the actual state topics (i.e. all the metadata topics did try to subscribe). I have TRACE logging on for org.openhab.core.io.transport.mqtt and org.openhab.binding.mqtt - I don't see any logging that shows when a value was actually received though.

[openhab.log](https://github.com/openhab/openhab-addons/files/6496559/openhab.log)
[mqtt_topics.log](https://github.com/openhab/openhab-addons/files/6496563/mqtt_topics.log)

Add a discovered Homie Thing from the inbox.

## Context
I've looked at #7252, but AFAICT I should have the fix for that already in the code I'm running.

## Your Environment
* Version used: 3.4.0.M4
* Ubuntu Focal x86_64 (an Intel NUC)

ccutrer:
fixes #10711

technically this code is in mqtt.generic, but it's only used by Homie.

in particular, if an incoming string doesn't match an enum, this will now
just ignore the value instead of raising an exception to be caught somewhere inside
of Hive MQTT, and eventually timing out and logging that mandatory topics weren't
received, instead of logging a pointer to the actual problem. this makes it so that
if there's a homie $datatype openhab doesn't understand (like duration), it will be
able to get to the point of just choosing a string channel

also did some minor debug logging cleanup for mqtt:
 * fixed a typo
 * when logging homie device name from the thing handler, use the config deviceid,
   since we likely don't have the attributes from MQTT yet

### DIFF
--- a/bundles/org.openhab.binding.mqtt.generic/src/main/java/org/openhab/binding/mqtt/generic/ChannelState.java
+++ b/bundles/org.openhab.binding.mqtt.generic/src/main/java/org/openhab/binding/mqtt/generic/ChannelState.java
@@ -267,7 +267,7 @@ public class ChannelState implements MqttMessageSubscriber {
     }
 
     private void internalStop() {
-        logger.debug("Unsubscribed channel {} form topic: {}", this.channelUID, config.stateTopic);
+        logger.debug("Unsubscribed channel {} from topic: {}", this.channelUID, config.stateTopic);
         this.connection = null;
         this.channelStateUpdateListener = null;
         hasSubscribed = false;
--- a/bundles/org.openhab.binding.mqtt.generic/src/main/java/org/openhab/binding/mqtt/generic/mapping/SubscribeFieldToMQTTtopic.java
+++ b/bundles/org.openhab.binding.mqtt.generic/src/main/java/org/openhab/binding/mqtt/generic/mapping/SubscribeFieldToMQTTtopic.java
@@ -140,25 +140,35 @@ public class SubscribeFieldToMQTTtopic implements MqttMessageSubscriber {
         }
 
         String valueStr = new String(payload, StandardCharsets.UTF_8);
+        String originalValueStr = valueStr;
 
         // Check if there is a manipulation annotation attached to the field
-        final MQTTvalueTransform transform = field.getAnnotation(MQTTvalueTransform.class);
-        Object value;
-        if (transform != null) {
-            // Add a prefix/suffix to the value
-            valueStr = transform.prefix() + valueStr + transform.suffix();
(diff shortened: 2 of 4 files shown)


lolodomo (approved):
LGTM

PR #12259, opened 2022-02-10
[hdpowerview] Fix regression when bridge is offline while initializing child things

jlaur:
Fix recent regression because of misunderstanding: https://github.com/openhab/openhab-addons/pull/12061#discussion_r791120507

This pull request reverts:
- Repeater: d7e10c565a16ca40d8f04c386312449aeb5b878d
- Shade: 7626a88cabbc5337106cc0a8373be57d6fdfc711 (last change)

The logic is actually needed. The problem can be reproduced by configuring an invalid IP address for the bridge. This causes the bridge to be correctly marked as **OFFLINE**, but the shade/repeater things will be left in status **UNKNOWN**, when they should in fact also be **OFFLINE** with **BRIDGE_OFFLINE**.

### DIFF
--- a/bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/handler/HDPowerViewRepeaterHandler.java
+++ b/bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/handler/HDPowerViewRepeaterHandler.java
@@ -79,7 +79,12 @@ public class HDPowerViewRepeaterHandler extends AbstractHubbedThingHandler {
                     "@text/offline.conf-error.invalid-bridge-handler");
             return;
         }
-        updateStatus(ThingStatus.UNKNOWN);
+        ThingStatus bridgeStatus = bridge.getStatus();
+        if (bridgeStatus == ThingStatus.ONLINE) {
+            updateStatus(ThingStatus.UNKNOWN);
+        } else {
+            updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.BRIDGE_OFFLINE);
+        }
         scheduleRefreshJob();
     }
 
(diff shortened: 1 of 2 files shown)


lolodomo:
The correct pattern is even to first check if the bridge handler is not null. If null, set the thing status to OFFLINE / BRIDGE_UNINITIALIZED.
Then your fixed code is ok. 
The correct pattern can be found in many bindings.

jlaur:
> The correct pattern is even to first check if the bridge handler is not null. If null, set the thing status to OFFLINE / BRIDGE_UNINITIALIZED.

That's how it works (again), please look at the lines above.

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/handler/HDPowerViewRepeaterHandler.java
```diff
@@ -79,7 +79,12 @@ public void initialize() {
...
             return;
         }
-        updateStatus(ThingStatus.UNKNOWN);
+        ThingStatus bridgeStatus = bridge.getStatus();
+        if (bridgeStatus == ThingStatus.ONLINE) {
+            updateStatus(ThingStatus.UNKNOWN);
+        } else {
+            updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.BRIDGE_OFFLINE);
+        }
         scheduleRefreshJob();
```
lolodomo, line 88: Do you want to schedule refresh even if bridge is OFFLINE?

jlaur: Good point. This can be optimized, but require some more changes, so outside the scope of this PR. Now that you mention it, I'll reconsider the responsibility between hub handler and repeater handler. In case of more than one repeater this can be optimized if the bridge handler would get data for all repeaters in one call and notify all child repeater handlers.

lolodomo: So I merge as it is now in the PR, or not

jlaur: You can merge, this PR just fixes a regression (by reverting a change) caused by a misunderstanding with a PR comment.

lolodomo (approved):
LGTM

PR #12260, opened 2022-02-11
[actiontemplatehli] Initial contribution

GiviMAD:
NOTE: This interpreter was renamed from 'opennlphli' to 'actiontemplatehli' to avoid confusion.
This add-on adds an human language interpreter powered by openNLP that allows the user to define their own actions to read from items or send commands to them. Also supports to configure a  fallback interpreter, so you can use this one just to complement another.
Aims to be compatible with all languages and assumes the action templates are written in the configured openHAB language.
Supported feature list and samples are provided on the readme.
I hope that the community find it useful, for me is a solution for the custom commands I wanted to add to the standard interpreter.

### DIFF
--- a/bundles/org.openhab.voice.actiontemplatehli/README.md
+++ b/bundles/org.openhab.voice.actiontemplatehli/README.md
@@ -41,11 +41,11 @@ This is defined by the boolean field read which is 'false' by default.
 When read is false:
 
 * template: action template. (Required)
-* value: value to by send. It can be used to capture the transformed placeholder values. (Required unless if the target item type is String in that case rule mode is assumed true and the whole text is passed to item).
+* value: value to by send. It can be used to capture the transformed placeholder values. (Required unless if the target item type is String in that case silence mode is assumed true and the whole text is passed to item).
 * type: action template type.
 * requiredTags: allow to restrict the item targets by its tags.
 * placeholders: defined placeholders that can be used on the template and replaced on the value.
-* ruleMode: boolean used to avoid confirmation message.
+* silence: boolean used to avoid confirmation message.
 * targetMembers: when targeting a Group item, can be used to send the command to its member items instead.
 
 When read is true:
@@ -321,7 +321,7 @@ config:
         - Andrea
         - Jacob
         - Raquel
-  ruleMode: true
+  silence: true
   template: send message $* to $contact
   type: tokens
   value: $contact:$*
--- a/bundles/org.openhab.voice.actiontemplatehli/README.md
+++ b/bundles/org.openhab.voice.actiontemplatehli/README.md
@@ -1,0 +1,336 @@
+# Action Template Interpreter
+
+A human language interpreter implementation powered by OpenNLP.
+This is an attempt to provide you with a template system to match text commands to specific items and read its state or send command to them.
+For doing this the interpreter takes advantage of some nlp techniques.
+
+The (Apache OpenNLP)[https://opennlp.apache.org] library is a machine learning based toolkit for the processing of natural language text.
+This human language interpreter aims to have no language dependency as it does nothing out of the box, please report any incompatibility with your language.
+
+You can find models provided by OpenNLP for some languages [here](https://opennlp.apache.org/models.html) and [here](http://opennlp.sourceforge.net/models-1.5/).
+Those are not required, as you can use the build-in white space or simple tokenizers (from OpenNLP),they are just required to use the match by lemmas and the optional language tag functionalities.
+
+There are some examples at the end that you can review if you want a general idea of what can be done.
+
+## NLP Terminology
+
+I will briefly explain some terms that will be used:
+
+* Tokenize: first step of the recognition is to split the text, each of these parts is called token. 
+* Named Entity Recognition (NER): is the process of finding a subset of tokens on the input.
+* Part Of Speech (POS) tagging: categorizing tokens in a text, depending on the definition of the token and its context.
+* Lemmatize: is the process of getting a generic representation of the tokens, each of it is called lemma. (Example of one token to lemma conversion: 'is' -> 'be').
+
+## Action Template Target:
+
+This interpreter allows two ways of target and item:
+
+* You can link an action to a specific item by adding it the custom metadata namespace 'actiontemplatehli'.
+* You can link an action to all items of a type by providing the file '<OPENHAB_USERDATA>/actiontemplatehli/type_actions/<ITEM_TYPE_NAME>.json' (here you can restrict each action by item tags).
+
+Two important notes:
+
+* Actions linked to items have prevalence over actions linked to a whole item type.
+* On actions linked to an item type the itemLabel placeholder will always be applied (explained bellow), if there are multiple item label detected in the text this part will be aborted.
+
+## Action Templates Configuration:
+
+Actions can read from an item or send a command to it.
+This is defined by the boolean field read which is 'false' by default.
+
+When read is false:
+
+* template: action template. (Required)
+* value: value to by send. It can be used to capture the transformed placeholder values. (Required unless if the target item type is String in that case rule mode is assumed true and the whole text is passed to item).
+* type: action template type.
+* requiredTags: allow to restrict the item targets by its tags.
+* placeholders: defined placeholders that can be used on the template and replaced on the value.
+* ruleMode: boolean used to avoid confirmation message.
+* targetMembers: when targeting a Group item, can be used to send the command to its member items instead.
+
+When read is true:
+
+* template: action template. (Required)
+* value: read template, can use the placeholders symbols $itemLabel and $state.
+* emptyValue: An alternative template. Is used when the state value is empty or NULL after the post transformation. The $itemLabel is available.
+* type: action template type.
+* requiredTags: allow to restrict the item targets by its tags.
+* placeholders: only the placeholder with label state will be used, to process its POS transformation on the state.
+* targetMembers: when targeting a Group item, can be used access the state of one of its members, if multiples matches a warning is shown and the first one is used.
+
+## Placeholders
+
+This configuration allow you to define a symbols to use on your templates.
+You can define the sets of tokens to match using ner, and a transformation using pos.
+Those are its fields:
+
+ * label: label for the placeholder, is prefixed by '$' and spaces are replaced by '_' to create the symbol you can use on the template (Required).
+ * nerValues: list of string containing parts of the text to look for. Takes precedence over the ner field.
+ * ner: name for a file under the ner folder (<OPENHAB_USERDATA>/actiontemplatehli/ner), first it will look for a <ner>.bin model and then for a <ner>.xml dictionary for applying ner (prevalence over 'nerValues').
+ * posValues: apply a pos transformation with static values. Takes precedence over the pos field.
+ * pos: name for a file under the pos folder (<OPENHAB_USERDATA>/actiontemplatehli/pos), first it will look for a <pos>.bin model and then for a <pos>.xml dictionary (prevalence over 'posValues').
+
+The placeholder symbol replaces the text tokens matched using NER (before scoring the actions) and the captured value could be transformed using POS and will be accessible to the value under its symbol.
+As a summary, using the placeholders you can configure how parts of the speech are converted into valid item values and backward.
+
+There are some special placeholders, described in another section.
+
+### POS Transformation
+
+POS is a technique which produces tags for each token, here though we are going to use it to match a group of words with a value so we should transform those words to a single token.
+That's the reason why the whitespace character should be replaced by '__' in the POS dictionaries and static values, you can see some examples bellow.
+
+### Target members:
+
+When the target of an action is group item you can target its members instead.
+
+You can use the following fields:
+
+* itemName: name of the item member to target. If present the other fields are ignored.
+* itemType: type of the item members to target.
+* requiredTags: allow to restrict the member targets by its tags when matching by type.
+* recursive: when matching by itemType, look for group members in a recursive way, default true.
+* mergeState: on a read action when matching by itemType, merge the item states, only allowed for 'Switch' and 'Contact' item types, default false.
+
+## Special Placeholders
+
+### The 'itemLabel' Placeholder
+
+The itemLabel placeholder is always applied when scoring actions linked to item types. It's replaced using NER (no case-sensitive) with your item labels. Its value is only available for read actions.
+
+### The 'groupLabel' Placeholder
+
+The groupLabel placeholder is only available for read actions when targeting a group member.
+When it's present, the 'itemLabel' placeholder will take the value of the target member label and the 'groupLabel' the label of the group.
+
+### The 'state' Placeholder
+
+It's used to access the value on the read actions, you can configure a POS transformation for it.
+
+### The '*' Placeholder (the dynamic placeholder)
+
+The dynamic placeholder is designed to capture free text. The captured value is exposed to the value under the symbol '$*' as other placeholders.
+It has some restrictions:
+
+* Can not be the only token on the template.
+* Can not be used as an optional token.
+* Can not be used multiple type on the same template alternative.
+
+Note that the dynamic placeholder does not score.
+
+## Text Preprocessing
+
+The interpreter needs to match the input text with a target item and action config, to know what to do.
+To do so, it need the tokens and optionally the POS tags and the lemmas.
+
+### Tokenizer
+
+You can provide a custom model at '<OPENHAB_USERDATA>/actiontemplatehli/token.bin', otherwise it will use the build in simple tokenizer or whitespace tokenizer (configurable).
+
+Tokenizing the text is enough to use the action type 'tokens' as tokens are the only ones required for scoring (but the option 'optionalLanguageTags' will not take effect unless you have the POS language tags).
+
+### POSTagger (language tags)
+
+You need to provide a model for POS tagging at '<OPENHAB_USERDATA>/actiontemplatehli/pos.bin' for your language. 
+This will produce a language tag for each token, that can be used in 'optionalLanguageTags' to make some optional for scoring.
+Note that you need the correct language tags for the lemmatizer to work.
+
+### Lemmatizer
+
+You need to provide a model for the lemmatizer at '<OPENHAB_USERDATA>/actiontemplatehli/lemma.bin' for your language. 
+This will produce a lemma for each token, now you can use the action type 'lemmas'
+
+## Action Scoring
+
+The scoring mechanism scores each action config to look for the best scored.
+If some token fails the comparison the action scores 0.
+If all action scored 0 none is executed.
+
+By the value of the field 'type' it defines if the template should by compared with the tokens or the lemmas.
+Please note that the captured placeholder value is extracted from the tokens not from the lemmas, but the equivalent lemmas are replaced before scoring.
+
+The action template string is a list of tokens separated by white space.
+You can use the ';' separator to provide alternative templates and the '|' to provide alternative tokens.
+
+Take in account that, as this is a token basis comparison, matching changes depending on the tokenizer you are using as they can produce different tokens for the same text.
+
+### Interpreter Configuration
+
+| Config   |  Group  |  Type  |   Default  | Description  |
+|----------|----------|----------|----------|------------------------------|
+| lowerText | nlp | boolean | false | Convert the input text to lowercase before processing |
+| caseSensitive | nlp | boolean | false | Enable case sensitivity, do not apply to dictionaries and models, do not apply to the 'itemLabel' placeholder |
+| useSimpleTokenizer | nlp | boolean | false | Prefer simple tokenizer over white space tokenizer |
+| optionalLanguageTags | nlp | text |  | Comma separated POS language tags that will be optional when comparing |
+| fallbackHLI | nlp | text |  | Human language interpreter to use if command is unhandled |
+| commandSentMessage | messages | text | Done | Message for successful command |
+| unhandledMessage | messages | text | I can not do that | Message for unsuccessful action |
+| failMessage | messages | text | There was an error | Message for error during processing |
+
+## Examples:
+
+### String type action configs example:
+
+This example contain the files to add actions for open an android application and check what is open.
+These actions will target all String items with the tag 'launch_android_app'.
+
+These are the files needed:
+
+#### File '<OPENHAB_USERDATA>/actiontemplatehli/type_actions/String.json'
+
+```json
+[
+    {
+        "template": "launch|open $app on $itemLabel",
+        "value": "$app",
+	      "type": "tokens",
+        "requiredTags": ["launch_android_app"],
+        "placeholders": [
+            {
+                "label": "app",
+                "ner": "applications",
+                "pos": "application_to_package"
+            }
+        ]
+    },
+    {
+        "template": "what app|application is open on $itemLabel;what app|application is on $itemLabel",
+        "read": true,
+        "value": "the open app is $state",
+        "emptyValue": "no app open on $itemLabel",
+	      "type": "tokens",
+        "requiredTags": ["launch_android_app"],
+        "placeholders": [
+            {
+                "label": "state",
+                "pos": "package_to_application"
+            }
+        ]
+    }
+]
+```
+
+#### File '<OPENHAB_USERDATA>/actiontemplatehli/ner/applications.xml'
+
+```xml
+<?xml version="1.0" encoding="UTF-8"?>
+<dictionary case_sensitive="false">
+  <entry>
+      <token>youtube</token>
+  </entry>
+  <entry>
+      <token>jellyfin</token>
+  </entry>
+  <entry>
+      <token>amazon</token>
+      <token>video</token>
+  </entry>
+  <entry>
+      <token>netflix</token>
+  </entry>
+</dictionary>
+```
+
+#### File '<OPENHAB_USERDATA>/actiontemplatehli/pos/package_to_application.xml'
+
+```xml
+<?xml version="1.0" encoding="UTF-8"?>
+<dictionary>
+  <entry tags="youtube">
+    <token>com.google.android.youtube</token>
+  </entry>
+  <entry tags="netflix">
+    <token>com.netflix.ninja</token>
+  </entry>
+  <entry tags="jellyfin">
+    <token>org.jellyfin.androidtv</token>
+  </entry>
+  <entry tags="amazon__video"> // note the __
+    <token>com.amazon.amazonvideo.livingroom</token>
+  </entry>
+</dictionary>
+```
+
+#### File '<OPENHAB_USERDATA>/actiontemplatehli/pos/application_to_package.xml'
+
+```xml
+  <?xml version="1.0" encoding="UTF-8"?>
+  <dictionary>
+    <entry tags="com.google.android.youtube">
+      <token>youtube</token>
+    </entry>
+    <entry tags="com.netflix.ninja">
+      <token>netflix</token>
+    </entry>
+    <entry tags="org.jellyfin.androidtv">
+      <token>jellyfin</token>
+    </entry>
+    <entry tags="com.amazon.amazonvideo.livingroom">
+      <token>amazon__video</token> // note the __
+    </entry>
+  </dictionary>
+```
+
+### Switch type action configs example:
+
+This example contain the files to add actions for turn a switch on or off.
+These actions will target all Switch items.
+
+#### File '<OPENHAB_USERDATA>/actiontemplatehli/type_actions/Switch.json'
+
+```json
+[
+  {
+    "template": "$onOff $itemLabel",
+    "value": "$onOff",
+    "type": "tokens",
+    "placeholders": [
+      {
+        "label": "onOff",
+        "nerValues": [
+          "turn on",
+          "turn off"
+        ],
+        "posValues": {
+          "turn__on": "ON", // note the __
+          "turn__off": "OFF"
+        }
+      }
+    ]
+  },
+  {
+    "template": "how be the $itemLabel",
+    "read": true,
+    "type": "lemmas",
+    "value": "$itemLabel is $state"
+  }
+]
+```
+
+### Dynamic placeholder example:
+
+This example contain the item metadata to add an action that uses the dynamic placeholder.
+Add a custom metadata 'actiontemplatehli' to a String item with the following:
+
+```yaml
+value: ""
+config:
+  placeholders:
+    - label: contact
+      nerValues:
+        - Andrea
+        - Jacob
+        - Raquel
+  ruleMode: true
+  template: send message $* to $contact
+  type: tokens
+  value: $contact:$*
+```
+
+# Note From The Original Author
+
+This human language interpreter takes advantage of just a subset of the capabilities offered by OpenNLP.
+
+Don't hesitate on creating an issue proposing how to take advance of capabilities or opening a PR with any improvements.
+
+And feel free to ping me if you need something, I'm not an expert on this field but I will try to help, @GiviMAD.
--- a/bundles/org.openhab.voice.actiontemplatehli/README.md
+++ b/bundles/org.openhab.voice.actiontemplatehli/README.md
@@ -101,7 +101,7 @@ That's the reason why the whitespace character should be replaced by '__' in the
 
 ### Target members:
 
-When the target of an action is group item, you can target its members instead.
+When the target of an action is a group item, you can target its members instead.
 
 You can use the following fields:
 
--- a/bundles/org.openhab.voice.actiontemplatehli/README.md
+++ b/bundles/org.openhab.voice.actiontemplatehli/README.md
@@ -154,7 +154,7 @@ To do so, it needs the tokens and optionally the POS tags and the lemmas.
 
 ### Tokenizer
 
-You can provide a custom model at '<OPENHAB_USERDATA>/actiontemplatehli/token.bin', otherwise it will use the build in simple tokenizer or whitespace tokenizer (configurable).
+You can provide a custom model at '<OPENHAB_USERDATA>/actiontemplatehli/token.bin', otherwise it will use the build in [simple tokenizer](https://opennlp.apache.org/docs/1.9.4/manual/opennlp.html#tools.cli.tokenizer.SimpleTokenizer) or whitespace tokenizer (configurable).
 
 Tokenizing the text is enough to use the action type 'tokens' as tokens are the only ones required for scoring (but the option 'optionalLanguageTags' will not take effect unless you have the POS language tags).
 
--- a/bundles/org.openhab.voice.actiontemplatehli/README.md
+++ b/bundles/org.openhab.voice.actiontemplatehli/README.md
@@ -92,26 +92,7 @@ The placeholder symbol replaces the text tokens matched using NER (before scorin
 As a summary, using the placeholders you can configure how parts of the speech are converted into valid item values and backward.
 The examples at the end of the document can help you to see it clearer.
 
-There are some special placeholders, described in their own sections bellow.
-
-### POS Transformation
-
-POS is a technique which produces tags for each token, here though we are going to use it to match a group of words with a value so we should transform those words to a single token.
-That's the reason why the whitespace character should be replaced by '__' in the POS dictionaries and static values, you can see some examples bellow.
-
-### Target members:
-
-When the target of an action is a group item, you can target its members instead.
-
-You can use the following fields:
-
-* itemName: name of the item member to target. If present the other fields are ignored.
-* itemType: type of the item members to target.
-* requiredTags: allow to restrict the members targeted by tags when matching by type.
-* recursive: when matching by itemType, look for group members in a recursive way, default true.
-* mergeState: on a read action when matching by itemType, merge the item states by performing an AND operation, only allowed for 'Switch' and 'Contact' item types, default false.
-
-## Special Placeholders
+There are some special placeholders:
 
 ### The 'itemLabel' Placeholder
 
@@ -156,6 +137,23 @@ The action with template "play $musicAuthor on living room" will be executed.
 
 The sentence "play beethoven on living room" will score 4 when compared with the template containing the dynamic placeholder and 0 when compared with the one without it.
 The action with template "play $* on living room" will be executed.
+
+### POS Transformation
+
+POS is a technique which produces tags for each token, here though we are going to use it to match a group of words with a value so we should transform those words to a single token.
+That's the reason why the whitespace character should be replaced by '__' in the POS dictionaries and static values, you can see some examples bellow.
+
+### Target members:
+
+When the target of an action is a group item, you can target its members instead.
+
+You can use the following fields:
+
+* itemName: name of the item member to target. If present the other fields are ignored.
+* itemType: type of the item members to target.
+* requiredTags: allow to restrict the members targeted by tags when matching by type.
+* recursive: when matching by itemType, look for group members in a recursive way, default true.
+* mergeState: on a read action when matching by itemType, merge the item states by performing an AND operation, only allowed for 'Switch' and 'Contact' item types, default false.
 
 ## Text Preprocessing
 
--- a/bundles/org.openhab.voice.actiontemplatehli/README.md
+++ b/bundles/org.openhab.voice.actiontemplatehli/README.md
@@ -334,6 +334,27 @@ These actions will target all Switch items.
 ]
 ```
 
+### Switch item action configs example:
+
+This example contains the item metadata to add an action to an item with type 'Switch''.
+Add a custom metadata 'actiontemplatehli' to the 'Switch' item with the following:
+
+```yaml
+value: ""
+config:
+  placeholders:
+    - label: onOff
+      nerValues:
+        - turn on
+        - turn off
+      posValues:
+        turn__on: ON
+        turn__off: OFF
+  template: $onOff $itemLabel
+  type: tokens
+  value: $onOff
+```
+
 ### Dynamic placeholder, sending a message example:
 
 This example contains the item metadata to add an action that uses the dynamic placeholder.
--- a/bundles/org.openhab.voice.actiontemplatehli/README.md
+++ b/bundles/org.openhab.voice.actiontemplatehli/README.md
@@ -126,6 +126,16 @@ When it's present, the 'itemLabel' placeholder will take the value of the target
 
 It's used to access the value on the read actions, you can configure a POS transformation for it.
 
+### The 'itemOption' Placeholder
+
+This placeholder is available for both write and read actions and doesn't need to be configured.
+
+When read is false, the 'itemOption' placeholder will be computed from the item command description options, or from the state description options if the command description options are not present.
+
+When read is true, the 'itemOption' placeholder will be computed from the item state description options.
+
+Note that, when targeting multiple group members, 'ner' (value matching) is done by merging all available member options but 'pos' (value transformation) is done using just the target member item options.
+
 ### The '*' Placeholder (the dynamic placeholder)
 
 The dynamic placeholder is designed to capture free text. The captured value is exposed to the value under the symbol '$*' as other placeholders.
--- a/bundles/org.openhab.voice.actiontemplatehli/README.md
+++ b/bundles/org.openhab.voice.actiontemplatehli/README.md
@@ -50,8 +50,8 @@ Take in account that, as this is a token basis comparison, matching depends on t
 
 ## Action Template Options:
 
-The location were action configurations are placed changes whetter you are targeting an item or many, so take a look to the 'Action Template Target' to understand were to put this configurations.
-Also you can check the paths indicated on the examples at the end.
+The location where action configurations are placed changes whether you are targeting an item or many, so take a look to the 'Action Template Target' to understand where to put those configurations.
+Also, you can check the paths indicated on the examples at the end.
 
 Actions can read the state from an item or send a command to it.
 This is defined by the boolean field read which is 'false' by default.
@@ -61,7 +61,7 @@ When read is false:
 * template: action template. (Required)
 * value: value to be sent. It can be used to capture the transformed placeholder values. (Required unless the target item type is String, in which case silent mode is assumed to be true and the whole text is passed to the item).
 * type: action template type, either "tokens" or "lemmas".
-* requiredTags: allow to restrict the items targeted by its tags.
+* requiredTags: allow to restrict the items targeted by its tags by ignoring items not having all these tags.
 * placeholders: defined placeholders that can be used on the template and replaced on the value.
 * silent: boolean used to avoid confirmation message.
 * targetMembers: when targeting a Group item, can be used to send the command to its member items instead.
@@ -72,7 +72,7 @@ When read is true:
 * value: read template, can use the placeholders symbols $itemLabel and $state.
 * emptyValue: An alternative template. Is used when the state value is empty or NULL after the post transformation. The $itemLabel is available.
 * type: action template type.
-* requiredTags: allow to restrict the items targeted by its tags.
+* requiredTags: allow to restrict the items targeted by its tags by ignoring items not having all these tags.
 * placeholders: only the placeholder with label state will be used, to process its POS transformation on the state.
 * targetMembers: when targeting a Group item, can be used to access the state of one of its members. In case of multiple matches, a warning is shown and the first one is used.
 
@@ -107,7 +107,7 @@ You can use the following fields:
 
 * itemName: name of the item member to target. If present the other fields are ignored.
 * itemType: type of the item members to target.
-* requiredTags: allow to restrict the members targeted by its tags when matching by type.
+* requiredTags: allow to restrict the members targeted by tags when matching by type.
 * recursive: when matching by itemType, look for group members in a recursive way, default true.
 * mergeState: on a read action when matching by itemType, merge the item states by performing an AND operation, only allowed for 'Switch' and 'Contact' item types, default false.
 
@@ -164,7 +164,12 @@ To do so, it needs the tokens and optionally the POS tags and the lemmas.
 
 ### Tokenizer
 
-You can provide a custom model at '<OPENHAB_USERDATA>/actiontemplatehli/token.bin', otherwise it will use the build in [simple tokenizer](https://opennlp.apache.org/docs/1.9.4/manual/opennlp.html#tools.cli.tokenizer.SimpleTokenizer) or whitespace tokenizer (configurable).
+You can provide a custom model at '<OPENHAB_USERDATA>/actiontemplatehli/token.bin', otherwise it will use the built-in simple tokenizer or whitespace tokenizer (configurable).
+
+Here you have an example of the built-in ones:
+
+* Using the white space tokenizer "What time is it?" produces the tokens "what" "time" "is" "it?"
+* Using the simple tokenizer "What time is it?" produces the tokens "what" "time" "is" "it" "?"
 
 Tokenizing the text is enough to use the action type 'tokens' as tokens are the only ones required for scoring (but the option 'optionalLanguageTags' will not take effect unless you have the POS language tags).
 
@@ -191,18 +196,18 @@ This will produce a lemma for each token, then you can use the action type 'lemm
 
 Note that you need the POS language tags for your language, the ones covered on the previous section, for the lemmatizer to work.
 
-### Interpreter Configuration
+## Interpreter Configuration
 
-| Config   |  Group  |  Type  |   Default  | Description  |
-|----------|----------|----------|----------|------------------------------|
-| lowerText | nlp | boolean | false | Convert the input text to lowercase before processing |
-| caseSensitive | nlp | boolean | false | Enable case sensitivity, do not apply to dictionaries and models, do not apply to the 'itemLabel' placeholder |
-| useSimpleTokenizer | nlp | boolean | false | Prefer simple tokenizer over white space tokenizer |
-| optionalLanguageTags | nlp | text |  | Comma separated POS language tags that will be optional when comparing |
-| fallbackHLI | nlp | text |  | Human language interpreter to use if command is unhandled |
-| commandSentMessage | messages | text | Done | Message for successful command |
-| unhandledMessage | messages | text | I can not do that | Message for unsuccessful action |
-| failMessage | messages | text | There was an error | Message for error during processing |
+| Config                  |  Group   |  Type   |   Default            | Description                                                                                                   |
+|-------------------------|----------|---------|----------------------|---------------------------------------------------------------------------------------------------------------|
+| lowerText               | nlp      | boolean | false                | Convert the input text to lowercase before processing                                                         |
+| caseSensitive           | nlp      | boolean | false                | Enable case sensitivity, do not apply to dictionaries and models, do not apply to the 'itemLabel' placeholder |
+| useSimpleTokenizer      | nlp      | boolean | false                | Prefer simple tokenizer over white space tokenizer                                                            |
+| detokenizeOptimization  | nlp      | boolean | true                 | Enables build-in detokenization based on original text, otherwise string join by space is used                |
+| optionalLanguageTags    | nlp      | text    |                      | Comma separated POS language tags that will be optional when comparing                                        |
+| commandSentMessage      | messages | text    | Done                 | Message for successful command                                                                                |
+| unhandledMessage        | messages | text    | I can not do that    | Message for unsuccessful action                                                                               |
+| failureMessage          | messages | text    | There was an error   | Message for error during processing                                                                           |
 
 ## Examples:
 
@@ -384,11 +389,3 @@ config:
   type: tokens
   value: $contact:$*
 ```
-
-# Note From The Original Author
-
-This human language interpreter takes advantage of just a subset of the capabilities offered by OpenNLP.
-
-Don't hesitate on creating an issue proposing how to take advance of capabilities or opening a PR with any improvements.
-
-And feel free to ping me if you need something, I'm not an expert on this field but I will try to help, @GiviMAD.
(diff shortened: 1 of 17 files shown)


GiviMAD:
@lolodomo I think the initial version is ready in case you want to give it a try, for me it work pretty well using Spanish.
I will add a post on the forum tomorrow to see if someone is interested on going testing it.
I just need to add some rule actions to allow to take advance of the interpreter features from there.
Also I will try to add some tests before removing the WIP.
Also I think this should be changed for the non response mode to work correctly https://github.com/openhab/openhab-core/pull/2780

GiviMAD:
I think is not possible to expose actions from this kind of add-on, I will let that for the future.

GiviMAD:
when you have time, it's ready for review!

GiviMAD:
@lolodomo, can you remove the work in progress label? Thank you in advance.

GiviMAD:
I have updated how getSupportedLocale works, and renamed the 'ruleMode' option to 'silence'.

Let me know whenever you have time to start reviewing this one, and thanks for the effort.

bundles/org.openhab.voice.actiontemplatehli/src/main/resources/OH-INF/config/config.xml
```diff
@@ -0,0 +1,55 @@
...
+		</parameter>
+		<parameter name="optionalLanguageTags" type="text" groupName="nlp">
+			<label>Optional Language Tags</label>
+			<description>Comma separated POS language tags that will be optional when comparing</description>
+		</parameter>
+		<parameter name="fallbackHLI" type="text" groupName="nlp">
+			<label>Fallback Human Language Interpreter</label>
+			<description>Human language interpreter to use if text is unhandled.</description>
+		</parameter>
+		<parameter name="commandSentMessage" type="text" groupName="messages" required="true">
```
lolodomo, line 39: No need to mark it as required as you define a default value.

bundles/org.openhab.voice.actiontemplatehli/src/main/resources/OH-INF/config/config.xml
```diff
@@ -0,0 +1,55 @@
...
+			<label>Command Sent Message</label>
+			<description>Message for successful command</description>
+			<default>Done</default>
+		</parameter>
+		<parameter name="unhandledMessage" type="text" groupName="messages" required="true">
+			<label>Unhandled Message</label>
+			<description>Message for unsuccessful processing</description>
+			<default>I can not do that</default>
+		</parameter>
+		<parameter name="failMessage" type="text" groupName="messages" required="true">
```
lolodomo, line 49: No need to mark it as required as you define a default value.

bundles/org.openhab.voice.actiontemplatehli/README.md
```diff
@@ -0,0 +1,336 @@
...
+These are the files needed:
+
+#### File '<OPENHAB_USERDATA>/actiontemplatehli/type_actions/String.json'
+
+```json
+[
+    {
+        "template": "launch|open $app on $itemLabel",
+        "value": "$app",
+	      "type": "tokens",
```
lolodomo, line 186: That is minor, but could you align correctly please (in all examples)

bundles/org.openhab.voice.actiontemplatehli/README.md
```diff
@@ -0,0 +1,336 @@
...
+| fallbackHLI | nlp | text |  | Human language interpreter to use if command is unhandled |
+| commandSentMessage | messages | text | Done | Message for successful command |
+| unhandledMessage | messages | text | I can not do that | Message for unsuccessful action |
+| failMessage | messages | text | There was an error | Message for error during processing |
+
+## Examples:
+
+### String type action configs example:
+
+This example contain the files to add actions for open an android application and check what is open.
```
lolodomo, line 174: This example **contains** the files to add actions for **opening** an android application and **checking** what **application** is **opened**.

lolodomo (changes requested):
Review part 1

lolodomo:
I mainly reviewed the README and stuff relative to service configuration.
Looks promising but the documentation is really technical, not easy to understand. Hopefully, you proposed several examples.
As you mentioned in the documentation, there is nothiong provided out of the box. Maybe we should provide the more common action templates ?

bundles/org.openhab.voice.actiontemplatehli/README.md
```diff
@@ -0,0 +1,336 @@
...
+
+## Action Templates Configuration:
+
+Actions can read from an item or send a command to it.
+This is defined by the boolean field read which is 'false' by default.
+
+When read is false:
+
+* template: action template. (Required)
+* value: value to by send. It can be used to capture the transformed placeholder values. (Required unless if the target item type is String in that case silence mode is assumed true and the whole text is passed to item).
```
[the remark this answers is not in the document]
GiviMAD: I've also renamed the "silence" option to "silent", as it seems more appropriate.

bundles/org.openhab.voice.actiontemplatehli/README.md
```diff
@@ -0,0 +1,336 @@
...
+### Tokenizer
+
+You can provide a custom model at '<OPENHAB_USERDATA>/actiontemplatehli/token.bin', otherwise it will use the build in simple tokenizer or whitespace tokenizer (configurable).
+
+Tokenizing the text is enough to use the action type 'tokens' as tokens are the only ones required for scoring (but the option 'optionalLanguageTags' will not take effect unless you have the POS language tags).
+
+### POSTagger (language tags)
+
+You need to provide a model for POS tagging at '<OPENHAB_USERDATA>/actiontemplatehli/pos.bin' for your language. 
+This will produce a language tag for each token, that can be used in 'optionalLanguageTags' to make some optional for scoring.
```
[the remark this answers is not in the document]
GiviMAD: Small example added.

bundles/org.openhab.voice.actiontemplatehli/README.md
```diff
@@ -0,0 +1,336 @@
...
+
+* You can link an action to a specific item by adding it the custom metadata namespace 'actiontemplatehli'.
+* You can link an action to all items of a type by providing the file '<OPENHAB_USERDATA>/actiontemplatehli/type_actions/<ITEM_TYPE_NAME>.json' (here you can restrict each action by item tags).
+
+Two important notes:
+
+* Actions linked to items have prevalence over actions linked to a whole item type.
+* On actions linked to an item type the itemLabel placeholder will always be applied (explained bellow), if there are multiple item label detected in the text this part will be aborted.
+
+## Action Templates Configuration:
```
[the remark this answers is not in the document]
GiviMAD: Yes, it's explained at the  'Action Templates Target' section. I see it's not quite clear so I have added a comment highlighting it.

bundles/org.openhab.voice.actiontemplatehli/README.md
```diff
@@ -0,0 +1,336 @@
...
+You need to provide a model for POS tagging at '<OPENHAB_USERDATA>/actiontemplatehli/pos.bin' for your language. 
+This will produce a language tag for each token, that can be used in 'optionalLanguageTags' to make some optional for scoring.
+Note that you need the correct language tags for the lemmatizer to work.
+
+### Lemmatizer
+
+You need to provide a model for the lemmatizer at '<OPENHAB_USERDATA>/actiontemplatehli/lemma.bin' for your language. 
+This will produce a lemma for each token, now you can use the action type 'lemmas'
+
+## Action Scoring
```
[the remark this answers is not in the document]
GiviMAD: I have renamed the "Action Templates Configuration" to "Action Templates Options", as the whole configuration gets explained through the different sections seems a better name.

bundles/org.openhab.voice.actiontemplatehli/src/main/java/org/openhab/voice/actiontemplatehli/internal/ActionTemplateInterpreterConfiguration.java
```diff
@@ -0,0 +1,60 @@
...
+     */
+    public String optionalLanguageTags = "";
+    /**
+     * Prefer simple tokenizer over white space tokenizer
+     */
+    public boolean useSimpleTokenizer = false;
+    /**
+     * Enables build-in detokenization based on original text, other way string join by space is used
+     */
+    public boolean detokenizeOptimization = true;
```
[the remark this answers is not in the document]
GiviMAD: I had forgotten to add it to the configuration.

bundles/org.openhab.voice.actiontemplatehli/README.md
```diff
@@ -0,0 +1,336 @@
...
+
+### Interpreter Configuration
+
+| Config   |  Group  |  Type  |   Default  | Description  |
+|----------|----------|----------|----------|------------------------------|
+| lowerText | nlp | boolean | false | Convert the input text to lowercase before processing |
+| caseSensitive | nlp | boolean | false | Enable case sensitivity, do not apply to dictionaries and models, do not apply to the 'itemLabel' placeholder |
+| useSimpleTokenizer | nlp | boolean | false | Prefer simple tokenizer over white space tokenizer |
+| optionalLanguageTags | nlp | text |  | Comma separated POS language tags that will be optional when comparing |
+| fallbackHLI | nlp | text |  | Human language interpreter to use if command is unhandled |
```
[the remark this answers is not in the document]
GiviMAD: I'll open a PR in the core to reflect that. In the meantime is it ok if I add a comment and let the feature as it is?
Something like "(the option will be remove at short time as this behavior will be handled by the core)" ?

bundles/org.openhab.voice.actiontemplatehli/src/main/resources/OH-INF/config/config.xml
```diff
@@ -0,0 +1,55 @@
...
+			<default>false</default>
+		</parameter>
+		<parameter name="optionalLanguageTags" type="text" groupName="nlp">
+			<label>Optional Language Tags</label>
+			<description>Comma separated POS language tags that will be optional when comparing</description>
+		</parameter>
+		<parameter name="fallbackHLI" type="text" groupName="nlp">
+			<label>Fallback Human Language Interpreter</label>
+			<description>Human language interpreter to use if text is unhandled.</description>
+		</parameter>
```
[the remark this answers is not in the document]
GiviMAD: Same comment as above, can it persist for now with a comment that indicates it will be removed shortly?

bundles/org.openhab.voice.actiontemplatehli/README.md
```diff
@@ -0,0 +1,336 @@
...
+ * pos: name for a file under the pos folder (<OPENHAB_USERDATA>/actiontemplatehli/pos), first it will look for a <pos>.bin model and then for a <pos>.xml dictionary (prevalence over 'posValues').
+
+The placeholder symbol replaces the text tokens matched using NER (before scoring the actions) and the captured value could be transformed using POS and will be accessible to the value under its symbol.
+As a summary, using the placeholders you can configure how parts of the speech are converted into valid item values and backward.
+
+There are some special placeholders, described in another section.
+
+### POS Transformation
+
+POS is a technique which produces tags for each token, here though we are going to use it to match a group of words with a value so we should transform those words to a single token.
```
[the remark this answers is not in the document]
GiviMAD: No it's "should", as said at the beginning of the statement "POS is a technique which produces tags for each token" so we need to do that in order to use it for this.

bundles/org.openhab.voice.actiontemplatehli/README.md
```diff
@@ -0,0 +1,336 @@
...
+
+Take in account that, as this is a token basis comparison, matching changes depending on the tokenizer you are using as they can produce different tokens for the same text.
+
+### Interpreter Configuration
+
+| Config   |  Group  |  Type  |   Default  | Description  |
+|----------|----------|----------|----------|------------------------------|
+| lowerText | nlp | boolean | false | Convert the input text to lowercase before processing |
+| caseSensitive | nlp | boolean | false | Enable case sensitivity, do not apply to dictionaries and models, do not apply to the 'itemLabel' placeholder |
+| useSimpleTokenizer | nlp | boolean | false | Prefer simple tokenizer over white space tokenizer |
```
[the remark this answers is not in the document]
lolodomo: If you like, 
Or just copy / paste the explanation in case it can be explained very shortly.

bundles/org.openhab.voice.actiontemplatehli/README.md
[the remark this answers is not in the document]
lolodomo: No, it is better to remove it there and you can in parallel propose a clean and full solution in the core framework, a solution that can work whatever the HLI.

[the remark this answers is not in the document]
lolodomo: Open an issue in github so that it can be discussed.

GiviMAD:
@lolodomo I have added this issue https://github.com/openhab/openhab-core/issues/2904 to ask for a solution to use multiple interpreters.
I will remove the fallback functionality.

bundles/org.openhab.voice.actiontemplatehli/README.md
```diff
@@ -0,0 +1,336 @@
...
+Note that the dynamic placeholder does not score.
+
+## Text Preprocessing
+
+The interpreter needs to match the input text with a target item and action config, to know what to do.
+To do so, it need the tokens and optionally the POS tags and the lemmas.
+
+### Tokenizer
+
+You can provide a custom model at '<OPENHAB_USERDATA>/actiontemplatehli/token.bin', otherwise it will use the build in simple tokenizer or whitespace tokenizer (configurable).
```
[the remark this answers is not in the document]
GiviMAD: Added link to OpenNLP docs on the simple tokenizer.

bundles/org.openhab.voice.actiontemplatehli/README.md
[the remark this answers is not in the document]
GiviMAD: Added link to OpenNLP docs on the simple tokenizer.

bundles/org.openhab.voice.actiontemplatehli/README.md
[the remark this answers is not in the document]
GiviMAD: Removed

bundles/org.openhab.voice.actiontemplatehli/src/main/resources/OH-INF/config/config.xml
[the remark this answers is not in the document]
GiviMAD: Removed

bundles/org.openhab.voice.actiontemplatehli/README.md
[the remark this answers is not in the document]
GiviMAD: Is it more clear if I replace "we should" by "we need to"?

bundles/org.openhab.voice.actiontemplatehli/README.md
```diff
@@ -0,0 +1,336 @@
...
+    "type": "lemmas",
+    "value": "$itemLabel is $state"
+  }
+]
+```
+
+### Dynamic placeholder example:
+
+This example contain the item metadata to add an action that uses the dynamic placeholder.
+Add a custom metadata 'actiontemplatehli' to a String item with the following:
```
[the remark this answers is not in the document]
GiviMAD: Oh, I didn't understood the comment, I read it wrong. Yes, it makes sense, I going to add it.

GiviMAD:
> I mainly reviewed the README and stuff relative to service configuration.
> Looks promising but the documentation is really technical, not easy to understand. Hopefully, you proposed several examples.
> As you mentioned in the documentation, there is nothiong provided out of the box. Maybe we should provide the more common action templates ?

I think there is just one left to solve, the one in the POS transformation section

I can add an example similar to the one described in the dynamic placeholder section to explain the score (added as part of the review). 
Other than that I have no more ideas at the moment. Let me know if any good use case for another example came to your mind.

GiviMAD:
@lolodomo I have realize that I can add another placeholder that work based on the target item state description and the command description, will make the add-on simpler to use and more connected to openHAB. 

Two questions:

* Is it ok to added it now, or is better wait to finish this PR? It will relay on the current implementation so no too much change is needed.
* By reading the docs I understand that to get all the possible values for an item I need to merge the options for its stateDescrition and commandDescription. Is this assumption correct? I don't totally understand the difference between them.

bundles/org.openhab.voice.actiontemplatehli/README.md
```diff
@@ -0,0 +1,384 @@
...
+
+The action template string is a list of tokens separated by white space.
+You can use the ';' separator to provide alternative templates and the '|' to provide alternative tokens.
+Here is an example of string: "what app|application is open on $itemLabel;what app|application is on $itemLabel".
+
+Take in account that, as this is a token basis comparison, matching depends on the tokenizer you are using as they can produce different tokens for the same text.
+
+## Action Template Options:
+
+The location were action configurations are placed changes whetter you are targeting an item or many, so take a look to the 'Action Template Target' to understand were to put this configurations.
```
lolodomo, line 53: The location **where** configurations ... **whether** you are ... to understand **where** to put ...

bundles/org.openhab.voice.actiontemplatehli/README.md
```diff
@@ -0,0 +1,384 @@
...
+
+Actions can read the state from an item or send a command to it.
+This is defined by the boolean field read which is 'false' by default.
+
+When read is false:
+
+* template: action template. (Required)
+* value: value to be sent. It can be used to capture the transformed placeholder values. (Required unless the target item type is String, in which case silent mode is assumed to be true and the whole text is passed to the item).
+* type: action template type, either "tokens" or "lemmas".
+* requiredTags: allow to restrict the items targeted by its tags.
```
lolodomo, line 64: by ignoring items not having all these tags.

bundles/org.openhab.voice.actiontemplatehli/README.md
```diff
@@ -0,0 +1,384 @@
...
+* silent: boolean used to avoid confirmation message.
+* targetMembers: when targeting a Group item, can be used to send the command to its member items instead.
+
+When read is true:
+
+* template: action template. (Required)
+* value: read template, can use the placeholders symbols $itemLabel and $state.
+* emptyValue: An alternative template. Is used when the state value is empty or NULL after the post transformation. The $itemLabel is available.
+* type: action template type.
+* requiredTags: allow to restrict the items targeted by its tags.
```
lolodomo, line 75: by ignoring items not having all these tags.

bundles/org.openhab.voice.actiontemplatehli/README.md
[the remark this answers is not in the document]
lolodomo: Keep it as it is.

bundles/org.openhab.voice.actiontemplatehli/README.md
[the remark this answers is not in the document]
lolodomo: Your link does not help !
I found a little more information here: https://www.tutorialspoint.com/opennlp/opennlp_tokenization.htm
"SimpleTokenizer − This class tokenizes the given raw text using character classes."
There is an example showing the difference with the two tokenizers. Maybe you can include a simplified example.
"What time is it?" == white space tokenizer => tokens "what" "time" "is" "it?"
"What time is it?" == simple tokenizer => tokens "what" "time" "is" "it" "?"

bundles/org.openhab.voice.actiontemplatehli/src/main/java/org/openhab/voice/actiontemplatehli/internal/ActionTemplateInterpreter.java
```diff
@@ -0,0 +1,1159 @@
...
+        nlpItemMaps = null;
+    }
+
+    private void reloadConfigs() {
+        optionalLanguageTags = Arrays.stream(this.config.optionalLanguageTags.split(",")).filter(i -> !i.isEmpty())
+                .collect(Collectors.toList());
+        cacheMap.invalidateAll();
+        try {
+            tokenizer = getTokenizer();
+        } catch (IOException e) {
```
lolodomo, line 1070: I think it will be better to catch the exception directly in method getTokenizer.

bundles/org.openhab.voice.actiontemplatehli/src/main/java/org/openhab/voice/actiontemplatehli/internal/ActionTemplateInterpreter.java
```diff
@@ -0,0 +1,1159 @@
...
+
+    @Override
+    public Set<String> getSupportedGrammarFormats() {
+        return Set.of();
+    }
+
+    @Override
+    public String interpret(Locale locale, String words) throws InterpretationException {
+        if (words.isEmpty()) {
+            new InterpretationException(config.unhandledMessage);
```
lolodomo, line 205: I think "trow" is missing

bundles/org.openhab.voice.actiontemplatehli/src/main/java/org/openhab/voice/actiontemplatehli/internal/ActionTemplateInterpreter.java
```diff
@@ -0,0 +1,1159 @@
...
+                return sendItemCommand(result.targetItem, words, result.actionConfig, result.placeholderValues);
+            } else {
+                return readItemState(result.targetItem, result.actionConfig);
+            }
+        } else {
+            throw new InterpretationException(config.unhandledMessage);
+        }
+    }
+
+    public NLPInfo getNLPInfo(String text) throws IOException {
```
lolodomo, line 247: Method could be private ?

bundles/org.openhab.voice.actiontemplatehli/src/main/java/org/openhab/voice/actiontemplatehli/internal/ActionTemplateInterpreter.java
```diff
@@ -0,0 +1,1159 @@
...
+        if (words.isEmpty()) {
+            new InterpretationException(config.unhandledMessage);
+        }
+        try {
+            if (config.lowerText) {
+                words = words.toLowerCase(locale);
+            }
+            var info = getNLPInfo(words);
+            if (info.tokens.length == 0) {
+                logger.warn("no tokens produced; aborting");
```
lolodomo, line 213: DEBUG level should be sufficient

bundles/org.openhab.voice.actiontemplatehli/src/main/java/org/openhab/voice/actiontemplatehli/internal/ActionTemplateInterpreter.java
```diff
@@ -0,0 +1,1159 @@
...
+            var lemmas = languageLemmatize(tokens, tags);
+            nlpInfo = new NLPInfo(tokens, lemmas, tags);
+        } else {
+            logger.debug("resolved from cache");
+        }
+        logger.debug("tokens: {}", List.of(nlpInfo.tokens));
+        logger.debug("tags: {}", List.of(nlpInfo.tags));
+        logger.debug("lemmas: {}", List.of(nlpInfo.lemmas));
+        var finalNlpInfo = nlpInfo;
+        cacheMap.put(text, () -> finalNlpInfo);
```
lolodomo, line 262: Looks strange: You provide a supplier something that will return always the same data ?

bundles/org.openhab.voice.actiontemplatehli/src/main/java/org/openhab/voice/actiontemplatehli/internal/ActionTemplateInterpreter.java
```diff
@@ -0,0 +1,1159 @@
...
+                logger.warn("Fallback to simple tokenizer");
+                tokenizer = SimpleTokenizer.INSTANCE;
+            } else {
+                logger.warn("Fallback to white space tokenizer");
+                tokenizer = WhitespaceTokenizer.INSTANCE;
+            }
+        }
+    }
+
+    public static class NLPInfo {
```
lolodomo, line 1082: Could be a private class ?

bundles/org.openhab.voice.actiontemplatehli/src/main/java/org/openhab/voice/actiontemplatehli/internal/ActionTemplateInterpreter.java
```diff
@@ -0,0 +1,1159 @@
...
+    }
+    protected ActionTemplateInterpreterConfiguration config = new ActionTemplateInterpreterConfiguration();
+    private static final Pattern COLOR_HEX_PATTERN = Pattern.compile("^#([a-fA-F0-9]{6}|[a-fA-F0-9]{3})$");
+    private final Logger logger = LoggerFactory.getLogger(ActionTemplateInterpreter.class);
+    private final LocaleService localeService;
+    private final ItemRegistry itemRegistry;
+    private final MetadataRegistry metadataRegistry;
+    private final EventPublisher eventPublisher;
+    private Tokenizer tokenizer = WhitespaceTokenizer.INSTANCE;
+    private ExpiringCacheMap<String, NLPInfo> cacheMap = new ExpiringCacheMap<>(Duration.of(1, ChronoUnit.MINUTES));
```
lolodomo, line 128: The cache is to avoid tokenizing again the same text. But why a one minute validity ?

lolodomo (changes requested):
Reivew part 2

lolodomo:
> @lolodomo I have realize that I can add another placeholder that work based on the target item state description and command description, will make the add-on simpler to use and more connected to openHAB.
> 
> Two questions:
> 
>     * Is it ok to add it now, or is better wait to finish this PR? It will relay on the current implementation so I think no too much change is needed.
> 
>     * By reading the docs I understand that to get all the possible values for an item I need to merge the options for its stateDescrition and commandDescription. Is this assumption correct? I don't totally understand the difference between them.

You can do it in that PR, I have not yet the major part of the code.

commandDescription is for mapping the commands that are sent to an item.
stateDescrition  is for mapping the states of an item.
When "read" is true, you should consider stateDescrition .
When "read" is false (you send a command), you should consider commandDescription.

lolodomo:
Remains for me about 1000 lines to review in your main java class.

bundles/org.openhab.voice.actiontemplatehli/src/main/java/org/openhab/voice/actiontemplatehli/internal/ActionTemplateInterpreter.java
```diff
@@ -0,0 +1,1159 @@
...
+                    List<NLPPlaceholderData> currentPlaceholderValues = new ArrayList<>();
+                    var currentItem = entry.getKey();
+                    var scoreResult = getScoreWithPlaceholders(text, tokens, tags, lemmas, actionConfig, template,
+                            currentPlaceholderValues);
+                    if (scoreResult.score != 0 && scoreResult.score == matchScore) {
+                        if (targetItem == currentItem) {
+                            logger.warn("multiple alternative templates with same score '{}' item '{}'", template,
+                                    currentItem);
+                        } else {
+                            logger.warn("multiple templates with same score '{}' item '{}'", template, currentItem);
```
lolodomo, line 302: These 2 logs are not at the right level. Probably DEBUG or INFO if this is very important for the user.

bundles/org.openhab.voice.actiontemplatehli/src/main/java/org/openhab/voice/actiontemplatehli/internal/ActionTemplateInterpreter.java
```diff
@@ -0,0 +1,1159 @@
...
+        }
+        return maps.itemLabelByTokens;
+    }
+
+    private Map<Item, ActionTemplateConfiguration[]> getItemsWithActionConfigs() throws IOException {
+        var maps = this.nlpItemMaps;
+        if (maps == null) {
+            return getItemsMaps().itemsWithActionConfigs;
+        }
+        return maps.itemsWithActionConfigs;
```
lolodomo, line 945: One unique line is sufficient: return getItemsMaps().itemsWithActionConfigs;

bundles/org.openhab.voice.actiontemplatehli/src/main/java/org/openhab/voice/actiontemplatehli/internal/ActionTemplateInterpreter.java
```diff
@@ -0,0 +1,1159 @@
...
+        }
+        return tag.replaceAll("__", " ");
+    }
+
+    private Map<String[], Item> getItemsByLabelTokensMap() throws IOException {
+        var maps = this.nlpItemMaps;
+        if (maps == null) {
+            return getItemsMaps().itemLabelByTokens;
+        }
+        return maps.itemLabelByTokens;
```
lolodomo, line 937: One unique line is sufficient: return getItemsMaps().itemLabelByTokens;

bundles/org.openhab.voice.actiontemplatehli/src/main/java/org/openhab/voice/actiontemplatehli/internal/ActionTemplateInterpreter.java
```diff
@@ -0,0 +1,1159 @@
...
+    private NLPItemMaps getItemsMaps() throws IOException {
+        var itemMaps = this.nlpItemMaps;
+        if (itemMaps == null) {
+            var itemByLabelTokens = new HashMap<String[], Item>();
+            var itemsWithActionConfigs = new HashMap<Item, ActionTemplateConfiguration[]>();
+            for (Item item : itemRegistry.getAll()) {
+                var label = item.getLabel();
+                if (label != null) {
+                    var lowerLabel = label.toLowerCase();
+                    itemByLabelTokens.put(tokenizeText(lowerLabel), item);
```
lolodomo, line 957: In case several items have the same label, you keep trace of only one.
This is not a problem ?

lolodomo (changes requested):
Review part 3

lolodomo:
When I compile, I see this warning. Could you please investigate ?
```
[INFO] --- sat-plugin:0.12.0:pmd (sat-all) @ org.openhab.voice.actiontemplatehli ---
[INFO] PMD version: 6.39.0
[WARNING] There are 1 PMD processing errors:
[WARNING] D:\dev\openhab3\git\openhab-addons\bundles\org.openhab.voice.actiontemplatehli\src\main\java\org\openhab\voice\actiontemplatehli\internal\ActionTemplateInterpreter.java: PMDException: Error while parsing D:\dev\openhab3\git\openhab-addons\bundles\org.openhab.voice.actiontemplatehli\src\main\java\org\openhab\voice\actiontemplatehli\internal\ActionTemplateInterpreter.java
```

bundles/org.openhab.voice.actiontemplatehli/src/main/java/org/openhab/voice/actiontemplatehli/internal/ActionTemplateInterpreter.java
GiviMAD: I think it's important. It indicates a miss configuration because one of the actions will never be executed so one of them should be changed/removed by the user. I will change it to info.

bundles/org.openhab.voice.actiontemplatehli/src/main/java/org/openhab/voice/actiontemplatehli/internal/ActionTemplateInterpreter.java
GiviMAD: Yes, there is no support for that. I will a warning in case that happens so the user is aware, thanks for pointing it.

bundles/org.openhab.voice.actiontemplatehli/src/main/java/org/openhab/voice/actiontemplatehli/internal/ActionTemplateInterpreter.java
```diff
@@ -0,0 +1,1159 @@
...
+    protected ActionTemplateInterpreterConfiguration config = new ActionTemplateInterpreterConfiguration();
+    private static final Pattern COLOR_HEX_PATTERN = Pattern.compile("^#([a-fA-F0-9]{6}|[a-fA-F0-9]{3})$");
+    private final Logger logger = LoggerFactory.getLogger(ActionTemplateInterpreter.class);
+    private final LocaleService localeService;
+    private final ItemRegistry itemRegistry;
+    private final MetadataRegistry metadataRegistry;
+    private final EventPublisher eventPublisher;
+    private Tokenizer tokenizer = WhitespaceTokenizer.INSTANCE;
+    private ExpiringCacheMap<String, NLPInfo> cacheMap = new ExpiringCacheMap<>(Duration.of(1, ChronoUnit.MINUTES));
+    private ActionTemplateInterpreter.@Nullable NLPItemMaps nlpItemMaps;
```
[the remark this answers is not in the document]
GiviMAD: IntelliJ tells me that is not correct.

bundles/org.openhab.voice.actiontemplatehli/src/main/java/org/openhab/voice/actiontemplatehli/internal/ActionTemplateInterpreter.java
GiviMAD: True, it was due to type requirements, for me it will be enough to put the value but the ExpiringCacheMap seems to require this. Not sure if there is an alternative class.

bundles/org.openhab.voice.actiontemplatehli/src/main/java/org/openhab/voice/actiontemplatehli/internal/ActionTemplateInterpreter.java
GiviMAD: Yes sorry, it has an explanation.

My initial idea (but I decided not to do it for now) was to add some rule actions to this add-on to expose the getNLPInfo the NER and the POS to the rules so the user has like a little nlp framework to use on the rules. 

That is the reason the NLPInfo class and the related method was public (changed to private right now). The cache was intended to reuse the processing in case a rule calls getNLPInfo with the same text to do some custom processing there with the tokens or lemmas.

Maybe it has value on its own if we allow to customize the expiration time. WDYT?

bundles/org.openhab.voice.actiontemplatehli/src/main/java/org/openhab/voice/actiontemplatehli/internal/ActionTemplateInterpreter.java
GiviMAD: I realized when looking this one I had also missed to return the unhandledMessage by throwing an exception in some places, that's also changed.

bundles/org.openhab.voice.actiontemplatehli/src/main/java/org/openhab/voice/actiontemplatehli/internal/ActionTemplateInterpreter.java
GiviMAD: I've added a warning. But maybe this situation is common (it's not my case), should it rather be a warning?

bundles/org.openhab.voice.actiontemplatehli/src/main/java/org/openhab/voice/actiontemplatehli/internal/ActionTemplateInterpreter.java
GiviMAD: Looking at it again I still think it should be a warning as it indicates you have an unused action. 

But the message really looks like simple info right now, I will change it to indicate that you have something to fix. 

Is it ok if I keep it as a warning after this? or still is better to change it to info?

bundles/org.openhab.voice.actiontemplatehli/src/main/java/org/openhab/voice/actiontemplatehli/internal/ActionTemplateInterpreter.java
lolodomo: Now, that is not my comment. 
If U am correct, this parameter should be the supplier to get a fresh new value.

bundles/org.openhab.voice.actiontemplatehli/src/main/java/org/openhab/voice/actiontemplatehli/internal/ActionTemplateInterpreter.java
lolodomo: I rather understand this cache is now useless. 
So I would suggest to remove it.

bundles/org.openhab.voice.actiontemplatehli/src/main/java/org/openhab/voice/actiontemplatehli/internal/ActionTemplateInterpreter.java
lolodomo: We should just not flood the logs.
Keep a warning if you think it is really necessary for the user and this is not a regular log that will be logged at each call to the interpreter.

bundles/org.openhab.voice.actiontemplatehli/src/main/java/org/openhab/voice/actiontemplatehli/internal/ActionTemplateInterpreter.java
GiviMAD: I still not happy with this, I'll add support for item 'Synonyms' and just show a debug log for the collisions.

GiviMAD:
> When I compile, I see this warning. Could you please investigate ?
> 
> ```
> [INFO] --- sat-plugin:0.12.0:pmd (sat-all) @ org.openhab.voice.actiontemplatehli ---
> [INFO] PMD version: 6.39.0
> [WARNING] There are 1 PMD processing errors:
> [WARNING] D:\dev\openhab3\git\openhab-addons\bundles\org.openhab.voice.actiontemplatehli\src\main\java\org\openhab\voice\actiontemplatehli\internal\ActionTemplateInterpreter.java: PMDException: Error while parsing D:\dev\openhab3\git\openhab-addons\bundles\org.openhab.voice.actiontemplatehli\src\main\java\org\openhab\voice\actiontemplatehli\internal\ActionTemplateInterpreter.java
> ```

@lolodomo I think the problem is about this error https://github.com/pmd/pmd/issues/1367 
I have refactored the code and the warning has disappear, I'll push in a moment.

lolodomo:
@GiviMAD : I will restart the review soon and also test the binding.

GiviMAD:
> @GiviMAD : I will restart the review soon and also test the binding.

Thank you!

lolodomo:
@GiviMAD : I would like to try merging this PR before feature freeze (Sunday). If I finish the review during the next coming couple of days, will you have some time to consider the comments ?

GiviMAD:
Hi @lolodomo, can we try instead to review this one? https://github.com/openhab/openhab-addons/pull/12606
Still has a pending change that I will try to add this afternoon, but is just related to release the library version 1.0.0.
All the rest is ready for the review I think.
Is it ok for you?

lolodomo:
> Hi @lolodomo, can we try instead to review this one? #12606
> Still has a pending change that I will try to add this afternoon, but is just related to release the library version 1.0.0.
> All the rest is ready for the review I think.
> Is it ok for you?

Ok, done ;)

bundles/org.openhab.voice.actiontemplatehli/src/main/java/org/openhab/voice/actiontemplatehli/internal/ActionTemplateInterpreter.java
```diff
@@ -0,0 +1,1251 @@
...
+            Span[] nerSpans;
+            Map<String[], String> possibleValuesByTokensMap = null;
+            if (nerStaticValues != null) {
+                possibleValuesByTokensMap = getStringsByTokensMap(nerStaticValues);
+                nerSpans = nerValues(tokens, possibleValuesByTokensMap.keySet().toArray(String[][]::new),
+                        placeholder.label);
+            } else if (nerFile != null) {
+                nerSpans = nerWithFile(tokens, nerFile);
+            } else {
+                logger.warn("Placeholder {} could not be applied missing ner config", placeholder.label);
```
lolodomo, line 565: ... not be applied **due to** missing ner config

bundles/org.openhab.voice.actiontemplatehli/src/main/java/org/openhab/voice/actiontemplatehli/internal/ActionTemplateInterpreter.java
```diff
@@ -0,0 +1,1251 @@
...
+                placeholdersCopy.add(itemOptionPlaceholder);
+            }
+        }
+        // replace placeholder symbols
+        for (var placeholder : placeholdersCopy) {
+            var placeholderValue = placeholderValues.getOrDefault(placeholder.label, "");
+            if (!placeholderValue.isBlank()) {
+                placeholderValue = applyPOSTransformation(placeholderValue, placeholder);
+            }
+            text = text.replace(getPlaceholderSymbol(placeholder.label), placeholderValue);
```
lolodomo, line 772: As mentioned by the warning, it would be better to not assign a parameter.

bundles/org.openhab.voice.actiontemplatehli/src/main/java/org/openhab/voice/actiontemplatehli/internal/ActionTemplateInterpreter.java
```diff
@@ -0,0 +1,1251 @@
...
+                        .filter(entry -> Arrays.equals(spanTokens, entry.getKey())).findFirst();
+                if (match.isPresent()) {
+                    value = match.get().getValue();
+                } else {
+                    value = getSpanTokens(tokens, span, text);
+                }
+            } else {
+                value = getSpanTokens(tokens, span, text);
+            }
+            if (value != null) {
```
lolodomo, line 825: value cannot be null there

bundles/org.openhab.voice.actiontemplatehli/src/main/java/org/openhab/voice/actiontemplatehli/internal/ActionTemplateInterpreter.java
```diff
@@ -0,0 +1,1251 @@
...
+                replacements.add(new NLPPlaceholderData(placeholderName, value, span));
+            }
+        }
+        var spanStart = span.getStart();
+        var stream = spanStart != 0 ? Arrays.stream(Arrays.copyOfRange(tokens, 0, spanStart)) : Stream.of();
+        stream = Stream.concat(stream, Arrays.stream(new String[] { getPlaceholderSymbol(placeholderName) }));
```
lolodomo, line 831: Eclipse detects a resource leak on this line (stream not closed)

bundles/org.openhab.voice.actiontemplatehli/src/main/java/org/openhab/voice/actiontemplatehli/internal/ActionTemplateInterpreter.java
```diff
@@ -0,0 +1,1251 @@
...
+        var spanEnd = span.getEnd();
+        if (spanEnd != tokens.length) {
+            stream = Stream.concat(stream, Arrays.stream(Arrays.copyOfRange(tokens, spanEnd, tokens.length)));
```
lolodomo, line 834: Eclipse detects a resource leak on this line (stream not closed)

bundles/org.openhab.voice.actiontemplatehli/src/main/java/org/openhab/voice/actiontemplatehli/internal/ActionTemplateInterpreter.java
```diff
@@ -0,0 +1,1251 @@
...
+        return itemMaps;
+    }
+
+    private NLPTokenComparisonResult compareTokens(String[] tokens, String[] tokenTags, String[] tokensTemplate) {
+        if (tokens.length == 0 || tokensTemplate.length == 0) {
+            return NLPTokenComparisonResult.ZERO;
+        }
+        int score = 0;
+        int processedIndex = 0;
+        // avoid use tags is not available for all tokens
```
lolodomo, line 1085: if not

lolodomo (changes requested):
Final review

bundles/org.openhab.voice.actiontemplatehli/src/main/java/org/openhab/voice/actiontemplatehli/internal/configuration/ActionTemplateGroupTargets.java
```diff
@@ -0,0 +1,37 @@
...
+ */
+package org.openhab.voice.actiontemplatehli.internal.configuration;
+
+import org.eclipse.jdt.annotation.NonNullByDefault;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+/**
+ * The {@link org.openhab.voice.actiontemplatehli.internal.configuration.ActionTemplateGroupTargets} represents each
+ * configured action
```
lolodomo, line 21: Please adjust the class description.

bundles/org.openhab.voice.actiontemplatehli/src/main/java/org/openhab/voice/actiontemplatehli/internal/configuration/ActionTemplatePlaceholder.java
```diff
@@ -0,0 +1,46 @@
...
+import java.util.Map;
+
+import org.eclipse.jdt.annotation.NonNullByDefault;
+import org.eclipse.jdt.annotation.Nullable;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+/**
+ * The {@link org.openhab.voice.actiontemplatehli.internal.configuration.ActionTemplatePlaceholder} represents each
+ * configured action
```
lolodomo, line 24: Please adjust the class description.

lolodomo (changes requested):

lolodomo:
@GiviMAD : my final review only add relatively minor comments.
It is not easy to understand all the code, I sometimes forgave.

lolodomo:
I will try playing with this interpreter next weekend. 
IMHO, something important is missing. Everything is mainly based on items l abel. The problem is that, due to UI constraints on mobile phones, the item label you define is generally a short label and not the one you will use when speaking. As an example, for temperature in the bedroom, my item label could be in French "Température chambre" but when speaking I will use "température dans la chambre". This was solved with other voice assistants (google and Amazon) with the ability to define for each item the label to consider by the assistant (replacing the OH label). I believe we should have this feature also for this interpreter.

lolodomo:
Another difficulty is that for requesting the temperature in the bedroom, in French we have at least 2 natural ways to request it, either "quelle est la température dans la chambre ? " and "quelle température fait-il dans la chambre ? ". If my item label is "température dans la chambre", I don't think there is a way to make the current interpreter matching my second request.

bundles/org.openhab.voice.actiontemplatehli/src/main/java/org/openhab/voice/actiontemplatehli/internal/ActionTemplateInterpreter.java
```diff
@@ -0,0 +1,1251 @@
...
+        var itemsWithActions = getItemsWithActionConfigs();
+        Item targetItem = null;
+        ActionTemplateConfiguration targetActionConfig = null;
+        // store data to restore placeholder values
+        List<NLPPlaceholderData> placeholderValues = null;
+        // store span of dynamic placeholder, to invalidate others
+        Span dynamicSpan = null;
+        int matchScore = 0;
+        for (var entry : itemsWithActions.entrySet()) {
+            var actionConfigs = entry.getValue();
```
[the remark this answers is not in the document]
GiviMAD: This part applies to actions added to the item metadata so I think it makes no sense to check the items tags, as you are defining the action just for one item. On the other place it's checking actions linked by multiple items by its item type so I see the benefit on filter by tags there.
I just checked it, in case it was wrong but it's correctly indicated on the readme
```
requiredTags: allow to restrict the members targeted by tags when matching by type.
```

GiviMAD:
> Another difficulty is that for requesting the temperature in the bedroom, in French we have at least 2 natural ways to request it, either "quelle est la température dans la chambre ? " and "quelle température fait-il dans la chambre ? ". If my item label is "température dans la chambre", I don't think there is a way to make the current interpreter matching my second request.

For this kind of situation I think you can use an action that matches groups with tag 'location' and filter the member targets by another tag that is only on the temperature items. This way the itemLabel will be the group name (room name).

GiviMAD:
> I will try playing with this interpreter next weekend.
> IMHO, something important is missing. Everything is mainly based on items label. The problem is that, due to UI constraints on mobile phones, the item label you define is generally a short label and not the one you will use when speaking. As an example, for temperature in the bedroom, my item label could be in French "Température chambre" but when speaking I will use "température dans la chambre". This was solved with other voice assistants (google and Amazon) with the ability to define for each item the label to consider by the assistant (replacing the OH label). I believe we should have this feature also for this interpreter.

Right now you can add synonyms to the items and those will be also considered as a valid values for the itemLabel placeholder.

bundles/org.openhab.voice.actiontemplatehli/src/main/java/org/openhab/voice/actiontemplatehli/internal/ActionTemplateInterpreter.java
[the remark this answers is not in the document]
lolodomo: Ok, makes sense.

lolodomo:
> Right now you can add synonyms to the items and those will be also considered as valid values for the itemLabel placeholder.

Ok, will try that.

lolodomo:
There are still few cases where you use a method parameter to set it to a new value inside the method.
Could you fix these few cases ?

GiviMAD:
> There are still few cases where you use a method parameter to set it to a new value inside the method. Could you fix these few cases ?

I did have that warning enabled on the IDE.
Must be corrected now.
Let me know for anything else.

GiviMAD:
@lolodomo, is there something pending in this one?

bundles/org.openhab.voice.actiontemplatehli/README.md
```diff
@@ -0,0 +1,389 @@
...
+* placeholders: defined placeholders that can be used on the template and replaced on the value.
+* silent: boolean used to avoid confirmation message.
+* targetMembers: when targeting a Group item, can be used to send the command to its member items instead.
+
+When read is true:
+
+* template: action template. (Required)
+* value: read template, can use the placeholders symbols $itemLabel and $state.
+* emptyValue: An alternative template. Is used when the state value is empty or NULL after the post transformation. The $itemLabel is available.
+* type: action template type.
```
lolodomo, line 74: Like in the previous block, can you add this: either "tokens" or "lemmas".

bundles/org.openhab.voice.actiontemplatehli/README.md
```diff
@@ -0,0 +1,389 @@
...
+
+Tokenizing the text is enough to use the action type 'tokens' as tokens are the only ones required for scoring (but the option 'optionalLanguageTags' will not take effect unless you have the POS language tags).
+
+### POSTagger (language tags)
+
+You need to provide a model for POS tagging at '<OPENHAB_USERDATA>/actiontemplatehli/pos.bin' for your language. 
+This will produce a language tag for each token, that can be used in 'optionalLanguageTags' to make some optional for scoring.
+As an example:
+
+The tokens "that,sounds,good" produces the tags "DT,VBZ,JJ".
```
lolodomo, line 180: How can you guess the values of these tags ?

lolodomo (changes requested):

lolodomo:
> @lolodomo, is there something pending in this one?

Just one improvement to do in the documentation and a question.

I will not have time to test today. I will merge today as soon as you give a feedback to these 2 comments.
I will probably then test it during the next week.

bundles/org.openhab.voice.actiontemplatehli/README.md
GiviMAD: Those depends on the lemmatization model you are using.

GiviMAD: I can find this link https://www.ling.upenn.edu/courses/Fall_2003/ling001/penn_treebank_pos.html 
But I think there is not an official standard for those, and that can different by language.

GiviMAD: I will add a note to recall this.

bundles/org.openhab.voice.actiontemplatehli/src/main/java/org/openhab/voice/actiontemplatehli/internal/configuration/ActionTemplateConfiguration.java
```diff
@@ -0,0 +1,70 @@
...
+import java.io.IOException;
+import java.util.List;
+
+import org.eclipse.jdt.annotation.NonNullByDefault;
+import org.eclipse.jdt.annotation.Nullable;
+import org.openhab.core.items.Metadata;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.ObjectMapper;
```
lolodomo, line 25: I do not see these dependencies in your POM file. Are they inherited with NLP dependency ?
Is there no need to mention it in the NOTICE file ?

GiviMAD: I'm not sure.
I though those are available to all the project, because I see more bindings using those without any notice around.
But maybe those are in fact included by the library.

GiviMAD: No I think those are available to all the project because I don't see the opennlp-tools depending on those.
But not totally sure.

lolodomo: I have the feeling it is included with the OH core framework:
https://github.com/openhab/openhab-core/blob/66953210898d4729dff08eee6dc72778cfeff28f/bom/runtime/pom.xml#L944
But this is not mentioned as default library in the binding development documentation: https://www.openhab.org/docs/developer/guidelines.html#default-libraries

GiviMAD: The chrome cast binding is doing this:
```
    <dependency>
      <groupId>com.fasterxml.jackson.core</groupId>
      <artifactId>jackson-core</artifactId>
      <version>${jackson.version}</version>
      <scope>compile</scope>
    </dependency>
    <dependency>
      <groupId>com.fasterxml.jackson.core</groupId>
      <artifactId>jackson-annotations</artifactId>
      <version>${jackson.version}</version>
      <scope>compile</scope>
    </dependency>
    <dependency>
      <groupId>com.fasterxml.jackson.core</groupId>
      <artifactId>jackson-databind</artifactId>
      <version>${jackson.version}</version>
      <scope>compile</scope>
    </dependency>
```
Uses a global version variable. I think I can go for the same.

GiviMAD: @lolodomo I have added it as in the chromecast binding and in the notice file

bundles/org.openhab.voice.actiontemplatehli/src/test/java/org/openhab/voice/actiontemplatehli/internal/ActionTemplateInterpreterTest.java
```diff
@@ -0,0 +1,248 @@
...
+        groupNPLOptionWriteAction.requiredItemTags = new String[] { "Location" };
+        groupNPLOptionWriteAction.value = "$itemOption";
+        groupNPLOptionWriteAction.memberTargets = new ActionTemplateGroupTargets();
+        groupNPLOptionWriteAction.memberTargets.itemType = "Number";
+        groupNPLOptionWriteAction.memberTargets.requiredItemTags = new String[] { "tv_channel" };
+        // Prepare group read action using item option
+        var groupNPLOptionReadAction = new ActionTemplateConfiguration();
+        groupNPLOptionReadAction.read = true;
+        groupNPLOptionReadAction.requiredItemTags = new String[] { "Location" };
+        groupNPLOptionReadAction.template = "what channel is the on the $itemLabel tv";
```
lolodomo, line 137: what channel is on ...

bundles/org.openhab.voice.actiontemplatehli/src/test/java/org/openhab/voice/actiontemplatehli/internal/ActionTemplateInterpreterTest.java
```diff
@@ -0,0 +1,248 @@
...
+        var response = interpreter.interpret(Locale.ENGLISH, "how is the light in the bedroom");
+        assertThat(response, is("bedroom light in bedroom is off"));
+    }
+
+    /**
+     * Test target a group member item using the itemOption placeholder
+     */
+    @Test
+    public void groupItemOptionTest() throws InterpretationException {
+        var response = interpreter.interpret(Locale.ENGLISH, "what channel is the on the bedroom tv");
```
lolodomo, line 229: what channel is on the bedroom tv

lolodomo (changes requested):

lolodomo:
Thank you for the tests.

GiviMAD:
> Thank you for the tests.

Never added tests in a binding before, but they were really useful for me while implementing the latest features.

bundles/org.openhab.voice.actiontemplatehli/src/main/java/org/openhab/voice/actiontemplatehli/internal/configuration/ActionTemplateConfiguration.java
lolodomo: Yes and the pattern was the same in the iCalendar binding. This is probably the correct approach.

lolodomo (approved):
LGTM, thank you

Issue #12262, opened 2022-02-12
Check existing semantic tags defined on channel types


lolodomo:
The guidelines about semantic tags are not yet really clear.
We found the following temporary agreement with @kaikreuzer to start adding semantic tags. Here are the current guidelines.  Kai, please correct me if something is wrong or not enough clear:

1. Semantic tags should be added to a channel type only when the assignment is absolutely obvious. For example, the tags "Measurement" + "Temperature" should not be set on a channel type used for forecasted temperature as forecast is not a measurement.
2. It is not allowed to set only "Control" or "Switch" tag on a writable channel type. Also not expected to set "Status" tag on all read-only channel types.
3. No property tag is expected with the "Status" point tag; a property tag is generally expected with a "Control" or "Measurement" point tag. As an example, Status + Temperature is not expected.

Now here are all the bindings which defined semantic tags on static channel types. We have to check if they respect the rules defined above:

- [x] airquality (6 channels) => PR #12290
- [x] danfossairunit (1 channel) => the system channel type could have probably rather be used.
- [x] deconz (5 channels) => PR #12287
- [x] dmx (3 channels) => the system channel types could have probably rather be used.  PR #12265
- [x] feican (1 channel) => the system channel type could have probably rather be used.
- [x] homeconnect (1 channel) => I see several other similar channel types without any tag.
- [x] hue (3 channels)
- [x] ipobserver  (13 channels) => PR #12379 
- [x] lifx (6 channels) => the system channel types could have probably rather be used. PR #12274
- [x] max (3 channels) => PR #12268
- [x] mihome (3 channels) => the system channel types could have probably rather be used.
- [x] milight (2 channels) => the system channel types could have probably rather be used.
- [x] millheat (7 channels) => PR #12289
- [x] mqtt.espmilighthub (3 channels) => the system channel types could have probably rather be used.
- [x] nikohomecontrol (2 channels)
- [x] omnilink (8 channels) =>PR #12288
- [x] openweathermap (6 channels) => PR #12285
- [x] openwebnet (5 channels) => PR #12271
- [x] qbus (2 channels)
- [x] rotel (1 channel)
- [x] sagercaster (4 channels) => PR #12284
- [x] sensibo (2 channels)
- [x] #12326
- [x] sonyprojector (1 channel) => PR #12283
- [x] souliss (10 channels) => PR #12269 and PR #12282
- [x] touchwand (1 channel)
- [x] vigicrues (7 channels) => PR #12286
[...]

PR #12269, opened 2022-02-12
[souliss] Fix semantic tags

lolodomo:
Related to #12262

### DIFF
--- a/bundles/org.openhab.binding.souliss/src/main/resources/OH-INF/thing/thing-types.xml
+++ b/bundles/org.openhab.binding.souliss/src/main/resources/OH-INF/thing/thing-types.xml
@@ -1142,9 +1142,6 @@
 		<item-type>Number</item-type>
 		<label>Healthy</label>
 		<description>Souliss Healthy</description>
-		<tags>
-			<tag>QualityOfService</tag>
-		</tags>
 		<state readOnly="true" pattern="%.0f"/>
 	</channel-type>
 
@@ -1152,9 +1149,6 @@
 		<item-type>Number</item-type>
 		<label>Value</label>
 		<description>Floating Point Input</description>
-		<tags>
-			<tag>sensors</tag>
-		</tags>
 		<state readOnly="true" pattern="%.1f "/>
 	</channel-type>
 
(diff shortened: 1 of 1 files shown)


bundles/org.openhab.binding.souliss/src/main/resources/OH-INF/thing/thing-types.xml
```diff
@@ -1238,7 +1238,8 @@
 		<description>Current Power</description>
 		<category>Energy</category>
 		<tags>
-			<tag>sensors</tag>
+			<tag>Measurement</tag>
+			<tag>Power</tag>
```
lolodomo, line 1242: I am not sure if Power or Energy must be used here.

lolodomo:
This binding does not use UoM except for temperature, while it could use it for many other channels. I don't believe it should prevent adding semantic tags. To be discussed in #12262

bundles/org.openhab.binding.souliss/src/main/resources/OH-INF/thing/thing-types.xml
wborn: I think this binding has some modeling issues because it seems that this power type is used for both Power and Pressure:

https://github.com/openhab/openhab-addons/blob/8b91e2172f48025efab0fa3f7701347e79b71bea/bundles/org.openhab.binding.souliss/src/main/resources/OH-INF/thing/thing-types.xml#L744-L751


https://github.com/openhab/openhab-addons/blob/8b91e2172f48025efab0fa3f7701347e79b71bea/bundles/org.openhab.binding.souliss/src/main/resources/OH-INF/thing/thing-types.xml#L769-L776

So until that is resolved it might be better to remove the tags?

lolodomo: Yes, I agree. I will remove it,

wborn (approved):

bundles/org.openhab.binding.souliss/src/main/resources/OH-INF/thing/thing-types.xml
wborn: Sorry I thought it was already removed. 😆

PR #12276, opened 2022-02-12
[nuki] Fixed configuration reload on initialization

janvyb:
Fixed bug mentionend in https://community.openhab.org/t/new-openhab2-binding-for-nuki-smart-locks/25940/186.

When user changes configuration property (unlatch in this case), openhab calls `initialize()` on Thing, but since configuration was only loaded in constructor, any changes user made will not take effect until openhab restart or reinstallation of binding.

I've also added logging of request received from bridge when handling callback, since I noticed some weird `NullpointerException`s in my logfile which were probably caused by invalid request, but without seeing the data sent it's hard to debug.

### DIFF
--- a/bundles/org.openhab.binding.nuki/src/main/java/org/openhab/binding/nuki/internal/dataexchange/NukiApiServlet.java
+++ b/bundles/org.openhab.binding.nuki/src/main/java/org/openhab/binding/nuki/internal/dataexchange/NukiApiServlet.java
@@ -114,7 +114,12 @@ public class NukiApiServlet extends HttpServlet {
             responseEntity = new ResponseEntity(HttpStatus.BAD_REQUEST_400,
                     new NukiHttpServerStatusResponseDto("Invalid BCB-Request!"));
         } else {
-            responseEntity = doHandle(bridgeApiLockStateRequestDto, request.getParameter("bridgeId"));
+            try {
+                responseEntity = doHandle(bridgeApiLockStateRequestDto);
+            } catch (Exception e) {
+                logger.warn("Error processing request '{}'", gson.toJson(bridgeApiLockStateRequestDto), e);
+                throw e;
+            }
         }
 
         setHeaders(response);
@@ -122,7 +127,7 @@ public class NukiApiServlet extends HttpServlet {
         response.getWriter().write(gson.toJson(responseEntity.getData()));
     }
 
-    private ResponseEntity doHandle(BridgeApiLockStateRequestDto request, @Nullable String bridgeId) {
+    private ResponseEntity doHandle(BridgeApiLockStateRequestDto request) {
         String nukiId = request.getNukiId().toString();
         for (NukiBridgeHandler nukiBridgeHandler : nukiBridgeHandlers) {
             logger.trace("Searching Bridge[{}] with NukiBridgeHandler[{}] for nukiId[{}].",
--- a/bundles/org.openhab.binding.nuki/src/main/java/org/openhab/binding/nuki/internal/handler/AbstractNukiDeviceHandler.java
+++ b/bundles/org.openhab.binding.nuki/src/main/java/org/openhab/binding/nuki/internal/handler/AbstractNukiDeviceHandler.java
@@ -32,6 +32,7 @@ import org.openhab.binding.nuki.internal.dataexchange.NukiBaseResponse;
 import org.openhab.binding.nuki.internal.dataexchange.NukiHttpClient;
 import org.openhab.binding.nuki.internal.dto.BridgeApiDeviceStateDto;
 import org.openhab.binding.nuki.internal.dto.BridgeApiListDeviceDto;
+import org.openhab.core.config.core.Configuration;
 import org.openhab.core.library.types.DateTimeType;
 import org.openhab.core.library.types.DecimalType;
 import org.openhab.core.library.types.OnOffType;
@@ -70,6 +71,11 @@ public abstract class AbstractNukiDeviceHandler<T extends NukiDeviceConfiguratio
     @Nullable
     private NukiHttpClient nukiHttpClient;
 
+    public AbstractNukiDeviceHandler(Thing thing) {
+        super(thing);
+        this.configuration = getConfigAs(getConfigurationClass());
+    }
+
     private static String hexToDecimal(String hexString) {
         return String.valueOf(Integer.parseInt(hexString, 16));
     }
@@ -92,23 +98,34 @@ public abstract class AbstractNukiDeviceHandler<T extends NukiDeviceConfiguratio
         }
     }
 
-    public AbstractNukiDeviceHandler(Thing thing) {
-        super(thing);
-        this.configuration = getConfigAs(getConfigurationClass());
+    /**
+     * Performs migration of old device configuration
+     * 
+     * @return true if configuration was change and reload is needed
+     */
+    protected boolean migrateConfiguration() {
+        String nukiId = getConfig().get(NukiBindingConstants.PROPERTY_NUKI_ID).toString();
         // legacy support - check if nukiId is hexadecimal (which might have been set by previous binding version)
         // and convert it to decimal
-        if (NUKI_ID_HEX_PATTERN.matcher(this.configuration.nukiId).matches()) {
+        if (NUKI_ID_HEX_PATTERN.matcher(nukiId).matches()) {
             logger.warn(
                     "SmartLock '{}' was created by old version of binding. It is recommended to delete it and discover again",
                     thing.getUID());
-            this.thing.getConfiguration().put(NukiBindingConstants.PROPERTY_NUKI_ID,
-                    hexToDecimal(configuration.nukiId));
-            this.configuration = getConfigAs(getConfigurationClass());
+            Configuration newConfig = editConfiguration();
+            newConfig.put(NukiBindingConstants.PROPERTY_NUKI_ID, hexToDecimal(nukiId));
+            updateConfiguration(newConfig);
+            return true;
+        } else {
+            return false;
         }
     }
 
     @Override
     public void initialize() {
+        this.configuration = getConfigAs(getConfigurationClass());
+        if (migrateConfiguration()) {
+            this.configuration = getConfigAs(getConfigurationClass());
+        }
         scheduler.execute(this::initializeHandler);
     }
 
--- a/bundles/org.openhab.binding.nuki/src/main/java/org/openhab/binding/nuki/internal/NukiHandlerFactory.java
+++ b/bundles/org.openhab.binding.nuki/src/main/java/org/openhab/binding/nuki/internal/NukiHandlerFactory.java
@@ -21,7 +21,6 @@ import org.openhab.binding.nuki.internal.dataexchange.NukiApiServlet;
 import org.openhab.binding.nuki.internal.handler.NukiBridgeHandler;
 import org.openhab.binding.nuki.internal.handler.NukiOpenerHandler;
 import org.openhab.binding.nuki.internal.handler.NukiSmartLockHandler;
-import org.openhab.core.config.core.Configuration;
 import org.openhab.core.id.InstanceUUID;
 import org.openhab.core.io.net.http.HttpClientFactory;
 import org.openhab.core.net.HttpServiceUtil;
@@ -86,11 +85,6 @@ public class NukiHandlerFactory extends BaseThingHandlerFactory {
         }
         logger.warn("No valid Handler found for Thing[{}]!", thingTypeUID);
         return null;
-    }
-
-    @Override
-    protected @Nullable Thing createThing(ThingTypeUID thingTypeUID, Configuration configuration, ThingUID thingUID) {
-        return super.createThing(thingTypeUID, configuration, thingUID);
     }
 
     @Override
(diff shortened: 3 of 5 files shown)


bundles/org.openhab.binding.nuki/src/main/java/org/openhab/binding/nuki/internal/dataexchange/NukiApiServlet.java
```diff
@@ -114,15 +114,20 @@ protected void service(HttpServletRequest request, HttpServletResponse response)
             responseEntity = new ResponseEntity(HttpStatus.BAD_REQUEST_400,
                     new NukiHttpServerStatusResponseDto("Invalid BCB-Request!"));
         } else {
-            responseEntity = doHandle(bridgeApiLockStateRequestDto, request.getParameter("bridgeId"));
+            try {
+                responseEntity = doHandle(bridgeApiLockStateRequestDto);
+            } catch (Exception e) {
```
lolodomo, line 119: Probably more something only for you as developer to find a potential bug.

bundles/org.openhab.binding.nuki/src/main/java/org/openhab/binding/nuki/internal/handler/AbstractNukiDeviceHandler.java
```diff
@@ -92,23 +97,24 @@ protected void withHttpClient(Consumer<NukiHttpClient> consumer) {
...
+    protected T loadConfiguration() {
+        T configuration = getConfigAs(getConfigurationClass());
         // legacy support - check if nukiId is hexadecimal (which might have been set by previous binding version)
         // and convert it to decimal
-        if (NUKI_ID_HEX_PATTERN.matcher(this.configuration.nukiId).matches()) {
+        if (NUKI_ID_HEX_PATTERN.matcher(configuration.nukiId).matches()) {
             logger.warn(
                     "SmartLock '{}' was created by old version of binding. It is recommended to delete it and discover again",
                     thing.getUID());
             this.thing.getConfiguration().put(NukiBindingConstants.PROPERTY_NUKI_ID,
```
lolodomo, line 108: It looks like a non standard (and unexpected) way to update the thing configuration. Can you please fix that?

lolodomo (changes requested):

bundles/org.openhab.binding.nuki/src/main/java/org/openhab/binding/nuki/internal/handler/AbstractNukiDeviceHandler.java
lolodomo: https://www.openhab.org/docs/developer/bindings/#updating-the-configuration

bundles/org.openhab.binding.nuki/src/main/java/org/openhab/binding/nuki/internal/dataexchange/NukiApiServlet.java
janvyb: I mean, yes, that's exactly why I added it. If someone reports problem with callback I won't be able to debug it if I don't know what was sent.

bundles/org.openhab.binding.nuki/src/main/java/org/openhab/binding/nuki/internal/handler/AbstractNukiDeviceHandler.java
janvyb: Ok, I've updated it, but now I'm getting another warning in log since this is done before the thing is initialized:
```
19:29:56.486 [WARN ] [internal.handler.NukiSmartLockHandler] - SmartLock 'nuki:smartlock:4c89efd66d:647771828' was created by old version of binding. It is recommended to delete it and discover again
19:30:43.432 [WARN ] [b.core.thing.binding.BaseThingHandler] - Handler NukiSmartLockHandler tried updating its configuration although the handler was already disposed.
```

lolodomo: Yes, you should init configuration only in intialize, not in the constructor. Or at least you should not call loadConfiguration in the constructor because it updates the thing configuration too early, when not expected. 
Just call loadConfiguration in initialize and it will work.

bundles/org.openhab.binding.nuki/src/main/java/org/openhab/binding/nuki/internal/dataexchange/NukiApiServlet.java
lolodomo: But do you have an example of exception by a user? 
Why not rather analyzing the current code and only catch more specific exceptions? 
In case this change makes sense, logging at warn level would be more appropriate.

bundles/org.openhab.binding.nuki/src/main/java/org/openhab/binding/nuki/internal/handler/AbstractNukiDeviceHandler.java
janvyb: I was trying to avoid marking that `configuration` field as @Nullable since it's massive pain to deal with it (putting if != null every time I want to read it makes the code unreadable really fast), and if I don't initialize it in constructor it throws error.

bundles/org.openhab.binding.nuki/src/main/java/org/openhab/binding/nuki/internal/dataexchange/NukiApiServlet.java
janvyb: As I've said I've added it because I found some Nullpointers in logfile which I can't replicate (which means it could happen to someone else) and knowing what data were sent would really help. I agree that warn would be better.

bundles/org.openhab.binding.nuki/src/main/java/org/openhab/binding/nuki/internal/handler/AbstractNukiDeviceHandler.java
lolodomo: Yes you can do it but your method loadConfiguration is inappropriate for that. Just keep the initialize of the variable. 
And the stuff about updating the thing configuration, you do it only on initialize().

bundles/org.openhab.binding.nuki/src/main/java/org/openhab/binding/nuki/internal/dataexchange/NukiApiServlet.java
lolodomo: Why not fixing the NPE in this case, rather than catching the  exception?

janvyb: I've fixed the one place it throws NPE but it still would be helpful to see the data sent since that boolean flag should not be null

bundles/org.openhab.binding.nuki/src/main/java/org/openhab/binding/nuki/internal/handler/AbstractNukiDeviceHandler.java
janvyb: Fixed, guess marking it as nullable does not really throw errors elsewhere

bundles/org.openhab.binding.nuki/src/main/java/org/openhab/binding/nuki/internal/handler/AbstractNukiDeviceHandler.java
```diff
@@ -66,10 +67,15 @@
 
     @Nullable
     protected ScheduledFuture<?> reInitJob;
+    @Nullable
     protected T configuration;
```
lolodomo, line 71: No need to have a nullable member.
See my next comment.

bundles/org.openhab.binding.nuki/src/main/java/org/openhab/binding/nuki/internal/handler/AbstractNukiDeviceHandler.java
```diff
@@ -66,10 +67,15 @@
...
     @Nullable
     private NukiHttpClient nukiHttpClient;
 
+    public AbstractNukiDeviceHandler(Thing thing) {
+        super(thing);
+    }
```
lolodomo, line 77:
```
    public AbstractNukiDeviceHandler(Thing thing) {
        super(thing);
        this.configuration = getConfigAs(getConfigurationClass());
    }
```

bundles/org.openhab.binding.nuki/src/main/java/org/openhab/binding/nuki/internal/handler/AbstractNukiDeviceHandler.java
```diff
@@ -92,23 +98,25 @@ protected void withHttpClient(Consumer<NukiHttpClient> consumer) {
...
         // and convert it to decimal
-        if (NUKI_ID_HEX_PATTERN.matcher(this.configuration.nukiId).matches()) {
+        if (NUKI_ID_HEX_PATTERN.matcher(configuration.nukiId).matches()) {
             logger.warn(
                     "SmartLock '{}' was created by old version of binding. It is recommended to delete it and discover again",
                     thing.getUID());
-            this.thing.getConfiguration().put(NukiBindingConstants.PROPERTY_NUKI_ID,
-                    hexToDecimal(configuration.nukiId));
-            this.configuration = getConfigAs(getConfigurationClass());
+            Configuration newConfig = new Configuration(this.thing.getConfiguration());
```
lolodomo, line 109: Configuration newConfig = editConfiguration();

bundles/org.openhab.binding.nuki/src/main/java/org/openhab/binding/nuki/internal/handler/AbstractNukiDeviceHandler.java
```diff
@@ -92,23 +98,25 @@ protected void withHttpClient(Consumer<NukiHttpClient> consumer) {
         }
     }
 
-    public AbstractNukiDeviceHandler(Thing thing) {
-        super(thing);
-        this.configuration = getConfigAs(getConfigurationClass());
+    protected void loadConfiguration() {
```
lolodomo, line 101:
```
protected T loadConfiguration() {
```

bundles/org.openhab.binding.nuki/src/main/java/org/openhab/binding/nuki/internal/handler/AbstractNukiDeviceHandler.java
```diff
@@ -92,23 +98,25 @@ protected void withHttpClient(Consumer<NukiHttpClient> consumer) {
...
                     thing.getUID());
-            this.thing.getConfiguration().put(NukiBindingConstants.PROPERTY_NUKI_ID,
-                    hexToDecimal(configuration.nukiId));
-            this.configuration = getConfigAs(getConfigurationClass());
+            Configuration newConfig = new Configuration(this.thing.getConfiguration());
+            newConfig.put(NukiBindingConstants.PROPERTY_NUKI_ID, hexToDecimal(configuration.nukiId));
+            updateConfiguration(newConfig);
+            configuration = getConfigAs(getConfigurationClass());
         }
+        this.configuration = configuration;
```
lolodomo, line 114:
```
return configuration;
```

bundles/org.openhab.binding.nuki/src/main/java/org/openhab/binding/nuki/internal/handler/AbstractNukiDeviceHandler.java
```diff
@@ -92,23 +98,25 @@ protected void withHttpClient(Consumer<NukiHttpClient> consumer) {
...
     }
 
     @Override
     public void initialize() {
+        loadConfiguration();
```
lolodomo, line 119:
```
this.configuration = loadConfiguration();
```

lolodomo (changes requested):

bundles/org.openhab.binding.nuki/src/main/java/org/openhab/binding/nuki/internal/handler/AbstractNukiDeviceHandler.java
janvyb: The problem with this solution is that until `initialize()` is called there might be invalid configuration loaded. Although nothing should really happen before initialize so if it's ok with you I can live with it.

janvyb:
How about this? I renamed the method, I didn't like the `loadConfiguration()` name since it did too much and you would not expect that method to have any side effects (like also changing thing configuration).

lolodomo:
That looks even better now. 
But please reload the configuration at first in initialize because I am not sure the load in the constructor is a guarantee to have the proper configuration. Normally, the loading is done in initialize.

lolodomo:
Just to be clear, here is what I was suggesting:
```
    public void initialize() {
        this.configuration = getConfigAs(getConfigurationClass());
        if (migrateConfiguration()) {
            this.configuration = getConfigAs(getConfigurationClass());
        }
        scheduler.execute(this::initializeHandler);
    }
```
Note that when the thing configuration is updated, I believe the core framework will dispo/initialize again. So I believe the proper code should be:
```
    public void initialize() {
        this.configuration = getConfigAs(getConfigurationClass());
        if (migrateConfiguration()) {
            return;
        }
        scheduler.execute(this::initializeHandler);
    }
```
but that would be better if you can test it but I am not 100% sure.

lolodomo (approved):
LGTM

PR #12277, opened 2022-02-12
[danfossairunit] Use system channel types for humidity and outdoor temperature

jlaur:
Related to #12262 

Use system channel types where possible:
- Relative humidity
- Outdoor temperature

Please note that openHAB needs to be restarted after updating the binding with new channel type id's.

See also: #11121


lolodomo (approved):
LGTM

lolodomo:
> Please note that openHAB needs to be restarted after updating the binding with new channel type id's.

I have not yet investigated that myself so I still doubt when I consider how many complaints we got in the community forum when the exactly same change was done for the hue binding.
But as you are apparently sure of you, let's go.

jlaur:
Created openhab/openhab-distro#1369 for mentioning the **humidity** channel type reassignment as a possibly breaking change. The **outdoor_temp** channel type reassignment is non-breaking as the previously referenced locally defined channel type still exists. See https://github.com/openhab/openhab-addons/issues/12299#issuecomment-1042215452 for details.

PR #12285, opened 2022-02-13
[openweathermap] Cleanup semantic tags

lolodomo:
Related to #12262

### DIFF
--- a/bundles/org.openhab.binding.openweathermap/src/main/resources/OH-INF/thing/channel-types.xml
+++ b/bundles/org.openhab.binding.openweathermap/src/main/resources/OH-INF/thing/channel-types.xml
@@ -455,10 +455,6 @@
 		<label>Apparent Temperature</label>
 		<description>Current apparent temperature.</description>
 		<category>Temperature</category>
-		<tags>
-			<tag>Measurement</tag>
-			<tag>Temperature</tag>
-		</tags>
 		<state readOnly="true" pattern="%.1f %unit%"/>
 	</channel-type>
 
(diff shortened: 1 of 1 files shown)


bundles/org.openhab.binding.openweathermap/src/main/resources/OH-INF/thing/channel-types.xml
```diff
@@ -507,10 +503,6 @@
 		<label>Precipitation</label>
 		<description>Precipitation volume of the related timespan.</description>
 		<category>Rain</category>
-		<tags>
-			<tag>Measurement</tag>
-			<tag>Rain</tag>
```
kaikreuzer, line 512: Which rule do these tags break? Is it because it implicitly refers to an (undefined) measurement period?

lolodomo: In fact, it is forecasted precipitations. It is a channel type only member of group named oneCallMinutely having this label "One Call API Minutely Forecast". Maybe the channel label or description could be improved.

kaikreuzer: Ok, I indeed didn't notice this as I didn't look at the channel group of it. Thanks for the explanation!

kaikreuzer (approved):
Thanks!

bundles/org.openhab.binding.openweathermap/src/main/resources/OH-INF/thing/channel-types.xml
lolodomo: And you have another channel type "rain" with tags Measurment + Rain.

PR #12287, opened 2022-02-13
[deconz] Cleanup semantic tags

lolodomo:
Related to #12262

### DIFF
--- a/bundles/org.openhab.binding.deconz/src/main/resources/OH-INF/thing/group-thing-types.xml
+++ b/bundles/org.openhab.binding.deconz/src/main/resources/OH-INF/thing/group-thing-types.xml
@@ -35,10 +35,6 @@
 		<item-type>Switch</item-type>
 		<label>Any On</label>
 		<description>"On" if any light in this group is "On", otherwise "Off".</description>
-		<tags>
-			<tag>Measurement</tag>
-			<tag>Light</tag>
-		</tags>
 		<state readOnly="true"/>
 	</channel-type>
 
--- a/bundles/org.openhab.binding.deconz/src/main/resources/OH-INF/thing/group-thing-types.xml
+++ b/bundles/org.openhab.binding.deconz/src/main/resources/OH-INF/thing/group-thing-types.xml
@@ -28,9 +28,6 @@
 		<item-type>Switch</item-type>
 		<label>All On</label>
 		<description>"On" if all lights in this group are "On", otherwise "Off".</description>
-		<tags>
-			<tag>Status</tag>
-		</tags>
 		<state readOnly="true"/>
 	</channel-type>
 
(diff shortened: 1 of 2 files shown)


bundles/org.openhab.binding.deconz/src/main/resources/OH-INF/thing/group-thing-types.xml
```diff
@@ -30,7 +30,6 @@
 		<description>"On" if all lights in this group are "On", otherwise "Off".</description>
 		<tags>
 			<tag>Status</tag>
```
jlaur, line 32: Is it on purpose that Status is left for **all_on**, but removed for **any_on**?

lolodomo: Yes, my idea is that it could make sense to have as status for the group that all lights are ON or not.
But now, I told me that I could rather use Measurement + Light. Should I put them on all_on or any_on ?
Finally, the more appropriate could be Measurement + Light only on any_on ?

lolodomo: In fact, this is the same case as "is_raining" on another binding. As discussed in #12262, I finally tagged with Measurement and Rain. So, I believe I should use here Measurement + Light. I am just not sure on which of both channel types.

jlaur: These channels are for `lightgroup`. I don't think they represent any kind of measurement, but simply logic. @J-N-K, please correct me if I'm wrong.

lolodomo: They are read-only channels type. The tags could help to know if this group is on or off.

lolodomo: I pushed the change as it is more coherent with what was proposed in #12262

jlaur: My point is that they represent logic, not measurements. "Measurement" + "Light" seems to indicate that it's a measurement of light?

lolodomo: Measurement if there is light or not light.
Maybe @kaikreuzer can comment as it looks to me relatively similar to the case "is raining" which was discussed earlier.

kaikreuzer: I'm with @jlaur here - in this case, we don't have a sensor that measures anything, but it is a channel which reports a (logical) state of the lights. Maybe the light is on, but the bulb is broken, so it is still dark ;-) A sensor would notice, a logical state reporting channel doesn't.
Especially on such grouping features, I'd remove the tags completely.

jlaur: I think in that scenario ("is raining") it was based on sensor data, i.e. an actual measurement, just in low resolution (boolean).

lolodomo: Ok guys, I remove it.

jlaur (approved):
LGTM

PR #12303, opened 2022-02-16
[danfossairunit] Fix null annotation issues

jlaur:
Fix two issues:

```
[INFO] openhab-addons\bundles\org.openhab.binding.danfossairunit\src\main\java\org\openhab\binding\danfossairunit\internal\discovery\DanfossAirUnitDiscoveryService.java:[85,53] Unsafe interpretation of method return type as '@NonNull' based on the receiver type 'java.util.Enumeration<java.net.@NonNull NetworkInterface>'. Type 'java.util.Enumeration<E>' doesn't seem to be designed with null type annotations in mind
[INFO] openhab-addons\bundles\org.openhab.binding.danfossairunit\src\test\java\org\openhab\binding\danfossairunit\internal\DanfossAirUnitTest.java:[43,40] Unsafe interpretation of method return type as '@NonNull' based on substitution 'T=org.openhab.binding.danfossairunit.internal.@NonNull CommunicationController'. Declaring type 'org.mockito.Mockito' doesn't seem to be designed with null type annotations in mind
```

### DIFF
--- a/bundles/org.openhab.binding.danfossairunit/src/main/java/org/openhab/binding/danfossairunit/internal/discovery/DanfossAirUnitDiscoveryService.java
+++ b/bundles/org.openhab.binding.danfossairunit/src/main/java/org/openhab/binding/danfossairunit/internal/discovery/DanfossAirUnitDiscoveryService.java
@@ -28,6 +28,7 @@ import java.util.Map;
 import java.util.Set;
 
 import org.eclipse.jdt.annotation.NonNullByDefault;
+import org.eclipse.jdt.annotation.Nullable;
 import org.openhab.core.config.discovery.AbstractDiscoveryService;
 import org.openhab.core.config.discovery.DiscoveryResult;
 import org.openhab.core.config.discovery.DiscoveryResultBuilder;
@@ -82,6 +83,7 @@ public class DanfossAirUnitDiscoveryService extends AbstractDiscoveryService {
         try (DatagramSocket socket = new DatagramSocket()) {
             Enumeration<NetworkInterface> interfaces = NetworkInterface.getNetworkInterfaces();
             while (interfaces.hasMoreElements()) {
+                @Nullable
                 NetworkInterface networkInterface = interfaces.nextElement();
                 if (networkInterface.isLoopback() || !networkInterface.isUp()) {
                     continue;
--- a/bundles/org.openhab.binding.danfossairunit/src/test/java/org/openhab/binding/danfossairunit/internal/DanfossAirUnitTest.java
+++ b/bundles/org.openhab.binding.danfossairunit/src/test/java/org/openhab/binding/danfossairunit/internal/DanfossAirUnitTest.java
@@ -20,8 +20,11 @@ import java.io.IOException;
 import java.time.ZoneId;
 import java.time.ZonedDateTime;
 
-import org.junit.jupiter.api.BeforeEach;
+import org.eclipse.jdt.annotation.NonNullByDefault;
 import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.ExtendWith;
+import org.mockito.Mock;
+import org.mockito.junit.jupiter.MockitoExtension;
 import org.openhab.core.library.types.DateTimeType;
 import org.openhab.core.library.types.DecimalType;
 import org.openhab.core.library.types.OnOffType;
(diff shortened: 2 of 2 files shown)


pravussum (approved):

bundles/org.openhab.binding.danfossairunit/src/main/java/org/openhab/binding/danfossairunit/internal/discovery/DanfossAirUnitDiscoveryService.java
```diff
@@ -82,6 +83,7 @@ private synchronized void discover() {
         try (DatagramSocket socket = new DatagramSocket()) {
             Enumeration<NetworkInterface> interfaces = NetworkInterface.getNetworkInterfaces();
             while (interfaces.hasMoreElements()) {
+                @Nullable
                 NetworkInterface networkInterface = interfaces.nextElement();
```
lolodomo, line 87: I am not sure to understand why you want to add `@Nullable` ?
If interfaces.hasMoreElements() returns true, I assume interfaces.nextElement() will never be null.
And if networkInterface could be null, your next test should be updated to avoid a NPE.

jlaur: > I am not sure to understand why you want to add `@Nullable` ?

To fix the first of the two mentioned issues where null annotations are not used by Enumerable but was treated as NonNull.

> If interfaces.hasMoreElements() returns true, I assume interfaces.nextElement() will never be null.

Correct, that's also the analysis result, hence no null annotation warning introduced.

lolodomo: Sorry but I am not understanding. My feeling is that add `@Nullable` should introduce a SAT warning at line 88.
@fwolter : please have the lead to merge this PR.

fwolter: > If interfaces.hasMoreElements() returns true, I assume interfaces.nextElement() will never be null.

I'd go further and say: interfaces.nextElement() will never be null.

See the javadoc:
```
    /**
     * Returns the next element of this enumeration if this enumeration
     * object has at least one more element to provide.
     *
     * @return     the next element of this enumeration.
     * @exception  NoSuchElementException  if no more elements exist.
     */
```

So, I wouldn't declare it as `@Nullable`.

Note that the quoted compiler outputs are only info, not warnings.

jlaur: @fwolter, we agree that interfaces.nextElement() will never be null. The compiler also agrees to that, hence no warnings when annotated as nullable.

We have to either annotate it as nullable or non-nullable. When doing the latter we get  a compiler info that annotations are not supported/used by the API we are using. So we should assume the worst, i.e. nullable. Annotation is just annotation, so this doesn't imply that it can ever actually be null. If we would check for null, it would yield a warning, as analysis concludes that it can't be null.

So I don't see the problem in getting rid of a compiler info this way? I guess it must be common when bridging code with and without annotations.

fwolter: It's the combination of using a non-annotated class with generics. A thirs solution is to declare the generic type as nullable:
```

fwolter (approved):

bundles/org.openhab.binding.danfossairunit/src/main/java/org/openhab/binding/danfossairunit/internal/discovery/DanfossAirUnitDiscoveryService.java
jlaur: @fwolter - thanks for additional perspective. For this case I would stay with the current approach, since it eliminates compiler info, warning as well as redundant null check - so IMHO it wins. I tried your approach also, but having to add a redundant null check to avoid a warning seems less desirable.

fwolter: Yes, that'd be necessary, then. Lets merge it!

PR #12305, opened 2022-02-17
[homekit] fix "No Response" issue due missing pairing information

yfre:
there are many reports (almost since 2 years) about missing "user" entry in homekit.json which leads to repairing need. this issue was difficult to reproduce and often disappeared after some time. 
finally the root cause found  - different storage caches.

2 classes write to homekit.json storage:
- HomekitChangeListener writes number of accessories and configuration version 
- HomekitAuthInfoImpl writes "user" entry after device pairing

on binding start both classes got their own Storage instance via StorageService.getStorage("homekit.json") call. 

however, JsonStorageService.getStorage reads homekit.json every time into a new cache object (map) 
https://github.com/openhab/openhab-core/blob/44da7a4e0ee0b7b1f33c49f3f13f6cf9c2747058/bundles/org.openhab.core.storage.json/src/main/java/org/openhab/core/storage/json/internal/JsonStorageService.java#L130

as result, HomekitChangeListener and HomekitAuthInfoImpl had 2 different cache instances. 

On pairing, HomekitAuthInfoImpl added "user" entry to its cache and flushed/wrote the cache to homekit.json. homekit.json had "user" entry. But this was not reflected in HomekitChangeListener storage cache, i.e. "user" entry was missing in HomekitChangeListener cache.

now, if one would change item configuration, HomekitChangeListener would update configuration number in his storage cache and flushed to homekit.json. as result, user entry was removed from homekit.json. On next restart, HomekitAuthInfoImpl would report missing pairing information.

if homekit binding was restarted after pairing and before changes to items, then HomekitChangeListener got the cache with "user" entry and the issue disappeared. 

this PR ensures that getStorage is called only once and both classes use the same instance of Storage/Cache.

### DIFF
--- a/bundles/org.openhab.io.homekit/src/main/java/org/openhab/io/homekit/internal/HomekitChangeListener.java
+++ b/bundles/org.openhab.io.homekit/src/main/java/org/openhab/io/homekit/internal/HomekitChangeListener.java
@@ -34,7 +34,6 @@ import org.openhab.core.items.ItemRegistryChangeListener;
 import org.openhab.core.items.Metadata;
 import org.openhab.core.items.MetadataRegistry;
 import org.openhab.core.storage.Storage;
-import org.openhab.core.storage.StorageService;
 import org.openhab.io.homekit.internal.accessories.HomekitAccessoryFactory;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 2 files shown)


lolodomo (approved):
LGTM

lolodomo:
@andylintner for review

PR #12313, opened 2022-02-18
[jdbc] Catch exception in case of connection to database failed during activation of service

cweitkamp:
- Catch exception in case of connection to database failed during activation of service

See https://github.com/openhab/openhab-core/issues/2630


kaikreuzer (approved):
Thank you!

Issue #11503, opened 2021-11-03
[Network] Binding: arp_state is always Disabled, presence_detection_type is always emtpy


FranzSchi:
## Expected Behavior

A new instance of OH is setup. (in my case with docker)
Network Binding is added, arping is part of the Docker image and installed
A "pingable device" is added.
arp_state should show the arp_state
and presence_detection_type should show "arp", "system_ping", "ping" ..-. whatever 

## Current Behavior

Binding: arp_state is always Disabled, presence_detection_type is always emtpy even if the arping binary is detected and arping is executed.
```
2021-11-03 14:54:53.689 [DEBUG] [twork.internal.NetworkHandlerFactory] - bundle org.openhab.binding.network:3.1.0 (300)[org.openhab.binding.network.internal.NetworkHandlerFactory(349)] : invoking activate: activate: parameters [org.apache.felix.scr.impl.manager.ComponentContextImpl, org.apache.felix.scr.impl.helper.ReadOnlyDictionary]
2021-11-03 14:54:53.704 [DEBUG] [twork.internal.NetworkHandlerFactory] - Updated binding configuration to NetworkBindingConfiguration{allowSystemPings=true, allowDHCPlisten=true, cacheDeviceStateTimeInMS=2000, arpPingToolPath='/usr/sbin/arping', arpPingUtilMethod=THOMAS_HABERT_ARPING, preferResponseTimeAsLatency=false}
2021-11-03 14:54:53.706 [DEBUG] [twork.internal.NetworkHandlerFactory] - bundle org.openhab.binding.network:3.1.0 (300)[org.openhab.binding.network.internal.NetworkHandlerFactory(349)] : invoked activate: activate
2021-11-03 14:54:53.707 [DEBUG] [twork.internal.NetworkHandlerFactory] - bundle org.openhab.binding.network:3.1.0 
```

and here the arping is done

```
2021-11-03 14:54:53.926 [TRACE] [g.network.internal.PresenceDetection] - host name resolved to other address, (re-)setup presence detection
2021-11-03 14:54:53.935 [TRACE] [g.network.internal.PresenceDetection] - Perform ARP ping presence detection for 192.168.3.26 on interface: veth132adc8
2021-11-03 14:54:53.937 [TRACE] [g.network.internal.PresenceDetection] - Perform ARP ping presence detection for 192.168.3.26 on interface: veth128711e
[...]

PR #12316, opened 2022-02-18
[network] Adressing issue #11503

clinique:
Adressing some SAT complains and removed a reference to Apache StringUtils.

### DIFF
--- a/bundles/org.openhab.binding.network/src/main/java/org/openhab/binding/network/internal/utils/LatencyParser.java
+++ b/bundles/org.openhab.binding.network/src/main/java/org/openhab/binding/network/internal/utils/LatencyParser.java
@@ -28,7 +28,7 @@ import org.slf4j.LoggerFactory;
 @NonNullByDefault
 public class LatencyParser {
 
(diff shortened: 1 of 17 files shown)


bundles/org.openhab.binding.network/src/main/java/org/openhab/binding/network/internal/utils/LatencyParser.java
```diff
@@ -24,9 +25,10 @@
  *
  * @author Andreas Hirsch - Initial contribution
  */
+@NonNullByDefault
 public class LatencyParser {
 
-    private static Pattern LATENCY_PATTERN = Pattern.compile(".*time=(.*) ?ms");
+    private static final Pattern latencyPattern = Pattern.compile(".*time=(.*) ?ms");
```
wborn, line 31: I think SAT will complain about this casing now that the final modifier was?

```suggestion
    private static final Pattern LATENCY_PATTERN = Pattern.compile(".*time=(.*) ?ms");
```

wborn (approved):
Thanks for the fix and cleanup!

PR #12318, opened 2022-02-18
[watsonstt] lazy abort

GiviMAD:

Do not force close background task.

@lolodomo, for reference, this is the change I was talking about in the VoskSTT PR https://github.com/openhab/openhab-addons/pull/12249#discussion_r807709755

### DIFF
--- a/bundles/org.openhab.voice.watsonstt/src/main/java/org/openhab/voice/watsonstt/internal/WatsonSTTService.java
+++ b/bundles/org.openhab.voice.watsonstt/src/main/java/org/openhab/voice/watsonstt/internal/WatsonSTTService.java
@@ -134,7 +134,7 @@ public class WatsonSTTService implements STTService {
                 .build();
         final AtomicReference<@Nullable WebSocket> socketRef = new AtomicReference<>();
         final AtomicBoolean aborted = new AtomicBoolean(false);
-        var task = executor.submit(() -> {
+        executor.submit(() -> {
             int retries = 2;
             while (retries > 0) {
                 try {
@@ -272,8 +272,10 @@ public class WatsonSTTService implements STTService {
                 return;
             }
             logger.warn("TranscriptionError: {}", errorMessage);
-            sttListener.sttEventReceived(
-                    new SpeechRecognitionErrorEvent(errorMessage != null ? errorMessage : "Unknown error"));
+            if (!aborted.get()) {
+                sttListener.sttEventReceived(
+                        new SpeechRecognitionErrorEvent(errorMessage != null ? errorMessage : "Unknown error"));
+            }
         }
 
         @Override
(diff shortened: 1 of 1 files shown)


bundles/org.openhab.voice.watsonstt/src/main/java/org/openhab/voice/watsonstt/internal/WatsonSTTService.java
```diff
@@ -132,12 +133,13 @@ public STTServiceHandle recognize(STTListener sttListener, AudioStream audioStre
                 .speechDetectorSensitivity(config.speechDetectorSensitivity).inactivityTimeout(config.inactivityTimeout)
                 .build();
         final AtomicReference<@Nullable WebSocket> socketRef = new AtomicReference<>();
+        final AtomicBoolean aborted = new AtomicBoolean(false);
         var task = executor.submit(() -> {
```
lolodomo, line 137: You can now suppress "var task = "

bundles/org.openhab.voice.watsonstt/src/main/java/org/openhab/voice/watsonstt/internal/WatsonSTTService.java
```diff
@@ -275,16 +280,18 @@ public void onError(@Nullable Exception e) {
         public void onDisconnected() {
```
lolodomo, line 280: Nothing similar to do in onError ?

lolodomo (changes requested):

bundles/org.openhab.voice.watsonstt/src/main/java/org/openhab/voice/watsonstt/internal/WatsonSTTService.java
```diff
@@ -157,16 +159,17 @@ public STTServiceHandle recognize(STTListener sttListener, AudioStream audioStre
...
             public void abort() {
-                var socket = socketRef.get();
-                if (socket != null) {
-                    socket.close(1000, null);
-                    socket.cancel();
-                    try {
-                        Thread.sleep(100);
-                    } catch (InterruptedException ignored) {
+                if (!aborted.getAndSet(true)) {
+                    var socket = socketRef.get();
```
lolodomo, line 163: Is there no risk to not close the socket ?
This would be the case when onDisconnected is first called. You are setting aborted to true. Then when abort is called, you will not enter in the if.

GiviMAD: It shouldn't as the sdk executes onDisconnect after the socket is closed, so if it was executed the socket is already closed, so no needed to close it again. 
On the error method I'm not setting abort to true because I can not asume there that the socket is closed.

lolodomo: Ok

lolodomo (approved):
LGTM

PR #12320, opened 2022-02-19
[hueemulation] Provide non-null context to TriggerHandlerCallback

wborn:
This prevents NPEs and a compilation issue when openhab/openhab-core#2763 is merged.


jlaur (approved):
Thanks!

kaikreuzer:
@jlaur I was just reviewing this at the same time and was actually not happy with the change.
https://github.com/openhab/openhab-core/pull/2763 introduces a backward-compatibility breaking change by now not allowing null values anymore here.
@wborn I am in general a fan of the [robustness principle](https://en.wikipedia.org/wiki/Robustness_principle), hence I prefer to have such method inputs being treated conservatively - i.e. also allowing null values as it is here seems to be an optional parameter anyway.
Why do you think it should be made a non-null param?

jlaur:
@kaikreuzer - my thinking was that even if openhab/openhab-core#2763 will not be merged (and make the API more strict), it will not hurt to avoid null when consuming the API, thus being more strict here?

kaikreuzer:
@jlaur Well yes, you are right that my comment would have better done on https://github.com/openhab/openhab-core/pull/2763 then. :-)

kaikreuzer:
Done that now: https://github.com/openhab/openhab-core/pull/2763/files#r810536664

Issue #10454, opened 2021-04-04
Sending request with Jetty HTTP client: catching InterruptedException or not ?


lolodomo:
@octa22 : I was asking myself if it is really useful to catch InterruptedException to then call
```
Thread.currentThread().interrupt();
```
You are doing that just to change the thing status.
But if the thread is interrupted, we can imagine that is for a severe reason and the thing handler will be certainly destroyed when it occurs.

My feeling is that InterruptedException should not be caught.
WDYT ? What was you reason to do that ?

lolodomo:
Interesting, I checked how I handled exceptions in the remote openHAB binding when sending a Jetty HTTP request and I see that I caught `Exception`, meaning at the same time `TimeoutException` and `InterruptedException`. Doing that way, I am masking the interrupt !
Your approach might be the correct one, @octa22 
It would be interesting to get the advice from an expert and then check and fix all bindings using Jetty HTTP client.
@openhab/add-ons-maintainers for advice.

fwolter:
I see lots of code catching `Exception` in reviews. The resulting problems are exactly like @lolodomo describes. Additionally, all unchecked exceptions like `RuntimeException` are cought, which might not be the intention. I try to throw `InterruptedExceptions` as high as possible. `Thread.currentThread().interrupt()` is not necessary in most cases, then.

lolodomo:
I just checked few bindings and each one is doing differently.
The ones that thrown `InterruptedException` to a higher level exist but are a minority.
In most cases, the 3 exceptions are just caught, including `InterruptedException`. Some bindings, like Somfy Tahoma, are calling `Thread.currentThread().interrupt();`.

If we would like to throw this exception to a higher level, the change is more difficult, it would require to change the method calling `request.send()` but also all methods calling it.

At first steps, we could change:

1. all bindings just catching with `Exception` for the call to request.send() to replace it by each of the 3 exceptions triggered by the call. My remote openHAB binding is an example to fix.
2. all binding catching `InterruptedException` triggered by the call to request.send() to add a call to `Thread.currentThread().interrupt();` if missing.

WDYT ?

fwolter:
I'm fine with that.

You could combine the catch clause for `TimeoutException` and `ExecutionException` with the `|` operator.

lolodomo:
More generally in relation to your remark, I found 625 catch of `Exception` !!!
I have not yet any idea how Exception catching are relative to Jetty HTTP requests but it should not so much.

[...]

PR #12314, opened 2022-02-18
[openweathermap] Call Thread.currentThread().interrupt() on InterruptedException

cweitkamp:
- Call `Thread.currentThread().interrupt()` on `InterruptedException`

Related to https://github.com/openhab/openhab-addons/issues/10454

Signed-off-by: Christoph Weitkamp <github@christophweitkamp.de>

PR #12323, opened 2022-02-19
[hdpowerview] Interrupt current thread on InterruptedException

jlaur:
Related to #10454


jlaur:
@lolodomo - I saw #12314, then read #10454 and https://www.baeldung.com/java-interrupted-exception and concluded that we need a similar fix in the hdpowerview binding. Can you assess if this is intended/correct behavior?

lolodomo (approved):
LGTM

PR #12324, opened 2022-02-19
[daikin] Fix Special Mode and Streamer support

jimtng:
Some Daikin models have the ability to set 
* Streamer mode on/off
* Special mode: Eco / Powerful / Normal

Currently:
* The channel `specialmode` reports the current combination of modes set on the device: streamer / eco_streamer / powerful_streamer / eco / powerful / off
* It is read-only
* Only allows to turn on/off the "powerful" mode through `specialmode-powerful` channel
* Does not allow setting the mode to eco 
* Does not allow setting the streamer mode on/off

This PR introduced the following changes:
* Split Streamer and Normal/Eco/Powerful mode into two separate and independent channels
* The `specialmode` channel now only reports NORMAL/ECO/POWERFUL, removing Streamer from this channel. This greatly simplifies the usage for this channel.
* The `specialmode` channel is changed from Read-only to Read/write
* The `specialmode-powerful` channel is removed, since setting the powerful mode can now be done through the `specialmode` channel
* A new read/write channel `streamer` is introduced to set the streamer mode

This makes the two features (streamer and normal/eco/powerful) more straight forward to deal with.

The issue / feature was reported/requested and tested here: 
https://community.openhab.org/t/daikin-the-channel-specialmode-can-be-read-but-not-modified-from-oh-both-2-5-12-and-3-1/125561


bundles/org.openhab.binding.daikin/src/main/java/org/openhab/binding/daikin/internal/api/Enums.java
```diff
@@ -148,65 +148,63 @@ public String getValue() {
         OFF(""),
         UNKNOWN("??");
 
-        private static final Logger LOGGER = LoggerFactory.getLogger(SpecialMode.class);
+        private static final Logger logger = LoggerFactory.getLogger(AdvancedMode.class);
```
lolodomo, line 151: As it is a static, the good practice us uppercase so LOGGER was good.

bundles/org.openhab.binding.daikin/src/main/java/org/openhab/binding/daikin/internal/handler/DaikinBaseHandler.java
```diff
@@ -92,6 +92,10 @@ public DaikinBaseHandler(Thing thing, DaikinDynamicStateDescriptionProvider stat
 
     @Override
     public void handleCommand(ChannelUID channelUID, Command command) {
+        if (webTargets == null) {
+            logger.error("Unable to handle commands. Check Daikin binding configurations");
```
lolodomo, line 96: Rather WARN level

lolodomo (changes requested):

lolodomo:
This is a breaking change to be explained in release notes.

bundles/org.openhab.binding.daikin/src/main/java/org/openhab/binding/daikin/internal/handler/DaikinBaseHandler.java
jimtng: if `webTargets` is null the binding is not going to work at all. Shouldn't that warrant an ERROR level? It is an exceptional condition that shouldn't be encountered in the normal operation, even when the binding has been misconfigured, this condition should still not occur. I don't even know how to cause webTargets to become null, if OSGI works properly.

bundles/org.openhab.binding.daikin/src/main/java/org/openhab/binding/daikin/internal/api/Enums.java
jimtng: > This is a breaking change to be explained in release notes.

Where is the "Release notes"? Is it something for the openhab maintainer (i.e. not me) to do?

jimtng:
> This is a breaking change to be explained in release notes.

Where is the "Release notes"? Is it something for the openhab maintainer (i.e. not me) to do?

bundles/org.openhab.binding.daikin/src/main/java/org/openhab/binding/daikin/internal/api/Enums.java
jimtng: fixed

bundles/org.openhab.binding.daikin/src/main/java/org/openhab/binding/daikin/internal/handler/DaikinBaseHandler.java
jimtng: However, the log message should be changed to clarify the situation though

jimtng: I've changed this to warn. It seems that I should never use `logger.error` in any case, within a binding.

lolodomo (approved):
LGTM

lolodomo:
> Where is the "Release notes"? Is it something for the openhab maintainer (i.e. not me) to do?

Look at this example: https://github.com/openhab/openhab-distro/pull/1369

PR #12325, opened 2022-02-20
[shelly] TRV+Motion2 support, various fixes & enhancements

markus7017:
Includes
- Shelly TRV support incl. new channels
- Shelly Motion 2 support (like Motion 1 + temp channel)
- Translations updated
- Meta data added to XML (category etc.)
- README updated
- Several bugfixes


hmerk:
@markus7017 As I learned, milestones are added by maintainers after merge, therefore I removed it.

mhupfauer:
Is there an ETA when this will be merged?

bundles/org.openhab.binding.shelly/src/main/java/org/openhab/binding/shelly/internal/handler/ShellyBaseHandler.java
```diff
@@ -158,6 +158,9 @@ public void initialize() {
                 updateStatus(ThingStatus.UNKNOWN, ThingStatusDetail.CONFIGURATION_PENDING,
                         messages.get("status.unknown.initializing"));
                 start = initializeThing();
+
+                // Promote Shelly Manager usage
+                logger.info("{}: {}", thingName, messages.get("status.managerstarted", localIP, localPort));
```
fwolter, line 163: Logging to info should be used sparsely e.g. a newly started component or a user file that has been loaded. This could be debug. See https://next.openhab.org/docs/developer/guidelines.html#f-logging

bundles/org.openhab.binding.shelly/src/main/java/org/openhab/binding/shelly/internal/handler/ShellyBaseHandler.java
```diff
@@ -344,19 +347,49 @@ public void handleCommand(ChannelUID channelUID, Command command) {
...
+                    int value = (int) getNumber(command);
                     value = value > 0 ? Math.max(SHELLY_MOTION_SLEEPTIME_OFFSET, value - SHELLY_MOTION_SLEEPTIME_OFFSET)
                             : 0;
                     api.setSleepTime(value);
                     break;
+                case CHANNEL_CONTROL_PROFILE:
+                    logger.debug("{}: Select profile {}", thingName, command);
+                    int profile = (int) getNumber(command);
+                    if (profile < 0 || profile > 5) {
+                        logger.info("{}: Invalid profile Id {} requested", thingName, profile);
```
fwolter, line 359: See above. Logging to Info. Please check all.

bundles/org.openhab.binding.shelly/src/main/java/org/openhab/binding/shelly/internal/handler/ShellyBaseHandler.java
```diff
@@ -486,6 +533,8 @@ public void setThingOnline() {
             updateStatus(ThingStatus.ONLINE);
 
             // request 3 updates in a row (during the first 2+3*3 sec)
+            logger.debug("{}: Thing went online, request status update", thingName);
+            ;
```
fwolter, line 537: Can the single semicolon be removed?

bundles/org.openhab.binding.shelly/src/main/java/org/openhab/binding/shelly/internal/handler/ShellyComponents.java
```diff
@@ -288,8 +297,44 @@ public static boolean updateSensors(ShellyBaseHandler thingHandler, ShellySettin
...
+                    updated |= thingHandler.updateChannel(CHANNEL_GROUP_SENSOR, CHANNEL_SENSOR_TEMP,
+                            toQuantityType(temp.doubleValue(), DIGITS_TEMP, SIUnits.CELSIUS));
+                    temp = convertToC(t.targetTemp.value, getString(t.targetTemp.unit));
+                    updated |= thingHandler.updateChannel(CHANNEL_GROUP_CONTROL, CHANNEL_CONTROL_SETTEMP,
+                            /* t.targetTemp.enabled ? */toQuantityType(t.targetTemp.value, DIGITS_TEMP, SIUnits.CELSIUS)
+                    /*
+                     * toQuantityType(t.targetTemp.value, DIGITS_NONE,
+                     * Units.PERCENT)
+                     * : UnDefType.UNDEF
+                     */);
```
fwolter, line 330: Can the comments be removed?

bundles/org.openhab.binding.shelly/src/main/resources/OH-INF/i18n/shelly_de.properties
```diff
@@ -1,53 +1,507 @@
+# binding
```
fwolter, line 1: @openhab/add-ons-maintainers There are quite a lot changes in the translation in this PR. If I understood correctly, translations should only be handled in Crowdin, otherwise they will be overwritten in the repo. How do we handle this?

fwolter (changes requested):

bundles/org.openhab.binding.shelly/src/main/resources/OH-INF/i18n/shelly_de.properties
markus7017: Most of the translations are even no changed. I moved all str resources from xml to the .properties file. Changes must not be overwritten depending on tooling. Please advise how to proceed

lolodomo: First, the Shelly binding is one of the bindings for which i18n is not yet properly implemented.
What is done here is surprising. I am not a fan of replacing all labels/descriptions in XML by `@text` entries but this is a way to do. What is strange here is that it looks like that `@text` entries look like the default paths that will have been considered even without the `@text`. So it was relatively useless to update all the XML.
**I would suggest to undo these changes from this PR and do it in a separate PR.** This can be done automatically by simply run a maven command. Using this command has the big advantage as it can be rerun later to simply add a new entry.
And regarding the German translations, it is better to update them in Crowdin in a second time. Because if the PR is merged and no manual action is done in Crowdin, there is a risk to loose them in the future.  An exception could be done as this is more or less a first massive injection of translations.

My suggestion: separate PR and simply use the maven command to generate the default translations.

mschoenbein:
Is there an ETA when this will be merged? Wait for TVR support.

hmerk:
> Is there an ETA when this will be merged? Wait for TVR support.

No, in general, we don't have ETAs, as all maintainers and contributors are doing this on voluntary basis in their spare time.
If you want to use the feature right now follow the Shelly binding topic in our community and install the latest dev version, like I do. TRV support is woking great.

bundles/org.openhab.binding.shelly/src/main/java/org/openhab/binding/shelly/internal/handler/ShellyBaseHandler.java
markus7017: Many people don't recognize the Shelly Manager UI, because OH doesn't provide an interface for UI extensions provided by the binding (as already requested a long time ago) and many don't reads the manual (RTFM). Therefore I'm adding this info to the log to promote this useful feature.

bundles/org.openhab.binding.shelly/src/main/java/org/openhab/binding/shelly/internal/handler/ShellyBaseHandler.java
markus7017: Doesn't it make use to make the user aware of a mis-configuration. Otherwise nothing happens and there is no indication in the log with default config. How to find this out without a log message?

bundles/org.openhab.binding.shelly/src/main/java/org/openhab/binding/shelly/internal/handler/ShellyBaseHandler.java
markus7017: fixed

bundles/org.openhab.binding.shelly/src/main/java/org/openhab/binding/shelly/internal/handler/ShellyComponents.java
markus7017: done

bundles/org.openhab.binding.shelly/src/main/resources/OH-INF/i18n/shelly_de.properties
markus7017: - " is not yet properly implemented." - where is a proper way described. 
- "This can be done automatically by simply run a maven command. " - which one?
- "And regarding the German translations, it is better to update them in Crowdin in a second time. " - looking to the volume of translations this is a huge effort. Where is the Crowdin integration described? Where documented that this is a formal requirement?
- "Yes please"
- "My suggestion: separate PR" - again, a huge effort. Everything has to be done manually. I think we should work on another one to make the move to CrowedIn.

bundles/org.openhab.binding.shelly/src/main/java/org/openhab/binding/shelly/internal/handler/ShellyBaseHandler.java
fwolter: If this is supposed to warn the user something is wrong, this should be WARN.

bundles/org.openhab.binding.shelly/src/main/resources/OH-INF/i18n/shelly_de.properties
ijasan: I supported @markus7017 on the translations, I could also support on moving resources to CrowdIn, but I also recommend the exception for this PR to avoid blocking the process. If the @text statements are not nessesary we could remove them, but maybe @fwolter could also proceed.

bundles/org.openhab.binding.shelly/src/main/java/org/openhab/binding/shelly/internal/handler/ShellyBaseHandler.java
markus7017: ok, I reviewed and adjusted all log levels in this module

bundles/org.openhab.binding.shelly/src/main/resources/OH-INF/i18n/shelly_de.properties
markus7017: @lolodomo any advice?

markus7017: @fwolter @lolodomo who do we solve this issue

markus7017:
@fwolter I'm ready to continue

bundles/org.openhab.binding.shelly/src/main/resources/OH-INF/i18n/shelly_de.properties
fwolter: I'd suggest to populate Crowding correctly once this is merged. Also because @ijasan offered his support.

fwolter (approved):

PR #12329, opened 2022-02-20
[bluetooth] Add missing default translations

wborn:
This will allow for translating the Bluetooth Binding in Crowdin.


lolodomo (approved):
LGTM

PR #12330, opened 2022-02-20
[hdpowerview] Let the framework change the child Thing statuses

fwolter:
@jlaur @lolodomo @andrewfg If you set the Thing status of the bridge and the child Things to UNKNOWN in `initialize()` like in this PR, the framework takes care of the statuses of the child Things.

I tested this code and the behavior is now, after enabling the Bridge it will stay in UNKNOWN and will eventually turn to OFFLINE when the service is not reachable. The child Things will change to BRIDGE_OFFLINE. I couldn't test the ONLINE transition.


jlaur:
@fwolter, interesting - thanks! Looks like setting the bridge status to **unknown** is what I missed when doing #12259. I'm eager to test this, but won't be home until late Wednesday. If @andrewfg will not verify sooner than that, I'll have a closer look Thursday.

lolodomo:
The current code matches the proper pattern required historically because a thing could be initiated before its bridge.
But then the core framework was enhanced and maybe it is now safe to do it like you propose.

jlaur (approved):
Thanks! Tested with offline transition as well as online transition, and works correctly (with core 3.2).

PR #12332, opened 2022-02-20
Remove some of the empty XML tag "description"

lolodomo:
Signed-off-by: Laurent Garnier <lg.hc@free.fr>

PR #12317, opened 2022-02-18
[googlestt] lazy abort

GiviMAD:
Signed-off-by: Miguel Álvarez Díez <miguelwork92@gmail.com>

Do not force close background task.

@lolodomo, for reference, this is the change I was talking about in the VoskSTT PR https://github.com/openhab/openhab-addons/pull/12249#discussion_r807709755

PR #12334, opened 2022-02-20
[flicbutton] Fix SAT `@NonNull` findings

wborn:
SAT is right. :slightly_smiling_face: 

See: https://github.com/openhab/static-code-analysis/issues/431


pfink:
LGTM, thanks!

pfink (approved):

lolodomo (approved):
LGTM

wborn:
Hi @lolodomo, I think it is better to not label this PR as "enhancement" because it will end up in the release notes and I doubt end users get very excited about null-annotations enhancements. :wink:

For the same reasons I also would not label https://github.com/openhab/openhab-addons/pull/12332 as enhancement so it ends up in the release notes, because it doesn't change any end user functionality.

I'd also remove it from https://github.com/openhab/openhab-addons/pull/12317, https://github.com/openhab/openhab-addons/pull/12318, because these are new add-ons in 3.3 so nobody ever ran into this if they upgrade from 3.2 to 3.3. :wink: 

CC: @openhab/add-ons-maintainers

PR #9180, opened 2020-11-30
[FreeboxOs] Introducing new version of the Freebox binding.

clinique:
As this version introduces many many modifications, it is presented as a new binding, targetted to repleace the current Freebox Binding in the future.

Signed-off-by: clinique <gael@lhopital.org>

PR #11833, opened 2021-12-21
pom reference update from http to https

marcelrv:
This fixes issue in eclipse where pom appears to have an error. Replacing the http with https fixes the issue.

note:  only the 2nd url changed, this is in line with the examples on the maven site https://maven.apache.org/pom.html#the-basics

image

Signed-off-by: Marcel Verpaalen <marcel@verpaalen.com>

PR #12342, opened 2022-02-21
[Freeboxos] New binding alternative to Freebox binding

clinique:
Supersedes [this PR](https://github.com/openhab/openhab-addons/pull/9180)


clinique:
@lolodomo : I propose that we come back on this one once the v3.3 is out.

bundles/org.openhab.binding.freeboxos/pom.xml
```diff
@@ -0,0 +1,17 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+  xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+
+  <modelVersion>4.0.0</modelVersion>
+
+  <parent>
+    <groupId>org.openhab.addons.bundles</groupId>
+    <artifactId>org.openhab.addons.reactor.bundles</artifactId>
+    <version>3.3.0-SNAPSHOT</version>
```
wborn, line 10:
```suggestion
    <version>3.4.0-SNAPSHOT</version>
```

jlaur:
@clinique - for a PR of this size, perhaps you can add some more context/description to the PR description? Also, without this context, the first question for me would be: Why introduce it as a new binding instead of updating/replacing the existing one - i.e., what are the differences and where do they overlap? Cc @lolodomo.

clinique:
> @clinique - for a PR of this size, perhaps you can add some more context/description to the PR description? Also, without this context, the first question for me would be: Why introduce it as a new binding instead of updating/replacing the existing one - i.e., what are the differences and where do they overlap? Cc @lolodomo.

@jlaur : answer is [here](https://github.com/openhab/openhab-addons/pull/9180#issuecomment-890450826)

bundles/org.openhab.binding.freeboxos/pom.xml
clinique: Done

lolodomo:
We even have a discussion about that and this was my proposal. The new binding has many new features but also has some missing features compared to the current binding and breaking changes if I correctly remember. The idea was to remove the old binding the day the new binding can fully replace the old binding, keeping the two bindings in the distribution during 6 months.
@jlaur : note that I already did a partial review. I even tested the binding and push code to @clinique. Several things were not good but @clinique certainly corrected them during the time.
This was in my TODO list to review again and test it but you are welcome @jlaur to make it too.

clinique:
Binding now published in the add-on market place.

jlaur:
> We even have a discussion about that and this was my proposal.

OK. What about the naming, where does the "OS" come from?

> The new binding has many new features but also has some missing features compared to the current binding and breaking changes if I correctly remember.

Would it be possible to implement those missing features, or are they somehow obsoleted? Otherwise they will still disappear when the old binding is finally removed.

> The idea was to remove the old binding the day the new binding can fully replace the old binding, keeping the two bindings in the distribution during 6 months.

I get that migration and breaking changes is always an issue with openHAB. Let me just brainstorm a little - sorry if you already went through those discussions previously: We now have Community Marketplace, which is sometimes used for pushing new binding versions before they make it into the openHAB distribution. Perhaps it could be used the other way around: Publishing the old "legacy" Freebox binding to marketplace. Then users needing a longer migration period and immediate compatibility when upgrading openHAB could simply uninstall the binding and install the legacy version from marketplace instead. This could be mentioned in the release notes for breaking changes. Would it work - and WDYT? This way the new version could still replace the old one without any naming issues, i.e. it could still be named "freebox".

The general idea is that the current approach is to deliver a "legacy" version and a new version through the same openHAB distribution channel, where binding names must be unique. But it doesn't have to be that way, at least some other options could be explored.

Secondly: Although I'm not sure if the new Netatmo binding was a success story in terms of migration, is the situation any different from that? I know we were forced to deliver a new version due to API changes, so it would eventually have broken anyway, but still interested in why you are seeking a different approach for Freebox compared to Netatmo.

@kaikreuzer - tagging you in case you would be interested as they are also some general issues about binding migrations, breaking changes and approaches to this, and you might also have some feedback.

> @jlaur : note that I already did a partial review. I even tested the binding and push code to @clinique. Several things were not good but @clinique certainly corrected them during the time.
> This was in my TODO list to review again and test it but you are welcome @jlaur to make it too.

I was really hoping you would resume the review, @lolodomo. :-) I might have a quick look and post some lose comments here and there if you are interested, but currently I will probably not commit fully to reviewing this PR.

bundles/org.openhab.binding.freeboxos/README.md
```diff
@@ -0,0 +1,177 @@
+# Freebox Binding
```
jlaur, line 1: I guess it should be:
```suggestion
# FreeboxOS Binding
```

? But please see latest PR comments before changing anything.

bundles/org.openhab.binding.freeboxos/README.md
```diff
@@ -0,0 +1,177 @@
...
+
+## Discovery
+
+The API bridge is discovered automatically through mDNS in the local network.
+After the bridge is discovered and available to openHAB, the binding will automatically discover phone, network devices available on the local network.
+Note that the discovered thing will be setup to use only HTTP API (and not HTTPS) because only an IP can be automatically determined while a domain name is required to use HTTPS with a valid certificate.
+
+## Binding configuration
+
+FreeboxOs binding has the following configuration parameters:
```
jlaur, line 31: FreeboxOs or FreeboxOS?

clinique:
According to the documentation : 
image

It should be spelled FreeboxOS

lolodomo:
> We now have Community Marketplace, which is sometimes used for pushing new binding versions before they make it into the openHAB distribution. Perhaps it could be used the other way around: Publishing the old "legacy" Freebox binding to marketplace. Then users needing a longer migration period and immediate compatibility when upgrading openHAB could simply uninstall the binding and install the legacy version from marketplace instead. This could be mentioned in the release notes for breaking changes. Would it work - and WDYT?

That looks like a good idea. I think the marketplace was not yet existing when we had this discussion with @clinique ;) But in that case, I consider that at least the new binding should not remove any feature from the legacy binding. At least, we should discuss each of them.

The problem with updating the current binding is that there are so many changes that the review is very hard. This is almost as if all the previous code is replaced by new code.

jlaur:
> That looks like a good idea. I think the marketplace was not yet existing when we had this discussion with @clinique ;) But in that case, I consider that at least the new binding should not remove any feature from the legacy binding. At least, we should discuss each of them.

That probably applies no matter how the migration is approached, assuming final action still would be to remove the "legacy" binding?

> The problem with updating the current binding is that there are so many changes that the review is very hard. This is almost as if all the previous code is replaced by new code.

That's a separate concern. I'm sure we'll manage as reviewers to not look too much at diffs, but concentrate on the new code itself.

> So having a new binding named freeboxos will help everybody, the reviewers and the final users.

Assuming the binding rename was out of necessity in order to have both bindings exist simultaneously, this contradicts the idea of providing the legacy binding through Marketplace. So I'd propose to either:
- Provide legacy binding through Marketplace and keep the existing binding name when providing the new version, i.e. have only one binding included in the openHAB distribution.
- Or go ahead with your original plan to provide the new version with a new name and keep the legacy binding, i.e. have two bindings for the same purpose and with overlapping functionality in order to work-around issues with versioning/migration/breaking changes.

kaikreuzer:
> @kaikreuzer - tagging you in case you would be interested as they are also some general issues about binding migrations, breaking changes and approaches to this, and you might also have some feedback.

Thanks, @jlaur.

I agree that we should try to only have one version of a binding in the official distro at any time. The marketplace sounds like the best option to make alternative versions available. So once this PR is merged, the "old" binding should be moved to the marketplace. Wrt the name, I personally do not care too much - I think we could leave the name of the new one to be "freeboxos" as it is anyhow not compatible and people will have to recreate their things, if I understand it correctly.
It might then even have the benefit that the old binding (then from the marketplace) could be installed by users in parallel (e.g. during migration), since both would have a different namespace and would not interfere with each other.

bundles/org.openhab.binding.freeboxos/pom.xml
```diff
@@ -0,0 +1,17 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+  xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
```
jlaur, line 3:
```suggestion
  xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd">
```

See #11833

clinique: Thanks for identifying this @jlaur . Will submit correction right now.

bundles/org.openhab.binding.freeboxos/pom.xml
```diff
@@ -0,0 +1,17 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+  xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd">
+
+  <modelVersion>4.0.0</modelVersion>
+
+  <parent>
+    <groupId>org.openhab.addons.bundles</groupId>
+    <artifactId>org.openhab.addons.reactor.bundles</artifactId>
+    <version>3.4.0-SNAPSHOT</version>
```
wborn, line 10:
```suggestion
    <version>4.0.0-SNAPSHOT</version>
```

clinique:
@lolodomo : when you'll have a minute, I would welcome a review. I would really like have this merged in OH4.0

clinique:
@ben12 : please take a look at this last version. I made huge modifications on the Home Node side - resulting in an important simplification of the configuration of Nodes.

ben12:
> @ben12 : please take a look at this last version. I made huge modifications on the Home Node side - resulting in an important simplification of the configuration of Nodes.

@clinique Code seems OK. (but I have not tested, I tried to run OpenHab 4 into Eclipse IDE without success)

clinique:
@lolodomo : sorry to ping you here - I would really like to have a chance to merge it before OH4 goes final.

ben12:
> @ben12 : please take a look at this last version. I made huge modifications on the Home Node side - resulting in an important simplification of the configuration of Nodes.

@clinique 
Finally, my "basic" shutter works with the last version 4.0.0-SNAPSHOT of openhab-core / freeboxos.

But check this:
https://github.com/clinique/openhab-addons/commit/9f518ff5eb27d374213a05ad1bad459732fb875e#r111107027

bundles/org.openhab.binding.freeboxos/src/main/java/org/openhab/binding/freeboxos/internal/config/ApiConsumerConfiguration.java
```diff
@@ -0,0 +1,31 @@
...
+ */
+@NonNullByDefault
+public class ApiConsumerConfiguration {
+    // public static final String PASSWORD = "password";
+    // public static final String PORT = "port";
+    public static final String REFRESH_INTERVAL = "refreshInterval";
+
+    public int refreshInterval = 30;
+    public String password = "";
+    // public int port = 24322;
```
fwolter, line 29: Can the comments be removed?

fwolter (changes requested):
I reviewed 63 / 118 files for now.

bundles/org.openhab.binding.freeboxos/README.md
```diff
@@ -0,0 +1,204 @@
...
+| Freebox Server Address   | apiDomain         | The domain to use in place of hardcoded Freebox ip     | No       | mafreebox.freebox.fr |
+| Application Token        | appToken          | Token generated by the Freebox Server.                 | Yes      |                      |
+| Network Device Discovery | discoverNetDevice | Enable the discovery of network device things.         | No       | false                |
+| HTTPS Available          | httpsAvailable    | Tells if https has been configured on the Freebox      | No       | false                |
+| HTTPS port               | httpsPort         | Port to use for remote https access to the Freebox Api | No       | 15682                |
+
+If the parameter *apiDomain* is not set, the binding will use the default address used by Free to access your Freebox Server (mafreebox.freebox.fr).
+The bridge thing will initialize only if a valid application token (parameter *appToken*) is filled.
+
+### Server : Revolution or Delta
```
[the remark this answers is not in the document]
clinique: Done.

bundles/org.openhab.binding.freeboxos/README.md
```diff
@@ -0,0 +1,204 @@
+# FreeboxOS Binding
+
+This binding integrates the [Freebox Revolution](https://www.free.fr/freebox/freebox-revolution/) and [Freebox Delta](https://www.free.fr/freebox/freebox-delta/) to your openHAB installation.
```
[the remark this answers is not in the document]
clinique: Done

bundles/org.openhab.binding.freeboxos/README.md
```diff
@@ -0,0 +1,204 @@
...
+FreeboxOS binding has the following configuration parameters:
+
+| Name            | Description                                        | Mandatory |
+|-----------------|----------------------------------------------------|-----------|
+| timeout         | The timeout for reading from the device in seconds | yes       |
+
+
+## Thing Configuration
+
+### Api bridge
```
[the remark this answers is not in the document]
clinique: Done

bundles/org.openhab.binding.freeboxos/README.md
```diff
@@ -0,0 +1,204 @@
...
+The *host* and *wifihost* things requires the following configuration parameters:
+
+| Parameter Label  | Parameter ID    | Description                                                            | Required | Default |
+|------------------|-----------------|------------------------------------------------------------------------|----------|---------|
+| MAC Address      | macAddress      | The MAC address of the network host .                                  | Yes      |         |
+| Refresh Interval | refreshInterval | The refresh interval in seconds which is used to poll for phone state. | No       | 30      |
+
+### Repeater and Vm thing 
+
+The *repeater* thing is a specialized case of a *wifihost*. The *vm* derives from *host*. They share the same configuration definition :
```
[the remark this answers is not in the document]
clinique: Done

bundles/org.openhab.binding.freeboxos/README.md
```diff
@@ -0,0 +1,204 @@
...
+
+| Parameter Label  | Parameter ID    | Description                                                      | Required | Default |
+|------------------|-----------------|------------------------------------------------------------------|----------|---------|
+| MAC Address      | macAddress      | The MAC address of the player device.                            | No       |         |
+| ID               | id              | Id of the repeater within Freebox Api                            | Yes      | 1       |
+| Refresh Interval | refreshInterval | The refresh interval in seconds which is used to poll the player | Yes      | 30      |
+
+### Basic shutter thing
+
+The *basic_shutter* thing requires the following configuration parameters:
```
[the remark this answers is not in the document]
clinique: Ok, understood changed accordingly. Replacing also `active_player` by `active-player`

bundles/org.openhab.binding.freeboxos/README.md
```diff
@@ -0,0 +1,204 @@
...
+|------------------|-----------------|------------------------------------------------------------------|----------|---------|
+| ID               | id              | Id of the Home Node                                              | Yes      | 1       |
+| UP Slot ID       | upSlotId        | Id of the UP Slot.                                               | Yes      | 0       |
+| STOP Slot ID     | stopSlotId      | Id of the STOP Slot.                                             | Yes      | 1       |
+| DOWN Slot ID     | downSlotId      | Id of the DOWN Slot.                                             | Yes      | 2       |
+| STATE Signal ID  | stateSignalId   | Id of the STATE Signal.                                          | Yes      | 3       |
+
+## Authentication
+
+You will have to authorize openHAB to connect to your Freebox. Here is the process described :
```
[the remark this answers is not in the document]
clinique: Done

bundles/org.openhab.binding.freeboxos/README.md
```diff
@@ -0,0 +1,204 @@
...
+| call          | accepted#number             | Call          | R           | Last accepted call number                                                      |
+| call          | accepted#duration           | Number        | R           | Last accepted call duration in seconds                                         |
+| call          | accepted#timestamp          | DateTime      | R           | Last accepted call creation timestamp                                          |
+| call          | accepted#name               | String        | R           | Last accepted caller name                                                      |
+| call          | missed#number               | Call          | R           | Last missed call number                                                        |
+| call          | missed#timestamp            | DateTime      | R           | Last missed call creation timestamp                                            |
+| call          | missed#name                 | String        | R           | Last missed  caller name                                                       |
```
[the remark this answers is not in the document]
clinique: Done

bundles/org.openhab.binding.freeboxos/README.md
```diff
@@ -0,0 +1,204 @@
...
+| net_interface | reachable                   | Switch        | R           | Indicates whether the network interface is reachable                           |
+| airplay       | playurl                     | String        | W           | Play an audio or video media from the given URL                                |
+| airplay       | stop                        | Switch        | W           | Stop the media playback                                                        |
+| basic_shutter | basic-shutter#basic-shutter | RollerShutter | W           | Up, stop and down commands for a RTS shutter                                   |
+
+(*) : server means *delta* or *revolution*
+
+## Actions for rules
+
+The following actions are available in rules/scripting:
```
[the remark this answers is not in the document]
clinique: Done

bundles/org.openhab.binding.freeboxos/pom.xml
```diff
@@ -0,0 +1,30 @@
...
+    <version>4.0.0-SNAPSHOT</version>
+  </parent>
+
+  <artifactId>org.openhab.binding.freeboxos</artifactId>
+
+  <name>openHAB Add-ons :: Bundles :: FreeboxOS Binding</name>
+
+  <dependencies>
+    <dependency>
+      <groupId>com.google.code.gson</groupId>
```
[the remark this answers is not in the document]
clinique: Should be but I used version 2.10 of gson that accepts deserialization of records. There's an [open issue](https://github.com/openhab/openhab-core/issues/3321) on this in core

bundles/org.openhab.binding.freeboxos/pom.xml
```diff
@@ -0,0 +1,30 @@
...
+      <artifactId>gson</artifactId>
+      <version>2.10.1</version>
+    </dependency>
+    <dependency>
+      <groupId>com.github.seancfoley</groupId>
+      <artifactId>ipaddress</artifactId>
```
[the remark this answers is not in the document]
clinique: Done

bundles/org.openhab.binding.freeboxos/src/main/java/org/openhab/binding/freeboxos/internal/FreeboxOsHandlerFactory.java
```diff
@@ -0,0 +1,166 @@
...
+
+        callbackURL = (String) config.getOrDefault(CALLBACK_URL, "");
+        int port = HttpServiceUtil.getHttpServicePort(bundleContext);
+        if (callbackURL.isEmpty() && port != -1) {
+            String openHabIp = Objects.requireNonNull(networkAddressService.getPrimaryIpv4HostAddress());
+            // we do not use SSL as it can cause certificate validation issues.
+            callbackURL = "http://%s:%d".formatted(openHabIp, port);
+        }
+        if (callbackURL.isEmpty()) {
+            logger.info("Unable to build a correct call back URL to stream media contents");
```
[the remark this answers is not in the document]
clinique: Correct ! and corrected

bundles/org.openhab.binding.freeboxos/src/main/java/org/openhab/binding/freeboxos/internal/api/ApiHandler.java
```diff
@@ -0,0 +1,146 @@
...
+            logger.trace("executeUrl {} - {} returned {}", method, uri, content);
+            T result = deserialize(clazz, content);
+            if (statusCode == Code.OK) {
+                return result;
+            } else if (statusCode == Code.FORBIDDEN) {
+                logger.debug("Fobidden, serviceReponse was {}, ", content);
+                Response<?> error = (Response<?>) result;
+                throw new FreeboxException(error.getErrorCode(), error.getMsg());
+            }
+            throw new FreeboxException("Error '%s' requesting : %s", statusCode.getMessage(), uri.toString());
```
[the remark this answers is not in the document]
clinique: Done

bundles/org.openhab.binding.freeboxos/src/main/java/org/openhab/binding/freeboxos/internal/api/ApiHandler.java
```diff
@@ -0,0 +1,146 @@
...
+        } catch (InterruptedException | TimeoutException | ExecutionException e) {
```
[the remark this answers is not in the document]
clinique: Good to know, thanks for this. Changed accordingly.

bundles/org.openhab.binding.freeboxos/src/main/java/org/openhab/binding/freeboxos/internal/api/FreeboxOsIconProvider.java
```diff
@@ -0,0 +1,91 @@
...
+        URI uri = uriBuilder.clone().path(resourceName).build();
+        Request request = httpClient.newRequest(uri).method(HttpMethod.GET);
+
+        try {
+            ContentResponse response = request.send();
+            if (HttpStatus.getCode(response.getStatus()) == Code.OK) {
+                return new ByteArrayInputStream(response.getContent());
+            }
+        } catch (InterruptedException | TimeoutException | ExecutionException e) {
+        }
```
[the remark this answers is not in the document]
clinique: Sure, no. Changed

bundles/org.openhab.binding.freeboxos/src/main/java/org/openhab/binding/freeboxos/internal/api/rest/FreeboxOsSession.java
```diff
@@ -0,0 +1,207 @@
...
+
+    public void closeSession() {
+        Session currentSession = session;
+        if (currentSession != null) {
+            try {
+                getManager(WebSocketManager.class).closeSession();
+                getManager(LoginManager.class).closeSession();
+                session = null;
+            } catch (FreeboxException e) {
+                logger.info("Error closing session : {}", e.getMessage());
```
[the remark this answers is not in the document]
clinique: Changed

bundles/org.openhab.binding.freeboxos/src/main/java/org/openhab/binding/freeboxos/internal/api/rest/FreeboxOsSession.java
```diff
@@ -0,0 +1,207 @@
...
+        try {
+            T response = apiHandler.executeUri(uri, method, clazz, getSessionToken(), aPayload);
+            if (response.getErrorCode() == ErrorCode.INTERNAL_ERROR && retryCount > 0) {
+                return execute(uri, method, clazz, false, retryCount - 1, aPayload);
+            } else if (retryAuth && response.getErrorCode() == ErrorCode.AUTH_REQUIRED) {
+                initiateConnection();
+                return execute(uri, method, clazz, false, retryCount, aPayload);
+            }
+            if (!response.isSuccess()) {
+                throw new FreeboxException("Api request failed : %s", response.getMsg());
```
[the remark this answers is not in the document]
clinique: Changed

bundles/org.openhab.binding.freeboxos/src/main/java/org/openhab/binding/freeboxos/internal/api/rest/LoginManager.java
```diff
@@ -0,0 +1,159 @@
...
+    private Status trackAuthorize(int trackId) throws FreeboxException {
+        return getSingle(AuthStatus.class, AUTHORIZE_ACTION, Integer.toString(trackId)).status();
+    }
+
+    public String grant() throws FreeboxException {
+        Authorization authorize = post(new AuthorizeData(APP_ID, BUNDLE), AuthResponse.class, AUTHORIZE_ACTION);
+        Status track = Status.PENDING;
+        try {
+            while (Status.PENDING.equals(track)) {
+                Thread.sleep(2000);
```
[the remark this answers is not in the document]
clinique: Normally it should not lead to an endless loop because the box expires but lets be cautious and double the security on OH side also.

bundles/org.openhab.binding.freeboxos/src/main/java/org/openhab/binding/freeboxos/internal/api/rest/SystemManager.java
```diff
@@ -0,0 +1,103 @@
...
+public class SystemManager extends ConfigurableRest<SystemManager.Config, SystemManager.ConfigurationResponse> {
+
+    protected static class ConfigurationResponse extends Response<Config> {
+    }
+
+    public static record Sensor(String id, String name, int value) {
+        public enum SensorKind {
+            FAN("Vitesse"),
+            TEMP("Température"),
+            UNKNOWN("Uknown");
```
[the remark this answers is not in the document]
clinique: Useless, removed.

bundles/org.openhab.binding.freeboxos/src/main/java/org/openhab/binding/freeboxos/internal/api/rest/WebSocketManager.java
```diff
@@ -0,0 +1,201 @@
...
+        if (result.success) {
+            switch (result.action) {
+                case REGISTER:
+                    logger.debug("Event registration successfull");
+                    break;
+                case NOTIFICATION:
+                    handleNotification(result);
+                    break;
+                default:
+                    logger.info("Unhandled notification received : {}", result.action);
```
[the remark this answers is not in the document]
clinique: Yes, this is a call to expand deserialization capabilities - switched to warn.

bundles/org.openhab.binding.freeboxos/src/main/java/org/openhab/binding/freeboxos/internal/config/ApiConsumerConfiguration.java
clinique: Done

bundles/org.openhab.binding.freeboxos/src/main/java/org/openhab/binding/freeboxos/internal/handler/FreeDeviceIntf.java
```diff
@@ -0,0 +1,56 @@
...
+        }
+        return newUptime;
+    }
+
+    default void processReboot(Runnable actualReboot) {
+        triggerChannel(getEventChannelUID(), "reboot_requested");
+        actualReboot.run();
+        updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.DUTY_CYCLE, "System rebooting...");
+        stopRefreshJob();
+        getScheduler().schedule(this::initialize, 30, TimeUnit.SECONDS);
```
fwolter, line 54: This task needs to be cancelled in `dispose()`.

bundles/org.openhab.binding.freeboxos/src/main/java/org/openhab/binding/freeboxos/internal/handler/FreeboxOsHandler.java
```diff
@@ -0,0 +1,141 @@
...
+        super(thing);
+        this.session = session;
+        this.callbackURL = callbackURL;
+        this.bundleContext = bundleContext;
+        this.audioHTTPServer = audioHTTPServer;
+    }
+
+    @Override
+    public void initialize() {
+        logger.debug("Initializing Freebox OS API handler for thing {}.", getThing().getUID());
```
fwolter, line 65: See above. Can be removed.

bundles/org.openhab.binding.freeboxos/src/main/java/org/openhab/binding/freeboxos/internal/handler/FreeboxOsHandler.java
```diff
@@ -0,0 +1,141 @@
...
+                }
+
+                String currentAppToken = session.initialize(config);
+
+                if (!currentAppToken.equals(config.appToken)) {
+                    Configuration thingConfig = editConfiguration();
+                    thingConfig.put(FreeboxOsConfiguration.APP_TOKEN, currentAppToken);
+                    updateConfiguration(thingConfig);
+                    logger.info(
+                            "App token updated in Configuration, please give needed permissions to openHAB app in your Freebox management console");
```
fwolter, line 86: Shouldn't this be set as the Thing status detail message to be visible in the UI?

bundles/org.openhab.binding.freeboxos/src/main/java/org/openhab/binding/freeboxos/internal/handler/FreeboxOsHandler.java
```diff
@@ -0,0 +1,141 @@
...
+    }
+
+    private void freeConnectionJob() {
+        openConnectionJob.ifPresent(job -> job.cancel(true));
+        openConnectionJob = Optional.empty();
+    }
+
+    @Override
+    public void dispose() {
+        logger.debug("Disposing Freebox OS API handler for thing {}", getThing().getUID());
```
fwolter, line 109: See above. Remove.

bundles/org.openhab.binding.freeboxos/src/main/java/org/openhab/binding/freeboxos/internal/handler/FreeplugHandler.java
```diff
@@ -0,0 +1,102 @@
...
+
+    private void updateRateChannel(String channel, int rate) {
+        QuantityType<?> qtty = rate != -1 ? new QuantityType<>(rate, Units.MEGABIT_PER_SECOND) : null;
+        updateChannelQuantity(channel, qtty);
+    }
+
+    public void reset() {
+        try {
+            getManager(FreeplugManager.class).reboot(getMac());
+            logger.info("Freeplug {} succesfully restarted", getMac());
```
fwolter, line 92: This could be debug.

bundles/org.openhab.binding.freeboxos/src/main/java/org/openhab/binding/freeboxos/internal/handler/HomeNodeHandler.java
```diff
@@ -0,0 +1,134 @@
...
+            if (slotId instanceof Integer) {
+                try {
+                    EndpointState state = homeManager.getEndpointsState(getClientId(), slotId);
+                    if (state != null) {
+                        result = getChannelState(homeManager, channel.getUID().getIdWithoutGroup(), state);
+                    } else {
+                        result = getChannelState(homeManager, channel.getUID().getIdWithoutGroup());
+                    }
+                } catch (FreeboxException e) {
+                    logger.info("Error updating channel : {}", e.getMessage());
```
fwolter, line 91: This should be debug if it is self-healing or warn if it is fatal.

bundles/org.openhab.binding.freeboxos/src/main/java/org/openhab/binding/freeboxos/internal/handler/PlayerHandler.java
```diff
@@ -0,0 +1,114 @@
...
+    public void updateConnectivityChannels(LanHost host) {
+        super.updateConnectivityChannels(host);
+        ipAddress = host.getIpv4();
+    }
+
+    public void sendKey(String key, boolean longPress, int count) {
+        String aKey = key.toLowerCase();
+        IPAddress ip = ipAddress;
+        if (ip == null) {
+            logger.info("Player IP is unknown");
```
fwolter, line 86: This should be warn.

bundles/org.openhab.binding.freeboxos/src/main/java/org/openhab/binding/freeboxos/internal/handler/PlayerHandler.java
```diff
@@ -0,0 +1,114 @@
...
+                try {
+                    getManager(PlayerManager.class).sendKey(ip.toCanonicalString(), remoteCode, aKey, longPress, count);
+                } catch (FreeboxException e) {
+                    logger.info("Error sending key : {}", e.getMessage());
+                }
+            } else {
+                logger.warn("A remote code must be configured in the on the player thing.");
+            }
+        } else {
+            logger.info("Key '{}' is not a valid key expression", key);
```
fwolter, line 99: Warn.

bundles/org.openhab.binding.freeboxos/src/main/java/org/openhab/binding/freeboxos/internal/handler/PlayerHandler.java
```diff
@@ -0,0 +1,114 @@
...
+        IPAddress ip = ipAddress;
+        if (ip == null) {
+            logger.info("Player IP is unknown");
+        } else if (VALID_REMOTE_KEYS.contains(aKey)) {
+            String remoteCode = (String) getConfig().get(PlayerConfiguration.REMOTE_CODE);
+            if (remoteCode != null) {
+                try {
+                    getManager(PlayerManager.class).sendKey(ip.toCanonicalString(), remoteCode, aKey, longPress, count);
+                } catch (FreeboxException e) {
+                    logger.info("Error sending key : {}", e.getMessage());
```
fwolter, line 93: Debug or warn.

bundles/org.openhab.binding.freeboxos/src/main/resources/OH-INF/addon/addon.xml
```diff
@@ -0,0 +1,24 @@
...
+	xsi:schemaLocation="https://openhab.org/schemas/addon/v1.0.0 https://openhab.org/schemas/addon-1.0.0.xsd">
+
+	<type>binding</type>
+	<name>Freebox OS Binding</name>
+	<description>The Freebox OS binding integrates Free equipments in your home automation.</description>
+	<connection>local</connection>
+	<countries>fr,it</countries>
+
+	<config-description>
+		<parameter name="timeout" type="integer" required="false" min="1" max="30" unit="s">
```
fwolter, line 13: Is it reasonable to specify a max value?

bundles/org.openhab.binding.freeboxos/src/main/resources/OH-INF/i18n/freeboxos.properties
```diff
@@ -0,0 +1,354 @@
+# add-on
+
+addon.freeboxos.name = Freebox OS Binding
+addon.freeboxos.description = The Freebox OS binding integrates Free equipments in your home automation.
+
+# add-on config
+
+addon.config.freeboxos.callbackUrl.label = Callback URL
+addon.config.freeboxos.callbackUrl.description = URL to use for playing notification sounds hosted by openHab, e.g. 'http://192.168.0.2:8080'
```
fwolter, line 9:
```suggestion
addon.config.freeboxos.callbackUrl.description = URL to use for playing notification sounds hosted by openHAB, e.g. 'http://192.168.0.2:8080'
```

bundles/org.openhab.binding.freeboxos/src/main/resources/OH-INF/thing/channel-types.xml
```diff
@@ -0,0 +1,465 @@
...
+	<channel-type id="connection-status">
+		<item-type>String</item-type>
+		<label>Connection</label>
+		<description>Connection Status</description>
+		<state readOnly="true" pattern="%s"/>
+	</channel-type>
+
+	<channel-type id="hardware-status">
+		<item-type>String</item-type>
+		<label>Hardware status</label>
```
fwolter, line 389:
```suggestion
		<label>Hardware Status</label>
```

bundles/org.openhab.binding.freeboxos/src/main/resources/OH-INF/thing/channel-types.xml
```diff
@@ -0,0 +1,465 @@
...
+			<parameter name="slot" type="integer"/>
+			<parameter name="signal" type="integer"/>
+			<parameter name="stop" type="integer"/>
+		</config-description>
+	</channel-type>
+
+	<channel-type id="alarm-timeout">
+		<item-type>Number:Time</item-type>
+		<label>Alarm Duration</label>
+		<category>zone_temporisee</category>
```
fwolter, line 447: Are you sure this category exists?

bundles/org.openhab.binding.freeboxos/src/main/resources/OH-INF/thing/channel-types.xml
```diff
@@ -0,0 +1,465 @@
...
+		<state pattern="%d %unit%"/>
+	</channel-type>
+
+	<channel-type id="alarm-volume">
+		<item-type>Number:Dimensionless</item-type>
+		<label>Alarm Volume</label>
+		<category>sirene</category>
```
fwolter, line 454:
```suggestion
		<category>Siren</category>
```

bundles/org.openhab.binding.freeboxos/src/main/resources/OH-INF/thing/channel-types.xml
```diff
@@ -0,0 +1,465 @@
...
+		<state min="0" max="100" step="1" pattern="%d %unit%"/>
+	</channel-type>
+
+	<channel-type id="alarm-pin">
+		<item-type>String</item-type>
+		<label>PIN Code</label>
+		<category>pin_code</category>
```
fwolter, line 461: Are you sure this category exists? https://www.openhab.org/docs/concepts/categories.html

fwolter (changes requested):
I reviewed 100 / 118 files for now.

bundles/org.openhab.binding.freeboxos/src/main/java/org/openhab/binding/freeboxos/internal/api/rest/FreeboxOsSession.java
```diff
@@ -0,0 +1,210 @@
...
+
+    public void closeSession() {
+        Session currentSession = session;
+        if (currentSession != null) {
+            try {
+                getManager(WebSocketManager.class).closeSession();
+                getManager(LoginManager.class).closeSession();
+                session = null;
+            } catch (FreeboxException e) {
+                logger.info("Error closing session: {}", e.getMessage());
```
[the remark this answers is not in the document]
clinique: Done

bundles/org.openhab.binding.freeboxos/src/main/java/org/openhab/binding/freeboxos/internal/api/rest/WebSocketManager.java
```diff
@@ -0,0 +1,201 @@
...
+
+    @Override
+    public void onWebSocketClose(int statusCode, @NonNullByDefault({}) String reason) {
+        logger.debug("Socket Closed: [{}] - reason {}", statusCode, reason);
+        this.wsSession = null;
+    }
+
+    @Override
+    public void onWebSocketError(@NonNullByDefault({}) Throwable cause) {
+        logger.warn("Error on websocket : {}", cause.getMessage());
```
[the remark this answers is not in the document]
clinique: > Under which circumstances is the returned status PENDING?

It is returned as PENDING as long as the pairing process is ongoing, until success or timeout.

> I still see it as info.

Right. I changed it in a wrong place. Done.

> Please check all.
Done.

bundles/org.openhab.binding.freeboxos/src/main/java/org/openhab/binding/freeboxos/internal/handler/ApiConsumerHandler.java
```diff
@@ -0,0 +1,353 @@
...
+    private @Nullable ScheduledFuture<?> globalJob;
+    private @Nullable ServiceRegistration<?> reg;
+
+    ApiConsumerHandler(Thing thing) {
+        super(thing);
+    }
+
+    @Override
+    public void initialize() {
+        logger.debug("Initializing handler for thing {}", getThing().getUID());
```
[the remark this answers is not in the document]
clinique: Done

bundles/org.openhab.binding.freeboxos/src/main/java/org/openhab/binding/freeboxos/internal/handler/ApiConsumerHandler.java
```diff
@@ -0,0 +1,353 @@
...
+                    reg = bridgeHandler.getBundleContext().registerService(
+                            AudioSink.class.getName(), new AirMediaSink(this, bridgeHandler.getAudioHTTPServer(),
+                                    callbackURL, receiver.name(), config.password, config.acceptAllMp3),
+                            new Hashtable<>());
+                } else {
+                    logger.info("A password needs to be configured to enable Air Media capability.");
+                }
+            }
+        } catch (FreeboxException e) {
+            logger.warn("Unable to retrieve Media Receivers");
```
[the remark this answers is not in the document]
clinique: Done

bundles/org.openhab.binding.freeboxos/src/main/java/org/openhab/binding/freeboxos/internal/handler/ApiConsumerHandler.java
```diff
@@ -0,0 +1,353 @@
...
+            }
+        } else {
+            updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.BRIDGE_UNINITIALIZED);
+        }
+        return null;
+    }
+
+    @Override
+    public void dispose() {
+        logger.debug("Disposing handler for thing {}", getThing().getUID());
```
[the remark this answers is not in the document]
clinique: Removed.

bundles/org.openhab.binding.freeboxos/src/main/java/org/openhab/binding/freeboxos/internal/handler/ApiConsumerHandler.java
```diff
@@ -0,0 +1,353 @@
...
+                    }
+                }
+            }
+
+            globalJob = scheduler.scheduleWithFixedDelay(() -> {
+                try {
+                    internalPoll();
+                } catch (FreeboxException e) {
+                    logger.warn("Error polling thing {} : {}", getThing().getUID(), e.getMessage());
+                    updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.COMMUNICATION_ERROR, e.getMessage());
```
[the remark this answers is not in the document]
clinique: Done.

bundles/org.openhab.binding.freeboxos/src/main/java/org/openhab/binding/freeboxos/internal/handler/CallHandler.java
```diff
@@ -0,0 +1,109 @@
...
+        // Do not consider duration for Missed & incoming calls
+        if (lastType == Type.ACCEPTED || lastType == Type.OUTGOING) {
+            updateChannelQuantity(group, DURATION, call.duration(), Units.SECOND);
+        }
+    }
+
+    public void emptyQueue() {
+        try {
+            getManager(CallManager.class).emptyQueue();
+            logger.info("Call log succesfully cleared");
```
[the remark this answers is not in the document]
clinique: Removed.

bundles/org.openhab.binding.freeboxos/src/main/java/org/openhab/binding/freeboxos/internal/handler/FreeboxOsHandler.java
clinique: Done

bundles/org.openhab.binding.freeboxos/src/main/java/org/openhab/binding/freeboxos/internal/handler/FreeboxOsHandler.java
clinique: I can't make this a permanent status info. It's a reminder sent to the user so he does not fall in errors due to missing rights. In that case, he'll have errors raised when using parts of the binding mentioning that some rights are missing. BTW I'm going to document this in the README.

bundles/org.openhab.binding.freeboxos/src/main/java/org/openhab/binding/freeboxos/internal/handler/FreeboxOsHandler.java
clinique: Done.

bundles/org.openhab.binding.freeboxos/src/main/java/org/openhab/binding/freeboxos/internal/handler/FreeplugHandler.java
clinique: Changed

bundles/org.openhab.binding.freeboxos/src/main/java/org/openhab/binding/freeboxos/internal/handler/HomeNodeHandler.java
clinique: Changed.

bundles/org.openhab.binding.freeboxos/src/main/java/org/openhab/binding/freeboxos/internal/handler/PlayerHandler.java
clinique: Correct.

bundles/org.openhab.binding.freeboxos/src/main/java/org/openhab/binding/freeboxos/internal/handler/PlayerHandler.java
clinique: Done

bundles/org.openhab.binding.freeboxos/src/main/java/org/openhab/binding/freeboxos/internal/handler/PlayerHandler.java
clinique: Done

bundles/org.openhab.binding.freeboxos/src/main/resources/OH-INF/addon/addon.xml
clinique: You're correct

bundles/org.openhab.binding.freeboxos/src/main/resources/OH-INF/i18n/freeboxos.properties
clinique: Changed

bundles/org.openhab.binding.freeboxos/src/main/resources/OH-INF/thing/channel-types.xml
clinique: Changed

bundles/org.openhab.binding.freeboxos/src/main/resources/OH-INF/thing/channel-types.xml
clinique: Well, yes...and no.
The binding incorporates its own IconServer so correcting this to '<category>oh:freeboxos:zone_temporisee</category>' is good.

bundles/org.openhab.binding.freeboxos/src/main/resources/OH-INF/thing/channel-types.xml
clinique: Changed to `<category>oh:freeboxos:sirene</category>`

bundles/org.openhab.binding.freeboxos/src/main/resources/OH-INF/thing/channel-types.xml
clinique: Changed to `<category>oh:freeboxos:pin_code</category>`

bundles/org.openhab.binding.freeboxos/src/main/java/org/openhab/binding/freeboxos/internal/handler/FreeDeviceIntf.java
clinique: Correct, I reviewed the way I handle scheduledJobs globally.

bundles/org.openhab.binding.freeboxos/src/main/java/org/openhab/binding/freeboxos/internal/handler/ApiConsumerHandler.java
```diff
@@ -0,0 +1,353 @@
...
+            ThingStatusDetail detail = thing.getStatusInfo().getStatusDetail();
+            if (ThingStatusDetail.DUTY_CYCLE.equals(detail)) {
+                boolean rebooting = true;
+                while (rebooting) {
+                    try {
+                        internalPoll();
+                        rebooting = false;
+                    } catch (FreeboxException ignore) {
+                        try {
+                            Thread.sleep(20000);
```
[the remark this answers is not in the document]
clinique: Correct - I changed the approach here.

clinique:
@fwolter : you can proceed with review the PR build failed for a SAT issue.
mvn spotless:apply (as well as Eclipse IDE) keeps changing 
`
private·static·record·EndpointValue<T>·(T·value)·{
`
to this 

`
private·static·record·EndpointValue<T>(T·value)·{
`
mvn install does not complain but the CI/Build fails.

bundles/org.openhab.binding.freeboxos/src/main/java/org/openhab/binding/freeboxos/internal/api/rest/WebSocketManager.java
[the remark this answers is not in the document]
fwolter: Is the pairing process done once e.g. during configuration or does it need to be done regularily? Is there any chance you can return a try-again-later error if PENDING is returned?

fwolter:
I'm a bit puzzled, are you saying the CI build succeeds without the space, but the local build fails and with the space the CI build fails, but the local build succeeds?

In any case, building locally succeeds only with the space on my machine. But obviously the CI build succeeded.

Please note that the sign-off message needs to contain your real name. You can keep it for this PR.

The rest LGTM, except the remaining `Thread.sleep()`.

bundles/org.openhab.binding.freeboxos/src/main/java/org/openhab/binding/freeboxos/internal/api/rest/WebSocketManager.java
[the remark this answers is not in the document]
clinique: It is done only once at first install

clinique:
> I'm a bit puzzled, are you saying the CI build succeeds without the space, but the local build fails and with the space the CI build fails, but the local build succeeds?
> 
> In any case, building locally succeeds only with the space on my machine. But obviously the CI build succeeded.
> 
> Please note that the sign-off message needs to contain your real name. You can keep it for this PR.
> 
> The rest LGTM, except the remaining `Thread.sleep()`.

I reversed cases. Spotless:apply removes the space and CI breaks

bundles/org.openhab.binding.freeboxos/src/main/java/org/openhab/binding/freeboxos/internal/api/rest/WebSocketManager.java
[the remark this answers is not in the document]
fwolter: If I see correctly you call `Thread.sleep()` on every API interaction, if it's not authenticated, making it hard to wait asynchronously for the result of the interaction. Why not doing the authentication once, e.g. in `initialize()`. That would make it much easier to do the authentication asynchronously. Then, normal commands during runtime require only one API interaction and tend not to block for seconds.

fwolter:
Seems a bit strange. Before merging this, we should figure out why this is building locally, but the CI build breaks and vice versa.

bundles/org.openhab.binding.freeboxos/src/main/java/org/openhab/binding/freeboxos/internal/api/rest/WebSocketManager.java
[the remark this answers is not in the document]
clinique: Not exactly. If you're not authenticated you go this path, the binding will stay offline until this is satisfied, so no API call is made until the process is completed. Once authenticated you never get there.

clinique:
> Seems a bit strange. Before merging this, we should figure out why this is building locally, but the CI build breaks and vice versa.

I think this recent construct (generic record) is not well handled by spotless:apply. It's not a huge issue, noticeable and will have to be adressed at some point, unless I'm the only one facing it.

Before spotless: 
```
    private static record EndpointValue<T>(T value) {
    }
```
==> this compiles fine locally and in CI

after spotless:
```
    private static record EndpointValue<T> (T value) {
    }

```
Note the space inserted after the diamond operator ( and BTW Eclipse IDE does the same when saving file).

==> this compiles fine locally but not in CI

bundles/org.openhab.binding.freeboxos/src/main/java/org/openhab/binding/freeboxos/internal/api/rest/WebSocketManager.java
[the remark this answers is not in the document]
fwolter: That sounds like you could trigger the authentication process in `initialize()` and remove it in all API calls when the Thing is online. Did I miss something?

fwolter:
This doesn't build on my side (current Git HEAD):
```
    private static record EndpointValue<T>(T value) {
    }
```
```
PS C:\Users\Fabian Wolter\Documents\GitHub\openhab-addons\bundles\org.openhab.binding.freeboxos> mvn clean install
[INFO] Scanning for projects...
[INFO]
[INFO] Using the MultiThreadedBuilder implementation with a thread count of 64
[INFO]
[INFO] ------< org.openhab.addons.bundles:org.openhab.binding.freeboxos >------
[INFO] Building openHAB Add-ons :: Bundles :: FreeboxOS Binding 4.0.0-SNAPSHOT
[INFO] --------------------------------[ jar ]---------------------------------
[WARNING] The POM for org.eclipse.orbit.bundles:net.i2p.crypto.eddsa:jar:0.3.0.v20220506-1020 is missing, no dependency information available
[INFO]
[INFO] --- maven-clean-plugin:3.0.0:clean (default-clean) @ org.openhab.binding.freeboxos ---
[INFO] Deleting C:\Users\Fabian Wolter\Documents\GitHub\openhab-addons\bundles\org.openhab.binding.freeboxos\target
[INFO]
[INFO] --- maven-enforcer-plugin:3.0.0-M2:enforce (enforce-java) @ org.openhab.binding.freeboxos ---
[INFO]
[INFO] --- directory-maven-plugin:1.0:directory-of (directories) @ org.openhab.binding.freeboxos ---
[INFO] Directory of org.openhab.addons:org.openhab.addons.reactor set to: C:\Users\Fabian Wolter\Documents\GitHub\openhab-addons
[INFO]
[INFO] --- spotless-maven-plugin:2.28.0:check (codestyle_check) @ org.openhab.binding.freeboxos ---
[INFO] ------------------------------------------------------------------------
[INFO] BUILD FAILURE
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  5.400 s (Wall Clock)
[INFO] Finished at: 2023-06-25T21:17:02+02:00
[INFO] ------------------------------------------------------------------------
[ERROR] Failed to execute goal com.diffplug.spotless:spotless-maven-plugin:2.28.0:check (codestyle_check) on project org.openhab.binding.freeboxos: The following files had format violations:
[ERROR]     src\main\java\org\openhab\binding\freeboxos\internal\api\rest\HomeManager.java
[ERROR]         @@ -60,7 +60,7 @@
[ERROR]          ········UNKNOWN;
[ERROR]          ····}
[ERROR]
[ERROR]         -····private·static·record·EndpointValue<T>(T·value)·{
[ERROR]         +····private·static·record·EndpointValue<T>·(T·value)·{
[ERROR]          ····}
[ERROR]
[ERROR]          ····private·static·record·EndpointUi(AccessType·access,·DisplayType·display,·String·iconUrl,·@Nullable·String·unit)·{
[ERROR] Run 'mvn spotless:apply' to fix these violations.
[ERROR] -> [Help 1]
[ERROR]
[ERROR] To see the full stack trace of the errors, re-run Maven with the -e switch.
[ERROR] Re-run Maven using the -X switch to enable full debug logging.
[ERROR]
[ERROR] For more information about the errors and possible solutions, please read the following articles:
[ERROR] [Help 1] http://cwiki.apache.org/confluence/display/MAVEN/MojoExecutionException
```
I'm wondering why the CI build succeeds...

fwolter:
@jlaur could you have a quick look wether the current Git HEAD or the version with the space builds on your side?

jlaur:
> could you have a quick look wether the current Git HEAD or the version with the space builds on your side?

Current head doesn't build for me:
```
[ERROR] Failed to execute goal com.diffplug.spotless:spotless-maven-plugin:2.28.0:check (codestyle_check) on project org.openhab.binding.freeboxos: The following files had format violations:
[ERROR]     src\main\java\org\openhab\binding\freeboxos\internal\api\rest\HomeManager.java
[ERROR]         @@ -60,7 +60,7 @@
[ERROR]          ········UNKNOWN;
[ERROR]          ····}
[ERROR] 
[ERROR]         -····private·static·record·EndpointValue<T>(T·value)·{
[ERROR]         +····private·static·record·EndpointValue<T>·(T·value)·{
[ERROR]          ····}
[ERROR] 
[ERROR]          ····private·static·record·EndpointUi(AccessType·access,·DisplayType·display,·String·iconUrl,·@Nullable·String·unit)·{
[ERROR] Run 'mvn spotless:apply' to fix these violations.
```

Spotless changes to:
```java
    private static record EndpointValue<T> (T value) {
    }
```

and then it builds.

Unrelated, but now seeing the build - @clinique, there are a few compiler warnings:
```
[WARNING] openhab-addons\bundles\org.openhab.binding.freeboxos\src\main\java\org\openhab\binding\freeboxos\internal\api\rest\FreeboxOsSession.java:[151,20] Null comparison always yields false: The variable result cannot be null at this location
[WARNING] openhab-addons\bundles\org.openhab.binding.freeboxos\src\main\java\org\openhab\binding\freeboxos\internal\api\ApiHandler.java:[124,44] Potential null pointer access: The variable error may be null at this location
[WARNING] openhab-addons\bundles\org.openhab.binding.freeboxos\src\main\java\org\openhab\binding\freeboxos\internal\api\rest\APManager.java:[124,17] Redundant null check: The variable stations cannot be null at this 
location
```

jlaur:
> ```
> [ERROR] Failed to execute goal com.diffplug.spotless:spotless-maven-plugin:2.28.0:check (codestyle_check) on project org.openhab.binding.freeboxos: The following files had format violations:
> [ERROR]     src\main\java\org\openhab\binding\freeboxos\internal\api\rest\HomeManager.java
> [ERROR]         @@ -60,7 +60,7 @@
> [ERROR]          ········UNKNOWN;
> [ERROR]          ····}
> [ERROR] 
> [ERROR]         -····private·static·record·EndpointValue<T>(T·value)·{
> [ERROR]         +····private·static·record·EndpointValue<T>·(T·value)·{
> [ERROR]          ····}
> [ERROR] 
> [ERROR]          ····private·static·record·EndpointUi(AccessType·access,·DisplayType·display,·String·iconUrl,·@Nullable·String·unit)·{
> [ERROR] Run 'mvn spotless:apply' to fix these violations.
> ```

@clinique - when did you last rebase towards main? The difference in local vs. CI builds could perhaps be explained by upstream changes?

clinique:
@fwolter : rebased and it builds.

fwolter:
I can confirm that.

fwolter:
Just to avoid confusion. There is one open point left.

> Under which circumstances is the returned status PENDING?

clinique:
@fwolter : let me know your thoughts on this new implementation proposal.

bundles/org.openhab.binding.freeboxos/src/main/java/org/openhab/binding/freeboxos/internal/api/rest/LoginManager.java
```diff
@@ -138,31 +150,67 @@ public void closeSession() throws FreeboxException {
...
-                    throw new FreeboxException("Unable to grant session, access was denied");
-                default:
-                    throw new FreeboxException("Unable to grant session");
+
+        grantJob = Optional
+                .of(scheduler.schedule(new GrantLoop(timeLimit, authorize, track), 2000, TimeUnit.MILLISECONDS));
+
+        while (track == Status.PENDING) {
+            try {
+                grantJob.wait();
```
fwolter, line 192: Are you sure this is working? You're calling `wait()` on the Optional. However, `wait()` shouldn't be used directly, and if so, it must be called in a loop. https://wiki.sei.cmu.edu/confluence/display/java/THI03-J.+Always+invoke+wait%28%29+and+await%28%29+methods+inside+a+loop You could call `get()` on the `ScheduledFuture`.

fwolter (changes requested):
The current implementation spawns a new thread pool, which should only be done in specific cases like creating a server thread listening on a socket or reading data from a serial interface.

We had a very similar case here https://github.com/openhab/openhab-addons/pull/13570#discussion_r1240911512. I'm sure we can work out a viable solution, but I need a bit more context about the API returning PENDING. Is this returned on a regular basis or only in error cases? Can you somehow preemptively prevent the situation of receiving a PENDING?

clinique:
> The current implementation spawns a new thread pool, which should only be done in specific cases like creating a server thread listening on a socket or reading data from a serial interface.
> 
> We had a very similar case here [#13570 (comment)](https://github.com/openhab/openhab-addons/pull/13570#discussion_r1240911512). I'm sure we can work out a viable solution, but I need a bit more context about the API returning PENDING. Is this returned on a regular basis or only in error cases? Can you somehow preemptively prevent the situation of receiving a PENDING?

Here's the current process : 
1- The binding tries to authenticated with its current token (that may be empty on first run)
2- If no token has been stored previously, the authentication will fail (this is expected) and then the binding will send a pairing request
3- During the pairing request, the user has to move to its Free router to approve the pairing request. During this time lapse, the API will answer PENDING, until either TIMEOUT, or GRANTED if the user pressed "OK" on its device.
4- If GRANTED is received, the token is validated and given back to the handler so it is stored in the handler configuration. On next run of (1) it will success, so we'll never enter again in this loop.

jlaur:
@clinique - I don't know if relevant for this PR, but @lolodomo pushed some changes in #15121.

fwolter:
What about moving the pairing code into `initialize()` of the Thing handler (asynchronously)? The Thing status could be ONLINE with the Thing status detail CONFIGURATION_PENDING as long as the API returns PENDING. The Thing status detail message could point the user to press the pairing button.

Thing commands will fail as long as the Thing status detail is CONFIGURATION_PENDING, but this should be OK.

During CONFIGURATION_PENDING, you could poll the API every 5s or so by scheduling another task via the framework's scheduler and finally set the Thing ONLINE without a Thing status detail or to OFFLINE when the API returned the TIMEOUT state. In the latter case you could set a Thing status detail message pointing the user to re-enable the Thing to start the pairing process again.

Then, you don't need to block Thing commands when the granting process is ongoing.

clinique:
> What about moving the pairing code into `initialize()` of the Thing handler (asynchronously)? The Thing status could be ONLINE with the Thing status detail CONFIGURATION_PENDING as long as the API returns PENDING. The Thing status detail message could point the user to press the pairing button.
> 
> Thing commands will fail as long as the Thing status detail is CONFIGURATION_PENDING, but this should be OK.
> 
> During CONFIGURATION_PENDING, you could poll the API every 5s or so by scheduling another task via the framework's scheduler and finally set the Thing ONLINE without a Thing status detail or to OFFLINE when the API returned the TIMEOUT state. In the latter case you could set a Thing status detail message pointing the user to re-enable the Thing to start the pairing process again.
> 
> Then, you don't need to block Thing commands when the granting process is ongoing.

Thanks for your ideas. I'll have a look at this no later than tomorrow.

fwolter:
Did you finish making all changes?

clinique:
> Did you finish making all changes?

Yes. Tested and working.

fwolter (approved):
LGTM

Please make sure your sign-off contains your realname for future PRs.

fwolter:
Are you going to remove the legacy freebox binding from this repo and submit it to the marketplace?

lolodomo:
> Are you going to remove the legacy freebox binding from this repo and submit it to the marketplace?

No ;)

I will first intensively test the new binding and check that no feature from the original binding is missing.
You can plan to remove it in a next version.

jlaur:
> > Are you going to remove the legacy freebox binding from this repo and submit it to the marketplace?
> 
> No ;)
> 
> I will first intensively test the new binding and check that no feature from the original binding is missing. You can plan to remove it in a next version.

Please see https://github.com/openhab/openhab-addons/pull/12342#issuecomment-1193136175. I think the agreement was to upload the legacy binding the Community Marketplace and remove it from the distribution, so that 4.0 will only contain the new `freeboxos` version. Users not ready to migrate can then uninstall the bundled version and install the community version, and everything will still work.

lolodomo:
I did not remember about that !
So please don't remove it from the distribution until it is available in the marketplace and I have verified that it is working.

jlaur:
> I did not remember about that !
> So please don't remove it from the distribution until it is available in the marketplace and I have verified that it is working.

@lolodomo, @clinique - what is the status here? We are about to release 4.0 with two overlapping bindings, which was not the intention. Cc @kaikreuzer

clinique:
@jlaur : there's not much I can do here - pending @lolodomo on this.

PR #12345, opened 2022-02-22
[plugwiseha] Improve cache and timeout handling

lsiepel:
This PR fixes two issues with the current way the binding interacts with the Plugwise API (Adam).

1. Previously it fired Location_count + Appliance_count +1 times a http request to the api every 5 seconds. With this change it is 1 request per 5 seconds.

2. Somehow the Adam is a bit slow every 15 minutes and requests take longer then 5 seconds. I also added a better timeouthandling with 3 retries.

No documentation, configuration or other changes needed for the end-user.


lsiepel:
One second i thought this was some kind of demo github page, but for real i have PR #12345 :-)

bundles/org.openhab.binding.plugwiseha/src/main/java/org/openhab/binding/plugwiseha/internal/api/model/PlugwiseHAControllerRequest.java
```diff
@@ -224,18 +219,24 @@ private String getContent() throws PlugwiseHAException {
...
-        if (logger.isTraceEnabled()) {
-            logger.trace(">> {} {}", request.getMethod(), request.getURI());
-        }
+        this.logger.debug("Performing API request: {} {}", request.getMethod(), request.getURI());
 
         try {
             response = request.send();
-        } catch (TimeoutException | InterruptedException e) {
+        } catch (InterruptedException e) {
+            this.logger.trace("InterruptedException occured {} {}", e.getMessage(), e.getStackTrace());
```
lolodomo, line 231: Please add this line just after logging:

lolodomo (changes requested):

bundles/org.openhab.binding.plugwiseha/src/main/java/org/openhab/binding/plugwiseha/internal/api/model/PlugwiseHAControllerRequest.java
lsiepel: Added it

lolodomo (approved):
LGTM

PR #12348, opened 2022-02-22
[zoneminder] Properly handle when zm installed on root directory

mhilbush:
Binding wasn't handling the case where ZoneMinder is installed on the web server root directory.


lolodomo (approved):
LGTM

PR #12350, opened 2022-02-22
[http] Allow pre-escaped URLs

jamesmelville:
Whilst it's helpful that the binding automatically escapes characters in URLs for us, there are some scenarios where it doesn't escape characters which need to be escaped for a specific API call. Because the binding escapes the URL for us, any attempts to include escaped characters ourselves result in the % sign also being included.

An example is an API I wanted to use which expects a query in this format:
https://example.com/api/query?q=deleted%3Dfalse%26creatorId%3Djames&sort=updatedTime+desc&count=1

`%3D` is `=`, and `%25` is `&`, these are both valid characters in a URL (see other examples in the URL), so reverting them to their unescaped values results in the API request being rejected.

Using the URL with the HTTP binding results in an error because of the use of java's string.format to include the date in the URL:
`Creating request for 'https://example.com/api/query?q=deleted%3Dfalse%26creatorId%3Djames&sort=updatedTime+desc&count=1' failed: Conversion = 'D'`

Including a double percentage sign results in the wrong URL being called and the API returning 400, as the remaining % is also encoded:
`Requesting 'https://example.com/api/query?q=deleted%253Dfalse%2526creatorId%253Djames&sort=updatedTime+desc&count=1' (method='GET', content='null') failed: 400 null`

I'm not the only one who needs a similar capability: https://community.openhab.org/t/oh3-http-binding-in-url/133503

I wanted to propose this PR to allow those who need to to escape the URL themselves. It's a fairly simple change to add an extra parameter on the channel to flag that the URL formed from the baseURL and it's extensions is already escaped.


jamesmelville:
Zipped Jar file of built code [org.openhab.binding.http-3.3.0-SNAPSHOT.zip](https://github.com/openhab/openhab-addons/files/8120144/org.openhab.binding.http-3.3.0-SNAPSHOT.zip)

bundles/org.openhab.binding.http/README.md
```diff
@@ -35,23 +35,26 @@ Authentication might fail if redirections are involved as headers are stripper p
 
 *Note:* If you rate-limit requests by using the `delay` parameter you have to make sure that the time between two refreshes is larger than the time needed for one refresh cycle.
 
-**Attention:** `baseUrl` (and `stateExtension`/`commandExtension`) should not use escaping (e.g. `%22` instead of `"` or `%2c` instead of `,`).
+**Attention:** `baseUrl` (and `stateExtension`/`commandExtension`) should not normally use escaping (e.g. `%22` instead of `"` or `%2c` instead of `,`).
 URLs are properly escaped by the binding itself before the request is sent.
 Using escaped strings in URL parameters may lead to problems with the formatting (see below).
 
+In certain scenarios you may need to manually escape your URL, for example if you need to include an escaped `=` (`%3D`) in this scenario include `%%3D` in the URL to preserve the `%` during formatting, and set the parameter `escapedUrl` to true on the channel.
```
lolodomo, line 42: r example if you need to include an escaped `=` (`%3D`)**.
I**n this scenario**,** include

bundles/org.openhab.binding.http/README.md
```diff
@@ -35,23 +35,26 @@ Authentication might fail if redirections are involved as headers are stripper p
...
-The `image` channel-type supports `stateExtension` only.
+The `image` channel-type supports `stateExtension`, `stateContent` and `escapedUrl` only.
 
 | parameter               | optional | default     | description |
 |-------------------------|----------|-------------|-------------|
 | `stateExtension`        | yes      |      -      | Appended to the `baseURL` for requesting states. |
 | `commandExtension`      | yes      |      -      | Appended to the `baseURL` for sending commands. If empty, same as `stateExtension`. |
 | `stateTransformation  ` | yes      |      -      | One or more transformation applied to received values before updating channel. |
 | `commandTransformation` | yes      |      -      | One or more transformation applied to channel value before sending to a remote. |
+| `escapedUrl`            | yes      |      -      | This specifies whether the URL is already escaped. |
```
lolodomo, line 57: You set a default, which is false

bundles/org.openhab.binding.http/README.md
```diff
@@ -23,7 +23,7 @@ It can be extended with different channels.
 | `commandMethod`   | no       |   GET   | Method used for sending commands: `GET`, `PUT`, `POST`. |
 | `contentType`     | yes      |    -    | MIME content-type of the command requests. Only used for  `PUT` and `POST`. |
 | `encoding`        | yes      |    -    | Encoding to be used if no encoding is found in responses (advanced parameter). |
-| `headers`         | yes      |    -    | Additional headers that are sent along with the request. Format is "header=value". Multiple values can be stored as `headers="key1=value1", "key2=value2", "key3=value3",`|
+| `headers`         | yes      |    -    | Additional headers that are sent along with the request. Format is "header=value". Multiple values can be stored as `headers="key1=value1", "key2=value2", "key3=value3",`. When using text based configuration include at minimum 2 headers to avoid parsing errors.|
```
lolodomo, line 26: This is something fully unrelated to your enhancedment. I would suggest you exclude from this PR and you either propose a simple update of the doc for this or even better you could open an issue.

lolodomo (changes requested):
Only few remarks about updated documentation.

lolodomo:
@jamesmelville : did you see my review comments?

lolodomo:
As there is no feedback but my comments were relatively minor and only about documentation, I propose to merge.

PR #12353, opened 2022-02-23
Fix Java 17 compilation and test issues

wborn:
* Adds --add-opens to the maven-surefire-plugin config required for deserialization using Gson/XStream
* Upgrades plugin dependencies to JDK 17 compatible versions
* Replaces some reflection that no longer works on JDK 17
* Fixes issues when mocking Random
* Run Nashorn dependant tests only on JDK < 15


fwolter (approved):
LGTM

Setting a private static final field via reflection :)

wborn:
The mielecloud tests would be more maintainable/future proof if they didn't use any reflection trickery at all @BjoernLange. :slightly_smiling_face:

PR #12355, opened 2022-02-23
[voskstt] add linux arm and linux aarch64 binaries 

GiviMAD:
I realize the binaries for this architectures were not included when I tryed to run it on raspberry OS.
They are available on the vosk-api github release page so I've added them, now it works perfect on my raspberry pi 4.
Also changed the default value for model preload to false, which seems more logical.


GiviMAD:
> I am a little lost. This libraries are loaded dynamically by the code ? As there is no provided library for win64 for example, does it mean the addon is not working on Windows 64 ?

Sorry I will explain it correctly, the sdk already includes de binaries for linux-x86_64, win64, and darwin. I assumed all the supported architectures were already included with the sdk, but no.

I will remove Win32 from the list. It's available also on the release page but I didn't include it.

lolodomo:
> Sorry I will explain it correctly, the sdk already includes de binaries for linux-x86_64, win64, and darwin. I assumed all the supported architectures were already included with the sdk, but no.

But where is the code that loads the right library depending on the running architecture ?

GiviMAD:
> But where is the code that loads the right library depending on the running architecture ?
That's handled by the vosk sdk. It's using [this library](https://github.com/java-native-access/jna) for that.

PR #12356, opened 2022-02-23
[velux] Improve shutdown exception handling

### LINKED ISSUE #12327: [velux] catching InterruptedException or not?
This refers to issue #10454 whereby I wonder if the Velux binding also needs to handle InterruptedException differently?

The binding does catch InterruptedException in at least seven locations; it logs such exceptions to debug; but it does NOT call `Thread.currentThread().interrupt()` -- which now seems to be the recommended approach..

The KLF bridge is notorious for entering a so called 'zombie' state where the binding thinks the connection socket has been destroyed, whereas the bridge thinks the socket is still open, and so it refuses to accept a new socket connection attempt from the binding. This error most often occurs when shutting down OpenHab; however it does not occur every time, and the circumstances are hard to determine, so it is has proven so far to be impossible to debug. Therefore it is not obvious if catching InterruptedException and calling `Thread.currentThread().interrupt()` might actually solve this issue, or not. ??

This issue is raised as a placeholder for a discussion, before (eventually, or not) opening a PR.

andrewfg:
Potentially fixes #12327

This PR addresses the issues discussed in #10454 and #12327 in a further attempt to solve the KLF200 'zombie state' problem.

Two things have been improved..
- Added an isDisposing flag so that the binding can shut down faster if it was already in a zombie state.
- Refactored the code that reads messages from the KLF200 to handle `InterruptedException` , `ExecutionException`, and `IOException` more properly and completely.

### DIFF
--- a/bundles/org.openhab.binding.velux/src/main/java/org/openhab/binding/velux/internal/bridge/slip/io/Connection.java
+++ b/bundles/org.openhab.binding.velux/src/main/java/org/openhab/binding/velux/internal/bridge/slip/io/Connection.java
@@ -147,14 +147,18 @@ public class Connection implements Closeable {
             } catch (IOException ioe) {
                 logger.info("io() on {}: Exception occurred during I/O: {}.", host, ioe.getMessage());
                 lastIOE = ioe;
-                // Error Retries with Exponential Backoff
-                long waitTime = ((long) Math.pow(2, retryCount)
-                        * bridgeInstance.veluxBridgeConfiguration().timeoutMsecs);
-                logger.trace("io() on {}: wait time {} msecs.", host, waitTime);
-                try {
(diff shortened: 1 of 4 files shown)


andrewfg:
Ping @gs4711 for comments if any..

gs4711:
@andrewfg, being busy within the last days... therefore, now an answer after a delay of some days.

The changes look good to me. Just out of curiosity: how did you came to the value of MAX_WAIT_SECONDS ?

andrewfg:
> how did you came to the value of MAX_WAIT_SECONDS

The OH maintainers request that a binding shutdown should try not to take longer than 15 seconds.

fwolter (approved):

andrewfg:
@fwolter many thanks.

PR #12360, opened 2022-02-23
Add Java 17 to GHA CI build matrix

wborn:
This adds Java 17 to the GitHub Actions CI build matrix so we can make sure the build keeps working with both Java 11 and Java 17. It also updates the required Java version range used by the enforcer plugin so it is also possible to build with the supported Java versions.

### DIFF
--- a/.github/workflows/ci-build.yml
+++ b/.github/workflows/ci-build.yml
@@ -17,7 +17,7 @@ jobs:
   build:
     strategy:
       matrix:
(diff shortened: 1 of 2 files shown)


jlaur (approved):
Good idea, thanks!

wborn:
The full builds are also 11 minutes faster with Java 17 which is quite significant. :slightly_smiling_face:

PR #12363, opened 2022-02-23
new goechargerapiv2 binding

ranzen84:
# Title

new goechargerapiv2 binding

# Description

This Binding reads and writes data from the [Go-eCharger](https://go-e.co/) HOME+ and HOMEfix with API version 2 (Hardware version 3).
The HOME+ is a mobile wallbox for charging EVs, HOMEfix is a mounted wallbox and they have an open REST API for reading and writing data and configuration.

Signed-off-by: Reinhard Plaim <reinhardplaim@gmail.com>

### DIFF
--- a/bundles/org.openhab.binding.goechargerapiv2/NOTICE
+++ b/bundles/org.openhab.binding.goechargerapiv2/NOTICE
@@ -1,0 +1,13 @@
+This content is produced and maintained by the openHAB project.
+
+* Project home: https://www.openhab.org
+
+== Declared Project Licenses
+
+This program and the accompanying materials are made available under the terms
+of the Eclipse Public License 2.0 which is available at
+https://www.eclipse.org/legal/epl-2.0/.
+
+== Source Code
+
+https://github.com/openhab/openhab-addons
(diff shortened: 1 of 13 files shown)


jlaur:
@ranzen84 - a few initial questions: Why did you create a new binding when there's already an existing binding? Wouldn't it be possible to extend the existing binding to support the new API v2? Will a new binding be needed for each new API/hardware revision? I don't think "api" should be part of a binding name since it denotes a technical detail not relevant to users. Probably not "v2" either as it simply denotes a specific revision of the API, which again should probably not be of much concern to users.

lsiepel (commented):
Did my first review, but i'm still a freshmen here :-)
I think you should also fix the PR title to: [goeChargerapiv2] Initial binding.  And also consider to remove the apiv2 part of the name. If you can look into the failed build and signoff someone more rights should review.

I'm also wondering if this is tested, is there a community thread? Did you have discussion with the other go-charger binding developer Samuel Brucksch?  Why not integrate both?

ranzen84:
Thank you for your reply. I think you are right, I will get in contact with Samuel Brucksch and will try to integrate the adaptions for API v2 to the existing binding. There are some differences/breaking changes but it should be able to handle this.

PR #12349, opened 2022-02-22
[plugwiseha] Add and enhance channels

lsiepel:
1. Add Zone => cooling_allowed channel (r/w)
2. Add Zone => regulation_control channel (r/w)
3. Fix documentation mistake=> preheating_allowed channel is r/w
4. Changed Zone => scene channel from read-only to r/w.
5. Minor codestyle improvements

No configuration or other changes needed for the end-user.

PR #12366, opened 2022-02-24
[plugwiseha] Fix bug introduced with #12349

lsiepel:
Sorry @lolodomo i should have paid more attention to your comment about the StringType returning null in my previous PR. I thought it was fine. When testing i had no problems, but after some user feedback we discovered some situations that it does lead to a problem.


bundles/org.openhab.binding.plugwiseha/src/main/java/org/openhab/binding/plugwiseha/internal/handler/PlugwiseHABridgeHandler.java
```diff
@@ -175,12 +175,16 @@ private void run() {
         } catch (PlugwiseHAUnauthorizedException | PlugwiseHANotAuthorizedException e) {
             updateStatus(OFFLINE, CONFIGURATION_ERROR, STATUS_DESCRIPTION_INVALID_CREDENTIALS);
         } catch (PlugwiseHACommunicationException e) {
+            this.logger.trace("Bridge encountered an error {} {}", e.getMessage(), e.getStackTrace());
```
lolodomo, line 178: To log stack trace:
`this.logger.trace("Bridge encountered an error {}", e.getMessage(), e);`
Same for others.

lolodomo (changes requested):

bundles/org.openhab.binding.plugwiseha/src/main/java/org/openhab/binding/plugwiseha/internal/handler/PlugwiseHABridgeHandler.java
lsiepel: Fixed

lolodomo (approved):
LGTM

PR #1375, opened 2016-11-01
Hotfix

Mr-Eskildsen:

PR #12367, opened 2022-02-24
[openthermgateway] Add support for Ventilation/Heat Recovery units 

### LINKED ISSUE #11202: [openthermgateway] Add support for Ventilation/Heat Recovery and Solar Boiler units

andrewfg:
Continues from  #11203 (i.e. closes #11203)
Resolves #11202 

As discussed with @fwolter and @ArjenKorevaar


bundles/org.openhab.binding.openthermgateway/src/main/java/org/openhab/binding/openthermgateway/handler/OpenThermGatewayHandler.java
```diff
@@ -108,184 +102,186 @@ public void handleCommand(ChannelUID channelUID, Command command) {
...
+            logger.debug("Starting OpenThermGateway connector");
+
+            autoReconnect = true;
 
-            connector = new OpenThermGatewaySocketConnector(this, conf.ipaddress, conf.port);
+            connector = new OpenThermGatewaySocketConnector(this, config);
 
             Thread thread = new Thread(connector, "OpenTherm Gateway Binding - socket listener thread");
             thread.setDaemon(true);
             thread.start();
```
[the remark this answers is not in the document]
lolodomo: Scheduler should only handle short time thread. If your thread is a long time thread, it must not be started by scheduler (as it will lock an entry in the thread pool.
Also note that the thread pool name does not respect our convention.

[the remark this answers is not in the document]
fwolter: You could do `new NamedThreadFactory("binding-xyz")` to create a custom thread. This way, the thread will be automatically named correctly.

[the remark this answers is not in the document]
andrewfg: Done

andrewfg:
@fwolter apropos your https://github.com/openhab/openhab-addons/pull/11203#discussion_r815459410 .. it is nevertheless still a little bit different in the case of this binding.. 
- In `hdpowerview` the Thing state is set in two steps 1) to 'UNKNOWN' when `initialize()` is called, and 2) to 'ONLINE' when it receives the first data from the Bridge.
- In `openthermgateway` the Thing state can be set directly in one step when `initialize()` is called and the Bridge 'ONLINE' then the state is also 'ONLINE' (reason is that the Things are more tightly coupled with the Bridge), and it is wrong to set the Thing to 'UNKNOWN' can only be 'ONLINE' or 'BRIDGE:ERROR'..

fwolter:
OK, but the framework should manage to set the child Things' states correctly, also in this case. The LCN binding sets the child Things to ONLINE in `initialize()` directly and the automatism works correctly. See https://github.com/openhab/openhab-addons/blob/923f720a5c256613fbefb4d4bae7634732a14d87/bundles/org.openhab.binding.lcn/src/main/java/org/openhab/binding/lcn/internal/LcnModuleHandler.java#L143

andrewfg:
> the framework should manage to set the child Things' states correctly, also in this case

Ok, I will try this tomorrow.

andrewfg:
Note: I am also going to implement some lessons about socket exception handling that I learned while writing the code for #12356

andrewfg:
@fwolter apropos https://github.com/openhab/openhab-addons/pull/12367#issuecomment-1053650055 I have made the requested changes, and confirm that it works :)

andrewfg:
@fwolter as this PR is a _'breaking change'_ I would like to add whatever I can to ameliorate the migration for users. To be specific the change is as follows..
- **Prior Version**: there is just one single Thing which combines the functionality of the OpenThermGateway communication tasks (functions you would expect of a Bridge), and the Boiler (functions you would expect of a Thing)
- **New Version**: the communication functions have been moved to the OpenThermGateway Bridge, the Boiler functions have been moved to the Boiler Thing, and the Ventilation/Heat-pump have been added in a new Ventilation/Heat-pump Thing.

So the Thing and Item definitions have to be migrated as shown below.

@fwolter I am wondering if there are "automations" in the framework that would simplify such migration? e.g. Dynamic Thing/Channel/Item creation? Unfortunately I don't really know enough about this topic myself.

**Old Thing and Item Definition**
```
Thing openthermgateway:otgw:XXX [ ipaddress="192.168.1.100", port=8000, connectionRetryInterval=60 ]
----
Number:Temperature Boiler_DHW_Temperature "Boiler DHW Temperature [%.1f %unit%]" <temperature> {"openthermgateway:otgw:XXX:dhwtemp"}
```
**New Thing and Item Definition**
```
Bridge openthermgateway:openthermgateway:XXX "OpenTherm Gateway" @ "Kitchen" [ipaddress="192.168.1.100", port=20108, connectionRetryInterval=60] {
    Thing boiler viessmann "Boiler" @ "Kitchen"
}
-----
Number:Temperature Boiler_DHW_Temperature "Boiler DHW Temperature [%.1f %unit%]" <temperature> {channel="openthermgateway:boiler:XXX:viessmann:dhwtemp"}
```

andrewfg:
IMHO there are still bugs in this code => converting back to DRAFT

andrewfg:
@fwolter do you have any idea why I get the following compiler warning ?? (Google did not help me) I think it is not a big deal because it builds and runs anyway..

```
[WARNING] G:\git\andrewfg\openhab-addons\bundles\org.openhab.binding.openthermgateway\pom.xml [0:0]: Export org.openhab.binding.openthermgateway.handler,  has 1,  private references [org.openhab.binding.openthermgateway.internal]
```

andrewfg:
Yesterday I made some further commits as follows..
1. remove all (but the last one) compiler warnings
2. fix code style warnings
3. removed duplicate OpenthermgatewayBindingConstants class (removed 'legacy' version)
4. removed xml thing descriptions for 'legacy' `otgw` thing
5. added a chapter in ReadMe to explain how users should migrate from legacy `otgw` thing to new `openthermgateway` bridge resp. `boiler` thing

=> So I am again now ready to re-request your @fwolter review. :) i.e. changing from DRAFT to active PR

@fwolter in order to make your re-review a bit easier, I summarise the take-overs from your prior review of #11203 as follows..

A) @fwolter #11203 review change requests that **WERE** implemented
- Checkstyle code analysis: => all issues resolved
- TSP / FHB channels: => closed as agreed
- ReadMe chapter on TSP / FHB: => chapter moved as requested
- Thing status updated by the framework: => resolved as discussed in this and other bindings
- verifyTspFhbChannels method name: => fixed as requested
- logger.info() message: => logger.info() moved to logger.debug()
- Number:Dimensionless: reverted to Number:time as requested (and I agree with you)

B) @fwolter #11203 review change requests that were **NOT** implemented
- Channel name 'vh_' prefixes: => not changed because, as @ArjenKorevaar says (and I agree), the 'Channel' names should match the 'Data-Object` names already previously published in the OpenTherm Specification documents.
- Channel labels: NOT changed because (as above) names / descriptions should match the 'Data-Object` names / descriptions already previously published in the OpenTherm Specification documents.
- logger.debug(): => NOT moved to debugger because, as @ArjenKorevaar says (and I agree), the debug messages are sometimes needed to help users with problems on the public fora
- Ventilation Enabled vs. ON / OFF: NOT changed because ENABLED is indeed a different state than ON / OFF
- Converting Channels to Properties: NOT implemented, because you said you won't insist, and because it would require a major re-work of the binding and risks breaking something else (I checked this)

bundles/org.openhab.binding.openthermgateway/src/main/java/org/openhab/binding/openthermgateway/handler/BaseDeviceHandler.java
```diff
@@ -54,15 +53,8 @@ public BaseDeviceHandler(Thing thing) {
 
     @Override
     public void initialize() {
-        Bridge bridge = getBridge();
-
-        if (bridge == null) {
-            updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.BRIDGE_UNINITIALIZED);
```
jlaur: I think this is still needed in case thing is configured without bridge, but in this case it should be offline with **CONFIGURATION_ERROR**. Please see #12331.

andrewfg: I thought that @fwolter was saying that the framework takes care not to initialize the things if a bridge is a) offline, b) wrongly configured, or c) missing. Or??

jlaur: I don't remember any such statement about c) missing? Could be tested by simply creating a things file with a thing not assigned to any bridge. In the linked PR he did not remove the check for bridge being null.

bundles/org.openhab.binding.openthermgateway/src/main/resources/OH-INF/thing/openthermgateway.xml
```diff
@@ -0,0 +1,18 @@
...
+	xmlns:thing="https://openhab.org/schemas/thing-description/v1.0.0"
+	xsi:schemaLocation="https://openhab.org/schemas/thing-description/v1.0.0 https://openhab.org/schemas/thing-description-1.0.0.xsd">
+	<bridge-type id="openthermgateway">
+		<label>OpenTherm Gateway</label>
+		<description>OpenTherm Gateway</description>
+		<channels>
+			<channel id="sendcommand" typeId="sendcommand"/>
+		</channels>
+		<properties>
+			<property name="version">2.1.4</property>
```
ArjenKorevaar, line 13: Could you update the version number of the binding ? I use this when providing support as bindings don't otherwise seem to have their own version number. I moved to major 2 with the introduction of bridge and separate things, so the next version number would be 2.2.0

ArjenKorevaar: Otherwise LGTM, thanks for the effort you've put into completing this PR

andrewfg: Ok.

bundles/org.openhab.binding.openthermgateway/src/main/java/org/openhab/binding/openthermgateway/handler/BaseDeviceHandler.java
andrewfg: Ok.

ArjenKorevaar (approved):

bundles/org.openhab.binding.openthermgateway/src/main/java/org/openhab/binding/openthermgateway/handler/OpenThermGatewayHandler.java
```diff
@@ -108,183 +101,184 @@ public void handleCommand(ChannelUID channelUID, Command command) {
...
-                    logger.debug("Received update for channel '{}': {}", channelId, state);
-                    updateState(channelId, state);
-                }
+    private void connectionStateChangedTask(ConnectionState state) {
+        if (this.state != state) {
+            this.state = state;
+
+            switch (state) {
+                case INITIALIZING:
+                    updateStatus(ThingStatus.UNKNOWN, ThingStatusDetail.NONE, "Initializing");
```
jlaur, line 168: Please provide i18n text for this as well as next two strings below.

bundles/org.openhab.binding.openthermgateway/src/main/java/org/openhab/binding/openthermgateway/handler/OpenThermGatewayHandler.java
```diff
@@ -108,183 +101,184 @@ public void handleCommand(ChannelUID channelUID, Command command) {
...
+                    break;
+                case DISCONNECTED:
+                    updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.NONE, "Disconnected");
+                    autoReconnect();
+                    break;
+                case CONNECTING:
+                    updateStatus(ThingStatus.UNKNOWN, ThingStatusDetail.NONE, "Connecting");
```
jlaur, line 175: What could be the reason for this ConnectionState? Initially **UNKNOWN** is correct, but if this can happen later because device is offline and we now need to reconnect, **OFFLINE** would be more appropriate. Didn't check how it's used.

jlaur (changes requested):
Looks good, just a few comments. Also: Can you regenerate i18n texts?

bundles/org.openhab.binding.openthermgateway/src/main/java/org/openhab/binding/openthermgateway/handler/OpenThermGatewayHandler.java
andrewfg: > Can you regenerate i18n texts?

Sorry but what do you mean by this?

bundles/org.openhab.binding.openthermgateway/src/main/java/org/openhab/binding/openthermgateway/handler/OpenThermGatewayHandler.java
andrewfg: This is a callback which is called several times during the various stages of establishing a TCP socket connection to the gateway device. 

I suppose one could argue that if the Bridge is not specifically ONLINE then it is _per definition_ OFFLINE. In that case the switch `case CONNECTED` could return ONLINE and the switch `default` case should return OFFLINE. So in other words the switch statement could be simplified to the following. PS also removes the need for special i18n texts :)

```
            switch (state) {
                case CONNECTED:
                    cancelAutoReconnect();
                    break;
                case DISCONNECTED:
                    autoReconnect();
                    // fall through
                default:
            }
```

andrewfg: ^
Probably the following is even better than the above..
```
            switch (state) {
                case CONNECTED:
                    cancelAutoReconnect();
                    break;
                case DISCONNECTED:
                    updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.COMMUNICATION_ERROR);
                    autoReconnect();
                    break;
                case CONNECTING:
                case INITIALIZING:
                    updateStatus(ThingStatus.INITIALIZING);
            }
```

andrewfg: Actually `case DISCONNECTED` can occur both a) if there was a real communication error and b) if the Bridge was disabled by the framework; so strictly in the latter case, `ThingStatusDetail.COMMUNICATION_ERROR` is not correct a correct statement; but I think it doesn't really matter because the framework will set 'ThingStatusDetail.DISABLED' immediately afterwards. Or ??

bundles/org.openhab.binding.openthermgateway/src/main/java/org/openhab/binding/openthermgateway/handler/OpenThermGatewayHandler.java
jlaur: Simply to move the texts to the i18n file so they could be translated. But from comment below seems like it will not be needed. :)

bundles/org.openhab.binding.openthermgateway/src/main/java/org/openhab/binding/openthermgateway/handler/OpenThermGatewayHandler.java
```diff
@@ -108,183 +101,181 @@ public void handleCommand(ChannelUID channelUID, Command command) {
...
+                    updateStatus(ThingStatus.ONLINE);
+                    cancelAutoReconnect();
+                    break;
+                case DISCONNECTED:
+                    updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.COMMUNICATION_ERROR);
+                    autoReconnect();
+                    break;
+                case CONNECTING:
+                case INITIALIZING:
+                    updateStatus(ThingStatus.INITIALIZING);
```
jlaur, line 177: Now this seems wrong - please see https://www.openhab.org/docs/concepts/things.html#status-transitions. **INITIALIZING** is set by the framework before calling `initialize()`. After setting UNKNOWN, ONLINE or OFFLINE in `initialize()`, status **INITIALIZING** should not be explicitly set by the binding.

bundles/org.openhab.binding.openthermgateway/src/main/java/org/openhab/binding/openthermgateway/handler/OpenThermGatewayHandler.java
jlaur: Not sure, sounds plausible, but probably you would need to test it.

andrewfg: > you would need to test it
Ok, I have tested it now :) and I found out that setting `ThingStatus.INITIALIZING` is not allowed, so I have pushed a (hopefully) final commit that (hopefully) closes this topic.

bundles/org.openhab.binding.openthermgateway/src/main/java/org/openhab/binding/openthermgateway/handler/OpenThermGatewayHandler.java
andrewfg: Yup. Fixed. See above.

bundles/org.openhab.binding.openthermgateway/src/main/java/org/openhab/binding/openthermgateway/handler/OpenThermGatewayHandler.java
```diff
@@ -108,183 +103,183 @@ public void handleCommand(ChannelUID channelUID, Command command) {
...
+                    cancelAutoReconnect();
+                    break;
+                case DISCONNECTED:
+                    if (!disposing) {
+                        updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.COMMUNICATION_ERROR);
+                        autoReconnect();
+                    }
+                    break;
+                default:
+                    updateStatus(ThingStatus.UNKNOWN);
```
jlaur, line 180: Is this needed? If I understand it correctly (which I may not), this method will be called with `ConnectionState.CONNECTING` when trying to connect, which could be for transitioning from **OFFLINE** back to **ONLINE**. In this case it probably wouldn't make much sense to switch to **UNKNOWN** briefly during this transition? On the other hand, before leaving `initialize()`, **UNKNOWN** should be set.

andrewfg: > wouldn't make much sense to switch to UNKNOWN briefly 

Ok. Done.

> before leaving initialize(), UNKNOWN should be set.

Ok. Done.

jlaur (approved):
Great work, thanks! Please remember to create a PR for adding a (broken) compatibility note [here](https://github.com/openhab/openhab-distro/blob/54c62770724dadf758524647fbffb66b9b17cda8/distributions/openhab/src/main/resources/bin/update.lst#L63-L68).

andrewfg:
Ok. Many thanks. I will create the compatibilty note asap.

andrewfg:
=> compatibilty note: [#1375](https://github.com/openhab/openhab-distro/pull/1375)

PR #7625, opened 2020-05-13
Don't limit serial port options

wborn:
When serial port options are limited it is only possible to configure discovered ports.
RXTX discovery only detects standard serial ports.
The serial transport adds undiscovered ports to 'gnu.io.rxtx.SerialPorts' so this way users do not need to manually configure it.
RFC2217 ports cannot be detected so if the ports are limited it is not possible to configure these using UIs.

Related to:

* https://github.com/openhab/openhab-core/issues/1029
* https://github.com/openhab/openhab-core/issues/1462

PR #12370, opened 2022-02-24
[novafinedust] Fix RFC2217 serial port in configuration

### LINKED ISSUE #12369: [novafinedust] Broken if RFC2217 ports (serial over ethernet) is used in port config
See: https://github.com/openhab/openhab-core/issues/2798

I will provide a workaround here.

t2000:
Fixes #12369


wborn (approved):
Very good! They should also not be limited because ports with custom names created using udev rules are also not discovered, see #7625.

PR #12376, opened 2022-02-26
[pulseaudio] register audio sources in openhab

GiviMAD:

This PR allows audio sources to be used by openHAB.

@peuter as you are the code owner, maybe you want to check the changes.

@lolodomo you asked me about this a some time ago, I just checked this with the dialog processor and same problem that with windows, as the stream is shared and the ks keeps consuming the stream, execution of the stt behaves poorly, I will open an issue on the core describing the problem because I think that adding some buffering mechanism by stream can solve the problem, then I can apply the solution on both sides.


lolodomo:
> @lolodomo you asked me about this a some time ago

If this allows using remote audio source, this is just excellent.

> I just checked this with the dialog processor and same problem that with windows, as the stream is shared and the ks keeps consuming the stream, execution of the stt behaves poorly, I will open an issue on the core describing the problem because I think that adding some buffering mechanism by stream can solve the problem, then I can apply the solution on both sides.

IMHO the problem is simply that the dialog processor should close the KS source before running the STT service, and then reopen it when the STT service ends. What makes no sense for me is to have several services opening the same audio source at the same time. This is even not necessary.

GiviMAD:
I've added a late change as my way to reconnect the stream was incorrect.

@peuter do you think that the socket SoTimeout could be increased in PulseaudioSimpleProtocolStream?

GiviMAD:
Yes I can confirm that after set the soTimeout of the simple protocol socket I stop facing the unexpected closing errors.
@peuter @lolodomo what do you think on make this value configurable through the sink/source configuration?
This way it can be tuned to match the Pulseaudio server host capabilities which I think is what you need to try to match?

bundles/org.openhab.binding.pulseaudio/README.md
```diff
@@ -59,6 +59,14 @@ Sink things can register themselves as audio sink in openHAB. MP3 and WAV files
 Use the appropriate parameter in the sink thing to activate this possibility (activateSimpleProtocolSink).
 This requires the module **module-simple-protocol-tcp** to be present on the server which runs your openHAB instance. The binding will try to command (if not discovered first) the load of this module on the pulseaudio server.
 
+
+## Audio source
+
+Source things can register themselves as audio source in openHAB.
+WAV input format, rate and channels can be configured on the thing configuration. (defaults to pcm_signed,16000,1)
+Use the appropriate parameter in the source thing to activate this possibility (activateSimpleProtocolSource).
+This requires the module **module-simple-protocol-tcp** to be present on the target pulseaudio server. The binding will load this module on the pulseaudio server.
```
lolodomo, line 68: What is the "target pulseaudio server". Is it "server which runs your openHAB instance" like for the sink thing ? Or is it the remote server with the audio source ?

bundles/org.openhab.binding.pulseaudio/src/main/java/org/openhab/binding/pulseaudio/internal/PulseAudioAudioSink.java
```diff
@@ -34,102 +31,48 @@
...
 import org.openhab.core.audio.UnsupportedAudioStreamException;
-import org.openhab.core.library.types.PercentType;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 /**
  * The audio sink for openhab, implemented by a connection to a pulseaudio sink
  *
  * @author Gwendal Roulleau - Initial contribution
+ * @author Miguel Álvarez - Register audio source and refactor
```
lolodomo, line 41: Please update your comment to match your changes in this class.

bundles/org.openhab.binding.pulseaudio/src/main/java/org/openhab/binding/pulseaudio/internal/PulseaudioBindingConstants.java
```diff
@@ -20,6 +20,7 @@
  * used across the whole binding.
  *
  * @author Tobias Bräutigam - Initial contribution
+ * @author Miguel Álvarez - Register audio source and refactor
```
lolodomo, line 23: Please update your comment to match your changes in this class.

bundles/org.openhab.binding.pulseaudio/src/main/java/org/openhab/binding/pulseaudio/internal/PulseaudioClient.java
```diff
@@ -47,14 +50,16 @@
  * On the pulseaudio server the module-cli-protocol-tcp has to be loaded.
  *
  * @author Tobias Bräutigam - Initial contribution
+ * @author Miguel Álvarez - Register audio source and refactor
```
lolodomo, line 53: Please update your comment to match your changes in this class.

lolodomo (changes requested):
Review part 1 of 2

bundles/org.openhab.binding.pulseaudio/README.md
GiviMAD: It could be both, it's a server with a running Pulseaudio instance.
What could be a better expression for it?

lolodomo: Ok, I understand, it could be a remote server or the same server as the OH server. In the case of a local source, it makes this an alternative to javasound if I correctly understand.
So you have nothing to change.

bundles/org.openhab.binding.pulseaudio/src/main/java/org/openhab/binding/pulseaudio/internal/PulseAudioAudioSource.java
```diff
@@ -0,0 +1,186 @@
...
+                        break;
+                    }
+                    var sourceFormat = pulseaudioHandler.getSourceAudioFormat();
+                    if (sourceFormat == null) {
+                        throw new AudioException("Unable to get source audio format");
+                    }
+                    if (!audioFormat.isCompatible(sourceFormat)) {
+                        throw new AudioException("Incompatible audio format requested");
+                    }
+                    setIdle(true);
```
lolodomo, line 77: I have a doubt, is it true or false you want to set here ?

bundles/org.openhab.binding.pulseaudio/src/main/java/org/openhab/binding/pulseaudio/internal/PulseAudioAudioSource.java
```diff
@@ -0,0 +1,186 @@
...
+        }
+
+        @Override
+        public AudioFormat getFormat() {
+            return format;
+        }
+
+        @Override
+        public int read() throws IOException {
+            logger.trace("reading from pulseaudio stream");
```
lolodomo, line 145: You have 3 times the same log message in 3 different methods.
I would suggest to keep only one, the third one as it will be always be called finally.
Or if you want to have a log in each method, make a difference in the log message.

lolodomo (changes requested):
Review part 2 of 2

lolodomo:
And do not forget to update the default translations.

GiviMAD:
> I have an off-topic question: if I would like to use the remote audio source connected to one RPI (which is not my OH server), do I need to install pulseaudio on the server running my OH server or is it enough to install it only on the RPI having the microphone connected to ?

Just on the remote server, you need module-cli-protocol-tcp enabled there, so the binding could send remote commands to it. That can be done by adding "load-module module-cli-protocol-tcp" at the end of /etc/pulse/default.pa.
Let me know if you encounter any problem.

bundles/org.openhab.binding.pulseaudio/src/main/java/org/openhab/binding/pulseaudio/internal/PulseAudioAudioSource.java
```diff
@@ -0,0 +1,186 @@
...
+                    disconnect(); // disconnect force to clear connection in case of socket not cleanly shutdown
+                    if (countAttempt == 2) { // we won't retry : log and quit
+                        if (logger.isWarnEnabled()) {
+                            String port = clientSocket != null ? Integer.toString(clientSocket.getPort()) : "unknown";
+                            logger.warn(
+                                    "Error while trying to get audio from pulseaudio audio source. Cannot connect to {}:{}, error: {}",
+                                    pulseaudioHandler.getHost(), port, e.getMessage());
+                        }
+                        setIdle(true);
+                        throw e;
```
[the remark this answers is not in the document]
GiviMAD: it's captured in a higher catch to have this finally block, it's similar to what is done in the sink, I "copied" the retry mechanism from there
```
       } catch (IOException e) {
            throw new AudioException(e);
        } finally {
            scheduleDisconnect();
        }
```

bundles/org.openhab.binding.pulseaudio/src/main/java/org/openhab/binding/pulseaudio/internal/PulseAudioAudioSource.java
GiviMAD: True, so the underling socket will be closed if the stream remains unused, that's ok as it will reconnect on first read.

bundles/org.openhab.binding.pulseaudio/src/main/java/org/openhab/binding/pulseaudio/internal/PulseAudioAudioSource.java
```diff
@@ -0,0 +1,186 @@
...
+                    return new PulseAudioStream(sourceFormat, this::getSourceInputStream, (idle) -> {
+                        setIdle(idle);
+                        if (idle) {
+                            scheduleDisconnect();
+                        }
+                    });
+                } catch (IOException e) {
+                    disconnect(); // disconnect force to clear connection in case of socket not cleanly shutdown
+                    if (countAttempt == 2) { // we won't retry : log and quit
+                        if (logger.isWarnEnabled()) {
```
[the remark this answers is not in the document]
GiviMAD: Also copied from the sink along the retry mechanism. For me it's just for optimization purposes, you will avoid to compute the log call when warn is not enabled. But as it's almost always enabled I can remove it if you want.

bundles/org.openhab.binding.pulseaudio/src/main/java/org/openhab/binding/pulseaudio/internal/PulseAudioAudioSource.java
GiviMAD: I didn't realize that was happening, there is also an extra call to setIdle because of the same, thanks!

GiviMAD:
I will try to modify the dialog processor to stop the ks while running the stt to see if it solves the transcription problem

bundles/org.openhab.binding.pulseaudio/src/main/java/org/openhab/binding/pulseaudio/internal/PulseAudioAudioSource.java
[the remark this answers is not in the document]
lolodomo: Yes, please remove it in both cases, because I don't see any value here, the code inside the if is basic.

bundles/org.openhab.binding.pulseaudio/src/main/java/org/openhab/binding/pulseaudio/internal/PulseAudioAudioSource.java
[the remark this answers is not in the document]
lolodomo: Yes, you're right, I see it now.

bundles/org.openhab.binding.pulseaudio/src/main/java/org/openhab/binding/pulseaudio/internal/PulseAudioAudioSource.java
```diff
@@ -0,0 +1,184 @@
...
+            Byte bb = Byte.valueOf(b[0]);
+            return bb.intValue();
+        }
+
+        @SuppressWarnings("null")
+        @Override
+        public int read(byte @Nullable [] b) throws IOException {
+            logger.trace("reading from pulseaudio stream");
+            setIdle.accept(false);
+            return getInputStream().read(b, 0, b.length);
```
lolodomo, line 159: If I am not wrong, you could here just replace these 3 lines by just:
```
return read(b, 0, b.length);
```

lolodomo (changes requested):

GiviMAD:
@lolodomo I have solved the problem with multiple stream open at same time using pipe streams.

GiviMAD:
Tested with the latest changes of the pr https://github.com/openhab/openhab-core/pull/2809 where ks and stt run in parallel.
Now everything seems to work.

bundles/org.openhab.binding.pulseaudio/src/main/java/org/openhab/binding/pulseaudio/internal/PulseAudioAudioSource.java
```diff
@@ -0,0 +1,246 @@
...
+    }
+
+    private synchronized void registerPipe(PipedOutputStream pipeOutput) {
+        this.pipeOutputs.add(pipeOutput);
+        startPipeWrite();
+    }
+
+    private void startPipeWrite() {
+        if (pipeWriteTask == null) {
+            this.pipeWriteTask = scheduler.submit(() -> {
```
lolodomo, line 134: Please do not use scheduler (and the common thread pool) for long life thread.
Please check that such thing was not done in your recent addon implementation.

lolodomo (changes requested):

lolodomo:
@dalgwen: you are the last to contribute several PRs for this binding. Would you like to provide a code review of this new PR?

bundles/org.openhab.binding.pulseaudio/src/main/java/org/openhab/binding/pulseaudio/internal/PulseAudioAudioSource.java
GiviMAD: oh, no in the other plugins I didn't do this as I created a separate executor:
```
    private final ScheduledExecutorService executor = ThreadPoolManager.getScheduledPool("OH-voice-voskstt");
```
But I did it because I didn't have the common scheduler accesible there, as I didn't know this.
I will do the same here and create a new executor, thanks  for pointing it.

bundles/org.openhab.binding.pulseaudio/src/main/java/org/openhab/binding/pulseaudio/internal/PulseAudioAudioSource.java
```diff
@@ -44,20 +45,23 @@
...
     }
 
     @Override
     public Set<AudioFormat> getSupportedFormats() {
+        var supportedFormats = new HashSet<AudioFormat>();
+        var audioFormat = pulseaudioHandler.getSourceAudioFormat();
+        if (audioFormat != null) {
+            supportedFormats.add(audioFormat);
+        }
+        ;
```
lolodomo, line 64: To be removed.

lolodomo (changes requested):

bundles/org.openhab.binding.pulseaudio/src/main/java/org/openhab/binding/pulseaudio/internal/PulseAudioAudioSource.java
```diff
@@ -0,0 +1,250 @@
...
+ * @author Miguel Álvarez - Initial contribution
+ *
+ */
+@NonNullByDefault
+public class PulseAudioAudioSource extends PulseaudioSimpleProtocolStream implements AudioSource {
+
+    private final Logger logger = LoggerFactory.getLogger(PulseAudioAudioSource.class);
+    private final Set<PipedOutputStream> pipeOutputs = new HashSet<>();
+    private final ScheduledExecutorService executor = ThreadPoolManager
+            .getScheduledPool("OH-binding-pulseaudio-source");
```
lolodomo, line 49: The convention is to use the thing UID in the name. If I correctly remember OH-<thingUID>-xxx
You have the thing handler, so you can the thing UID.

lolodomo: I am sorry, after checking, the naming convention is `OH-binding-<thingUID>-xxx` with the last part being optional.

And I think `executor` should be set in the constructor when `pulseaudioHandler` is defined.

lolodomo (approved):
Code LGTM

lolodomo:
Will wait a little before merging in case @peuter or @dalgwen would like to have a look.

GiviMAD:
@lolodomo I have found an error, that I think I can solve.
The problem, the pulseaudio module is not loaded, the binding load it, after that it's sending a list modules call, that one is failing most of the times when is done after the module loading, so another try is raised. The result, modules for sink and sources been loaded twice in the pulseaudio server.
It's not a big problem because after this everything works, but I think I can easily solve it by adding retries also on the listModules call.
I'll add a final commit for it.

GiviMAD:
Done, no more errors or duplicated modules with the latests changes.

I was using 'pactl list modules' to check the modules created by the binding.

lolodomo:
> The problem, the pulseaudio module is not loaded, the binding load it, after that it's sending a list modules call, that one is failing most of the times when is done after the module loading, so another try is raised. The result, modules for sink and sources been loaded twice in the pulseaudio server.

Rather than a reyry mechanism, the good approach could be to wait for the end of the module loading? Maybe just wait a little before the load and the command?

GiviMAD:
> > The problem, the pulseaudio module is not loaded, the binding load it, after that it's sending a list modules call, that one is failing most of the times when is done after the module loading, so another try is raised. The result, modules for sink and sources been loaded twice in the pulseaudio server.
> 
> Rather than a retry mechanism, the good approach could be to wait for the end of the module loading? Maybe just wait a little before the load and the command?

But the error is not related to the module loading I think is because a collision between the sink and the source sending commands to thee same server. Since I've added the retries in all that places the binding is performing really better for my and I didn't see frequent errors, which was happened before many times for me not just in the module load. So I will prefer to keep it like this as for me the retries have solved most of the problems I was facing with the binding.

bundles/org.openhab.binding.pulseaudio/README.md
```diff
@@ -31,7 +31,7 @@ You can use the GUI on the bindings page (click on the pulseaudio binding then "
 
 ```
 binding.pulseaudio:sink=true
-binding.pulseaudio:source=false
+binding.pulseaudio:source=true
```
dalgwen, line 34: As you set the "source" boolean to the default value "true" in the readme and in the PulseAudioBindingConfiguration file, could you also change it in the binding.xml ?
(also line 23 in the readme, the sink support is not anymore the only one activated by default, you coud mention the source support)

dalgwen:
Thanks @lolodomo for calling me out.
I planned to do this audio source, so I'm glad to see this PR and don't have to do it :) !
I left only one very minor comment for a default configuration value
Otherwise, I didn't manage to test the audio source functionnality, because I still have issue with the other voice services. I hope to make it soon.
So, even if I can't try, LGTM.

bundles/org.openhab.binding.pulseaudio/src/main/java/org/openhab/binding/pulseaudio/internal/PulseaudioClient.java
```diff
@@ -145,10 +145,29 @@ public boolean isConnected() {
...
+                if (updateRetries == 1) {
+                    logger.warn("IOException updating state: {}", e.getMessage());
+                }
+                updateRetries--;
+                try {
+                    Thread.sleep(0);
+                } catch (InterruptedException ignored) {
+                }
+            }
+            break;
```
lolodomo, line 163: With a break at this place, a retry will never happen. 
So if you tell me that the new code solved many problems, I doubt it is due any retry.
WDYT?

lolodomo (changes requested):

bundles/org.openhab.binding.pulseaudio/src/main/java/org/openhab/binding/pulseaudio/internal/PulseaudioClient.java
GiviMAD: I don't know what to think, 🤔 I've tested the changes and see the duplicate problem solved. I don't know if I break the mechanism before the commit or it was fixed by other reason. I will test it again this afternoon and try to understand what happened.

GiviMAD: I think I understood what is happening, the modules info was loaded before, so my changes only catch the error on the update call, and let the process continue. So it was normal I stop seeing errors as they were all ignored. 
And everything works as the modules are loaded always before turn the things online.

That makes me realise that I can change the place were the update is taking place as it's not really required to reload the modules before looking for a valid module, just after load a new one.

GiviMAD:
I've undo those changes, seems like syncronizing the sendRawCommand method in the pulseaudio client was the key, to avoid sinks and sources connected to the same server to give errors.

Also I have moved update the to occur after the module load. This was the reason the socket reconnection was taking so long, because it was trigger an update before checking if there is already a compatible module in memory. I was not able to identify the reason.
Thanks for the help @lolodomo

bundles/org.openhab.binding.pulseaudio/src/main/java/org/openhab/binding/pulseaudio/internal/PulseAudioAudioSource.java
```diff
@@ -0,0 +1,250 @@
...
+            byte[] b = new byte[1];
+            int bytesRead = read(b);
+            if (-1 == bytesRead) {
+                return bytesRead;
+            }
+            Byte bb = Byte.valueOf(b[0]);
+            return bb.intValue();
+        }
+
+        @SuppressWarnings("null")
```
lolodomo, line 227: Is it really necessary?

bundles/org.openhab.binding.pulseaudio/src/main/java/org/openhab/binding/pulseaudio/internal/handler/PulseaudioHandler.java
```diff
@@ -367,18 +415,73 @@ public String getHost() {
...
     }
 
     public int getIdleTimeout() {
-        return ((BigDecimal) getThing().getConfiguration()
-                .get(PulseaudioBindingConstants.DEVICE_PARAMETER_AUDIO_SINK_IDLE_TIMEOUT)).intValue();
+        var handler = getPulseaudioBridgeHandler();
+        if (handler == null) {
+            return 30000;
+        }
+        AbstractAudioDeviceConfig device = getPulseaudioBridgeHandler().getDevice(name);
```
lolodomo, line 475: Use variable `handler`

lolodomo (changes requested):
Final review

lolodomo (approved):
LGTM
Thank you @GiviMAD

PR #12380, opened 2022-02-26
[wemo] Refactor Insight Switch parser

### LINKED ISSUE #12105: [wemo] Refactoring WeMo Insight Switch parser
This issue described some possible next step after #12100.

Suggestions/plan:
- Create new subclasses from `WemoHandler` targeting each thing type: **Insight**, **Motion** and **Switch**/**Socket**.
- Make `WemoHandler` abstract.
- Extract relevant `onValueReceived` parts into each handler.
- For `WemoInsightHandler` method `onValueReceived()`: Extract parser to separate class and provide unit test for this. The unit test must cover variables **EnergyPerUnitCost** (currently not supported), **BinaryState** and **InsightParams**.

### Service variables
Example data to support in parser:
```
Received pair 'EnergyPerUnitCost':'4|2000|4' (service 'basicevent1') for thing 'wemo:insight:insight1'
Received pair 'BinaryState':'0|1642884254|0|0|0|1209600|13|0|0|0' (service 'basicevent1') for thing 'wemo:insight:insight1'
Received pair 'InsightParams':'1|1632573998|10371148|52746|1174820|1209600|39|26440|28068134|548930560|8000' (service 'insight1') for thing 'wemo:insight:insight2'
```

jlaur:
Fixes #12105
Continuation of #12120

Refactoring: Parsing of WeMo Insight Switch data string has been extracted to separate class:
- Improved robustness: Number of provided values is no longer fixed/assumed.
- Improved robustness: Non-numeric values could throw `NumberFormatException`.
- Reduced redundancy: Each conversion if only written once (in private methods).
- Unit tests are provided.

### DIFF
--- a/bundles/org.openhab.binding.wemo/src/main/java/org/openhab/binding/wemo/internal/InsightParser.java
+++ b/bundles/org.openhab.binding.wemo/src/main/java/org/openhab/binding/wemo/internal/InsightParser.java
@@ -1,0 +1,150 @@
+/**
+ * Copyright (c) 2010-2022 Contributors to the openHAB project
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package org.openhab.binding.wemo.internal;
+
+import java.math.BigDecimal;
+import java.math.RoundingMode;
+import java.time.Instant;
+import java.time.ZonedDateTime;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.TimeZone;
+
+import org.eclipse.jdt.annotation.NonNullByDefault;
+import org.openhab.core.library.types.DateTimeType;
+import org.openhab.core.library.types.DecimalType;
+import org.openhab.core.library.types.OnOffType;
+import org.openhab.core.library.types.QuantityType;
+import org.openhab.core.library.unit.Units;
+import org.openhab.core.types.State;
+import org.openhab.core.types.UnDefType;
+
+/**
+ * Parser for WeMo Insight Switch values.
+ *
+ * @author Jacob Laursen - Initial contribution
+ */
+@NonNullByDefault
+public class InsightParser {
+
+    private static final int INSIGHT_POSITION_STATE = 0;
+    private static final int INSIGHT_POSITION_LASTCHANGEDAT = 1;
+    private static final int INSIGHT_POSITION_LASTONFOR = 2;
+    private static final int INSIGHT_POSITION_ONTODAY = 3;
+    private static final int INSIGHT_POSITION_ONTOTAL = 4;
+    private static final int INSIGHT_POSITION_TIMESPAN = 5;
+    private static final int INSIGHT_POSITION_AVERAGEPOWER = 6;
+    private static final int INSIGHT_POSITION_CURRENTPOWER = 7;
+    private static final int INSIGHT_POSITION_ENERGYTODAY = 8;
+    private static final int INSIGHT_POSITION_ENERGYTOTAL = 9;
+    private static final int INSIGHT_POSITION_STANDBYLIMIT = 10;
+
+    private final String value;
+
+    public InsightParser(String value) {
+        this.value = value;
+    }
+
+    /**
+     * Parse provided string of values.
+     *
+     * @return Map of channel id's with states
+     */
+    public Map<String, State> parse() {
+        HashMap<String, State> result = new HashMap<>();
+        String[] params = value.split("\\|");
+        for (int i = 0; i < params.length; i++) {
+            String value = params[i];
+            switch (i) {
+                case INSIGHT_POSITION_STATE:
+                    result.put(WemoBindingConstants.CHANNEL_STATE, getOnOff(value));
+                    break;
+                case INSIGHT_POSITION_LASTCHANGEDAT:
+                    result.put(WemoBindingConstants.CHANNEL_LASTCHANGEDAT, getDateTime(value));
+                    break;
+                case INSIGHT_POSITION_LASTONFOR:
+                    result.put(WemoBindingConstants.CHANNEL_LASTONFOR, getNumber(value));
+                    break;
+                case INSIGHT_POSITION_ONTODAY:
+                    result.put(WemoBindingConstants.CHANNEL_ONTODAY, getNumber(value));
+                    break;
+                case INSIGHT_POSITION_ONTOTAL:
+                    result.put(WemoBindingConstants.CHANNEL_ONTOTAL, getNumber(value));
+                    break;
+                case INSIGHT_POSITION_TIMESPAN:
+                    result.put(WemoBindingConstants.CHANNEL_TIMESPAN, getNumber(value));
+                    break;
+                case INSIGHT_POSITION_AVERAGEPOWER:
+                    result.put(WemoBindingConstants.CHANNEL_AVERAGEPOWER, getPowerFromWatt(value));
+                    break;
+                case INSIGHT_POSITION_CURRENTPOWER:
+                    result.put(WemoBindingConstants.CHANNEL_CURRENTPOWER, getPowerFromMilliWatt(value));
+                    break;
+                case INSIGHT_POSITION_ENERGYTODAY:
+                    result.put(WemoBindingConstants.CHANNEL_ENERGYTODAY, getEnergy(value));
+                    break;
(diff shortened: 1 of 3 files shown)


bundles/org.openhab.binding.wemo/src/main/java/org/openhab/binding/wemo/internal/InsightParser.java
```diff
@@ -0,0 +1,150 @@
...
+                case INSIGHT_POSITION_STANDBYLIMIT:
+                    result.put(WemoBindingConstants.CHANNEL_STANDBYLIMIT, getPowerFromMilliWatt(value));
+                    break;
+            }
+        }
+        return result;
+    }
+
+    private State getOnOff(String value) {
+        return "0".equals(value) ? OnOffType.OFF : OnOffType.ON;
```
lolodomo, line 108: You prefer that option to OnOffType.from(value) ?
It could make sense if the value could have other value than 0 and 1.

bundles/org.openhab.binding.wemo/src/main/java/org/openhab/binding/wemo/internal/InsightParser.java
```diff
@@ -0,0 +1,150 @@
...
+
+    private State getDateTime(String value) {
+        long lastChangedAt = 0;
+        try {
+            lastChangedAt = Long.parseLong(value);
+        } catch (NumberFormatException e) {
+            return UnDefType.UNDEF;
+        }
+        ZonedDateTime zoned = ZonedDateTime.ofInstant(Instant.ofEpochSecond(lastChangedAt),
+                TimeZone.getDefault().toZoneId());
```
lolodomo, line 119: Maybe it would be better to consider the timezone setup by the OH server ?

bundles/org.openhab.binding.wemo/src/main/java/org/openhab/binding/wemo/internal/handler/WemoInsightHandler.java
```diff
@@ -71,94 +64,21 @@ public void onValueReceived(@Nullable String variable, @Nullable String value, @
...
+                    updateState(channel, state);
+                });
+
+                // Update helper channel onStandBy by checking if currentPower > standByLimit.
+                var standByLimit = (QuantityType<?>) results.get(WemoBindingConstants.CHANNEL_STANDBYLIMIT);
+                if (standByLimit != null) {
+                    var currentPower = (QuantityType<?>) results.get(WemoBindingConstants.CHANNEL_CURRENTPOWER);
+                    if (currentPower != null) {
+                        updateState(WemoBindingConstants.CHANNEL_ONSTANDBY,
+                                currentPower.intValue() > standByLimit.intValue() ? OnOffType.OFF : OnOffType.ON);
```
lolodomo, line 81: Here, you can rather use `OnOffType(currentPower.intValue() > standByLimit.intValue())`

lolodomo (changes requested):

bundles/org.openhab.binding.wemo/src/main/java/org/openhab/binding/wemo/internal/InsightParser.java
jlaur: I don't prefer it, but unfortunately it can be non-binary. When receiving the parameters from a HTTP call the value can be 8 for ON (just verified by testing). I have added a test case for this.

bundles/org.openhab.binding.wemo/src/main/java/org/openhab/binding/wemo/internal/handler/WemoInsightHandler.java
jlaur: Simplified.

bundles/org.openhab.binding.wemo/src/main/java/org/openhab/binding/wemo/internal/InsightParser.java
jlaur: Right, fixed.

hmerk (approved):
LGTM

bundles/org.openhab.binding.wemo/src/main/java/org/openhab/binding/wemo/internal/InsightParser.java
lolodomo: Ok, that is a good argument.

bundles/org.openhab.binding.wemo/src/main/java/org/openhab/binding/wemo/internal/InsightParser.java
lolodomo: I believe you misunderstood my comment. You are considering the default system timezone. My proposal was to consider the timezone setup in OH server by the user. This is delivered by the service `TimeZoneProvider `

jlaur: Ah, yes I did. That proposal would require some additional refactoring, if I understand it correctly, as the provider is injected into the handler factory and would need to be passed to the handler and into the parser instance.

Before doing that I need some more understanding. I just tried setting a different openHAB timezone than my system timezone, and I'm actually surprised that the DateTime item is displayed like "2022-02-27T19:51:31.000+0100" and not converted to local time. If used in a rule, for example, I don't think it makes a difference whether value is "2022-02-27T19:51:31.000+0100" or "2022-02-27T18:51:31.000+0200" as it's the same time. If value is presented in a UI it should be converted to local time in either case.

So a little confused now about actual consequence of changing anything. Can you point me to some documentation and/or explain benefits of providing configured timezone?

lolodomo: At my knowledge, UI is doing no date conversion but just display the date value as it is. That is why it is important to set it with the right timezone.
If you want to consider that later in another PR, that is fine to me, just let me know.

jlaur: Yes, I think would prefer to keep this PR as a clean technical refactoring of the parser without functional changes. I'll need to dig deeper into DateTime in openHAB, compare with other bindings and probably also fix other channels in the binding, as this PR is focusing only on the WeMo Insight Switch parser.

lolodomo (approved):
LGTM

bundles/org.openhab.binding.wemo/src/main/java/org/openhab/binding/wemo/internal/InsightParser.java
jlaur: @lolodomo - I've now investigated a bit more, so sharing my findings so far:
- [ZonedDateTime](https://docs.oracle.com/javase/8/docs/api/java/time/ZonedDateTime.html) quote: _should be viewed primarily as the combination of a LocalDateTime and a ZoneId_.
- 2022-02-28T21:24:52.000+0000 is the same unique moment in time as 2022-02-28T22:24:52.000+0100, but since ZonedDateTime contains local datetime and zone, not just a UTC timestamp, they are by definition different and can be shown differently when formatting with timezone, for example when displayed raw like "2022-02-28T22:24:52.000+0100".
- The Hue binding creates timestamps from local time as UTC: `new DateTimeType(ZonedDateTime.ofInstant(LocalDateTime.parse(String.valueOf(lastUpdated), ateTimeFormatter.ISO_LOCAL_DATE_TIME), ZoneOffset.UTC, ZoneId.systemDefault()));` and uses explicit formatting in the corresponding channel definition: `<state readOnly="true" pattern="%1$tY-%1$tm-%1$td %1$tH:%1$tM:%1$tS"/>` so will be displayed like "2022-02-28 22:24:52" as local time.
- The Netatmo binding creates timestamps from epoch and uses `TimeZoneProvider` to store them with configured timezone in openHAB. No pattern is provided for the channels, but an item can be created with custom formatting, for example `DateTime Netatmo_LastSeen "Last Seen [%1$td.%1tm.%1$tY %1$tR]" <time> { channel = "netatmo:NAMain:indoor:timestamp#last-seen" }` which would output "28.02.2022 22:24" as local time.
- When using custom formatting like shown, the timezone doesn't make a difference as it will result in the same local time. In this case it seems that openHAB doesn't use the configured timezone when displaying the item in MainUI. I don't know if other UI's respect that.
- When NOT using custom formatting, the raw timestamp is displayed, which of course will be different depending on how the ZonedDateTime was constructed (as this is the internal type within DateTimeType).

IMHO: I think it will always be preferable to convert to local time in UI's. Displaying a raw ZonedDateTime doesn't seem user-friendly under most circumstances. I would say the difference between using system timezone and openHAB configured timezone when **constructing** ZonedDateTimes is quite minor/theoretical. However, when **converting** those ZonedDateTimes into a UI representation, the openHAB-configured timezone **should** be considered - quite similar to language.

PR #12386, opened 2022-02-28
[jrubyscripting] Filter out empty require options

jimtng:
Found a small issue when dealing with empty or erroneous data. Avoid injecting require statement when the script name is empty.


wborn (approved):
Thanks for the filtering!

PR #12393, opened 2022-03-01
[snowboyks] Initial contribution

GiviMAD:

Another keyword spotter.
It was a product from the closed Kitt-AI startup, but still there are many projects using it, and since they close the code for model generation is available.
More details in the readme.


GiviMAD:
@lolodomo the code under the "generated" folder was generated with swig, I added the autor comment and the top Copyright notice. 
Should I solve other problems reported in them (like incorrect methods names and other problems in the report.html)? or there is a way to ignore them?

GiviMAD:
I have to recompile the Debian binaries, I didn't link a library statically so it's required at runtime.

GiviMAD:
Closing this until I tested it a little more.

PR #12398, opened 2022-03-02
[rrd4j] Upgrade to rrd4j 3.8.1

wborn:
This should prevent a resource leak when generating graphs.

See: https://github.com/rrd4j/rrd4j/blob/master/changelog.txt


jlaur (approved):
Thanks!

PR #12399, opened 2022-03-02
[ipcamera] Fix Digest bug and Doorbird fixes.

Skinah:
These changes fix an issue reported on the forum here and some general bugs I noticed whilst looking over the users logs and the code:
https://community.openhab.org/t/ipcamera-blank-imageurl-channel/133457/14

Changes are:
1. Doorbird is not ONVIF so this stops the binding trying to connect with ONVIF.
2. Digest was reporting to the camera to `keep alive` the connection right before the binding closed the connection right before making a new connection. This appears to throw off the camera which only allows a single connection at a time for some Doorbird cameras.
3. When a non 200 OK reply came in like a 500 busy error code, the binding was trying to still capture a snapshot.


lolodomo (approved):
LGTM

PR #12406, opened 2022-03-03
[deutschebahn] Fixed order of stops

### LINKED ISSUE #12405: [deutschebahn] Order of stops wrong
The order of trains is depending from the setting of the train-filter attribute.
If only arrivals are containd within the timetable they must be sorted by the time of arrival.
If only departures or all trains are displayed, they must be ordered by depareture.

Before this fix they get always sorted by arrival.

soenkekueper:
Fixes #12405


jlaur (approved):
Excellent, thanks!

PR #12410, opened 2022-03-03
[porcupineks] fix build-in keywords on windows, fix service name and add missed modified

GiviMAD:
This pr:
* fixes the service name 
* fixes the extraction of the build in keywords in windows
* adds missed modified method

### DIFF
--- a/bundles/org.openhab.voice.porcupineks/src/main/java/org/openhab/voice/porcupineks/internal/PorcupineKSConstants.java
+++ b/bundles/org.openhab.voice.porcupineks/src/main/java/org/openhab/voice/porcupineks/internal/PorcupineKSConstants.java
@@ -25,7 +25,7 @@ public class PorcupineKSConstants {
     /**
      * Service name
      */
-    public static final String SERVICE_NAME = "Porcupine Keyword Spotter";
+    public static final String SERVICE_NAME = "Porcupine";
 
     /**
      * Service id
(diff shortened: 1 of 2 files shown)


bundles/org.openhab.voice.porcupineks/src/main/java/org/openhab/voice/porcupineks/internal/PorcupineKSConstants.java
```diff
@@ -25,7 +25,7 @@
     /**
      * Service name
      */
-    public static final String SERVICE_NAME = "Porcupine Keyword Spotter";
+    public static final String SERVICE_NAME = "Porcupine";
```
lolodomo, line 28: Ok fine but in this case please add "Keyword Spotter" to label in `@ConfigurableService(category`.
See how it is done in other TTS or STT service.

lolodomo (changes requested):

bundles/org.openhab.voice.porcupineks/src/main/java/org/openhab/voice/porcupineks/internal/PorcupineKSConstants.java
GiviMAD: I have the feeling I have to move it to somewhere else, thank you!

lolodomo (approved):
LGTM

PR #12414, opened 2022-03-04
[pulseaudio] fix null pointer exception and ensure source bg task stops

GiviMAD:

@lolodomo this is a small change to ensure the bg task for the source stops if no more streams are open.
I don't know how this can happen but it happened for me once.

Also includes a fix for null pointer exception that is raised when you have a Sink/Source with the register of the openHAB Sink/Source option enabled, and it is not offline (disconnected from the pulseaudio server).
I discover this earlier when I change the speaker for a new one, a Jabra I bought from amazon warehouse.

### DIFF
--- a/bundles/org.openhab.binding.pulseaudio/src/main/java/org/openhab/binding/pulseaudio/internal/PulseAudioAudioSource.java
+++ b/bundles/org.openhab.binding.pulseaudio/src/main/java/org/openhab/binding/pulseaudio/internal/PulseAudioAudioSource.java
@@ -138,7 +138,7 @@ public class PulseAudioAudioSource extends PulseaudioSimpleProtocolStream implem
             this.pipeWriteTask = executor.submit(() -> {
                 int lengthRead;
                 byte[] buffer = new byte[1024];
-                while (true) {
+                while (pipeOutputs.size() > 0) {
                     var stream = getSourceInputStream();
                     if (stream != null) {
                         try {
@@ -156,6 +156,7 @@ public class PulseAudioAudioSource extends PulseaudioSimpleProtocolStream implem
                         logger.warn("Unable to get source input stream");
                     }
                 }
+                this.pipeWriteTask = null;
             });
         }
     }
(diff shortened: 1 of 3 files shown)


bundles/org.openhab.binding.pulseaudio/src/main/java/org/openhab/binding/pulseaudio/internal/handler/PulseaudioHandler.java
```diff
@@ -414,9 +423,12 @@ public String getHost() {
      * @return the port on which the pulseaudio server is listening for this sink
      * @throws InterruptedException when interrupted during the loading module wait
      */
-    public int getSimpleTcpPort() throws InterruptedException {
+    public int getSimpleTcpPort() throws IOException, InterruptedException {
```
lolodomo, line 426: Please update the javadoc and check that the new exception is correctly handled by the caller.

bundles/org.openhab.binding.pulseaudio/src/main/java/org/openhab/binding/pulseaudio/internal/PulseAudioAudioSource.java
```diff
@@ -138,7 +138,7 @@ private void startPipeWrite() {
             this.pipeWriteTask = executor.submit(() -> {
                 int lengthRead;
                 byte[] buffer = new byte[1024];
-                while (true) {
+                while (pipeOutputs.size() > 0) {
```
lolodomo, line 141: Rather use isEmpty

lolodomo (changes requested):

bundles/org.openhab.binding.pulseaudio/src/main/java/org/openhab/binding/pulseaudio/internal/handler/PulseaudioHandler.java
GiviMAD: Yes the only call is from connectIfNeeded which already throws IOException

lolodomo (approved):
LGTM

Issue #12299, opened 2022-02-16
[danfossairunit] After upgrading from 3.1.1 to 3.2.0 my Danfoss Air A2 does not provide the renamed manual_fan_step channel


sonnin:
I just upgraded from 3.1.1 to 3.2.0 and renamed my manual_fan_speed channel link to manual_fan_step. But when I look at the danfossair thing I do not see the manua

## Expected Behavior
The danfossairunit thing to have a manual_fan_step channel as described in the documentation

## Current Behavior
I see not "main#manual_fan_step" channel

## Possible Solution
?
## Your Environment
Docker with OH 3.2.0 image
running on ubunto 
Linux openhab3 5.4.0-89-generic #100-Ubuntu SMP Fri Sep 24 14:50:10 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux

lolodomo:
Very interesting. 
@jlaur : that's for you ;) 
It looks to me like what I mentioned it could happen. I am interested by the conclusion of this issue... for the future.

jlaur:
@sonnin - thanks for reporting this issue. A few questions to trouble-shooting:
- Is your thing configured in UI or text file?
- How did you rename the channel:
- If UI, I would assume by removing linked item(s) and creating new ones?
- If text, did you simply rename the channel and saved? In this case I suspect openHAB have an issue, and I think you would first need to comment the item and save, then uncomment with new channel name, so it would be two steps.
- Can you provide a screenshot showing what you mean by not seeing the new channel?

jlaur:
For reference, here are the related issues:
- #10621
- #11667

sonnin:
The thing was autodiscovered in earlier version.
UID: danfossairunit:airunit:-1407909348
label: Danfoss HRV
thingTypeUID: danfossairunit:airunit
configuration:
  host: danfossair.xxx.yy
  refreshInterval: 10
  updateUnchangedValuesEveryMillis: 60000

Items linking to the channels is written in files.

After upgrading I updated my item(s) to match the new channel and data types (like temperature).

When I say that I cannot see the channel. I mean that it does not show up in the UI:
![Screenshot_20220216_130919](https://user-images.githubusercontent.com/28400676/154262378-499f34ef-24ae-4d5f-a537-604f0caa7948.png)
Checkings the logs I do however not see any statements that the item is linked to an unknown channel.

I will try to change play with the items and perhaps remove the thing and re-add it. And revert back.

[...]

PR #12420, opened 2022-03-05
[fronius] Add Autonomy and Self Consumption channels

jimtng:
* Add Autonomy and Self Consumption channels
* Change Inverter channels from DecimalType to QuantityType
* Convert ValueUnit to QuantityType, honour the unit from ValueUnit
* Replace custom unit converter with QuantityType's unit conversion
* Return QuantityType of zero for missing ValueUnit data, to be consistent with P_AC.
* Return UnDefType.NULL for other missing data
* De-duplicate URL parsing routine


bundles/org.openhab.binding.fronius/src/main/java/org/openhab/binding/fronius/internal/handler/FroniusSymoInverterHandler.java
```diff
@@ -84,65 +84,57 @@ protected Object getValue(String channelId) {
...
-                case FroniusBindingConstants.INVERTER_DATA_CHANNEL_YEAR:
-                    ValueUnit year = inverterRealtimeResponse.getBody().getData().getYearEnergy();
-                    if (year != null) {
-                        year.setUnit("MWh");
-                    }
-                    return year;
+                    return getQuantityOrZero(inverterData.getPac(), Units.WATT);
                 case FroniusBindingConstants.INVERTER_DATA_CHANNEL_FAC:
-                    return inverterRealtimeResponse.getBody().getData().getFac();
+                    return inverterData.getFac();
```
lolodomo, line 92: You define a dimension for the channel: Number:Frequency
So a QuantityType is expected.

lolodomo (changes requested):

lolodomo:
As a side note, using ultra generic channel types is not a good practice.

bundles/org.openhab.binding.fronius/src/main/java/org/openhab/binding/fronius/internal/handler/FroniusSymoInverterHandler.java
jimtng: InverterRealtimeBodyData::getFac() returns a `ValueUnit`, which in FroniusBaseThingHandler::updateChannel() gets converted into QuantityType. 

I understand this may not be immediately apparent, so do you think I should do the conversion here and return a QuantityType?

jimtng:
> As a side note, using ultra generic channel types is not a good practice.

Could you please elaborate?

bundles/org.openhab.binding.fronius/src/main/java/org/openhab/binding/fronius/internal/handler/FroniusSymoInverterHandler.java
lolodomo: Maybe getValue should even be now updated to return a `State` rather than `Object` ?

lolodomo:
> Could you please elaborate?

Technically, it is possible to define one unique channel type "string" that will then be used for any string channel by renaming its label and description. But the original idea was more to have channel types carrying concrete meaning.

You did not do it for string but for power, energy, ... This is maybe not as extreme as my example but almost.

jimtng:
> Technically, it is possible to define one unique channel type "string" that will then be used for any string channel by renaming its label and description. But the original idea was more to have channel types carrying concrete meaning.
> 
> You did not do it for string but for power, energy, ... This is maybe not as extreme as my example but almost.

I think you're referring to the xml files? So what's the purpose of channel type if they are merely duplicating the information for channels? It seems highly redundant.

bundles/org.openhab.binding.fronius/src/main/java/org/openhab/binding/fronius/internal/handler/FroniusSymoInverterHandler.java
jimtng: I've refactored it so getValue returns a State.

jimtng:
@lolodomo I've rewritten the channel-types so that they're not being shared by multiple channels.

bundles/org.openhab.binding.fronius/src/main/java/org/openhab/binding/fronius/internal/handler/FroniusBaseThingHandler.java
```diff
@@ -100,31 +99,16 @@ protected void updateChannel(String channelId) {
             return;
         }
 
-        Object value = getValue(channelId);
-        if (value == null) {
-            logger.debug("Value retrieved for channel '{}' was null. Can't update.", channelId);
-            return;
-        }
+        State state = getValue(channelId);
+        ;
```
lolodomo, line 103: Line to be removed

bundles/org.openhab.binding.fronius/src/main/resources/OH-INF/thing/inverter.xml
```diff
@@ -0,0 +1,162 @@
...
+			<channel id="inverterdatachanneludc" typeId="inverter_udc"/>
+			<channel id="inverterdatachannelidc" typeId="inverter_idc"/>
+			<channel id="inverterdatachannelfac" typeId="inverter_fac"/>
+
+			<channel id="inverterdatachanneldayenergy" typeId="inverter_energy_day"/>
+			<channel id="inverterdatachannelyear" typeId="inverter_energy_year"/>
+			<channel id="inverterdatachanneltotal" typeId="inverter_energy_total"/>
+
+			<channel id="inverterdatadevicestatusstatuscode" typeId="inverter_statuscode"/>
+			<channel id="inverterdatadevicestatuserrorcode" typeId="inverter_errorcode"/>
```
lolodomo, line 29: Now, you make the reverse, a channel type for each channel.
Here for example, `devicestatus_errorcode` looks OK as channel type for inverterdatadevicestatusstatuscode and statecode. No need to create 2 different channel types with exactly the same label and description.
Same for `devicestatus_statuscode`.
Maybe there are other cases, I have not checked in detail.

bundles/org.openhab.binding.fronius/src/main/resources/OH-INF/thing/inverter.xml
```diff
@@ -0,0 +1,162 @@
...
+	<thing-type id="powerinverter">
+		<supported-bridge-type-refs>
+			<bridge-type-ref id="bridge"/>
+		</supported-bridge-type-refs>
+
+		<label>Fronius Symo Inverter</label>
+		<description>Fronius Symo Power Inverter</description>
+
+		<channels>
+			<channel id="inverterdatachannelpac" typeId="inverter_pac"/>
```
lolodomo, line 17: You renamed all channel types, this is of course a breaking change (for users using UI to define things).
I mean you renamed `pac` into `inverter_pac`. Is it really necessary ?
I would suggest, if possible, to not rename any channel type name (and channel name).

lolodomo (changes requested):

lolodomo:
@jimtng : you probably misunderstood my previous comment as you took the opposite direction. Of course, some channel types can be used in several thing types, when they have the exact same meaning (like status code or error code for example). You can even sometimes override the label/description if this looks appropriate.
And please avoid renaming channel types when not required (as it is a breaking change).

bundles/org.openhab.binding.fronius/src/main/resources/OH-INF/thing/inverter.xml
jimtng: I didn't know that changing channel type id would break anything? I thought only channel id matters?

lolodomo: Yes, it is (only) a problem when the thing was created with UI.

jimtng: @lolodomo are you sure that changing typeId is a user breaking change?

lolodomo: See #12299

jimtng:
> @jimtng : you probably misunderstood my previous comment as you took the opposite direction. Of course, some channel types can be used in several thing types, when they have the exact same meaning (like status code or error code for example). You can even sometimes override the label/description if this looks appropriate. And please avoid renaming channel types when not required (as it is a breaking change).

I am completely confused now. The xml structure was a mess and I thought I tidied it up with this PR. It appears that I do not understand how it should be done.

Should I just revert all the changes to the xml file and just add the two new channels and leave the rest like before, even though I think it was a mess?

bundles/org.openhab.binding.fronius/src/main/resources/OH-INF/thing/inverter.xml
jimtng: > See #12299

That was the change of the `channel id` that caused the problem, not `channel-type id`.
My understanding is that the change in `channel-type id` has no bearing in the referenced issue whatsoever. My understanding is that you could even call it any random character and it would not matter.  

In any case I'll find some time to test this.

I would really rather rename the `channel-type id` (not `channel id`) to tidy things up.

bundles/org.openhab.binding.fronius/src/main/java/org/openhab/binding/fronius/internal/handler/FroniusBaseThingHandler.java
jimtng: Oops thanks for spotting this. removed.

lolodomo:
> Should I just revert all the changes to the xml file and just add the two new channels and leave the rest like before, even though I think it was a mess?

I agree that it was not perfect before.
For example, `energyrealsumconsumed` on meter and ohmpilot have different channel types while I understand they are the same.
I think a good approach would have been to define 2 channel types: energy_consumed and energy_produced. The channel energyrealsumconsumed on thing meter and the channel energyrealsumconsumed on thing ohmpilot will be of type energy_consumed while the channel energyrealsumproduced on thing meter will be of type energy_produced.
As another example, it looks to me that meter_powerreal and ohmpilot_powerreal are similar and a common channel type would have been appropriate. Using meter_powerreal for each phase and the sum is ok for me.

So yes, it was a mess. And cleaning that will be impossible without breaking changes.

bundles/org.openhab.binding.fronius/src/main/resources/OH-INF/thing/inverter.xml
jimtng: I did some tests and looked into the json database, and it appears that you're correct and I was mistaken. A change in the channel type id matters for things created with UI as you said. I will revert my changes.

jimtng:
I've reverted the xml file reorganisation. It now only includes the added channels and type changes.

lolodomo (approved):
LGTM, thank you

PR #12423, opened 2022-03-05
[pulseaudio]  Fix exception handling when connecting

### LINKED ISSUE #12419: [pulseaudio] Bridge thing remaining in INITIALIZING status in case of bad configuration
If you add a new Pulseaudio server with a wrong IP or wrong port, the thing will remain in INITIALIZING status for ever.
If you try to fix the thing configuration, the thing remains in INITIALIZING status. You have to stop/start the bundle to get the ONLINE status.

I will propose a fix.

### LINKED ISSUE #12424: [pulseaudio] Bridge status remains ONLINE when the pulseaudio server is no more reachable
When the pulseaudio server becomes unreachable (shutdown for example), the thing (bridge) status remains ONLINE and log file is flooded with errors including full stack trace.

What is expected is the thing status switching to OFFLINE without any regular logs (except if DEBUG logs are enabled by the user).

I will fix that issue too.

lolodomo:
Fix bridge/thing status update
Also update log levels

Fix #12419
Fix #12424


wborn (approved):
Thanks for the fixes and cleanup!

PR #12425, opened 2022-03-05
GHA CI build workflow improvements

wborn:
* Add error annotations only in Java 11 matrix build to prevent duplicates
* Make sure Java 11 build is not cancelled when Java 17 build fails so it can add annotations by using `fail-fast: false`
* Use changed files only for incremental PR builds so a full build is done for changes merged into 'main' branch
* Add GHA build status badge


jlaur (approved):
Thanks!

PR #8649, opened 2020-10-03
removed embedded broker

kaikreuzer:
Replacement for https://github.com/openhab/openhab-addons/pull/8643

Signed-off-by: Kai Kreuzer <kai@openhab.org>

PR #12431, opened 2022-03-06
[mqtt] Revive disabled itests

wborn:
This fixes all the compilation/dependency issues in the MQTT itests so they can be reenabled again.
The tests now create and use their own Moquette instance instead of the removed embedded MQTT broker.
The moquette-broker JAR is also included in the test bundles as workaround for its missing OSGi bundle manifest headers.

See also: https://github.com/openhab/openhab-addons/pull/8649


jlaur (approved):
Great to have these tests back, thank you.

PR #12434, opened 2022-03-06
[knx] Allow decoding of KNX Data Secure frames

### LINKED ISSUE #7239: [knx] initial addresses read at startup breaks after exception
## Expected Behavior
all group addresses marked with "<" should be read on OH start and corresponding items get the current value set

## Current Behavior
On OH start it's possible to see a sequence of GroupValueRead/GroupValueResponse telegrams which terminates after the following exception in the log file:
```2020-03-27 19:48:44.642 [WARN ] [.internal.handler.DeviceThingHandler] - Ignoring KNX bus data: couldn't transform to any Type (destination='3/6/230', datapoint='command DP 3/6/230 'knx:device:231_6128-01-500', DPT id 9.001, low priority', data='0x02')
2020-03-27 19:48:44.643 [WARN ] [mmon.WrappedScheduledExecutorService] - Scheduled runnable ended with an exception:
tuwien.auto.calimero.KNXIllegalArgumentException: DPT 9.001 Temperature: data length 1 < required datapoint type width 2
        at tuwien.auto.calimero.dptxlator.DPTXlator.setData(DPTXlator.java:257) ~[?:?]
        at tuwien.auto.calimero.process.ProcessCommunicatorImpl.extractGroupASDU(ProcessCommunicatorImpl.java:508) ~[?:?]
        at tuwien.auto.calimero.process.ProcessCommunicatorImpl.read(ProcessCommunicatorImpl.java:345) ~[?:?]
        at org.openhab.binding.knx.internal.client.AbstractKNXClient.readNextQueuedDatapoint(AbstractKNXClient.java:287) ~[?:?]
        at org.openhab.binding.knx.internal.client.AbstractKNXClient.lambda$1(AbstractKNXClient.java:199) ~[?:?]
        at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511) ~[?:1.8.0_201]
        at java.util.concurrent.FutureTask.runAndReset(FutureTask.java:308) ~[?:1.8.0_201]
        at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.access$301(ScheduledThreadPoolExecutor.java:180) ~[?:1.8.0_201]
        at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:294) ~[?:1.8.0_201]
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149) [?:1.8.0_201]
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624) [?:1.8.0_201]
        at java.lang.Thread.run(Thread.java:748) [?:1.8.0_201]
```
most of items remain not initialised having value "-" or "NULL".

## Possible Solution
The workaround is to specify DPT value in channel configuration that will match in size to data obtained from the bus, to allow conversion.

1. Identify a group address with size 1 byte (3/6/240)
2. Create a channel with configuration similar to this ```{"uid":"knx:device:241_6128-01-500:Climat_FF_Red_Mode_Status","id":"Climat_FF_Red_Mode_Status","channelTypeUID":"knx:number","kind":"STATE","itemType":"Number","label":"Red Climat Mode Status","linkedItems":["Climat_FF_Red_Mode_Status"],"configuration":{"ga":"<3/6/240"},"defaultTags":[]}```
3. Link this channel to some item
4. Restart OH
5. monitor openhab.log

## Context
I need all values to be read after a restart.

## Your Environment
* Version used: OH 2.5, knx  add-on 2

holgerfriedrich:
* add passive (listening only) access for KNX Data Secure frames, #8872
* add config options for KNX keyring file and password
* ease setup if IP Secure, as required parameters can be read from keyring
* add tests for security functions
* update user documentation


holgerfriedrich:
This PR brings support to KNX IP secure, and partial support for KNX data secure (listening only). See previous discussions in #8872.

What has been tested: 
* Secure tunneling, seems to work fine for a few days

What is to be tested by someone else:
* Secure routing (don't have a secure router at hand)
* CEMI frame support  (don't have a kberry at hand)

TODO for me: fix CI problems, for me the code compiles warning-free

holgerfriedrich:
@kaikreuzer I followed your advise and have only included the passive support for data secure. Outgoing writes and poll requests for secure group adresses are dropped at lowest level. 

Though, when I change a switch connected to a secure GA, the event log still shows the "Item ... predicted to become ..." (which of course will not happen as I drop the write). Is there a clean way to signal a failed write to the upper layer? Throwing an exception is probably too much, as it gets logged with a full stack trace.

bundles/org.openhab.binding.knx/src/main/resources/OH-INF/thing/serial.xml
```diff
@@ -34,6 +38,24 @@
 				<description>Seconds between connect retries when KNX link has been lost, 0 means never retry</description>
 				<default>0</default>
 			</parameter>
+			<parameter name="useCemi" type="boolean">
+				<label>Use CEMI</label>
+				<description>Use newer CEMI frame format. May be useful for newer serial devices like KNX RF sticks, kBerry, etc.</description>
+				<default>false</default>
+				<advanced>true</advanced>
+			</parameter>
+			<parameter name="keyringFile" type="text" groupName="knxsecure">
```
J-N-K, line 47: Wouldn't it make sense to add that directly to the JSON database (i.e. add the file content directly here instead of a filename).

slueder:
@holgerfriedrich 

I retested the keyring approach with the latest version of ETS (6.0.4) and reasonably strong keyring export passwords (8 characters, consisting of upper & lower letters, digits and a limited set of special chars :#,;) and it worked now.

I am using commit ID https://github.com/holgerfriedrich/openhab-addons/commit/c7114f50bed46a35f7e3d317709943c2d95f7ae5

Happy to re-test as soon as you let me know.

holgerfriedrich:
As initial support for KNX IP Secure has been merged with PR #12709 and is available since OH 3.0M1, this PR contains the next bits required for full support of KNX Secure. It extends the current implementation by using the KNX keyring export, which allows to read the credentials for IP Secure connection from keyring. More important, is allows to **decode KNX Data Secure frames** as well.
This is not a complete implementation, but allows to run on current version of Calimero library.

kaikreuzer:
>  Is there a clean way to signal a failed write to the upper layer? Throwing an exception is probably too much, as it gets logged with a full stack trace.

Sorry, I am probably a bit late with my answer here 🙄 . You might want to set the `autoUpdatePolicy` to `veto` for such channels - see https://www.openhab.org/docs/developer/bindings/thing-xml.html#auto-update-policies. In that case, the framework won't predict any state change and leaves everything to the binding. When you do not do anything, the state will simply stay as it is, so no need to throw an exception or similar.

Your latest comment sounds as if the PR is ready for review - but the title still has a [WIP] in it. Could you clarify the latest status of it, please?

holgerfriedrich:
Converting to draft again, as Calimero 2.5.1 was released a few hours ago.

holgerfriedrich:
@kaikreuzer I think we can start the review process.

As descibed above, this is the next step towards supporting KNX Data Secure, adding passive "listen only" mode to decode KNX frames.
Sending and initial read is disabled. This leads to the issue that event log shows "Item ... predicted to become ..." but the item will of course not change. I have not looked at setting ```autoUpdatePolicy``` yet.

We already use Calimero 2.5.1 on the main branch and thus I updated this PR.

The PR brings the following improvements:
- KNX keyring file import and unit test, this allows for decoding incoming encrypted frames
- SecureApplicationLayer provided by binding, i.e. allows for several KNX gateways at a time

jlaur:
@kaikreuzer, @J-N-K - would this be something for you to review for 4.0?

holgerfriedrich:
@J-N-K is there any easy way to set the autoupdate policy for a specific channel dynamically from the addon code? ```Channel``` has a method ```getAutoUpdatPolicy()``` but no obvious setter.... Device initialization loops through all channels using ```getThing().getChannels().forEach(channel...```, and there I would like to set the policy to VETO for the channels involving data secure....

J-N-K:
You can't change the auto-update policy dynamically. You can modify the thing (i.e. create a `ThingBuilder`, remove the channel, add a new channel with changed policy, update the thing).

J-N-K:
@holgerfriedrich What is the status here?

NorbertHD:
Has @holgerfriedrich forgotten this PR?

florian-h05:
@holgerfriedrich What is the state here?

holgerfriedrich:
@florian-h05 Basically the (read-only) part of data secure is ready for use. I am a bit undecided, because a worried about putting the password of the keyring export in clear text in the config.
This is why I have tried to provide a proof of concept for protecting passwords with a TPM - still not perfect, but better than passwords in clear text. #15326.

florian-h05:
Great to hear! (and sorry for my very delayed answer)
So this PR is finally finished?

lsiepel:
@holgerfriedrich what is the state if this PR?

florian-h05:
@holgerfriedrich Can we start with the review process here?

I understand your security concerns wrt to putting the keyring password in cleartext into the config, but IMO this is no major security problem. I would consider the openHAB server as a secure environment (use reasonable strong SSH password or even better public key auth), and in that case we have two layers of security that have to be breached to get access to the keyring: One needs to get the password for the keyring, either via REST API or via filesystem, and one has to access the keyring itself via filesystem. Only attacking REST API therefore wouldn‘t breach the security, and access to the filesystem can be secured very well (no SMB, FTP, NFS, only SSHFS and strong SSH security).
There is a remaining risk but the added security compared to unsecured KNX outweighs that risk, hence openHAB should implement KNX Data Secure even if not a „perfect“ implementation.

holgerfriedrich:
@florian-h05 agreed, let me rebase and go over the docs again. Then we can start the review.

lsiepel:
> I would consider the openHAB server as a secure environment (use reasonable strong SSH password or even better public key auth), and in that case we have two layers of security that have to be breached to get access to the keyring: One needs to get the password for the keyring, either via

How bad would it if some unauthorized individual got access to this keyring?
Does it only store the KNX secret?

florian-h05:
The keyring together with the password would allow an attacker who has physical access to the KNX bus or the network where the IP interface is located, to gain full access to the bus. 
Current situation is that you cannot use KNX data secure at all, so if you had the physical or network access to KNX at the moment, you could directly attack it. With data secure support in openHAB and data secure enabled, there is the additional step of stealing the keyring.

I have already had a conversation with Holger and the best way would be to be a KNX member as openHAB and register openHAB as a bus member via the ETS software, because this would allow us to get data secure keys that are only valid for a limited set of group addresses (simply said: parts of the bus) and not everything; but this would be a huge effort.

For a successful attack, you would hence either need to break into the building and get access to the KNX wire or in case of KNX RF be in range of the radio frequency, or have access to the network where the IP interface is. A obviously doesn’t allow remote attacks, and B usually requires to breach the firewall of the internet router or possibly additional firewalls.

If I were to attack such a system, I would rather use openHAB to send commands to KNX instead of stealing the keyring — because to steal the keyring, I need access to the server which I would then use to attack openHAB directly.

lsiepel:
> TLDR: It would allow an attacker to have full access to the bus IF he has physical (wired or radio frequency) or network-based (breach a firewall, create a VPN tunnel …) access. This keyring only stores keys required for de- and encrypting secured KNX telegrams.

Thanks for the extensive outline. My concerns are not toward the KNX bus, adding this security is an imrovement eitherway, totally agree. You confirmed the thoughts i had. 
My concern are that this keyring would also be a store to other secrets. But as you confirmed it is exclusive to KNX, i'm fine and let's move forward.

florian-h05:
Just FYI you program KNX using a program called ETS and the keyring required here is exported from the ETS, it’s the same keyring that allows the ETS full access to the KNX bus.

lsiepel:
@holgerfriedrich did this PR get some test mileage on different installations ?

holgerfriedrich:
@lsiepel unfortunately not yet, at least not after the rebase and latest modifications. 

I typically deploy to RPI, and I am using Enertex IP interface (connected via IP,  with IP secure), Weinzierl 730 (IP w/o IP secure), or Weinzierl kBerry (RPI hat, using serial connection).

Give me some time to double check that data secure (decoding incoming encrypted frames) still works as expected.....

bundles/org.openhab.binding.knx/README.md
```diff
@@ -66,6 +66,9 @@ At its base, the _ip_ bridge accepts the following configuration parameters:
...
 | tunnelUserPassword  | No           | KNX secure: Tunnel user key for secure tunnel mode                                                           | -                                                    |
 | tunnelDeviceAuthentication  | No   | KNX secure: Tunnel device authentication for secure tunnel mode                                              | -                                                    |
+| keyringFile         | No           | KNX secure: Keyring file exported from ETS and placed in openHAB config/misc folder. Mandatory to decode secure GAs. | -                                                    |
+| keyringPassword     | No           | KNX secure: Keyring file password (set during export from ETS)                                               | -                                                    |
+| tunnelSourceAddress | No           | KNX secure: Physical KNX address of tunnel in secure mode to identify tunnel. If given, openHAB will read tunnelUserId, tunnelUserPassword, tunnelDeviceAuthentication from keyring  | -         |
```
lsiepel, line 71: Minopr: could you reformat the table (other one too)

bundles/org.openhab.binding.knx/src/main/resources/OH-INF/thing/ip.xml
```diff
@@ -73,29 +73,60 @@
 				<description>Seconds between connection retries when KNX link has been lost, 0 means never retry, minimum 30s</description>
 				<default>60</default>
 			</parameter>
+			<parameter name="keyringFile" type="text" groupName="knxsecure">
+				<label>Keyring file</label>
+				<description>Keyring file exported from ETS and placed in openHAB config/misc folder, e.g.
+					knx.knxkeys. This file is
+					mandatory to decode secure GAs. It can provide settings
```
lsiepel, line 80: GAs ?

lsiepel (commented):
Left two comments. LGTM, but ift would also be nice of @J-N-K and @florian-h05 to confirm.

bundles/org.openhab.binding.knx/src/main/resources/OH-INF/thing/ip.xml
florian-h05: GAs = Group Addresses
It’s a KNX specific term and I’m pretty sure the users of the KNX binding should understand it, but probably would be good to write it out.

florian-h05:
I am currently busy till beginning of October, but then should be able to review.

holgerfriedrich:
Accidentally pushed to this PR when working on a followup PR, rolled back, rebased to current main as a force push was anyway necessary.
Sorry for the noise.

lsiepel (approved):

florian-h05 (approved):
Code LGTM, thanks! Just one minor comment:

I would have loved to give it a try, but I don't have KNX Data nor IP Secure devices.
I have of course tested it against my system - everything still works as expected.

bundles/org.openhab.binding.knx/src/main/java/org/openhab/binding/knx/internal/client/AbstractKNXClient.java
```diff
@@ -206,38 +207,46 @@ private synchronized boolean connect() {
...
+            // Protected ctor using given ManagementClientImpl is available (custom class to be inherited)
+            managementProcedures = new CustomManagementProceduresImpl(managementClient, tl);
+
             // OH helper for reading device info, based on managementClient above
             deviceInfoClient = new DeviceInfoClientImpl(managementClient);
 
             // ProcessCommunicator provides main KNX communication (Calimero).
-            // Note for KNX Secure: SAL to be provided
-            ProcessCommunicator processCommunicator = new ProcessCommunicatorImpl(link);
+            final var useGoDiagnostics = true;
```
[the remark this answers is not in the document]
holgerfriedrich: @florian-h05 done

florian-h05:
@lsiepel You can merge now :rocket:

PR #12437, opened 2022-03-06
[homematic] Fix invalid default values ending up in the thing type

Flole998:
Sometimes invalid default values ended up in the default value for a channel of a thing type. Initializing the thing would fail completely complaining that it is not an allowed option. This patch makes sure those values are actually valid and attempts to correct them if they are invalid.

This fixes a regression introduced by the core now checking validity of values even from text-based configs.

### DIFF
--- a/bundles/org.openhab.binding.homematic/src/main/java/org/openhab/binding/homematic/internal/type/HomematicTypeGeneratorImpl.java
+++ b/bundles/org.openhab.binding.homematic/src/main/java/org/openhab/binding/homematic/internal/type/HomematicTypeGeneratorImpl.java
@@ -342,6 +342,33 @@ public class HomematicTypeGeneratorImpl implements HomematicTypeGenerator {
                                     }
                                 });
                         builder.withOptions(options);
+                        if (dp.isEnumType()) {
+                            logger.trace("Checking if default option {} is valid",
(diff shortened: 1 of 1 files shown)


bundles/org.openhab.binding.homematic/src/main/java/org/openhab/binding/homematic/internal/type/HomematicTypeGeneratorImpl.java
```diff
@@ -342,6 +342,32 @@ public ParameterOption createOption(String value, String description) {
...
+                            boolean needsChange = true;
+                            for (ParameterOption option : options) {
+                                if (option.getValue().equals(Objects.toString(dp.getDefaultValue(), ""))) {
+                                    needsChange = false;
+                                    break;
+                                }
+                            }
+                            if (needsChange) {
+                                String defStr = Objects.toString(dp.getDefaultValue(), "0");
+                                if (defStr == null)
```
lolodomo, line 357: Can defStr be null after the previous call? My feeling is no. 

In case it could be null, keep the if but surround it with {}

lolodomo (changes requested):

lolodomo (approved):
LGTM

Issue #12435, opened 2022-03-06
[androiddebugbridge] discovery for non firestick devices is broken


GiviMAD:
Device discovery for normal android devices was broken.

@cweitkamp I didn't understand some changes you did there.
You try to use the interface macAddress as if it was the device Mac address. (also the conversion from bytes to string was wrong).
Apart from that I don't see the use of the Mac address as a reliable identifier. Maybe I'm wrong but you could have different Mac address in your different interfaces so this is not a valid identifier. 

I will revert to use the device serial number as identifier, which will require to connect through adb from the mdns discovery service. But this is the only way I think both discovery methods could behave fine together. Also will ensure you have enabled adb before consider the thing discovered.

Let me know your thoughts.

PR #12438, opened 2022-03-06
[androiddebugbridge] add start intent channel

GiviMAD:

This pr add a new channel for start application intents.
Also fixes the discovery and revert to use the device serial number as identifier on both discovery methods.


GiviMAD:
related to https://github.com/openhab/openhab-addons/issues/12435

jaywiseman:
I'm using this version of the binding.

https://docs.smarthomej.org/3.2.12-SNAPSHOT/org.smarthomej.binding.androiddebugbridge.html

This version has the Intent channel defined already.

I have these items working with the binding I'm referencing above.

https://community.openhab.org/t/android-debug-binding-adb-notes-for-others/133777

Best, Jay

lsiepel:
Did a quick scan for this PR. Usually it is preferred to seperate bug and enhancements PR. Anyway, did you get in contact with previous contributer that changed the identifier for discovery?

GiviMAD:
@lsiepel, yes he is one of the reviewers also I mentioned him in the issue, and for the reference I'm the original contributor.

I was working on the new channel when I discovered the issue so I fixed it to been able to tested it.

GiviMAD:
@jaywiseman I have no clue, I don't think it will work without changes.

GiviMAD:
@cweitkamp did you have a change to look at this?
We can take a different path and use the ip as identifier if you think that connect through add from the discovery is too aggressive.

cweitkamp:
Sry for my late reply. As I already stated in https://github.com/openhab/openhab-addons/pull/11881#discussion_r776251787 for me it looks like the Mac address is the only unique identifier available in both discovery service implementations.

The mDNS discovery in general does not require to enable usb debugging in general as requested in the documentation. Imo your approach will fail to discover a new device as long as this requirement is not given. The question now is if we want to restrict the discovery results to rely on this setting or not.

May I ask you to slit this PR into two separate ones? One for the discovery issue and one for the enhancement to add the new channel. Thanks.

GiviMAD:
Sorry but I can’t agree @cweitkamp.

  First of all we don’t have the mac address in both discovery methods, the way currently implemented on the normal discovery is totally wrong and I couldn’t find an easy way to get it, also the serial number used util that pr seems the best option and with this changes is available on both sides.
  
    Also currently we have broken a rudimentary but functional discovery that works for all android devices, just for adding one for discover FireTV devices, which is not the only target of this binding, so for me it’s more a problem right now, so if we could not find a good solution quickly I think its better to remove the new discovery method and implemented it again when we know how to do without break the current one.  
    
    And finally as I said before I have added the fix to this pr because I discovered it was broken at that moment, I can move it to a separate one if is needed, but seems like and extra work for me and I really don’t see the point as the pr is not huge.

GiviMAD:
Removed all discovery related changes. 
So the problem with the discovery remains open as the MAC address is not available in the adb discovery method.

bundles/org.openhab.binding.androiddebugbridge/README.md
```diff
@@ -101,6 +102,16 @@ Please note that events could fail if the input method is removed, for example i
 | wake-lock            | Number | Power wake lock value                                                                                                         |
 | screen-state         | Switch | Screen power state                                                                                                            |
 
+#### Start Intent
+
+This channel allows to call am start, the sintax por it is:
```
jlaur, line 107: Some typos here.

bundles/org.openhab.binding.androiddebugbridge/README.md
```diff
@@ -101,6 +102,16 @@ Please note that events could fail if the input method is removed, for example i
...
+<package/activity>||<<arg name>> <arg value>||...
+
+This is a sample:
+com.netflix.ninja/.MainActivity||<a>android.intent.action.VIEW||<d>netflix://title/80025384||<f>0x10000020||<es>amzn_deeplink_data 80025384
+
+Not all the (arguments)[https://developer.android.com/studio/command-line/adb#IntentSpec] are supported, open an issue or pr if you need more.
```
jlaur, line 113:
```suggestion
Not all the (arguments)[https://developer.android.com/studio/command-line/adb#IntentSpec] are supported. Please open an issue or pull request if you need more.
```

bundles/org.openhab.binding.androiddebugbridge/src/main/java/org/openhab/binding/androiddebugbridge/internal/AndroidDebugBridgeDevice.java
```diff
@@ -299,7 +304,8 @@ public String getSerialNo() throws AndroidDebugBridgeDeviceException, Interrupte
 
     public String getMacAddress() throws AndroidDebugBridgeDeviceException, InterruptedException,
             AndroidDebugBridgeDeviceReadException, TimeoutException, ExecutionException {
-        return getDeviceProp("ro.boot.wifimacaddr").toLowerCase();
+        var mac = runAdbShell("cat", "/sys/class/net/wlan0/address").replace("\n", "").replace("\r", "");
+        return mac;
```
jlaur, line 308:
```suggestion
        return runAdbShell("cat", "/sys/class/net/wlan0/address").replace("\n", "").replace("\r", "");
```

bundles/org.openhab.binding.androiddebugbridge/src/main/java/org/openhab/binding/androiddebugbridge/internal/AndroidDebugBridgeDevice.java
```diff
@@ -374,6 +380,225 @@ public void powerOffDevice()
         }
     }
 
+    public void startIntent(String command) throws NumberFormatException, AndroidDebugBridgeDeviceException,
+            ExecutionException, InterruptedException, TimeoutException {
+        String[] commandParts = command.split("\\|\\|");
+        String targetPackage = commandParts[0];
+        var targetPackageParts = targetPackage.split("/");
+        var packageName = targetPackageParts[0];
+        var activityName = targetPackageParts[1];
```
jlaur, line 389: Possible `ArrayIndexOutOfBoundsException` here?

bundles/org.openhab.binding.androiddebugbridge/src/main/java/org/openhab/binding/androiddebugbridge/internal/AndroidDebugBridgeDevice.java
```diff
@@ -374,6 +380,225 @@ public void powerOffDevice()
...
+                logger.warn("{} is not a valid package name", packageName);
+                return;
+            }
+            if (!SECURE_SHELL_INPUT_PATTERN.matcher(activityName).matches()) {
+                logger.warn("{} is not a valid activity name", activityName);
+                return;
+            }
+        } else {
+            throw new AndroidDebugBridgeDeviceException("Invalid target package " + targetPackage);
+        }
```
jlaur, line 406: Maybe this could be slightly rewritten for simplification and to avoid redundancy. Something like:

```
if (targetPackageParts.length > 2) {
    throw new AndroidDebugBridgeDeviceException("Invalid target package " + targetPackage);
}
if (!PACKAGE_NAME_PATTERN.matcher(packageName).matches()) {
    logger.warn("{} is not a valid package name", packageName);
    return;
}
if (targetPackageParts.length == 2 && !SECURE_SHELL_INPUT_PATTERN.matcher(activityName).matches()) {
    logger.warn("{} is not a valid activity name", activityName);
    return;
}
```

bundles/org.openhab.binding.androiddebugbridge/src/main/resources/OH-INF/thing/thing-types.xml
```diff
@@ -447,4 +448,10 @@
 		<state readOnly="true"/>
 	</channel-type>
 
+	<channel-type id="start-intent-channel" advanced="true">
+		<item-type>String</item-type>
+		<label>Start Intent</label>
+		<description>Start application intent</description>
+	</channel-type>
```
jlaur, line 455: New texts, so i18n strings needs to be added.

jlaur (changes requested):
Thanks for the PR. I have made a few minor comments.

bundles/org.openhab.binding.androiddebugbridge/src/main/java/org/openhab/binding/androiddebugbridge/internal/AndroidDebugBridgeDevice.java
GiviMAD: That's right, thank you!
Should be fixed with the new changes.

bundles/org.openhab.binding.androiddebugbridge/src/main/java/org/openhab/binding/androiddebugbridge/internal/AndroidDebugBridgeDevice.java
```diff
@@ -374,6 +379,220 @@ public void powerOffDevice()
         }
     }
 
+    public void startIntent(String command) throws NumberFormatException, AndroidDebugBridgeDeviceException,
```
lolodomo, line 382: Take care, the method `handleCommandInternal` could then throw NumberFormatException which is not caught by the method handleCommand

bundles/org.openhab.binding.androiddebugbridge/src/main/java/org/openhab/binding/androiddebugbridge/internal/AndroidDebugBridgeDevice.java
```diff
@@ -374,6 +379,220 @@ public void powerOffDevice()
...
+                    flags = argValue;
+                    break;
+                case "<e>":
+                case "<es>":
+                    valueParts = argValue.split(" ");
+                    if (!SECURE_SHELL_INPUT_PATTERN.matcher(valueParts[0]).matches()) {
+                        logger.warn("{}, insecure input value", valueParts[0]);
+                        return;
+                    }
+                    if (!SECURE_SHELL_INPUT_PATTERN.matcher(valueParts[1]).matches()) {
```
lolodomo, line 482: You should check that `valueParts` has at least two elements

bundles/org.openhab.binding.androiddebugbridge/src/main/java/org/openhab/binding/androiddebugbridge/internal/AndroidDebugBridgeDevice.java
```diff
@@ -374,6 +379,220 @@ public void powerOffDevice()
...
+                    }
+                    extraStrings.put(valueParts[0], valueParts[1]);
+                    break;
+                case "<ez>":
+                    valueParts = argValue.split(" ");
+                    if (!SECURE_SHELL_INPUT_PATTERN.matcher(valueParts[0]).matches()) {
+                        logger.warn("{}, insecure input value", valueParts[0]);
+                        return;
+                    }
+                    if (!SECURE_SHELL_INPUT_PATTERN.matcher(valueParts[1]).matches()) {
```
lolodomo, line 494: You should check that `valueParts` has at least two elements

bundles/org.openhab.binding.androiddebugbridge/src/main/java/org/openhab/binding/androiddebugbridge/internal/AndroidDebugBridgeDevice.java
```diff
@@ -374,6 +379,220 @@ public void powerOffDevice()
...
+                    }
+                    extraBooleans.put(valueParts[0], Boolean.parseBoolean(valueParts[1]));
+                    break;
+                case "<ei>":
+                    valueParts = argValue.split(" ");
+                    if (!SECURE_SHELL_INPUT_PATTERN.matcher(valueParts[0]).matches()) {
+                        logger.warn("{}, insecure input value", valueParts[0]);
+                        return;
+                    }
+                    if (!SECURE_SHELL_INPUT_PATTERN.matcher(valueParts[1]).matches()) {
```
lolodomo, line 506: You should check that `valueParts` has at least two elements

bundles/org.openhab.binding.androiddebugbridge/src/main/java/org/openhab/binding/androiddebugbridge/internal/AndroidDebugBridgeDevice.java
```diff
@@ -374,6 +379,220 @@ public void powerOffDevice()
...
+                    }
+                    extraIntegers.put(valueParts[0], Integer.parseInt(valueParts[1]));
+                    break;
+                case "<el>":
+                    valueParts = argValue.split(" ");
+                    if (!SECURE_SHELL_INPUT_PATTERN.matcher(valueParts[0]).matches()) {
+                        logger.warn("{}, insecure input value", valueParts[0]);
+                        return;
+                    }
+                    if (!SECURE_SHELL_INPUT_PATTERN.matcher(valueParts[1]).matches()) {
```
lolodomo, line 518: You should check that `valueParts` has at least two elements

bundles/org.openhab.binding.androiddebugbridge/src/main/java/org/openhab/binding/androiddebugbridge/internal/AndroidDebugBridgeDevice.java
```diff
@@ -374,6 +379,220 @@ public void powerOffDevice()
...
+                    }
+                    extraLongs.put(valueParts[0], Long.parseLong(valueParts[1]));
+                    break;
+                case "<ef>":
+                    valueParts = argValue.split(" ");
+                    if (!SECURE_SHELL_INPUT_PATTERN.matcher(valueParts[0]).matches()) {
+                        logger.warn("{}, insecure input value", valueParts[0]);
+                        return;
+                    }
+                    if (!SECURE_SHELL_INPUT_PATTERN.matcher(valueParts[1]).matches()) {
```
lolodomo, line 530: You should check that `valueParts` has at least two elements

lolodomo (changes requested):

lolodomo:
@GiviMAD : there is not too much to do so that I can merge your PR.

GiviMAD:
> @GiviMAD : there is not too much to do so that I can merge your PR.

Sorry, almost forgot about it, I was messing around trying other things. Thanks for pinging me, I'll go for it.

lolodomo (approved):
LGTM

lolodomo:
@jlaur : is it ok for you (as you reviewed this PR without approving it) ?

PR #12439, opened 2022-03-06
Fix links to SmartThings capabilities reference doc

glassl:


glassl:
I'm new to this repo and I'm not sure what this DCO check is about. I tried to fix it but I wasn't able. Sorry if I caused any problems.

wborn (approved):
Thanks! For these small fixes the DCO is not important. You can read about adding a sign off  [here](https://www.openhab.org/docs/developer/contributing.html#sign-your-work). If you add a small line to the commit message next time, the DCO check will succeed. :heavy_check_mark:

PR #12443, opened 2022-03-07
[ipp] Update cups4j and dependencies

wborn:
Updates cups4j to 0.7.8 and syncs the cups4j dependencies.


jlaur (approved):
Looks good, thanks! Verified that cups4j actually changed the `CupsPrinter` constructor in 0.7.7 and that credentials being null is handled.

PR #12450, opened 2022-03-08
[homekit] do unit conversion for min/max heating/cooling temps

ccutrer:
this is especially important if you use the same item for both
TargetTemperature and HeatingThresholdTemperature characteristics,
since the former was already doing unit conversion for min/max.


andylintner:
I remember feeling like something was off with temperature scales. Glad you figured it out @ccutrer !

yfre (approved):
LGTM

@ccutrer  thank you for fixing this!

lolodomo (approved):
LGTM

Issue #12451, opened 2022-03-09
[stt] Better default value for maxSilenceSeconds / inactivityTimeout setting


lolodomo:
The default value for maxSilenceSeconds setting is 5 seconds in the Google STT and Vosk STT.
For the Watson STT, the similar setting is named inactivityTimeout and its default value is 3 seconds.

5 seconds is largely too much in practice.
I would suggest to switch to 2 seconds, max 3 seconds.

Additionaly, it could be good to rename the setting in the Watson STT service to have the same setting name in all STT services.

@GiviMAD : WDYT ?

GiviMAD:
The reason behind this is that those services (Google STT and Vosk STT) support the single utterance mode which made that you don't need to wait the maxSilenceSeconds for the transcription to stop so having a large value there just allows you to wait in silence before start talking.

But yes, I think that unify the option names and default values for does services makes sense and also I will try to add support for the single utterance mode in Watson STT.

I will assign the issue to myself and check to check it in the afternoon.

lolodomo:
> The reason behind this is that those services (Google STT and Vosk STT) support the single utterance mode which made that you don't need to wait the maxSilenceSeconds for the transcription to stop so having a larger value there just allows you to wait in silence before start talking.

Ok. I am using the default settings so singleUtteranceMode is true and I noticed a very long delay before SST ended. I thought it was due to maxSilenceSeconds but I understand now that it is not. 
Do you notice this big delay too? 
Does it mean singleUtteranceMode is not a good mode?

lolodomo:
> And no I didn't face a long delay, how long are we talking?

I can measure it more precisely this evening but I would say around 5 seconds

GiviMAD:
> I can measure it more precisely this evening but I would say around 5 seconds

No, I'm not facing this problem tested with the latest openHAB snapshot and vosk. With both the javasound source and the pulse audio source (last version from the active pr).
I think is just the time that vosk takes to transcript the audio (which can be different based on the model and the computer), you should be able to check if this is the case by enabling the debug logs on the vosk add-on.

lolodomo:
Here are logs with timings:
```
22:45:19.567 [DEBUG] [.core.voice.internal.VoiceManagerImpl] - Executing a simple dialog for source Entrée audio Sony Eye (pulseaudio:source:rpi2:entree)
22:45:19.571 [DEBUG] [ternal.PulseaudioSimpleProtocolStream] - Simple TCP Stream connecting
[...]

PR #12453, opened 2022-03-10
STT service improvements

GiviMAD:
googlestt/voskstt: change default maxSilenceSeconds to 3
watsonstt: add singleUtterance mode, rename inativityTimeout to maxSilenceSeconds and minor improvements


GiviMAD:
Proposed solution to https://github.com/openhab/openhab-addons/issues/12451.

@lolodomo the build failure is not related to the changes.

lolodomo:
Finally, when not in single utterance mode  is 3 seconds enough? WDYT?

bundles/org.openhab.voice.watsonstt/src/main/resources/OH-INF/i18n/watsonstt_it.properties
```diff
@@ -6,8 +6,8 @@ voice.config.watsonstt.group.authentication.label = Autenticazione
 voice.config.watsonstt.group.authentication.description = Informazioni per la connessione all'istanza Watson Speech-to-Text.
 voice.config.watsonstt.group.stt.label = Configurazione STT
 voice.config.watsonstt.group.stt.description = Parametri per l'API Watson Speech-to-Text.
-voice.config.watsonstt.inactivityTimeout.label = Timeout Inattività
-voice.config.watsonstt.inactivityTimeout.description = Il tempo in secondi dopo il quale, se viene rilevato solo silenzio (nessun discorso) nell'audio, la connessione viene chiusa.
+voice.config.watsonstt.maxSilenceSeconds.label = Timeout Inattività
```
lolodomo, line 9: To be removed. 
To be done later in Crowdin.

lolodomo (changes requested):

GiviMAD:
> Finally, when not in single utterance mode is 3 seconds enough? WDYT?

Yes I think it's enough, it's the value that I was using. 
Also as the singleUtteranceMode is enabled by default I think the user will tune it as desired when disabling the singleUtteranceMode.

bundles/org.openhab.voice.watsonstt/src/main/resources/OH-INF/i18n/watsonstt_it.properties
GiviMAD: So, should I avoid modify property files other than the English one?

bundles/org.openhab.voice.watsonstt/src/main/java/org/openhab/voice/watsonstt/internal/WatsonSTTService.java
```diff
@@ -130,16 +132,16 @@ public STTServiceHandle recognize(STTListener sttListener, AudioStream audioStre
...
                 .build();
         final AtomicReference<@Nullable WebSocket> socketRef = new AtomicReference<>();
         final AtomicBoolean aborted = new AtomicBoolean(false);
         executor.submit(() -> {
             int retries = 2;
             while (retries > 0) {
                 try {
-                    socketRef.set(speechToText.recognizeUsingWebSocket(wsOptions,
-                            new TranscriptionListener(sttListener, config, aborted)));
+                    int internalRetries = 0;
```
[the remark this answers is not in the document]
GiviMAD: > This looks to me as a possible workaround to a wrong implementation, if required. WDYT?

Yes, it's a workaround for a problem that I'm facing with this SDK from the beginning. 
After the first execution, a SSLPeerUnverifiedException is thrown. The cause seems to be that the underling library used for that (bouncycastle I think) seems to be unloaded after the first execution.
I'm been unable to identify what is causing the issue.

I saw there is some available updates in the dependencies, I will check if that changes something.

bundles/org.openhab.voice.watsonstt/src/main/resources/OH-INF/i18n/watsonstt_it.properties
lolodomo: Yes

bundles/org.openhab.voice.watsonstt/src/main/java/org/openhab/voice/watsonstt/internal/WatsonSTTService.java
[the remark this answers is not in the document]
GiviMAD: After a little nightmare I discovered the problem, a bug on the java version I was using with TLSv3, described (here)[https://stackoverflow.com/questions/55161904/javax-net-ssl-sslhandshakeexception-from-ibms-watson-assistant-api].
I updated to the last java 11 version available and the problem is solved, so I will remove both retry mechanisms.

lolodomo (approved):
LGTM

PR #12461, opened 2022-03-11
[wemo] Prevent excessive currentPower channel updates

### LINKED ISSUE #12460: [wemo] Excessive channel updates of currentPower for WeMo Insight Switch
For my WeMo Insight Switches, this is a quite common scenario:
- 2022-03-11 23:53:58.859: 23 W
- 2022-03-11 23:54:06.902: 24 W
- 2022-03-11 23:54:15.952: 23 W
- 2022-03-11 23:54:24.005: 24 W
- 2022-03-11 23:54:32.025: 23 W
- 2022-03-11 23:54:37.557: 24 W
- 2022-03-11 23:54:53.677: 23 W
- 2022-03-11 23:55:01.761: 24 W

In this period of ~a minute these values are actually received from the device:
- 2022-03-11 23:53:58.845: 23.27 W
- 2022-03-11 23:54:00.355: 23.24 W
- 2022-03-11 23:54:02.867: 23.26 W
- 2022-03-11 23:54:04.375: 23.275 W
- 2022-03-11 23:54:06.885: 23.765 W
- 2022-03-11 23:54:08.395: 23.98 W
- 2022-03-11 23:54:10.904: 24.13 W
- 2022-03-11 23:54:12.414: 24.16 W
- 2022-03-11 23:54:15.937: 23.415 W
- 2022-03-11 23:54:17.440: 23.38 W
- 2022-03-11 23:54:19.952: 23.275 W
- 2022-03-11 23:54:21.457: 23.26 W
- 2022-03-11 23:54:23.972: 23.615 W
- 2022-03-11 23:54:25.478: 23.64 W
- 2022-03-11 23:54:27.993: 23.645 W
- 2022-03-11 23:54:29.499: 23.655 W
- 2022-03-11 23:54:32.044: 23.405 W
- 2022-03-11 23:54:33.520: 23.31 W
- 2022-03-11 23:54:36.035: 23.495 W
- 2022-03-11 23:54:37.543: 23.565 W
- 2022-03-11 23:54:40.055, 23.64 W
- 2022-03-11 23:54:41.562: 23.685 W
- 2022-03-11 23:54:44.076: 23.805 W
- 2022-03-11 23:54:45.585: 23.775 W
- 2022-03-11 23:54:48.099: 23.81 W
- 2022-03-11 23:54:49.601: 23.765 W
- 2022-03-11 23:54:53.662: 23.37 W
- 2022-03-11 23:54:57.697: 23.38 W
- 2022-03-11 23:54:59.191: 23.42 W
- 2022-03-11 23:55:01.710: 24.05 W

When calculating the average current power consumption for this sample period, it will actually be closer to 23 W than 24 W, so we can safely round down.

Introducing an algorithm for calculating a moving average in a sliding window would prevent such excessive channel updates which is causing a lot of logging and data growth when using persistence.

jlaur:
Fixes #12460

Please see linked issue for description of the problem being solved with this pull request.

### Example
This example shows the previous behavior compared to new behavior with actual data from a recuperator unit connected to the WeMo Insight Switch.

#### Summary
Reduction in number of updates during a 24 hour period:
- Actual updates from device with full precision: **41.915**
- Previous simple rounding: **4.638**
- With new algorithm applied (default parameters): **314**

#### Data

Raw data:
- 2022-03-11 23:53:58.845: 23.27 W
- 2022-03-11 23:54:00.355: 23.24 W
- 2022-03-11 23:54:02.867: 23.26 W
- 2022-03-11 23:54:04.375: 23.275 W
- 2022-03-11 23:54:06.885: 23.765 W
- 2022-03-11 23:54:08.395: 23.98 W
- 2022-03-11 23:54:10.904: 24.13 W
- 2022-03-11 23:54:12.414: 24.16 W
- 2022-03-11 23:54:15.937: 23.415 W
- 2022-03-11 23:54:17.440: 23.38 W
- 2022-03-11 23:54:19.952: 23.275 W
- 2022-03-11 23:54:21.457: 23.26 W
- 2022-03-11 23:54:23.972: 23.615 W
- 2022-03-11 23:54:25.478: 23.64 W

Previous behavior:
- 2022-03-11 23:53:58.859: 23 W
- 2022-03-11 23:54:06.902: 24 W
- 2022-03-11 23:54:15.952: 23 W
- 2022-03-11 23:54:24.005: 24 W

New behavior (full data not included, but moving average for the example is less than 23.5 W):
- 2022-03-11 23:53:58.859: 23 W


jlaur:
JAR for testing: https://drive.google.com/file/d/13az_CkBXNenTpwdU-taXph2d1PicGYI-/view?usp=sharing

bundles/org.openhab.binding.wemo/README.md
```diff
@@ -52,6 +76,7 @@ Devices support some of the following channels:
 | timespan            | Number        | Time in seconds over which onTotal applies. Typically 2 weeks except first used.                                           | Insight                                              |
 | averagePower        | Number:Power  | Average power consumption in Watts.                                                                                        | Insight                                              |
 | currentPower        | Number:Power  | Current power consumption of an Insight device. 0 if switched off.                                                         | Insight                                              |
+| currentPowerRaw     | Number:Power  | Current power consumption of an Insight device with full precision (5 mW accuracy, three decimals). 0 if switched off.     | Insight                                              |
 | energyToday         | Number:Energy | Energy in Wh used today.                                                                                                   | Insight                                              |
```
hmerk, line 80: You are introducing a breaking change. I would rather suggest to add e new channel with the reduced precision.

hmerk (changes requested):
Thanks @jlaur for your contribution.
In general, the idea behind looks good, even your implementation.
But to not break existing installations. I would rather introduce a new channel with reduced accuracy instead of changing the default behaviour.

bundles/org.openhab.binding.wemo/README.md
jlaur: > You are introducing a breaking change. I would rather suggest to add e new channel with the reduced precision.

Actually, I'm not. :-) Previously the accuracy was 1 W, i.e. rounding to nearest integer, and it still is. I tried to balance the default configuration parameters to not optimize too aggressively, so most users shouldn't notice anything, except fewer updates.

To be fully backwards compatible in terms of behavior, the sliding window could be defaulted to 0 to disable it. However, in my opinion it makes more sense to fix the problem with sensible default parameters (one minute window and 1 W forced update) and leave it to the users to adjust. I don't think any users would prefer the old behavior.

jlaur: Just to be clear: The new **currentPowerRaw** channel with the full provided precision of 5 mW accuracy is not the same as the previous **currentPower** - it never had any decimals.

hmerk: OK, convinced

hmerk (approved):
LGTM

jlaur:
@openhab/add-ons-maintainers - any chance of a review before next milestone? This PR **might** fix unstable integration test.

wborn:
I wonder if it would also make sense to implement such an algorithm in a profile so you can use it with any binding and don't need to create additional channels and duplicate this kind of logic?

jlaur:
> I wonder if it would also make sense to implement such an algorithm in a profile so you can use it with any binding and don't need to create additional channels and duplicate this kind of logic?

I completely agree with you. I had the same thought, but didn't know how to proceed with this. I don't think it's currently possible with any existing profiles and/or transformations? Implementing a new profile might be the way to go. How are new profiles provided - in core or addons? This might be something for a new project. :) I can see that transformations are in addons, but not sure about profiles even after reading https://www.openhab.org/docs/developer/transformations/.

In any case, if I (or someone else) would manage to provide a profile for this functionality, the functionality in this binding would of course be obsoleted, and I would remove it again. As short-term solution I'm wondering what you would think about accepting the algorithm provided in this PR to fix the issue at hand? If/when a profile would be available, probably it would make most sense to remove channel **currentPowerRaw** again, and provide full precision in **currentPower** as it would then be possible to provide rounding as well as moving averages in the profile.

wborn:
For the rounding you can already use the JavaScript profile/transformation or probably any other available scripting language when https://github.com/openhab/openhab-core/pull/2872 is merged.

For the moving average it would have to maintain state for calculating the average. Transformations are stateless but for profiles you can keep track of state in a profile, see [Developing a profile](https://www.openhab.org/docs/developer/transformations/#developing-a-profile):

> Profiles are created by ProfileFactories and are retained for the lifetime of their link. This means that they are, in contrast to transformations, allowed to retain a transient state, like e.g. the timestamp of the the last event or the last state. With this, it is possible to take into account the temporal dimension when calculating the appropriate action in any situation.

I haven't implemented any profiles myself so far. So maybe @kaikreuzer and @cweitkamp can tell if would be a good idea to implement transformations/profiles for rounding and moving averages in the core?

bundles/org.openhab.binding.wemo/README.md
```diff
@@ -18,23 +18,47 @@ The binding does not need any configuration.
...
 
-For a WeMo Link bridge and paired LED Lights, please use the following Thing definition
+| Configuration Parameter | Description                        |
+|-------------------------|------------------------------------|
+| udn                     | The UDN identifies the WeMo device |
 
-```
-Bridge wemo:bridge:Bridge-1_0-231445B01006A0 [udn="Bridge-1_0-231445B010xxxx"] {
-MZ100 94103EA2B278xxxx [ deviceID="94103EA2B278xxxx" ]
-MZ100 94103EA2B278xxxx [ deviceID="94103EA2B278xxxx" ]
```
lolodomo, line 33: Please keep the example with `deviceID` below in the documentation.

bundles/org.openhab.binding.wemo/README.md
```diff
@@ -18,23 +18,47 @@ The binding does not need any configuration.
...
+For LED Lights paired to a WeMo Link bridge, please use the following configuration parameter:
+
+| Configuration Parameter | Description                                     |
+|-------------------------|-------------------------------------------------|
+| deviceID                | The device ID identifies one certain WeMo light |
+
+### WeMo Insight Switch
 
+The WeMo Insight Switch has some additional parameters for controlling the behavior for channel `currentPower`.
+This channel reports the current power consumption in Watt. The internal theoretical accuracy is 5 mW, i.e.
```
lolodomo, line 41: Please cut, one sentence per line is the recommendation.
Same remark for all the file (and all your changes).

bundles/org.openhab.binding.wemo/README.md
```diff
@@ -94,14 +119,14 @@ Devices support some of the following channels:
...
 
-
 ## Full Example
 
 demo.things:
 
 ```
 wemo:socket:Switch1     "DemoSwitch"   @ "Office"   [udn="Socket-1_0-221242K11xxxxx"]
 wemo:motion:Sensor1     "MotionSensor" @ "Entrance" [udn="Sensor-1_0-221337L11xxxxx"]
+wemo:insight:Insight1   "Insight"      @ "Attic"    [udn="Insight-1_0-xxxxxxxxxxxxxx", currentPowerSlidingSeconds="120", currentPowerDeltaTrigger=2]
```
lolodomo, line 129: currentPowerSlidingSeconds=120

bundles/org.openhab.binding.wemo/src/main/resources/OH-INF/config/insight.xml
```diff
@@ -0,0 +1,29 @@
...
+
+	<config-description uri="thing-type:wemo:insight">
+		<parameter name="udn" type="text" required="true">
+			<label>Unique Device Name</label>
+			<description>The UDN identifies the WeMo Insight Switch</description>
+		</parameter>
+		<parameter name="currentPowerSlidingSeconds" type="integer" min="0" unit="s">
+			<label>Current Power sliding window</label>
+			<description>Sliding window in seconds for which moving average power is calculated (0 = disabled)</description>
+			<unitLabel>seconds</unitLabel>
```
lolodomo, line 16: Just for my information: what is this tag compared to "unit" attribute ?

bundles/org.openhab.binding.wemo/src/main/java/org/openhab/binding/wemo/internal/WemoBindingConstants.java
```diff
@@ -57,7 +57,8 @@
     public static final String CHANNEL_ONTOTAL = "onTotal";
     public static final String CHANNEL_TIMESPAN = "timespan";
     public static final String CHANNEL_AVERAGEPOWER = "averagePower";
-    public static final String CHANNEL_CURRENTPOWER = "currentPower";
+    public static final String CHANNEL_CURRENT_POWER = "currentPower";
+    public static final String CHANNEL_CURRENT_POWER_RAW = "currentPowerRaw";
```
lolodomo, line 61: IMHO, this is better with _ for readability but as all the other constants are without `_`, maybe it would be better to keep something homogeneous ?

bundles/org.openhab.binding.wemo/src/main/java/org/openhab/binding/wemo/internal/WemoPowerBank.java
```diff
@@ -0,0 +1,123 @@
...
+    private class CacheItem {
+
+        public CacheItem(double power, Instant start) {
+            this.start = start;
+            this.power = power;
+        }
+
+        public Instant start;
+        public @Nullable Instant end;
+        public double power;
```
lolodomo, line 48: Members are generally more expected before the constructor.

bundles/org.openhab.binding.wemo/src/main/java/org/openhab/binding/wemo/internal/config/WemoInsightConfiguration.java
```diff
@@ -0,0 +1,33 @@
...
+@NonNullByDefault
+public class WemoInsightConfiguration {
+
+    public static final String CURRENT_POWER_SLIDING_SECONDS = "currentPowerSlidingSeconds";
+    public static final String CURRENT_POWER_DELTA_TRIGGER = "currentPowerDeltaTrigger";
+
+    @Nullable
+    public String udn;
+    public int currentPowerSlidingSeconds;
+    public int currentPowerDeltaTrigger;
```
lolodomo, line 32: A good practice is to set as default value the default value that is defined in the XML definition file.

bundles/org.openhab.binding.wemo/src/main/java/org/openhab/binding/wemo/internal/handler/WemoHandler.java
```diff
@@ -62,21 +61,13 @@ public WemoHandler(Thing thing, UpnpIOService upnpIOService, WemoHttpCall wemoHt
     @Override
     public void initialize() {
         super.initialize();
-        Configuration configuration = getConfig();
 
-        if (configuration.get(UDN) != null) {
```
lolodomo, line 67: udn parameter can now never be null when initialize() is called ?

lolodomo (changes requested):

bundles/org.openhab.binding.wemo/README.md
jlaur: It's already in the "Full Example" chapter, it was redundant here. I changed it so only description is given here, and examples are provided only in that chapter.

bundles/org.openhab.binding.wemo/README.md
jlaur: Done for whole file.

bundles/org.openhab.binding.wemo/README.md
jlaur: Good catch, fixed.

bundles/org.openhab.binding.wemo/src/main/resources/OH-INF/config/insight.xml
jlaur: Please see https://www.openhab.org/docs/developer/bindings/config-xml.html#xml-structure-for-configuration-descriptions.

bundles/org.openhab.binding.wemo/src/main/java/org/openhab/binding/wemo/internal/config/WemoInsightConfiguration.java
jlaur: Done.

bundles/org.openhab.binding.wemo/src/main/java/org/openhab/binding/wemo/internal/WemoPowerBank.java
jlaur: Of course - fixed.

bundles/org.openhab.binding.wemo/src/main/java/org/openhab/binding/wemo/internal/WemoBindingConstants.java
jlaur: All constants renamed for readability and consistency.

bundles/org.openhab.binding.wemo/src/main/java/org/openhab/binding/wemo/internal/handler/WemoHandler.java
jlaur: No, this check is redundant in 3.3 as the parameter is defined as _required_ and we now have proper configuration validation: openhab/openhab-core#2691.

bundles/org.openhab.binding.wemo/src/main/resources/OH-INF/config/insight.xml
lolodomo: Makes no sense IMHO to hardcode an English unit label. The label should be automatically derived from the unit attribute.

bundles/org.openhab.binding.wemo/README.md
lolodomo: OK

lolodomo (approved):
LGTM

bundles/org.openhab.binding.wemo/src/main/resources/OH-INF/config/insight.xml
lolodomo: Is this label considered by MainUI ?
Is it translated (i18n) through an entry in the i18n properties ?

lolodomo: I found this tag used 86 times.

jlaur: It might be obsoleted, in which case we need to update the documentation and remove it from the codebase. I didn't check it yet.

PR #12463, opened 2022-03-12
[fronius] Add a note about enabling Night Mode on the device

jimtng:
See: https://community.openhab.org/t/fronius-symo-goes-offline/134119/7


jlaur (approved):
Thanks!

PR #12469, opened 2022-03-13
Add mielecloud.tests to CODEOWNERS file

wborn:
The mielecloud.tests dir is missing from the CODEOWNERS file.


jlaur (approved):
Correct, thanks.

PR #12471, opened 2022-03-13
[mail] Fix unstable MailBuilderTest

wborn:
When a URL attachment is used a check is done if an InputStream to the URL can be opened to see if it is valid.
Because the openHAB website was used the test would fail when the website has issues:

```
[ERROR]  org.openhab.binding.mail.MailBuilderTest.withURLAttachmentReturnsMultiPartEmail  Time elapsed: 10.391 s  <<< ERROR!
org.apache.commons.mail.EmailException: Invalid URL set:http://www.openhab.org
	at org.openhab.binding.mail.MailBuilderTest.withURLAttachmentReturnsMultiPartEmail(MailBuilderTest.java:77)
Caused by: java.io.IOException: Server returned HTTP response code: 520 for URL: http://www.openhab.org
	at org.openhab.binding.mail.MailBuilderTest.withURLAttachmentReturnsMultiPartEmail(MailBuilderTest.java:77)
```

It will now use a file URL instead so it no longer depends on a website.

--- 

See: https://github.com/openhab/openhab-addons/runs/5528459800

### DIFF
--- a/bundles/org.openhab.binding.mail/src/test/java/org/openhab/binding/mail/MailBuilderTest.java
+++ b/bundles/org.openhab.binding.mail/src/test/java/org/openhab/binding/mail/MailBuilderTest.java
@@ -18,6 +18,7 @@ import static org.junit.jupiter.api.Assertions.*;
 
 import java.io.IOException;
 import java.net.MalformedURLException;
+import java.nio.file.Path;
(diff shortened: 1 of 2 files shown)


jlaur (approved):
Great finding, thanks!

PR #12472, opened 2022-03-13
[mqtt] Remove MQTT system broker leftovers

wborn:
The MqttService was used to keep track of the system MQTT brokers.
Since there are none the dependency can be removed.

Related to: openhab/openhab-core#2842
See also: #12157


jlaur (approved):
LGTM

PR #12482, opened 2022-03-15
[icalendar] Adding ability to handle events without DTEND

clinique:
According to section 3.6.1 of RFC 5545 :

```For cases where a "VEVENT" calendar component specifies a "DTSTART" property with a DATE value type but no "DTEND" nor "DURATION" property, the event's duration is taken to be one day. For cases where a "VEVENT" calendar component specifies a "DTSTART" property with a DATE-TIME value type but no "DTEND" property, the event ends on the same calendar date and time of day specified by the "DTSTART" property.```

Currently icalendar binding just ignores calendar entries if they have no DTEND defined.


lolodomo (approved):
LGTM

lolodomo:
@daMihe : is it ok for you?

daMihe:
I have tomorrow time to review.

daMihe (approved):
Great :)

PR #12487, opened 2022-03-18
[renault] API key update.

dougculnane:
Signed-off-by: Doug Culnane <doug@culnane.net>

I and 2 other users have had connection errors. See: (https://community.openhab.org/t/betatest-renault-ze-services-binding/72226/190?u=doug_culnane)

The API key has changed which it does every year apparently... So I used the new key that the python developers have extracted:
https://github.com/hacf-fr/renault-api/commit/03f4dbd9f27dc84aa748510892fbb3f55eed0f1d

This worked on my installation.

I will submit a pull request with this as a configuration variable with the default being the current key.  This will allow easier updating of running systems in the future.


lolodomo (approved):
LGTM

dougculnane:
@lolodomo  Thanks.

PR #12489, opened 2022-03-18
[openwebnet] fixed generic_device thing-type, improved refresh at boot

mvalla:
Fixed `generic_device` thing-type (used when a BTicino device cannot be recognized) and added a time limit to devices refresh at boot/reconnect.

### DIFF
--- a/bundles/org.openhab.binding.openwebnet/README.md
+++ b/bundles/org.openhab.binding.openwebnet/README.md
@@ -231,7 +231,8 @@ See [openwebnet.sitemap](#openwebnet-sitemap) & [openwebnet.rules](#openwebnet-r
 
 There are three WEEKLY and sixteen SCENARIO programs defined for the Central Unit.
 
-In order to activate one of them you have to use two different channels: 
+In order to activate one of them you have to use two different channels:
+
 - with `mode` you can set the mode (`WEEKLY` or `SCENARIO`)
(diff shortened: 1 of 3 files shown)


lolodomo:
Changing a thing type is a breaking change for any user that already defined these things. Do you really want to change that? In this case, a specific notr in the 3.3 release notes would be expected.

mvalla:
> Changing a thing type is a breaking change for any user that already defined these things.

actually this is a thing-type never used practically: it’s just used to highlight not recognized devices during discovery (where the right name “generic_thing” was already set programmatically when creating the inbox entry). 

Formally is a breaking change but practically no one should have never used (and will never use) this thing type as it’s useless from manual configuration perspective. 

Should one user have used it in file configuration, it would not work anyway as current name is not aligned with the rest of the code, so practically it’s a bug.

However if you prefer I can put a breaking change Notice.

lolodomo (approved):
LGTM

lolodomo:
> However if you prefer I can put a breaking change Notice.

Not necessary if you think it adds no value for the final user.

lolodomo:
Before I merge, I read that in your description:

> Removed a line from README.

What I see is no line removed but a blank line added. 
I am not sure it is necessary to mention it.

mvalla:
ok so no breaking change notice (should the tag be removed then?). I removed the (wrong) mention to the README. Thanks!

lolodomo:
> ok so no breaking change notice (should the label be removed then?)

I think we can keep the label. As you said: "Formally is a breaking change".

Issue #12458, opened 2022-03-10
[somfytahoma] Detected a new unsupported device: Shutter with widgetName: DynamicShutter


ElectricBlue66:
Hi, I'm not too sure whether this is the right place to post this. But I'm sure someone can move it if it is wrong.

I recently installed my Somfy Tahoma Connectivity Kit and set everything up. In OH3 I added the SomfyTahoma binding and it automatically discovered my Somfy awning. However, OH3 does not automatically discover my roller shutters. I noticed that the log file reports the following:

`==> /var/log/openhab/openhab.log <== 2022-03-10 17:43:32.680 [INFO ] [very.SomfyTahomaItemDiscoveryService] - Detected a new unsupported device: Shutter with widgetName: DynamicShutter 2022-03-10 17:43:32.695 [INFO ] [very.SomfyTahomaItemDiscoveryService] - If you want to add the support, please create a new issue and attach the information below 2022-03-10 17:43:32.697 [INFO ] [very.SomfyTahomaItemDiscoveryService] - Device definition: Commands: { close (params: 0); goToAlias (params: 1); identify (params: 0); open (params: 0); saveAlias (params: 1); setName (params: 1); setNodeDefinition (params: 3); stop (params: 0); } States: {core:AvailabilityState (values: available, unavailable); core:NameState (values: ); core:RemovableState (values: ); } 2022-03-10 17:43:32.700 [INFO ] [very.SomfyTahomaItemDiscoveryService] - Current device states: {name='core:AvailabilityState', type=3, value=available}`

This is reported 7 times, which matches the number of roller shutters. Looks like they are not supported. I tried to add them as things, but I don't know the Somfy Tahoma Device URL.

Can I find the URL somewhere? Or are the roller shutters really not supported?

Thanks for any help.

Rob

octa22:
Hi,

I have prepared a version, which adds the support. Can you test it please?
https://www.dropbox.com/s/4i37burwfkhc2zl/org.openhab.binding.somfytahoma-3.3.0-SNAPSHOT.jar?dl=1
Thanks.
Ondrej

ElectricBlue66:
Hey, that sounds great. I'm in Berlin for the weekend. Will test next week.

Rob

Holen Sie sich Outlook für Android<https://aka.ms/AAb9ysg>

________________________________

ElectricBlue66:
Hi Ondrej,

[...]

PR #12495, opened 2022-03-19
[somfytahoma] added support for the Shutter thing

octa22:
This PR adds the support for the Shutter thing as requested here:
https://github.com/openhab/openhab-addons/issues/12458
and more over it fixes the functionality for simple dynamic venetian blinds which is discussed here:
https://community.openhab.org/t/re-somfy-tahoma-binding-not-working-properly-with-new-somfy-connectivity-kit/133969/14

Signed-off-by: Ondrej Pecta <opecta@gmail.com>


bundles/org.openhab.binding.somfytahoma/src/main/java/org/openhab/binding/somfytahoma/internal/handler/SomfyTahomaShutterHandler.java
```diff
@@ -0,0 +1,74 @@
...
+            case "CLOSE":
+                return COMMAND_CLOSE;
+            case "ON":
+            case "UP":
+            case "OPEN":
+                return COMMAND_OPEN;
+            case "MOVE":
+            case "MY":
+            case "STOP":
+                return COMMAND_STOP;
```
lolodomo, line 69: Sending the same command (CLOSE) for MOVE and STOP looks a little strange, no ?
What is this OH command "MY" ? From which type does it come from ?

octa22: You are right, not necessary here - removed MOVE and MY.

bundles/org.openhab.binding.somfytahoma/src/main/java/org/openhab/binding/somfytahoma/internal/handler/SomfyTahomaShutterHandler.java
```diff
@@ -0,0 +1,72 @@
...
+            case "CLOSE":
+                return COMMAND_CLOSE;
+            case "ON":
+            case "UP":
+            case "OPEN":
+                return COMMAND_OPEN;
+            case "STOP":
+                return COMMAND_STOP;
+            default:
+                return command.toLowerCase();
```
lolodomo, line 69: That means you will send "move" for example. 
If only open close and stop are valid commands, this is a problem, no?

octa22: there is no right behaviour for the MOVE command. So I changed it to issue a STOP command.

lolodomo: Why not simply ignore the OH commands which are not supported by the device? 
It would make more sense than sending a supported command but an inappropriate command?

octa22: OK, I have done it as you suggested - the unsupported/not recognized commands are just ignored.

lolodomo (approved):
LGTM, thank you

Issue #12490, opened 2022-03-18
[openwebnet] Thermo things are refreshed twice at boot/reconnect


mvalla:
Thermo things are refreshed 2 times at boot/reconnect, first from bridge then from Thermo handler (in bridgeStatusChanged()). 
This causes unnecessary messages at boot/reconnect.

## Possible Solution
Remove method bridgeStatusChanged() from Thermo handler class

lolodomo:
Closed by #12502

PR #12502, opened 2022-03-21
[openwebnet] removed (duplicate) refreshDevice() at bridgeStatusChanged

aconte80:
Thermo things are refreshed 2 times at boot/reconnect, first from bridge then from Thermo handler (in bridgeStatusChanged()).

See issue https://github.com/openhab/openhab-addons/issues/12490


lolodomo (approved):
LGTM

PR #12509, opened 2022-03-22
[plugwiseha] Fix Zone thermostat not updating

lsiepel:
Previous PR reduced http request rate/load by enabling incremental updates from the PlugwiseHA API. This PR fixes a situation that was overlooked in that PR. This bug made it impossible to set the thermostat for a zone.


lolodomo (approved):
LGTM

PR #12510, opened 2022-03-22
[miele] Fix configuration validation and optimize performance

### LINKED ISSUE #12504: [miele] Thing status handling and other issues
While working on #12497 I saw some issues with the way status transitions are handled in the Miele@home binding:
- Theoretical timing issue: Bridge status is set to **UNKNOWN** right after scheduling polling job, which could set status **ONLINE** at the same time. However, since a network call is made before setting this status, it is unlikely to happen.
- Configuration errors are not correctly detected/handled by the appliance handler: Missing bridge or UID should be reflected as status **OFFLINE** with **CONFIGURATION_ERROR**.
- Appliances configured with an invalid (unknown) UID are marked as **ONLINE** and a warning will be logged every 15 seconds. Should be marked as **OFFLINE** with **GONE**.
- Bridge handler type is checked unneededly.
- Each thing registration in bridge handler will cause full list of devices to be fetched.
- Each thing registration will attempt to schedule polling job and multicast thread. It will only happen once, but a lot of redundant logging during initialization.
- Each thing will be notified about any registration (i.e. all other things) and filters irrelevant information.
- `getMieleBridgeHandler()` subscribes to events by calling `registerApplianceStatusListener()`. And it's called in `dispose()`, after which `unregisterApplianceStatusListener()` is called.
- Some static properties like **deviceType** and **productType** are updated every 15 seconds.

Additional refactoring ideas:
- JSON-RPC methods could be extracted from `MieleBridgeHandler`.
- Appliance status listeners are kept in a list so could risk duplication.
- Every 15th second the polling job will call a dummy method to check if gateway is reachable. This could easily be deducted from real calls instead of this additional call.
- Some logging with **error** level should be reduced to **warning**.
- `MieleTranslationProvider` could be initialized with default language (provided by openHAB) in constructor to avoid being `@Nullable`.

jlaur:
This pull request contains various fixes and optimizations for the Miele@home binding. These issues were discovered by analyzing the code, backed up by manual testing as proof. See also #12504 for findings. Please read this PR commit by commit for better understanding.

### Bugfixes
- Appliance configuration is now validated: Missing or empty UID will now result in status **OFFLINE** with **CONFIGURATION_ERROR**.
- Appliances removed from bridge (i.e. non-existent appliances) are now marked as **OFFLINE** with **GONE**. Previously they were marked as **ONLINE**.

### Optimizations
- An unneeded call to the Miele gateway every 15 seconds has been removed. This call received a payload of almost 6 KB.
- Appliance handlers are no longer notified about updates for other appliances. Everything was broadcasted and each handler would filter.
- Listeners will receive immediate callback when appliances are already cached.
- Listener registrations will no longer cause all appliances to be fetched from gateway for each registration.
- Jobs are no longer checked/scheduled upon each listener registration.

### Refactoring
- JSON-RPC handling has been extracted to a separate class.
- Listener interface has been divided into a distinct interface for discovery and another one for appliance status.
- Online/offline status transitions has been simplified.
- Null annotation improvement for translation provider.
- Log levels have been adjusted for some messages.
- Access levels for some variables were reduced to prevent being exposed to subclasses for no reason.
- Many other simplifications.

# Testing

Besides general regression testing with appliances running programs, the following specific scenarios have been addressed:
- Online/offline transitions have been tested by disabling/enabling a port on the switch for the XGW 3000 gateway.
- Configuration of thing with missing/empty UID resulting in status **OFFLINE** with **CONFIGURATION_ERROR**: "Appliance ID is not set"
- Configuration of non-existent UID resulting in status **OFFLINE** with **GONE**.

### DIFF
--- a/bundles/org.openhab.binding.miele/src/main/java/org/openhab/binding/miele/internal/handler/DiscoveryListener.java
+++ b/bundles/org.openhab.binding.miele/src/main/java/org/openhab/binding/miele/internal/handler/DiscoveryListener.java
@@ -1,0 +1,39 @@
+/**
+ * Copyright (c) 2010-2022 Contributors to the openHAB project
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package org.openhab.binding.miele.internal.handler;
+
+import org.eclipse.jdt.annotation.NonNullByDefault;
+import org.openhab.binding.miele.internal.api.dto.HomeDevice;
+
+/**
+ * The {@link DiscoveryListener} is notified when any appliance has been removed or added.
+ *
+ * @author Jacob Laursen - Initial contribution
+ */
+@NonNullByDefault
+public interface DiscoveryListener {
+
+    /**
+     * This method is called whenever any appliance is removed.
+     *
+     * @param appliance The XGW homedevice definition of the appliance that was removed
+     */
+    void onApplianceRemoved(HomeDevice appliance);
+
+    /**
+     * This method is called whenever any appliance is added.
+     *
+     * @param appliance The XGW homedevice definition of the appliance that was removed
+     */
+    void onApplianceAdded(HomeDevice appliance);
+}
--- a/bundles/org.openhab.binding.miele/src/main/java/org/openhab/binding/miele/internal/handler/DiscoveryListener.java
+++ b/bundles/org.openhab.binding.miele/src/main/java/org/openhab/binding/miele/internal/handler/DiscoveryListener.java
@@ -33,7 +33,7 @@ public interface DiscoveryListener {
     /**
      * This method is called whenever any appliance is added.
      *
-     * @param appliance The XGW homedevice definition of the appliance that was removed
+     * @param appliance The XGW homedevice definition of the appliance that was added
      */
     void onApplianceAdded(HomeDevice appliance);
 }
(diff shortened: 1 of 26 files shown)


bundles/org.openhab.binding.miele/src/main/java/org/openhab/binding/miele/internal/handler/MieleBridgeHandler.java
```diff
@@ -276,50 +213,85 @@ public void run() {
...
+        if (!isInitialized()) {
+            return devices;
+        }
+
+        String[] args = new String[1];
+        args[0] = "(type=SuperVision)";
+        JsonElement result = gatewayCommunication.invokeRPC("HDAccess/getHomeDevices", args);
+
+        for (JsonElement obj : result.getAsJsonArray()) {
+            HomeDevice hd = gson.fromJson(obj, HomeDevice.class);
```
lolodomo, line 292: No need to catch exception ?

lolodomo (changes requested):

bundles/org.openhab.binding.miele/src/main/java/org/openhab/binding/miele/internal/handler/MieleBridgeHandler.java
jlaur: It's a runtime exception, so technically it is not required to catch it. It should only occur either in case there's a bug in the code or if the firmware would change and produce some unexpected response. I never saw this happen (and it was never caught), and the firmware for sure will never change as Miele abandoned it. So I think it's safe to leave like is.

bundles/org.openhab.binding.miele/src/main/java/org/openhab/binding/miele/internal/handler/DiscoveryListener.java
```diff
@@ -0,0 +1,39 @@
...
+     * This method is called whenever any appliance is removed.
+     *
+     * @param appliance The XGW homedevice definition of the appliance that was removed
+     */
+    void onApplianceRemoved(HomeDevice appliance);
+
+    /**
+     * This method is called whenever any appliance is added.
+     *
+     * @param appliance The XGW homedevice definition of the appliance that was removed
```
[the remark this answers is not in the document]
jlaur: Nice catch. This was a copy'n'paste bug when copied method signatures from the old interface which also had the copy'n'paste bug since the binding was created. :) Fixed.

PR #12518, opened 2022-03-25
[knx] Reduce compiler warnings

holgerfriedrich:
added NonNullByDefault annotations
rename functions and local variables not matching the naming scheme


holgerfriedrich:
factored out from #12434

wborn (approved):
Thanks for the cleanup!

PR #12519, opened 2022-03-25
[yamahamusiccast]  New channel for volume in decibel

florian-h05:
# [yamahamusiccast] New channel for volume in decibel

# Description

This PR adds one more volume channel (personally for me, the most important one), the **volume in decibel channel**.
This channel allows you to retrieve the volume level in decibel as shown on the Yamaha device and to control the volume level in decibel (same functionality as the volumeDB channel of the Yamaha Receiver binding).
It also features some small improvements and fixes (have a look at the commit message).

Please note that some imports were changed due to the openHAB codestyle auto-format and spotless.

# Testing

Tested with RX-A750 and RX-V481D.

The built jar can be downloaded here: https://1drv.ms/u/s!Al8cKKxosMjV7gEAVWkq_VDdoCVa?e=VsvJND (outdated).


bundles/org.openhab.binding.yamahamusiccast/src/main/java/org/openhab/binding/yamahamusiccast/internal/YamahaMusiccastHandler.java
```diff
@@ -451,6 +504,7 @@ private void createChannels(String zone) {
         createChannel(zone, CHANNEL_MUTE, CHANNEL_TYPE_UID_MUTE, "Switch");
         createChannel(zone, CHANNEL_VOLUME, CHANNEL_TYPE_UID_VOLUME, "Dimmer");
         createChannel(zone, CHANNEL_VOLUMEABS, CHANNEL_TYPE_UID_VOLUMEABS, "Number");
+        createChannel(zone, CHANNEL_VOLUMEDB, CHANNEL_TYPE_UID_VOLUMEDB, "Number");
```
jlaur, line 507: This could be `Number:Dimensionless` with `Units.DECIBEL`. See https://www.openhab.org/docs/concepts/units-of-measurement.html.

bundles/org.openhab.binding.yamahamusiccast/src/main/java/org/openhab/binding/yamahamusiccast/internal/YamahaMusiccastHandler.java
```diff
@@ -1014,6 +1085,27 @@ private void setVolumeLinkedDevice(int value, @Nullable String zone, String host
...
+                    setVolumeDb(value, "main", host);
+                    break;
+                case 2:
+                    setVolumeDb(value, "zone2", host);
+                    break;
+                case 3:
+                    setVolumeDb(value, "zone3", host);
+                    break;
+                case 4:
+                    setVolumeDb(value, "zone2", host);
```
jlaur, line 1103:
```suggestion
                    setVolumeDb(value, "zone4", host);
```

probably?

bundles/org.openhab.binding.yamahamusiccast/src/main/resources/OH-INF/thing/thing-types.xml
```diff
@@ -38,8 +38,23 @@
 				<options>
 					<option value="none">None</option>
 					<option value="net_radio">Net Radio</option>
+					<option value="tuner">Radio</option>
```
jlaur, line 41: File _src/main/resources/OH-INF/i18n/yamahamusiccast.properties_ needs to be regenerated after adding new strings. Run `mvn i18n:generate-default-translations` in the bundle folder.

bundles/org.openhab.binding.yamahamusiccast/src/main/java/org/openhab/binding/yamahamusiccast/internal/YamahaMusiccastHandler.java
```diff
@@ -1140,6 +1232,29 @@ private String makeRequest(@Nullable String topicAVR, String url) {
...
+    /**
+     * Sets the volume in decibels (dB).
+     *
+     * @param value volume in dB (decibels)
+     * @param zone name of zone
+     * @param host hostname or ip address
+     * @return HTTP request
+     */
+    private @Nullable String setVolumeDb(float value, @Nullable String zone, @Nullable String host) {
+        if (value < Float.parseFloat(getThing().getConfiguration().get("volumeDbMin").toString())) {
```
jlaur, line 1244: Consider introducing a local variable for the parsed values to avoid parsing twice when out of valid range.

bundles/org.openhab.binding.yamahamusiccast/src/main/java/org/openhab/binding/yamahamusiccast/internal/YamahaMusiccastHandler.java
```diff
@@ -12,7 +12,40 @@
...
+import static org.openhab.binding.yamahamusiccast.internal.YamahaMusiccastBindingConstants.CHANNEL_TYPE_UID_SOUNDPROGRAM;
+import static org.openhab.binding.yamahamusiccast.internal.YamahaMusiccastBindingConstants.CHANNEL_TYPE_UID_VOLUME;
+import static org.openhab.binding.yamahamusiccast.internal.YamahaMusiccastBindingConstants.CHANNEL_TYPE_UID_VOLUMEABS;
+import static org.openhab.binding.yamahamusiccast.internal.YamahaMusiccastBindingConstants.CHANNEL_TYPE_UID_VOLUMEDB;
+import static org.openhab.binding.yamahamusiccast.internal.YamahaMusiccastBindingConstants.CHANNEL_VOLUME;
+import static org.openhab.binding.yamahamusiccast.internal.YamahaMusiccastBindingConstants.CHANNEL_VOLUMEABS;
+import static org.openhab.binding.yamahamusiccast.internal.YamahaMusiccastBindingConstants.CHANNEL_VOLUMEDB;
+import static org.openhab.binding.yamahamusiccast.internal.YamahaMusiccastBindingConstants.HTTP;
+import static org.openhab.binding.yamahamusiccast.internal.YamahaMusiccastBindingConstants.LONG_CONNECTION_TIMEOUT_MILLISEC;
+import static org.openhab.binding.yamahamusiccast.internal.YamahaMusiccastBindingConstants.YAMAHA_EXTENDED_CONTROL;
```
jlaur, line 48: It seems wildcard import might be preferable here considering the number of needed constants from this file?

bundles/org.openhab.binding.yamahamusiccast/src/main/resources/OH-INF/thing/thing-types.xml
```diff
@@ -29,7 +29,7 @@
 			<parameter name="syncVolume" type="boolean">
 				<label>Sync Volume</label>
 				<description>Sync Volume across linked Music Cast models</description>
-				<default>true</default>
+				<default>false</default>
```
jlaur, line 32: This could be a small breaking change for some users, why is this default changed?

jlaur (changes requested):
Looks good! Some minor comments to consider.

bundles/org.openhab.binding.yamahamusiccast/src/main/java/org/openhab/binding/yamahamusiccast/internal/YamahaMusiccastHandler.java
florian-h05: Oops, thanks for that comment!

bundles/org.openhab.binding.yamahamusiccast/src/main/resources/OH-INF/thing/thing-types.xml
florian-h05: I have changed this default as the README says that default is false "Sync volume across linked models (default=false)".
Furthermore, I consider false to be more practical as I don't want to automatically sync the volume by default.

bundles/org.openhab.binding.yamahamusiccast/src/main/resources/OH-INF/thing/thing-types.xml
florian-h05: Done!

florian-h05: @jlaur I encountered one big problem that took me quite a while to solve.
See my newest comment in this PR.

florian-h05:
@jlaur @cweitkamp 

That might be interesting for you:

When I run `mvn i18n:generate-default-translations` maven gives the following error:
```
[ERROR] No plugin found for prefix 'i18n' in the current project and in the plugin groups [org.apache.maven.plugins, org.codehaus.mojo] available from the repositories [local (/home/florian/.m2/repository), central (https://repo1.maven.org/maven2), openhab-release (https://openhab.jfrog.io/openhab/libs-release)] -> [Help 1]
```

I solved this by modifyng the `pom.xml` in the repos root and setting the version of the `i18n-maven-plugin` to `3.2.0`. Before it was `${ohc.version}` which is currently `3.3.0-SNAPSHOT`. 

So there is a problem that the i18n plugin for 3.3.0-SNAPSHOT is not on the maven central repo??

jlaur:
> So there is a problem that the i18n plugin for 3.3.0-SNAPSHOT is not on the maven central repo??

Now that you mention it, I believe I've experienced the same problem. I solved it by:
```
mvn org.openhab.core.tools:i18n-maven-plugin:3.2.0:generate-default-translations
```

bundles/org.openhab.binding.yamahamusiccast/src/main/java/org/openhab/binding/yamahamusiccast/internal/YamahaMusiccastHandler.java
florian-h05: Sorry, I am not sure how to implement the Units.DECIBEL. Therefore I only use Number:Dimensionless.

florian-h05:
Is that something interesting for everyone/maintainers?
Should we open an issue?

jlaur:
@florian-h05 - unfortunately the recent merge of #12528 has caused a conflict, can you resolve this?

bundles/org.openhab.binding.yamahamusiccast/src/main/java/org/openhab/binding/yamahamusiccast/internal/YamahaMusiccastHandler.java
```diff
@@ -593,6 +619,11 @@ private void updateStateFromUDPEvent(String zoneToUpdate, UdpMessage targetObjec
             updateState(channel, new DecimalType(volumeState));
         }
 
+        if (volumeDbState != -80f) {
+            channel = new ChannelUID(getThing().getUID(), zoneToUpdate, CHANNEL_VOLUMEDB);
+            updateState(channel, new DecimalType(volumeDbState));
```
jlaur, line 624: To support UoM here you could use:
```suggestion
            updateState(channel, new QuantityType<>(volumeDbState, Units.DECIBEL));
```

bundles/org.openhab.binding.yamahamusiccast/src/main/resources/OH-INF/thing/thing-types.xml
```diff
@@ -78,6 +93,14 @@
...
 		<label>Volume</label>
 		<description>Volume channel - Absolute value</description>
+		<category>SoundVolume</category>
+	</channel-type>
+	<channel-type id="volumeDB">
+		<item-type>Number:Dimensionless</item-type>
+		<label>Volume in dB</label>
+		<description>Volume channel - in decibel (dB)</description>
+		<category>SoundVolume</category>
+		<state min="-80" max="12" step="0.5" pattern="%.1f dB"/>
```
jlaur, line 103: If you will implement UoM, this could be adapted to:
```suggestion
		<state min="-80" max="12" step="0.5" pattern="%.1f %unit%"/>
```

florian-h05:
@jlaur 

Ready for review, I have resolved the conflict in README.

bundles/org.openhab.binding.yamahamusiccast/src/main/java/org/openhab/binding/yamahamusiccast/internal/YamahaMusiccastHandler.java
```diff
@@ -593,6 +621,11 @@ private void updateStateFromUDPEvent(String zoneToUpdate, UdpMessage targetObjec
             updateState(channel, new DecimalType(volumeState));
         }
 
+        if (volumeDbState != -80f) {
```
jlaur, line 624: Float comparison with the `==` or `!=` operators is not recommended. Consider even replacing this type with something else, for example `BigDecimal` in order to use exact comparison.

florian-h05: This comparison is only used to check whether the var had an update.
Does exact comparison matter here?

florian-h05: The volume returned from the Yamaha device always is .0 or .5 and you are not able to set anything else.

jlaur: I'm not a Java expert, so I'm not very much into the details, but I would presume you can risk comparing with a value like -79.9999999997 which would not equal -80. Since your volume steps are exactly 0.5 dB, probably it would be safer to switch to `BigDecimal` where accuracy is guaranteed. See for example https://howtodoinjava.com/java-examples/correctly-compare-float-double/ or many other articles for this topic.

florian-h05: Hmm okay, I can switch if you want me to.

I would just like to mention that the yamahareceiver binding implements volumeDB with float and does the same float comparison.

jlaur: There is of course a possibility that I'm wrong, so if you know that this comparison will always work on all platforms, I'm fine with it. :) Just wanted to share my concern when seeing this and checking articles online. Let me know what you prefer.

florian-h05: > Just wanted to share my concern when seeing this and checking articles online.

So, I do not know if it works on all platforms and in every case, but float is used in the yamahareceiver binding and I did never encounter any issue with volumeDB in that binding.

> Let me know what you prefer.

I prefer the float solution as it is easier to read (no BigDecimal Constructing etc.), but I had a look at the article and I think I will implement the Threshold based comparison (chapter 2 of the article).

But currently, I am wondering if this comparison is even needed. I think I will have a look at the code and then take a decision.

florian-h05: So I have tested much, it does not matter whether that comparison is 100% exact.

florian-h05: @jlaur 
I would leave it as it is now.

bundles/org.openhab.binding.yamahamusiccast/README.md
```diff
@@ -47,6 +50,7 @@ Default value for *defaultAfterMCLink* is *NET RADIO* as most of the models have
 | mute           | Switch | Mute ON/OFF                                                         |
 | volume         | Dimmer | Volume as % (recalculated based on Max Volume Model)                |
 | volumeAbs      | Number | Volume as absolute value                                            |
+| volumeDB       | Number | Volume in decibel (dB) (returns -90 dB if not available on device)  |
```
jlaur, line 53: This should now be updated to `Number:Dimensionless`.

bundles/org.openhab.binding.yamahamusiccast/src/main/java/org/openhab/binding/yamahamusiccast/internal/YamahaMusiccastHandler.java
jlaur: Sure, let's leave it. Just a small update of the README.md needed before we are ready to merge.

florian-h05:
@jlaur 

This should be ready to merge.

jlaur (approved):
Thanks for this addition!

PR #12520, opened 2022-03-25
[knx] handle exceptions during initial read

### LINKED ISSUE #7239: [knx] initial addresses read at startup breaks after exception

holgerfriedrich:
fixes #7239

### DIFF
--- a/bundles/org.openhab.binding.knx/src/main/java/org/openhab/binding/knx/internal/client/AbstractKNXClient.java
+++ b/bundles/org.openhab.binding.knx/src/main/java/org/openhab/binding/knx/internal/client/AbstractKNXClient.java
@@ -14,6 +14,7 @@ package org.openhab.binding.knx.internal.client;
 
 import java.time.Duration;
 import java.util.Set;
+import java.util.concurrent.CancellationException;
 import java.util.concurrent.CopyOnWriteArraySet;
 import java.util.concurrent.LinkedBlockingQueue;
 import java.util.concurrent.ScheduledExecutorService;
@@ -291,6 +292,8 @@ public abstract class AbstractKNXClient implements NetworkLinkListener, KNXClien
                 logger.trace("Sending a Group Read Request telegram for {}", datapoint.getDatapoint().getMainAddress());
                 processCommunicator.read(datapoint.getDatapoint());
             } catch (KNXException e) {
+                // Note: KnxException does not cover KnxRuntimeException and subclasses KnxSecureException,
+                // KnxIllegArgumentException
                 if (datapoint.getRetries() < datapoint.getLimit()) {
                     readDatapoints.add(datapoint);
                     logger.debug("Could not read value for datapoint {}: {}. Going to retry.",
@@ -299,9 +302,15 @@ public abstract class AbstractKNXClient implements NetworkLinkListener, KNXClien
                     logger.warn("Giving up reading datapoint {}, the number of maximum retries ({}) is reached.",
                             datapoint.getDatapoint().getMainAddress(), datapoint.getLimit());
                 }
-            } catch (InterruptedException e) {
+            } catch (InterruptedException | CancellationException e) {
                 logger.debug("Interrupted sending KNX read request");
                 return;
+            } catch (Exception e) {
+                // Any other exception: Fail gracefully, i.e. notify user and continue reading next DP.
+                // Not catching this would end the scheduled read for all DPs in case of an error.
+                // Severity is warning as this is likely caused by a configuration error.
+                logger.warn("Error reading datapoint {}: {}", datapoint.getDatapoint().getMainAddress(),
+                        e.getMessage());
             }
         }
     }


holgerfriedrich:
factored out from #12434

bundles/org.openhab.binding.knx/src/main/java/org/openhab/binding/knx/internal/client/AbstractKNXClient.java
```diff
@@ -299,6 +300,11 @@ private void readNextQueuedDatapoint() {
                     logger.warn("Giving up reading datapoint {}, the number of maximum retries ({}) is reached.",
                             datapoint.getDatapoint().getMainAddress(), datapoint.getLimit());
                 }
+            } catch (KnxRuntimeException e) {
+                // KnxRuntimeException is _not_ a subclass of KnxException. Fail gracefully for this case as well,
+                // fixes #7239
```
lolodomo, line 305: I don't think this is necessary to mention the github issue number directly in code. Please remove it.

bundles/org.openhab.binding.knx/src/main/java/org/openhab/binding/knx/internal/client/AbstractKNXClient.java
```diff
@@ -299,6 +300,11 @@ private void readNextQueuedDatapoint() {
...
+                logger.warn("Error reading datapoint {}: {}", datapoint.getDatapoint().getMainAddress(),
+                        e.getMessage());
```
lolodomo, line 307: Is the WARN level the good level ? As you finally ignore the error, it looks like something no really important...
Is it something you would like the user to be alerted ?
Is there no risk to flood the log file with recurrent logging ?

lolodomo (changes requested):

bundles/org.openhab.binding.knx/src/main/java/org/openhab/binding/knx/internal/client/AbstractKNXClient.java
```diff
@@ -299,6 +300,11 @@ private void readNextQueuedDatapoint() {
                     logger.warn("Giving up reading datapoint {}, the number of maximum retries ({}) is reached.",
                             datapoint.getDatapoint().getMainAddress(), datapoint.getLimit());
                 }
+            } catch (KnxRuntimeException e) {
```
jlaur, line 303: I think you are hiding a bug here. `KnxRuntimeException` is inherited from `RuntimeException` by design. In the linked issue, `KNXIllegalArgumentException` is thrown, so this seems to suggest that `read()` is called with an invalid parameter, `datapoint.getDatapoint()`, in this scenario?

holgerfriedrich: Currently, all kind of KnxRuntimeExceptions will end the scheduled runnable, so all scheduled reads will not be executed after the exception caused by one broken item (as shown in #7239). This effectivly breaks all scheduled reads, not only for the problematic item.

The approach here is targeted to improve the behaviour in case a problem occurs.
I think using the warning for the log this will be sufficient to show that there is an issue.

Remark: KnxSecureException is derived from KnxRuntimeException. AFAIR I was able to trigger it by resetting the secure IP Interface (work for #12434), but let me double check this....

bundles/org.openhab.binding.knx/src/main/java/org/openhab/binding/knx/internal/client/AbstractKNXClient.java
holgerfriedrich: will do

bundles/org.openhab.binding.knx/src/main/java/org/openhab/binding/knx/internal/client/AbstractKNXClient.java
holgerfriedrich: see comment to jlaur below.... maybe info is fine as well.
Yes, because in item is configured in a way which is not supported or something strange happened during runtime.
Don't think so, as this should happen only at the interval of the scheduled read.

lolodomo:
@jlaur : please take the lead to review this PR, your comments are more detailed than mine.

bundles/org.openhab.binding.knx/src/main/java/org/openhab/binding/knx/internal/client/AbstractKNXClient.java
jlaur: I don't know any details here, so something just seemed wrong when catching a RuntimeException as this usually hides deeper problems. So now checked the method throwing the exception:

```
	/**
	 * Sets the data array with KNX datapoint type items for translation.
	 * <p>
	 * The <code>data</code> array contains at least one DPT item, the new item(s) will
	 * replace any other items set in the translator before.<br>
	 * The number of items (KNX data values) for translation in <code>data</code> is
	 * inferred from the length of the usable <code>data</code> range:<br>
	 * <code>items = (data.length - offset) / (length of KNX data type)</code>
	 * <p>
	 * In general, the KNX data type width is implicitly known in the context where a
	 * translator is invoked (e.g., by appropriate DP configuration). Therefore,
	 * <code>data.length</code> will satisfy the minimum acceptable length. If this is
	 * not the case, {@link KNXIllegalArgumentException} has to be caught and handled in
	 * the caller's context.
	 *
	 * @param data byte array containing KNX DPT item(s)
	 * @param offset offset into <code>data</code> from where to start, 0 &lt;= offset
	 *        &lt; <code>data.length</code>
	 * @throws KNXIllegalArgumentException if <code>data.length</code> - offset &lt;
	 *         data type width of this DPTXlator
	 */
	public void setData(final byte[] data, final int offset)
	{
		if (offset < 0 || offset > data.length)
			throw new KNXIllegalArgumentException("illegal offset " + offset);
		final int size = Math.max(1, getTypeSize());
		final int length = (data.length - offset) / size * size;
		if (length == 0)
			throw new KNXIllegalArgumentException("DPT " + dpt.getID() + " " + dpt.getDescription() + ": data length "
					+ (data.length - offset) + " < required datapoint type width " + size);
		this.data = new short[length];
		for (int i = 0; i < length; ++i)
			this.data[i] = ubyte(data[offset + i]);
	}
```

So if in your case the KNX data type width is not known in your context, I guess it's okay to catch and handle the exception. However, if you could somehow validate before calling, that might be preferable.

holgerfriedrich:
@jlaur I had a closer look into the implementation of the Calimero library. The function called in the `try` block may throw more exceptions than declared,,,,

> ProcessCommunicatior.read(...) declares KNXException, InterruptedException

> but it may throw other exceptions as well:
> 
> readFromGroup declares InterruptedException and subclasses of KNXException -> may throw 
> IllegalStateException extends RuntimeException extends Exception   if detached
> InterruptedException extends Exception    in case secure GO is interrupted
> CancellationException extends IllegalStateException extends RuntimeException extends Exception   in case secure GO is interrupted
> KnxTimeoutException extends KnxException    in case secure GO times out
> 
> extractGroupASDU -> may throw KnxIllegalArgumentException  extends KnxRuntimeException extends RuntimeException extends Exception in case APDU length does not match configured data type
> setdata -> may throw KnxIllegalArgumentException extends KnxRuntimeException extends RuntimeException extends Exception
> 
> createTranslator -> may throw
> [IndexOutOfBoundsException extends RuntimeException extends Exception]
> NumberFormatException extends IllegalArgumentException extends RuntimeException extends Exception     DPT type cannot be parsed
> KnxException     no translator, other exceptions re-thrown as KnxException
> 
> DataUnitBuilder.extractASDU -> may throw KNXIllegalArgumentException extends RuntimeException extends Exception    wrong APDU data

So I now changed the code do catch _any_ exception.

The rationale behind it is as follows: any exception will end _all_ scheduled reads, as the scheduled job ends. Catching exceptions and printing error messages is the approach, as it only skips the one faulty DP and continues for all others. 
All exceptions which could be resolved with a retry should be handled in the first catch(KNXException) clause. Interruption is handled separately, as no log with stack trace is needed. All other stuff is printed as a warning, as I do not want to hide any problem.

jlaur:
> So I now changed the code do catch _any_ exception. The rationale behind it is as follows: any exception will end _all_ scheduled reads, as the scheduled job ends.
[....]

Thanks for following up on this. I think it's fine, since you are calling an external library you have no control over. So you need to encapsulate the risk of anything than can happen outside your boundary.

jlaur (approved):
Thanks for the fix.

PR #12015, opened 2022-01-10
[tesla] Clarify and update how refresh tokens can be obtained

kaikreuzer:
Also improved some logging

Signed-off-by: Kai Kreuzer <kai@openhab.org>

Issue #12533, opened 2022-03-28
[tesla] Exception occurred - communication_error


kojid0:
After the [fix of a recent API token problem](https://github.com/openhab/openhab-addons/pull/12530), another problem with the tesla binding has come up.
@tkranz gets this error

```2022-03-27 17:52:41.378 [ERROR] [internal.handler.TeslaAccountHandler] - An exception occurred while connecting to the Tesla back-end: 'null'
at org.openhab.binding.tesla.internal.handler.TeslaAccountHandler.authenticate(TeslaAccountHandler.java:259) ~[?:?]
at org.openhab.binding.tesla.internal.handler.TeslaAccountHandler.lambda$0(TeslaAccountHandler.java:377) ~[?:?]
2022-03-27 17:52:41.443 [INFO ] [ab.event.ThingStatusInfoChangedEvent] - Thing 'tesla:models:12345:54321' changed from OFFLINE (BRIDGE_OFFLINE) to ONLINE
2022-03-27 17:52:41.526 [INFO ] [ab.event.ThingStatusInfoChangedEvent] - Thing 'tesla:models:12345:54321' changed from OFFLINE (BRIDGE_OFFLINE) to ONLINE
2022-03-27 17:52:56.379 [ERROR] [internal.handler.TeslaAccountHandler] - An exception occurred while connecting to the Tesla back-end: 'null'
at org.openhab.binding.tesla.internal.handler.TeslaAccountHandler.authenticate(TeslaAccountHandler.java:259) ~[?:?]
at org.openhab.binding.tesla.internal.handler.TeslaAccountHandler.lambda$0(TeslaAccountHandler.java:377) ~[?:?]
```

and I'm getting this error
```
2022-03-27 17:01:36.557 [WARN ] [internal.handler.TeslaVehicleHandler] - Exception occurred in slowStateRunnable
java.util.ConcurrentModificationException: null
	at java.util.ArrayList.sort(ArrayList.java:1752) ~[?:?]
	at org.apache.cxf.jaxrs.provider.ProviderFactory.sortReaders(ProviderFactory.java:734) ~[?:?]
	at org.apache.cxf.jaxrs.provider.ProviderFactory.setCommonProviders(ProviderFactory.java:662) ~[?:?]
	at org.apache.cxf.jaxrs.client.ClientProviderFactory.setProviders(ClientProviderFactory.java:73) ~[?:?]
[...]

PR #12537, opened 2022-03-28
[tesla] Remove (broken) options to obtain refresh token through credentials

kaikreuzer:
This is a follow-up on #12015 as I no longer have the hope that this feature will come back.


martinvw (approved):

bundles/org.openhab.binding.tesla/src/main/java/org/openhab/binding/tesla/internal/handler/TeslaAccountHandler.java
```diff
@@ -269,33 +267,11 @@ private ThingStatusInfo authenticate() {
         }
 
         if (hasExpired) {
-            String username = (String) getConfig().get(CONFIG_USERNAME);
```
martinvw, line 272: Should/could we delete CONfIG_USERNAME? And password?

kaikreuzer: Certainly so! Done.

lolodomo:
@kaikreuzer : code is not compiling (CONFIG_USERNAME).

kaikreuzer:
Thanks @lolodomo. Looking at it, I think I need to understand why this parameter is still used at that place in the code.
I'll do some further testing and fixing, especially as I'll have to look into #12533 as well...

kaikreuzer:
@lolodomo Compilation issues resolved!

bundles/org.openhab.binding.tesla/src/main/java/org/openhab/binding/tesla/internal/handler/TeslaVehicleHandler.java
```diff
@@ -1027,8 +1027,7 @@ protected boolean establishEventStream() {
                 if (!isEstablished) {
                     eventBufferedReader = null;
 
-                    eventClient = clientBuilder.build()
-                            .register(new Authenticator((String) getConfig().get(CONFIG_USERNAME), vehicle.tokens[0]));
+                    eventClient = clientBuilder.build().register(new Authenticator("", vehicle.tokens[0]));
```
jlaur, line 1030: It seems that this is the only usage of `Authenticator`, so `user` could be safely removed from the constructor. However, in either case please also check this method:

```
        private String getBasicAuthentication() {
            String token = this.user + ":" + this.password;
            return "Basic " + Base64.getEncoder().encodeToString(token.getBytes(StandardCharsets.UTF_8));
        }
```

It seems like this could also be simplified, or actually might need to be rewritten to some extend, if class would now only host a password (or token)?

jlaur (changes requested):
Thanks for cleaning the code. Just one comment for something more to check.

bundles/org.openhab.binding.tesla/src/main/java/org/openhab/binding/tesla/internal/handler/TeslaVehicleHandler.java
kaikreuzer: Haha, you caught me there @jlaur!
I was actually just at this piece of code last night, when I wondered why it still uses the credentials and then I became too tired and went to bed.
I had forgotten about it today and was just happy that it compiled (because I had put in "" as a test last night). 🤣 
Looking at it now again, I noticed that this is very old code for event streaming, which didn't work since years anymore - I hence now completely removed these parts.

bundles/org.openhab.binding.tesla/src/main/resources/OH-INF/thing/account.xml
```diff
@@ -11,18 +11,7 @@
...
-				<advanced>true</advanced>
-				<label>Username</label>
-				<description>Username for the Tesla Remote Service, e.g email address.</description>
-			</parameter>
-			<parameter name="password" type="text" required="false">
-				<advanced>true</advanced>
-				<context>password</context>
-				<label>Password</label>
-				<description>Password for the Tesla Remote Service</description>
+				<description>Refresh token for account authentication. Please see the documentation on how to get hold of it.</description>
```
jlaur, line 14: The i18n file needs to be regenerated after these text changes.

bundles/org.openhab.binding.tesla/src/main/java/org/openhab/binding/tesla/internal/handler/TeslaAccountHandler.java
```diff
@@ -269,33 +261,11 @@ private ThingStatusInfo authenticate() {
...
-                    } else {
-                        return new ThingStatusInfo(ThingStatus.OFFLINE, ThingStatusDetail.CONFIGURATION_ERROR,
-                                "Failed to obtain refresh token with username/password.");
-                    }
-                } else {
-                    return new ThingStatusInfo(ThingStatus.OFFLINE, ThingStatusDetail.CONFIGURATION_ERROR,
-                            "Neither a refresh token nor credentials are provided.");
-                }
+                return new ThingStatusInfo(ThingStatus.OFFLINE, ThingStatusDetail.CONFIGURATION_ERROR,
+                        "No refresh token is provided.");
```
jlaur, line 268: Consider adding this string to the i18n file for being localizable.

jlaur (changes requested):
Thanks! Some minor localization comments added.

bundles/org.openhab.binding.tesla/src/main/resources/OH-INF/thing/account.xml
kaikreuzer: Done.

bundles/org.openhab.binding.tesla/src/main/java/org/openhab/binding/tesla/internal/handler/TeslaAccountHandler.java
kaikreuzer: This isn't done for thing status updates throughout the whole binding yet, so I'd not want to just do it for this single occasion.

bundles/org.openhab.binding.tesla/src/main/resources/OH-INF/i18n/tesla.properties
```diff
@@ -296,3 +292,10 @@ channel-type.tesla.wakeup.label = Wake Up
 channel-type.tesla.wakeup.description = Wake up the vehicle from a (deep) sleep
 channel-type.tesla.wiperbladeheater.label = Wiperblade Heater
 channel-type.tesla.wiperbladeheater.description = Indicates if the wiperblade heater is switched on
+
+# thing types config
+
+thing-type.config.tesla.account.password.label = Password
+thing-type.config.tesla.account.password.description = Password for the Tesla Remote Service
+thing-type.config.tesla.account.username.label = Username
+thing-type.config.tesla.account.username.description = Username for the Tesla Remote Service, e.g email address.
```
jlaur, line 301: Is this auto-generated? Seems strange that a new thing types config section would be generated here when there's already a section from line 19.

kaikreuzer: Yes, I purely ran the generator.

kaikreuzer: But you are right, these entries should actually have disappeared as we do no longer have these parameters...

kaikreuzer: I've now deleted it and regenerated - it didn't appear again.

jlaur: Perfect! Thanks.

jlaur (approved):
Thanks for cleaning up the code as well as the documentation, everything looks good.

PR #12539, opened 2022-03-29
[jrubyscripting] Accept generic global variable names

jimtng:
Remove hard coded list of scoped global variables and use a more generic criteria. This will allow introduction of other / new scoped variables whose names are not already in the hard coded list.


wborn (approved):
Thanks!

Issue #12456, opened 2022-03-10
[homekit]  [ssories.HomekitCharacteristicFactory] - Item type Dimmer is not supported for FanSwimSPA1. Only Number type is supported.


SchmidSP:
## Expected Behavior
as written in https://www.openhab.org/addons/integrations/homekit/
Fan.RotationSpeed should accept dimmer and number should work

Stops working since Openhab 3.3.0 M2.
All versions before work as excepted 

## Current Behavior
as written in https://www.openhab.org/addons/integrations/homekit/
Fan.RotationSpeed should accept dimmer and number

since 3.3.0 M2 only number is accepted

Openhab Log:
[WARN ] [ssories.HomekitCharacteristicFactory] - Item type Dimmer is not supported for FanSwimSPA1. Only Number type is supported.

## Possible Solution
add dimmer in homekit.fan.rotationspeed again

## Steps to Reproduce (for Bugs)
1. add item fanrotation speed:
example: 
Dimmer	Fan "Fan"	(Fan)		{homekit = "RotationSpeed" }
2. See Openhab log for error:
[WARN ] [ssories.HomekitCharacteristicFactory] - Item type Dimmer is not supported for FanSwimSPA1. Only Number type is supported.
if you change the value

## Context
can't use fan.rotationspeed anymore with dimmer

## Your Environment
Rasp 4
Openhab 3.3.0 M2

PR #12540, opened 2022-03-29
[homekit] fix Fan Rotation Speed. add support for dimmer

yfre:
this PR adds support for Dimmer item type for the rotation speed and fixes the issue reported here 
#12456 


lolodomo (approved):
LGTM

PR #12541, opened 2022-03-29
[synopanalyzer] Incorrect octa reported

clinique:
Some stations does not report octa dimension (value = -1), thus leading the misinterpret
Enhanced discovery process (progressive extension of the search radius) 
Code enhancements
SAT corrections


bundles/org.openhab.binding.synopanalyzer/src/main/java/org/openhab/binding/synopanalyzer/internal/stationdb/StationDbService.java
```diff
@@ -0,0 +1,56 @@
...
+public class StationDbService {
+    private final List<Station> stations;
+
+    @Activate
+    public StationDbService() {
+        try (InputStream is = Thread.currentThread().getContextClassLoader().getResourceAsStream("/db/stations.json");
+                Reader reader = new InputStreamReader(is, StandardCharsets.UTF_8);) {
+            Gson gson = new GsonBuilder().setFieldNamingPolicy(FieldNamingPolicy.LOWER_CASE_WITH_UNDERSCORES).create();
+            stations = Arrays.asList(gson.fromJson(reader, Station[].class));
+        } catch (IOException e) {
```
lolodomo, line 48: I am not sure that Gson exceptions are catched

bundles/org.openhab.binding.synopanalyzer/src/main/resources/db/stations.json
```diff
@@ -2420,5 +2419,4 @@
          "elevation":1133,
          "station_type":1
       }
-   ]
-}
\ No newline at end of file
+]
```
lolodomo, line 2422: Probably add a return at the end

lolodomo (changes requested):

bundles/org.openhab.binding.synopanalyzer/src/main/java/org/openhab/binding/synopanalyzer/internal/discovery/SynopAnalyzerDiscoveryService.java
```diff
@@ -71,23 +72,28 @@ public void startScan() {
...
-        Optional<Station> station = stationDB.stations.stream().filter(s -> s.idOmm == nearestId).findFirst();
-        thingDiscovered(DiscoveryResultBuilder.create(new ThingUID(THING_SYNOP, nearestId.toString()))
-                .withLabel(String.format("Synop : %s", station.get().usualName))
-                .withProperty(SynopAnalyzerConfiguration.STATION_ID, nearestId)
-                .withRepresentationProperty(SynopAnalyzerConfiguration.STATION_ID).build());
+            thingDiscovered(DiscoveryResultBuilder.create(new ThingUID(THING_SYNOP, Integer.toString(station.idOmm)))
+                    .withLabel(String.format("Synop : %s", station.usualName))
+                    .withProperty(SynopAnalyzerConfiguration.STATION_ID, station.idOmm)
+                    .withRepresentationProperty(SynopAnalyzerConfiguration.STATION_ID).build());
+        } catch (NoSuchElementException e) {
```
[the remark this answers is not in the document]
clinique: There is no way to test it. `distances.entrySet().iterator().next();` raises `NoSuchElementException` if there is no element available.

bundles/org.openhab.binding.synopanalyzer/src/main/java/org/openhab/binding/synopanalyzer/internal/stationdb/StationDbService.java
lolodomo: My request was not to remove this exception catch which has probably an utility but rather to mention that Gson exceptions are not caught. So I was more expecting additional exception catching.

lolodomo: In case of any problem, you could log something and consider an empty list of stations.

bundles/org.openhab.binding.synopanalyzer/src/main/java/org/openhab/binding/synopanalyzer/internal/discovery/SynopAnalyzerDiscoveryService.java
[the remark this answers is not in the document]
lolodomo: If I correctly remember, it exists a method hasNext() you could use

bundles/org.openhab.binding.synopanalyzer/src/main/java/org/openhab/binding/synopanalyzer/internal/stationdb/StationDbService.java
clinique: > My request was not to remove this exception catch which has probably an utility but rather to mention that Gson exceptions are not caught. So I was more expecting additional exception catching.

Sure, but Gson....create() does not raise any exception, so what should I be catching ?

clinique: > In case of any problem, you could log something and consider an empty list of stations.

Yes, this is better.

lolodomo: I am of course meaning the call to gson.fromJson which can trigger exceptions.
Of course, as you are parsing a local file you created in resources, the risk of exception is more or less null.
But that would be better to catch these exceptions.

clinique: Sure. I don't know where I had the head .??.??..
Done.

lolodomo (approved):
LGTM, thank you

PR #12545, opened 2022-03-29
[miele] Use framework's HTTP client

### LINKED ISSUE #12544: [miele] Use framework HTTP client
Communication with the Miele XGW 3000 gateway could easily be factored into using the framework Jetty HTTP client. This would simplify the code and eliminate some risk of bugs in the current custom implementation of GZIP decompression.

Furthermore, HTTP return codes should be handled correctly. Currently failure is only logged, but not handled appropriately:

```
2022-03-29 18:56:58.420 [DEBUG] [.MieleGatewayCommunicationController] - An unexpected status code was returned: '503'
2022-03-29 18:56:58.423 [TRACE] [.MieleGatewayCommunicationController] - The request '{"jsonrpc":"2.0","id":641713446,"method":"HDAccess/getHomeDevices","params":["(type=SuperVision)"]}' yields '<!DOCTYPE HTML PUBLIC "-//IETF//DTD HTML 2.0//EN">
<html><head>
<title>503 Service Temporarily Unavailable</title>
</head><body>
<h1>Service Temporarily Unavailable</h1>
<p>The server is temporarily unable to service your
request due to maintenance downtime or capacity
problems. Please try again later.</p>
</body></html>
'
2022-03-29 18:56:58.425 [DEBUG] [.internal.handler.MieleBridgeHandler] - An exception occurred while polling an appliance: 'Error parsing JSON response' -> 'com.google.gson.stream.MalformedJsonException: Use JsonReader.setLenient(true) to accept malformed JSON at line 1 column 12 path $'
```

jlaur:
Resolves #12544

Replaces custom implementation in binding with the framework's Jetty HTTP client.

Additionally fails faster in case of HTTP failures instead of trying to parse invalid response content (e.g. HTML) as JSON.

### DIFF
--- a/bundles/org.openhab.binding.miele/src/main/java/org/openhab/binding/miele/internal/MieleGatewayCommunicationController.java
+++ b/bundles/org.openhab.binding.miele/src/main/java/org/openhab/binding/miele/internal/MieleGatewayCommunicationController.java
@@ -12,21 +12,19 @@
  */
 package org.openhab.binding.miele.internal;
 
-import java.io.BufferedInputStream;
-import java.io.ByteArrayOutputStream;
-import java.io.IOException;
-import java.io.InputStream;
-import java.io.OutputStream;
 import java.io.StringReader;
-import java.net.HttpURLConnection;
-import java.net.MalformedURLException;
-import java.net.URL;
-import java.util.Collections;
-import java.util.Map;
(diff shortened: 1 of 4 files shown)


lolodomo (approved):
LGTM, thank you

PR #12246, opened 2022-02-08
[OmniLink] Refactor parts of code

ecdye:
This also removes an unused trigger channel definition that was in
OmnilinkBindingConstants.java.

Related-to [openhab-distro#1367](https://github.com/openhab/openhab-distro/pull/1367)

Signed-off-by: Ethan Dye <mrtops03@gmail.com>

PR #12546, opened 2022-03-30
[OmniLink] Fix daylight savings when setting date/time

ecdye:
Signed-off-by: Ethan Dye <mrtops03@gmail.com>

### DIFF
--- a/bundles/org.openhab.binding.omnilink/README.md
+++ b/bundles/org.openhab.binding.omnilink/README.md
@@ -26,7 +26,6 @@ The OmniPro/Lumina controller acts as a "bridge" for accessing other connected d
 | Access Control Reader Lock | Leviton Access Control Reader                    | `lock`                            |
 
 
-
 ## Discovery
 
 ### Controller
@@ -88,7 +87,7 @@ The devices support some of the following channels:
 | `zone_volume`               | Dimmer               | Volume level of this audio zone.                                                     | `audio_zone`                                        |
 | `zone_source`               | Number               | Source for this audio zone.                                                          | `audio_zone`                                        |
 | `zone_control`              | Player               | Control the audio zone, e.g. start/stop/next/previous.                               | `audio_zone`                                        |
-| `system_date`               | DateTime             | Set controller date/time.                                                            | `controller`                                        |
+| `system_date`               | DateTime             | Controller date/time.                                                            | `controller`                                        |
 | `last_log`                  | String               | Last log message on the controller, represented in JSON.                             | `controller`                                        |
 | `enable_disable_beeper`     | Switch               | Enable/Disable the beeper for this/all console(s).                                   | `controller`, `console`                             |
 | `beep`                      | Switch               | Send a beep command to this/all console(s).                                          | `controller`, `console`                             |
@@ -146,6 +145,56 @@ The devices support some of the following trigger channels:
 | `activated_event`             | Event sent when a button is activated.                                               | `button`                            |
 | `switch_press_event`          | Event sent when an ALC, UPB, Radio RA, or Starlite switch is pressed.                | `dimmable`, `upb`                   |
 
+
+## Rule Actions
+
+This binding includes a rule action, which allows synchronizing the controller time to match the openHAB system time with a user specified zone.
+There is a separate instance for each contoller, which can be retrieved through:
+
+:::: tabs
+
+::: tab JavaScript
+
+``` javascript
+var omnilinkActions = actions.get("omnilink", "omnilink:controller:home");
+```
+
+:::
+
+::: tab DSL
+
+``` php
+val omnilinkActions = getActions("omnilink", "omnilink:controller:home")
+
+```
+
+:::
+
+::::
+
+where the first parameter always has to be `omnilink` and the second is the full Thing UID of the controller that should be used.
+Once this action instance is retrieved, you can invoke the `synchronizeControllerTime(String zone)` method on it:
+
+:::: tabs
+
+::: tab JavaScript
+
+``` javascript
+omnilinkAction.synchronizeControllerTime("America/Denver");
+```
+
+:::
+
+::: tab DSL
+
+``` php
+omnilinkAction.synchronizeControllerTime("America/Denver")
+
+```
+
+:::
+
+::::
 
 ## Full Example
 
@@ -307,12 +356,4 @@ DateTime   OmniProTime   "Last Time Update [%1$ta %1$tR]"   <time>   {channel="o
 13=Arming day instant
 14=Arming night delay
 =Unknown
-```
-
-### Example EMCAScript rule
-
-``` javascript
-var omnilinkAction = actions.get("omnilink", "omnilink:controller:home");
-
-omnilinkAction.setDateTime("America/Denver");
 ```
--- a/bundles/org.openhab.binding.omnilink/README.md
+++ b/bundles/org.openhab.binding.omnilink/README.md
@@ -312,9 +312,7 @@ DateTime   OmniProTime   "Last Time Update [%1$ta %1$tR]"   <time>   {channel="o
 ### Example EMCAScript rule
 
 ``` javascript
-var ZonedDateTime = Java.type("java.time.ZonedDateTime");
-var ZoneId = Java.type("java.time.ZoneId");
-var zdt = ZonedDateTime.now(ZoneId.of("America/Denver"));
+var omnilinkAction = actions.get("omnilink", "omnilink:controller:home");
 
-events.sendCommand('OmniProTime', new StringType(zdt));
+omnilinkAction.setDateTime("America/Denver");
 ```
--- a/bundles/org.openhab.binding.omnilink/README.md
+++ b/bundles/org.openhab.binding.omnilink/README.md
@@ -316,5 +316,5 @@ var ZonedDateTime = Java.type("java.time.ZonedDateTime");
 var ZoneId = Java.type("java.time.ZoneId");
 var zdt = ZonedDateTime.now(ZoneId.of("America/Denver"));
 
-events.sendCommand('DateTime', new StringType(zdt));
+events.sendCommand('OmniProTime', new StringType(zdt));
 ```
--- a/bundles/org.openhab.binding.omnilink/README.md
+++ b/bundles/org.openhab.binding.omnilink/README.md
@@ -309,13 +309,12 @@ DateTime   OmniProTime   "Last Time Update [%1$ta %1$tR]"   <time>   {channel="o
 =Unknown
 ```
 
-### Example `omnilink.rules`
+### Example EMCAScript rule
 
-```
-rule "Update OmniPro Time"
-when
-  Time cron "0 0 0/1 1/1 * ? *"
-then
-  OmniProTime.sendCommand( new DateTimeType() )
-end
+``` javascript
+var ZonedDateTime = Java.type("java.time.ZonedDateTime");
+var ZoneId = Java.type("java.time.ZoneId");
+var zdt = ZonedDateTime.now(ZoneId.of("America/Denver"));
+
+events.sendCommand('DateTime', new StringType(zdt));
 ```
--- a/bundles/org.openhab.binding.omnilink/README.md
+++ b/bundles/org.openhab.binding.omnilink/README.md
@@ -8,7 +8,6 @@ It connects to many other devices through serial ports or wired contacts and exp
 
 The OmniPro/Lumina controller acts as a "bridge" for accessing other connected devices.
 
-
 | Omni type                  | Hardware Type                                    | Things                            |
 |:---------------------------|:-------------------------------------------------|:----------------------------------|
 | Controller                 | Omni (Pro II, IIe, LTe), Lumina                  | `controller` (omni, lumina)       |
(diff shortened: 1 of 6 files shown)


ecdye:
Assuming that this goes through before the 3.3 release a breaking change notice is already present for this channel due to unrelated changes made previously see https://github.com/openhab/openhab-addons/pull/12246.

lolodomo:
Why a date should be stored in a string item rather than a date item? I don't understand the logic. 
Is there an issue describing the problem you are trying to resolve?

lsiepel:
> Because in order to allow calculation of DST we need to preserve the location data but DateTimeType does not do that which causes the DST flag to always be false. This was brought up in the forum discussion that was auto posted here above.

The problem is that the DateTimeType lacks timezone information. As a solution DateTimeType is switched to a StringType item so one can add the timezone and convert that string to a ZonedDateTime, right? Not very user friendly and intuitive. 

Not sure how feasible this is, but you might look into another scenario: Get the openhab Time Zone Settings from the i18n services. But not sure if that is always available to the binding. 

I also wonder how other bindings solve this problem.

jlaur:
> The problem is that the DateTimeType lacks timezone information.

`DateTimeType` internally stores a `ZonedDateTime`? See also https://www.openhab.org/javadoc/latest/org/openhab/core/library/types/datetimetype.

lsiepel:
Sorry, i menat not only timezine. It is the combination timezone / locale that is needed to determine if DST is active or not.

ecdye:
Correct, time zone and locale are required for DST calculation. However, DateTimeType drops the locale information when converting to/from ZonedDateTime. 

I forgot to update the rules documentation to describe how to send the command with the changes but will do that in the morning.

jlaur:
> Correct, time zone and locale are required for DST calculation. However, DateTimeType drops the locale information when converting to/from ZonedDateTime.

IMHO `ZonedDateTime` has even too much information for a DateTimeType. :) A UTC timestamp would have been sufficient for recording an exact unambiguous moment in time. This could then be used with the openHAB/system timezone (including DST) at any moment to produce a local datetime or to calculate if timestamp was generated in DST or not. This will also be resilient to any changes in the openHAB/system timezone after timestamp was generated.

I still don't understand the need, so as also requested by @lolodomo - maybe a description of the exact problem being solved would help?

jlekhter:
Hi,

The need for the request is the following. The Ommi system requires you to identify if your current location is in DST. In openHAB, DateTimeType stores the Date and Time in ISO8601 format with just a simple timezone offset. This is not sufficient for determining whether you are in DST or not. For instance "-0700" is in DST if you are in California, but it is not in DST if you are in Arizona. You need a location to determine DST. Using the DateTimeType in openHAB doesnt give you the location.

@jlaur @lolodomo What exactly are you recommending so that we can be unambiguous about the location of the UTC offset?

bundles/org.openhab.binding.omnilink/src/main/java/org/openhab/binding/omnilink/internal/handler/OmnilinkBridgeHandler.java
```diff
@@ -492,7 +493,7 @@ private void getSystemStatus() throws IOException, OmniNotConnectedException, Om
                 .append(String.format("%02d", status.getHour())).append(":")
                 .append(String.format("%02d", status.getMinute())).append(":")
                 .append(String.format("%02d", status.getSecond())).toString();
-        updateState(CHANNEL_SYSTEM_DATE, new DateTimeType(dateString));
+        updateState(CHANNEL_SYSTEM_DATE, new StringType(dateString));
```
jlaur, line 496: Leaving DateTimeType/StringType discussion aside for a moment, I wonder if something like this would have been a better approach:
```
        updateState(CHANNEL_SYSTEM_DATE,
                new DateTimeType(ZonedDateTime.ofLocal(LocalDateTime.of(status.getYear(), status.getMonth(),
                        status.getDay(), status.getHour(), status.getMinute(), status.getSecond()),
                        ZoneId.systemDefault(), null)));
```

Also, I can see that `SystemStatus` returned by the controller contains `daylightSavings`, but unfortunately that seems useless without also knowing the time zone. Do you need time zone/DST information from the controller, or simply from the openHAB system, so you can assume that system time returned by the controller is local time in openHAB's timezone?

The example is for simplicity using system time zone. Using `TimeZoneProvider` it's possible to get the time zone configured in openHAB.

jlaur:
> The Ommi system requires you to identify if your current location is in DST.

When setting the date/time on the controller, or for what purpose?

> In openHAB, `DateTimeType `stores the Date and Time in ISO8601 format with just a simple timezone offset. This is not sufficient for determining whether you are in DST or not. For instance "-0700" is in DST if you are in California, but it is not in DST if you are in Arizona. You need a location to determine DST. Using the `DateTimeType `in openHAB doesnt give you the location.

Thanks for clarifying this part. I think I'm beginning to understand the issue, but I'm still not entirely there. :-) I guess I still don't understand why and when the Omni system needs the DST flag.

> @jlaur @lolodomo What exactly are you recommending so that we can be unambiguous about the location of the UTC offset?

No recommendation yet, but getting there. Please see my code comments - for now raises more questions than answered, but hopefully that will turn around soon. Is this problem inconsistency in the controller wanting DST, but not time zone? What can the controller do when having this information? Is the controller configured with local time (and DST on/off)? How do you experience the current problem as a user - what exactly happens?

jlekhter:
We are writing to the controller to set its time. The Omni controllers are notoriously bad a keeping a good clock. My clock drifts by 2-3min per day. Therefore at 1am nightly, I set the controller time through openHAB. The set time API to the controller requires you to let the controller know if the user is currently in DST. It has nothing to do with setting DST or ST, it simply part of the API to the controller

Doing this `boolean inDaylightSavings = ZoneId.systemDefault().getRules().isDaylightSavings(zdt.toInstant());`  fixes the issue too. However @ecdye was concerned that using the systemDefault  zoneId broke the API contract with the user. If the user passed in their own zoneId for whatever reason, the code would ignore it, hence the string.

Let me know if I can answer any other questions.

Thanks,

Jerry

ecdye:
Everything that @jlekhter said is correct, sorry about the silence on my part, I have been extra busy these last two days. Let me know if there are any further questions.

bundles/org.openhab.binding.omnilink/src/main/java/org/openhab/binding/omnilink/internal/handler/OmnilinkBridgeHandler.java
ecdye: Perhaps, but I actually didn't write that code in the first place. I migrated the existing code and got it up to spec for inclusion in the openHAB addons proper repo and never really changed it. At this point it doesn't make a to much sense to have it be a date time type because of the issues discussed above but that has yet to be resolved, so for the moment I plan to leave it.

jlaur:
Thanks for this additional explanation, I think I finally understand the details at play now.

> We are writing to the controller to set its time. The Omni controllers are notoriously bad a keeping a good clock. My clock drifts by 2-3min per day. Therefore at 1am nightly, I set the controller time through openHAB. The set time API to the controller requires you to let the controller know if the user is currently in DST. It has nothing to do with setting DST or ST, it simply part of the API to the controller

Will you need to manually flip the DST flag twice per year when switching between summer time and normal time, or will the controller automatically do that when you are **not** interfering in any way by using the API to set the clock?

If it can do that automatically, all you would need to do is read the DST flag from the controller and apply this again when setting the new time. That should fix the issue, but it wouldn't give you full control, i.e. the ability to flip the flag by openHAB - will be addressed in next comment.

>> How do you experience the current problem as a user - what exactly happens?
 
> Because the DST flag isnt set when you set the controller time, all of the controller rules happen an hour too early. For instance, I lower a blackout shade when it gets dark via the controller. Its lowered an hour earlier. I turn on my landscape lights when it get dark -- they turn on too early. I water my lawn at dawn, it gets watered earlier than when I want. Etc.

> I could use openHAB rules for all of that, but the controller is really reliable (except for the time drift) and I lose network more frequently than I lose the controller.

Completely understandable. I have the same situation with my shades having automations defined in a Hunter Douglas PowerView Hub. I have winter and summer automations depending on sunrise time (and DST actually), so recently added ability from openHAB to just flip the automations four times per year instead of letting openHAB actually be in charge of them: #11516

> Let me know if I can answer any other questions.

Thanks for the insight. I now have some propositions:
- If you don't need manual control over DST, just could perhaps simply fetch and reapply the DST flag when setting the time?
- If you need manual control, I would propose to keep the DateTime, but add a dedicated Switch channel for reading/writing the DST flag.
- Perhaps it could be nice to add a thing action which would simply synchronize the controller with the openHAB time and DST flag. This could be called from a rule without any parameters, so would be really simple to use.

jlekhter:
The controller should flip the flag when DST occurs. Its not something that I've tested, but the controller does have settings for changing to Daylight Savings. The DST flag could be read using a REQUEST SYSTEM STATUS command to the controller. So that's an option.

Having a separate switch channel would only store data in the binding, not really update the controller. The DST flag cannot be set without setting the controller date and time.  It seems more complicated than necessary to use 2 channels to update the date/time of the controller.

My favorite option is the Thing action which would synch the controller time with openHAB time.  But that would leave the current channel broken.

I think to fix the current channel, @ecdye's idea of using a String to store the ZonedDateTime value is a good idea. We could always add Thing action later. That's just my opinion though :)

jlaur:
> The controller should flip the flag when DST occurs. Its not something that I've tested, but the controller does have settings for changing to Daylight Savings. The DST flag could be read using a REQUEST SYSTEM STATUS command to the controller. So that's an option.
[...]
> I think to fix the current channel, @ecdye's idea of using a String to store the ZonedDateTime value is a good idea. We could always add Thing action later. That's just my opinion though :)

Why is it better than keeping the DateTime and simply reapplying the current DST flag? The String solution seems like a work-around that could easily be avoided this way:
- Read system time from controller (with DST flag).
- Set system time from command's DateTime with DST flag received from controller.
- Time is now set and DST flag is preserved.

Anything I'm missing?

> Having a separate switch channel would only store data in the binding, not really update the controller. The DST flag cannot be set without setting the controller date and time. It seems more complicated than necessary to use 2 channels to update the date/time of the controller.

That channel would only be needed if there is any need/requirement to be able to change the controller's DST flag from the binding. It could work the same way as described above:
- Read system time from controller.
- Set system time from command's DST flag with time received from controller.
- DST flag is now set and system time is preserved.

But if changing DST is not needed, there is no reason to introduce this channel.

> My favorite option is the Thing action which would synch the controller time with openHAB time. But that would leave the current channel broken.

We could have both. :-)

ecdye:
I don't think that it actually does flip it automatically on all models. I can't test but at least in my testing of the models I have only the OmniPro II controller did, the LTE mode didn't.

jlaur:
> And besides that, what if the user wants to update the daylight savings setting because for some reason it is wrong? Then your approach won't work.

It will when introducing an additional Switch channel for managing the DST flag: https://github.com/openhab/openhab-addons/pull/12546#issuecomment-1086580340

ecdye:
Once again that leaves the issue of in order to change the DST flag on the controller you have to update the time too. I personally think that separating the channels in this case is more complicated and less user friendly. I would prefer to stick with the string approach.

What would be most ideal is if DateTimeType could be fixed to properly preserve the locale information so at some point in the future we could move back to it. Also given the user base of the binding this change is not going to have a large impact as to may knowledge there are only a handful of users who use the binding.

jlekhter:
Can we also consider the use cases being solved? There is no use case in the HAI controller for flipping the DST flag. The only use case we're considering is to set the time of the controller properly. The controller requires local date/time + DST flag. As @ecdye mentioned, the real fix is to fix DateTimeType. Everything else is a work around.  The StringType is a good workaround for now. This allows people to update the time of a remote controller and a local controller through openHAB, which is really the only use case.

jlaur:
> Once again that leaves the issue of in order to change the DST flag on the controller you have to update the time too.

No, that was addressed here: https://github.com/openhab/openhab-addons/pull/12546#issuecomment-1086847185

> I personally think that separating the channels in this case is more complicated and less user friendly. I would prefer to stick with the string approach.

Well, IMHO is actually simplifies the logic when separating time and DST flag:
- System time is contained in the appropriate type (DateTime) like you will see in all other bindings.
- When you need to flip the DST flag, you simply update the DST channel (without having to consider time).
- When you need to set the system time, you simply update the time (without having to consider DST).

With the String approach:
- You need to learn about [ISO_ZONED_DATE_TIME](https://docs.oracle.com/javase/8/docs/api/java/time/format/DateTimeFormatter.html#ISO_ZONED_DATE_TIME) for providing something like "2011-12-03T10:15:30+01:00[Europe/Paris]" to set the time/DST.
- When setting the time, you also have to set DST.
- When setting DST, you also have to set time.

> What would be most ideal is if DateTimeType could be fixed to properly preserve the locale information

I disagree here. DateTimeType should actually internally store an Instant instead of a ZonedDateTime. At any moment this Instant could be converted to ZonedDateTime or LocalDateTime using TimeZoneProvider and/or LocationProvider.

jlaur:
> Can we also consider the use cases being solved? There is no use case in the HAI controller for flipping the DST flag. The only use case we're considering is to set the time of the controller properly.

If that's the only use case, we are back to: https://github.com/openhab/openhab-addons/pull/12546#issuecomment-1086847185 without need for a separate channel. So DateTime can be preserved at the same time as solving the issue?

ecdye:
You don't need to learn about ISO_ZONED_DATE_TIME as I have already provided a rule that constructs it for you. I have to disagree with the DateTimeType only storing an instant. It is dropping valuable metadata by removing the locale description.

jlekhter:
Let's look at the use case. As an end user, I want to set the time of my controller.

My openHAB system is in California and I have 2 controllers.The first one is colocated with my openHAB instance in California and the other is in a remote location in Arizona. 

All I need to do to set the time of both controllers is:

` 
       OmniCA_SetTime.sendCommand(new DateTimeType().zonedDateTime.now().toString )
       OmniAZ_SetTime.sendCommand(new DateTimeType().zonedDateTime.now().zoneIdof("Arizona/Pheonix").toString )
`
That seems a lot easier than setting a switch to ON or OFF before I have set the time. And there is NO Api on the controller to simply toggle the DST flag. You simply cannot do it. You have to update the date/time when you update the DST flag.

What time would you set for the flag? it seems like a problem that the folks over at HAI hadn't seen a reason to solve.

ecdye:
@fwolter I am ok with implementing your suggestion. I have made a preliminary attempt which seems to work. However I can't figure out how to get the TimeZone from TimeZoneProvider only from the system using the Java default which doesn't feel right. Any insight on how to get the configured time zone from openHAB?

fwolter:
There's a bunch of bindings using it. See https://github.com/openhab/openhab-addons/search?q=TimeZoneProvider

You get the time zone with `timeZoneProvider.getTimeZone()`.

ecdye:
@fwolter Alright, I think it's ready for review in the new form now. Let me know what you think

bundles/org.openhab.binding.omnilink/src/main/java/org/openhab/binding/omnilink/internal/action/OmnilinkActions.java
```diff
@@ -0,0 +1,75 @@
...
+            this.handler = (OmnilinkBridgeHandler) handler;
+        }
+    }
+
+    @Override
+    public @Nullable ThingHandler getThingHandler() {
+        return handler;
+    }
+
+    @RuleAction(label = "@text/actionLabel", description = "@text/actionDesc")
```
jlaur, line 59: I don't know if still needed, but according to the [documentation](https://www.openhab.org/docs/developer/bindings/#actions-bound-to-a-thing) a static method of the same name should exist in order to support DSL rules. Have you tested this action from a DSL rule?

bundles/org.openhab.binding.omnilink/src/main/java/org/openhab/binding/omnilink/internal/action/OmnilinkActions.java
```diff
@@ -0,0 +1,75 @@
...
+    public void setDateTime(
```
jlaur, line 60: Consider the name of the action: It sets the date, time and DST flag according to the zone provided as parameter. Perhaps **synchronizeClock** or something similar? Only for consideration.

bundles/org.openhab.binding.omnilink/README.md
```diff
@@ -309,13 +309,10 @@ DateTime   OmniProTime   "Last Time Update [%1$ta %1$tR]"   <time>   {channel="o
...
-rule "Update OmniPro Time"
-when
-  Time cron "0 0 0/1 1/1 * ? *"
-then
-  OmniProTime.sendCommand( new DateTimeType() )
-end
+``` javascript
+var omnilinkAction = actions.get("omnilink", "omnilink:controller:home");
+
+omnilinkAction.setDateTime("America/Denver");
```
jlaur, line 317: I believe I read somewhere that rule examples still need to be documented as DSL rules for consistency across all bindings. And **only** DSL rules to avoid redundancy.

bundles/org.openhab.binding.omnilink/src/main/java/org/openhab/binding/omnilink/internal/action/OmnilinkActions.java
```diff
@@ -0,0 +1,75 @@
...
+            @ActionInput(name = "zone", label = "@text/actionInputZoneLabel", description = "@text/actionInputZoneDesc") @Nullable String zone) {
+        OmnilinkBridgeHandler actionsHandler = handler;
+        ZonedDateTime zdt;
+        if (zone != null) {
+            zdt = ZonedDateTime.now(ZoneId.of(zone));
```
jlaur, line 65: Can throw `DateTimeException` (including `ZoneRulesException` inherited from `DateTimeException`).

bundles/org.openhab.binding.omnilink/src/main/java/org/openhab/binding/omnilink/internal/handler/OmnilinkBridgeHandler.java
```diff
@@ -105,7 +108,8 @@ public OmnilinkBridgeHandler(Bridge bridge) {
 
     @Override
     public Collection<Class<? extends ThingHandlerService>> getServices() {
-        return Collections.singleton(OmnilinkDiscoveryService.class);
+        return Collections.unmodifiableSet(
+                Stream.of(OmnilinkDiscoveryService.class, OmnilinkActions.class).collect(Collectors.toSet()));
```
jlaur, line 112:
```suggestion
        return Set.of(OmnilinkDiscoveryService.class, OmnilinkActions.class);
```

bundles/org.openhab.binding.omnilink/src/main/resources/OH-INF/i18n/omnilink.properties
```diff
@@ -318,3 +318,10 @@ channel-type.omnilink.zone_latched_alarm_status.state.option.1 = Tripped
 channel-type.omnilink.zone_latched_alarm_status.state.option.2 = Reset, but previously tripped
 channel-type.omnilink.zone_restore.label = Restore Zone
 channel-type.omnilink.zone_restore.description = Send a 4 digit user code to restore this zone.
+
+# thing actions
+
+actionInputZoneLabel = Time zone
+actionInputZoneDesc = The time zone of the controller, provided in the "America/Denver" format.
+actionLabel = Set the Date/Time of the controller
+actionDesc = Sets the current Date/Time of the controller to the correct value, including daylight savings.
```
jlaur, line 327: The "correct value" is in this case the date and time of the system where openHAB is running. Maybe this could be written more precisely to avoid any doubts? Perhaps something like: "Synchronizes the system time and DST of the controller with openHAB's system time."

jlaur (changes requested):
Thanks for this improvement which seems to really simplify the use case of synchronizing the time on the controller. I have added a few minor comments.

bundles/org.openhab.binding.omnilink/src/main/resources/OH-INF/i18n/omnilink.properties
ecdye: Well, that is not necessarily true, because it synchronizes it with the provided locale in the action input, right?

bundles/org.openhab.binding.omnilink/src/main/java/org/openhab/binding/omnilink/internal/action/OmnilinkActions.java
ecdye: Because the internal method used by the jomnilink library is setTime, it made more sense for my sanity and keeping track to keep it in the same style here. If you really think that it would be beneficial I can change it. It was also what the system channel name was before.

bundles/org.openhab.binding.omnilink/README.md
ecdye: Hmm, well I would prefer to use JS because it is the direction OH3 is headed. It also makes it easier to implement in a new system because JS is so much more flexible. I haven't seen any formal declaration like that in the documentation any other thoughts from the other @openhab/add-ons-maintainers?

digitaldan: I'm not aware that we standardize on DSL rules in docs, but i agree that while out DSL is not going away anytime soon, it is seeing a decline in usage in favor of our JS automation engine, I'm ok with starting to see more docs reference JS rules, but welcome others point of view.

wborn: Supporting multiple scripting languages using tabs might also be an option. 

Like this:

https://www.openhab.org/docs/configuration/jsr223.html#example-rules-for-a-first-impression
https://github.com/openhab/openhab-docs/blob/main/configuration/jsr223.md#example-rules-for-a-first-impression

But I don't know if that also works here. :slightly_smiling_face:

ecdye: > But I don't know if that also works here. 🙂

I think that it does, I like that idea.

ecdye:
@jlaur I've update to address all of your review comments.

ecdye:
bump

jlaur:
> @jlaur I've updated the code to address all of your review comments. 

I'm currenty on Easter holiday, will be able to resume review within the next couple of days.

bundles/org.openhab.binding.omnilink/README.md
```diff
@@ -88,7 +87,7 @@ The devices support some of the following channels:
 | `zone_volume`               | Dimmer               | Volume level of this audio zone.                                                     | `audio_zone`                                        |
 | `zone_source`               | Number               | Source for this audio zone.                                                          | `audio_zone`                                        |
 | `zone_control`              | Player               | Control the audio zone, e.g. start/stop/next/previous.                               | `audio_zone`                                        |
-| `system_date`               | DateTime             | Set controller date/time.                                                            | `controller`                                        |
+| `system_date`               | DateTime             | Controller date/time.                                                            | `controller`                                        |
```
jlaur, line 90: You could also refer/link to the rule actions chapter here for instruction how to set system date/time.

bundles/org.openhab.binding.omnilink/README.md
```diff
@@ -147,6 +146,56 @@ The devices support some of the following trigger channels:
 | `switch_press_event`          | Event sent when an ALC, UPB, Radio RA, or Starlite switch is pressed.                | `dimmable`, `upb`                   |
 
 
+## Rule Actions
+
+This binding includes a rule action, which allows synchronizing the controller time to match the openHAB system time with a user specified zone.
+There is a separate instance for each contoller, which can be retrieved through:
+
+:::: tabs
```
jlaur, line 154: Curious: How can this be tested? GitHub doesn't display it correctly, so not sure how to verify.

bundles/org.openhab.binding.omnilink/src/main/resources/OH-INF/i18n/omnilink.properties
```diff
@@ -318,3 +318,10 @@ channel-type.omnilink.zone_latched_alarm_status.state.option.1 = Tripped
 channel-type.omnilink.zone_latched_alarm_status.state.option.2 = Reset, but previously tripped
 channel-type.omnilink.zone_restore.label = Restore Zone
 channel-type.omnilink.zone_restore.description = Send a 4 digit user code to restore this zone.
+
+# thing actions
+
+actionInputZoneLabel = Time zone
+actionInputZoneDesc = The time zone of the controller, provided in the "America/Denver" format.
+actionLabel = Synchronize the Date/Time and DST flag of the controller
```
jlaur, line 326: Can you shorten the label? Perhaps simply "Synchronize controller Date/Time" and let the description supply the detailed version.

jlaur (changes requested):
Thanks, looks good! Only a few additional minor comments.

bundles/org.openhab.binding.omnilink/README.md
ecdye: I think that the only way right now is testing by doing. I'm not sure that there is another good way to test it.

ecdye:
@jlaur Concerns addressed, I think we are ready for merge now.

jlaur:
> @jlaur Concerns addressed, I think we are ready for merge now.

Let's get it merged then - thanks!

jlaur (approved):
LGTM

jlaur:
@ecdye - now it would be nice if you could create separate PR for correcting this line in the upgrade notes:
https://github.com/openhab/openhab-distro/blob/c4a1d03455cff53712d07ef0c86588097f02f1b7/distributions/openhab/src/main/resources/bin/update.lst#L67

jlekhter:
@ecdye sorry, for the late testing. It looks like the javascript version of this fix is working fine, but the DSL Rules portion is not working. 

I get the following error when I run DSL rules:
`2022-04-24 13:38:19.731 [ERROR] [internal.handler.ScriptActionHandler] - Script execution of rule with UID 'omnilink-2' failed: 'synchronizeControllerTime' is not a member of 'org.openhab.core.thing.binding.ThingActions'; line 26, column 9, length 60 in omnilink`

It seems like the DSL interpreter cant find the action.

ecdye:
@jlekhter Did you delete and re-add your controller after updating the binding?

jlekhter:
@ecdye. Yes I did. I can try doing it again when I get home, but I deleted the controller several times during the upgrade to make sure. 

Using JavaScript works, but if I use a rules file or a script from within OpenHAB I get the same error as above. 

I never restarted OpenHAB. I could try restarting openhab too later today.

jlekhter:
I deleted the controller, re-added it, restarted openHAB and still the same problem with DSL rules. Here is the error from within the script interface in openHAB

```
2022-04-25 14:27:10.252 [ERROR] [internal.handler.ScriptActionHandler] - Script execution of rule with UID 'test_setting_time' failed: val omniActions = getActions("omnilink", "omnilink:controller:home")
omniActions.synchronizeControllerTime("America/Los_Angeles")
      
   The method synchronizeControllerTime(String) is undefined for the type ThingActions; line 2, column 81, length 25
```

jlaur:
> I deleted the controller, re-added it, restarted openHAB and still the same problem with DSL rules. Here is the error from within the script interface in openHAB
> 
> ```
> 2022-04-25 14:27:10.252 [ERROR] [internal.handler.ScriptActionHandler] - Script execution of rule with UID 'test_setting_time' failed: val omniActions = getActions("omnilink", "omnilink:controller:home")
> omniActions.synchronizeControllerTime("America/Los_Angeles")
>       
>    The method synchronizeControllerTime(String) is undefined for the type ThingActions; line 2, column 81, length 25
> ```

Here is the problem: https://github.com/openhab/openhab-addons/blob/fdc3206d84aa394890c44c7c088939f4acc1ad19/bundles/org.openhab.binding.omnilink/src/main/java/org/openhab/binding/omnilink/internal/action/OmnilinkActions.java#L77-L78

jlekhter:
Yes that works. Changing the DSL rule to synchronizeSystemTime works.  Javascript still wants synchronizeControllerTime as the method.

The methods should be consistent across both rule sets.

jlaur:
> The methods should be consistent across both rule sets. 

Created #12655 for that.

jlekhter:
Just to finish this off, I retested with the fix and it seems to work with DSL rules now. Thank you. But is this also supposed to work when you do not pass an argument into the action? 

If its supposed to work, then it doesnt. I was under the impression that you could write this and it would work:
```
      val omniActions = getActions("omnilink", "omnilink:controller:home")
      omniActions.synchronizeControllerTime()
```
But I get a scripting error when the parameter is not there.

ecdye:
> But I get a scripting error when the parameter is not there.

That is because of the JavaScript, it requires a parameter try `omniActions.synchronizeControllerTime("")` and it should set it to your default time zone.

jlekhter:
Thanks -- an empty string works fine.

All good then, Also as an FYI, this does require an openHab restart. Works like a charm after that.

Thanks for all the work.

Jerry

PR #12553, opened 2022-03-31
[pulseaudio] Fix misconfiguration example in README

dalgwen:
As seen [on the community forum](https://community.openhab.org/t/pulseaudio-thing-initialization-during-startup-to-early-and-therefore-causing-an-exception/134657) by a user encountering an issue, the configuration file example in the pulseaudio README has an error with the parameter activateSimpleProtocolSink


lolodomo (approved):
LGTM

PR #12192, opened 2022-02-03
[homematic] Remove state description step size handling

maniac103:
The RPC protocol doesn't provide this value, so it always was made up
more or less arbitrarily. Since the UI now uses this value for
validation purposes, there are cases where valid values can not be
entered due to this step size (particularly for datapoints with more
than 1 decimal digit).

See #12183 for a concrete breakage example.

Fixes #12183

PR #12557, opened 2022-04-01
[homematic] Validate datapoint values before writing to config

maniac103:
HM config datapoint values on some devices can be out of their valid
range in some cases, particularly if they are unused by the device
currently. Since openhab-core now validates attempts to update
configuration by the thing handlers, make sure we always send a valid
configuration even for those affected datapoints.

One example for such behaviour is HmIPW-DRBL4, which has multiple datapoints which depend on each other:
- POWERUP_JUMPTARGET can have values OFF, ON, OFF_DELAY, ON_DELAY
- POWERUP_ONDELAY_VALUE, POWERUP_ONDELAY_UNIT are only used if POWERUP_JUMPTARGET has value ON_DELAY
- likewise, POWERUP_OFFDELAY_VALUE and POWERUP_OFFDELAY_UNIT are only used if POWERUP_JUMPTARGET has value OFF_DELAY
- if e.g. the POWERUP_JUMPTARGET is OFF, e.g. POWERUP_ONDELAY_VALUE might stay uninitialized by CCU and/or device itself, in which case it may be reported with an invalid value

Fixes the issue reported [here](https://github.com/openhab/openhab-addons/pull/12192#issuecomment-1080436944)

### DIFF
--- a/bundles/org.openhab.binding.homematic/src/main/java/org/openhab/binding/homematic/internal/handler/HomematicThingHandler.java
+++ b/bundles/org.openhab.binding.homematic/src/main/java/org/openhab/binding/homematic/internal/handler/HomematicThingHandler.java
@@ -26,6 +26,7 @@ import java.util.Map.Entry;
 import java.util.Objects;
 import java.util.concurrent.Future;
 
+import org.eclipse.jdt.annotation.Nullable;
 import org.openhab.binding.homematic.internal.HomematicBindingConstants;
 import org.openhab.binding.homematic.internal.common.HomematicConfig;
 import org.openhab.binding.homematic.internal.communicator.HomematicGateway;
@@ -419,7 +420,7 @@ public class HomematicThingHandler extends BaseThingHandler {
         }
     }
 
-    private Object getValueForConfiguration(HmDatapoint dp) {
+    private @Nullable Object getValueForConfiguration(HmDatapoint dp) {
         if (dp.getValue() == null) {
             return null;
         }
@@ -434,12 +435,12 @@ public class HomematicThingHandler extends BaseThingHandler {
             final boolean minValid, maxValid;
             if (dp.isFloatType()) {
                 Double numValue = dp.getDoubleValue();
-                minValid = dp.getMinValue() == null || numValue >= dp.getMaxValue().doubleValue();
-                maxValid = dp.getMinValue() == null || numValue <= dp.getMinValue().doubleValue();
+                minValid = dp.getMinValue() == null || numValue >= dp.getMinValue().doubleValue();
+                maxValid = dp.getMaxValue() == null || numValue <= dp.getMaxValue().doubleValue();
             } else {
                 Integer numValue = dp.getIntegerValue();
-                minValid = dp.getMinValue() == null || numValue >= dp.getMaxValue().intValue();
-                maxValid = dp.getMinValue() == null || numValue <= dp.getMinValue().intValue();
+                minValid = dp.getMinValue() == null || numValue >= dp.getMinValue().intValue();
+                maxValid = dp.getMaxValue() == null || numValue <= dp.getMaxValue().intValue();
             }
             if (minValid && maxValid) {
                 return dp.getValue();
--- a/bundles/org.openhab.binding.homematic/src/main/java/org/openhab/binding/homematic/internal/handler/HomematicThingHandler.java
+++ b/bundles/org.openhab.binding.homematic/src/main/java/org/openhab/binding/homematic/internal/handler/HomematicThingHandler.java
@@ -140,8 +140,7 @@ public class HomematicThingHandler extends BaseThingHandler {
             loadHomematicChannelValues(channel);
             for (HmDatapoint dp : channel.getDatapoints()) {
                 if (dp.getParamsetType() == HmParamsetType.MASTER) {
-                    config.put(MetadataUtils.getParameterName(dp),
-                            dp.isEnumType() ? dp.getOptionValue() : dp.getValue());
+                    config.put(MetadataUtils.getParameterName(dp), getValueForConfiguration(dp));
                 }
             }
         }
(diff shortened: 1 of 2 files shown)


maniac103:
@fwolter @MHerbst Please review as well ;-)

bundles/org.openhab.binding.homematic/src/main/java/org/openhab/binding/homematic/internal/handler/HomematicThingHandler.java
```diff
@@ -420,6 +419,37 @@ protected void updateDatapointState(HmDatapoint dp) {
...
+        if (dp.isNumberType()) {
+            // For number datapoints that are only used depending on the value of other datapoints,
+            // the CCU may return invalid (out of range) values if the datapoint currently is not in use.
+            // Make sure to not invalidate the whole configuration by returning the datapoint's default
+            // value in that case.
+            final boolean minValid, maxValid;
+            if (dp.isFloatType()) {
+                Double numValue = dp.getDoubleValue();
+                minValid = dp.getMinValue() == null || numValue >= dp.getMaxValue().doubleValue();
+                maxValid = dp.getMinValue() == null || numValue <= dp.getMinValue().doubleValue();
```
lolodomo, line 438: It looks to me wrong: the first should I believe consider dp.getMinValue() while the second should consider dp.getMaxValue(). WDYT ?
Same remark in the else branch

bundles/org.openhab.binding.homematic/src/main/java/org/openhab/binding/homematic/internal/handler/HomematicThingHandler.java
```diff
@@ -140,8 +140,7 @@ private void doInitializeInBackground() throws GatewayNotAvailableException, Hom
             loadHomematicChannelValues(channel);
             for (HmDatapoint dp : channel.getDatapoints()) {
                 if (dp.getParamsetType() == HmParamsetType.MASTER) {
-                    config.put(MetadataUtils.getParameterName(dp),
-                            dp.isEnumType() ? dp.getOptionValue() : dp.getValue());
+                    config.put(MetadataUtils.getParameterName(dp), getValueForConfiguration(dp));
```
lolodomo, line 143: As getValueForConfiguration can retrun null, I am not sure whether this config entry should be set to null, or not set at all ?
I am even not sure if this is valid to set a null value for a config parameter.

bundles/org.openhab.binding.homematic/src/main/java/org/openhab/binding/homematic/internal/model/HmDatapoint.java
```diff
@@ -150,19 +150,34 @@ public int getOptionIndex(String option) {
      */
     public String getOptionValue() {
         if (options != null && value != null) {
-            int idx = 0;
-            if (value instanceof Integer) {
-                idx = (int) value;
-            } else {
-                idx = Integer.parseInt(value.toString());
-            }
+            int idx = getIntegerValue();
```
lolodomo, line 153: As getIntegerValue can return null, there is probably a test to include here before casting it to an int.

bundles/org.openhab.binding.homematic/src/main/java/org/openhab/binding/homematic/internal/handler/HomematicThingHandler.java
```diff
@@ -420,6 +419,37 @@ protected void updateDatapointState(HmDatapoint dp) {
         }
     }
 
+    private Object getValueForConfiguration(HmDatapoint dp) {
```
lolodomo, line 422: Please add @Nullable to the return type.

bundles/org.openhab.binding.homematic/src/main/java/org/openhab/binding/homematic/internal/model/HmDatapoint.java
```diff
@@ -150,19 +150,34 @@ public int getOptionIndex(String option) {
...
             if (idx < options.length) {
                 return options[idx];
             }
         }
         return null;
     }
 
+    public Integer getIntegerValue() {
```
lolodomo, line 161: Please add @Nullable to the return type.

bundles/org.openhab.binding.homematic/src/main/java/org/openhab/binding/homematic/internal/model/HmDatapoint.java
```diff
@@ -150,19 +150,34 @@ public int getOptionIndex(String option) {
...
+        if (value instanceof Integer) {
+            return (int) value;
+        } else if (value != null) {
+            return Integer.parseInt(value.toString());
+        } else {
+            return null;
+        }
+    }
+
+    public Double getDoubleValue() {
```
lolodomo, line 171: Please add @Nullable to the return type.

lolodomo (changes requested):

bundles/org.openhab.binding.homematic/src/main/java/org/openhab/binding/homematic/internal/handler/HomematicThingHandler.java
maniac103: Yes, you're completely right. Not sure what I was thinking there, probably copy'n'paste gone wrong. Fixed now.

bundles/org.openhab.binding.homematic/src/main/java/org/openhab/binding/homematic/internal/model/HmDatapoint.java
maniac103: Added null check

bundles/org.openhab.binding.homematic/src/main/java/org/openhab/binding/homematic/internal/handler/HomematicThingHandler.java
maniac103: I've done so (same as in the occasions below), but I guess this binding will need a lot more work before being fully nullability-correct.

bundles/org.openhab.binding.homematic/src/main/java/org/openhab/binding/homematic/internal/handler/HomematicThingHandler.java
maniac103: The only case under which `getValueForConfiguration()` returns null is if the datapoint holds a `null` value. In that case I'd say that writing `null` into the config is correct.
The places I checked that care about the keys present in the map (those that call `containsKey()` and `keySet()` seem to be able to deal with null values being present. Not sure about e.g. the UI, but given the previous code put `null` values in this case (value being `null`) as well, I guess it should be fine.

lolodomo (approved):
LGTM

PR #12563, opened 2022-04-02
[nanoleaf] Fix LayoutTest failing on Windows

### LINKED ISSUE #12562: [nanoleaf] LayoutTest fails on Windows because of CRLF
The method org.openhab.binding.nanoleaf.internal.model.Layout#getLayoutView() adds System dependent line separators inside the resulting string 
` view += System.lineSeparator();`

which is on Windows environment CRLF

but the unit tests only accepts LF

`String layoutView = layout.getLayoutView();
        assertThat(layoutView, is(equalTo(
                "  186       42064       61141                   45187       33626       10523                   39163 \n"
                        + "                                                                                                      \n"
                        + "19209                   60147       36604       37121                   54086        3512       16398 \n")));`

Therefore, build of that module fails on Windows hosts.

Would suggest to replace the the static LF in the expected test result with `System.lineSeparator()`.

wborn:
On Windows the view has different line endings causing some tests to fail.
For readability the big strings have also been moved into files.

Fixes #12562


lolodomo (approved):
LGTM

PR #12565, opened 2022-04-02
[pidcontroller] Add ability to limit the I-part

fwolter:
@lnagel "tomorrow" is now :) Sorry for the delay.

Signed-off-by: Fabian Wolter <github@fabian-wolter.de>

### DIFF
--- a/bundles/org.openhab.automation.pidcontroller/src/main/java/org/openhab/automation/pidcontroller/internal/handler/PIDController.java
+++ b/bundles/org.openhab.automation.pidcontroller/src/main/java/org/openhab/automation/pidcontroller/internal/handler/PIDController.java
@@ -34,12 +34,17 @@ class PIDController {
     private double ki;
     private double kd;
     private double derivativeTimeConstantSec;
+    private double iMinValue;
+    private double iMaxValue;
 
-    public PIDController(double kpAdjuster, double kiAdjuster, double kdAdjuster, double derivativeTimeConstantSec) {
+    public PIDController(double kpAdjuster, double kiAdjuster, double kdAdjuster, double derivativeTimeConstantSec,
+            double iMinValue, double iMaxValue) {
         this.kp = kpAdjuster;
         this.ki = kiAdjuster;
         this.kd = kdAdjuster;
         this.derivativeTimeConstantSec = derivativeTimeConstantSec;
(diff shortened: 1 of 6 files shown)


bundles/org.openhab.automation.pidcontroller/src/main/java/org/openhab/automation/pidcontroller/internal/handler/PIDController.java
```diff
@@ -34,12 +34,25 @@
...
+
+        // prepare min/max for the integral result accumulator
+        if (Double.isFinite(kiAdjuster) && Math.abs(kiAdjuster) > 0.0) {
+            if (Double.isFinite(iMinValue)) {
+                this.iMinResult = iMinValue / kiAdjuster;
+            }
+            if (Double.isFinite(iMaxValue)) {
+                this.iMaxResult = iMaxValue / kiAdjuster;
+            }
+        }
```
lolodomo, line 55: Shouldn't you set iMinResult/iMaxResult to Double.NaN; by default, in case you did not enter in the different if ?

fwolter: I think you're right. If I see correctly the integral result will always be zero if you leave the upper limit empty. @lnagel did you test leaving the upper limit blank?

lnagel: Thanks for the remarks to both of you. I haven't been working with Java for 10 years (switched to Scala and Python) so it would seem that I forgot that there are default values for uninitialised class variables. I'll fix that in the code.

lnagel: https://github.com/fwolter/openhab-addons/pull/5

fwolter: Thank you very much you two! The code should be fine now.

lolodomo (approved):
LGTM

PR #12571, opened 2022-04-04
[marytts] Fix incorrect WAV header for some voices

dalgwen:
With this fix, MaryTTS creates the correct wav header for the audio inputstream (audioformat is now assigned before we use it in the method 'getWavHeaderInputStream')

This bug was detected [in a community post](https://community.openhab.org/t/pulseaudio-as-sink-is-playing-back-way-to-fast/129349/6), and the sound was playing too fast.

I made [a snapshot release](https://github.com/dalgwen/openhab-addons/releases/tag/3.3.0_SNAPSHOT-marytts_wav_header).


wborn (approved):
Thanks for fixing it!

PR #12572, opened 2022-04-04
Don't limit serial port options

lolodomo:
Related to openhab/openhab-core#2798


mlobstein (approved):
LGTM for benqprojector, epsonprojector, kaleidescape, monopriceaudio, nuvo & oppo

wborn:
I think it would also be better to remove the default from the herzborg config. So you consciously configure the port. That default port may not be available at all (especially on Windows).

wborn (approved):
Thanks!

PR #12583, opened 2022-04-06
[mielecloud] Fix washing machine can be started channel is not updated

### LINKED ISSUE #12500: [mielecloud] Channel CanBeStarted is sometimes not activated for washing machine
## Expected Behavior

The `CanBeStarted` channel should be set to `ON` when the washing machine is pre-programmed, mobile control is enabled, a delayed start time is set and the timer has been started.

## Current Behavior

Often the channel remains `OFF`.

## Possible Solution

The issue is caused by a combination of problems on the cloud side and how the value for the channel is fetched from the cloud. The field `"processAction"` from [`/devices/{deviceId}/actions`](https://www.miele.com/developer/swagger-ui/swagger.html#/Action/get_devices__deviceId__actions) is not immediately updated after the device changes its state and the binding fetches the not-yet-updated value. Because of the way action fetching is implemented the endpoint is only fetched again when the device changes it's state. This only happens when the timer for the delayed start runs out and the machine starts by itself or when the user interacts with it directly. Additionally, the live updates SSE endpoint does not send the same values as the actions endpoint. The value from the endpoint is the correct one.

As a workaround, always fetch the actions endpoint when an `actions` update arrives via SSE.

In the future it is reasonable to directly use the SSE payload instead of fetching the actions endpoint. This has to be delayed until the discrepancies between the SSE events and the actions endpoint have been resolved on the cloud side.

## Steps to Reproduce

1. Turn washing machine on
2. Pre-programm machine, enable mobile control, set a delayed start time
3. Start the delay timer
4. Check channel value in openHAB

## Context

Automatic early starting of delayed washing machine runs.

## Your Environment
* Version used: openHAB 3.2.0 with binding version 3.2.0
* Environment name and version: all

BjoernLange:
Closes #12500.


lolodomo (approved):
LGTM

PR #12586, opened 2022-04-07
[epsonprojector] Fix discovery service to prevent erroneous inbox items

mlobstein:
Fix an issue with the Epson binding creating erroneous inbox items when AMX discovery beacon packets from other devices (GlobalCache, etc) are received. The code was improved to ignore non Epson discovery beacon packets and will handle multiple Epson projectors on the network properly (thread safe).

First discovered in this post: https://community.openhab.org/t/is-the-epson-projector-binding-abandoned/107769/82

### DIFF
--- a/bundles/org.openhab.binding.epsonprojector/src/main/java/org/openhab/binding/epsonprojector/internal/discovery/MulticastListener.java
+++ b/bundles/org.openhab.binding.epsonprojector/src/main/java/org/openhab/binding/epsonprojector/internal/discovery/MulticastListener.java
@@ -12,6 +12,11 @@
  */
 package org.openhab.binding.epsonprojector.internal.discovery;
 
+import static org.openhab.binding.epsonprojector.internal.EpsonProjectorBindingConstants.DEFAULT_PORT;
+import static org.openhab.binding.epsonprojector.internal.EpsonProjectorBindingConstants.THING_PROPERTY_HOST;
+import static org.openhab.binding.epsonprojector.internal.EpsonProjectorBindingConstants.THING_PROPERTY_MAC;
+import static org.openhab.binding.epsonprojector.internal.EpsonProjectorBindingConstants.THING_PROPERTY_PORT;
+
 import java.io.IOException;
 import java.net.DatagramPacket;
 import java.net.InetAddress;
@@ -20,8 +25,11 @@ import java.net.NetworkInterface;
 import java.net.SocketException;
 import java.net.SocketTimeoutException;
 import java.nio.charset.StandardCharsets;
+import java.util.HashMap;
+import java.util.Map;
 
 import org.eclipse.jdt.annotation.NonNullByDefault;
+import org.eclipse.jdt.annotation.Nullable;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -37,10 +45,6 @@ public class MulticastListener {
     private final Logger logger = LoggerFactory.getLogger(MulticastListener.class);
 
     private MulticastSocket socket;
-
-    // Epson-specific properties defined in this binding
-    private String uid = "";
-    private String ipAddress = "";
 
     // Epson projector devices announce themselves on a multicast port
     private static final String EPSON_MULTICAST_GROUP = "239.255.250.250";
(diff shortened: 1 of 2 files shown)


bundles/org.openhab.binding.epsonprojector/src/main/java/org/openhab/binding/epsonprojector/internal/discovery/MulticastListener.java
```diff
@@ -103,47 +107,26 @@ public boolean waitForBeacon() throws IOException {
...
      * AMXB<-UUID=000048746B33><-SDKClass=VideoProjector><-GUID=EPSON_EMP001><-Revision=1.0.0>
      */
-    private void parseAnnouncementBeacon(DatagramPacket packet) {
+    private @Nullable Map<String, Object> parseAnnouncementBeacon(DatagramPacket packet) {
         String beacon = (new String(packet.getData(), StandardCharsets.UTF_8)).trim();
-
         logger.trace("Multicast listener parsing announcement packet: {}", beacon);
 
-        clearProperties();
+        if (beacon.contains("EPSON") && beacon.contains("VideoProjector")) {
```
mhilbush, line 114: Did you purposely remove the `toUpperCase` calls? I thought those were there in case a future projector changed the case of those items (which admittedly seems unlikely).

mhilbush (approved):
Other than my one question, this looks like a good solution.

bundles/org.openhab.binding.epsonprojector/src/main/java/org/openhab/binding/epsonprojector/internal/discovery/MulticastListener.java
mlobstein: This was something that I had on the back burner for a while. There was a problem with another binding where similar code had issues on systems that were using a non-english language locale. It seems a little safer to match the actual value. I might revisit this.

bundles/org.openhab.binding.epsonprojector/src/main/java/org/openhab/binding/epsonprojector/internal/discovery/MulticastListener.java
```diff
@@ -103,47 +108,26 @@ public boolean waitForBeacon() throws IOException {
...
-        for (String parameter : parameterList) {
-            String[] keyValue = parameter.split("=");
-
-            if (keyValue.length != 2) {
-                continue;
-            }
-
-            if (keyValue[0].contains("UUID")) {
-                uid = keyValue[1].substring(0, keyValue[1].length() - 1);
+                if (keyValue.length == 2 && keyValue[0].contains("UUID") && !"".equals(keyValue[1])) {
```
lolodomo, line 121: Please rather use isEmpty

lolodomo (changes requested):

bundles/org.openhab.binding.epsonprojector/src/main/java/org/openhab/binding/epsonprojector/internal/discovery/MulticastListener.java
mlobstein: done

bundles/org.openhab.binding.epsonprojector/src/main/java/org/openhab/binding/epsonprojector/internal/discovery/MulticastListener.java
lolodomo: Is it ok now? 
Why no more using toUpprrcase for VideoProjector?

mlobstein: It is not necessary. The string 'VideoProjector' is the given SDK class name and will always be capital case as defined by the whitepaper I found on the AMX beacon spec. The few other examples that I found for a projector beacon supported this. Here is the whitepaper, the device class names are on the last page: https://trade.amx.com/assets/whitePapers/AMX.Dynamic.Device.Discovery.White.Paper.pdf

lolodomo (approved):

PR #12587, opened 2022-04-08
[icloud] Fix ONLINE state when device response status is 203

### LINKED ISSUE #12501: [icloud] Devices report OFFLINE state while ONLINE
The iCloud binding shows OFFLINE state while devices are ONLINE.

After some investigation it seems that it happens because these devices report "deviceStatus" as 203. While in the binding, it only checks for "200" deviceStatus, all other state means OFFLINE.

It is not really clear what does "203" mean, because some devices which are online in icloud shows this state and also offline devices have this state.

## Your Environment

* openHAB 3.2.0

radokristof:
This minor patch fixes an issue when the device response status code is 203.

This seems to work, but I don't know if this is enough/acceptable. Devices reports 203 status sometimes when they are online, sometimes when they are offline. However there seems to be another status code (201) for offline devices.

I couldn't find any info online what these status codes mean, nor any other reported value which can indicate the real status of an iCloud device.

Closes #12501


lolodomo:
@pgfeller : is it ok for you?

lolodomo:
HTTP code 203 means the server response was updated by a proxy. This is not an error. 
That being said, I have no idea if this is applicable in iCloud context.

radokristof:
@lolodomo yes if you treat it as an http response code then yes.

lolodomo:
@radokristof : as there is no feedback from the binding maintainer, let's go with the merge of your change.

lolodomo (approved):

PR #12594, opened 2022-04-09
[http] Support all 2xx response codes

### LINKED ISSUE #11369: [http] failed calls in logging that are actually successful
## Expected Behavior
HTTP calls that have a 20x response should be considered successful 

## Current Behavior
Only HTTP calls that have a 200 response are considered successful. All other 20x responses (so 201 Created, 204 No Content, etc) are marked as failed by the binding. This results in a lot of incorrect warning lines in the log.  

## Possible Solution

Change [this line](https://github.com/openhab/openhab-addons/blob/main/bundles/org.openhab.binding.http/src/main/java/org/openhab/binding/http/internal/http/HttpResponseListener.java#L67) to use `response.getStatus().isSuccess()`

1. Do a http call that returns a 201, 204 or other non-200 response
2. See this line appear in the OH log: `[WARN ] [p.internal.http.HttpResponseListener] - Requesting '<url>' (method='POST', content='org.eclipse.jetty.client.util.StringContentProvider@704d4f0e') failed: 201 Created`

## Context
I'm calling the [DSMR-reader](https://github.com/dsmrreader/dsmr-reader) REST api  from OH. This endpoint returns 201 created. 

Workaround for now: `log:set ERROR org.openhab.binding.http.internal.http.HttpResponseListener`

## Your Environment
openhab 3.2.0.M3

corubba:
Until now, the http binding did only regard the `HTTP/200 OK` response code as successful, all the other 2xx codes were treated as failed. This behaviour made it impossible to work with non-trivial REST APIs that may return "HTTP/201 Created" or "HTTP/204 No Content".

I thought about simply extending the switch-case with all to other 2xx constant from `HttpStatus`, but that felt cumbersome. I am not really happy that the switch now only has one case and a default left, but I kept it this way (instead of dissolving it into the outer `if-elseif-else`) to keep the possibility for future special-handling of specific return codes.

I also removed that content null check, because `getContent()` never returns null; it always returns a byte-array (which may has the length 0).

Fixes #11369

### DIFF
--- a/bundles/org.openhab.binding.http/src/main/java/org/openhab/binding/http/internal/http/HttpResponseListener.java
+++ b/bundles/org.openhab.binding.http/src/main/java/org/openhab/binding/http/internal/http/HttpResponseListener.java
@@ -13,6 +13,7 @@
 package org.openhab.binding.http.internal.http;
 
 import java.nio.charset.StandardCharsets;
+import java.util.Objects;
 import java.util.concurrent.CompletableFuture;
 import java.util.stream.Collectors;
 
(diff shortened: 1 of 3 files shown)


lolodomo (commented):
Code LGTM
But maybe you should add an explanation in README ?

corubba:
Can do, but I am not really sure what to write/document. From my point of view it is expected behaviour that a http client (which this basically acts as) follows the semantics laid out by [rfc7231](https://datatracker.ietf.org/doc/html/rfc7231#section-6.3), and does not print a "Request failed" warning when the server sends a response clearly indicating success.

lolodomo:
Ok, this might be not necessary.

lolodomo (approved):
LGTM

PR #12603, opened 2022-04-11
[tesla] Remove synchronization to avoid deadlocks

kaikreuzer:
I just experienced a deadlock situation due to this synchronization here.
The `BaseThingHandler` does a synchronize on `this` within `updateState()` - this means that handler implementations must not do the same, if they also use other kinds of locks. 

Since the method in question is only called at a define interval from a job within the binding, the synchronization is not necessary at this place. Therefore removing it should fix the deadlock situations.


lolodomo (approved):
LGTM

PR #12607, opened 2022-04-15
[sonos] Support added for models Five and Roam SL

### LINKED ISSUE #12596: [Sonos] Device Type "Roam SL" is not recognized (invalid Character in Device ID)
## Expected Behavior
Device is found in a scan and can be added

## Current Behavior
Error in Log:

"java.lang.IllegalArgumentException: ID segment 'Roam SL' contains invalid characters. Each segment of the ID must match the pattern [\w-]*."

Observed in OpenHAB 3.20 Release Version

## Steps to Reproduce

1. Add Sonos Roam SL to Network
2. Error can be seen in Log

## Your Environment
OpenHAB 3.20 Release Version 
Running on a Raspberry Pi4/openhabian

### LINKED ISSUE #12610: [sonos] Support for Sonos Five should be added
Sonos Five is certainly not yet supported by the binding.
Is there someone with such device that could confirm ?

lolodomo:
Fix #12596
Fix #12610

### DIFF
--- a/bundles/org.openhab.binding.sonos/src/main/java/org/openhab/binding/sonos/internal/discovery/ZonePlayerDiscoveryParticipant.java
+++ b/bundles/org.openhab.binding.sonos/src/main/java/org/openhab/binding/sonos/internal/discovery/ZonePlayerDiscoveryParticipant.java
@@ -95,6 +95,9 @@ public class ZonePlayerDiscoveryParticipant implements UpnpDiscoveryParticipant 
                     case "Arc SL":
                         modelName = "ArcSL";
                         break;
+                    case "Roam SL":
+                        modelName = "RoamSL";
+                        break;
                     case "Sub":
                         // The Sonos Sub is ignored
                         ignored = true;
--- a/bundles/org.openhab.binding.sonos/README.md
+++ b/bundles/org.openhab.binding.sonos/README.md
@@ -11,11 +11,11 @@ This happens because of a TTL=1 for ALIVE packets send by Sonos devices, resulti
 ## Supported Things
 
 All available Sonos (playback) devices are supported by this binding.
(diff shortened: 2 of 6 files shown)


bundles/org.openhab.binding.sonos/src/main/java/org/openhab/binding/sonos/internal/discovery/ZonePlayerDiscoveryParticipant.java
```diff
@@ -95,6 +95,9 @@ public Set<ThingTypeUID> getSupportedThingTypeUIDs() {
                     case "Arc SL":
                         modelName = "ArcSL";
                         break;
+                    case "Roam SL":
+                        modelName = "RoamSL";
```
jlaur, line 99: Perhaps it would make sense as `default` either to strip invalid characters or set `ignored = true` if such characters are present. Otherwise we'll get `IllegalArgumentException` anyway when trying to create the ThingTypeUID. This could also be logged. What do you think?

jlaur (changes requested):
Thanks, looks good! One suggestion added in order to avoid the exception mentioned in the linked issue in the future.

bundles/org.openhab.binding.sonos/src/main/java/org/openhab/binding/sonos/internal/discovery/ZonePlayerDiscoveryParticipant.java
jlaur: As reading PR's in sequence I hadn't seen #12609 when writing this comment. Please disregard. :)

jlaur (approved):
LGTM

PR #12609, opened 2022-04-15
[sonos] Discovery of unsupported models without exception

lolodomo:
Transform the found model name into a valid thing type ID  when trying 
to match a thing type

This could avoid an `IllegalArgumentException` thrown by the discovery
service when new models will be introduced on the market.

Should be compatible with Bubble UPnP Server which modifies the model
name

A message is now logged for unsupported Sonos models.

### DIFF
--- a/bundles/org.openhab.binding.sonos/src/main/java/org/openhab/binding/sonos/internal/SonosXMLParser.java
+++ b/bundles/org.openhab.binding.sonos/src/main/java/org/openhab/binding/sonos/internal/SonosXMLParser.java
@@ -1014,26 +1014,41 @@ public class SonosXMLParser {
     }
 
     /**
-     * The model name provided by upnp is formated like in the example form "Sonos PLAY:1" or "Sonos PLAYBAR"
+     * Build a valid thing type ID from the model name provided by UPnP
      *
-     * @param sonosModelName Sonos model name provided via upnp device
-     * @return the extracted players model name without column (:) character used for ThingType creation
+     * @param sonosModelName Sonos model name provided via UPnP device
+     * @return a valid thing type ID that can then be used for ThingType creation
      */
-    public static String extractModelName(String sonosModelName) {
-        String ret = sonosModelName;
+    public static String buildThingTypeIdFromModelName(String sonosModelName) {
+        // For Ikea SYMFONISK models, the model name now starts with "SYMFONISK" with recent firmwares
+        if (sonosModelName.toUpperCase().contains("SYMFONISK")) {
+            return "SYMFONISK";
+        }
+        String id = sonosModelName;
         // Remove until the first space (in practice, it removes the leading "Sonos " from the model name)
-        Matcher matcher = Pattern.compile("\\s(.*)").matcher(ret);
+        Matcher matcher = Pattern.compile("\\s(.*)").matcher(id);
         if (matcher.find()) {
-            ret = matcher.group(1);
+            id = matcher.group(1);
             // Remove a potential ending text surrounded with parenthesis
-            matcher = Pattern.compile("(.*)\\s\\(.*\\)").matcher(ret);
+            matcher = Pattern.compile("(.*)\\s\\(.*\\)").matcher(id);
             if (matcher.find()) {
-                ret = matcher.group(1);
+                id = matcher.group(1);
             }
         }
-        // Finally remove any : and any space
-        ret = ret.replaceAll(":", "").replaceAll(" ", "");
-        return ret;
+        // Finally remove unexpected characters in a thing type ID
+        id = id.replaceAll("[^a-zA-Z0-9_]", "");
+        // ZP80 is translated to CONNECT and ZP100 to CONNECTAMP
+        switch (id) {
+            case "ZP80":
+                id = "CONNECT";
+                break;
+            case "ZP100":
+                id = "CONNECTAMP";
+                break;
+            default:
+                break;
+        }
+        return id;
     }
 
     public static String compileMetadataString(SonosEntry entry) {
--- a/bundles/org.openhab.binding.sonos/src/main/java/org/openhab/binding/sonos/internal/SonosXMLParser.java
+++ b/bundles/org.openhab.binding.sonos/src/main/java/org/openhab/binding/sonos/internal/SonosXMLParser.java
@@ -1031,9 +1031,8 @@ public class SonosXMLParser {
                 ret = matcher.group(1);
             }
         }
-        if (ret.contains(":")) {
-            ret = ret.replace(":", "");
-        }
+        // Finally remove any : and any space
+        ret = ret.replaceAll(":", "").replaceAll(" ", "");
         return ret;
     }
 
--- a/bundles/org.openhab.binding.sonos/src/main/java/org/openhab/binding/sonos/internal/SonosXMLParser.java
+++ b/bundles/org.openhab.binding.sonos/src/main/java/org/openhab/binding/sonos/internal/SonosXMLParser.java
@@ -1021,7 +1021,7 @@ public class SonosXMLParser {
      */
     public static String extractModelName(String sonosModelName) {
         String ret = sonosModelName;
-        // Remove until the first space (in practice, it removes the leading "Sonos " form the model name)
+        // Remove until the first space (in practice, it removes the leading "Sonos " from the model name)
         Matcher matcher = Pattern.compile("\\s(.*)").matcher(ret);
         if (matcher.find()) {
             ret = matcher.group(1);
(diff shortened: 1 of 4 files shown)


bundles/org.openhab.binding.sonos/src/main/java/org/openhab/binding/sonos/internal/SonosXMLParser.java
```diff
@@ -1021,9 +1021,15 @@ public void characters(char @Nullable [] ch, int start, int length) throws SAXEx
      */
     public static String extractModelName(String sonosModelName) {
         String ret = sonosModelName;
+        // Remove until the first space (in practice, it removes the leading "Sonos " form the model name)
```
jlaur, line 1024:
```suggestion
        // Remove until the first space (in practice, it removes the leading "Sonos " from the model name)
```

bundles/org.openhab.binding.sonos/src/main/java/org/openhab/binding/sonos/internal/SonosXMLParser.java
```diff
@@ -1021,9 +1021,15 @@ public void characters(char @Nullable [] ch, int start, int length) throws SAXEx
...
         Matcher matcher = Pattern.compile("\\s(.*)").matcher(ret);
         if (matcher.find()) {
             ret = matcher.group(1);
+            // Remove a potential ending text surrounded with parenthesis
+            matcher = Pattern.compile("(.*)\\s\\(.*\\)").matcher(ret);
+            if (matcher.find()) {
+                ret = matcher.group(1);
+            }
```
jlaur, line 1032: This method might be good candidate for some unit test coverage which would clearly document how it works.

bundles/org.openhab.binding.sonos/src/main/java/org/openhab/binding/sonos/internal/discovery/ZonePlayerDiscoveryParticipant.java
```diff
@@ -100,22 +100,26 @@ public Set<ThingTypeUID> getSupportedThingTypeUIDs() {
                         ignored = true;
                         break;
                     default:
+                        modelName = modelName.replaceAll("[^a-zA-Z0-9_]", "_");
```
jlaur, line 103: This seems inconsistent with the current hardcoded translations where spaces are simply stripped, for example "Arc SL" -> "ArcSL" while this would produce "Arc_SL" instead. If this would be consistent, cases for "One SL" and "Arc SL" could be dropped.

jlaur (changes requested):
Thanks! A few minor suggestions added.

bundles/org.openhab.binding.sonos/src/main/java/org/openhab/binding/sonos/internal/discovery/ZonePlayerDiscoveryParticipant.java
lolodomo: All those for which a space is expected (like "Arc SL", 'One SL" are handled in switch cases ; the default handles all other cases and space is then not expected.

One solution to avoid the special cases for "Arc SL" or "One SL" would be to rename thing type ArcSL into Arc_SL but this is a breaking change so I do not want to do it and I prefer to do it that way.

So for me, the new code in the default case is OK.

bundles/org.openhab.binding.sonos/src/main/java/org/openhab/binding/sonos/internal/SonosXMLParser.java
lolodomo: Typo fixed.

bundles/org.openhab.binding.sonos/src/main/java/org/openhab/binding/sonos/internal/SonosXMLParser.java
lolodomo: Unit tests added

bundles/org.openhab.binding.sonos/src/main/java/org/openhab/binding/sonos/internal/discovery/ZonePlayerDiscoveryParticipant.java
jlaur: Okay, let me try to explain. It seems that the "standard" translation to thing type is by removing spaces, for example "Arc SL" to "ArcSL". But now default is to replace them by underscores. If this would be harmonized you wouldn't need the two special cases for "One SL" and "Arc SL" as this would be correctly handled by the default case.

If a new "Arc XY" would be introduced, would you add another case for it being translated to "ArcXY" or would you leave it for the default case to translate it to "Arc_XY"? If you prefer underscores for future additions and "Arc SL" and "One SL" are only legacy translations, then current code is fine.

lolodomo: Yes, exactly, I don't want to break existing setup by changing thing types.
In the future, either I will add a new specific case in the discovery code or I will use the _ in the thing type. Adding one _ in future thing types would be the best option I believe.

lolodomo: Ok, I understand what you would like. You would like that I replace non standard characters by "" rather than by "_".
I think you are right, iut is the best option.

lolodomo: Ok I think it is now how you expect.

jlaur: Exactly.

bundles/org.openhab.binding.sonos/src/test/java/org/openhab/binding/sonos/internal/SonosXMLParserTest.java
```diff
@@ -0,0 +1,34 @@
...
+ */
+@NonNullByDefault
+public class SonosXMLParserTest {
+
+    @Test
+    public void extractModelName() {
+        assertEquals("Move", SonosXMLParser.extractModelName("Sonos Move"));
+        assertEquals("Roam SL", SonosXMLParser.extractModelName("Sonos Roam SL"));
+        assertEquals("Five", SonosXMLParser.extractModelName("Sonos Five (OpenHome)"));
+        assertEquals("One SL", SonosXMLParser.extractModelName("Sonos One SL (OpenHome)"));
```
jlaur, line 32: Now getting really picky/detailed, but consider if you want also to add a case for the removing of ":", so test coverage would be complete. :-) I'm ready to merge as-is since you already provided coverage for your changes and then some, so in much better condition now.

lolodomo: I forget the case of PLAY:5.
I am adding it just now...

lolodomo: Done

jlaur: Perfect!

jlaur (approved):
Thanks, looks good!

lolodomo:
Wait a second. To have the migration code working well in the thing handler, `extractModelName` should also remove the space.
I am going to make a little change.

lolodomo:
This time, it is OK, you can merge.

bundles/org.openhab.binding.sonos/src/main/java/org/openhab/binding/sonos/internal/discovery/ZonePlayerDiscoveryParticipant.java
```diff
@@ -89,36 +89,31 @@ public Set<ThingTypeUID> getSupportedThingTypeUIDs() {
...
-                        break;
-                    case "Roam SL":
-                        modelName = "RoamSL";
-                        break;
                     case "Sub":
                         // The Sonos Sub is ignored
                         ignored = true;
                         break;
                     default:
+                        modelName = modelName.replaceAll("[^a-zA-Z0-9_]", "");
```
jlaur, line 97: Now that you are at it, two more ideas: Since` SonosXMLParser.extractModelName` now replaces ":" and " " in a similar fashion, maybe this logic could be entirely moved there instead of having three layers of model interpretation/translation?

And second, logic method `getModelName()` in this class also applies some logic regarding SYMFONISK. This could also be moved to `extractModelName()` and covered with tests.

lolodomo: Done

jlaur: Excellent!

PR #12611, opened 2022-04-15
[evcc] Initial contribution

florian-h05:
# New binding: `evcc` - electric vehicle charging control

# Description

evcc controls your wallbox(es) with multiple charging modes and allows you to charge your ev with your photovoltaik's excess current. To provide an intelligent charging control, evcc supports over 30 wallboxes and over 20 energy meters/home energy managment systems from many manufacturers as well as electric vehicles from over 20 car manufacturers. Furthermore, evcc calculates your money savings.

This binding enables openHAB to retrieve status data from your evcc installation and to control the charging process. For more advanced features like calculated savings, you have to visit the web UI of evcc.

This binding is currently available at the marketplace: https://community.openhab.org/t/evcc-binding-electric-vehicle-charging-control/135209

# Testing

Testing can be performed with the evcc demo available at https://demo.evcc.io.

Binding JAR at Jenkins: https://openhab.jfrog.io/artifactory/libs-pullrequest-local/org/openhab/addons/bundles/org.openhab.binding.evcc/3.3.0-SNAPSHOT/org.openhab.binding.evcc-3.3.0-SNAPSHOT.jar


bundles/org.openhab.binding.evcc/src/main/java/org/openhab/binding/evcc/internal/EvccHandler.java
```diff
@@ -0,0 +1,547 @@
...
+            updateStatus(ThingStatus.ONLINE);
+            batteryConfigured = status.getResult().getBatteryConfigured();
+            gridConfigured = status.getResult().getGridConfigured();
+            pvConfigured = status.getResult().getPvConfigured();
+            updateChannelsGeneral();
+            for (int i = 0; i < numberOfLoadpoints; i++) {
+                updateChannelsLoadpoint(i);
+            }
+        } catch (Exception e) {
+            updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.OFFLINE.CONFIGURATION_ERROR,
```
jlaur, line 193: This seems more like `COMMUNICATION_ERROR` if related to connection failure as the cause text indicates. However, catching raw `Exception` could indicate that you may not know what actually went wrong. So I would suggest also to catch only needed, specific exception types here.

bundles/org.openhab.binding.evcc/src/main/java/org/openhab/binding/evcc/internal/EvccHandler.java
```diff
@@ -0,0 +1,547 @@
...
+        // "Could not control device at IP address x.x.x.x");
+    }
+
+    @Override
+    public void initialize() {
+        config = getConfigAs(EvccConfiguration.class);
+        updateStatus(ThingStatus.UNKNOWN);
+
+        if ("".equals(config.url)) {
+            updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.OFFLINE.CONFIGURATION_ERROR, "No host configured");
```
jlaur, line 153: Please provide this text in the i18n file **evcc.properties** so it can be localized.

bundles/org.openhab.binding.evcc/src/main/java/org/openhab/binding/evcc/internal/EvccHandler.java
```diff
@@ -0,0 +1,547 @@
...
+                try {
+                    sitename = status.getResult().getSiteTitle();
+                    numberOfLoadpoints = status.getResult().getLoadpoints().length;
+                    logger.debug("Found {} loadpoints on site {} (host: {}).", numberOfLoadpoints, sitename,
+                            config.url);
+                    updateStatus(ThingStatus.ONLINE);
+                    refreshOnStartup();
+                } catch (Exception e) {
+                    updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.OFFLINE.CONFIGURATION_ERROR,
+                            "Failed to connect to evcc: " + e);
```
jlaur, line 167: Please provide this text in the i18n file **evcc.properties** so it can be localized.

bundles/org.openhab.binding.evcc/src/main/java/org/openhab/binding/evcc/internal/EvccHandler.java
```diff
@@ -0,0 +1,547 @@
...
+     * @return the response body or response_code 999 if request faild
+     */
+    private String httpRequest(@Nullable String description, String url, String method) {
+        String response = "";
+        try {
+            response = HttpUtil.executeUrl(method, url, LONG_CONNECTION_TIMEOUT_MILLISEC);
+            logger.trace("{} - {}, {} - {}", description, url, method, response);
+            return response;
+        } catch (IOException e) {
+            logger.warn("IO Exception - {} - {}, {} - {}", description, url, method, e.toString());
```
jlaur, line 485: This should probably be reduced to **debug** level - see https://www.openhab.org/docs/developer/guidelines.html#f-logging.

bundles/org.openhab.binding.evcc/src/main/java/org/openhab/binding/evcc/internal/EvccHandler.java
```diff
@@ -0,0 +1,547 @@
...
+        }
+    }
+
+    /**
+     * Make a HTTP request.
+     * 
+     * @param description request description for logger
+     * @param url full request URL
+     * @param method reguest method, e.g. GET, POST
+     * @return the response body or response_code 999 if request faild
```
jlaur, line 476:
```suggestion
     * @return the response body or response_code 999 if request failed
```

bundles/org.openhab.binding.evcc/src/main/java/org/openhab/binding/evcc/internal/EvccHandler.java
```diff
@@ -0,0 +1,547 @@
...
+        if (command.equals(RefreshType.REFRESH)) {
+            refresh();
+        } else {
+            logger.debug("Handling command {} for channel {}", command, channelUID);
+            String channelIdWithoutGroup = channelUID.getIdWithoutGroup();
+            int loadpoint = Integer.parseInt(channelUID.getGroupId().toString().substring(9));
+            targetSoC = status.getResult().getLoadpoints()[loadpoint].getTargetSoC();
+            switch (channelIdWithoutGroup) {
+                case CHANNEL_LOADPOINT_MODE:
+                    setMode(config.url, loadpoint, command.toString());
```
jlaur, line 95: Here and for all other set methods: Error handling is missing. The set methods returns a `@Nullable String` received from `httpRequest()`. In case of failed calls, this string appears to have the content "{\"response_code\":\"999\"}". Perhaps it would be simpler to throw an exception in this case, which could then be handled?

bundles/org.openhab.binding.evcc/src/main/java/org/openhab/binding/evcc/internal/EvccHandler.java
```diff
@@ -0,0 +1,547 @@
...
+    // API calls to evcc
+    /**
+     * Get the status from evcc.
+     * 
+     * @param host hostname of IP address of the evcc instance
+     * @return Status object or null if request failed
+     */
+    private @Nullable Status getStatus(@Nullable String host) {
+        final String response = httpRequest("Status", host + EVCC_REST_API + "state", "GET");
+        return gson.fromJson(response, Status.class);
```
jlaur, line 500: This call can throw `JsonSyntaxException` which appears to be unhandled. In usage within `refresh()`, method is called outside of any `try`.

jlaur (changes requested):
Thanks for your contribution! I have posted some feedback after my first code reading.

bundles/org.openhab.binding.evcc/src/main/java/org/openhab/binding/evcc/internal/EvccHandler.java
florian-h05: I will handle `JsonSyntaxException` directly in `getStatus`.

bundles/org.openhab.binding.evcc/src/main/java/org/openhab/binding/evcc/internal/EvccHandler.java
florian-h05: I will return `null` if request failed and therefore set methods would also return `null` which then will be handled.

bundles/org.openhab.binding.evcc/src/main/java/org/openhab/binding/evcc/internal/EvccHandler.java
```diff
@@ -0,0 +1,547 @@
...
+        String targetTime = loadpoint.getTargetTime();
+        ZonedDateTime newTargetTimeZDT = ZonedDateTime.now().plusSeconds(30);
+        try {
+            Instant instant = Instant.parse(targetTime);
+            newTargetTimeZDT = ZonedDateTime.ofInstant(instant, ZoneId.systemDefault());
+        } catch (Exception e) {
+            try {
+                newTargetTimeZDT = ZonedDateTime.parse(targetTime);
+            } catch (Exception f) {
+                logger.debug("Failed parsing targetTime {}. Error: {}.", targetTime, f.toString());
```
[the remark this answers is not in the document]
florian-h05: The problem is, that due to the lack of API docs which describe the format of targetTime, I had to try out and I got multiple strings back. I just checked the API again and now it seems like the following:

When targetTime is set, a String like `2022-05-09T07:00:00+02:00` is returned, if no time is set the string `null` is returned.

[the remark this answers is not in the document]
florian-h05: Which looks like ISO-8601 calendar system so I can use `ZonedDateTime` now.

bundles/org.openhab.binding.evcc/README.md
```diff
@@ -0,0 +1,175 @@
+# evcc Binding
+
+This binding integrates [evcc - electric vehicle charging control](https://evcc.io), a project that provides a control center for electric vehicle charging.
+
+evcc controls your wallbox(es) with multiple charging modes and allows you to charge your ev with your photovoltaik's excess current.
+To provide an intelligent charging control, evcc supports over 30 wallboxes and over 20 energy meters/home energy managment systems from many manufacturers as well as electric vehicles from over 20 car manufacturers.
```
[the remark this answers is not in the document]
jlaur: Please resolve comments after pushing commits to the branch. This will ease the process of verifying comment resolution.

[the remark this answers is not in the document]
florian-h05: Okay, I have unresolved all comments

bundles/org.openhab.binding.evcc/src/main/java/org/openhab/binding/evcc/internal/EvccHandler.java
```diff
@@ -0,0 +1,547 @@
...
+                status = getStatus(config.url);
+                try {
+                    sitename = status.getResult().getSiteTitle();
+                    numberOfLoadpoints = status.getResult().getLoadpoints().length;
+                    logger.debug("Found {} loadpoints on site {} (host: {}).", numberOfLoadpoints, sitename,
+                            config.url);
+                    updateStatus(ThingStatus.ONLINE);
+                    refreshOnStartup();
+                } catch (Exception e) {
+                    updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.OFFLINE.CONFIGURATION_ERROR,
```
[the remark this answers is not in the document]
florian-h05: I solved this by using `COMMUNICATION_ERROR` When the request to evcc failed, on any other failure I will use `HANDLER_INITIALIZING_ERROR`.

bundles/org.openhab.binding.evcc/src/main/java/org/openhab/binding/evcc/internal/EvccHandler.java
florian-h05: How would this exactly work? Could you please provide an example?

bundles/org.openhab.binding.evcc/src/main/java/org/openhab/binding/evcc/internal/EvccHandler.java
```diff
@@ -0,0 +1,547 @@
...
+         * }
+         */
+        targetSoC = loadpoint.getTargetSoC();
+        channel = new ChannelUID(getThing().getUID(), loadpointName, CHANNEL_LOADPOINT_TARGET_SOC);
+        updateState(channel, new QuantityType<>(targetSoC, Units.PERCENT));
+        String targetTime = loadpoint.getTargetTime();
+        ZonedDateTime newTargetTimeZDT = ZonedDateTime.now().plusSeconds(30);
+        try {
+            Instant instant = Instant.parse(targetTime);
+            newTargetTimeZDT = ZonedDateTime.ofInstant(instant, ZoneId.systemDefault());
```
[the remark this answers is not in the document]
florian-h05: Not needed anymore.

bundles/org.openhab.binding.evcc/src/main/java/org/openhab/binding/evcc/internal/EvccHandler.java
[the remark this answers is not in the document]
florian-h05: Fixed with a71486a

bundles/org.openhab.binding.evcc/src/main/java/org/openhab/binding/evcc/internal/EvccHandler.java
florian-h05: Fixed by catching generic communication error with `null` checking, I have preserved this for any other non-communication error (I could not really imagine any for now).

bundles/org.openhab.binding.evcc/src/main/java/org/openhab/binding/evcc/internal/EvccHandler.java
jlaur: Code:
```
updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.OFFLINE.CONFIGURATION_ERROR, "@text/offline.configuration-error.no-host");
```

Properties:
```
offline.configuration-error.no-host = No host configured
```

bundles/org.openhab.binding.evcc/src/main/java/org/openhab/binding/evcc/internal/EvccHandler.java
```diff
@@ -150,8 +148,9 @@ public void initialize() {
                         updateStatus(ThingStatus.ONLINE);
                         refreshOnStartup();
                     } catch (Exception e) {
-                        updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.OFFLINE.CONFIGURATION_ERROR,
-                                "Failed to connect to evcc: " + e);
+                        logger.debug("Initialization failed:", e);
+                        updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.HANDLER_INITIALIZING_ERROR,
```
jlaur, line 152: **HANDLER_INITIALIZING_ERROR** is not correct for status **OFFLINE**, this should only be set by the framework.

bundles/org.openhab.binding.evcc/src/main/java/org/openhab/binding/evcc/internal/EvccHandler.java
```diff
@@ -359,36 +363,22 @@ private void updateChannelsLoadpoint(int loadpointId) {
...
+        } else {
             try {
-                newTargetTimeZDT = ZonedDateTime.parse(targetTime);
-            } catch (Exception f) {
-                logger.debug("Failed parsing targetTime {}. Error: {}.", targetTime, f.toString());
+                ZonedDateTime targetTimeZDT = ZonedDateTime.parse(targetTime);
+                channel = new ChannelUID(getThing().getUID(), loadpointName, CHANNEL_LOADPOINT_TARGET_TIME);
+                updateState(channel, new DateTimeType(targetTimeZDT));
+
+            } catch (NullPointerException e) {
```
jlaur, line 376: `NullPointerException` should not be caught, it should be avoided. What can cause it here?

bundles/org.openhab.binding.evcc/src/main/java/org/openhab/binding/evcc/internal/EvccHandler.java
```diff
@@ -0,0 +1,547 @@
...
+            refresh();
+        } else {
+            logger.debug("Handling command {} for channel {}", command, channelUID);
+            String channelIdWithoutGroup = channelUID.getIdWithoutGroup();
+            int loadpoint = Integer.parseInt(channelUID.getGroupId().toString().substring(9));
+            targetSoC = status.getResult().getLoadpoints()[loadpoint].getTargetSoC();
+            switch (channelIdWithoutGroup) {
+                case CHANNEL_LOADPOINT_MODE:
+                    setMode(config.url, loadpoint, command.toString());
+                    refresh();
```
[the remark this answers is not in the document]
jlaur: What I meant here was simply to provide the `refresh()` at the very end of the method, but letting the `default` return from the method without calling `refresh()`. This way the call wouldn't be repeated for each `case`. It seems you by mistake removed the `break`'s, thus all set methods below the desired one are called.

bundles/org.openhab.binding.evcc/src/main/java/org/openhab/binding/evcc/internal/EvccHandler.java
florian-h05: As thus us not needed, I will completely remove it.

bundles/org.openhab.binding.evcc/src/main/java/org/openhab/binding/evcc/internal/EvccHandler.java
florian-h05: It can be caused when a String `null` is passed in to `ZonedDateTime.parse()`, I will avoid this by using `if (mysttring != null)`

bundles/org.openhab.binding.evcc/src/main/java/org/openhab/binding/evcc/internal/EvccHandler.java
```diff
@@ -0,0 +1,547 @@
...
+    }
+
+    @Override
+    public void handleCommand(ChannelUID channelUID, Command command) {
+        if (command.equals(RefreshType.REFRESH)) {
+            refresh();
+        } else {
+            logger.debug("Handling command {} for channel {}", command, channelUID);
+            String channelIdWithoutGroup = channelUID.getIdWithoutGroup();
+            int loadpoint = Integer.parseInt(channelUID.getGroupId().toString().substring(9));
```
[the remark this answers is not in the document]
florian-h05: How can I avoid that warnings??

For example here:
```java
private @Nullable Result result;

result = getResult(url); // returns @Nullable Result

if (result != null) {
  // do something
}
```
Gives a warning for potential null pointer access in the part `do something`.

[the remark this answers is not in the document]
jlaur: You can fix it by keeping a local reference, like this:
```
private @Nullable Result result;

this.result = getResult(url); // returns @Nullable Result
Result result = this.result;
if (result != null) {
  // do something
}
```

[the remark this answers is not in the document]
florian-h05: Compiler warnings should now be gone 😃

florian-h05:
@jlaur 

I think I made all requested changes, just tested it, works perfect.
Ready for re-review.

bundles/org.openhab.binding.evcc/src/main/java/org/openhab/binding/evcc/internal/EvccAPI.java
```diff
@@ -0,0 +1,130 @@
...
+
+import com.google.gson.Gson;
+import com.google.gson.JsonSyntaxException;
+
+/**
+ * The {@link EvccAPI} is responsible for API calls to evcc.
+ * 
+ * @author Florian Hotze - Initial contribution
+ */
+public class EvccAPI {
```
jlaur, line 37: Please annotate with `@NonNullByDefault`.

bundles/org.openhab.binding.evcc/src/main/java/org/openhab/binding/evcc/internal/EvccAPI.java
```diff
@@ -0,0 +1,130 @@
...
+    private static final Logger logger = LoggerFactory.getLogger(EvccAPI.class);
+    private static final Gson gson = new Gson();
+    private final @Nullable String host;
+
+    public EvccAPI(@Nullable String host) {
```
jlaur, line 42: Does it make sense to instantiate without a host, since it won't be able to do anything? The `@Nullable` annotation could be removed, and caller will need to handle configuration problems etc.

bundles/org.openhab.binding.evcc/src/main/java/org/openhab/binding/evcc/internal/EvccHandler.java
```diff
@@ -0,0 +1,443 @@
...
+                    break;
+                case CHANNEL_LOADPOINT_MIN_SOC:
+                    evccAPI.setMinSoC(loadpoint, Integer.parseInt(command.toString().replaceAll(" %", "")));
+                    break;
+                case CHANNEL_LOADPOINT_TARGET_SOC:
+                    evccAPI.setTargetSoC(loadpoint, Integer.parseInt(command.toString().replaceAll(" %", "")));
+                    break;
+                case CHANNEL_LOADPOINT_TARGET_TIME:
+                    if (targetTimeEnabled == true) {
+                        targetTimeZDT = new DateTimeType(command.toString()).getZonedDateTime();
```
jlaur, line 134: Hmm, nice trick. :) But if you know the format, it would be more efficient to provide the format string yourself and deal directly with `ZonedDateTime`. The `DateTimeType` string constructor use this private method for parsing:

https://github.com/openhab/openhab-core/blob/cabb3f73150d13f44d647f46af8362e396684b92/bundles/org.openhab.core/src/main/java/org/openhab/core/library/types/DateTimeType.java#L221-L236

bundles/org.openhab.binding.evcc/src/main/java/org/openhab/binding/evcc/internal/EvccHandler.java
```diff
@@ -0,0 +1,443 @@
...
+        if (config != null) {
+            String url = config.url;
+            if (url == null || "".equals(url)) {
+                updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.OFFLINE.CONFIGURATION_ERROR,
+                        "@text/offline.configuration-error.no-host");
+                return;
+            }
+            this.evccAPI = new EvccAPI(url);
+            logger.debug("Setting up refresh job ...");
+            statePollingJob = scheduler.scheduleWithFixedDelay(this::refresh, 0, 15, TimeUnit.SECONDS);
```
jlaur, line 179: Should the value 15 here be replaced by **refreshInterval** from the configuration?

bundles/org.openhab.binding.evcc/src/main/java/org/openhab/binding/evcc/internal/EvccHandler.java
```diff
@@ -0,0 +1,443 @@
...
+                default:
+                    return;
+            }
+            refresh();
+        }
+    }
+
+    @Override
+    public void initialize() {
+        this.config = getConfigAs(EvccConfiguration.class);
```
jlaur, line 168: `getConfigAs()` will not return null. And also it seems that the values obtained from the configuration is only used locally within `initialize()`. So this can be simplified by just validating URL and using the values directly without storing in any instance variables.

bundles/org.openhab.binding.evcc/src/main/java/org/openhab/binding/evcc/internal/EvccHandler.java
```diff
@@ -0,0 +1,443 @@
...
+        EvccConfiguration config = this.config;
+        if (config != null) {
+            String url = config.url;
+            if (url == null || "".equals(url)) {
```
jlaur, line 172:
```suggestion
            if (url == null || url.isBlank()) {
```

bundles/org.openhab.binding.evcc/src/main/java/org/openhab/binding/evcc/internal/EvccHandler.java
```diff
@@ -0,0 +1,443 @@
...
+     * Second, creates all available channels.
+     * Third, updates all channels.
+     */
+    private void refresh() {
+        logger.debug("Running refresh job ...");
+        EvccConfiguration config = this.config;
+        if (config == null)
+            return;
+        if (config.url == null)
+            return;
```
jlaur, line 196: This can be omitted if you change the logging a few lines below to omit the URL, which seems unneeded.

bundles/org.openhab.binding.evcc/src/main/java/org/openhab/binding/evcc/internal/EvccHandler.java
```diff
@@ -0,0 +1,443 @@
...
+        EvccAPI evccAPI = this.evccAPI;
+        if (evccAPI == null)
+            return;
```
jlaur, line 199:
```suggestion
        if (evccAPI == null) {
            return;
        }
```

bundles/org.openhab.binding.evcc/src/main/java/org/openhab/binding/evcc/internal/EvccHandler.java
```diff
@@ -0,0 +1,443 @@
...
+            }
+        }
+    }
+
+    @Override
+    public void dispose() {
+        ScheduledFuture<?> statePollingJob = this.statePollingJob;
+        if (statePollingJob != null) {
+            statePollingJob.cancel(true);
+            statePollingJob = null;
```
jlaur, line 228:
```suggestion
            this.statePollingJob = null;
```

bundles/org.openhab.binding.evcc/src/main/java/org/openhab/binding/evcc/internal/EvccHandler.java
```diff
@@ -0,0 +1,443 @@
...
+        }
+    }
+
+    // Utility functions
+    private void createChannelsGeneral() {
+        if (batteryConfigured == true) {
```
jlaur, line 234:
```suggestion
        if (batteryConfigured) {
```

bundles/org.openhab.binding.evcc/src/main/java/org/openhab/binding/evcc/internal/EvccHandler.java
```diff
@@ -0,0 +1,443 @@
...
+            createChannel(CHANNEL_BATTERY_POWER, "general", CHANNEL_TYPE_UID_BATTERY_POWER, "Number:Power");
```
jlaur, line 235: Consider introducing a constant for this channel group, "general".

bundles/org.openhab.binding.evcc/src/main/java/org/openhab/binding/evcc/internal/EvccHandler.java
```diff
@@ -0,0 +1,443 @@
...
+        createChannel(CHANNEL_LOADPOINT_TARGET_TIME, loadpointName, CHANNEL_TYPE_UID_LOADPOINT_TARGET_TIME, "DateTime");
+        createChannel(CHANNEL_LOADPOINT_TARGET_TIME_ENABLED, loadpointName,
+                CHANNEL_TYPE_UID_LOADPOINT_TARGET_TIME_ENABLED, "Switch");
+        createChannel(CHANNEL_LOADPOINT_TITLE, loadpointName, CHANNEL_TYPE_UID_LOADPOINT_TITLE, "String");
+        createChannel(CHANNEL_LOADPOINT_VEHICLE_CAPACITY, loadpointName, CHANNEL_TYPE_UID_LOADPOINT_VEHICLE_CAPACITY,
+                "Number:Energy");
+        createChannel(CHANNEL_LOADPOINT_VEHICLE_ODOMETER, loadpointName, CHANNEL_TYPE_UID_LOADPOINT_VEHICLE_ODOMETER,
+                "Number:Length");
+        createChannel(CHANNEL_LOADPOINT_VEHICLE_PRESENT, loadpointName, CHANNEL_TYPE_UID_LOADPOINT_VEHICLE_PRESENT,
+                "Switch");
```
jlaur, line 290:
```suggestion
                CoreItemFactory.SWITCH);
```

bundles/org.openhab.binding.evcc/src/main/java/org/openhab/binding/evcc/internal/EvccHandler.java
```diff
@@ -0,0 +1,443 @@
...
+        createChannel(CHANNEL_HOME_POWER, "general", CHANNEL_TYPE_UID_HOME_POWER, "Number:Power");
+        if (pvConfigured == true) {
+            createChannel(CHANNEL_PV_POWER, "general", CHANNEL_TYPE_UID_PV_POWER, "Number:Power");
+        }
+    }
+
+    private void createChannelsLoadpoint(int loadpointId) {
+        String loadpointName = "loadpoint" + loadpointId;
+        createChannel(CHANNEL_LOADPOINT_ACTIVE_PHASES, loadpointName, CHANNEL_TYPE_UID_LOADPOINT_ACTIVE_PHASES,
+                "Number");
```
jlaur, line 252:
```suggestion
                CoreItemFactory.NUMBER);
```

bundles/org.openhab.binding.evcc/src/main/java/org/openhab/binding/evcc/internal/EvccHandler.java
```diff
@@ -0,0 +1,443 @@
...
+                "Number:ElectricCurrent");
+        createChannel(CHANNEL_LOADPOINT_MIN_SOC, loadpointName, CHANNEL_TYPE_UID_LOADPOINT_MIN_SOC,
+                "Number:Dimensionless");
+        createChannel(CHANNEL_LOADPOINT_MODE, loadpointName, CHANNEL_TYPE_UID_LOADPOINT_MODE, "String");
+        createChannel(CHANNEL_LOADPOINT_PHASES, loadpointName, CHANNEL_TYPE_UID_LOADPOINT_PHASES, "Number");
+        createChannel(CHANNEL_LOADPOINT_TARGET_SOC, loadpointName, CHANNEL_TYPE_UID_LOADPOINT_TARGET_SOC,
+                "Number:Dimensionless");
+        createChannel(CHANNEL_LOADPOINT_TARGET_TIME, loadpointName, CHANNEL_TYPE_UID_LOADPOINT_TARGET_TIME, "DateTime");
+        createChannel(CHANNEL_LOADPOINT_TARGET_TIME_ENABLED, loadpointName,
+                CHANNEL_TYPE_UID_LOADPOINT_TARGET_TIME_ENABLED, "Switch");
```
jlaur, line 283:
```suggestion
                CHANNEL_TYPE_UID_LOADPOINT_TARGET_TIME_ENABLED, CoreItemFactory.SWITCH);
```

bundles/org.openhab.binding.evcc/src/main/java/org/openhab/binding/evcc/internal/EvccHandler.java
```diff
@@ -0,0 +1,443 @@
...
+                CHANNEL_TYPE_UID_LOADPOINT_CONNECTED_DURATION, "Number:Time");
+        createChannel(CHANNEL_LOADPOINT_ENABLED, loadpointName, CHANNEL_TYPE_UID_LOADPOINT_ENABLED, "Switch");
+        createChannel(CHANNEL_LOADPOINT_HAS_VEHICLE, loadpointName, CHANNEL_TYPE_UID_LOADPOINT_HAS_VEHICLE, "Switch");
+        createChannel(CHANNEL_LOADPOINT_MAX_CURRENT, loadpointName, CHANNEL_TYPE_UID_LOADPOINT_MAX_CURRENT,
+                "Number:ElectricCurrent");
+        createChannel(CHANNEL_LOADPOINT_MIN_CURRENT, loadpointName, CHANNEL_TYPE_UID_LOADPOINT_MIN_CURRENT,
+                "Number:ElectricCurrent");
+        createChannel(CHANNEL_LOADPOINT_MIN_SOC, loadpointName, CHANNEL_TYPE_UID_LOADPOINT_MIN_SOC,
+                "Number:Dimensionless");
+        createChannel(CHANNEL_LOADPOINT_MODE, loadpointName, CHANNEL_TYPE_UID_LOADPOINT_MODE, "String");
```
jlaur, line 277:
```suggestion
        createChannel(CHANNEL_LOADPOINT_MODE, loadpointName, CHANNEL_TYPE_UID_LOADPOINT_MODE, CoreItemFactory.STRING);
```

bundles/org.openhab.binding.evcc/src/main/java/org/openhab/binding/evcc/internal/EvccHandler.java
```diff
@@ -0,0 +1,443 @@
...
+        createChannel(CHANNEL_LOADPOINT_PHASES, loadpointName, CHANNEL_TYPE_UID_LOADPOINT_PHASES, "Number");
+        createChannel(CHANNEL_LOADPOINT_TARGET_SOC, loadpointName, CHANNEL_TYPE_UID_LOADPOINT_TARGET_SOC,
+                "Number:Dimensionless");
+        createChannel(CHANNEL_LOADPOINT_TARGET_TIME, loadpointName, CHANNEL_TYPE_UID_LOADPOINT_TARGET_TIME, "DateTime");
```
jlaur, line 281:
```suggestion
        createChannel(CHANNEL_LOADPOINT_TARGET_TIME, loadpointName, CHANNEL_TYPE_UID_LOADPOINT_TARGET_TIME, CoreItemFactory.DATETIME);
```

bundles/org.openhab.binding.evcc/src/main/java/org/openhab/binding/evcc/internal/EvccHandler.java
```diff
@@ -0,0 +1,443 @@
...
+        createChannel(CHANNEL_LOADPOINT_ENABLED, loadpointName, CHANNEL_TYPE_UID_LOADPOINT_ENABLED, "Switch");
+        createChannel(CHANNEL_LOADPOINT_HAS_VEHICLE, loadpointName, CHANNEL_TYPE_UID_LOADPOINT_HAS_VEHICLE, "Switch");
+        createChannel(CHANNEL_LOADPOINT_MAX_CURRENT, loadpointName, CHANNEL_TYPE_UID_LOADPOINT_MAX_CURRENT,
+                "Number:ElectricCurrent");
+        createChannel(CHANNEL_LOADPOINT_MIN_CURRENT, loadpointName, CHANNEL_TYPE_UID_LOADPOINT_MIN_CURRENT,
+                "Number:ElectricCurrent");
+        createChannel(CHANNEL_LOADPOINT_MIN_SOC, loadpointName, CHANNEL_TYPE_UID_LOADPOINT_MIN_SOC,
+                "Number:Dimensionless");
+        createChannel(CHANNEL_LOADPOINT_MODE, loadpointName, CHANNEL_TYPE_UID_LOADPOINT_MODE, "String");
+        createChannel(CHANNEL_LOADPOINT_PHASES, loadpointName, CHANNEL_TYPE_UID_LOADPOINT_PHASES, "Number");
```
jlaur, line 278:
```suggestion
        createChannel(CHANNEL_LOADPOINT_PHASES, loadpointName, CHANNEL_TYPE_UID_LOADPOINT_PHASES, CoreItemFactory.NUMBER);
```

bundles/org.openhab.binding.evcc/src/main/java/org/openhab/binding/evcc/internal/EvccHandler.java
```diff
@@ -0,0 +1,443 @@
...
+                "Number:Time");
+        createChannel(CHANNEL_LOADPOINT_CHARGE_POWER, loadpointName, CHANNEL_TYPE_UID_LOADPOINT_CHARGE_POWER,
+                "Number:Power");
+        createChannel(CHANNEL_LOADPOINT_CHARGE_REMAINING_DURATION, loadpointName,
+                CHANNEL_TYPE_UID_LOADPOINT_CHARGE_REMAINING_DURATION, "Number:Time");
+        createChannel(CHANNEL_LOADPOINT_CHARGE_REMAINING_ENERGY, loadpointName,
+                CHANNEL_TYPE_UID_LOADPOINT_CHARGE_REMAINING_ENERGY, "Number:Energy");
+        createChannel(CHANNEL_LOADPOINT_CHARGED_ENERGY, loadpointName, CHANNEL_TYPE_UID_LOADPOINT_CHARGED_ENERGY,
+                "Number:Energy");
+        createChannel(CHANNEL_LOADPOINT_CHARGING, loadpointName, CHANNEL_TYPE_UID_LOADPOINT_CHARGING, "Switch");
```
jlaur, line 265:
```suggestion
        createChannel(CHANNEL_LOADPOINT_CHARGING, loadpointName, CHANNEL_TYPE_UID_LOADPOINT_CHARGING, CoreItemFactory.SWITCH);
```

bundles/org.openhab.binding.evcc/src/main/java/org/openhab/binding/evcc/internal/EvccHandler.java
```diff
@@ -0,0 +1,443 @@
...
+        createChannel(CHANNEL_LOADPOINT_CONNECTED, loadpointName, CHANNEL_TYPE_UID_LOADPOINT_CONNECTED, "Switch");
+        createChannel(CHANNEL_LOADPOINT_CONNECTED_DURATION, loadpointName,
+                CHANNEL_TYPE_UID_LOADPOINT_CONNECTED_DURATION, "Number:Time");
+        createChannel(CHANNEL_LOADPOINT_ENABLED, loadpointName, CHANNEL_TYPE_UID_LOADPOINT_ENABLED, "Switch");
```
jlaur, line 269:
```suggestion
        createChannel(CHANNEL_LOADPOINT_ENABLED, loadpointName, CHANNEL_TYPE_UID_LOADPOINT_ENABLED, CoreItemFactory.SWITCH);
```

bundles/org.openhab.binding.evcc/src/main/java/org/openhab/binding/evcc/internal/EvccHandler.java
```diff
@@ -0,0 +1,443 @@
...
+        createChannel(CHANNEL_LOADPOINT_HAS_VEHICLE, loadpointName, CHANNEL_TYPE_UID_LOADPOINT_HAS_VEHICLE, "Switch");
```
jlaur, line 270:
```suggestion
        createChannel(CHANNEL_LOADPOINT_HAS_VEHICLE, loadpointName, CHANNEL_TYPE_UID_LOADPOINT_HAS_VEHICLE, CoreItemFactory.SWITCH);
```

bundles/org.openhab.binding.evcc/src/main/java/org/openhab/binding/evcc/internal/EvccHandler.java
```diff
@@ -0,0 +1,443 @@
...
+            createChannel(CHANNEL_BATTERY_POWER, "general", CHANNEL_TYPE_UID_BATTERY_POWER, "Number:Power");
+            createChannel(CHANNEL_BATTERY_SOC, "general", CHANNEL_TYPE_UID_BATTERY_SOC, "Number:Dimensionless");
+            createChannel(CHANNEL_BATTERY_PRIORITY_SOC, "general", CHANNEL_TYPE_UID_BATTERY_PRIORITY_SOC,
+                    "Number:Dimensionless");
+        }
+        if (gridConfigured == true) {
+            createChannel(CHANNEL_GRID_POWER, "general", CHANNEL_TYPE_UID_GRID_POWER, "Number:Power");
+        }
+        createChannel(CHANNEL_HOME_POWER, "general", CHANNEL_TYPE_UID_HOME_POWER, "Number:Power");
+        if (pvConfigured == true) {
```
jlaur, line 244:
```suggestion
        if (pvConfigured) {
```

bundles/org.openhab.binding.evcc/src/main/java/org/openhab/binding/evcc/internal/EvccHandler.java
```diff
@@ -0,0 +1,443 @@
...
+
+    // Utility functions
+    private void createChannelsGeneral() {
+        if (batteryConfigured == true) {
+            createChannel(CHANNEL_BATTERY_POWER, "general", CHANNEL_TYPE_UID_BATTERY_POWER, "Number:Power");
+            createChannel(CHANNEL_BATTERY_SOC, "general", CHANNEL_TYPE_UID_BATTERY_SOC, "Number:Dimensionless");
+            createChannel(CHANNEL_BATTERY_PRIORITY_SOC, "general", CHANNEL_TYPE_UID_BATTERY_PRIORITY_SOC,
+                    "Number:Dimensionless");
+        }
+        if (gridConfigured == true) {
```
jlaur, line 240:
```suggestion
        if (gridConfigured) {
```

bundles/org.openhab.binding.evcc/src/main/java/org/openhab/binding/evcc/internal/EvccHandler.java
```diff
@@ -0,0 +1,443 @@
...
+        createChannel(CHANNEL_LOADPOINT_VEHICLE_ODOMETER, loadpointName, CHANNEL_TYPE_UID_LOADPOINT_VEHICLE_ODOMETER,
+                "Number:Length");
+        createChannel(CHANNEL_LOADPOINT_VEHICLE_PRESENT, loadpointName, CHANNEL_TYPE_UID_LOADPOINT_VEHICLE_PRESENT,
+                "Switch");
+        createChannel(CHANNEL_LOADPOINT_VEHICLE_RANGE, loadpointName, CHANNEL_TYPE_UID_LOADPOINT_VEHICLE_RANGE,
+                "Number:Length");
+        createChannel(CHANNEL_LOADPOINT_VEHICLE_SOC, loadpointName, CHANNEL_TYPE_UID_LOADPOINT_VEHICLE_SOC,
+                "Number:Dimensionless");
+        createChannel(CHANNEL_LOADPOINT_VEHICLE_TITLE, loadpointName, CHANNEL_TYPE_UID_LOADPOINT_VEHICLE_TITLE,
+                "String");
```
jlaur, line 296:
```suggestion
               CoreItemFactory.STRING);
```

bundles/org.openhab.binding.evcc/src/main/java/org/openhab/binding/evcc/internal/EvccHandler.java
```diff
@@ -0,0 +1,443 @@
...
+    }
+
+    // Units and description for vars: https://docs.evcc.io/docs/reference/configuration/messaging/#msg
+    private void updateChannelsGeneral() {
+        Result result = this.result;
+        if (result == null)
+            return;
```
jlaur, line 303:
```suggestion
        if (result == null) {
            return;
        }
```

bundles/org.openhab.binding.evcc/src/main/java/org/openhab/binding/evcc/internal/EvccHandler.java
```diff
@@ -0,0 +1,443 @@
...
+        ChannelUID channel;
+        boolean batteryConfigured = this.batteryConfigured;
+        if (batteryConfigured == true) {
```
jlaur, line 306:
```suggestion
        if (batteryConfigured) {
```

bundles/org.openhab.binding.evcc/src/main/java/org/openhab/binding/evcc/internal/EvccHandler.java
```diff
@@ -0,0 +1,443 @@
...
+            updateState(channel, new QuantityType<>(batteryPower, Units.WATT));
+            batterySoC = result.getBatterySoC();
+            channel = new ChannelUID(getThing().getUID(), "general", CHANNEL_BATTERY_SOC);
+            updateState(channel, new QuantityType<>(batterySoC, Units.PERCENT));
+            batteryPrioritySoC = result.getBatterySoC();
+            channel = new ChannelUID(getThing().getUID(), "general", CHANNEL_BATTERY_PRIORITY_SOC);
+            updateState(channel, new QuantityType<>(batteryPrioritySoC, Units.PERCENT));
+        }
+        boolean gridConfigured = this.gridConfigured;
+        if (gridConfigured == true) {
```
jlaur, line 318:
```suggestion
        if (gridConfigured) {
```

bundles/org.openhab.binding.evcc/src/main/java/org/openhab/binding/evcc/internal/EvccHandler.java
```diff
@@ -0,0 +1,443 @@
...
+            gridPower = result.getGridPower();
+            channel = new ChannelUID(getThing().getUID(), "general", CHANNEL_GRID_POWER);
+            updateState(channel, new QuantityType<>(gridPower, Units.WATT));
+        }
+        homePower = result.getHomePower();
+        channel = new ChannelUID(getThing().getUID(), "general", CHANNEL_HOME_POWER);
+        updateState(channel, new QuantityType<>(homePower, Units.WATT));
+        boolean pvConfigured = this.pvConfigured;
+        if (pvConfigured == true) {
```
jlaur, line 327:
```suggestion
        if (pvConfigured) {
```

bundles/org.openhab.binding.evcc/src/main/java/org/openhab/binding/evcc/internal/EvccHandler.java
```diff
@@ -0,0 +1,443 @@
...
+            pvPower = result.getPvPower();
+            channel = new ChannelUID(getThing().getUID(), "general", CHANNEL_PV_POWER);
+            updateState(channel, new QuantityType<>(pvPower, Units.WATT));
+        }
+    }
+
+    private void updateChannelsLoadpoint(int loadpointId) {
+        Result result = this.result;
+        if (result == null)
+            return;
```
jlaur, line 337:
```suggestion
        if (result == null) {
            return;
        }
```

bundles/org.openhab.binding.evcc/src/main/resources/OH-INF/thing/thing-types.xml
```diff
@@ -0,0 +1,328 @@
...
+			<channel-group id="loadpoint8" typeId="loadpoint"/>
+			<channel-group id="loadpoint9" typeId="loadpoint"/>
+		</channel-groups>
+
+		<config-description>
+			<parameter name="url" type="text" required="true">
+				<context>network-address</context>
+				<label>URL</label>
+				<description>URL of evcc web UI, e.g. https://demo.evcc.io</description>
+				<default>https://demo.evcc.io</default>
```
jlaur, line 31: Does it make sense to default a demo URL? Parameter is required, so by leaving a default, user will not be forced to input anything meaningful.

bundles/org.openhab.binding.evcc/src/main/resources/OH-INF/thing/thing-types.xml
```diff
@@ -0,0 +1,328 @@
...
+		<item-type>Number:Power</item-type>
+		<label>Home Power</label>
+		<description>Current power taken by home</description>
+		<category>Energy</category>
+		<state pattern="%.1f %unit%" readOnly="true"/>
+	</channel-type>
+	<channel-type id="pvPower">
+		<item-type>Number:Power</item-type>
+		<label>PV Power</label>
+		<description> Current power from photovoltaik</description>
```
jlaur, line 88:
```suggestion
		<description>Current power from photovoltaik</description>
```

bundles/org.openhab.binding.evcc/src/main/resources/OH-INF/thing/thing-types.xml
```diff
@@ -0,0 +1,328 @@
...
+		<autoUpdatePolicy>veto</autoUpdatePolicy>
+	</channel-type>
+	<channel-type id="mode">
+		<item-type>String</item-type>
+		<label>Charging mode</label>
+		<description>Charging mode: "off", "now", "minpv", "pv"</description>
+		<category>String</category>
+		<state readOnly="false">
+			<options>
+				<option value="off">Stop</option>
```
jlaur, line 210: Stop or Off?

bundles/org.openhab.binding.evcc/README.md
```diff
@@ -0,0 +1,171 @@
...
+No auto discovery supported.
+
+## Thing Configuration
+
+### `device` Thing Configuration
+
+| Parameter       | Type   | Description                                              | Advanced | Required |
+|-----------------|--------|----------------------------------------------------------|----------|----------|
+| url             | String | URL of evcc web UI, e.g. *https://demo.evcc.io*          | false    | true     |
+| refreshInterval | Number | Interval the status is polled in seconds (minimum is 15) | true     | false    |
```
jlaur, line 27: true/false could be replaced by Yes/No or something else which would be more user-oriented. :)

jlaur (changes requested):
I went through the latest version of the code one more time, and added some additional feedback. Please let me know if anything is unclear, as in some cases I simply added suggestions without explicit comments.

bundles/org.openhab.binding.evcc/src/main/java/org/openhab/binding/evcc/internal/EvccAPI.java
```diff
@@ -0,0 +1,132 @@
...
+     */
+    private @Nullable String httpRequest(@Nullable String description, String url, String method) {
+        String response = "";
+        try {
+            response = HttpUtil.executeUrl(method, url, LONG_CONNECTION_TIMEOUT_MILLISEC);
+            logger.trace("{} - {}, {} - {}", description, url, method, response);
+            return response;
+        } catch (IOException e) {
+            logger.debug("IO Exception - {} - {}, {} - {}", description, url, method, e.toString());
+            return null;
```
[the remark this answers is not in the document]
florian-h05: Which exception should I throw then?

bundles/org.openhab.binding.evcc/src/main/java/org/openhab/binding/evcc/internal/EvccHandler.java
```diff
@@ -0,0 +1,443 @@
...
+public class EvccHandler extends BaseThingHandler {
+    private final Logger logger = LoggerFactory.getLogger(EvccHandler.class);
+    private @Nullable EvccAPI evccAPI;
+    private @Nullable ScheduledFuture<?> statePollingJob;
+
+    private @Nullable EvccConfiguration config;
+
+    private @Nullable Result result;
+
+    private int numberOfLoadpoints = 0;
```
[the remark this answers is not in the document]
florian-h05: I had them as local variables, but have changed them to global when I tried to remove the compiler warnings.

bundles/org.openhab.binding.evcc/src/main/java/org/openhab/binding/evcc/internal/EvccHandler.java
florian-h05: Oops, yes the refresh interval from the config should be there.

bundles/org.openhab.binding.evcc/src/main/resources/OH-INF/thing/thing-types.xml
florian-h05: In the evcc UI it's called `Stop`, but off might be more appropriate.
(It switches the charging off).

What do you think?

bundles/org.openhab.binding.evcc/src/main/java/org/openhab/binding/evcc/internal/EvccAPI.java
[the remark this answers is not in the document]
jlaur: I was thinking some custom exception where you could include URL and method, but actually `IOException` could be sufficient for this case, so you could simply add "throws IOException" to the method signature and remove the local handling here. Then, in the handler, `handleCommand()`, you could catch it and handle it by logging it. You would need to adapt `getResult()` also for unified error handling, so either introduce custom exception which could be thrown from here, or handle both the Json exception and IOException in the handler.

bundles/org.openhab.binding.evcc/src/main/resources/OH-INF/thing/thing-types.xml
jlaur: IMHO "Off" would be best as it best expresses state and action. "Stop" can only denote an action, whereas "Off" can mean "charging mode is **off**" (state) or "turn **off** charging mode" (action).

bundles/org.openhab.binding.evcc/src/main/java/org/openhab/binding/evcc/internal/EvccAPI.java
```diff
@@ -0,0 +1,132 @@
...
+
+    /**
+     * Make a HTTP request.
+     * 
+     * @param description request description for logger
+     * @param url full request URL
+     * @param method reguest method, e.g. GET, POST
+     * @return the response body or null if request failed
+     */
+    private @Nullable String httpRequest(@Nullable String description, String url, String method) {
```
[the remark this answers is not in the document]
florian-h05: I can also remove it, but do we loose anything if we keep it?

bundles/org.openhab.binding.evcc/src/main/java/org/openhab/binding/evcc/internal/EvccHandler.java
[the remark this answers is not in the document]
florian-h05: Some of those globals are required, e.g. targetSoC and targetTime, as the command handler needs them.

bundles/org.openhab.binding.evcc/src/main/java/org/openhab/binding/evcc/internal/EvccAPI.java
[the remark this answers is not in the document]
jlaur: I won't insist, but for me it reduces readability that first argument is the least important one (and even nullable), which is only used for debug logging - and in fact somewhat redundant.

bundles/org.openhab.binding.evcc/src/main/java/org/openhab/binding/evcc/internal/EvccAPI.java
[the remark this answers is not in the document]
jlaur: Hmm, after seeing last commits: I'm sorry if I got you started on this error handling refactoring without fully explaining my thoughts. So please just take the following as a description of my intentions here, and you can decide how much or little you want to do.

After defining a custom exception, for example:

```
@NonNullByDefault
public class EvccApiException extends Exception {

    private static final long serialVersionUID = 0L; // uniquely generated UID

    public EvccApiException(String message) {
        super(message);
    }

    public EvccApiException(String message, Throwable cause) {
        super(message, cause);
    }
}
```

this could then be thrown in case of failure:

```
private String httpRequest(String url, String method) throws EvccApiException {
    try {
        String response = HttpUtil.executeUrl(method, url, LONG_CONNECTION_TIMEOUT_MILLISEC);
        logger.trace("{}, {} - {}", url, method, response);
        return response;
    } catch (IOException e) {
        throw new EvccApiException("HTTP request failed for URL " + url, e);
    }
}
```

Then, in `getResult`:

```
public Result getResult() throws EvccApiException {
    final String response = httpRequest(this.host + EVCC_REST_API + "state", "GET");
    try {
        Status status = gson.fromJson(response, Status.class);
        if (status == null) {
            throw new EvccApiException("Status is null");
        }
        return status.getResult();
    } catch (JsonSyntaxException e) {
        throw new EvccApiException("Error parsing response: " + response, e);
    }
}
```

And the set methods:

```
public String setMode(int loadpoint, String mode) throws EvccApiException {
    return httpRequest(this.host + EVCC_REST_API + "loadpoints/" + loadpoint + "/mode/" + mode, "POST");
}
```

So now nothing is nullable and we only have one exception to deal with centrally - we'll either get a valid result or an exception thrown. So all we need to do in handler:

```
try {
    switch (channelIdWithoutGroup) {
        case CHANNEL_LOADPOINT_MODE:
            evccAPI.setMode(loadpoint, command.toString());
            break;
        case CHANNEL_LOADPOINT_MIN_SOC:
            evccAPI.setMinSoC(loadpoint, Integer.parseInt(command.toString().replaceAll(" %", "")));
            break;
} catch (EvccApiException e) {
    Throwable cause = e.getCause();
    if (cause == null) {
        logger.debug("Failed to update channel {}: {}", channelUID, e.getMessage());
    } else {
        logger.debug("Failed to update channel {}: {} -> {}", channelUID, e.getMessage(), cause.getMessage());
    }
}
```
(switch could even be extracted into separate private method for simplicity)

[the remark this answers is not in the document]
florian-h05: That looks good, I will have a look into this.

```java
private static final long serialVersionUID = 0L; // uniquely generated UID
```

Do I need and if yes, how would I generate that UID?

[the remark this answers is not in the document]
florian-h05: > (switch could even be extracted into separate private method for simplicity)

On one hand, yes, but on the other hand (at least for me) more methods make code more complicated to understand/read as you have to scroll more.

I will leave the switch inside of `handleCommand`

florian-h05:
@jlaur 
Please have a look at my two latest commits.

bundles/org.openhab.binding.evcc/src/main/java/org/openhab/binding/evcc/internal/EvccAPI.java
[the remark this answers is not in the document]
jlaur: >> private static final long serialVersionUID = 0L; // uniquely generated UID

> Do I need and if yes, how would I generate that UID?

Yes, you need to - your IDE can most likely generate it for you. At least I know IntelliJ and Eclipse can do it, so if you're using one of them, it would probably be easiest to google how to do it with your IDE and the specific version of it.

>> (switch could even be extracted into separate private method for simplicity)

> On one hand, yes, but on the other hand (at least for me) more methods make code more complicated to
> understand/read as you have to scroll more.

Well, I have to disagree here. :) By having smaller methods they will usually be simpler to read and understand as they would do fewer things, thus complexity would be smaller.

In the example here, having a private method for handling commands calling the set methods, let's call it `handleLoadpointCommand` (I don't know if it makes sense, so just for the example), it would be easy to see how errors are handled and that refresh() is called after calling the API set methods. Currently you will have to scroll down to understand this flow.

That said, I'm fine with your implementation, it was just a side note with an additional refactoring proposal. :)

bundles/org.openhab.binding.evcc/src/main/java/org/openhab/binding/evcc/internal/EvccHandler.java
```diff
@@ -0,0 +1,436 @@
...
+                    case CHANNEL_LOADPOINT_MAX_CURRENT:
+                        evccAPI.setMaxCurrent(loadpoint, Integer.parseInt(command.toString().replaceAll(" A", "")));
+                        break;
+                    default:
+                        return;
+                }
+            } catch (EvccApiException e) {
+                Throwable cause = e.getCause();
+                if (cause == null) {
+                    logger.debug("Failed to command channel {}: {}", channelUID, e.getMessage());
```
jlaur, line 143: Probably some rephrasing needed here, perhaps "Failed to handle command for channel"?

bundles/org.openhab.binding.evcc/src/main/java/org/openhab/binding/evcc/internal/api/EvccAPI.java
```diff
@@ -0,0 +1,125 @@
...
+     * @throws IOException if HTTP request failed
+     * @throws JsonSyntaxException if response JSON read failed
+     */
+    public @Nullable Result getResult() throws EvccApiException {
+        final String response = httpRequest(this.host + EVCC_REST_API + "state", "GET");
+        try {
+            Status status = gson.fromJson(response, Status.class);
+            if (status == null) {
+                return null;
+            }
```
jlaur, line 83: This is inconsistent as caller will need to handle errors in two different ways: Checking for null and catching exception. If you change this to something like:
```suggestion
            if (status == null) {
                throw new EvccApiException("Status is null");
            }
```

you can drop `@Nullable` from this method as well as all set methods below.

bundles/org.openhab.binding.evcc/src/main/java/org/openhab/binding/evcc/internal/EvccHandler.java
```diff
@@ -0,0 +1,436 @@
...
+    public EvccHandler(Thing thing) {
+        super(thing);
+    }
+
+    @Override
+    public void handleCommand(ChannelUID channelUID, Command command) {
+        if (command.equals(RefreshType.REFRESH)) {
+            refresh();
+        } else {
+            logger.debug("Handling command {} for channel {}", command, channelUID);
```
jlaur, line 83: Which command type do you expect here? You might want to check it as it channel could be mapped to any item type.

bundles/org.openhab.binding.evcc/src/main/java/org/openhab/binding/evcc/internal/EvccHandler.java
```diff
@@ -0,0 +1,436 @@
...
+            if (evccAPI == null) {
+                return;
+            }
+            try {
+                switch (channelIdWithoutGroup) {
+                    case CHANNEL_LOADPOINT_MODE:
+                        evccAPI.setMode(loadpoint, command.toString());
+                        break;
+                    case CHANNEL_LOADPOINT_MIN_SOC:
+                        evccAPI.setMinSoC(loadpoint, Integer.parseInt(command.toString().replaceAll(" %", "")));
```
jlaur, line 100: Will this work?
```suggestion
                        if (command instanceof PercentType) {
                            evccAPI.setMinSoC(loadpoint, ((PercentType) command).intValue());
                        }
```

bundles/org.openhab.binding.evcc/src/main/java/org/openhab/binding/evcc/internal/api/EvccAPI.java
```diff
@@ -0,0 +1,125 @@
...
+        this.host = host;
+    }
+
+    /**
+     * Make a HTTP request.
+     * 
+     * @param url full request URL
+     * @param method reguest method, e.g. GET, POST
+     * @return the response body
+     * @throws IOException if HTTP request failed
```
jlaur, line 54:
```suggestion
     * @throws EvccApiException if HTTP request failed
```

or remove Javadoc as this is private method.

bundles/org.openhab.binding.evcc/src/main/java/org/openhab/binding/evcc/internal/api/EvccAPI.java
```diff
@@ -0,0 +1,125 @@
...
+    // End utility functions
+
+    // API calls to evcc
+    /**
+     * Get the status from evcc.
+     * 
+     * @param host hostname of IP address of the evcc instance
+     * @return {@link org.openhab.binding.evcc.internal.api.dto.Result} object or null
+     * @throws IOException if HTTP request failed
+     * @throws JsonSyntaxException if response JSON read failed
```
jlaur, line 75:
```suggestion
     * @throws EvccApiException if HTTP request failed
```

bundles/org.openhab.binding.evcc/src/main/java/org/openhab/binding/evcc/internal/EvccHandler.java
florian-h05: Yes, in fact that works, but your approach looks much better.

florian-h05: How would that work for (current is QuantityType):

```java
 case CHANNEL_LOADPOINT_MIN_CURRENT:
   evccAPI.setMinCurrent(loadpoint, Integer.parseInt(command.toString().replaceAll(" A", "")));
   break;
```

bundles/org.openhab.binding.evcc/src/main/java/org/openhab/binding/evcc/internal/EvccAPI.java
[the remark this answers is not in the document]
florian-h05: > Yes, you need to - your IDE can most likely generate it for you. At least I know IntelliJ and Eclipse can do it, so if you're using one of them, it would probably be easiest to google how to do it with your IDE and the specific version of it.

I found no way to do that in VS Code.
Maybe you could generate that UID in your IDE and post it here?

bundles/org.openhab.binding.evcc/src/main/java/org/openhab/binding/evcc/internal/EvccHandler.java
florian-h05: Sorry, I am not sure what you mean.

Should I check the command type in each case of channel name??

bundles/org.openhab.binding.evcc/src/main/java/org/openhab/binding/evcc/internal/EvccHandler.java
florian-h05: I will try do handle the command that way for all channels.

jlaur: > Yes, in fact that works, but your approach looks much better.

Sorry, I meant to ask if my proposal would work. :)

bundles/org.openhab.binding.evcc/src/main/java/org/openhab/binding/evcc/internal/api/EvccAPI.java
florian-h05: See 7ade98f

bundles/org.openhab.binding.evcc/src/main/java/org/openhab/binding/evcc/internal/EvccAPI.java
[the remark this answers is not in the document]
florian-h05: I installed Ecplise, got the serialVersionUID generated.

See 7ade98f

bundles/org.openhab.binding.evcc/src/main/java/org/openhab/binding/evcc/internal/EvccHandler.java
florian-h05: Your proposal works slightly modified, thanks.
I have adjusted all command channels accordingly.

See 978656d

bundles/org.openhab.binding.evcc/src/main/java/org/openhab/binding/evcc/internal/EvccHandler.java
```diff
@@ -0,0 +1,449 @@
...
+                            evccAPI.setMaxCurrent(loadpoint, ((QuantityType<?>) command).intValue());
+                        }
+                        break;
+                    default:
+                        return;
+                }
+            } catch (EvccApiException e) {
+                Throwable cause = e.getCause();
+                if (cause == null) {
+                    logger.debug("Failed to handle command {} for channel {}: {}", command.toString(), channelUID,
```
jlaur, line 155: `toString()` is redundant:
```suggestion
                    logger.debug("Failed to handle command {} for channel {}: {}", command, channelUID,
```

bundles/org.openhab.binding.evcc/src/main/java/org/openhab/binding/evcc/internal/EvccHandler.java
```diff
@@ -0,0 +1,449 @@
...
+                            e.getMessage());
+                } else {
+                    logger.debug("Failed to handle command {} for channel {}: {} -> {}", command.toString(), channelUID,
```
jlaur, line 158:
```suggestion
                    logger.debug("Failed to handle command {} for channel {}: {} -> {}", command, channelUID,
```

bundles/org.openhab.binding.evcc/src/main/java/org/openhab/binding/evcc/internal/EvccHandler.java
jlaur: Very good.

bundles/org.openhab.binding.evcc/src/main/java/org/openhab/binding/evcc/internal/EvccHandler.java
```diff
@@ -0,0 +1,449 @@
...
+                                    logger.debug("Failed to set target charge", e);
+                                }
+                            }
+                        }
+                        break;
+                    case CHANNEL_LOADPOINT_TARGET_TIME_ENABLED:
+                        if (command == OnOffType.ON) {
+                            evccAPI.setTargetCharge(loadpoint, targetSoC, targetTimeZDT);
+                            targetTimeEnabled = true;
+                        } else {
```
jlaur, line 129:
```suggestion
                        } else if (command == OnOffType.OFF) {
```

florian-h05:
@jlaur 
If you do not have any more comments, this is ready to merge now.
I have tested it a few minutes ago and it works without any problems.

bundles/org.openhab.binding.evcc/src/main/java/org/openhab/binding/evcc/internal/api/dto/Loadpoint.java
```diff
@@ -0,0 +1,316 @@
...
+
+    /**
+     * @return the charged energy
+     */
+    public double getChargedEnergy() {
+        return chargedEnergy;
+    }
+
+    /**
+     * @return wtether loadpoint is charging a vehicle
```
jlaur, line 164:
```suggestion
     * @return whether loadpoint is charging a vehicle
```

bundles/org.openhab.binding.evcc/src/main/java/org/openhab/binding/evcc/internal/api/dto/Loadpoint.java
```diff
@@ -0,0 +1,316 @@
...
+     * @return the mode: off, now, minpv, pv
+     */
+    public String getMode() {
+        return mode;
+    }
+
+    /**
+     * @return the phases
+     */
+    public int getPhases() {
```
jlaur, line 236: Is it number of phases being returned?

jlaur: Okay - consider if this is worth mentioning in the method Javadoc and/or method name.

jlaur (approved):
Thanks for your contribution!

florian-h05:
Thanks for your review.
As I am relatively new to openHAB Add-On development (and Java development in general), I learned really much from your review.

jlaur:
@florian-h05 - I'm glad I could help. :-) Now, you could add your binding's logo to the openHAB website. See https://www.openhab.org/docs/developer/bindings/#add-your-binding-s-logo-to-the-openhab-website

florian-h05:
PR for icon is https://github.com/openhab/openhab-docs/pull/1815

jlaur:
@florian-h05 - since you are now code owner for this binding, you should now be added to the [contributors](https://github.com/orgs/openhab/teams/contributors) group, so that you will be automatically added as a reviewer there new pull requests will be opened for the binding. Can you check if you can request this access?

kaikreuzer:
I have added @florian-h05 to the contributors group!

florian-h05:
@jlaur @kaikreuzer 
I have accepted the invitation and now should be a member of the contributor group.

PR #12632, opened 2022-04-20
[upnpcontrol] Fix null pointer exception for invalid protocolInfo

mherwege:

According to the UPnP standard, res should always contain protocolInfo. Some devices do not respect this standard in their AVTransport implementation. To avoid null pointer exceptions, take care of this special case.


lolodomo (approved):
LGTM

PR #12649, opened 2022-04-25
[mybmw] Add default translations

wborn:
This makes it possible to translate the strings added in #12006 with Crowdin.


jlaur (approved):
Thanks!

weymann (commented):
LGTM

PR #12650, opened 2022-04-25
[jrubyscripting] Fix UI configuration property id

jimtng:
There was a bug in the component definition, causing configurations via the UI not saved to the correct Pid, as reported here:
https://community.openhab.org/t/jruby-openhab-rules-system/110598/191


wborn (approved):
Thanks!

PR #12652, opened 2022-04-25
[tellstick] Update README.md

ulfwin:
I have tried this binding with Tellstick Net v1, and it works (through Telldus Live)


jlaur (approved):
Thanks.

PR #12653, opened 2022-04-25
[elroconnects] Hub discovery, device management console commands, signal strength channels

mherwege:

This PR extends the functionality of the already reviewed binding.

Following functionality has been added:

- Alarm channels on the bridge, without a need to have the individual device alarm channels configured
- Signal strength channels for the devices wireless connection to the hub
- Rename devices from openHAB will rename them in the ELRO connects system
- Console command to list all devices on a hub (allows finding the device Id for manual configuration)
- Console commands to add, remove, rename or replace devices (no need for the ELRO connects app)
- Extra bridge type that makes connection to ELRO connects cloud service with automatic discovery of hubs
- Support for older versions of the firmware, version is set from cloud, but can be set manually

A beta version of this binding is available in the [marketplace](https://community.openhab.org/t/elro-connects-binding-beta/131152).


bundles/org.openhab.binding.elroconnects/src/main/java/org/openhab/binding/elroconnects/internal/discovery/ElroConnectsBridgeDiscoveryService.java
```diff
@@ -0,0 +1,149 @@
...
+                    // Assume new firmware if we cannot decode firmwareVersion
+                    logger.debug("Cannot get firmware version from {}, assume new firmware", firmwareVersion);
+                }
+                final Map<String, Object> properties = new HashMap<>();
+                properties.put(CONFIG_CONNECTOR_ID, connectorId);
+                properties.put(CONFIG_LEGACY_FIRMWARE, legacy);
+                properties.put("binVersion", c.getValue().binVersion);
+                properties.put("binType", c.getValue().binType);
+                properties.put("sdkVer", c.getValue().sdkVer);
+                properties.put("model", c.getValue().model);
```
jlaur, line 95: Here you can use standard property `Thing.PROPERTY_MODEL_ID`.

bundles/org.openhab.binding.elroconnects/src/main/java/org/openhab/binding/elroconnects/internal/discovery/ElroConnectsDiscoveryService.java
```diff
@@ -106,14 +108,15 @@ protected void startBackgroundDiscovery() {
...
-        if (job != null) {
+        if (job != null && !job.isCancelled()) {
             job.cancel(true);
             discoveryJob = null;
         }
     }
 
     @Override
     public void deactivate() {
+        removeOlderResults(new Date().getTime());
```
jlaur, line 119: Newer API can be used:
```suggestion
        removeOlderResults(Instant.now().toEpochMilli());
```

bundles/org.openhab.binding.elroconnects/src/main/java/org/openhab/binding/elroconnects/internal/handler/ElroConnectsBridgeHandler.java
```diff
@@ -528,7 +544,21 @@ private void processAlarmTriggerMessage(ElroConnectsMessage message) {
...
+        // Also trigger an alarm on the bridge, so the alarm also comes through when no thing for the device is
+        // configured
+        triggerChannel(ALARM, Integer.toString(deviceId));
         logger.debug("Device ID {} alarm", deviceId);
+
+        if (answerContent.length() < 22) {
+            logger.debug("Could not get device status from alarm message for device {}", deviceId);
+            return;
+        }
+        String deviceStatus = answerContent.substring(14, 22);
```
jlaur, line 556: I don't know how this string looks, but interpretation seems to carry some assumptions. Can it be sanity checked somehow?

jlaur (changes requested):
Thanks, the code looks really good. Only a few comments to look into.

bundles/org.openhab.binding.elroconnects/README.md
```diff
@@ -44,89 +58,114 @@ It will not be possible to receive alarms and control them from openHAB in this
...
+The `account` bridge thing does not have any channels.
+
 ### K1 connector hub
 
 The `connector` bridge thing has only one channel:
 
-| Channel ID         | Item Type            | Access Mode | Description                                        |
-|--------------------|----------------------|:-----------:|----------------------------------------------------|
-| `scene`            | String               | RW          | current scene                                      |
+| Channel ID         | Item Type      | Access Mode | Advanced | Description                                        |
```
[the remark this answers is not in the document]
mherwege: That should indeed not be there.

bundles/org.openhab.binding.elroconnects/src/main/java/org/openhab/binding/elroconnects/internal/handler/ElroConnectsBridgeHandler.java
mherwege: The content of the device status string is content specific. It is mapped and sanity checked in the ElroConnectsDevice class and its concrete subclasses. I can only pass on the info here if I don't want to bring all the device specific logic in this class.

bundles/org.openhab.binding.elroconnects/src/main/java/org/openhab/binding/elroconnects/internal/handler/ElroConnectsAccountHandler.java
```diff
@@ -0,0 +1,299 @@
...
+                            final HttpResponse response = (HttpResponse) result.getResponse();
+                            if (response.getStatus() == 200) {
+                                f.complete(getContentAsString());
+                            } else if (response.getStatus() == 401) {
+                                // Access token expired, so clear it
+                                accessToken = null;
+                                f.completeExceptionally(
+                                        new ElroConnectsAccountException("@text/offline.token-expired"));
+                            } else {
+                                f.completeExceptionally(new ElroConnectsAccountException(getContentAsString()));
```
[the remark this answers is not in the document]
mherwege: Yes, probably useful. Please check if the way I do it is OK.

bundles/org.openhab.binding.elroconnects/README.md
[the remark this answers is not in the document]
mherwege: It was actually intended to be on the parameters, not the channels. It makes the parameter tables consistent.

bundles/org.openhab.binding.elroconnects/README.md
```diff
@@ -44,89 +58,114 @@ It will not be possible to receive alarms and control them from openHAB in this
...
 
-## Door/window contact
+### Door/window contact
 
 The `entrysensor` thing has the following channels:
 
 | Channel ID         | Item Type            | Access Mode | Description                                        |
 |--------------------|----------------------|:-----------:|----------------------------------------------------|
 | `entry`            | Contact              | R           | open/closed door/window                            |
+| `signal`            | Number               | R           | signal strength between 0 and 4, higher is stronger |
```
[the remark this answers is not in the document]
mherwege: I put the whole README in a an online markup editor and aligned all the tables (except the end of the last column for long descriptions. They come out fine.

bundles/org.openhab.binding.elroconnects/README.md
```diff
@@ -44,89 +58,114 @@ It will not be possible to receive alarms and control them from openHAB in this
...
 
-## Motion sensor
+### Motion sensor
 
 The `motionsensor` thing has the following channels:
 
 | Channel ID         | Item Type            | Access Mode | Description                                        |
 |--------------------|----------------------|:-----------:|----------------------------------------------------|
 | `motion`           | Switch               | R           | on when motion detected                            |
+| `signal`            | Number               | R           | signal strength between 0 and 4, higher is stronger |
```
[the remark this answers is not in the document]
mherwege: See above.

bundles/org.openhab.binding.elroconnects/src/main/java/org/openhab/binding/elroconnects/internal/discovery/ElroConnectsBridgeDiscoveryService.java
mherwege: Thanks.

bundles/org.openhab.binding.elroconnects/src/main/java/org/openhab/binding/elroconnects/internal/discovery/ElroConnectsDiscoveryService.java
mherwege: Also adjusted in ElroConnectsBridgeDiscoveryService

mherwege:
@jlaur Thank you very much for your fast review. I have made adjustments and left a few comments. Can you check?

bundles/org.openhab.binding.elroconnects/README.md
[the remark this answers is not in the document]
jlaur: It looks fine now. It was rendered correctly before also, but looked messy in a text editor.

bundles/org.openhab.binding.elroconnects/src/main/java/org/openhab/binding/elroconnects/internal/handler/ElroConnectsAccountHandler.java
```diff
@@ -0,0 +1,300 @@
...
+                                    Map<String, String> content = gson.fromJson(getContentAsString(), loginType);
+                                    String accessToken = (content != null) ? content.get("access_token") : null;
+                                    f.complete(accessToken);
+                                } catch (JsonParseException parseException) {
+                                    logger.error("Parsing failed: {}", parseException.getMessage());
+                                    f.completeExceptionally(
+                                            new ElroConnectsAccountException("@text/offline.login-parsing-fail"));
+                                }
+                            } else {
+                                f.completeExceptionally(new ElroConnectsAccountException(getContentAsString()));
```
jlaur, line 227: If I'm reading this correctly, this exception is caught in `login()` and full content is set as `ThingStatusDetail`? This could possibly be quite long and/or technical?

bundles/org.openhab.binding.elroconnects/src/main/java/org/openhab/binding/elroconnects/internal/handler/ElroConnectsAccountHandler.java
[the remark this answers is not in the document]
jlaur: I found one more similar place, please see separate comment.

bundles/org.openhab.binding.elroconnects/src/main/java/org/openhab/binding/elroconnects/internal/discovery/ElroConnectsBridgeDiscoveryService.java
```diff
@@ -0,0 +1,150 @@
...
+
+    private volatile @Nullable ScheduledFuture<?> discoveryJob;
+
+    private static final int TIMEOUT_S = 5;
+    private static final int REFRESH_INTERVAL_S = 60;
+
+    public ElroConnectsBridgeDiscoveryService() {
+        super(ElroConnectsBindingConstants.SUPPORTED_CONNECTOR_TYPES_UIDS, TIMEOUT_S);
+        logger.debug("Bridge discovery service started");
+        super.activate(null); // Makes sure the background discovery for devices is enabled
```
jlaur, line 57: This looks like a hack. I think you should call the constructor in super class taking three parameters, where last parameter is `backgroundDiscoveryEnabledByDefault` (bool). The framework will then call `startBackgroundDiscovery` when the component is activated.

bundles/org.openhab.binding.elroconnects/src/main/java/org/openhab/binding/elroconnects/internal/handler/ElroConnectsBridgeHandler.java
jlaur: Hmm, you mean `deviceStatus` is passed to the concrete device classes below? I'm referring to the original string, `answerContent`. What are in the first skipped characters? How do you know how many to skip? Is it ASCII content only?

fwolter:
The title of this PR will show up in the changelog. Can you put the main points in the title so that anybody reading it without context gets a clue what this is about?

bundles/org.openhab.binding.elroconnects/src/main/java/org/openhab/binding/elroconnects/internal/handler/ElroConnectsBridgeHandler.java
mherwege: I know by looking at the wiresharks of the responses and other code available on the internet. So far, I have only been able to observe ASCII strings. Characters 6 to 10 are relevant and interpreted before (deviceId). 10 to 14 contains equipment type, but that is already known in the binding, so not relevant. Not everything in the message format is always clear, but this interpretation has worked so far.

bundles/org.openhab.binding.elroconnects/src/main/java/org/openhab/binding/elroconnects/internal/discovery/ElroConnectsBridgeDiscoveryService.java
mherwege: If you look at the declaration of the super constructor, it looks like it uses a default of true for this anyway, so making the call with the parameter set will not make a difference:
```
    public AbstractDiscoveryService(@Nullable Set<ThingTypeUID> supportedThingTypes, int timeout)
            throws IllegalArgumentException {
        this(supportedThingTypes, timeout, true);
    }
```

I don't know exactly where I got the description of this hack anymore, but I found it used 29 times in the repository.

bundles/org.openhab.binding.elroconnects/src/main/java/org/openhab/binding/elroconnects/internal/handler/ElroConnectsAccountHandler.java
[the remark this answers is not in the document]
mherwege: Commented on the similar line.

bundles/org.openhab.binding.elroconnects/src/main/java/org/openhab/binding/elroconnects/internal/handler/ElroConnectsAccountHandler.java
mherwege: I revisited it. I have heard conflicting messages before, sometimes people wanting to have all info in the UI if possible, sometimes favouring detailed info in the logs.
I now just put a general messages in the thing status, and have all the details when appropriate on warn level in the logs (I had some error messages in the log before, which I probably should not do). I hope this makes more sense. And thank you for pointing it out.

bundles/org.openhab.binding.elroconnects/src/main/java/org/openhab/binding/elroconnects/internal/discovery/ElroConnectsBridgeDiscoveryService.java
mherwege: Looking at it again, I should probably call startBackgroundDiscovery directly, without the activate method. The discovery configuration parameters are irrelevant anyway. I could expose an enableBackgroundDiscovery parameter on the thing to switch it on or off (like is the default for discovery configurations using the discovery service). @jlaur WDYT?

mherwege: I implemented this idea. It turns out to be more tricky than expected. The discovery is activated before the thing handler is initialized, leading to changes in the config not being visible (without extra logic). I could implement that as well, but need to think about it.

bundles/org.openhab.binding.elroconnects/src/main/java/org/openhab/binding/elroconnects/internal/handler/ElroConnectsAccountHandler.java
```diff
@@ -0,0 +1,332 @@
...
+        });
+    }
+
+    private CompletableFuture<@Nullable String> getAccessToken() {
+        CompletableFuture<@Nullable String> f = new CompletableFuture<>();
+
+        // The method call returns an invalid 401 response on authentication error, missing the www-authentication
+        // header. This header should be there according to RFC7235. This workaround removes the protocol handler and
+        // the check.
+        client.getProtocolHandlers().remove(WWWAuthenticationProtocolHandler.NAME);
```
mherwege, line 222: @jlaur Is this acceptable?

bundles/org.openhab.binding.elroconnects/src/main/java/org/openhab/binding/elroconnects/internal/handler/ElroConnectsAccountHandler.java
jlaur: Looks good. I don't think there are strict rules: https://www.openhab.org/docs/developer/bindings/#updating-the-thing-status. It can be technical - as in your previous implementation, so that part of my comments was misleading. However, my comment/question was mainly if `getContentAsString()` could risk being anything (like a long HTML page), in which case it would probably not be a good idea to directly use this.

bundles/org.openhab.binding.elroconnects/src/main/java/org/openhab/binding/elroconnects/internal/handler/ElroConnectsBridgeHandler.java
jlaur: Okay, seems like a crucial part of the protocol which you have fully under control, so let's leave it like it is. :)

bundles/org.openhab.binding.elroconnects/src/main/java/org/openhab/binding/elroconnects/internal/handler/ElroConnectsAccountHandler.java
jlaur: I don't have an opinion about this or ideas how to approach it differently. If this is needed for the desired behavior, I'm fine with it.

bundles/org.openhab.binding.elroconnects/src/main/java/org/openhab/binding/elroconnects/internal/discovery/ElroConnectsBridgeDiscoveryService.java
```diff
@@ -0,0 +1,166 @@
...
+    private static final int REFRESH_INTERVAL_S = 60;
+
+    public ElroConnectsBridgeDiscoveryService() {
+        super(ElroConnectsBindingConstants.SUPPORTED_CONNECTOR_TYPES_UIDS, TIMEOUT_S);
+        logger.debug("Bridge discovery service started");
+    }
+
+    @Override
+    protected void startScan() {
+        scheduler.submit(this::discoverConnectors); // If background account polling is not enabled for the handler,
```
jlaur, line 64: Since it seems to be "fire and forget", `execute` could be considered.

bundles/org.openhab.binding.elroconnects/src/main/java/org/openhab/binding/elroconnects/internal/handler/ElroConnectsAccountHandler.java
```diff
@@ -0,0 +1,332 @@
...
+            return;
+        }
+
+        login.put("username", username);
+        login.put("password", password);
+        loginJson = gson.toJson(login);
+
+        // Do initial login to retrieve access token and start polling with small initial delay if background discovery
+        // is enabled), so the access token is available
+        scheduler.submit(this::login);
```
jlaur, line 125: Just question: Is there any reason to explicitly attempt login here instead of having the job perform this without any initial delay? If background discovery is disabled, what will/should happen if this one-time login fails (for example due to temporary network problem)?

mherwege: This one time login (background discovery disabled) will take the thing online when the login is successful, or offline with error when not. That will be visible in the UI as a timeout exception. So a communication error should be apparent. The way to recover would be to disable and enable the thing to restart initialization, no automatic recovery. If background discovery is enabled, it will try to login again on the next iteration, 5 min. later.
As the login and polling to the cloud service is only really required for discovery, by disabling background discovery the cloud call every 5 min. is avoided. Manually triggering a scan will just do one cycle of cloud communication.

Do you think I should automate a retry of the initial login when there is a network problem? I see that as separate from background discovery.

bundles/org.openhab.binding.elroconnects/src/main/java/org/openhab/binding/elroconnects/internal/handler/ElroConnectsAccountHandler.java
mherwege: @jlaur After reconsideration, what I did is not a good idea. This will change the behaviour of the common http client and disable all BASIC and DIGEST automatic authentication handling on the common http client. By consequence other bindings using this automatic handling will fail.
I will change it, creating a dedicated http client in the handler factory.

bundles/org.openhab.binding.elroconnects/src/main/java/org/openhab/binding/elroconnects/internal/discovery/ElroConnectsBridgeDiscoveryService.java
mherwege: Will do.

bundles/org.openhab.binding.elroconnects/src/main/java/org/openhab/binding/elroconnects/internal/discovery/ElroConnectsBridgeDiscoveryService.java
mherwege: Modified it so background discovery is enabled/disabled from the handler, instead of discovery activation. That keeps it in sync and makes it possible to remove all initial discovery delays.

mherwege:
@jlaur I did 2 improvements in the last commit:
- properly handle disabling authentication header check, using a separate httpClient, to avoid issues with other consumers of the common httpClient that rely on the feature.
- Reworked how background discovery is started, so it gets started from the thing handler. This allows starting it when the handler is fully ready, and makes discovery immediate an smooth.

Short term network problems on background discovery should recover nicely. For one-off scans, they will fail if there is a network issue, but that is visible to the user in the account thing status.

bundles/org.openhab.binding.elroconnects/src/main/java/org/openhab/binding/elroconnects/internal/handler/ElroConnectsAccountHandler.java
```diff
@@ -156,71 +167,81 @@ private void stopPolling() {
...
+                    }
                 }
-            }
 
-            return null;
-        });
+                return null;
+            }).get();
+        } catch (InterruptedException e) {
+            updateStatus(ThingStatus.OFFLINE);
```
jlaur, line 236: I think you should call `Thread.currentThread().interrupt();` in this case.

bundles/org.openhab.binding.elroconnects/src/main/java/org/openhab/binding/elroconnects/internal/handler/ElroConnectsAccountHandler.java
```diff
@@ -156,71 +167,81 @@ private void stopPolling() {
...
```
jlaur, line 210: I think you should call `Thread.currentThread().interrupt();` in this case.

bundles/org.openhab.binding.elroconnects/src/main/java/org/openhab/binding/elroconnects/internal/handler/ElroConnectsAccountHandler.java
jlaur: Handled in [other comment](https://github.com/openhab/openhab-addons/pull/12653#issuecomment-1116174501), resolving this one.

jlaur:
> I did 2 improvements in the last commit:
> 
> * properly handle disabling authentication header check, using a separate httpClient, to avoid issues with other consumers of the common httpClient that rely on the feature.
> * Reworked how background discovery is started, so it gets started from the thing handler. This allows starting it when the handler is fully ready, and makes discovery immediate an smooth.
> 
> Short term network problems on background discovery should recover nicely. For one-off scans, they will fail if there is a network issue, but that is visible to the user in the account thing status.

Agreed - looks good!

bundles/org.openhab.binding.elroconnects/src/main/java/org/openhab/binding/elroconnects/internal/handler/ElroConnectsAccountHandler.java
mherwege: Done

bundles/org.openhab.binding.elroconnects/src/main/java/org/openhab/binding/elroconnects/internal/handler/ElroConnectsAccountHandler.java
mherwege: Done

mherwege:
@jlaur I made the last adjustments you requested.

jlaur:
> @jlaur I made the last adjustments you requested.

Well, let's get it merged then. :) Thanks for the contribution.

jlaur (approved):
LGTM

PR #12657, opened 2022-04-27
Update README.md

schmieeed:
Comma's are missing in the thing config + add optinal name placeholder


schmieeed (commented):
second try with signature

PR #12665, opened 2022-04-29
Bump commons-io

J-N-K:
This brings commons-io in-line with version in core.

Exclusion of `sun.nio.ch.*` is safe because it is only needed for classes available since 2.9.0 (which is above the version used before)


jlaur (approved):
Thanks.

PR #12675, opened 2022-05-01
[knx] Allow sending items with units to KNX bus.

### LINKED ISSUE #10706: [KNX] Items with Dimension are not written to the bus.
## Expected Behavior
When sending a value to an item with a dimension the corresponding should be written to the bus.  

## Current Behavior
Value sent to an item with a dimension are not written to the KNX bus.

## Possible Solution
All dimension types should be supported by the binding. Possible solution is mentioned here [https://community.openhab.org/t/can-not-write-value-to-knx-oh3/122363/19](https://community.openhab.org/t/can-not-write-value-to-knx-oh3/122363/19)

Sending a value to an item defined as follows will fail to send the value to the KNX bus.
`Number:Temperature Temperature "Temp [%.2f %unit%]" <temperature> {channel="knx:device:multicast:modes:temperature"}`

Sending a value to an item defined without the dimension will update the KNX bus correctly.
`Number Temperature "Temp [%.2f %unit%]" <temperature> {channel="knx:device:multicast:modes:temperature"}`

Issue also occurs for Number:Time dimension.

## Context
Item dimensions are well documented and their use helps improve the usability of a configured openHAB system.  therefore is would be sensible for the KNX binding to support dimensions. 

## Your Environment
OpenHAB 3.1.0.M4 and earlier (OH2)

holgerfriedrich:
Items with dimensions (QuantityType) are now translated and can be sent to the
KNX bus. This requires the correct DPT to be specified in the channel
definition. Fixes #10706.

### DIFF
--- a/bundles/org.openhab.binding.knx/README.md
+++ b/bundles/org.openhab.binding.knx/README.md
@@ -117,6 +117,10 @@ Note: After changing the DPT of already existing Channels, openHAB needs to be r
 |-----------|---------------|-------------|
 | ga        | Group address | 9.001       |
 
+
+Note: Using the Units Of Measurement feature of openHAB (Quantitytype) requires that the DPT value is set correctly.
+Automatic type conversion will be applied if required.
+
 ##### Channel Type "string"
 
 | Parameter | Description   | Default DPT |
--- a/bundles/org.openhab.binding.knx/src/main/java/org/openhab/binding/knx/internal/channel/KNXChannelType.java
+++ b/bundles/org.openhab.binding.knx/src/main/java/org/openhab/binding/knx/internal/channel/KNXChannelType.java
@@ -31,6 +31,8 @@ import org.openhab.binding.knx.internal.KNXTypeMapper;
 import org.openhab.binding.knx.internal.client.InboundSpec;
 import org.openhab.binding.knx.internal.client.OutboundSpec;
 import org.openhab.core.config.core.Configuration;
+import org.openhab.core.library.types.DecimalType;
+import org.openhab.core.library.types.QuantityType;
 import org.openhab.core.types.Type;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -171,9 +173,10 @@ public abstract class KNXChannelType {
                 }
                 Class<? extends Type> expectedTypeClass = typeHelper.toTypeClass(dpt);
                 if (expectedTypeClass != null) {
-                    if (expectedTypeClass.isInstance(command)) {
+                    if (expectedTypeClass.isInstance(command)
+                            || ((expectedTypeClass == DecimalType.class) && (command instanceof QuantityType))) {
                         logger.trace(
-                                "getCommandSpec key '{}' uses expectedTypeClass '{}' witch isInstance for command '{}' and dpt '{}'",
+                                "getCommandSpec key '{}' uses expectedTypeClass '{}' which isInstance for command '{}' and dpt '{}'",
                                 key, expectedTypeClass, command, dpt);
                         return new WriteSpecImpl(config, dpt, command);
                     }
--- a/bundles/org.openhab.binding.knx/src/main/java/org/openhab/binding/knx/internal/dpt/KNXCoreTypeMapper.java
+++ b/bundles/org.openhab.binding.knx/src/main/java/org/openhab/binding/knx/internal/dpt/KNXCoreTypeMapper.java
@@ -34,6 +34,7 @@ import org.openhab.core.library.types.IncreaseDecreaseType;
 import org.openhab.core.library.types.OnOffType;
 import org.openhab.core.library.types.OpenClosedType;
 import org.openhab.core.library.types.PercentType;
+import org.openhab.core.library.types.QuantityType;
 import org.openhab.core.library.types.StopMoveType;
 import org.openhab.core.library.types.StringType;
 import org.openhab.core.library.types.UpDownType;
@@ -231,7 +232,6 @@ public class KNXCoreTypeMapper implements KNXTypeMapper {
          * 7.011: DPT_Length_mm values: 0...65535 mm
          * 7.012: DPT_UElCurrentmA values: 0...65535 mA
          * 7.013: DPT_Brightness values: 0...65535 lx
-         * Calimero does not map: (map/use to 7.000 until then)
          * 7.600: DPT_Colour_Temperature values: 0...65535 K, 2000K 3000K 5000K 8000K
          */
         dptMainTypeMap.put(7, DecimalType.class);
@@ -250,6 +250,7 @@ public class KNXCoreTypeMapper implements KNXTypeMapper {
          * 8.007: DPT_DeltaTimeHrs
          * 8.010: DPT_Percent_V16
          * 8.011: DPT_Rotation_Angle
+         * 8.012: DPT_Length_m
          */
         dptMainTypeMap.put(8, DecimalType.class);
 
@@ -275,6 +276,8 @@ public class KNXCoreTypeMapper implements KNXTypeMapper {
          * 9.026: DPT_Rain_Amount values: -671088.64...670760.96 l/m²
          * 9.027: DPT_Value_Temp_F values: -459.6...670760.96 °F
          * 9.028: DPT_Value_Wsp_kmh values: 0...670760.96 km/h
+         * 9.029: DPT_Value_Absolute_Humidity: 0...670760 g/m³
+         * 9.030: DPT_Concentration_μgm3: 0...670760 µg/m³
          */
         dptMainTypeMap.put(9, DecimalType.class);
         /** Exceptions Datapoint Types "2-Octet Float Value", Main number 9 */
@@ -300,6 +303,11 @@ public class KNXCoreTypeMapper implements KNXTypeMapper {
          * MainType: 12
          * 12.000: General unsigned long
          * 12.001: DPT_Value_4_Ucount values: 0...4294967295 counter pulses
+         * 12.100: DPT_LongTimePeriod_Sec values: 0...4294967295 s
+         * 12.101: DPT_LongTimePeriod_Min values: 0...4294967295 min
+         * 12.102: DPT_LongTimePeriod_Hrs values: 0...4294967295 h
+         * 12.1200: DPT_VolumeLiquid_Litre values: 0..4294967295 l
+         * 12.1201: DPT_Volume_m3 values: 0..4294967295 m3
          */
         dptMainTypeMap.put(12, DecimalType.class);
         /** Exceptions Datapoint Types "4-Octet Unsigned Value", Main number 12 */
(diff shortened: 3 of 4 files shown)


bundles/org.openhab.binding.knx/src/main/java/org/openhab/binding/knx/internal/dpt/KNXCoreTypeMapper.java
```diff
@@ -571,6 +583,154 @@ public KNXCoreTypeMapper() {
...
+                        } else if (unit.contains("°F")) {
+                            targetOhUnit = "°F";
+                            scaleFactor = 5.0 / 9.0;
+                        } else if (unit.contains("K")) {
+                            targetOhUnit = "K";
+                        } else {
+                            targetOhUnit = "";
+                        }
+                    }
+                        break;
```
lolodomo, line 623: The more common way to do is I believe to put the break inside the block `{}`
Same for other switch cases

bundles/org.openhab.binding.knx/src/main/java/org/openhab/binding/knx/internal/dpt/KNXCoreTypeMapper.java
```diff
@@ -571,6 +583,154 @@ public KNXCoreTypeMapper() {
         defaultDptMap.put(HSBType.class, DPTXlatorRGB.DPT_RGB.getID());
     }
 
+    /*
+     * This function computes the target unit for type conversion from OH quantity type to DPT types.
+     * Calimero library provides units which can be used for most of the DPTs. There are some deviations
+     * from the OH unit scheme which are handled.
+     */
+    private String quantityTypeToDPTValue(QuantityType<?> qt, int mainNumber, int subNumber, String targetOhUnit)
```
lolodomo, line 591: If I correctly understand how the method is called, the last parameter is in fact a DP unit.
I would suggest to name the parameter `dpUnit` and then create a local variable named targetOhUnit that you will initialize with `dpUnit`

bundles/org.openhab.binding.knx/src/main/java/org/openhab/binding/knx/internal/dpt/KNXCoreTypeMapper.java
```diff
@@ -571,6 +583,154 @@ public KNXCoreTypeMapper() {
...
+                            targetOhUnit = "K/h";
+                        } else {
+                            targetOhUnit = "";
+                        }
+                    }
+                        break;
+                    case 23: {
+                        final String unit = qt.getUnit().toString();
+                        if (unit.contains("°C")) {
+                            targetOhUnit = "°C/%";
```
lolodomo, line 641: What is this unit ?

bundles/org.openhab.binding.knx/src/main/java/org/openhab/binding/knx/internal/dpt/KNXCoreTypeMapper.java
```diff
@@ -571,6 +583,154 @@ public KNXCoreTypeMapper() {
...
+                        break;
+                }
+                break;
+        }
+        // replace e.g. m3 by m³
+        targetOhUnit = targetOhUnit.replace("3", "³").replace("2", "²");
+
+        final QuantityType<?> result = qt.toUnit(targetOhUnit);
+        if (result == null)
+            throw new KNXException("incompatible types: " + qt.getUnit().toString() + ", " + targetOhUnit);
```
lolodomo, line 730: Please surround with `{}`:
```
       if (result == null) {
            throw new KNXException("incompatible types: " + qt.getUnit().toString() + ", " + targetOhUnit);
       }
```

lolodomo:
@sjka : would you like to review ?

bundles/org.openhab.binding.knx/src/main/java/org/openhab/binding/knx/internal/dpt/KNXCoreTypeMapper.java
holgerfriedrich: done

bundles/org.openhab.binding.knx/src/main/java/org/openhab/binding/knx/internal/dpt/KNXCoreTypeMapper.java
holgerfriedrich: done

bundles/org.openhab.binding.knx/src/main/java/org/openhab/binding/knx/internal/dpt/KNXCoreTypeMapper.java
holgerfriedrich: 9.023 DPT_KelvinPerPercent, this is a delta and we need to convert to the matching base unit first (automatic conversion °C to K or °F will fail due to the offset/scaling)

holgerfriedrich:
@lolodomo thanks for reviewing this

holgerfriedrich:
@lolodomo I have the feeling that our CODEOWNERS file is outdated.... @sjka joined github 2021....

@sjsf  Did you initially submit the code for the knx binding?

bundles/org.openhab.binding.knx/src/main/java/org/openhab/binding/knx/internal/dpt/KNXCoreTypeMapper.java
lolodomo: In fact, my question was: is "°C/%" is a valid unit supported by UoM ?

holgerfriedrich: Actually is seem so. DSL rules let you construct this
```
        var k = 25|K
        var p = 10|%
        var result = k/p
        var dimensionless = p.toUnit("")
```
Printing to the log gets you
```
...script.UoM   ] - this is k: 25 K
...script.UoM   ] - this is p: 10 %
...script.UoM   ] - this is k/p: 2.5 K/%
...script.UoM   ] - this is p converted to dimensionless: 0.1
```
The values are sent to the KNX bus when I use sendCommand, both tested with items defined as Number and Number:Temperature. It leads to same value on KNX side.

lolodomo: Ok, strange but if you tested it, I believe you.

lolodomo:
> @lolodomo I have the feeling that our CODEOWNERS file is outdated.... @SJKA joined github 2021....
> 
> @sjsf Did you initially submit the code for the knx binding?

I see "Simon Kaufmann" as author of a lot of classes. Is sjka =Simon Kaufmann ? This is probable...

@kaikreuzer : do you know if there still a maintainer for the KNX maintainer ? If not, I will merge this PR without waiting.

holgerfriedrich:
@lolodomo but to meet it seems that GitHub re-issued this user account to someone else... Just have a look at the two user profiles mentioned above

kaikreuzer:
@lolodomo Yes, I think Simon is no longer active here - I have thus created https://github.com/openhab/openhab-addons/pull/12694 and added me as a code owner as I was the original author of the 1.x binding (the first binding ever written for openHAB ;-)).
If you approve the changes, I can have a short second glance over it, so that we can merge it.

lolodomo (approved):

lolodomo:
> If you approve the changes, I can have a short second glance over it, so that we can merge it.

Yes please

bundles/org.openhab.binding.knx/README.md
```diff
@@ -117,6 +117,9 @@ Note: After changing the DPT of already existing Channels, openHAB needs to be r
 |-----------|---------------|-------------|
 | ga        | Group address | 9.001       |
 
+
+Note: Using the Units Of Measurement feature of OpenHAB (Quantitytype) requires that the DPT value is set correctly. Automatic type conversion will be applied if required.
```
kaikreuzer, line 121:
```suggestion
Note: Using the Units Of Measurement feature of openHAB (QuantityType) requires that the DPT value is set correctly.
Automatic type conversion will be applied if required.
```

bundles/org.openhab.binding.knx/src/main/java/org/openhab/binding/knx/internal/channel/KNXChannelType.java
```diff
@@ -173,11 +174,17 @@ protected final Set<String> asSet(String... values) {
                 if (expectedTypeClass != null) {
                     if (expectedTypeClass.isInstance(command)) {
                         logger.trace(
-                                "getCommandSpec key '{}' uses expectedTypeClass '{}' witch isInstance for command '{}' and dpt '{}'",
+                                "getCommandSpec key '{}' uses expectedTypeClass '{}' which isInstance for command '{}' and dpt '{}'",
                                 key, expectedTypeClass, command, dpt);
                         return new WriteSpecImpl(config, dpt, command);
                     }
                 }
+                if (command instanceof QuantityType) {
```
kaikreuzer, line 182: Hm, is it really correct to do an independent "if" here and not check at all the dpt anymore?
I would expect the `typeHelper` in line 173 to return `DecimalType` as the `expectedTypeClass`.
If this is the case, we would probably only need to adapt line 175 to
```
                    if (expectedTypeClass.isAssignableFrom(command.getClass())) {
```
so that it also evaluates to true, if `command` is a QuantityType and not only if it is a DecimalType.

Wdyt? Am I missing something?

kaikreuzer (changes requested):

bundles/org.openhab.binding.knx/src/main/java/org/openhab/binding/knx/internal/channel/KNXChannelType.java
holgerfriedrich: @kaikreuzer I just did a quick check: ```expectedTypeClass``` returns ```DecimalType```, ```command.getClass()```is ```QuantityType``` and ```isAssignableFrom``` returns false.

I don't think that we nedd to check the dpt - the type conversion will anyway only work if the unit conversion is possible.

kaikreuzer: You're right, I forgot that `QuantityType` does not extend `DecimalType`. 😞 
I would nonetheless prefer to do a proper check of the type in the if clause.
If `expectedTypeClass` is null, we should never proceed, so the code should imho move into the if statement above.
And then we could simply enhance line 175 to
```
                    if (expectedTypeClass.isInstance(command) || (expectedTypeClass==DecimalType.class && command instanceof QuantityType<?>)) {
```
(and slightly change the tracing message).

holgerfriedrich: This looks reasonable, updated code as requested.
Thanks Kai!

kaikreuzer (approved):
Great, thank you!

butterfly2501:
I debug the component to send items with unit yesterday because my humidity sensor was not working and always got getCommandSpec no Spec found.
It was never working wih the dpt 9.007 with or without unit. So my workaround is now with a seperate field and to change the value from % to °C and use instead dpt 9.001. Can you please check in the code if the unit for humidity % is correct processed.

PR #12679, opened 2022-05-03
[orbitbhyve] do not update status of a disabled sprinker

octa22:
This PR fixes a rare bug when the bridge sets the sprinkler's status to ONLINE/OFFLINE even if the sprinkler is disabled.

Signed-off-by: Ondrej Pecta <opecta@gmail.com>


lolodomo (approved):
LGTM

lolodomo:
Interesting, I imagine there is the same issue in many other bindings.

lolodomo:
@octa22 : in fact, I have two questions.
Has a disabled thing a non null thing handler ?
And what is the default thing status of a disabled thing ?
Maybe you know the answers.

octa22:
Yes, what I was experiencing was an ONLINE thing with a handler, but only constructor was called, no initialization. After an investigation I found that the sprinkler was disabled, but the bridge set its status to ONLINE according to a cloud status.
The disabled thing sets its state to UNINITIALIZED and extended status (DISABLED)

lolodomo:
> The disabled thing sets its state to UNINITIALIZED and extended status (DISABLED)

Thank you for your fast answer. It makes sense.

PR #12680, opened 2022-05-03
[boschshc] Support Bosch Smart Home Cameras (#12666)

### LINKED ISSUE #12666: [boschshc] Support for Bosch Security Cameras
Hi, I noticed that the Bosch security cameras (360° indoor camera and Eyes outdoor camera) are not supported by the Bosch Smart Home binding yet.

According to https://apidocs.bosch-smarthome.com/local/, both cameras offer only one service (`PrivacyMode`), which can be enabled or disabled. Not much, but better than nothing ;)

```
GET /devices/{deviceId}/services/PrivacyMode/state Retrieve the state of the PrivacyMode service.
PUT /devices/{deviceId}/services/PrivacyMode/state Executes the privacyModeState on the device.
```

Is someone already working on this? Otherwise I would consider contributing.

david-pace:
- add thing definitions for indoor security camera 360 and outdoor
security camera Eyes
- implement services for privacy mode and camera notifications
- implement camera handler
- add unit tests
- update documentation

closes #12666

### DIFF
--- a/bundles/org.openhab.binding.boschshc/src/main/java/org/openhab/binding/boschshc/internal/devices/camera/CameraHandler.java
+++ b/bundles/org.openhab.binding.boschshc/src/main/java/org/openhab/binding/boschshc/internal/devices/camera/CameraHandler.java
@@ -1,0 +1,167 @@
+/**
+ * Copyright (c) 2010-2022 Contributors to the openHAB project
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package org.openhab.binding.boschshc.internal.devices.camera;
+
+import static org.openhab.binding.boschshc.internal.devices.BoschSHCBindingConstants.CHANNEL_CAMERA_NOTIFICATION;
+import static org.openhab.binding.boschshc.internal.devices.BoschSHCBindingConstants.CHANNEL_PRIVACY_MODE;
+
+import java.util.List;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.TimeoutException;
+
+import org.eclipse.jdt.annotation.NonNullByDefault;
+import org.eclipse.jdt.annotation.Nullable;
+import org.openhab.binding.boschshc.internal.devices.BoschSHCHandler;
+import org.openhab.binding.boschshc.internal.exceptions.BoschSHCException;
+import org.openhab.binding.boschshc.internal.services.cameranotification.CameraNotificationService;
+import org.openhab.binding.boschshc.internal.services.cameranotification.CameraNotificationState;
+import org.openhab.binding.boschshc.internal.services.cameranotification.dto.CameraNotificationServiceState;
+import org.openhab.binding.boschshc.internal.services.privacymode.PrivacyModeService;
+import org.openhab.binding.boschshc.internal.services.privacymode.PrivacyModeState;
+import org.openhab.binding.boschshc.internal.services.privacymode.dto.PrivacyModeServiceState;
+import org.openhab.core.library.types.OnOffType;
+import org.openhab.core.thing.ChannelUID;
+import org.openhab.core.thing.Thing;
+import org.openhab.core.types.Command;
+
+/**
+ * Handler for security cameras.
+ * <p>
+ * This implementation handles services and commands that are common to all cameras, which are currently:
+ * 
+ * <ul>
+ * <li><code>PrivacyMode</code> - Controls whether the camera records images</li>
+ * <li><code>CameraNotification</code> - Enables or disables notifications for the camera</li>
+ * </ul>
+ * 
+ * <p>
+ * The Eyes outdoor camera advertises a <code>CameraLight</code> service, which unfortunately does not work properly.
+ * Valid states are <code>ON</code> and <code>OFF</code>.
+ * One of my two cameras returns <code>HTTP 204 (No Content)</code> when requesting the state.
+ * Once Bosch supports this service properly, a new subclass may be introduced for the Eyes outdoor camera.
+ * 
+ * @author David Pace - Initial contribution
+ *
+ */
+@NonNullByDefault
+public class CameraHandler extends BoschSHCHandler {
+
+    private PrivacyModeService privacyModeService;
+    private CameraNotificationService cameraNotificationService;
+
+    public CameraHandler(Thing thing) {
+        super(thing);
+        this.privacyModeService = new PrivacyModeService();
+        this.cameraNotificationService = new CameraNotificationService();
+    }
+
+    @Override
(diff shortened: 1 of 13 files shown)


bundles/org.openhab.binding.boschshc/src/main/java/org/openhab/binding/boschshc/internal/devices/camera/CameraHandler.java
```diff
@@ -0,0 +1,167 @@
...
+
+        this.registerService(this.privacyModeService, this::updateChannels, List.of(CHANNEL_PRIVACY_MODE));
+        this.registerService(this.cameraNotificationService, this::updateChannels,
+                List.of(CHANNEL_CAMERA_NOTIFICATION));
+    }
+
+    @Override
+    public void initialize() {
+        super.initialize();
+        requestInitialStates();
```
david-pace, line 80: The long poll results do not include the initial camera states. Is there a more elegent way to obtain the initial states?

bundles/org.openhab.binding.boschshc/src/main/java/org/openhab/binding/boschshc/internal/services/privacymode/PrivacyModeState.java
```diff
@@ -0,0 +1,55 @@
...
+
+    /**
+     * Privacy mode enabled / camera disabled
+     */
+    ENABLED,
+
+    /**
+     * Privacy mode disabled / camera enabled
+     */
+    DISABLED;
```
david-pace, line 35: The API exposes the so called _privacy mode_, which is the opposite of "camera enabled". With the current implementation we expose the API 1:1, meaning if the user switches the privacy mode off, the camera will be on. This might be counterintuitive for the uses, who probably expect that the camera is enabled when the switch is on.

Do you think we should rather offer an inverse logic channel/state? How would you call this channel? Maybe `camera-state`?

coeing:
@david-pace Cool, thanks a lot for your work! 👍 @GerdZanker is the main developer on the binding now, but I will have a quick look at your changes and send some comments if I see something odd :)

bundles/org.openhab.binding.boschshc/src/main/java/org/openhab/binding/boschshc/internal/services/privacymode/PrivacyModeState.java
coeing: I would stay as close to the official API as possible, so keeping the privacy mode here. It may include more than the camera deactivation in the future, maybe it already does. You could add a read-only `camera-state` channel of course which inverts this flag. But writing it should only be possible via the `privacy-mode` channel.

bundles/org.openhab.binding.boschshc/src/main/java/org/openhab/binding/boschshc/internal/devices/camera/CameraHandler.java
coeing: If this becomes more common, we could think about moving the code from your `requestInitial...` methods into the `registerService` method, e.g. with a flag that the initial state should be actively fetched. But if the long poll does not contain the information I do not see a way to get the state otherwise.

coeing (approved):
Very good code quality! 👍 Hope to see more contributions from you for this binding 🙂 

If you have any feedback on the general code architecture of the binding, let us know!

david-pace:
> Hope to see more contributions from you for this binding 🙂

I'm happy to contribute more code in the future, thank you very much for your feedback 🙂

> If you have any feedback on the general code architecture of the binding, let us know!

So far I found everything designed and thought out very well 👍 But I'll let you know if I have any ideas.

jlaur (approved):
Looks really good, I have nothing to add. Thanks! If you are ready to merge, please resolve the two comments and/or confirm.

david-pace:
Thank you @jlaur 👍 I would propose waiting for feedback on the two code comments by at least one of @GerdZanker and/or @stefan-kaestle, if that's fine with you.

GerdZanker:
Hello @david-pace and thank you for contributing to the BSH binding.

Your pull request arrived right in time to check my new development environment. I have to say that was block since a too long time because of my ten year old computer. Sorry for this, especially to @coeing. Now I replaced everything inside my mini-tower and the new hardware works perfectly and very fast. Same as your code - I was able to test your new eyes camera handler successfully with my hardware.

So please just check the logging statements to finish the pr.

GerdZanker (changes requested):

stefan-kaestle:
Awesome stuff, thanks @david-pace!

bundles/org.openhab.binding.boschshc/src/main/java/org/openhab/binding/boschshc/internal/devices/camera/CameraHandler.java
david-pace: ok, then I will leave it as it is for now and it can be generalized later if necessary.

bundles/org.openhab.binding.boschshc/src/main/java/org/openhab/binding/boschshc/internal/services/privacymode/PrivacyModeState.java
david-pace: I decided to go only for the `privacy-mode` channel in the initial version that exposes the API 1:1. If required, other channels can be added later.

bundles/org.openhab.binding.boschshc/src/main/java/org/openhab/binding/boschshc/internal/devices/camera/CameraHandler.java
```diff
@@ -0,0 +1,167 @@
...
+    private void requestInitialPrivacyState() {
+        try {
+            @Nullable
+            PrivacyModeServiceState serviceState = privacyModeService.getState();
+            if (serviceState != null) {
+                super.updateState(CHANNEL_PRIVACY_MODE, serviceState.value.toOnOffType());
+            }
+        } catch (InterruptedException e) {
+            Thread.currentThread().interrupt();
+            logger.error("Error while requesting initial privacy state of camera {}", getBoschID(), e);
```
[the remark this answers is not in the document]
david-pace: Thanks for calling my attention to the logging guidelines. In this case this would indeed not be a bug and the log level `error` is not applicable here. I changed the log level for all logger calls in those methods to `debug` now. I am not sure whether a stack trace would be allowed or even useful on debug level. In my current commit I have omitted the stack trace. If you are fine with that you can resolve this comment.

david-pace:
Thank you very much for your feedback, @GerdZanker and @stefan-kaestle 😃 I pushed a commit that adresses the logging enhancements 👍

bundles/org.openhab.binding.boschshc/src/main/java/org/openhab/binding/boschshc/internal/devices/camera/CameraHandler.java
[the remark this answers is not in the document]
jlaur: It's really fine without stack trace, especially for debug level. Nice catch about error level, @GerdZanker, I somehow missed that.

Issue #12688, opened 2022-05-05
[openuv] not able to see any SafeExposure value per Fitzpatrick Skin Type after migration from 3.3.0 snapshot to 3.3.0 M3. File config


nesys:
## Expected Behavior

With the config below in openuv.things, I should be able to receive by openuv a value in minutes for each of six safe exposures per Fitzpatrick Skin Type.

`Bridge openuv:openuvapi:local "OpenUV API" @ "Accounts" [ apikey="myAPIkey" ] {

    Thing uvreport milan "Local UV" @ "Outdoor" [ location="myLat,myLong", refresh=20 ] {
        Channels:
            Type SafeExposure : pale [       
                index="I"
            ]
            Type SafeExposure : white [
                index="II"
            ]
            Type SafeExposure : light_brown [
                index="III"
            ]
            Type SafeExposure : moderate_brown [
                index="IV"
            ]
            Type SafeExposure : dark_brown [
                index="V"
            ]
            Type SafeExposure : black [
                index="VI"
            ]
    }
}
`

## Current Behavior

After the migration from snapshot to 3.3.0 M3, the Binding is working great. But I'm not able to receive any data for each SafeExposure item anymore.

But if I try a normal curl request, like this:

`/usr/bin/curl -s -XGET "https://api.openuv.io/api/v1/uv?lat=mylat&lng=mylong" -H "x-access-token: myAPItoken" > openuv.txt
`
I see a value for each SafeExposure (st1, st2, st3, st4, st5 and st6).

## Steps to Reproduce (for Bugs)

In 3.3.0 M3 (and also M4), configure manually the Openuv Binding, via config files (things file with the config above, and items file like the following:

`
Number:Time Openuv_SafeExposure1            "Safe Exposure I [%d %unit%]"                   (gWeather)                            { channel = "openuv:uvreport:local:milan:pale" }
`

## Context
I would like to collect all the six values, one for each Fitzpatrick Skin Type, as exposed by 

result.safe_exposure_time.st1
result.safe_exposure_time.st2
result.safe_exposure_time.st3
result.safe_exposure_time.st4
result.safe_exposure_time.st5
result.safe_exposure_time.st6

Community topic: https://community.openhab.org/t/new-warn-in-log-since-fresh-3-0-1/119012/28?u=ariela

## Your Environment
openhabian 3.3.0 M4

nesys:
Hi @clinique 
[...]

PR #12695, opened 2022-05-07
[openuv] adressing issue #12688

clinique:

### DIFF
--- a/bundles/org.openhab.binding.openuv/src/main/java/org/openhab/binding/openuv/internal/handler/OpenUVBridgeHandler.java
+++ b/bundles/org.openhab.binding.openuv/src/main/java/org/openhab/binding/openuv/internal/handler/OpenUVBridgeHandler.java
@@ -23,8 +23,6 @@ import java.util.Set;
 import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.TimeUnit;
 
-import javax.ws.rs.HttpMethod;
-
 import org.eclipse.jdt.annotation.NonNullByDefault;
 import org.eclipse.jdt.annotation.Nullable;
 import org.openhab.binding.openuv.internal.OpenUVException;
@@ -118,7 +116,7 @@ public class OpenUVBridgeHandler extends BaseBridgeHandler {
         String url = String.format(QUERY_URL, latitude, longitude, altitude);
         String jsonData = "";
         try {
-            jsonData = HttpUtil.executeUrl(HttpMethod.GET, url, header, null, null, REQUEST_TIMEOUT_MS);
+            jsonData = HttpUtil.executeUrl("GET", url, header, null, null, REQUEST_TIMEOUT_MS);
             OpenUVResponse uvResponse = gson.fromJson(jsonData, OpenUVResponse.class);
             if (uvResponse != null) {
                 String error = uvResponse.getError();
--- a/bundles/org.openhab.binding.openuv/src/main/java/org/openhab/binding/openuv/internal/handler/OpenUVBridgeHandler.java
+++ b/bundles/org.openhab.binding.openuv/src/main/java/org/openhab/binding/openuv/internal/handler/OpenUVBridgeHandler.java
@@ -94,7 +94,7 @@ public class OpenUVBridgeHandler extends BaseBridgeHandler {
 
     @Override
     public void dispose() {
-        reconnectJob.ifPresent(job -> job.cancel(false));
+        reconnectJob.ifPresent(job -> job.cancel(true));
         reconnectJob = Optional.empty();
     }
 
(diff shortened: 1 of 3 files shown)


bundles/org.openhab.binding.openuv/src/main/java/org/openhab/binding/openuv/internal/handler/OpenUVBridgeHandler.java
```diff
@@ -17,11 +17,14 @@
 import java.time.LocalDate;
 import java.time.LocalDateTime;
 import java.util.Collection;
+import java.util.Optional;
 import java.util.Properties;
 import java.util.Set;
 import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.TimeUnit;
 
+import javax.ws.rs.HttpMethod;
```
lolodomo, line 26: I would suggest to rather use org.eclipse.jetty.http.HttpMethod

bundles/org.openhab.binding.openuv/src/main/java/org/openhab/binding/openuv/internal/handler/OpenUVBridgeHandler.java
```diff
@@ -93,11 +96,8 @@ public void initialize() {
 
     @Override
     public void dispose() {
-        ScheduledFuture<?> job = this.reconnectJob;
-        if (job != null && !job.isCancelled()) {
-            job.cancel(true);
-        }
-        reconnectJob = null;
+        reconnectJob.ifPresent(job -> job.cancel(false));
```
lolodomo, line 99: You switch from true to false. Why ?

lolodomo (changes requested):

bundles/org.openhab.binding.openuv/src/main/java/org/openhab/binding/openuv/internal/handler/OpenUVBridgeHandler.java
clinique: Correct.

bundles/org.openhab.binding.openuv/src/main/java/org/openhab/binding/openuv/internal/handler/OpenUVBridgeHandler.java
clinique: Because this can cause the same InterruptedException we had on Netatmo binding.

bundles/org.openhab.binding.openuv/src/main/java/org/openhab/binding/openuv/internal/handler/OpenUVBridgeHandler.java
lolodomo: Unchanged

clinique: It is removed no ?

lolodomo: Yes, I meant "Change undone" rather than "unchanged".

clinique: Yes, I kept it as is at the end.

bundles/org.openhab.binding.openuv/src/main/java/org/openhab/binding/openuv/internal/handler/OpenUVBridgeHandler.java
lolodomo: As cancel(true) is used almost everywhere, I need to think why false us required here. 
And I did not take about this change in the netatmo binding. I should have. 
I do not say this change is wrong, I just want to be sure and understand.

clinique: In Netatmo binding, true led to InterruptedException in the http request procedure, leading to unpredictable situation.
It is not a common issue in openuv binding because it is far more simple, but while debugging in searching the origin of this bug, I saw it. But I can revert if you prefer.

lolodomo (approved):
LGTM

bundles/org.openhab.binding.openuv/src/main/java/org/openhab/binding/openuv/internal/handler/OpenUVBridgeHandler.java
lolodomo: I see that you revert it to cancel(true). So I merge without delay the PR but I think we should clearly have a conviction if true or false should be used.

nesys:
Hi @clinique
it works at the beginning (cleaned cache, restarted the system), but after a while the values in SafeExposure items disappear, and the value is back to 'NULL'. Tested on 3.3.0 M5

kaikreuzer:
@clinique & @lolodomo: Please always make sure to have a speaking PR title, so that things turn up correctly in the release notes. Thanks!

PR #12701, opened 2022-05-08
[netatmo] Do not ignore updated events (same timestamp)

### LINKED ISSUE #12699: [netatmo] Camera event not systematically updated
While testing #12698 I encountered the case where the last event was not updated.
First the channels were updates while the video was recording. At next poll, the event was unchanged in UI (video recording) while the API returned the event with video status being now available. Even after another poll, the channels were not updated.
I was even able to reproduce it.
@clinique: any idea ?

lolodomo:
Fix #12699


clinique (approved):

jlaur (approved):
LGTM, thanks!

PR #12708, opened 2022-05-09
[elroconnects] Correct error for long thing names

mherwege:
Thing names longer than 15 characters caused a index out of bound exception when syncing the name with the name in the elro system.


jlaur (approved):
Thanks!

PR #12710, opened 2022-05-10
[knx] Improve identification of *-control items in UI

### LINKED ISSUE #9462: [knx] Thing channelTypeUID isn't displayed correctly when using *-control channel
## Expected Behavior
If using a *-control channel Type, "Control" should be displayed in addition to the general Type (Switch, Dimmer...)

Create a *-control Channel in a knx thing and save. Look at the configuration...

This issue is available in openHAB2.x as well as in openHAB3 :)

holgerfriedrich:
Control items are correctly added to json config and properly displayed.
Fixes #9462.


lolodomo:
You try to define invalid item types?! 
You must keep valid item type.

holgerfriedrich:
@lolodomo actually, I replicated what was already there for the DateTime control because it showed correcly in the UI.
I just checked that I can write anything into item-type and it will be rendered in the UI.

I will have a deeper look at this later. Converting to draft.

lolodomo:
Look at https://www.openhab.org/docs/developer/bindings/thing-xml.html#bridges-and-thing-descriptions and search item-type for the description, 
If unexpected item-type was used in the past, it has to be fixed. Will do a search on all code base when I have time.

holgerfriedrich:
@lolodomo Thanks for pointing me in the right direction.
As the UI does not show the channel type label (e.g. "Number", "Number Control") in the list of channels, the my first approach was to change the item-type tag. The xsd allows a string here and the result is shown in the UI on the first glance. But as I learned from #12712, I must stick to the defined types. So the item type does not help here.
As KNX has two possible channel types for each item type) which sould be clearly distinguishable for the user as they have different semantics on the bus side, I changed the approach and improved the channel description. -control channels are prefixed with "...-control:". When adding channels this seems a bit redundant, as the label and description are shown, but it seems the best we can do without adding new features to the UI.
Pls also merge #12717 to correct the item tpye for datatime control.

lolodomo:
The new descriptions are a little strange IMHO with these prefix which does really help. Why not clearly explain the difference in the description? 

By the way, I will let @kaikreuzer review and merge this PR.

holgerfriedrich:
@lolodomo @kaikreuzer Do you think it is better to rewrite all channel descriptions, e.g.
"switch: Control a KNX device using a Group Address with a DPT compatible to Switch items"
and "switch-control: Control an Switch device owned by openHAB from the KNX installation (i.e. the switch is native KNX, the device is a virtual KNX device in openHAB)"

kaikreuzer:
I'd agree with @lolodomo that those prefixes look strange and do not really belong there.
In general, I think the current descriptions of the control channels are pretty much correct and I don't really see a need to change them.
Sure, those channel types are at first difficult to understand, but that's why I anyhow expect everyone to consult [the documentation about it](https://www.openhab.org/addons/bindings/knx/#control-channel-types), before diving into this.

holgerfriedrich:
@kaikreuzer I don't mind to close this PR, but it would leave #9462 open.

Creating channels in the UI is fine as it is, but when you look at the tab channels of an item, it gets confusing.
The pattern for listing the channels is as follows: label (as given by user during creation), identifier (as given by user during creation), ```item-type``` (this is the associated OH type, without "-control"), and finally the description from ```device.xml```.
*You cannot distinguish between normal and -control items.*
(just via the description we could give a hint)

I do not see another way wi8thout introducing extensions to the UI which are not needed by other addons. Maybe there is an option I do not know of.

So shall I close?

kaikreuzer:
> but when you look at the tab channels of an item, it gets confusing.

Could you maybe post a screenshot so that it is easier to understand what you refer to? Thanks!

holgerfriedrich:
grafik
@kaikreuzer sure. In the list of channels, the switch (1) and switch-control (2) look nearly similar. This is why I proposed the change marked red.

For adding channels, this is not necessary, here we have a clear difference:
grafik

kaikreuzer:
Ok, thanks @holgerfriedrich!
I would argue that if all is clear enough when creating channels, the user should set the channel id and channel label to values that make it clear to him, what the channel is meant for. Even if he does not, he still sees the existing descriptions, which for a control channel clearly mention that the status is not owned by KNX.

The situation is imho not different to any other such situations in other bindings: An extensible Thing can have many channels of type "Switch" and even have the same channel type. Good naming of channels by the user is therefore important. Tbh, I would therefore also tend to reject #9462, but I can be convinced otherwise by good arguments. 😉

holgerfriedrich:
ping @udo1toni, anything to add?

PR #12711, opened 2022-05-10
[nest] Fix for missing refresh token after reauthorization

mhilbush:
By default, the Google OAuth API returns the refresh token only on the first authorization. On subsequent authorizations using the same Client ID and Secret, Google does not return the refresh token in the access token response. This causes the binding to fail on the first refresh of the access token after a reauthorization.

This affects the SDM and Pub/Sub APIs. However, the failure is more spectacular with the Pub/Sub API as it causes the binding to repeatedly do a pub/sub request against the Google API every 20-40 msec.

This PR adds a new query parameter (prompt=consent) to the URL used to get the authorization code. This parameter causes Google to always return the refresh token in the authorization response (via `oAuthService.getAccessTokenResponseByAuthorizationCode`).

In addition, this PR adds checks to make sure the refresh token in present in the access token response in order to fail more quickly and gracefully if the refresh token is missing.

Further details of my analysis are here.
https://community.openhab.org/t/nest-binding-throwing-pub-sub-errors-repeatedly/135721

I know @wborn is pretty time-constrained right now, but I'm hopeful he can find a few minutes to review. 😉 

### DIFF
--- a/bundles/org.openhab.binding.nest/README.md
+++ b/bundles/org.openhab.binding.nest/README.md
@@ -89,11 +89,11 @@ Finally, an SDM Account Thing can be created to access the SDM project using the
 1. Create an authorization code for the binding:
     1. Replace the **Project ID** and **Client ID** in the URL below with your SDM Project ID and SDM OAuth 2.0 Client ID and open the URL in a new browser tab:
        
-       `https://nestservices.google.com/partnerconnections/<ProjectID>/auth?scope=https://www.googleapis.com/auth/sdm.service&access_type=offline&include_granted_scopes=true&response_type=code&redirect_uri=https://www.google.com&client_id=<ClientID>`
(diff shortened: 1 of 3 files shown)


lolodomo:
I do not see in the code the addition of the prompt query parameter.

mhilbush:
That's correct. It's not in the code, but only in the URL in the documentation. You need to copy/paste the URL from the documentation into your browser's address bar, then replace the `<ProjectID>` and `<ClientID>` in the URL with your specific values. It's a completely manual process.

lolodomo (changes requested):
LGTM

lolodomo (approved):

lolodomo:
I let a chance to @wborn to review.

wborn (approved):
Thanks for the fix!

Issue #12712, opened 2022-05-10
Bad item-type value for channel


lolodomo:
After looking at code change proposed in the PR #12710 , I had the idea to search in our code database if we have unexpected values in item-type tag for channel definition:
Here is the result:
```
$ find */src -name "*.xml" -exec grep -H "<item-type>" {} \; | grep -v ">Switch<" | grep -v ">Number<" | grep -v ">Number:" | grep -v ">Dimmer<" | grep -v ">String<" | grep -v ">Image<" | grep -v ">Color<" | grep -v ">Contact<" | grep -v ">DateTime<" | grep -v ">Rollershutter<" | grep -v ">Call<" | grep -v ">Player<" | grep -v ">Location<"
org.openhab.binding.automower/src/main/resources/OH-INF/thing/thing-types.xml:          <item-type>Point</item-type>
org.openhab.binding.haywardomnilogic/src/main/resources/OH-INF/thing/bow.xml:           <item-type>system.power</item-type>
org.openhab.binding.haywardomnilogic/src/main/resources/OH-INF/thing/heater.xml:                <item-type>system.power</item-type>
org.openhab.binding.ihc/src/main/resources/OH-INF/thing/channels.xml:           <item-type>RollerShutter</item-type>
org.openhab.binding.ism8/src/main/resources/OH-INF/thing/thing-types.xml:               <item-type>number</item-type>
org.openhab.binding.ism8/src/main/resources/OH-INF/thing/thing-types.xml:               <item-type>number</item-type>
org.openhab.binding.knx/src/main/resources/OH-INF/thing/device.xml:             <item-type>DateTime Control</item-type>
org.openhab.binding.lutron/src/main/resources/OH-INF/thing/thing-types.xml:             <item-type>Text</item-type>
org.openhab.binding.oceanic/src/main/resources/OH-INF/thing/channel-types.xml:          <item-type>Decimal</item-type>
org.openhab.binding.oceanic/src/main/resources/OH-INF/thing/channel-types.xml:          <item-type>Decimal</item-type>
org.openhab.binding.paradoxalarm/src/main/resources/OH-INF/thing/partition.xml:         <item-type>switch</item-type>
org.openhab.binding.paradoxalarm/src/main/resources/OH-INF/thing/partition.xml:         <item-type>switch</item-type>
org.openhab.binding.paradoxalarm/src/main/resources/OH-INF/thing/partition.xml:         <item-type>switch</item-type>
[...]

PR #12714, opened 2022-05-10
[automower] Fix item-type value for channel type positionType

lolodomo:
Related to #12712 


jlaur (approved):
LGTM

PR #12716, opened 2022-05-10
[oceanic] Fix item-type value for channel types reserve and flow

lolodomo:
Related to #12712


jlaur (approved):
LGTM

PR #12717, opened 2022-05-10
[knx] Fix item-type value for channel type datetime-control

lolodomo:
Related to #12712


holgerfriedrich:
LGTM

PR #12721, opened 2022-05-11
[Helios] Remove support for obsolete firmware 2.7 including SOAP base…

kgoderis:
…d communication

Signed-off-by: Karel Goderis <karel.goderis@me.com>


kgoderis:
@kaikreuzer In case you review this, know that the Helios binding has not been functional over the last few years. Nobody noticed, but there were issues with resolution of of the Apache CXF javax.ws ClientBuilder. Reason unknown, but most likely because for the SOAP stuff it was making use of very old /other Apache packages. Anyways, Helios does not support anymore the very old firmwares, so my proposal it to simply remove support for it. AFAIK little, if not no, people besides myself do use the Helios IP Vario equipment. I prefer to update the binding in the main repo instead of having to run a custom compiled jar

lolodomo:
I remove the tag "awaiting feedback" which is more when a reviewer is awaiting a feedbacj from the contributor. With this tag, the risk is that no reviewer starts a review, thinking something is awaited from you.

bundles/org.openhab.binding.helios/src/main/java/org/openhab/binding/helios/internal/HeliosHandlerFactory.java
```diff
@@ -37,8 +36,8 @@
 @Component(service = ThingHandlerFactory.class, configurationPid = "binding.helios")
 public class HeliosHandlerFactory extends BaseThingHandlerFactory {
 
-    private static final Set<ThingTypeUID> SUPPORTED_THING_TYPES_UIDS = Collections.unmodifiableSet(
-            Stream.of(HELIOS_VARIO_IP_2_7_TYPE, HELIOS_VARIO_IP_2_21_TYPE).collect(Collectors.toSet()));
+    private static final Set<ThingTypeUID> SUPPORTED_THING_TYPES_UIDS = Collections
+            .unmodifiableSet(Stream.of(HELIOS_VARIO_IP_2_21_TYPE).collect(Collectors.toSet()));
```
jlaur, line 40:
```suggestion
    private static final Set<ThingTypeUID> SUPPORTED_THING_TYPES_UIDS = Set.of(HELIOS_VARIO_IP_2_21_TYPE);
```

jlaur (changes requested):
Thanks! For me it's fine to drop support for that old firmware as long as it's possible for all users to upgrade to 2.21. @kaikreuzer - do you have any feedback? @kgoderis - one minor optional simplification suggested.

kaikreuzer (approved):
lgtm, I'm fine with dropping support for the older firmware.

kaikreuzer:
@kgoderis Could you please add a PR for the [openhab-distro update list](https://github.com/openhab/openhab-distro/blob/main/distributions/openhab/src/main/resources/bin/update.lst#L63-L75) to mention this breaking change?

PR #12725, opened 2022-05-12
[jsscripting] Removed ClassExtender as it's no longer used and may be leaking memory

jpg0:
Looking at a heap dump this is a candidate for a memory leak as it tries to hang on to it's classloader. It's not used anyway (checked in openhab-js repo) so I have removed it.


J-N-K:
I can‘t say much here, but LGTM. I can test if it solves the problem when merged.

jpg0:
Yeah unfortunately I struggled to get a consistent test. Possibly it was because I am using Docker and it seems to take longer to notice file changes to reload.

Either way the change is a positive.

jlaur (approved):
Thanks for the fix!

PR #12728, opened 2022-05-12
[miele] Fix logging for appliances where consumption data is not supported

jlaur:
This is another iteration #11417 regarding extended device state.

A wrong assumption about the minimum amount of data being available in **ExtendedDeviceState** has been revealed after receiving data from an older appliance model.

Accordingly, logging has been fixed to more accurately describe this issue and log level has been reduced to **DEBUG**.


lolodomo (approved):
LGTM

PR #12734, opened 2022-05-14
[insteon] remove instructions about redirecting output from console commands

robnielsen:


robnielsen:
This is dependent on  https://github.com/openhab/openhab-core/pull/2961

jlaur (approved):
Thank you, and also for the core fix rendering these comments obsolete.

PR #12749, opened 2022-05-16
[voskstt] Update the description of the preloadModel parameter

lolodomo:

### DIFF
--- a/bundles/org.openhab.voice.voskstt/src/main/resources/OH-INF/config/config.xml
+++ b/bundles/org.openhab.voice.voskstt/src/main/resources/OH-INF/config/config.xml
@@ -32,8 +32,8 @@
 		<parameter name="preloadModel" type="boolean" groupName="stt">
 			<label>Preload Model</label>
 			<description>Keep the language model loaded. If the parameter is set to true, the model will be reloaded only on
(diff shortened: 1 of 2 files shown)


bundles/org.openhab.voice.voskstt/src/main/resources/OH-INF/config/config.xml
```diff
@@ -31,9 +31,10 @@
 		</parameter>
 		<parameter name="preloadModel" type="boolean" groupName="stt">
 			<label>Preload Model</label>
-			<description>Keep language model loaded. If true model is just reload the model on config updates, if not model will
-				be loaded and offloaded on each execution. It will fallback to try to load the model when executed if it was not
-				able to load it before.</description>
+			<description>Keep the language model loaded. If the parameter is set to true, the model will be reloaded only on
+				configuration updates, but If the model is not loaded when needed, the service will try to load it. If the parameter
```
jlaur, line 35:
```suggestion
				configuration updates. If the model is not loaded when needed, the service will try to load it. If the parameter
```

or alternatively correct "If" to "if", but sentence is quite long, so probably better to end it and start new one.

jlaur (changes requested):
Thanks, just a minor glitch detected.

jlaur (approved):
Thanks.

PR #12756, opened 2022-05-16
DSMR identifier cosem objects should be parsed as hex values

### LINKED ISSUE #12724: [dsmr] Convert logical device name from HEX to ASCII for Smarty meter
## Your Environment

* Version used: openHAB 3.2
* Operating System and version: openHABian 1.7.3

Currently, the `emeter_equipment_identifier_v2_x` channel type id shows the logical device name of the smarty e-meter as a hexadecimal value. It would be nice if this value would internally be converted to ASCII, so that it matches the e-meter serial number that is printed on the physical device (e.g. SAG1030.....).

Example HEX value: `53414731303330363930303032363136`
Resulting ASCII value: `SAG1030690002616`

Hilbrand:
Closes #12724


lolodomo (approved):
LGTM, thank you

PR #12758, opened 2022-05-16
[boschshc] Support for Bosch Intrusion Detection System (#12700)

### LINKED ISSUE #12700: [boschshc] Support for Bosch Intrusion Detection System
As requested in [this community thread](https://community.openhab.org/t/will-there-be-a-bosch-smart-home-binding/8910/128), it would be nice to support obtaining (and maybe even controlling) the intrusion detection alarm system.

According to the [API Documentation](https://apidocs.bosch-smarthome.com/local/) it should be possible to get and set some states from openHAB.

david-pace:
* Add thing and channel definitions for intrusion detection system
* Extract handler abstraction for devices with non-configurable device
IDs
* Extract service abstractions for write-only services (i.e. services
that can not receive states from the bridge)
* Add handler, services and DTO implementations for the intrusion
detection system
* Add detailed Javadocs
* Generalize mechanism to actively fetch initial states for certain
services
* Add unit tests
* Add documentation

closes #12700

### DIFF
--- a/bundles/org.openhab.binding.boschshc/src/main/resources/OH-INF/thing/thing-types.xml
+++ b/bundles/org.openhab.binding.boschshc/src/main/resources/OH-INF/thing/thing-types.xml
@@ -189,6 +189,89 @@
 
 	</thing-type>
 
+	<thing-type id="intrusion-detection-system">
+		<supported-bridge-type-refs>
+			<bridge-type-ref id="shc"/>
+		</supported-bridge-type-refs>
+
+		<label>Intrusion Detection System</label>
+		<description>Allows to retrieve and control the state of the intrusion detection alarm system.</description>
+
+		<channels>
+			<channel id="system-availability" typeId="system-availability"/>
+			<channel id="arming-state" typeId="arming-state"/>
+			<channel id="alarm-state" typeId="alarm-state"/>
+			<channel id="active-configuration-profile" typeId="active-configuration-profile"/>
+			<channel id="arm-action" typeId="arm-action"/>
+			<channel id="disarm-action" typeId="disarm-action"/>
+			<channel id="mute-action" typeId="mute-action"/>
+		</channels>
+
+	</thing-type>
+
+	<channel-type id="system-availability">
+		<item-type>Switch</item-type>
+		<label>System Availability</label>
+		<description>Indicates whether the intrusion detection system is available.</description>
+		<state readOnly="true"/>
+	</channel-type>
+
+	<channel-type id="arming-state">
+		<item-type>String</item-type>
+		<label>Arming State</label>
+		<description>The arming state of the intrusion detection system. Possible values are SYSTEM_ARMING, SYSTEM_ARMED and
+			SYSTEM_DISARMED. This channel is read-only. Use the arm-action and disarm-action channels to arm and disarm the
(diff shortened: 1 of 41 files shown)


bundles/org.openhab.binding.boschshc/src/main/resources/OH-INF/thing/thing-types.xml
```diff
@@ -189,6 +189,89 @@
...
+		<description>The name of the active configuration profile used for the intrusion detection system.</description>
+		<state readOnly="true"/>
+	</channel-type>
+
+	<channel-type id="arm-action">
+		<item-type>String</item-type>
+		<label>Arm Action</label>
+		<description>Arms the intrusion detection system using the given profile ID.</description>
+	</channel-type>
+
```
david-pace, line 262: I was wondering if some channels could be combined to one, for example the arm action and the disarm action. However, arming requires a string command (the profile to use) and disarming would then require handling for a special string like "OFF". Do you agree that using two separate channels is better in this case?

coeing:
@david-pace Just had a really quick glance at your changes. It is really great that you were able to adjust the existing architecture to support the intrusion detection system which works different to the previous devices 👍

jlaur (changes requested):
Thanks for the contribution. It was a joy reviewing a PR of this quality, and I have added only a few minor comments. @coeing, @GerdZanker - awaiting your review also before merging.

david-pace:
> Thanks for the contribution. It was a joy reviewing a PR of this quality, and I have added only a few minor comments. @coeing, @GerdZanker - awaiting your review also before merging.

Thank you very much @jlaur 😃 I have applied all your suggestions in a single commit 👍

bundles/org.openhab.binding.boschshc/src/main/resources/OH-INF/thing/thing-types.xml
jlaur: IMHO it's fine the way it is.

jlaur (approved):
Many thanks, LGTM.

bundles/org.openhab.binding.boschshc/src/main/resources/OH-INF/thing/thing-types.xml
GerdZanker: I'm fine sticking close to the SHC API, because we avoid own logic in the Bosch SHC binding . Either the API is providing already the right level of abstrasction or the users can use e.g. openhab rules if needed.

GerdZanker:
Thank you @david-pace for this big improvement including also your unit tests. 
I will need some free hours during the next days to test and review all your changes. Sorry for this.

david-pace:
Thanks for your response and please take your time @GerdZanker, it's not like we have a deadline or something like that 😉 And additionally it's not the most trivial change either 😇 To quote Joel Spolsky:

> Good software, like wine, takes time.

GerdZanker (approved):

GerdZanker:
I took the time to understand your code changes and because of your very good code comments it was easy to follow the changes and new class hierarchies.
I really think the binding will benefit from your improvements and it will be easier to support other system services from now on.
Thank you for your contribution!

PR #12766, opened 2022-05-19
[fineoffsetweatherstation] Align XML with default translations

lolodomo:
Makes i18n tool reusable.


jlaur (approved):
I missed that in the review, thanks.

PR #12777, opened 2022-05-21
[netatmo] Avoid requesting favorite stations when not needed

### LINKED ISSUE #12769: [netatmo] Requesting favorite stations only when needed
I have not enabled favorite stations:
image

But the binding requests favorite data anyway:
```
15:04:30.993 [TRACE] [tmo.internal.handler.ApiBridgeHandler] - executeUri GET  https://api.netatmo.com/api/getstationsdata?device_id=xxx&get_favorites=true
```

This is probably triggered at this line with the second parameter set to true:
https://github.com/openhab/openhab-addons/blob/main/bundles/org.openhab.binding.netatmo/src/main/java/org/openhab/binding/netatmo/internal/api/WeatherApi.java#L66

It could (should) be optimized to use false when getStationData is called for your local weather station.

@clinique  : WDYT ?

lolodomo:
Fix #12769

### DIFF
--- a/bundles/org.openhab.binding.netatmo/src/main/java/org/openhab/binding/netatmo/internal/api/WeatherApi.java
+++ b/bundles/org.openhab.binding.netatmo/src/main/java/org/openhab/binding/netatmo/internal/api/WeatherApi.java
@@ -45,12 +45,11 @@ public class WeatherApi extends RestManager {
     }
 
     /**
-     *
      * Returns data from a user's Weather Stations (measures and device specific data);
      *
      * @param deviceId Id of the device you want to retrieve information of (optional)
      * @param getFavorites Whether to include the user's favorite Weather Stations in addition to the user's
-     *            own Weather Stations (optional, default to false)
+     *            own Weather Stations
      * @return StationDataResponse
      * @throws NetatmoException If fail to call the API, e.g. server error or deserializing
      */
@@ -62,8 +61,34 @@ public class WeatherApi extends RestManager {
         return response;
     }
 
-    public NAMain getStationData(String deviceId, boolean getFavorites) throws NetatmoException {
-        ListBodyResponse<NAMain> answer = getStationsData(deviceId, getFavorites).getBody();
+    /**
+     * Returns data from a user's Weather Station, this stations can be a station owned by the user or a favorite
+     * station or a guest station.
+     *
+     * @param deviceId Id of the device you want to retrieve information
+     * @return NAMain
+     * @throws NetatmoException If fail to call the API, e.g. server error or deserializing
+     */
+    public NAMain getStationData(String deviceId) throws NetatmoException {
+        ListBodyResponse<NAMain> answer = getStationsData(deviceId, true).getBody();
+        if (answer != null) {
+            NAMain station = answer.getElement(deviceId);
+            if (station != null) {
+                return station;
+            }
+        }
+        throw new NetatmoException("Unexpected answer searching device '%s' : not found.", deviceId);
+    }
+
+    /**
+     * Returns data from a Weather Station owned by the user
+     *
+     * @param deviceId Id of the device you want to retrieve information
+     * @return NAMain
+     * @throws NetatmoException If fail to call the API, e.g. server error or deserializing
+     */
+    public NAMain getOwnedStationData(String deviceId) throws NetatmoException {
+        ListBodyResponse<NAMain> answer = getStationsData(deviceId, false).getBody();
         if (answer != null) {
             NAMain station = answer.getElement(deviceId);
             if (station != null) {
--- a/bundles/org.openhab.binding.netatmo/src/main/java/org/openhab/binding/netatmo/internal/api/WeatherApi.java
+++ b/bundles/org.openhab.binding.netatmo/src/main/java/org/openhab/binding/netatmo/internal/api/WeatherApi.java
@@ -53,7 +53,7 @@ public class WeatherApi extends RestManager {
      * @return StationDataResponse
      * @throws NetatmoException If fail to call the API, e.g. server error or deserializing
      */
-    public StationDataResponse getStationsData(@Nullable String deviceId, boolean getFavorites)
+    private StationDataResponse getStationsData(@Nullable String deviceId, boolean getFavorites)
             throws NetatmoException {
         UriBuilder uriBuilder = getApiUriBuilder(SUB_PATH_GETSTATION, PARAM_DEVICEID, deviceId, //
                 PARAM_FAVORITES, getFavorites);
@@ -96,6 +96,10 @@ public class WeatherApi extends RestManager {
             }
         }
         throw new NetatmoException("Unexpected answer searching device '%s' : not found.", deviceId);
+    }
+
+    public @Nullable ListBodyResponse<NAMain> getFavoriteAndGuestStationsData() throws NetatmoException {
+        return getStationsData(null, true).getBody();
     }
 
     public @Nullable Object getMeasures(String deviceId, @Nullable String moduleId, @Nullable String scale,
--- a/bundles/org.openhab.binding.netatmo/src/main/java/org/openhab/binding/netatmo/internal/api/WeatherApi.java
+++ b/bundles/org.openhab.binding.netatmo/src/main/java/org/openhab/binding/netatmo/internal/api/WeatherApi.java
@@ -83,6 +83,9 @@ public class WeatherApi extends RestManager {
     /**
      * Returns data from a Weather Station owned by the user
      *
+     * This method must be preferred to getStationData when you know that the device is a station owned by the user
+     * (because it avoids requesting additional data for favorite/guest stations).
+     *
      * @param deviceId Id of the device you want to retrieve information
      * @return NAMain
      * @throws NetatmoException If fail to call the API, e.g. server error or deserializing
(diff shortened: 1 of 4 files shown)


bundles/org.openhab.binding.netatmo/src/main/java/org/openhab/binding/netatmo/internal/api/WeatherApi.java
```diff
@@ -62,8 +62,8 @@ public StationDataResponse getStationsData(@Nullable String deviceId, boolean ge
         return response;
     }
 
-    public NAMain getStationData(String deviceId) throws NetatmoException {
-        ListBodyResponse<NAMain> answer = getStationsData(deviceId, true).getBody();
+    public NAMain getStationData(String deviceId, boolean getFavorites) throws NetatmoException {
```
jlaur, line 65: In line 52:
`@param getFavorites Whether to include the user's favorite Weather Stations in addition to the user's own Weather Stations (optional, default to false)`

Does it make sense?

First, there is no "default" for this parameter as it's a required parameter. Perhaps the intention was the mention that if parameter is left out in API call, it will default to false. But as written here, it's not the API which is described, but the actual parameter.

Second, perhaps we should introduce a method overload which could actually default to false. Or even a different method, `getStationDataWithFavorites` for clarity.

bundles/org.openhab.binding.netatmo/src/main/java/org/openhab/binding/netatmo/internal/handler/capability/WeatherCapability.java
```diff
@@ -39,7 +39,7 @@ public WeatherCapability(CommonInterface handler) {
     @Override
     protected List<NAObject> updateReadings(WeatherApi api) {
         try {
-            return List.of(api.getStationData(handler.getId()));
+            return List.of(api.getStationData(handler.getId(), readonly));
```
jlaur, line 42: Just for my understanding: If read-only it means that user doesn't have any favorite stations in addition to the user's own stations?

bundles/org.openhab.binding.netatmo/src/main/java/org/openhab/binding/netatmo/internal/handler/capability/DeviceCapability.java
```diff
@@ -28,12 +28,15 @@
 public class DeviceCapability extends Capability {
     private static final int DATA_AGE_LIMIT_S = 3600;
 
+    protected boolean readonly = true;;
```
jlaur, line 31: Is it intentional that default value is the opposite default value of the API?

bundles/org.openhab.binding.netatmo/src/main/java/org/openhab/binding/netatmo/internal/api/WeatherApi.java
lolodomo: You're right, the comment on the parameter is certainly more a comment for the API.
The binding systematically provides a value.
So the parameter description should be fixed.
No need for another method IMHO.
Just to clarify for you, either you want the data from your own station, in this case you have to provide the device_id and get_favorites can be omitted or can be set to false; setting it to true will be ok bu will return also some data you don't need. If you want the data from a favorite station, you have to provide the device_id and get_favorites must be set to true.

bundles/org.openhab.binding.netatmo/src/main/java/org/openhab/binding/netatmo/internal/handler/capability/WeatherCapability.java
lolodomo: readonly will be set to true in the returned data for a favorite or guest station.
Here you have to pass true as second parameter when you want the data associated to a station you do not own, either a favorite oir guest station.

bundles/org.openhab.binding.netatmo/src/main/java/org/openhab/binding/netatmo/internal/handler/capability/DeviceCapability.java
lolodomo: Yes, it is because the first time the API will be call, I want to be sure that the request will not fail even for a favorite/guest station.
If you call the API with the device id of a favorite station and without setting get_favorites to true, the request fails.
Maybe I should rename this class member into `owned`.

bundles/org.openhab.binding.netatmo/src/main/java/org/openhab/binding/netatmo/internal/api/WeatherApi.java
jlaur: So from your explanation, if I understand it correctly, perhaps something like this would be be easier to understand and read:

```
public NAMain getStationData(String deviceId) throws NetatmoException {
    return getStationData(deviceId, false);
}

public NAMain getGuestStationData(String deviceId) throws NetatmoException {
    return getStationData(deviceId, true);
}

private NAMain getStationData(String deviceId, boolean getFavorites) throws NetatmoException {
    ...
}
```

This would hide the complexity and avoid the combination of providing **own deviceId** and **getFavorites = true**, which will fetch unneeded data.

There is also this call from `NetatmoDiscoveryService` to be considered:
`getStationsData(null, true)`

lolodomo: Ok, why not

jlaur: `getStationsData` (mind the plural Stations) is only called externally from `NetatmoDiscoveryService` and always with deviceId null, so it would make sense to expose an explicit method for this case, e.g.:

```
public StationDataResponse getStationsData()
```

jlaur: There might also be a difference between favorite stations and guest stations, which I'm not aware of, so please keep this in mind also with the naming - if refactoring into something which will be more understandable at first sight.

lolodomo: I just pushed changes but without a dedicated method for discovery. I am going to add it.

lolodomo: Done. You can have a final look.

bundles/org.openhab.binding.netatmo/src/main/java/org/openhab/binding/netatmo/internal/api/WeatherApi.java
```diff
@@ -73,6 +80,31 @@ public NAMain getStationData(String deviceId) throws NetatmoException {
...
+            NAMain station = answer.getElement(deviceId);
+            if (station != null) {
+                return station;
+            }
+        }
+        throw new NetatmoException("Unexpected answer searching device '%s' : not found.", deviceId);
+    }
+
+    public @Nullable ListBodyResponse<NAMain> getFavoriteAndGuestStationsData() throws NetatmoException {
+        return getStationsData(null, true).getBody();
```
jlaur, line 105: I would suggest to drop `@Nullable` and instead throw exception if null like the other method variants.

jlaur (changes requested):
LGTM, just one comment about unified error handling for `NetatmoDiscoveryService`.

bundles/org.openhab.binding.netatmo/src/main/java/org/openhab/binding/netatmo/internal/api/WeatherApi.java
lolodomo: I am not sure. If you have no weather station s attached to your account, maybe the result will be null ? I don't know if the API will return null or an empty list. And I can't test it.

lolodomo: But Gael was testing if the body was not null in `startScan`, so I assume it could be null.

lolodomo: This is of course different for getStationData because it is called for a previously identified weather station.

So I would suggest to keep the code as it is now.

jlaur: You are right that `startScan` doesn't currently consider this as an error. However, it needs to handle the case. So I would then suggest to return an empty list from `getFavoriteAndGuestStationsData` when null is returned from `getStationsData`. This way the caller doesn't have to deal with this, and it will be clear that this is not considered an error. Currently caller has three possible outcomes: null, a list or an exception thrown, which can easily be reduced to two.

lolodomo: Done. I hope this is the last change now ;)

jlaur: It sure is. :-) Thanks for applying all the changes for my suggestions. Not only is the problem now fixed, but the confusion that led to the problem is also addressed by the refactoring.

jlaur (approved):
Thanks for the fix!

PR #12782, opened 2022-05-22
[netatmo] Discovery using current value of readFriends parameter

### LINKED ISSUE #12778: [netatmo] readFriends binding parameter update ignored until binding restart
The `readFriends` parameter is used by the discovery to search or not favorites/guests stations.
When you update this binding parameter (in UI) and then start a new manual discovery, the new value of the binding configuration is not considered.
If you restart openHAB, the new binding parameter value is then considered.

I guees this is probably the same for the other binding parameters.

lolodomo:
Fix #12778


jlaur (approved):
Thanks for the fix!

PR #12787, opened 2022-05-22
[insteon] ignore commands if device is not online

robnielsen:
Added as a defensive measure to ensure that commands are ignored if the device is offline.


jlaur:
@robnielsen - since not linked to any issue, can you add a description to this pull request? What is it solving, and what is the impact of the bug from user perspective?

jlaur (approved):
LGTM

PR #12790, opened 2022-05-23
[insteon] set device to offline if network is not online after initialization

robnielsen:
This solves an issue I noticed in my log file there were instances where devices would go back online when the network was offline. This solves that issue.

### DIFF
--- a/bundles/org.openhab.binding.insteon/src/main/java/org/openhab/binding/insteon/internal/handler/InsteonDeviceHandler.java
+++ b/bundles/org.openhab.binding.insteon/src/main/java/org/openhab/binding/insteon/internal/handler/InsteonDeviceHandler.java
@@ -41,6 +41,7 @@ import org.openhab.core.thing.ChannelUID;
 import org.openhab.core.thing.Thing;
 import org.openhab.core.thing.ThingStatus;
 import org.openhab.core.thing.ThingStatusDetail;
+import org.openhab.core.thing.ThingStatusInfo;
 import org.openhab.core.thing.binding.BaseThingHandler;
 import org.openhab.core.thing.binding.ThingHandlerCallback;
 import org.openhab.core.thing.type.ChannelTypeUID;
@@ -413,6 +414,16 @@ public class InsteonDeviceHandler extends BaseThingHandler {
         }
 
         super.dispose();
+    }
+
+    @Override
+    public void bridgeStatusChanged(ThingStatusInfo bridgeStatusInfo) {
+        if (bridgeStatusInfo.getStatus() == ThingStatus.ONLINE
+                && getThing().getStatusInfo().getStatusDetail() == ThingStatusDetail.BRIDGE_OFFLINE) {
+            updateStatus(ThingStatus.ONLINE, ThingStatusDetail.NONE);
+        } else if (bridgeStatusInfo.getStatus() == ThingStatus.OFFLINE) {
+            updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.BRIDGE_OFFLINE);
+        }
     }
 
     @Override
--- a/bundles/org.openhab.binding.insteon/src/main/java/org/openhab/binding/insteon/internal/handler/InsteonDeviceHandler.java
+++ b/bundles/org.openhab.binding.insteon/src/main/java/org/openhab/binding/insteon/internal/handler/InsteonDeviceHandler.java
@@ -372,7 +372,7 @@ public class InsteonDeviceHandler extends BaseThingHandler {
                     }
                 });
 
-                if (ThingStatus.ONLINE.equals(bridge.getStatus())) {
+                if (ThingStatus.ONLINE == bridge.getStatus()) {
                     updateStatus(ThingStatus.ONLINE);
                 } else {
                     updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.BRIDGE_OFFLINE);
--- a/bundles/org.openhab.binding.insteon/src/main/java/org/openhab/binding/insteon/internal/handler/InsteonDeviceHandler.java
+++ b/bundles/org.openhab.binding.insteon/src/main/java/org/openhab/binding/insteon/internal/handler/InsteonDeviceHandler.java
@@ -41,7 +41,6 @@ import org.openhab.core.thing.ChannelUID;
 import org.openhab.core.thing.Thing;
 import org.openhab.core.thing.ThingStatus;
 import org.openhab.core.thing.ThingStatusDetail;
-import org.openhab.core.thing.ThingStatusInfo;
 import org.openhab.core.thing.binding.BaseThingHandler;
 import org.openhab.core.thing.binding.ThingHandlerCallback;
 import org.openhab.core.thing.type.ChannelTypeUID;
@@ -135,8 +134,6 @@ public class InsteonDeviceHandler extends BaseThingHandler {
     @Override
     public void initialize() {
         config = getConfigAs(InsteonDeviceConfiguration.class);
-
-        updateStatus(ThingStatus.UNKNOWN);
 
         scheduler.execute(() -> {
             Bridge bridge = getBridge();
@@ -414,16 +411,6 @@ public class InsteonDeviceHandler extends BaseThingHandler {
         }
 
         super.dispose();
-    }
-
-    @Override
-    public void bridgeStatusChanged(ThingStatusInfo bridgeStatusInfo) {
-        if (bridgeStatusInfo.getStatus() == ThingStatus.ONLINE
-                && getThing().getStatusInfo().getStatusDetail() == ThingStatusDetail.BRIDGE_OFFLINE) {
-            updateStatus(ThingStatus.ONLINE, ThingStatusDetail.NONE);
-        } else if (bridgeStatusInfo.getStatus() == ThingStatus.OFFLINE) {
-            updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.BRIDGE_OFFLINE);
-        }
     }
 
     @Override
--- a/bundles/org.openhab.binding.insteon/src/main/java/org/openhab/binding/insteon/internal/handler/InsteonDeviceHandler.java
+++ b/bundles/org.openhab.binding.insteon/src/main/java/org/openhab/binding/insteon/internal/handler/InsteonDeviceHandler.java
@@ -136,7 +136,7 @@ public class InsteonDeviceHandler extends BaseThingHandler {
         config = getConfigAs(InsteonDeviceConfiguration.class);
 
         scheduler.execute(() -> {
-            Bridge bridge = getBridge();
+            final Bridge bridge = getBridge();
             if (bridge == null) {
                 String msg = "An Insteon network bridge has not been selected for this device.";
                 logger.warn("{} {}", thing.getUID().getAsString(), msg);
--- a/bundles/org.openhab.binding.insteon/src/main/java/org/openhab/binding/insteon/internal/handler/InsteonDeviceHandler.java
+++ b/bundles/org.openhab.binding.insteon/src/main/java/org/openhab/binding/insteon/internal/handler/InsteonDeviceHandler.java
@@ -135,8 +135,11 @@ public class InsteonDeviceHandler extends BaseThingHandler {
     public void initialize() {
         config = getConfigAs(InsteonDeviceConfiguration.class);
 
+        updateStatus(ThingStatus.UNKNOWN);
+
         scheduler.execute(() -> {
-            if (getBridge() == null) {
+            Bridge bridge = getBridge();
+            if (bridge == null) {
                 String msg = "An Insteon network bridge has not been selected for this device.";
                 logger.warn("{} {}", thing.getUID().getAsString(), msg);
 
@@ -371,7 +374,11 @@ public class InsteonDeviceHandler extends BaseThingHandler {
                     }
                 });
 
-                updateStatus(ThingStatus.ONLINE);
+                if (ThingStatus.ONLINE.equals(bridge.getStatus())) {
+                    updateStatus(ThingStatus.ONLINE);
+                } else {
+                    updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.BRIDGE_OFFLINE);
+                }
             } else {
                 String msg = "Product key '" + productKey
                         + "' does not have any features that match existing channels.";


jlaur:
@robnielsen - since not linked to any issue, can you add a description to this pull request? What is it solving, and what is the impact of the bug from user perspective?

jlaur:
@robnielsen - is this PR replacing #12787?

robnielsen:
> since not linked to any issue, can you add a description to this pull request? What is it solving, and what is the impact of the bug from user perspective?

I noticed in my log files that if I modified my .things file, the devices were going online before the network was ready. This solves that problem.

> is this PR replacing https://github.com/openhab/openhab-addons/pull/12787?
No

bundles/org.openhab.binding.insteon/src/main/java/org/openhab/binding/insteon/internal/handler/InsteonDeviceHandler.java
```diff
@@ -136,18 +136,27 @@ public void initialize() {
...
                 updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.CONFIGURATION_ERROR, msg);
                 return;
             }
 
+            if (!ThingStatus.ONLINE.equals(bridge.getStatus())) {
+                String msg = String.format("Insteon network is not online");
+                logger.debug("{} {}", thing.getUID().getAsString(), msg);
+
+                updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.BRIDGE_OFFLINE, msg);
+                return;
```
jlaur, line 153: It seems this will prevent it from ever coming **ONLINE**? Also, most likely the bridge is not **OFFLINE** in this scenario, but actually **UNKNOWN**. Otherwise child handlers' `initialize()` method shouldn't be called. As far as I can see the only place setting status to **ONLINE** is at the end of this method. This seems like a timing issue. I haven't seen your logs, but I can imagine something like this happening (before your changes) in order:
- Bridge status: INITIALIZING
- Bridge status: UNKNOWN
- Thing status: INITIALIZING
- Thing status: UNKNOWN
- Thing status: ONLINE
- Bridge status: ONLINE

Since the thing always communicates through the bridge, you need to synchronize online state closely with the bridge, as far as I understand.

Perhaps you could implement the Observer pattern where the child things would register as listeners with the bridge. Then they could get notified when the bridge would come online. This would be needed in the scenario where they are initialized faster than the bridge itself will transition from **UNKNOWN** to **ONLINE**.

bundles/org.openhab.binding.insteon/src/main/java/org/openhab/binding/insteon/internal/handler/InsteonDeviceHandler.java
```diff
@@ -136,18 +136,27 @@ public void initialize() {
...
 
                 updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.CONFIGURATION_ERROR, msg);
                 return;
             }
 
+            if (!ThingStatus.ONLINE.equals(bridge.getStatus())) {
+                String msg = String.format("Insteon network is not online");
+                logger.debug("{} {}", thing.getUID().getAsString(), msg);
+
+                updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.BRIDGE_OFFLINE, msg);
```
jlaur, line 152: Nice to have: i18n text for description.

jlaur (changes requested):
Thanks. I have put some remarks trying to understand the underlying problem, which I think is related to timing. See also #12330 and #12113.

bundles/org.openhab.binding.insteon/src/main/java/org/openhab/binding/insteon/internal/handler/InsteonDeviceHandler.java
robnielsen: I've tested this and it works.

jlaur: Can you post logs documenting the status transitions (for bridge and thing) with timestamps - if possible, also logs from before the changes?

robnielsen: I don't have logs before the change, but here's a snapshot of the current log:

```
2022-05-23 09:00:43.695 [INFO ] [ab.event.ThingStatusInfoChangedEvent] - Thing 'insteon:network:home' changed from UNINITIALIZED to UNINITIALIZED (HANDLER_MISSING_ERROR)
2022-05-23 09:00:45.648 [INFO ] [ab.event.ThingStatusInfoChangedEvent] - Thing 'insteon:network:home' changed from UNINITIALIZED (HANDLER_MISSING_ERROR) to INITIALIZING
2022-05-23 09:00:46.523 [INFO ] [ab.event.ThingStatusInfoChangedEvent] - Thing 'insteon:network:home' changed from INITIALIZING to UNKNOWN
2022-05-23 09:00:46.654 [INFO ] [ab.event.ThingStatusInfoChangedEvent] - Thing 'insteon:device:home:179387' changed from INITIALIZING to OFFLINE (BRIDGE_OFFLINE): Insteon network is not online
...
2022-05-23 09:00:47.919 [INFO ] [ab.event.ThingStatusInfoChangedEvent] - Thing 'insteon:device:home:5469D5' changed from INITIALIZING to OFFLINE (BRIDGE_OFFLINE): Insteon network is not online
2022-05-23 09:00:51.181 [INFO ] [ab.event.ThingStatusInfoChangedEvent] - Thing 'insteon:network:home' changed from UNKNOWN to ONLINE
2022-05-23 09:00:51.518 [INFO ] [ab.event.ThingStatusInfoChangedEvent] - Thing 'insteon:device:home:179387' changed from OFFLINE (BRIDGE_OFFLINE): Insteon network is not online to ONLINE
...
2022-05-23 09:00:51.909 [INFO ] [ab.event.ThingStatusInfoChangedEvent] - Thing 'insteon:device:home:5469D5' changed from OFFLINE (BRIDGE_OFFLINE): Insteon network is not online to ONLINE
```

What would happen is the devices would go straight from `INITIALIZING` to `ONLINE`

bundles/org.openhab.binding.insteon/src/main/java/org/openhab/binding/insteon/internal/handler/InsteonDeviceHandler.java
jlaur, line 152: Since bridge is not **OFFLINE** (but UNKNOWN):
```suggestion
                updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.NONE, msg);
```

bundles/org.openhab.binding.insteon/src/main/java/org/openhab/binding/insteon/internal/handler/InsteonDeviceHandler.java
jlaur: Thanks for the clarification. So child thing will track bridge transition to **ONLINE**.

bundles/org.openhab.binding.insteon/src/main/java/org/openhab/binding/insteon/internal/handler/InsteonDeviceHandler.java
jlaur: But now not sure if framework is only marking child things with **BRIDGE_OFFLINE** as ONLINE when bridge comes online, or if detail is not considered. Can you test if this change would break the transition to ONLINE?

jlaur: Indeed:
https://github.com/openhab/openhab-core/blob/86ee4dc2b5802cab3fc3d7a575847e7ab40314d4/bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/binding/BaseThingHandler.java#L645-L647

jlaur: Resolving this comment as it would break the current fix.

bundles/org.openhab.binding.insteon/src/main/java/org/openhab/binding/insteon/internal/handler/InsteonDeviceHandler.java
```diff
@@ -136,18 +136,27 @@ public void initialize() {
...
                 String msg = "An Insteon network bridge has not been selected for this device.";
-                logger.warn("{} {}", thing.getUID().getAsString(), msg);
+                logger.debug("{} {}", thing.getUID().getAsString(), msg);
 
                 updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.CONFIGURATION_ERROR, msg);
                 return;
             }
 
+            if (!ThingStatus.ONLINE.equals(bridge.getStatus())) {
+                String msg = String.format("Insteon network is not online");
```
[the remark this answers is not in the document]
robnielsen: fixed

robnielsen:
@jlaur, I guess the binding wasn't working quite right. The device would go to ONLINE but initialize() wasn't getting called.

robnielsen:
@jlaur, the binding is now working correctly. Now the device is properly initialized, then the device is set ONLINE if the bridge is ONLINE, or OFFLINE (BRIDGE_OFFLINE) otherwise.

BTW, `BaseThingHandler.bridgeStatusChanged()` will set the device to ONLINE when the bridge changes to ONLINE and the thing status detail is set to BRIDGE_OFFLINE.

robnielsen:
> If you do not override bridgeStatusChanged, all your code in initialize() will not be run when the bridge will switch to ONLINE. By default, your thing status will switch ONLINE and that's all.

Thats exactly what needs to happen

lolodomo:
> > If you do not override bridgeStatusChanged, all your code in initialize() will not be run when the bridge will switch to ONLINE. By default, your thing status will switch ONLINE and that's all.
> 
> Thats exactly what needs to happen, the device has been initialized and just needs to be switched to ONLINE.

Makes no sense. If your thing was OFFLINE due a configuration error for example, it should not simply switch to ONLINE because the bridge switch to ONLINE. It should remain OFFLINE until the configuration is fixed.

robnielsen:
>>>If you do not override bridgeStatusChanged, all your code in initialize() will not be run when the bridge will switch to ONLINE. By default, your thing status will switch ONLINE and that's all.
>>
>>Thats exactly what needs to happen, the device has been initialized and just needs to be switched to ONLINE.
>
>Makes no sense. If your thing was OFFLINE due a configuration error for example, it should not simply switch to ONLINE because the bridge switch to ONLINE. It should remain OFFLINE until the configuration is fixed

There are no configuration errors, configuration is complete. The device changes from UNKNOWN to ONLINE if the bridge is online. Otherwise it changes to OFFLINE (BRIDGE_OFFLINE).

lolodomo:
> There are no configuration errors

Here is a part of the code of the initialize() method:
```
        scheduler.execute(() -> {
            if (getBridge() == null) {
                String msg = "An Insteon network bridge has not been selected for this device.";
                logger.warn("{} {}", thing.getUID().getAsString(), msg);

                updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.CONFIGURATION_ERROR, msg);
                return;
            }

            InsteonDeviceConfiguration config = this.config;
            if (config == null) {
                String msg = "Insteon device configuration is null.";
                logger.warn("{} {}", thing.getUID().getAsString(), msg);

                updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.CONFIGURATION_ERROR, msg);
                return;
            }
            String address = config.getAddress();
            if (!InsteonAddress.isValid(address)) {
                String msg = "Unable to start Insteon device, the insteon or X10 address '" + address
                        + "' is invalid. It must be in the format 'AB.CD.EF' or 'H.U' (X10).";
                logger.warn("{} {}", thing.getUID().getAsString(), msg);

                updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.CONFIGURATION_ERROR, msg);
                return;
            }
```
There are clearly checks and the thing can be put OFFLINE/CONFIGURATION_ERROR.

robnielsen:
> There are clearly checks and the thing can be put OFFLINE/CONFIGURATION_ERROR.

Yes but in this case the device will stay OFFLINE. Here is the code for BaseThingHandler.bridgeStatusChanged()

```
    @Override
            updateStatus(ThingStatus.ONLINE, ThingStatusDetail.NONE);
        }
    }
```

It ONLY will change the thing to ONLINE if the bridge changed to online and the status detail is equal to BRIDGE_OFFLINE.

lolodomo:
> It ONLY will change the thing to ONLINE if the bridge changed to online and the status detail is equal to BRIDGE_OFFLINE.

OK, I did not remember of the second condition.
Interesting pattern, it is not the one I am using but ok it should work.

jlaur:
> > It ONLY will change the thing to ONLINE if the bridge changed to online and the status detail is equal to BRIDGE_OFFLINE.
> 
> OK, I did not remember of the second condition. Interesting pattern, it is not the one I am using but ok it should work.

I have only two comments to be considered in relation to this:
- Technically thing should not set **BRIDGE_OFFLINE** when actually bridge is not **OFFLINE**, but **UNKNOWN**.
- Dependency towards current `bridgeStatusChanged` implementation in `BaseThingHandler`. I don't fully understand the reasoning for this logic.

IMHO implementing `bridgeStatusChanged` might be the safer approach, but I'll not insist since the current implementation seems to work - although it seems a bit like a hack because it only works by abusing BRIDGE_OFFLINE to get back ONLINE, even though bridge is not really OFFLINE. @lolodomo - I'll await your input whether we should merge as is, and @robnielsen, please also consider my arguments and whether to make implementation a bit more solid.

robnielsen:
@jlaur, @lolodomo this pattern is used elsewhere. Here are a few examples, I'm sure there are others as well: 

https://github.com/openhab/openhab-addons/blob/main/bundles/org.openhab.binding.serial/src/main/java/org/openhab/binding/serial/internal/handler/SerialDeviceHandler.java

https://github.com/openhab/openhab-addons/blob/main/bundles/org.openhab.binding.velux/src/main/java/org/openhab/binding/velux/internal/handler/VeluxHandler.java

https://github.com/openhab/openhab-addons/blob/main/bundles/org.openhab.binding.unifi/src/main/java/org/openhab/binding/unifi/internal/handler/UniFiBaseThingHandler.java

jlaur:
> @jlaur, @lolodomo this pattern is used elsewhere. Here are a few examples, I'm sure there are others as well:
> 
> https://github.com/openhab/openhab-addons/blob/main/bundles/org.openhab.binding.serial/src/main/java/org/openhab/binding/serial/internal/handler/SerialDeviceHandler.java
> 
> https://github.com/openhab/openhab-addons/blob/main/bundles/org.openhab.binding.velux/src/main/java/org/openhab/binding/velux/internal/handler/VeluxHandler.java
> 
> https://github.com/openhab/openhab-addons/blob/main/bundles/org.openhab.binding.unifi/src/main/java/org/openhab/binding/unifi/internal/handler/UniFiBaseThingHandler.java

UniFi is not using the same logic, it explicitly checks for bridge OFFLINE. You are right about the other two examples, though. The fact that they are also doing it, doesn't make the approach correct though.

bundles/org.openhab.binding.insteon/src/main/java/org/openhab/binding/insteon/internal/handler/InsteonDeviceHandler.java
```diff
@@ -135,8 +135,11 @@ public InsteonDeviceHandler(Thing thing) {
     public void initialize() {
         config = getConfigAs(InsteonDeviceConfiguration.class);
 
+        updateStatus(ThingStatus.UNKNOWN);
+
         scheduler.execute(() -> {
-            if (getBridge() == null) {
+            Bridge bridge = getBridge();
+            if (bridge == null) {
```
lolodomo, line 142: I see many first checks done in this thread that could have been done outside (before scheduler.execute) and so before setting the status to UNKNOWN. This would avoid status being set to UNKNOWN and then immediately after to OFFLINE/CONFIGURATION_ERROR.
Everything that is fast to execute should remain in initialize().
I have not studied in details the big content run in the thread but if a thread was created, I assume this is at least required for a part of the code (because it can take several seconds to run).

lolodomo (changes requested):

lolodomo:
> IMHO implementing `bridgeStatusChanged` might be the safer approach, but I'll not insist since the current implementation seems to work - although it seems a bit like a hack because it only works by abusing BRIDGE_OFFLINE to get back ONLINE, even though bridge is not really OFFLINE. @lolodomo - I'll await your input whether we should merge as is, and @robnielsen, please also consider my arguments and whether to make implementation a bit more solid.

I am sure we have many other bindings with a bad management of bridge/childs. At least, this one seems to work (checked by @robnielsen) and thinking about it, I also believe it should work.
Note that it works in this case because OFFLINE/BRIDGE_OFFLINE is set at the very end in initialize() after all checks leading to OFFLINE
This is IMHO not the recommended way to manage bridge/childs but this is OK for me to merge this PR (after considering my other review comment).

bundles/org.openhab.binding.insteon/src/main/java/org/openhab/binding/insteon/internal/handler/InsteonDeviceHandler.java
robnielsen: @lolodomo, I was told by @kaikreuzer that this code needed to be inside of an thread when it was ported from the oh1 binding.  Why does it matter if it's first set to UNKNOWN then a thread is started up, it is in a UNKNOWN state.

robnielsen:
>This is IMHO not the recommended way to manage bridge/childs but this is OK for me to merge this PR (after considering my other review comment)

Can you point me to the documentation with the recommended approach?

bundles/org.openhab.binding.insteon/src/main/java/org/openhab/binding/insteon/internal/handler/InsteonDeviceHandler.java
lolodomo: This is better if initialize() can return with the final status set.
If it can't because there is stuff to run that needs many seconds, it is ok to set it first to UNKNOW and then update it later in the created thread.
Having useless status changes (in your case UNKNOW and then immediately after OFFLINE/CONFIGURATION_ERROR) could be apparently avoided so why not avoid it.
I am just proposing a better implementation but the code as it is in your PR will also work, of course.

lolodomo:
> Can you point me to the documentation with the recommended approach?

There is not a lot of documentation about that I am aware, except this chapter:
https://www.openhab.org/docs/developer/bindings/#bridge-status-changes

bundles/org.openhab.binding.insteon/src/main/java/org/openhab/binding/insteon/internal/handler/InsteonDeviceHandler.java
lolodomo: Maybe I am not clear with what I am suggesting ?
Of course, the thread must be kept if it contains code that needs several seconds to run.
My suggestion is to move outside the thread all the checks that can be run quickly and lead to a status OFFLINE/CONFIGURATION_ERROR.
Let me know if this is not clear.

lolodomo: Just to explain:
```
    public void initialize() {
        config = getConfigAs(InsteonDeviceConfiguration.class);

            Bridge bridge = getBridge();
            if (bridge == null) {
                String msg = "An Insteon network bridge has not been selected for this device.";
                logger.warn("{} {}", thing.getUID().getAsString(), msg);

                updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.CONFIGURATION_ERROR, msg);
                return;
            }

            InsteonDeviceConfiguration config = this.config;
            if (config == null) {
                String msg = "Insteon device configuration is null.";
                logger.warn("{} {}", thing.getUID().getAsString(), msg);

                updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.CONFIGURATION_ERROR, msg);
                return;
            }
            String address = config.getAddress();
            if (!InsteonAddress.isValid(address)) {
                String msg = "Unable to start Insteon device, the insteon or X10 address '" + address
                        + "' is invalid. It must be in the format 'AB.CD.EF' or 'H.U' (X10).";
                logger.warn("{} {}", thing.getUID().getAsString(), msg);

                updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.CONFIGURATION_ERROR, msg);
                return;
            }

        scheduler.execute(() -> {
```

robnielsen: None of it takes a long time to run, there's just a bunch of initialization code. During the initial port from OH1, I was told to have all the initialization run inside of a thread. It's also cleaner this way to have everything encapsulated in the same place.

BTW, Here's the lifecycle for one of my 60+ devices:

```
2022-05-25 11:00:11.066 [INFO ] [ab.event.ThingStatusInfoChangedEvent] - Thing 'insteon:device:home:22F688' changed from UNINITIALIZED to UNINITIALIZED (BRIDGE_UNINITIALIZED)
2022-05-25 11:00:11.574 [INFO ] [ab.event.ThingStatusInfoChangedEvent] - Thing 'insteon:device:home:22F688' changed from UNINITIALIZED (BRIDGE_UNINITIALIZED) to INITIALIZING
2022-05-25 11:00:11.649 [INFO ] [ab.event.ThingStatusInfoChangedEvent] - Thing 'insteon:device:home:22F688' changed from INITIALIZING to UNKNOWN
2022-05-25 11:00:14.847 [INFO ] [ab.event.ThingStatusInfoChangedEvent] - Thing 'insteon:device:home:22F688' changed from UNKNOWN to OFFLINE (BRIDGE_OFFLINE)
2022-05-25 11:00:24.879 [INFO ] [ab.event.ThingStatusInfoChangedEvent] - Thing 'insteon:device:home:22F688' changed from OFFLINE (BRIDGE_OFFLINE) to ONLINE
```


Prior to this change it would have looked like:

```
2022-05-25 11:00:11.066 [INFO ] [ab.event.ThingStatusInfoChangedEvent] - Thing 'insteon:device:home:22F688' changed from UNINITIALIZED to UNINITIALIZED (BRIDGE_UNINITIALIZED)
2022-05-25 11:00:11.574 [INFO ] [ab.event.ThingStatusInfoChangedEvent] - Thing 'insteon:device:home:22F688' changed from UNINITIALIZED (BRIDGE_UNINITIALIZED) to INITIALIZING
2022-05-25 11:00:14.847 [INFO ] [ab.event.ThingStatusInfoChangedEvent] - Thing 'insteon:device:home:22F688' changed from INITIALIZING to ONLINE
```

robnielsen:
>>Can you point me to the documentation with the recommended approach?
>
>There is not a lot of documentation about that I am aware, except this chapter:
https://www.openhab.org/docs/developer/bindings/#bridge-status-changes

Thanks for finding the documention about this. It says _Therefore, the method `ThingHandler.bridgeStatusChanged(ThingStatusInfo)` must be implemented_.  I added it.

bundles/org.openhab.binding.insteon/src/main/java/org/openhab/binding/insteon/internal/handler/InsteonDeviceHandler.java
lolodomo: Thread should be avoided if all the code can be run quickly. 

https://www.openhab.org/docs/developer/bindings/#the-thinghandler

robnielsen: I'd rather leave as is, it keeps the code cleaner since its all together in the same code block.

lolodomo:
> Thanks for finding the documention about this. It says _Therefore, the method `ThingHandler.bridgeStatusChanged(ThingStatusInfo)` must be implemented_. I added it.

What you added is fully useless as the method has exactly the same code as the overrided method.

bundles/org.openhab.binding.insteon/src/main/java/org/openhab/binding/insteon/internal/handler/InsteonDeviceHandler.java
robnielsen: Also, according to the logs I provided, it takes about 3 seconds to initialize. So it makes sense to do in a separate thread, correct? This could also be considerably longer for others based on factors such as the number of Insteon devices to configure, hardware that OH is running on, etc.

robnielsen:
>What you added is fully useless as the method has exactly the same code as the overrided method.

Yes, that's how I'm expecting it to work. That's why I didn't add it before.

bundles/org.openhab.binding.insteon/src/main/java/org/openhab/binding/insteon/internal/handler/InsteonDeviceHandler.java
lolodomo: Thread will be used for example if you need to execute a HTTP request to authenticate or get data.

lolodomo:
> Yes, that's how I'm expecting it to work. That's why I didn't add it before.

This is not the way @jlaur  or myself would recommend but ok I believe this will not cause any issue so ok for me. 

Remains your useless thread to remove.

bundles/org.openhab.binding.insteon/src/main/java/org/openhab/binding/insteon/internal/handler/InsteonDeviceHandler.java
lolodomo: After removing the thread, the UNKNOWN status is no more needed.

robnielsen: What's here is because of review comments in the past from @kaikreuzer, now you want me to remove the thread? It takes at least 2-3 seconds to initialize as you can see from the log I provided.

lolodomo: UNKNOWN state was added because of the thread as initialize must set a status before returning.

lolodomo: > It takes at least 2-3 seconds to initialize

So some code is really consuming a lot of time. That is probably why Kai suggested the thread. 
So comeback to my initial suggestion and just extract what should run very fast like checking if configuration parameyers ate ok in particular.

robnielsen: What is here works, and the configuration errors are very rare, but need to be covered. Since a thread is needed, I going to leave as is.

lolodomo:
I will not merge with the method `bridgeStatusChanged` added with an identical content compared to its overridden version.

robnielsen:
> I will not merge with the method bridgeStatusChanged added with an identical content compared to its overridden version.

I can remove it but then the binding is not following the requirements at https://www.openhab.org/docs/developer/bindings/#bridge-status-changes: 

> A ThingHandler is notified about Bridge status changes to ONLINE and OFFLINE after a BridgeHandler has been initialized. Therefore, the method ThingHandler.bridgeStatusChanged(ThingStatusInfo) must be implemented (this method is not called for a bridge status updated through the bridge initialization itself). If the Thing of this handler does not have a Bridge, this method is never called.
> 
> If the bridge status has changed to OFFLINE, the status of the handled thing must also be updated to OFFLINE with detail BRIDGE_OFFLINE. If the bridge returns to ONLINE, the thing status must be changed at least to OFFLINE with detail NONE or to another thing specific status.

jlaur:
> I can remove it but then the binding is not following the requirements at https://www.openhab.org/docs/developer/bindings/#bridge-status-changes:

Yes it is, it only states that `bridgeStatusChanged` must be implemented - which it is, by `BaseThingHandler`. If you didn't derive from `BaseThingHandler`, then you would have to implement it in your class. Just to be clear, my previous suggestion was not to duplicate this implementation, but to change it to a custom implementation that didn't require BRIDGE_OFFLINE as detail in order to switch back to ONLINE. At this point let's not go further into that, but simply remove the redundant implementation as suggested by @lolodomo.

robnielsen:
>At this point let's not go further into that, but simply remove the redundant implementation as suggested by @lolodomo.

Done

bundles/org.openhab.binding.insteon/src/main/java/org/openhab/binding/insteon/internal/handler/InsteonDeviceHandler.java
```diff
@@ -136,8 +135,6 @@ public InsteonDeviceHandler(Thing thing) {
     public void initialize() {
         config = getConfigAs(InsteonDeviceConfiguration.class);
 
-        updateStatus(ThingStatus.UNKNOWN);
```
lolodomo: Why do you remove that line ?

The initialize method must set a thing status before exiting. I thing this is mentioned in the documentation.
If you remove that line, this will not be the case. For the thing manager, your thing will be in INITIALIZING state ... until the thread you created sets a new state.

lolodomo (changes requested):

bundles/org.openhab.binding.insteon/src/main/java/org/openhab/binding/insteon/internal/handler/InsteonDeviceHandler.java
robnielsen: According to https://www.openhab.org/docs/concepts/things.html#thing-status, INITIALIZING is the correct state:

**INITIALIZING** This status is assigned while the binding initializes the Thing. It depends on the binding how long the initializing process takes. Commands sent to Channels will not be processed.

**UNKNOWN** The handler is fully initialized but due to the nature of the represented device/service it cannot really tell yet whether the Thing is ONLINE or OFFLINE. Therefore the Thing potentially might be working correctly already and may or may not process commands. But the framework is allowed to send commands, because some radio-based devices may go ONLINE if a command is sent to them. The handler should take care to switch the Thing to ONLINE or OFFLINE as soon as possible.

lolodomo: Sorry, I forgave, I am probably not enough patient.
@jlaur : please take the lead and merge if that is fine for you.

bundles/org.openhab.binding.insteon/src/main/java/org/openhab/binding/insteon/internal/handler/InsteonDeviceHandler.java
```diff
@@ -136,7 +136,8 @@ public void initialize() {
         config = getConfigAs(InsteonDeviceConfiguration.class);
 
         scheduler.execute(() -> {
-            if (getBridge() == null) {
+            Bridge bridge = getBridge();
+            if (bridge == null) {
```
jlaur, line 140: Why is this change needed?

bundles/org.openhab.binding.insteon/src/main/java/org/openhab/binding/insteon/internal/handler/InsteonDeviceHandler.java
```diff
@@ -371,7 +372,11 @@ public void initialize() {
                     }
                 });
 
-                updateStatus(ThingStatus.ONLINE);
+                if (ThingStatus.ONLINE.equals(bridge.getStatus())) {
```
jlaur, line 375: Simplification possible:
```suggestion
                if (ThingStatus.ONLINE == bridge.getStatus()) {
```

bundles/org.openhab.binding.insteon/src/main/java/org/openhab/binding/insteon/internal/handler/InsteonDeviceHandler.java
robnielsen: Bridge is used below

bundles/org.openhab.binding.insteon/src/main/java/org/openhab/binding/insteon/internal/handler/InsteonDeviceHandler.java
robnielsen: changed even though it really doesn't matter

jlaur: It was suggested as a possibility, not a requirement. Argument for using `==` over `equals` on enum:
- It's faster.
- It's safer at run-time.
- It's safer at compile-time.

jlaur (approved):
Thanks.

PR #12794, opened 2022-05-23
[jellyfin] fix multiple sessions for device and update deps

GiviMAD:

Fix error when multiple sessions for same device are returned by the server. 
Also update the sdk dependencies, I should have done it when I updated the library before the initial merge.


jlaur (approved):
Thanks!

PR #12802, opened 2022-05-24
[LuxtronikHeatpump] fix menuStateLine1 state options

sgiehl:
The state value meanings for `menuStateLine1` channel were migrated incorrectly.

The state value `3` actually does not exist (or is not used).

This was also the case in the addon for openhab 1. See https://github.com/openhab/openhab1-addons/blob/main/bundles/binding/org.openhab.binding.novelanheatpump/src/main/java/org/openhab/binding/novelanheatpump/internal/HeatPumpBinding.java#L468-L479


lolodomo (approved):
LGTM but it will require a manual action in Crowdin to resynchronize the German translation. It is better to update translations only in Crowdin.

sgiehl:
I'll update the translations there.

lolodomo:
> I'll update the translations there.

No, please let me upload the file.
And first the new reference file must be synchronized with Crowdin.

lolodomo:
It is done, I uploaded the German translations form Github into Crowdin.
I hope we will not have bad surprise when next synchronization will be triggered.

PR #12807, opened 2022-05-25
[Netatmo] Enhancement of documentation

### LINKED ISSUE #12799: [Netatmo] Calculation of ID
At the moment the following information is part of the documentation:

```
You have to calculate the ID for the outside module as follows: (it cannot be read from the app)

- if the first serial character is "h": start with "02"
- if the first serial character is "i": start with "03"

```

Looks like this can be enhanced to

```
You have to calculate the ID for the outside module as follows: (it cannot be read from the app)

- if the first serial character is "h": start with "02"
- if the first serial character is "i": start with "03"
- if the first serial character is "j": start with "04"
- if the first serial character is "k": start with "05"

```

Successfully tested with a device with starting character "k". Can anyone confirm?

AndreasIsenmann:
Resolves #12799


jlaur (approved):
Thanks!

PR #12819, opened 2022-05-26
[nikohomecontrol] Do not set thing status before end of initialization

mherwege:

There is a large [PR #11963](https://github.com/openhab/openhab-addons/pull/11963) with new functionality for the Niko Home Control binding that has been open, waiting for review, for a while.
One of the things that got fixed there is an annoying bug. Things did not get initialized properly as the bridge status was already set before the bridge initialization finished. This small PR isolates this fix. I hope that makes it quicker to at least get this fix merged.

### DIFF
--- a/bundles/org.openhab.binding.nikohomecontrol/src/main/java/org/openhab/binding/nikohomecontrol/internal/handler/NikoHomeControlActionHandler.java
+++ b/bundles/org.openhab.binding.nikohomecontrol/src/main/java/org/openhab/binding/nikohomecontrol/internal/handler/NikoHomeControlActionHandler.java
@@ -66,8 +66,8 @@ public class NikoHomeControlActionHandler extends BaseThingHandler implements Nh
     public void handleCommand(ChannelUID channelUID, Command command) {
         NikoHomeControlCommunication nhcComm = getCommunication();
         if (nhcComm == null) {
-            updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.BRIDGE_UNINITIALIZED,
-                    "@text/offline.bridge-unitialized");
+            updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.CONFIGURATION_ERROR,
+                    "@text/offline.configuration-error.invalid-bridge-handler");
             return;
         }
 
@@ -214,7 +214,8 @@ public class NikoHomeControlActionHandler extends BaseThingHandler implements Nh
 
         NikoHomeControlCommunication nhcComm = getCommunication();
         if (nhcComm == null) {
-            updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.BRIDGE_OFFLINE, "@text/offline.bridge-unitialized");
+            updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.CONFIGURATION_ERROR,
+                    "@text/offline.configuration-error.invalid-bridge-handler");
             return;
         }
 
@@ -338,8 +339,8 @@ public class NikoHomeControlActionHandler extends BaseThingHandler implements Nh
         if (nhcBridgeHandler != null) {
             nhcBridgeHandler.bridgeOnline();
         } else {
-            updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.BRIDGE_UNINITIALIZED,
-                    "@text/offline.bridge-unitialized");
+            updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.CONFIGURATION_ERROR,
+                    "@text/offline.configuration-error.invalid-bridge-handler");
         }
     }
 
--- a/bundles/org.openhab.binding.nikohomecontrol/src/main/java/org/openhab/binding/nikohomecontrol/internal/handler/NikoHomeControlActionHandler.java
+++ b/bundles/org.openhab.binding.nikohomecontrol/src/main/java/org/openhab/binding/nikohomecontrol/internal/handler/NikoHomeControlActionHandler.java
@@ -217,8 +217,6 @@ public class NikoHomeControlActionHandler extends BaseThingHandler implements Nh
             updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.BRIDGE_UNINITIALIZED,
                     "@text/offline.bridge-unitialized");
             return;
-        } else {
-            updateStatus(ThingStatus.UNKNOWN);
         }
 
         // We need to do this in a separate thread because we may have to wait for the communication to become active
--- a/bundles/org.openhab.binding.nikohomecontrol/src/main/java/org/openhab/binding/nikohomecontrol/internal/handler/NikoHomeControlActionHandler.java
+++ b/bundles/org.openhab.binding.nikohomecontrol/src/main/java/org/openhab/binding/nikohomecontrol/internal/handler/NikoHomeControlActionHandler.java
@@ -55,6 +55,8 @@ public class NikoHomeControlActionHandler extends BaseThingHandler implements Nh
 
     private volatile @Nullable NhcAction nhcAction;
 
+    private volatile boolean initialized = false;
+
     private String actionId = "";
     private int stepValue;
     private boolean invert;
@@ -65,10 +67,9 @@ public class NikoHomeControlActionHandler extends BaseThingHandler implements Nh
 
     @Override
     public void handleCommand(ChannelUID channelUID, Command command) {
-        NikoHomeControlCommunication nhcComm = getCommunication();
+        NikoHomeControlCommunication nhcComm = getCommunication(getBridgeHandler());
         if (nhcComm == null) {
-            updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.CONFIGURATION_ERROR,
-                    "@text/offline.configuration-error.invalid-bridge-handler");
+            logger.debug("communication not up yet, cannot handle command {} for {}", command, channelUID);
             return;
         }
 
@@ -201,6 +202,8 @@ public class NikoHomeControlActionHandler extends BaseThingHandler implements Nh
 
     @Override
     public void initialize() {
+        initialized = false;
+
         NikoHomeControlActionConfig config;
         if (thing.getThingTypeUID().equals(THING_TYPE_DIMMABLE_LIGHT)) {
             config = getConfig().as(NikoHomeControlActionDimmerConfig.class);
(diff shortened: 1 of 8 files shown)


lolodomo:
This is perfectly expected to set a thing status in thing handler initialize() before returning. So the code you removed is correct.
The problem is a bad bridge/thing implementation: your thing handler is apparently not checking the bridge status.
I can try to propose an alternative and proper correction of your problem if you are available to test it.

lolodomo:
> There is a large [PR #11963](https://github.com/openhab/openhab-addons/pull/11963) with new functionality for the Niko Home Control binding that has been open, waiting for review, for a while.

Prefer several small PRs and you will get a faster feedback.
This is what you did with this PR and we should conclude soon.

lolodomo:
> > The problem is a bad bridge/thing implementation: your thing handler is apparently not checking the bridge status.
> 
> I was wrong, it is checked ... but very late in the thing initialization. To be investigated.
You are updating very often the thing status in this binding, probably more than expected. Could be another source of problems.

bundles/org.openhab.binding.nikohomecontrol/src/main/java/org/openhab/binding/nikohomecontrol/internal/handler/NikoHomeControlActionHandler.java
```diff
@@ -217,8 +217,6 @@ public void initialize() {
             updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.BRIDGE_UNINITIALIZED,
```
jlaur, line 217: Additional improvement opportunity here: **BRIDGE_UNINITIALIZED** should be set by framework in combination with status **UNINITIALIZED**. When bridge is not initialized, `initialize()` in child things will not be called. Instead you can set **BRIDGE_OFFLINE** (if this is the case). For reference: https://www.openhab.org/docs/concepts/things.html#status-details

mherwege: @jlaur Correct, and this should not happen anymore at all. nhcComm cannot be null if the bridge has been initialized. I modified the ThingStatusDetail anyway.

bundles/org.openhab.binding.nikohomecontrol/src/main/java/org/openhab/binding/nikohomecontrol/internal/handler/NikoHomeControlActionHandler.java
```diff
@@ -214,11 +214,8 @@ public void initialize() {
 
         NikoHomeControlCommunication nhcComm = getCommunication();
         if (nhcComm == null) {
-            updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.BRIDGE_UNINITIALIZED,
-                    "@text/offline.bridge-unitialized");
+            updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.BRIDGE_OFFLINE, "@text/offline.bridge-unitialized");
```
jlaur, line 217: Please update i18n key and text accordingly:
`offline.bridge-unitialized = Bridge not initialized`

bundles/org.openhab.binding.nikohomecontrol/src/main/java/org/openhab/binding/nikohomecontrol/internal/handler/NikoHomeControlActionHandler.java
jlaur: It will also return null in case `getBridgeHandler()` returns null. This can occur in case of misconfiguration where thing is configured without a bridge. Please check if this is now handled correctly - **CONFIGURATION_ERROR** would be more correct in this scenario.

jlaur: See for example:
https://github.com/openhab/openhab-addons/blob/88e7ba3774045325c6e85c613eeefb468db2ecef/bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/handler/HDPowerViewShadeHandler.java#L97-L102

mherwege: @jlaur Indeed, I didn't think that through properly for configuration errors. I changed it.

bundles/org.openhab.binding.nikohomecontrol/src/main/java/org/openhab/binding/nikohomecontrol/internal/handler/NikoHomeControlActionHandler.java
mherwege: @jlaur Changed.

jlaur:
@lolodomo - I added a few comments which are now resolved. Please have a final look as you also had some concerns - and merge if everything is okay with you.

mherwege:
@lolodomo Is there anything left you would want me to look into? I looked again at the sequence of setting the bridge and thing status. There is nothing obvious to simplify it when I look at it. Let me know if you have suggestions.

bundles/org.openhab.binding.nikohomecontrol/src/main/java/org/openhab/binding/nikohomecontrol/internal/handler/NikoHomeControlActionHandler.java
```diff
@@ -214,11 +215,9 @@ public void initialize() {
 
         NikoHomeControlCommunication nhcComm = getCommunication();
         if (nhcComm == null) {
-            updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.BRIDGE_UNINITIALIZED,
-                    "@text/offline.bridge-unitialized");
+            updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.CONFIGURATION_ERROR,
+                    "@text/offline.configuration-error.invalid-bridge-handler");
             return;
-        } else {
-            updateStatus(ThingStatus.UNKNOWN);
```
lolodomo, line 221: The recommended practice is to have initialize() setting a valid status (ONLINE / OFFLINE / UNKNOWN before returning and so before running additional code in a separate thread.
Look at the template: https://github.com/openhab/openhab-core/blob/main/tools/archetype/binding/src/main/resources/archetype-resources/src/main/java/internal/__bindingIdCamelCase__Handler.java#L80

Same remark for everywhere where you removed the set of status to UNKNOWN.

lolodomo (changes requested):

lolodomo:
The PR title is already a derivation to the recommendation ;) Yes, a valid thing status must be set as quickly.

bundles/org.openhab.binding.nikohomecontrol/src/main/java/org/openhab/binding/nikohomecontrol/internal/handler/NikoHomeControlActionHandler.java
```diff
@@ -66,8 +67,8 @@ public NikoHomeControlActionHandler(Thing thing) {
     public void handleCommand(ChannelUID channelUID, Command command) {
         NikoHomeControlCommunication nhcComm = getCommunication();
         if (nhcComm == null) {
-            updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.BRIDGE_UNINITIALIZED,
-                    "@text/offline.bridge-unitialized");
+            updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.CONFIGURATION_ERROR,
+                    "@text/offline.configuration-error.invalid-bridge-handler");
```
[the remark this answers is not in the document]
mherwege: This goes back to a remark from @jlaur. You can configure a thing without a bridge (clearly a communication error), and this would result in this code being executed. I would think (and agree with @jlaur) the right thing to do is to put the thing offline and point out the configuration error.

bundles/org.openhab.binding.nikohomecontrol/src/main/java/org/openhab/binding/nikohomecontrol/internal/handler/NikoHomeControlActionHandler.java
mherwege: Initialize does set the status, and the removed lines make sure it is only set when all of the initialization logic has executed. However, the bridge status can change if communication is lost. I do agree that the UNKNOWN status could not happen, so the last else would not be needed, but I added it for completeness. Should I remove it and replace by some debug log?

mherwege:
The initialize methods do return quickly. And they ultimately do set the thing status, but in a separate thread to allow the initialize method to return quickly. Is there an obligation to already set the thing status in the same thread the initialize method was called from? If that’s the case, I misinterpreted and I will have to do many more changes. I thought returning quickly just had to do with not keeping a thread occupied that would initialize other things in other bindings.

mherwege:
@lolodomo Setting the ting status to unknown was introduced in the binding 7 months ago. It wasn’t there before. The binding pre-dates the creation of the archetype. I read the reasoning for setting thing status unknown (unit tests), but I don’t agree it is right to set it in this case. Introducing it introduced a bug in the binding, which I try to revert here.

lolodomo:
> Is there an obligation to already set the thing status in the same thread the initialize method was called from? If that’s the case, I misinterpreted and I will have to do many more changes. I thought returning quickly just had to do with not keeping a thread occupied that would initialize other things in other bindings.

This is the recommendation, here is the code from the binding template:
```
        // set the thing status to UNKNOWN temporarily and let the background task decide for the real status.
        // the framework is then able to reuse the resources from the thing handler initialization.
        // we set this upfront to reliably check status updates in unit tests.

        // Example for background initialization:
        scheduler.execute(() -> {
            boolean thingReachable = true; // <background task with long running initialization here>
            // when done do:
            if (thingReachable) {
            } else {
            }
        });
```

mherwege:
@lolodomo Then the developer documentation should be changed: https://www.openhab.org/docs/developer/bindings/
Again, please, I am willing to keep status UNKNOWN for the thing handlers, but it leads to wrong behavior for the bridge handler (calling the thing handlers initialize too early). Do you also want me to keep it in the bridge handler? I read the argument in the comment in the archetype implementation again. This was introduced to avoid issues with unit tests, that is the only reason. It cannot be right to break bindings with no unit tests because of this.

lolodomo:
> I read the argument in the comment in the archetype implementation again. This was introduced to avoid issues with unit tests, that is the only reason.

Why do you say it is the only reason ? I don't think so.

As reviewer, I think my main role is to focus on the good practices and guidelines are followed. So I will not approve this PR myself.

Now, this is only my feedback and of course another reviewer can approve and merge this PR. @jlaur was already involved in the review.

lolodomo:
> f this is a hard rule, it means all long running bridge initializations will force depending thing initializations to have extra logic to wait for the bridge to be fully initialized. This essentially voids the advantage of the thing-bridge dependency.

There is no particular difficulty when the bridge/thing initialization is well implemented. This case is properly handled in many bindings. In your thing handler, you just have to first check the status of the bridge. If not yet ONLINE, you set your thing status to OFFLINE/BRIDGE_OFFLINE.

lolodomo:
> I could keep it in the thing handlers, but not int bridge handler. That is where the real problem is, because it caused the things to initialize too early.

Yes, of course, but in this case, you set your thing status to OFFLINE/BRIDGE_OFFLINE.
When your bridge status is changed, this will trigger your thing handler and you can run your thing initialization stuff.

mherwege:
@lolodomo I don't want to argue this any further. I developed this binding something like 5 years ago, based on the documentation at that point in time (which still is the same in the documentation). It had always worked well and reliably. On Oct. 9th 2021, in another PR [#11319](https://github.com/openhab/openhab-addons/pull/11319), you asked me to put the thing status UNKNOWN lines in. From that point onwards, the binding stopped starting reliably. I now for the first time see the archetype, which was only created after my binding got created, reviewed and accepted, and I don't agree with that specific line in it.
I agree things can be made to work, but more of the effort is on the binding than on the framework in that case. To make it work reliably again, I need to lift much of the initialization logic from the thing handler initialize methods and only call it from initialize when the bridge is out of status UNKOWN, or from bridgeStatusChanged. I think it makes things more complicated then it needs to be, but will do so.

lolodomo:
> I agree things can be made to work, but more of the effort is on the binding than on the framework in that case. To make it work reliably again, I need to lift much of the initialization logic from the thing handler initialize methods and only call it from initialize when the bridge is out of status UNKOWN, or from bridgeStatusChanged. I think it makes things more complicated then it needs to be, but will do so.

Wait for @jlaur feedback, he is maybe ok with your changes and will merge.

jlaur:
> Wait for @jlaur feedback, he is maybe ok with your changes and will then merge your PR.

I will be away today, but will be back in the evening. If there will still be any doubts then, I will catch up and try to provide some feedback.

mherwege:
For reference, here is the original discussion on why the requirement was introduced in the archetype: [PR #6160](https://github.com/eclipse-archived/smarthome/pull/6160)

mherwege:
@lolodomo @jlaur I think I have made the necessary modifications to be able to set the thing status early now. As far as I could test, it still behaves as I would want it to behave. Could you check it out?

bundles/org.openhab.binding.nikohomecontrol/src/main/java/org/openhab/binding/nikohomecontrol/internal/handler/NikoHomeControlActionHandler.java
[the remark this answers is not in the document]
jlaur: Correct. See also #12331.

jlaur (approved):
Thanks!

lolodomo:
I did not yet check the final PR but I assume the PR title has to be updated (it will appear in the changelog for the user) ?

PR #12839, opened 2022-05-28
[roku] Add PowerOn option to Remote Button channel

mlobstein:
Add missing Power On option to Remote Button drop down for Roku TV.

Discussed in the community here: https://community.openhab.org/t/roku-binding-power-for-tvs/136223


lolodomo (approved):
LGTM

Issue #12837, opened 2022-05-28
Bindings not yet avaiable for translation in Crowdin


lolodomo:
Here is the list of bindings not available in Crowdin:

- [x] darksky => PR #12853
- [x] herzborg => PR #12838 
- [x] meteoalerte (known and accepted)
- [x] modbus.e3dc => PR #12844
- [x] modbus.studer => PR #12844
- [x] io.imperihome => PR #12845
- [x] persistence/jpa => nothing to translate, no configuration through UI
- [x] persistence/mapdb => nothing to translate, no configuration
- [x] persistence/mongodb => nothing to translate, no configuration through UI
- [x] persistence/rrd4j => nothing to translate, no configuration through UI
- [x] transform/bin2json => nothing to translate, the service has no configuration
- [x] voice/mactts => nothing to translate, the service has no configuration
- [x] voice/marytts => nothing to translate, the service has no configuration
- [x] voice/picotts => nothing to translate, the service has no configuration

PR #12844, opened 2022-05-29
[modbus.*] Updated default translations

lolodomo:
Related to #12837


lolodomo:
For modbus.e3dc, a German translation already exists so we should load it in Crowdin after the merge of this PR.

jlaur (approved):
LGTM

lolodomo:
> For modbus.e3dc, a German translation already exists so we should load it in Crowdin after the merge of this PR.

Done but entries are not yet all translated.

PR #12850, opened 2022-05-30
[androiddebugbridge] Fix GetCurrentPackage for Android 11

Albinoman887:
* In Android 11 google updated the dympsys service and
  changed the naming of some of the command options which
  causes the binding to not parse the current package
  name correctly.

* Added a check for the current android version and if it
  returns "11" use the new command, otherwise use original

Signed-off-by: Matt Filetto <matt.filetto@gmail.com>

### DIFF
--- a/bundles/org.openhab.binding.androiddebugbridge/src/main/java/org/openhab/binding/androiddebugbridge/internal/AndroidDebugBridgeDevice.java
+++ b/bundles/org.openhab.binding.androiddebugbridge/src/main/java/org/openhab/binding/androiddebugbridge/internal/AndroidDebugBridgeDevice.java
@@ -201,7 +201,14 @@ public class AndroidDebugBridgeDevice {
 
     public String getCurrentPackage() throws AndroidDebugBridgeDeviceException, InterruptedException,
             AndroidDebugBridgeDeviceReadException, TimeoutException, ExecutionException {
-        var out = runAdbShell("dumpsys", "window", "windows", "|", "grep", "mFocusedApp");
+        int version = Integer.parseInt(getAndroidVersion());
+        String out;
+        if (version >= 11) {
+            out = runAdbShell("dumpsys", "window", "displays", "|", "grep", "mFocusedApp");
+            logger.debug("detected android version: {} ", getAndroidVersion());
+        } else {
+            out = runAdbShell("dumpsys", "window", "windows", "|", "grep", "mFocusedApp");
+        }
         var targetLine = Arrays.stream(out.split("\n")).findFirst().orElse("");
         var lineParts = targetLine.split(" ");
         if (lineParts.length >= 2) {


jlaur:
@Albinoman887 - any reason for this new PR instead of updating #12065 with the version check? They seem to be fixing the same issue.

bundles/org.openhab.binding.androiddebugbridge/src/main/java/org/openhab/binding/androiddebugbridge/internal/AndroidDebugBridgeDevice.java
```diff
@@ -202,6 +202,10 @@ public void stopPackage(String packageName)
     public String getCurrentPackage() throws AndroidDebugBridgeDeviceException, InterruptedException,
             AndroidDebugBridgeDeviceReadException, TimeoutException, ExecutionException {
         var out = runAdbShell("dumpsys", "window", "windows", "|", "grep", "mFocusedApp");
+        if (getAndroidVersion().contains("11")) {
```
jlaur, line 205: Can you provide an example how this string looks? Is is possible to get exact major version instead of having to rely on string comparison?

jlaur: Also, if this fix only for Android 11, or 11 and higher? The previous PR mentioned "before Android 11".

bundles/org.openhab.binding.androiddebugbridge/src/main/java/org/openhab/binding/androiddebugbridge/internal/AndroidDebugBridgeDevice.java
```diff
@@ -202,6 +202,10 @@ public void stopPackage(String packageName)
...
+            out = runAdbShell("dumpsys", "window", "displays", "|", "grep", "mFocusedApp");
```
jlaur, line 206: Is is on purpose that you run both commands for Android 11, i.e. line 204 **and** 206?

bundles/org.openhab.binding.androiddebugbridge/src/main/java/org/openhab/binding/androiddebugbridge/internal/AndroidDebugBridgeDevice.java
```diff
@@ -202,6 +202,10 @@ public void stopPackage(String packageName)
...
+            logger.debug("detected android version: {} ", getAndroidVersion());
```
jlaur, line 207:
```suggestion
            logger.debug("detected android version 11: {} ", getAndroidVersion());
```

jlaur (changes requested):
Thanks, a few questions that might trigger actual remarks (or not).

bundles/org.openhab.binding.androiddebugbridge/src/main/java/org/openhab/binding/androiddebugbridge/internal/AndroidDebugBridgeDevice.java
GiviMAD, line 205: I think we should not rely on this call, this way we are adding an extra adb command per execution for an information that is already stored in the thing properties. 
We should instead pass the android version stored on the device properties to this class.

GiviMAD (changes requested):
Thank you for your changes!
I have added a comment, we should try to avoid the command for the device version as it's already stored in the thing.

bundles/org.openhab.binding.androiddebugbridge/src/main/java/org/openhab/binding/androiddebugbridge/internal/AndroidDebugBridgeDevice.java
Albinoman887: I believe it is Android 11 or higher....  the string looks like this (this is from my pixel 6 pro hence the 12)
raven:/ $ getprop ro.build.version.release
12

there are other build props we could use.. that was just already there.

bundles/org.openhab.binding.androiddebugbridge/src/main/java/org/openhab/binding/androiddebugbridge/internal/AndroidDebugBridgeDevice.java
Albinoman887: Honestly i'm not formally trained on coding or anything and I was having issues. My original idea was just to make a if else statement that depending on the string for getAndroidVersion would either declare the out variable as either line 204 or 206 depending. but when I made it like this it would fail to compile saying out was not defined

if (getAndroidVersion().contains("11")) {
var out = runAdbShell("dumpsys", "window", "displays", "|", "grep", "mFocusedApp");
} else {
var out = runAdbShell("dumpsys", "window", "windows", "|", "grep", "mFocusedApp");
}

and I don't understand the concept of classes and public vs private etc etc. I get a surprising amount of things done considering but a lot of trial and error.

bundles/org.openhab.binding.androiddebugbridge/src/main/java/org/openhab/binding/androiddebugbridge/internal/AndroidDebugBridgeDevice.java
Albinoman887: what ever you think is best. I thought that the result of "getAndroidVersion()" is populated inside of those two brackets is why I had it that way .

bundles/org.openhab.binding.androiddebugbridge/src/main/java/org/openhab/binding/androiddebugbridge/internal/AndroidDebugBridgeDevice.java
Albinoman887: oh ? where is it stored? I was kind of confused on that cuz I see in the mDNS discovery it seems to define that but I swear it wasn't a prop before

Albinoman887:
> @Albinoman887 - any reason for this new PR instead of updating #12065 with the version check? They seem to be fixing the same issue.

Honestly,,, just laziness. it was late and I was tired :)

bundles/org.openhab.binding.androiddebugbridge/src/main/java/org/openhab/binding/androiddebugbridge/internal/AndroidDebugBridgeDevice.java
jlaur: I'll withdraw this comment. I assumed that `getAndroidVersion()` would return some kind of longer string, and you wanted to log that the "version 11" path was chosen, based on that string.

jlaur:
> > @Albinoman887 - any reason for this new PR instead of updating #12065 with the version check? They seem to be fixing the same issue.
> 
> Honestly,,, just laziness. it was late and I was tired :) should I close this and update the other one? or vise versa?

Your call, both PR's now seem to have an equal amount of comments and additional information. Close the one you'd like. :)

bundles/org.openhab.binding.androiddebugbridge/src/main/java/org/openhab/binding/androiddebugbridge/internal/AndroidDebugBridgeDevice.java
jlaur:
```
String out;
if (getAndroidVersion().contains("11")) {
    out = runAdbShell("dumpsys", "window", "displays", "|", "grep", "mFocusedApp");
} else {
    out = runAdbShell("dumpsys", "window", "windows", "|", "grep", "mFocusedApp");
}
```

or perhaps:

```
String out = runAdbShell("dumpsys", "window", getAndroidVersion().contains("11") ? "displays" : "windows", "|", "grep", "mFocusedApp");
```

bundles/org.openhab.binding.androiddebugbridge/src/main/java/org/openhab/binding/androiddebugbridge/internal/AndroidDebugBridgeDevice.java
Albinoman887: what do you mean? that build prop has been around since android 1

bundles/org.openhab.binding.androiddebugbridge/src/main/java/org/openhab/binding/androiddebugbridge/internal/AndroidDebugBridgeDevice.java
Albinoman887: oh I see I had to declare out before the ifdef... gotcha.. but your suggestion seems better. I"ll update it and change it to >= 11

bundles/org.openhab.binding.androiddebugbridge/src/main/java/org/openhab/binding/androiddebugbridge/internal/AndroidDebugBridgeDevice.java
jlaur: Sorry for being unclear: I meant implementation of getting version number might change because of other comment by @GiviMAD. Maybe it won't be string, in which case you won't have to parse it to get integer value for comparison.

Albinoman887: so how do I access that from the AndroidDebugBridgeDevice class?

GiviMAD: It's described in the previous comment:
| I think you could just add a 'setVersion(String version)' method to the 'AndroidDebugBridgeDevice' class and call it from this one, so the version is propagated to the 'AndroidDebugBridgeDevice' instance each time the handler calls 'refreshProperties' and stored in a private property that we can rely on.

So my proposal is adding a @Nullable private String variable called version in the 'AndroidDebugBridgeDevice', adding there a public void setVersion(String version) to setting it from the 'refreshProperties' method in the handler.

Albinoman887: OK. I don't know how to do that (or I might but not know the terminology). Would you be up for doing that ? or show me an example of how to? Or if you want you can just push the change...

lolodomo:
@GiviMAD : is this PR still relevant now that #13390 is merged?

GiviMAD:
> @GiviMAD : is this PR still relevant now that #13390 is merged?

No, it can be closed. @lolodomo

GiviMAD:
Solved by https://github.com/openhab/openhab-addons/pull/13390

Albinoman887:
Yay !

PR #12854, opened 2022-06-01
[homekit] update configuration revision on start

yfre:
update configuration revision on openHAB restart. this will force all paired ios devices to request a new list of homekit accessories provided via openHAB.

this solves following issues:
- item config was change while openHAB was down.
- the orders and ids of homekit accessories has changed, as described here 
https://community.openhab.org/t/how-to-configure-homekit-light-with-colour/136137


jlaur (approved):
Thanks for the fix!

PR #12865, opened 2022-06-02
Exclude embedded module-info.class recursively

wborn:
See: https://github.com/openhab/openhab-addons/pull/12772#issuecomment-1145179643


jlaur (approved):
It works, thanks!

PR #12869, opened 2022-06-04
[netatmo] Correct Discovery service

clinique:
While previous behavior seemed to work, it made the weather station child of a home, being inconsistant with ModuleType description and expected API calls (e.g. calls to measurement channels were broken).

On netatmo side, this dependency exists but is partially reflected in API calls structure, most of them still being based on initial organization (before the introduction of homes for energy and security purpose).

### DIFF
--- a/bundles/org.openhab.binding.netatmo/src/main/java/org/openhab/binding/netatmo/internal/discovery/NetatmoDiscoveryService.java
+++ b/bundles/org.openhab.binding.netatmo/src/main/java/org/openhab/binding/netatmo/internal/discovery/NetatmoDiscoveryService.java
@@ -14,7 +14,6 @@ package org.openhab.binding.netatmo.internal.discovery;
 
 import java.util.HashMap;
 import java.util.Map;
-import java.util.Set;
 import java.util.stream.Collectors;
 
 import org.eclipse.jdt.annotation.NonNullByDefault;
@@ -48,13 +47,13 @@ import org.slf4j.LoggerFactory;
  */
 @NonNullByDefault
 public class NetatmoDiscoveryService extends AbstractDiscoveryService implements ThingHandlerService, DiscoveryService {
-    private static final Set<ModuleType> SKIPPED_TYPES = Set.of(ModuleType.UNKNOWN, ModuleType.ACCOUNT);
     private static final int DISCOVER_TIMEOUT_SECONDS = 3;
     private final Logger logger = LoggerFactory.getLogger(NetatmoDiscoveryService.class);
+
     private @Nullable ApiBridgeHandler handler;
 
     public NetatmoDiscoveryService() {
-        super(ModuleType.AS_SET.stream().filter(mt -> !SKIPPED_TYPES.contains(mt)).map(mt -> mt.thingTypeUID)
+        super(ModuleType.AS_SET.stream().filter(mt -> !mt.apiName.isBlank()).map(mt -> mt.thingTypeUID)
                 .collect(Collectors.toSet()), DISCOVER_TIMEOUT_SECONDS);
     }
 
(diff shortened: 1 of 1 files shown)


lolodomo (approved):
LGTM

lolodomo:
I propose @jlaur or me to test that before merging.

jlaur:
> I propose @jlaur or me to test that before merging.

I didn't previously focus on testing discovery, but I remember you did some testing with findings and are more into the problem, so please go ahead without me. :-)

lolodomo:
I have my account bridge setup in a configuration file.
Everything is then discovered when I start a discovery but my weather station and its modules are all in doubled.
One with the following UID `netatmo:weather-station:compte:0123456789ab` is good and when approved, the thing went ONLINE and the hierarchy is OK.
For the other one with the following UID `netatmo:weather-station:compte:xxxxxxxxxxxxxxxx:0123456789ab`, if I approve it, the thing remains OFFLINE because it is linked to an unknwon parent `netatmo:home:compte:xxxxxxxxxxxxxxxx`.

For the home, the camera and the persons, it looks good, I can approve them, the hierarchy is good and they went ONLINE.

If it works for security stuff, it should work too for the weather station.

lolodomo:
> For the other one with the following UID `netatmo:weather-station:compte:xxxxxxxxxxxxxxxx:0123456789ab`, if I approve it, the thing remains OFFLINE because it is linked to an unknwon parent `netatmo:home:compte:xxxxxxxxxxxxxxxx`.

This is certainly the discovery through the homesdata that creates these duplicated things. Wasn't your code trying to filter them ?

lolodomo:
@clinique, I think your new filter is not correct:
```
                    homeApi.getHomesData(null, null).stream()
                            .filter(h -> !(h.getFeatures().isEmpty()
                                    || h.getFeatures().contains(FeatureArea.WEATHER) && h.getFeatures().size() == 1))
```
In my case, I have several feature areas.

bundles/org.openhab.binding.netatmo/src/main/java/org/openhab/binding/netatmo/internal/discovery/NetatmoDiscoveryService.java
```diff
@@ -71,40 +70,44 @@ public void startScan() {
...
-                        home.getModules().values().stream().filter(module -> module.getBridge() == null)
-                                .forEach(device -> bridgesUids.put(device.getId(), createThing(device, homeUID)));
-                        // Then the others
-                        home.getModules().values().stream().filter(module -> module.getBridge() != null).forEach(
-                                device -> createThing(device, bridgesUids.getOrDefault(device.getBridge(), homeUID)));
-                    });
+                if (homeApi != null) { // Search those depending from a home that has modules + not only weather modules
+                    homeApi.getHomesData(null, null).stream()
+                            .filter(h -> !(h.getFeatures().isEmpty()
+                                    || h.getFeatures().contains(FeatureArea.WEATHER) && h.getFeatures().size() == 1))
```
lolodomo, line 86: I think your new filter is not correct.
In my case, I have certainly several feature areas and it leads to the create of a thing for my weather station attached to the home thing.

lolodomo (changes requested):

bundles/org.openhab.binding.netatmo/src/main/java/org/openhab/binding/netatmo/internal/discovery/NetatmoDiscoveryService.java
clinique: I want to avoid the creation of a home that only owns a station.
Normally, your station should have been created and discovered in the previous step, but maybe I made a mistake there.

The thing is, my own setup has three homes, one for weather, one for energy, one for security. Never found how to merge them. The weather home is pointless.

lolodomo: Ok, I understand, your filter is for your case you have a home only for your weather station.
In my case, my home contains a weather station and a camera. In this case, the weather station should be ignored. This is an additional test to do in in the loop so that only security and energy things are created at this place.

lolodomo: > Normally, your station should have been created and discovered in the previous step, but maybe I made a mistake there.

Exactly, that is why it should not be discovered a second time, attached to my home that time.

lolodomo: > If its the case, we must find how to avoid creation of a second station.

A simple test on the thing type ? I do not remember exactly how look the returned data.

lolodomo: I just checked the returned data, there is a "type" field for each "thing". Type "NAMain" should be ignored.

lolodomo: I just checked the returned data, there is a "type" field for each "thing" in the home. Type "NAMain" should be ignored.

lolodomo: Once the correct filter will be added, you can even remove your own filter which cover only a specific case that can be covered by the more general test.

lolodomo: @clinique: here are the changes which works for me:
```
                                // Creating modules that have no bridge first
                                home.getModules().values().stream()
                                        .filter(module -> module.getBridge() == null
                                                && module.getType() != ModuleType.WEATHER_STATION)
                                        .forEach(device -> bridgesUids.put(device.getId(),
                                                createThing(device, homeUID)));

                                // Then the others
                                home.getModules().values().stream()
                                        .filter(module -> module.getBridge() != null
                                                && module.getType().getBridge() != ModuleType.WEATHER_STATION)
                                        .forEach(device -> createThing(device,
                                                bridgesUids.getOrDefault(device.getBridge(), homeUID)));
```

lolodomo:
I will test again with the last version of the code and I believe it will work ;)

lolodomo (approved):
Code looks good to me and change was tested with success.
Thank you.

PR #12877, opened 2022-06-04
[elroconnects] Set thingStatus earlier in initialize

mherwege:

- Set thing status UNKNOWN in initialize before launching long running init
- Implement bridgeStatusChanged

### DIFF
--- a/bundles/org.openhab.binding.elroconnects/src/main/java/org/openhab/binding/elroconnects/internal/handler/ElroConnectsDeviceHandler.java
+++ b/bundles/org.openhab.binding.elroconnects/src/main/java/org/openhab/binding/elroconnects/internal/handler/ElroConnectsDeviceHandler.java
@@ -66,8 +66,6 @@ public class ElroConnectsDeviceHandler extends BaseThingHandler {
                     String msg = String.format("@text/offline.invalid-device-type [ \"%s\" ]", deviceType);
                     updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.CONFIGURATION_ERROR, msg);
                 } else {
-                    updateStatus(ThingStatus.UNKNOWN);
-
                     bridgeHandler.setDeviceHandler(deviceId, this);
                     updateProperties(bridgeHandler);
                     updateDeviceName(bridgeHandler);
--- a/bundles/org.openhab.binding.elroconnects/src/main/java/org/openhab/binding/elroconnects/internal/handler/ElroConnectsDeviceHandler.java
+++ b/bundles/org.openhab.binding.elroconnects/src/main/java/org/openhab/binding/elroconnects/internal/handler/ElroConnectsDeviceHandler.java
@@ -49,16 +49,19 @@ public class ElroConnectsDeviceHandler extends BaseThingHandler {
         ElroConnectsDeviceConfiguration config = getConfigAs(ElroConnectsDeviceConfiguration.class);
         deviceId = config.deviceId;
 
-        Bridge bridge = getBridge();
         ElroConnectsBridgeHandler bridgeHandler = getBridgeHandler();
+        if (bridgeHandler == null) {
+            // Thing status has already been updated in getBridgeHandler()
+            return;
+        }
 
-        if ((bridge != null) && (bridgeHandler != null) && (bridge.getStatus() == ThingStatus.ONLINE)) {
+        if (bridgeHandler.getThing().getStatus() == ThingStatus.ONLINE) {
             bridgeHandler.setDeviceHandler(deviceId, this);
             updateProperties(bridgeHandler);
             updateDeviceName(bridgeHandler);
             refreshChannels(bridgeHandler);
         } else {
-            updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.BRIDGE_UNINITIALIZED);
+            updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.BRIDGE_OFFLINE);
         }
     }
 
--- a/bundles/org.openhab.binding.elroconnects/src/main/java/org/openhab/binding/elroconnects/internal/handler/ElroConnectsDeviceHandler.java
+++ b/bundles/org.openhab.binding.elroconnects/src/main/java/org/openhab/binding/elroconnects/internal/handler/ElroConnectsDeviceHandler.java
@@ -63,7 +63,7 @@ public class ElroConnectsDeviceHandler extends BaseThingHandler {
             if (device != null) {
                 ElroDeviceType deviceType = TYPE_MAP.get(device.getDeviceType());
                 if ((deviceType == null) || !thing.getThingTypeUID().equals(THING_TYPE_MAP.get(deviceType))) {
-                    String msg = String.format("@text/offline.invalid-device-type [ \"%s\" ]", device.getDeviceType());
+                    String msg = String.format("@text/offline.invalid-device-type [ \"%s\" ]", deviceType);
                     updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.CONFIGURATION_ERROR, msg);
                 } else {
                     updateStatus(ThingStatus.UNKNOWN);
--- a/bundles/org.openhab.binding.elroconnects/src/main/java/org/openhab/binding/elroconnects/internal/handler/ElroConnectsDeviceHandler.java
+++ b/bundles/org.openhab.binding.elroconnects/src/main/java/org/openhab/binding/elroconnects/internal/handler/ElroConnectsDeviceHandler.java
@@ -63,7 +63,7 @@ public class ElroConnectsDeviceHandler extends BaseThingHandler {
             if (device != null) {
                 ElroDeviceType deviceType = TYPE_MAP.get(device.getDeviceType());
                 if ((deviceType == null) || !thing.getThingTypeUID().equals(THING_TYPE_MAP.get(deviceType))) {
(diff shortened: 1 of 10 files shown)


bundles/org.openhab.binding.elroconnects/src/main/java/org/openhab/binding/elroconnects/internal/handler/ElroConnectsDeviceHandler.java
```diff
@@ -48,13 +49,16 @@ public void initialize() {
...
 
-        if (bridgeHandler != null) {
+        if ((bridge != null) && (bridgeHandler != null) && (bridge.getStatus() == ThingStatus.ONLINE)) {
             bridgeHandler.setDeviceHandler(deviceId, this);
             updateProperties(bridgeHandler);
             updateDeviceName(bridgeHandler);
             refreshChannels(bridgeHandler);
+        } else {
+            updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.BRIDGE_UNINITIALIZED);
         }
```
lolodomo, line 62: You are already setting the thing status to OFFLINE in getBridgeHandler() if there is no bridge or bridge handler is null.
Rather use BRIDGE_OFFLINE is the last updateStatus.
So I would rather suggest the following code:
```
        ElroConnectsBridgeHandler bridgeHandler = getBridgeHandler();
        if (bridgeHandler == null) {
             // Thing status has already been updated in getBridgeHandler()
             return;
        }

        if (bridgeHandler.getThing().getStatus() == ThingStatus.ONLINE) {
            bridgeHandler.setDeviceHandler(deviceId, this);
            updateProperties(bridgeHandler);
            updateDeviceName(bridgeHandler);
            refreshChannels(bridgeHandler);
        } else {
        }
```

I do not see where the status is set to ONLINE. Maybe it should be added there.

I also found that two methods `updateStatus` were override just to call the super method. I do not understand why.

lolodomo (changes requested):

bundles/org.openhab.binding.elroconnects/src/main/java/org/openhab/binding/elroconnects/internal/handler/ElroConnectsDeviceHandler.java
mherwege: @lolodomo Thank you for the simplification, I changed it.

The thing status is utlimately set in the refreshChannels call.
I need to override the methods because they are protected in the BaseThingHandler. I call these from outside the package.

lolodomo: Shouldn't you cover the case your device id is invalid ?
Something like that:
```
        if (bridgeHandler.getThing().getStatus() == ThingStatus.ONLINE) {
            ElroConnectsDevice device = bridgeHandler.getDevice(deviceId);
            if (device != null) {
                bridgeHandler.setDeviceHandler(deviceId, this);
                updateProperties(bridgeHandler);
                updateDeviceName(bridgeHandler);
                refreshChannels(bridgeHandler);
            } else {
                updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.CONFIGURATION_ERROR, "xxxxxx");
            }
        } else {
        }
```

mherwege: @lolodomo Indeed, good catch. I typically use discovery in testing, and then the problem does not appear. And as the refresh did not work in the case of a wrong deviceId, the thing stayed in initializing state.
I actually needed one more check. The user could create a thing of the wrong type, even if the deviceId is valid in the controller. So I am now checking for that as well.
For safety, I have inserted setting thing status to unknown before calling refresh. If something is still not covered, it would not keep showing INITIALIZING.
Many thanks for your eye for detail and your torough review.

bundles/org.openhab.binding.elroconnects/src/main/java/org/openhab/binding/elroconnects/internal/handler/ElroConnectsDeviceHandler.java
```diff
@@ -56,10 +59,22 @@ public void initialize() {
...
         if (bridgeHandler.getThing().getStatus() == ThingStatus.ONLINE) {
-            bridgeHandler.setDeviceHandler(deviceId, this);
-            updateProperties(bridgeHandler);
-            updateDeviceName(bridgeHandler);
-            refreshChannels(bridgeHandler);
+            ElroConnectsDevice device = bridgeHandler.getDevice(deviceId);
+            if (device != null) {
+                ElroDeviceType deviceType = TYPE_MAP.get(device.getDeviceType());
+                if ((deviceType == null) || !thing.getThingTypeUID().equals(THING_TYPE_MAP.get(deviceType))) {
+                    String msg = String.format("@text/offline.invalid-device-type [ \"%d\" ]", deviceType);
```
lolodomo, line 66: Take care, deviceType  is apparently not a number and it can be also null.

lolodomo (changes requested):

bundles/org.openhab.binding.elroconnects/src/main/java/org/openhab/binding/elroconnects/internal/handler/ElroConnectsDeviceHandler.java
lolodomo: String.format will not throw an exception if deviceType is null ?
I do not remember if String.format handles properly null parameters.

mherwege: Indeed, should be %s in this case.
deviceType can indeed be null if it is not found in the TYPE_MAP. That would be an issue with the binding not knowing the device type. At the moment, the binding maps all device types also known by the official app. I considered separately logging it but decided to keep it simple. The formatter will just show null in the string. This can anyhow only happen when a user tries to manually configure, something which is not obvious, because the deviceId can only be found through discovery, or console commands. And it is a required parameter.

bundles/org.openhab.binding.elroconnects/src/main/java/org/openhab/binding/elroconnects/internal/handler/ElroConnectsDeviceHandler.java
```diff
@@ -49,12 +53,32 @@ public void initialize() {
...
-            refreshChannels(bridgeHandler);
+        if (bridgeHandler.getThing().getStatus() == ThingStatus.ONLINE) {
+            ElroConnectsDevice device = bridgeHandler.getDevice(deviceId);
+            if (device != null) {
+                ElroDeviceType deviceType = TYPE_MAP.get(device.getDeviceType());
+                if ((deviceType == null) || !thing.getThingTypeUID().equals(THING_TYPE_MAP.get(deviceType))) {
+                    String msg = String.format("@text/offline.invalid-device-type [ \"%s\" ]", device.getDeviceType());
+                    updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.CONFIGURATION_ERROR, msg);
+                } else {
+                    updateStatus(ThingStatus.UNKNOWN);
```
lolodomo, line 69: In case you are sure that `refreshChannels(bridgeHandler)` will set a thing status quickly and in the current thread, I do not see the need to set the thing status to UNKNOWN. It would lead to the sequence OFFLINE => UNKNOWN => ONLINE maybe for no reason.
I let you confirm as it simply depends on `refreshChannels` implementation (I have not checkoed).

bundles/org.openhab.binding.elroconnects/src/main/java/org/openhab/binding/elroconnects/internal/handler/ElroConnectsDeviceHandler.java
lolodomo: Ok

lolodomo:
Just one remaining comment/question.
After that, let me know if you want to do more testing or if I can merge.

bundles/org.openhab.binding.elroconnects/src/main/java/org/openhab/binding/elroconnects/internal/handler/ElroConnectsDeviceHandler.java
mherwege: I checked the javadoc, and String.format should write out null if the value is null.
Anyway, checking again, the value can actually not be null here. The bridge handler will not put the device in its devices map when the type is not known in the binding. So it should never be null. The only reason to keep the null check is for null type checking (get can return null in theory).
So my last fix is not required, and I will actually revert it. The type enum is more meaningful and easier to understand than the 4 digit hex codes behind it.

bundles/org.openhab.binding.elroconnects/src/main/java/org/openhab/binding/elroconnects/internal/handler/ElroConnectsDeviceHandler.java
mherwege: It should, but I found one case where it might fall through when an unexpected message was received. So I fixed that.

mherwege:
@lolodomo While I revisited all device classes touched by the refresh, I also updates the status messages to make them localizable. I hope you are OK with this last change from me.
Apart from that, my testing shows it working as expected. If you are OK with it, feel free to merge.

lolodomo (approved):
LGTM, thank you

PR #12880, opened 2022-06-04
[upnpcontrol] Set thingStatus UNKNOWN in intialize

mherwege:

Set thing status to UNKNOWN before long running init jobs

### DIFF
--- a/bundles/org.openhab.binding.upnpcontrol/src/main/java/org/openhab/binding/upnpcontrol/internal/handler/UpnpHandler.java
+++ b/bundles/org.openhab.binding.upnpcontrol/src/main/java/org/openhab/binding/upnpcontrol/internal/handler/UpnpHandler.java
@@ -189,10 +189,10 @@ public abstract class UpnpHandler extends BaseThingHandler implements UpnpIOPart
      * To be called from implementing classes when initializing the device, to start initialization refresh
      */
     protected void initDevice() {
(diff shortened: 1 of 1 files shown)


bundles/org.openhab.binding.upnpcontrol/src/main/java/org/openhab/binding/upnpcontrol/internal/handler/UpnpHandler.java
```diff
@@ -189,6 +189,8 @@ private void cancelPollingJob() {
      * To be called from implementing classes when initializing the device, to start initialization refresh
      */
     protected void initDevice() {
+        updateStatus(ThingStatus.UNKNOWN);
```
jlaur, line 192: This could be moved inside the `if` below to avoid two updates in case of configuration error.

jlaur (changes requested):
Thanks! Small improvement suggestion added.

bundles/org.openhab.binding.upnpcontrol/src/main/java/org/openhab/binding/upnpcontrol/internal/handler/UpnpHandler.java
mherwege: @jlaur Thanks and done.

jlaur (approved):
LGTM - thanks!

PR #12883, opened 2022-06-04
[luxom] Add missing default translations

lolodomo:

### DIFF
--- a/bundles/org.openhab.binding.luxom/src/main/resources/OH-INF/thing/thing-types.xml
+++ b/bundles/org.openhab.binding.luxom/src/main/resources/OH-INF/thing/thing-types.xml
@@ -6,22 +6,22 @@
 
 	<bridge-type id="bridge">
 		<label>Luxom Bridge</label>
-		<description>This bridge represents a Luxom IP-interface (for example a DS-65L)</description>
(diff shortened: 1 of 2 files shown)


bundles/org.openhab.binding.luxom/src/main/resources/OH-INF/thing/thing-types.xml
```diff
@@ -34,7 +34,7 @@
 			<bridge-type-ref id="bridge"/>
 		</supported-bridge-type-refs>
 		<label>switch</label>
```
jlaur, line 36:
```suggestion
		<label>Switch</label>
```

jlaur (changes requested):
`thing-type.luxom.switch.label` was inconsistent with the corresponding string in the XML, so needs to be corrected here.

bundles/org.openhab.binding.luxom/src/main/resources/OH-INF/thing/thing-types.xml
lolodomo: You're right.

lolodomo: Changed

jlaur (approved):
LGTM

PR #12884, opened 2022-06-04
[roku] Fix reporting of Roku Home app

mlobstein:
The '/query/active-app' endpoint now reports an id of 562859 instead of -1 when on the Roku Home screen. This change will revert the id back to -1 so that the Active App channel's drop down can correctly report when the Roku is on the home screen.

### DIFF
--- a/bundles/org.openhab.binding.roku/src/main/java/org/openhab/binding/roku/internal/RokuBindingConstants.java
+++ b/bundles/org.openhab.binding.roku/src/main/java/org/openhab/binding/roku/internal/RokuBindingConstants.java
@@ -70,6 +70,7 @@ public class RokuBindingConstants {
     public static final String EMPTY = "";
     public static final String ROKU_HOME = "Roku Home";
     public static final String ROKU_HOME_ID = "-1";
+    public static final String ROKU_HOME_ID_562859 = "562859";
     public static final String ROKU_HOME_BUTTON = "Home";
     public static final String NON_DIGIT_PATTERN = "[^\\d]";
     public static final String TV_APP = "tvinput.dtv";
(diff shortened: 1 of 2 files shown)


jlaur:
@mlobstein - do you know why it reports this id? Is it documented? On which systems is it observed?

mlobstein:
> @mlobstein - do you know why it reports this id? Is it documented? On which systems is it observed?

I just started integrating this binding into my setup (using a Roku Ultra) and noticed this today. I am assuming that this behavior started with the Roku v11.0.0 software which started going out to devices in late March of this year.

jlaur:
> > @mlobstein - do you know why it reports this id? Is it documented? On which systems is it observed?
> 
> I just started integrating this binding into my setup (using a Roku Ultra) and noticed this today. I am assuming that this behavior started with the Roku v11.0.0 software which started going out to devices in late March of this year.

Would it be possible to confirm this behavior and id from another Roku user? Asking because it seems strange, and I guess it's not impossible that something on your specific device is causing this?

mlobstein:
> Would it be possible to confirm this behavior and id from another Roku user? Asking because it seems strange, and I guess it's not impossible that something on your specific device is causing this?

I asked another user on the community for confirmation. Although after looking into it more I am pretty sure this is a new feature in Roku 11. When I check the '/query/active-app' end point while on the home screen, this is the response:
image

However, this app id is not reported in the installed app list:
image

I notice now that on the home screen there now appears an ad with a play button to play a preview of some content. I would surmise that this is the new "Roku Dynamic Menu" that replaced the original, non-dynamic one. But likely the original still exists for older devices that can't support the dynamic home screen. For the binding we need to support all permutations so adding this work-around is allow both versions of the home screen to be reported correctly in the Active App drop down.

jlaur:
@mlobstein - thanks for the explanation. To speed things up: Did you get confirmation from the user you asked?

mlobstein:
> Did you get confirmation from the user you asked?

Yes another user also confirmed this behavior started recently:
https://community.openhab.org/t/roku-binding-power-for-tvs/136223/5

PR #12889, opened 2022-06-05
[elroconnects] Fix array out of bound exception

mherwege:


jlaur (approved):
Thanks for the fix!

PR #16, opened 2014-10-03
Initial contribution of IRtrans 2.0 binding

kgoderis:
A few notes:
1. It uses a Bridge-Thing flow
2. It borrows code from the TCP 1.x binding, e.g. it based on java.nio Selectors, and therefore could be refactored into a TCP 2.0 binding. It also has the code to support a binding as a server/listener, and the code to support bindings that need to make/manage more than one outbound connection, although the IRtrans binding only have 1 outbound connection per BridgeHandler, and it does not listens for incoming connections from the remote devices.

Signed-off-by: Karel Goderis karel.goderis@me.com (github: kgoderis)

PR #12902, opened 2022-06-07
[hdpowerview] Corrections to shade database and capabilities

andrewfg:
In the Home Assistant integration [see Issue #16](https://github.com/sander76/aio-powerview-api/issues/16), the developers, (supported by many other kind users), have assembled an extended collection of shade types, and tested the actual capabilities of existing and newly discovered shade types => @kingy444 many thanks for this great effort :)

In this PR I have updated the OpenHAB binding to reflect the learnings from the Home Assistant team.

Specific changes are..
- Capabilities 0: Removed tilt support
- Capabilities 3: Changed from `tiltOnClosed()` to `tiltAnywhere()`
- Capabilities 4: Removed tilt support
- Capabilities 9: Changed from `tiltAnywhere()` to `tiltOnClosed()`
- Capabilities 8 & 9: Fixed bugs in the position getting and setting functions
- Shade type 1: Added to database
- Shade type 18: Model name corrected
- Shade type 44: Added shade specific `tiltOnClosed()` support to override the capabilities 0 change above
- Shade types 69 .. 71:  Changed capabilities from 3 to 4

Furthermore, I have modified the JUnit tests to test the Shade Position interlocking in accordance with the above functionality changes.


jlaur (changes requested):
Thanks! I'll have a closer look tomorrow and also verify if `tiltOnClosed()` removal for capabilities 0 has any impact on my Twist shades.

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/database/ShadeCapabilitiesDatabase.java
```diff
@@ -40,16 +40,17 @@ public class ShadeCapabilitiesDatabase {
...
+            new Capabilities(3).primary()        .tiltAnywhere().tilt180()                      .text("Vertical Tilt 180°"),
+            new Capabilities(4).primary()                                                       .text("Vertical"),
             new Capabilities(5)                  .tiltAnywhere().tilt180()                      .text("Tilt Only 180°"),
             new Capabilities(6).primaryInverted()                                               .text("Top Down"),
             new Capabilities(7).primary()                                 .secondary()          .text("Top Down Bottom Up"),
             new Capabilities(8).primary()                                 .secondaryOverlapped().text("Dual Overlapped"),
-            new Capabilities(9).primary()        .tiltAnywhere()          .secondaryOverlapped().text("Dual Overlapped Tilt 90°"),
+            new Capabilities(9).primary()        .tiltOnClosed()          .secondaryOverlapped().text("Dual Overlapped Tilt 90°")
+                               .comment("The 'tiltOnClosed()' applies to the primary shade"),
```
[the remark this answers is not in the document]
andrewfg: > suggest changing it to a source code comment to not be included runtime

I created these as runtime comments in case they may in future be needed. I think it is better that way.

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/database/ShadeCapabilitiesDatabase.java
```diff
@@ -58,18 +59,20 @@ public class ShadeCapabilitiesDatabase {
...
             new Type( 9).capabilities(7).text("Duette DuoLite Top Down Bottom Up"),
-            new Type(18).capabilities(1).text("Silhouette"),
+            new Type(18).capabilities(1).text("Pirouette"),
             new Type(23).capabilities(1).text("Silhouette"),
             new Type(38).capabilities(9).text("Silhouette Duolite"),
             new Type(42).capabilities(0).text("M25T Roller Blind"),
             new Type(43).capabilities(1).text("Facette"),
-            new Type(44).capabilities(0).text("Twist"),
+            new Type(44).capabilities(0).text("Twist")
+                        .comment("Shade has functionality of a capabilities 1 shade").typeCapabilities(1),
```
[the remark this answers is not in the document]
andrewfg: Ditto

andrewfg:
> verify if tiltOnClosed() removal for capabilities 0 has any impact on my Twist shades.

I am pretty sure it does not. My own shades are Twist..

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/database/ShadeCapabilitiesDatabase.java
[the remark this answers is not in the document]
andrewfg: Technically Type 44 (Twist) shades report themselves as Capabilities 0, which technically means that they are simple roller blinds, without any vanes. But as we both know, this type of shade does actually emulate a vane/tilt function by slightly moving the overlapping bands to cover / not cover each other; and they do therefore also support the OH binding vane/tilt functionality. In other words, they report no vane/tilt capability even though they do actually support it. So this `typeCapabilities()` attribute is used to tell the binding: _"even if the blind technically reports a certain capabilities level, please actually use this other typeCapabilities level instead"_.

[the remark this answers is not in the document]
jlaur: So difference compared to simply declaring `capabilities(1)` is that you use text from capabilities 0 (visible as property), but attributes from capabilities 1?

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/database/ShadeCapabilitiesDatabase.java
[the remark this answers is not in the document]
jlaur: In that case they could be added in the future when needed. :-) I guess only purpose would be for debug logging, but in which case would it be relevant?

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/database/ShadeCapabilitiesDatabase.java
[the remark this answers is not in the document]
andrewfg: Yes.

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/database/ShadeCapabilitiesDatabase.java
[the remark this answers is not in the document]
andrewfg: Removed comment attributes

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/database/ShadeCapabilitiesDatabase.java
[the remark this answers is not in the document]
jlaur: OK, finally I had time to understand it from your code and explanation, so basically a work-around to override capabilities when wrongly reported by the API. I see now from returned JSON as you wrote:
```
"type":44,
"capabilities":0
```

So indeed inconsistent. I checked also the code you have provided for now looking up capabilities in internal database first, and only falling back to capabilities provided by JSON response if not found. Looks good. The only thing nagging me is that capabilities in database overrides reported capabilities - but also typeCapabilities overrides the already overridden capabilities. Order of precedence:
- typeCapabilities in database (if present) from type
- capabilities in database (if present) from type
- capabilities in response

So some complexity introduced only to provide a different text for the overridden capabilities. But decoupling texts from capabilities in itself would also add complexity, so I don't have a good proposal right now. At least this is only for Twist. Resolving comment.

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/database/ShadeCapabilitiesDatabase.java
```diff
@@ -310,6 +328,16 @@ public Capabilities getCapabilities(int capabilities) {
...
+    /**
+     * Return a Capabilities class instance that corresponds to the given 'type' parameter.
+     *
+     * @param type the shade type.
+     * @return corresponding instance of Capabilities class.
+     */
+    public Capabilities getCapabilitiesForType(int type) {
+        return getCapabilities(TYPE_DATABASE.getOrDefault(type, new Type()).getTypeCapabilities());
+    }
+
```
jlaur, line 340: Suggestion: Could we encapsulate/hide the "type override" complexity within this class by merging this method with previous method into this signature:
```
public Capabilities getCapabilities(int type, int capabilities)
```

and then simply return capabilities according to type and capabilities? This would move the logic from HDPowerViewShadeHandler into this specialized class.

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/database/ShadeCapabilitiesDatabase.java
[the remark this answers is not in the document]
andrewfg: I hear what you say about the complexity. The capabilities and types database is a kind of internal record of the state of knowledge about shade types and how they function. For example it is used for testing the binding functionality for different types in the JUnit `ShadePositionTests` class, so the tests can be done without the need for developers to physically own an example of all the different types of shades. And when the binding is online, it reads the `type` and `capabilities` reported by the shade and cross checks these against what is in the database. And if there is any kind of discrepancy the binding produces a warning in the log to contact us (the developers). This will allow us to expand our knowledge over time whenever new types of shades and/or new capabilities are discovered, and update the code accordingly.

jlaur (changes requested):
Still some doubts, sorry.

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/database/ShadeCapabilitiesDatabase.java
andrewfg: But maybe we can use only the actual data in the property, vs the functional capabilties via that method. Let me look at it again..

andrewfg:
I will get back tomorrow..

andrewfg:
So for example a shade would only report a property "tilt detected" if a JSON packet actually contained a posKind 3 entry. And such property might only be actualised only after the user had actually successfully executed a tilt command. Which may be never, if the user never issues such command. Hmm..

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/database/ShadeCapabilitiesDatabase.java
andrewfg: I made changes. See comment below. So marking this now as resolved.

andrewfg:
@jlaur upon reflection overnight (I woke up at 4 AM) I have made some changes (committed just now)..

1. Adopted your suggestion to move the functionality of overridden capabilities from the handler to the second `getCapabilities()` method (see code below).
2. Refactored the field name `typeCapabilities` to `capabilitiesOverride` to make it clearer what is going on.
3. Improved the Java Doc for the two `getCapabilities()` methods to make it clearer what is going on.

``` 
    /**
     * Return a Capabilities class instance that corresponds to the given 'capabilitiesId' parameter. If the
     * 'capabilitiesId' parameter is for a valid capabilities entry in the database, then that respective Capabilities
     * class instance is returned. Otherwise a blank Capabilities class instance is returned.
     *
     * @param capabilitiesId the target capabilities Id.
     * @return corresponding Capabilities class instance.
     */
    public Capabilities getCapabilities(@Nullable Integer capabilitiesId) {
        return CAPABILITIES_DATABASE.getOrDefault(capabilitiesId != null ? capabilitiesId.intValue() : -1,
                new Capabilities());
    }

    /**
     * Return a Capabilities class instance that corresponds to the given 'typeId' parameter. If the 'typeId' parameter
     * is a valid type in the database, and it has a 'capabilitiesOverride' value, then an instance of the respective
     * overridden Capabilities class is returned. Otherwise if the 'capabilitiesId' parameter is for a valid
     * capabilities entry in the database, then that respective Capabilities class instance is returned. Otherwise a
     * blank Capabilities class instance is returned.
     *
     * @param typeId the target shade type Id (to check if it has a 'capabilitiesOverride' value).
     * @param capabilitiesId the target capabilities value (when type Id does not have a 'capabilitiesOverride').
     * @return corresponding Capabilities class instance.
     */
    public Capabilities getCapabilities(int typeId, @Nullable Integer capabilitiesId) {
        int targetCapabilities = TYPE_DATABASE.getOrDefault(typeId, new Type()).getCapabilitiesOverride();
        if (targetCapabilities < 0) {
            targetCapabilities = capabilitiesId != null ? capabilitiesId.intValue() : -1;
        }
        return getCapabilities(targetCapabilities);
    }
```

And furthermore, I propose the following change (not yet committed, awaiting your feedback)..

4. I think we should DELETE the `secondaryRailDetected` and `tiltAnywhereDetected` properties from the UI; these depend on dynamic data from the JSON `positions` element, so the UI display may actually change depending on the history of commands sent to or received from the shade. In my opinion if a user has a problem that would require us to know these 'properties', then we can (and probably should) get the information via `log:set TRACE`. 
5. Or alternatively if you insist to retain these properties, then to be fully logical we should also add the other dynamic properties `tiltOnCloseDetected` / `tilt180Detected` / `overlappedSecondaryDetected` to be consistent.

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/database/ShadeCapabilitiesDatabase.java
jlaur: Excellent!

jlaur:
> * I think we should DELETE the `secondaryRailDetected` and `tiltAnywhereDetected` properties from the UI; these depend on dynamic data from the JSON `positions` element, so the UI display may actually change depending on the history of commands sent to or received from the shade. In my opinion if a user has a problem that would require us to know these 'properties', then we can (and probably should) get the information via `log:set TRACE`.
> * Or alternatively if you insist to retain these properties, then to be fully logical we should also add the other dynamic properties `tiltOnCloseDetected` / `tilt180Detected` / `overlappedSecondaryDetected` to be consistent.

I'm totally fine with removing them as they don't provide much value as you also conclude, and could even cause unneeded confusion.

jlaur:
> > * I think we should DELETE the `secondaryRailDetected` and `tiltAnywhereDetected` properties from the UI;

> I'm totally fine with removing them as they don't provide much value as you also conclude, and could even cause unneeded confusion.

@andrewfg - do you want to include it in this PR or another one?

andrewfg:
> do you want to include it in this PR or another one?

@jlaur I will remove them now :)

andrewfg:
@jlaur the properties have now been hidden :)

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/HDPowerViewBindingConstants.java
```diff
@@ -67,8 +67,8 @@ public class HDPowerViewBindingConstants {
     // Shade properties
     public static final String PROPERTY_SHADE_TYPE = "type";
     public static final String PROPERTY_SHADE_CAPABILITIES = "capabilities";
-    public static final String PROPERTY_SECONDARY_RAIL_DETECTED = "secondaryRailDetected";
-    public static final String PROPERTY_TILT_ANYWHERE_DETECTED = "tiltAnywhereDetected";
+    public static final String PROPERTY_SECONDARY_RAIL_DETECTED = "secondaryRailDetected"; // hidden
+    public static final String PROPERTY_TILT_ANYWHERE_DETECTED = "tiltAnywhereDetected"; // hidden
```
jlaur, line 71: This is a bit obscure/confusing. I would suggest to remove the constants from this class and use a different terminology than  "properties". I see an additional advantage in doing that: We can get rid of the **soft** vs. **hard** properties concept.

How about `detectedCapabilities` or something like that? We could then rename method `updateHardProperties` to `updateDetectedCapabilities` and `updateSoftProperties` to simply `updateProperties`.

jlaur (changes requested):
One last thing... :-)

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/HDPowerViewBindingConstants.java
andrewfg: Ok.

andrewfg: Done.

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/handler/HDPowerViewShadeHandler.java
```diff
@@ -75,6 +76,10 @@ private enum RefreshKind {
     private static final String COMMAND_CALIBRATE = "CALIBRATE";
     private static final String COMMAND_IDENTIFY = "IDENTIFY";
 
+    private static final String PROPERTY_SECONDARY_RAIL_DETECTED = "secondaryRailDetected"; // hidden
+    private static final String PROPERTY_TILT_ANYWHERE_DETECTED = "tiltAnywhereDetected"; // hidden
```
jlaur, line 80:
```suggestion
    private static final String SECONDARY_RAIL_DETECTED_KEY = "secondaryRailDetected";
    private static final String TILT_ANYWHERE_DETECTED_KEY = "tiltAnywhereDetected";
```

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/HDPowerViewBindingConstants.java
jlaur: OK, I reject this comment as it's a bit more specific than that - and covers only type/capabilities, not all properties.

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/handler/HDPowerViewShadeHandler.java
andrewfg: Oops. Ok will do it.

andrewfg: Done

jlaur (approved):
Thanks for these improvements!

andrewfg:
@jlaur thanks to you too :)

Issue #12903, opened 2022-06-07
[insteon] device is improperly set to ONLINE when device doesn't exist in modem database


robnielsen:
Currently this condition is detected and a warning is logged:

> device AA.BB.CC not found in the modem database. Did you forget to link?

When this occurs, the device status should be set OFFLINE (CONFIGURATION_ERROR) and include the message that is logged above.

PR #12904, opened 2022-06-07
[insteon] set device offline if it doesn't exist in the plm/hub database

robnielsen:
This fixes https://github.com/openhab/openhab-addons/issues/12903

### DIFF
--- a/bundles/org.openhab.binding.insteon/src/main/java/org/openhab/binding/insteon/internal/handler/InsteonDeviceHandler.java
+++ b/bundles/org.openhab.binding.insteon/src/main/java/org/openhab/binding/insteon/internal/handler/InsteonDeviceHandler.java
@@ -126,6 +126,7 @@ public class InsteonDeviceHandler extends BaseThingHandler {
     private final Logger logger = LoggerFactory.getLogger(InsteonDeviceHandler.class);
 
     private @Nullable InsteonDeviceConfiguration config;
+    private boolean deviceLinked = true;
 
     public InsteonDeviceHandler(Thing thing) {
         super(thing);
@@ -134,6 +135,7 @@ public class InsteonDeviceHandler extends BaseThingHandler {
     @Override
     public void initialize() {
         config = getConfigAs(InsteonDeviceConfiguration.class);
+        deviceLinked = true;
 
         scheduler.execute(() -> {
             final Bridge bridge = getBridge();
@@ -373,7 +375,7 @@ public class InsteonDeviceHandler extends BaseThingHandler {
                 });
 
                 if (ThingStatus.ONLINE == bridge.getStatus()) {
-                    if (ThingStatus.OFFLINE != getThing().getStatus()) {
+                    if (deviceLinked) {
                         updateStatus(ThingStatus.ONLINE);
                     }
                 } else {
@@ -548,6 +550,8 @@ public class InsteonDeviceHandler extends BaseThingHandler {
         String msg = "device with the address '" + address
                 + "' was not found in the modem database. Did you forget to link?";
         updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.CONFIGURATION_ERROR, msg);
+
+        deviceLinked = false;
     }
 
     private InsteonNetworkHandler getInsteonNetworkHandler() {
--- a/bundles/org.openhab.binding.insteon/src/main/java/org/openhab/binding/insteon/internal/handler/InsteonDeviceHandler.java
+++ b/bundles/org.openhab.binding.insteon/src/main/java/org/openhab/binding/insteon/internal/handler/InsteonDeviceHandler.java
@@ -373,7 +373,9 @@ public class InsteonDeviceHandler extends BaseThingHandler {
                 });
 
                 if (ThingStatus.ONLINE == bridge.getStatus()) {
-                    updateStatus(ThingStatus.ONLINE);
+                    if (ThingStatus.OFFLINE != getThing().getStatus()) {
+                        updateStatus(ThingStatus.ONLINE);
+                    }
                 } else {
                     updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.BRIDGE_OFFLINE);
                 }
@@ -531,6 +533,21 @@ public class InsteonDeviceHandler extends BaseThingHandler {
         getInsteonNetworkHandler().unlinked(channelUID);
 
         logger.debug("channel {} unlinked ", channelUID.getAsString());
+    }
+
+    public InsteonAddress getInsteonAddress() {
+        if (config != null) {
+            return new InsteonAddress(config.getAddress());
+        } else {
+            throw new IllegalArgumentException("config is null");
+        }
+    }
+
+    public void deviceNotLinked() {
+        String address = config != null ? config.getAddress() : "unknown";
+        String msg = "device with the address '" + address
+                + "' was not found in the modem database. Did you forget to link?";
+        updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.CONFIGURATION_ERROR, msg);
     }
 
     private InsteonNetworkHandler getInsteonNetworkHandler() {
--- a/bundles/org.openhab.binding.insteon/src/main/java/org/openhab/binding/insteon/internal/handler/InsteonDeviceHandler.java
+++ b/bundles/org.openhab.binding.insteon/src/main/java/org/openhab/binding/insteon/internal/handler/InsteonDeviceHandler.java
@@ -26,7 +26,6 @@ import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
 import org.eclipse.jdt.annotation.NonNullByDefault;
-import org.eclipse.jdt.annotation.Nullable;
 import org.openhab.binding.insteon.internal.InsteonBinding;
 import org.openhab.binding.insteon.internal.InsteonBindingConstants;
 import org.openhab.binding.insteon.internal.config.InsteonChannelConfiguration;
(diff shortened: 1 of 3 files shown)


bundles/org.openhab.binding.insteon/src/main/java/org/openhab/binding/insteon/internal/handler/InsteonDeviceHandler.java
```diff
@@ -373,7 +373,9 @@ public void initialize() {
                 });
 
                 if (ThingStatus.ONLINE == bridge.getStatus()) {
-                    updateStatus(ThingStatus.ONLINE);
+                    if (ThingStatus.OFFLINE != getThing().getStatus()) {
```
lolodomo, line 376: I guess test should be against ONLINE and not OFFLINE.

lolodomo (changes requested):

bundles/org.openhab.binding.insteon/src/main/java/org/openhab/binding/insteon/internal/handler/InsteonDeviceHandler.java
robnielsen: No, it needs to be OFFLINE since the method `deviceNotLinked()` below could get called during initialization above if the device does not exist in the plm/hub database. Another option would be to set a flag and check that instead.

lolodomo (changes requested):

bundles/org.openhab.binding.insteon/src/main/java/org/openhab/binding/insteon/internal/handler/InsteonDeviceHandler.java
```diff
@@ -373,7 +375,9 @@ public void initialize() {
                 });
 
                 if (ThingStatus.ONLINE == bridge.getStatus()) {
-                    updateStatus(ThingStatus.ONLINE);
+                    if (deviceLinked) {
+                        updateStatus(ThingStatus.ONLINE);
+                    }
```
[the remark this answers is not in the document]
robnielsen: Unfortunately, that will not work since the network handler might still be in the process of downloading the database and parsing it. `if (deviceLinked)` only comes into play when a user manually adds a device from the UI and enters an address that isn't in the modem database.

bundles/org.openhab.binding.insteon/src/main/java/org/openhab/binding/insteon/internal/handler/InsteonNetworkHandler.java
```diff
@@ -205,6 +206,20 @@ public void addMissingDevices(List<String> missing) {
...
+            InsteonDeviceHandler handler = (InsteonDeviceHandler) thing.getHandler();
+            if (handler == null) {
+                throw new IllegalArgumentException("handler is null");
+            }
+
+            if (addr.equals(handler.getInsteonAddress())) {
+                handler.deviceNotLinked();
+                return;
+            }
+        });
```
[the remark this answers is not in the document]
robnielsen: six of one, half a dozen of the other...

bundles/org.openhab.binding.insteon/src/main/java/org/openhab/binding/insteon/internal/handler/InsteonDeviceHandler.java
```diff
@@ -533,6 +537,23 @@ public void channelUnlinked(ChannelUID channelUID) {
...
 
+    public InsteonAddress getInsteonAddress() {
+        if (config != null) {
+            return new InsteonAddress(config.getAddress());
+        } else {
+            throw new IllegalArgumentException("config is null");
+        }
+    }
+
+    public void deviceNotLinked() {
```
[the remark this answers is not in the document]
robnielsen: six of one, half a dozen of the other

lolodomo:
@jlaur : please have a look and merge if that is fine for you.

bundles/org.openhab.binding.insteon/src/main/java/org/openhab/binding/insteon/internal/handler/InsteonDeviceHandler.java
```diff
@@ -373,7 +375,9 @@ public void initialize() {
                 });
 
                 if (ThingStatus.ONLINE == bridge.getStatus()) {
-                    updateStatus(ThingStatus.ONLINE);
+                    if (deviceLinked) {
```
jlaur, line 378: If device is not linked, is it intentional that status is not updated?

bundles/org.openhab.binding.insteon/src/main/java/org/openhab/binding/insteon/internal/handler/InsteonDeviceHandler.java
```diff
@@ -533,6 +537,23 @@ public void channelUnlinked(ChannelUID channelUID) {
         logger.debug("channel {} unlinked ", channelUID.getAsString());
     }
 
+    public InsteonAddress getInsteonAddress() {
+        if (config != null) {
+            return new InsteonAddress(config.getAddress());
+        } else {
+            throw new IllegalArgumentException("config is null");
```
jlaur, line 544: This is not a correct exception to throw in this case. As documented: "Thrown to indicate that a method has been passed an illegal or inappropriate argument." Instead, you could consider making config `@NonNullByDefault({})` (from `@Nullable`) since it will always be non-null after `initialize()`.

bundles/org.openhab.binding.insteon/src/main/java/org/openhab/binding/insteon/internal/handler/InsteonDeviceHandler.java
```diff
@@ -533,6 +537,23 @@ public void channelUnlinked(ChannelUID channelUID) {
...
+        }
+    }
+
+    public void deviceNotLinked() {
+        String address = config != null ? config.getAddress() : "unknown";
```
jlaur, line 549: I suppose address could be configured as "unknown", so this is kind of a strange message (_"device with the address 'unknown' was not found"_). See previous comment as this might also solve this one.

bundles/org.openhab.binding.insteon/src/main/java/org/openhab/binding/insteon/internal/handler/InsteonNetworkHandler.java
```diff
@@ -205,6 +206,20 @@ public void addMissingDevices(List<String> missing) {
         });
     }
 
+    public void deviceNotLinked(InsteonAddress addr) {
+        getThing().getThings().stream().forEach((thing) -> {
+            InsteonDeviceHandler handler = (InsteonDeviceHandler) thing.getHandler();
+            if (handler == null) {
+                throw new IllegalArgumentException("handler is null");
```
jlaur, line 213: This is not a correct exception to throw here. If handler is `null`, thing is not initialized yet. In that case you shouldn't interrupt the program flow by throwing an exception that will not be handled. Probably you should store this temporarily so you can tell the thing when it has been initialized? See for example #12113.

jlaur (changes requested):
Thanks. A few comments to consider.

bundles/org.openhab.binding.insteon/src/main/java/org/openhab/binding/insteon/internal/handler/InsteonDeviceHandler.java
robnielsen: The deviceNotLinked() method below will set the status and also set deviceLinked to false.  This code will only come into play when a user manually adds a device and enters the wrong address.

bundles/org.openhab.binding.insteon/src/main/java/org/openhab/binding/insteon/internal/handler/InsteonDeviceHandler.java
robnielsen: changed

bundles/org.openhab.binding.insteon/src/main/java/org/openhab/binding/insteon/internal/handler/InsteonDeviceHandler.java
robnielsen: changed

bundles/org.openhab.binding.insteon/src/main/java/org/openhab/binding/insteon/internal/handler/InsteonNetworkHandler.java
robnielsen: This really can't happen just need to resolve the warning that is thrown.

jlaur: Why do you think it cannot happen? As far as I can see this method is called indirectly from the bridge handler itself (through `InsteonBinding`). If all child handlers are not initialized at this point, it will happen.

robnielsen: Because the way the binding is written above, it waits at least 5 seconds after the last device has been created before the code that falls through this is even hit.

jlaur: In that case I guess you can simply skip those (instead of throwing), and perhaps add it comment why this is safe and cannot happen?

robnielsen: done

jlaur (approved):
Thanks.

PR #12906, opened 2022-06-07
[neohub] Workaround for 'Legacy API Disabled' problem

andrewfg:
### Background
From early 2022 Heatmiser introduced NeoHub firmware that has the ability to enable / disable the NeoHub `portNumber` 4242. If this port is disabled the OpenHAB binding cannot connect and the binding will report a *"Connection Refused"* warning in the log. In prior firmware versions the port was always enabled. But in the new firmware the port is initially enabled on power up but if no communication occurs for 48 hours it is automatically disabled. Alternatively the Heatmiser mobile App has a setting (Settings | System | API Access | Legacy API Enable | On) whereby the port can be permanently enabled.

### Solution
In this PR when the handler `initialize()` method is called, we try to 'ping' the NeoHub. And if the ping fails with *"Connection Refused"* it sets the Bridge status as offline / configuration error, with a message to consult the readme, and also writes a warning message in the log to say the same. And a chapter has been added to the readme to explain how to resolve the issue.
 
### DIFF
--- a/bundles/org.openhab.binding.neohub/README.md
+++ b/bundles/org.openhab.binding.neohub/README.md
@@ -39,7 +39,15 @@ Before the binding can communicate with the hub, the following Configuration Par
 | portNumber              | Port number of the NeoHub (Default=4242)                                                    |
 | pollingInterval         | Time (seconds) between polling requests to the NeoHub (Min=4, Max=60, Default=60)           |
 | socketTimeout           | Time (seconds) to allow for TCP socket connections to the hub to succeed (Min=4, Max=20, Default=5) |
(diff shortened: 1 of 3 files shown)


lolodomo (approved):
LGTM

andrewfg:
@lolodomo many thanks for merging this PR :) -- just for info, probably motivated by my posts on their developer portal, Heatmiser uploaded their API specification v3.0 today. In it, they announced an additional authenticated / encrypted connection channel to their hubs. I have just created #12909 to describe it. And I am starting work on the binding development to implement it now. But in the meantime this PR, just merged, is the official work around.

PR #12913, opened 2022-06-09
[pulseaudio] Fix source format options

GiviMAD:

I made some mistakes when implementing the AudioSource describing the supported formats, I just realize it while testing a new keyword spotter with different bit depths and endianness.
This pr fixes it.
Reference here: https://www.freedesktop.org/wiki/Software/PulseAudio/Documentation/User/SupportedAudioFormats/


jlaur (approved):
Thanks!

lolodomo:
As options labels were updated  the i18n tool has to be re-run to update the default translations.

jlaur:
> As options labels were updated  the i18n tool has to be re-run to update the default translations. 

Sorry, too early for me, somehow missed that second file. @GiviMAD - can you create another PR with that?

GiviMAD:
Thanks @lolodomo, I totally forgot about it.
@jlaur, yes I'll do it.

PR #12915, opened 2022-06-10
[neohub] Add support for WebSocket connection to hub

### LINKED ISSUE #12899: [neohub] Add support for "new" API
Heatmiser has added a new API in order to improve security. Currently the old API is supported in parallel. However I plan to add support for the new API in a forthcoming PR. I am opening this issue as a place holder.

### LINKED ISSUE #12909: [neohub] Add support for WebSocket connection to hub
Heatmiser has released their API specification v3.0 (it is dated 12/12/2021 but was actually uploaded to their developer portal today).

In this new release Heatmiser announce the following..
- New support for discovery of NeoHub devices via UDP LAN broadcast
- New support for certificate authenticated and encrypted connection to the NeoHub via WSS WebSockets
- Existing (legacy) plain socket connection continues to be supported, albeit disabled by default - see #12906

I am planning to add support for these new features to the OpenHAB NeoHub binding (already working on it).
Specifically the following features will be added..
1. Add an auto discovery service for NeoHub devices.
2. Add support for WSS WebSocket connection; the choice of websocket vs. plain socket to be via an OH config param.

---

EDIT: in case anyone needs further background see the following links..
- [Heatmiser API Specification v3.0](https://dev.heatmiser.com/t/api-neohub-v3-0-protocol-8th-june-2022/222)
- [Heatmiser Developer Portal thread](https://dev.heatmiser.com/t/openhab-binding-migrate-from-old-to-new-api/213/16)

andrewfg:
Fixes #12909
Fixes #12899

As described in #12909 the manufacturer Heatmiser has implemented a new way of connecting to NeoHubs by means of a secure WSS web socket connection. This is in addition to the existing non-secure TCP socket connection.

This PR adds support for the OH binding to connect via such a web socket.

The web socket connection requires an additional configuration parameter `apiToken`. Such tokens have to be created by the user for each hub on their Heatmiser mobile App. The read me explains how to do this.

The support for the existing non-secure TCP socket connection is not removed. However if users want to continue using it, they now need to enable that via the Heatmiser mobile App. The read me already explains how to do this.

Note that #12909 also mentions that Heatmiser introduced a new way of discovering hubs on the LAN. However this has not been implemented, since the prior implemented method of discovery still works.


bundles/org.openhab.binding.neohub/src/main/java/org/openhab/binding/neohub/internal/NeoHubHandler.java
```diff
@@ -113,7 +113,7 @@ public void initialize() {
             logger.debug("hub '{}' port={}", getThing().getUID(), config.portNumber);
         }
 
-        if (config.portNumber <= 0 || config.portNumber > 0xFFFF) {
+        if (config.portNumber != PORT_TCP && config.portNumber != PORT_WSS) {
```
jlaur, line 116: This seems too strict, it means you can no longer configure the port to anything but default values for TCP/secure web socket.

bundles/org.openhab.binding.neohub/src/test/java/org/openhab/binding/neohub/test/NeohubsocketTests.java
```diff
@@ -0,0 +1,70 @@
...
+            config.hostName = "192.168.1.xxx";
+            config.socketTimeout = 5;
+            config.apiToken = "12345678-1234-1234-1234-123456789ABC";
+
+            // use web socket to send the request, and log the response
+            try {
+                config.portNumber = NeoHubBindingConstants.PORT_WSS;
+                NeoHubWebSocket socket = new NeoHubWebSocket(config);
+                String requestJson = NeoHubBindingConstants.CMD_CODE_FIRMWARE;
+                System.out.println(requestJson);
```
jlaur, line 48: Six SAT warnings from this class: "System.out.println is used"

bundles/org.openhab.binding.neohub/src/test/java/org/openhab/binding/neohub/test/NeohubsocketTests.java
```diff
@@ -0,0 +1,70 @@
...
+
+    private boolean runOnline = false;
+
+    @SuppressWarnings("null")
+    @Test
+    void testWebSocket() {
+        if (runOnline) {
+            // create the web socket class
+            NeoHubConfiguration config = new NeoHubConfiguration();
+            config.hostName = "192.168.1.xxx";
```
jlaur, line 39: What is actually tested by this test with this configuration?

bundles/org.openhab.binding.neohub/src/main/resources/OH-INF/thing/thing-types.xml
```diff
@@ -53,6 +53,11 @@
 				<default>false</default>
 				<advanced>true</advanced>
 			</parameter>
+
+			<parameter name="apiToken" type="text" required="false">
+				<label>API Access Token</label>
+				<description>API access token for the hub (created on the Heatmiser mobile App)</description>
```
jlaur, line 59:
```suggestion
				<description>API access token for the hub (created with the Heatmiser mobile app)</description>
```

bundles/org.openhab.binding.neohub/src/main/java/org/openhab/binding/neohub/internal/NeoHubWebSocket.java
```diff
@@ -0,0 +1,234 @@
...
+     * @throws NeoHubException
+     */
+    private void startSession() throws NeoHubException {
+        Session session = this.session;
+        if (session == null || !session.isOpen()) {
+            closeSession();
+            try {
+                URI uri = new URI(String.format("wss://%s:%d", config.hostName, config.portNumber));
+                webSocketClient.connect(this, uri).get();
+            } catch (InterruptedException | ExecutionException | IOException | URISyntaxException e) {
```
jlaur, line 105: For `InterruptedException`, you might want to:
```
```

bundles/org.openhab.binding.neohub/src/main/java/org/openhab/binding/neohub/internal/NeoHubWebSocket.java
```diff
@@ -0,0 +1,234 @@
...
+        super(config);
+
+        // initialise and start ssl context factory, http client, web socket client
+        SslContextFactory.Client sslContextFactory = new SslContextFactory.Client();
+        sslContextFactory.setTrustAll(true);
+        HttpClient httpClient = new HttpClient(sslContextFactory);
+        try {
+            httpClient.start();
+        } catch (Exception e) {
+            throw new NeoHubException(String.format("Error starting http client: '%s'", e.getMessage()));
```
jlaur, line 82: Consider chaining exceptions to provide more information for the caller. So in `NeoHubException`, add:

```
    public NeoHubException(String message, Throwable cause) {
        super(message, cause);
    }
```

and then here:
```suggestion
            throw new NeoHubException(String.format("Error starting http client: '%s'", e.getMessage()), e);
```

jlaur (changes requested):
Thanks, nice improvement! Some quick feedback provided.

bundles/org.openhab.binding.neohub/src/main/java/org/openhab/binding/neohub/internal/NeoHubHandler.java
andrewfg: I made it so strict, because the hub actually has those two ports (only) hard coded in the hub. I suppose you could imagine an edge case where a user system had some kind of port forwarding from other port numbers, but it is hardly a common use case.

bundles/org.openhab.binding.neohub/src/test/java/org/openhab/binding/neohub/test/NeohubsocketTests.java
andrewfg: Ok. Will fix.

bundles/org.openhab.binding.neohub/src/test/java/org/openhab/binding/neohub/test/NeohubsocketTests.java
andrewfg: If you set the IP address to a valid address, then it tests actual online communication with a hub on that address. The ‘xxx’ is just a placeholder to bypass online connection testing by default.

bundles/org.openhab.binding.neohub/src/main/resources/OH-INF/thing/thing-types.xml
andrewfg: Ok.

bundles/org.openhab.binding.neohub/src/main/java/org/openhab/binding/neohub/internal/NeoHubHandler.java
andrewfg: And to be more specific, the communication protocol depends on which of the two ports are used. If the client connects on the TCP port, it runs the TCP protocol; if it connects on the WSS port, it runs the WSS protocol; and if tries to connect on any other port, the connection is refused.

bundles/org.openhab.binding.neohub/src/main/java/org/openhab/binding/neohub/internal/NeoHubWebSocket.java
andrewfg: Ok. Done.

bundles/org.openhab.binding.neohub/src/test/java/org/openhab/binding/neohub/test/NeohubsocketTests.java
andrewfg: Done.

bundles/org.openhab.binding.neohub/src/main/java/org/openhab/binding/neohub/internal/NeoHubWebSocket.java
andrewfg: Done

bundles/org.openhab.binding.neohub/src/main/java/org/openhab/binding/neohub/internal/NeoHubHandler.java
jlaur: In that case the configuration seems backwards. Would it make sense to configure which protocol to use, then use appropriate port accordingly? In that case you could even have the protocol configurable, since you would know the protocol not from the port. And port forwarding configurations would then be supported.

andrewfg: Good point.

andrewfg: Done

bundles/org.openhab.binding.neohub/README.md
```diff
@@ -33,21 +33,31 @@ It signs on to the hub using the supplied connection parameters, and it polls th
...
+|----------------------------|----------------------------------------------------------------------------------------------------------|
+| hostName                   | Host name (IP address) of the NeoHub (example 192.168.1.123)                                             |
+| useWebSockets<sup>1)</sup> | Use secure WebSockets to connect to  the NeoHub (example `true`)                                         |
+| apiToken<sup>1)</sup>      | API Access Token for secure connection to hub. Create the token in the Heatmiser mobile App              |
+| pollingInterval            | Time (seconds) between polling requests to the NeoHub (Min=4, Max=60, Default=60)                        |
+| socketTimeout              | Time (seconds) to allow for TCP socket connections to the hub to succeed (Min=4, Max=20, Default=5)      |
+| preferLegacyApi            | ADVANCED: Prefer to use older API calls; but if not supported, it switches to new calls (Default=false)  |
+| portOverride<sup>2)</sup>  | ADVANCED: Port number for connection to the NeoHub (0 = automatic)                                       |
```
jlaur, line 44: Can we rename this back to `portNumber` for backwards compatibility? Also, I think the original name is still perfectly understandable.

andrewfg: > Can we rename this back to portNumber for backwards compatibility?

No.

In existing installations, the users will have defined a portNumber config parameter (normally 4242). In the new version the binding automatically selects the port based on the useWebSockets parameter. So we have to prevent that predefined portNumber config parameter from preventing the automatic selection.

jlaur: OK, I'll have to try out how this works for managed things, if default values will actually be written? This doesn't make sense, since it would then by definition override the default value after saving, and there the value would no longer be defaulted.

Under the assumption that default value will not be written, I guess the most likely scenario for users configuring this parameter would be overriding default port, in which they probably still want the port to be overwritten after upgrading to 3.4? There's no real reason for anyone to configure the port number to be the same as the default port number for this optional parameter.

andrewfg: Ok. Understood. I will try another commit.

bundles/org.openhab.binding.neohub/src/main/resources/OH-INF/i18n/neohub.properties
```diff
@@ -40,6 +40,8 @@ thing-type.config.neohub.neohub.preferLegacyApi.label = Prefer Legacy API
 thing-type.config.neohub.neohub.preferLegacyApi.description = Use the legacy API instead of the new API (if available)
 thing-type.config.neohub.neohub.socketTimeout.label = Socket Timeout
 thing-type.config.neohub.neohub.socketTimeout.description = Time (seconds) to wait for connections to the Hub (min/default=5, max=20)
+thing-type.config.neohub.neohub.apiToken.label = API Access Token
```
jlaur, line 43: Please regenerate this file.

andrewfg: Ok. Done.

bundles/org.openhab.binding.neohub/README.md
```diff
@@ -33,21 +33,31 @@ It signs on to the hub using the supplied connection parameters, and it polls th
...
+| Configuration Parameter    | Description                                                                                              |
+|----------------------------|----------------------------------------------------------------------------------------------------------|
+| hostName                   | Host name (IP address) of the NeoHub (example 192.168.1.123)                                             |
+| useWebSockets<sup>1)</sup> | Use secure WebSockets to connect to  the NeoHub (example `true`)                                         |
+| apiToken<sup>1)</sup>      | API Access Token for secure connection to hub. Create the token in the Heatmiser mobile App              |
+| pollingInterval            | Time (seconds) between polling requests to the NeoHub (Min=4, Max=60, Default=60)                        |
+| socketTimeout              | Time (seconds) to allow for TCP socket connections to the hub to succeed (Min=4, Max=20, Default=5)      |
+| preferLegacyApi            | ADVANCED: Prefer to use older API calls; but if not supported, it switches to new calls (Default=false)  |
```
jlaur, line 43: Forgive me for one more question, perhaps even a stupid one: Can the legacy API be used with web sockets, or are they mutually exclusive? I'm thinking if protocol selection could be something like "Legacy", "Insecure" (?) and "Web Sockets", unless of course all combinations of the two boolean parameters are supported. The other thing which is not clear to me from reading this README - below you describe the old protocol as "older and less secure". That makes me think only about encryption. So just to be clear: The old protocol keeps a TCP connection open and also receives push messages this way? So in terms of functionality they are kind of similar, but the web socket implementation uses HTTPS, thus the major benefit is security by encryption?

bundles/org.openhab.binding.neohub/README.md
```diff
@@ -33,21 +33,31 @@ It signs on to the hub using the supplied connection parameters, and it polls th
...
-| portNumber              | Port number of the NeoHub (Default=4242)                                                    |
-| pollingInterval         | Time (seconds) between polling requests to the NeoHub (Min=4, Max=60, Default=60)           |
-| socketTimeout           | Time (seconds) to allow for TCP socket connections to the hub to succeed (Min=4, Max=20, Default=5) |
-| preferLegacyApi         | ADVANCED: Prefer the binding to use older API calls; if these are not supported, it switches to the new calls (Default=false) |
+| Configuration Parameter    | Description                                                                                              |
+|----------------------------|----------------------------------------------------------------------------------------------------------|
+| hostName                   | Host name (IP address) of the NeoHub (example 192.168.1.123)                                             |
+| useWebSockets<sup>1)</sup> | Use secure WebSockets to connect to  the NeoHub (example `true`)                                         |
```
jlaur, line 39: Here and everywhere else: Shouldn't it be [WebSocket](https://en.wikipedia.org/wiki/WebSocket) (singular) as the name of the protocol?

bundles/org.openhab.binding.neohub/README.md
```diff
@@ -33,21 +33,31 @@ It signs on to the hub using the supplied connection parameters, and it polls th
...
-If this port is disabled the OpenHAB binding cannot connect and the binding will report a *"Connection Refused"* warning in the log.
-In prior firmware versions the port was always enabled.
-But in the new firmware the port is initially enabled on power up but if no communication occurs for 48 hours it is automatically disabled.
-Alternatively the Heatmiser mobile App has a setting (Settings | System | API Access | Legacy API Enable | On) whereby the port can be permanently enabled.
+From early 2022 Heatmiser introduced NeoHub firmware that has the ability to enable / disable connecting to it via a TCP port.
+If the TCP port is disabled the OpenHAB binding cannot connect and the binding will report a *"Connection Refused"* warning in the log.
+In prior firmware versions the TCP port was always enabled.
+But in the new firmware the TCP port is initially enabled on power up but if no communication occurs for 48 hours it is automatically disabled.
+Alternatively the Heatmiser mobile App has a setting (Settings | System | API Access | Legacy API Enable | On) whereby the TCP port can be permanently enabled.
```
jlaur, line 60:
```suggestion
Alternatively the Heatmiser mobile app has a setting (Settings | System | API Access | Legacy API Enable | On) whereby the TCP port can be permanently enabled.
```

bundles/org.openhab.binding.neohub/src/main/java/org/openhab/binding/neohub/internal/NeoHubWebSocket.java
```diff
@@ -0,0 +1,238 @@
...
+        public @Nullable String response;
+    }
+
+    public NeoHubWebSocket(NeoHubConfiguration config) throws NeoHubException {
+        super(config);
+
+        // initialise and start ssl context factory, http client, web socket client
+        SslContextFactory.Client sslContextFactory = new SslContextFactory.Client();
+        sslContextFactory.setTrustAll(true);
+        HttpClient httpClient = new HttpClient(sslContextFactory);
```
jlaur, line 78: Anything preventing the use of the shared framework HTTP client? If having self-singed certificate trust issues, I believe core was prepared for that in 3.3 - see openhab/openhab-core#2622. #11842 does something similar.

bundles/org.openhab.binding.neohub/src/main/java/org/openhab/binding/neohub/internal/NeoHubWebSocket.java
```diff
@@ -0,0 +1,238 @@
...
+        try {
+            httpClient.start();
+        } catch (Exception e) {
+            throw new NeoHubException(String.format("Error starting http client: '%s'", e.getMessage()));
+        }
+        webSocketClient = new WebSocketClient(httpClient);
+        webSocketClient.setConnectTimeout(config.socketTimeout * 1000);
+        try {
+            webSocketClient.start();
+        } catch (Exception e) {
```
jlaur, line 88: Which exceptions can actually occur from this call? You might not want to catch all exceptions here, including runtime exceptions.

bundles/org.openhab.binding.neohub/README.md
andrewfg: See this as background..

https://dev.heatmiser.com/t/api-neohub-v3-0-protocol-8th-june-2022/222?u=andrewfg

The words new/legacy apply (confusingly) to two different things; namely the application layer and the transport layer.
1. Application layer. Depending on the firmware, a hub can support two dialects of JSON request and response payloads. The binding supports both, and it can auto detect and auto select which to use. In the case that the hub supports both dialects, the ‘preferLegacyApi’ config param allows you to decide which dialect shall be used.
2. Transport layer. Depending on the firmware, and a setting in the app, a hub can support two transport layers. Namely a TCP connection and/or a Websocket connection. The TCP connection is plaintext, and neither encrypted nor authenticated. The web socket connection is both encrypted and authenticated via a user generated access key.
3. In principle a hub can support both types of application layer, and both types of transport layer. So four possible permutations.

PS neither of the application layers supports push notifications. Both are query / response models.

bundles/org.openhab.binding.neohub/README.md
andrewfg: Insofar as there is only one session open at any time, I suppose there is only one TCP- or Web-Socket in use at any time. However over time, the binding will create new Xxx- sockets as needed whenever a prior Xxx socket is dropped e.g. due to a communication error.

bundles/org.openhab.binding.neohub/README.md
andrewfg: Ok.

bundles/org.openhab.binding.neohub/src/main/java/org/openhab/binding/neohub/internal/NeoHubWebSocket.java
andrewfg: I did try that initially, but I could not get it to work. But I suppose I was using OH 3.2 so I will have a look again if I could get it working on OH 3.3

bundles/org.openhab.binding.neohub/src/main/java/org/openhab/binding/neohub/internal/NeoHubWebSocket.java
andrewfg: Hmm. For me, any exception is an error that needs to be caught. I cannot imagine why I would handle one type of exception differently than another, or in what way.

andrewfg: PS the source code of the `webSocketClient.start()` method is declared as 'throws Exception' so the binding code must comply with that declaration :)


```
@Override
public final void start() throws Exception {
...
}
```

jlaur: Sorry, you are right - it's completely fine to catch `Exception` for code out of our control (external API).

bundles/org.openhab.binding.neohub/README.md
andrewfg: Ok. Done.

bundles/org.openhab.binding.neohub/README.md
jlaur: Thanks. In that case the current solution seems optimal.

bundles/org.openhab.binding.neohub/src/main/java/org/openhab/binding/neohub/internal/NeoHubWebSocket.java
jlaur: The current solution is perfectly fine, so let's resolve the comment. My intention was to point you towards a possible solution and assess if it could be applied for this case also. You did, and apparently it's not easy. Unfortunately I don't have time/motivation currently to investigate this deeper and come up with alternative code.

bundles/org.openhab.binding.neohub/src/test/java/org/openhab/binding/neohub/test/NeoHubProtocolTests.java
```diff
@@ -0,0 +1,105 @@
...
+            config.apiToken = HUB_API_TOKEN;
+
+            try {
+                NeoHubWebSocket socket = new NeoHubWebSocket(config);
+                String requestJson = NeoHubBindingConstants.CMD_CODE_FIRMWARE;
+                String responseJson = socket.sendMessage(requestJson);
+                assertNotEquals(0, responseJson.length());
+                socket.close();
+            } catch (Exception e) {
+                fail(e.getMessage());
```
jlaur, line 74: If you remove this catch, you can potentially have more information shown as failed test output if any exception is thrown (e.g. stack trace). Checked exceptions can be added to the method signature.

jlaur (approved):
Thanks!

bundles/org.openhab.binding.neohub/src/test/java/org/openhab/binding/neohub/test/NeoHubProtocolTests.java
andrewfg: I am not really sure what you mean by this statement. The test is to confirm that one can connect to a hub via a websocket, send a query, and return a result. So it is a test for one positive; and NOT a test for multiple potential negatives. The catch clause is basically only present to satisfy the compiler. If I would want to test for 'N' multiple potential negatives then I would need to write 'N@ test cases..

jlaur: It was a small tip to simplify your test - I'll post it as a new comment so you can see the diff.

bundles/org.openhab.binding.neohub/src/test/java/org/openhab/binding/neohub/test/NeoHubProtocolTests.java
```diff
@@ -0,0 +1,105 @@
...
+            }
+        }
+    }
```
jlaur, line 77:
```suggestion
    @Test
    void testWssConnection() throws IOException, NeoHubException {
        if (RUN_WSS_TEST) {
            if (!NeoHubJsonTests.VALID_IP_V4_ADDRESS.matcher(HUB_IP_ADDRESS).matches()) {
                fail();
            }

            NeoHubConfiguration config = new NeoHubConfiguration();
            config.hostName = HUB_IP_ADDRESS;
            config.socketTimeout = SOCKET_TIMEOUT;
            config.apiToken = HUB_API_TOKEN;

            NeoHubWebSocket socket = new NeoHubWebSocket(config);
            String requestJson = NeoHubBindingConstants.CMD_CODE_FIRMWARE;
            String responseJson = socket.sendMessage(requestJson);
            assertNotEquals(0, responseJson.length());
            socket.close();
        }
    }
```

bundles/org.openhab.binding.neohub/src/test/java/org/openhab/binding/neohub/test/NeoHubProtocolTests.java
jlaur: The proposed version is not only simpler, it also provides better output in case an exception will be thrown. You can resolve both comments if you prefer not to change anything.

bundles/org.openhab.binding.neohub/src/test/java/org/openhab/binding/neohub/test/NeoHubProtocolTests.java
andrewfg: Ok. Done.

PR #12924, opened 2022-06-11
[workflows] Update actions/setup-maven to 4.3

wrt54g:
# Title
[workflows] Update actions/setup-maven to 4.3

# Description

I updated the setup-maven action to 4.3 (NOTE: I cannot test this update since workflows and actions do not run on forks.)
# Testing
I did not test this since GitHub Workflows and Actions do not run on forked repositories.


wborn:
Thanks for the suggestion! :+1:  There is even 4.4 now. :slightly_smiling_face: 
I've added it together with some other updates/upgrades to #13018.

Issue #12928, opened 2022-06-12
[hdpowerview] Repeater usage information


jlaur:
#12061 introduced support for repeaters.

For a week one of my shades did not react to automations. I finally found out that a family member had turned off the repeater by accident.

Unfortunately the Hub does not report if repeaters are reachable, so it is not possible to create a rule simply detecting status change to **OFFLINE**.

However, survey data is available. Example:

**Request URL:** http://powerview/api/shades/36137?survey=true
```
{
    "shade_id": 36137,
    "survey": [
        {
            "neighbor_id": 17805,
            "rssi": -81
        }
    ]
}
```

This indicates that shade 36137 is connected through repeater 17805. This could be exposed as a `Switch` channel for shade things. This way it would be possible to detect if a specific shade is connected through a repeater. However, we also need to be aware that response from time to time will be:
```
{
    "shade_id": 36137,
    "survey": []
}
```

this could be mapped to **UNDEF**.

A rule could then detect if channel has been != **ON** for a certain amount of time.

Shade not connected through a repeater:

```
{
    "shade_id": 46532,
    "survey": [
        {
            "neighbor_id": 0,
            "rssi": -64
        }
    ]
}
```

Also seen:

```
{
    "shade_id": 58120,
    "survey": [
        {
            "neighbor_id": 17805,
            "rssi": -89
        },
        {
            "neighbor_id": 0,
            "rssi": -71
        }
    ]
}
```

Not sure what this means, but at least is has connection to repeater, even if this is not used. Probably it will use the "neighbor" with highest [RSSI](https://en.wikipedia.org/wiki/Received_signal_strength_indication).

Additionally a new RSSI channel should be introduced.

PR #11944, opened 2022-01-02
[hdpowerview] Support refresh command for signal strength channel

jlaur:
Fixes #11942

Signed-off-by: Jacob Laursen <jacob-github@vindvejr.dk>

Channel **signalStrength** was added by #11198 while #11933 refactored the _REFRESH_ command for shades to be relative to channel for requested item. The missing piece now is the **signalStrength** channel which currently cannot be refreshed on request.

This pull request makes the _REFRESH_ command for shades feature complete as each channel now has a dedicated method assigned for requesting a hard refresh for the item context.

Tested with both negative result:
```
2022-01-03 00:52:58.796 [TRACE] [rview.internal.HDPowerViewWebTargets] - JSON response = null
2022-01-03 00:52:58.800 [WARN ] [rnal.handler.HDPowerViewShadeHandler] - No response from shade 36321 survey
```

and positive result:
```
2022-01-03 00:54:21.764 [INFO ] [openhab.event.ItemCommandEvent      ] - Item 'Blind11_SignalStrength' received command REFRESH
2022-01-03 00:54:40.781 [TRACE] [rview.internal.HDPowerViewWebTargets] - JSON response = {"shade_id":36321,"survey":[{"neighbor_id":17805,"rssi":-86}]}
2022-01-03 00:54:40.783 [DEBUG] [rnal.handler.HDPowerViewShadeHandler] - Survey response for shade 36321: {neighbor:17805, rssi:-86}
2022-01-03 00:54:40.947 [DEBUG] [ternal.handler.HDPowerViewHubHandler] - Updating shade '36321'
2022-01-03 00:54:40.964 [INFO ] [openhab.event.ItemStateChangedEvent ] - Item 'Blind11_SignalStrength' changed from 0 to 4
```

PR #12930, opened 2022-06-12
[hdpowerview] Fix signalStrength channel update after REFRESH command

jlaur:
While preparing for #12928 I encountered an issue with refreshing signal strength (introduced in #11944). Signal strength is refreshed in the Hub by this request:

```
http://<hub>/api/shades/<shadeId>?survey=true
```

However, the response is a survey which does not include **signalStrength** itself. Therefore an additional request must be made to fetch the new value:

```
http://<hub>/api/shades/<shadeId>
```

This pull request fixes this issue so that new value is obtained immediately instead of waiting for next poll (up to one minute).


lolodomo (approved):
LGTM

PR #12931, opened 2022-06-12
[homekit] fix issue with some hue lights, e.g. xiaomi

yfre:
As reported here, some hue light expect all 3 value (hue, saturation and brightness) even if only one value got update
see investigation here
https://community.openhab.org/t/how-to-configure-homekit-light-with-colour/136137/33?u=yfre
this was already the case for hue and saturation changes but not for brightness. 

this fix ensure that all 3 values are always sent for item of type color. 

### DIFF
--- a/bundles/org.openhab.io.homekit/src/main/java/org/openhab/io/homekit/internal/HomekitOHItemProxy.java
+++ b/bundles/org.openhab.io.homekit/src/main/java/org/openhab/io/homekit/internal/HomekitOHItemProxy.java
@@ -103,28 +103,37 @@ public class HomekitOHItemProxy {
         // if hue or saturation present, send an HSBType state update. no filter applied for HUE & Saturation
         if ((hue != null) || (saturation != null)) {
             if (item instanceof ColorItem) {
-                // logic for ColorItem = combine hue, saturation and brightness update to one command
-                final HSBType currentState = item.getState() instanceof UnDefType ? HSBType.BLACK
-                        : (HSBType) item.getState();
-                ((ColorItem) item).send(new HSBType(hue != null ? hue : currentState.getHue(),
-                        saturation != null ? saturation : currentState.getSaturation(),
-                        brightness != null ? brightness : currentState.getBrightness()));
-                logger.trace("send HSB command for item {} with following values hue={} saturation={} brightness={}",
-                        item, hue, saturation, brightness);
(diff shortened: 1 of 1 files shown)


andylintner:
Isn't this something to fix in the Hue binding rather than workaround here? It should be valid in OpenHAB to change brightness without updating hue/saturation...

yfre:
> Isn't this something to fix in the Hue binding rather than workaround here? It should be valid in OpenHAB to change brightness without updating hue/saturation...

agree, ideally it should be fix in the xiaomi binding. but it looks like openhab main ui sends also all 3 values all time.  so, we mimic ui behaviour here and it is an easy fix. 

i could make it configurable, just in case we have devices which needs only brightness.

bundles/org.openhab.io.homekit/src/main/java/org/openhab/io/homekit/internal/HomekitOHItemProxy.java
```diff
@@ -103,30 +103,39 @@ private void sendCommand() {
...
         commandCache.clear();
     }
 
+    private void sendHSBCommand(ColorItem item, @Nullable DecimalType hue, @Nullable PercentType saturation,
+            @Nullable PercentType brightness) {
+        final HSBType currentState = item.getState() instanceof UnDefType ? HSBType.BLACK : (HSBType) item.getState();
+        // logic for ColorItem = combine hue, saturation and brightness update to one command
+        final DecimalType target_hue = hue != null ? hue : currentState.getHue();
+        final PercentType target_saturation = saturation != null ? saturation : currentState.getSaturation();
+        final PercentType target_brightness = brightness != null ? brightness : currentState.getBrightness();
```
lolodomo, line 133: Please rename your variables to follow convention, `targetHue` for example.

lolodomo (changes requested):

bundles/org.openhab.io.homekit/src/main/java/org/openhab/io/homekit/internal/HomekitOHItemProxy.java
yfre: valid point. fixed

andylintner:
> it looks like openhab main ui sends also all 3 values all time.

Makes sense to do the same thing as the UI then.

lolodomo (approved):
LGTM

PR #12934, opened 2022-06-13
[Senechome] add state battery diagnosis

vich-667:
# Senechome: add state battery diagnosis
- [senechome][SenecSystemStatus] adding missins state battery diagnosis


jlaur (approved):
Thanks!

PR #12935, opened 2022-06-13
[influxdb] Partial ModifiablePersistenceService support

seime:
Signed-off-by: Arne Seime <arne.seime@gmail.com>

Partial support for persisting values with a specific timestamp by implementing `ModifiablePersistenceService` instead of `QueryablePersistenceService`.

By also changing the startlevel to 79 (before bindings), it allows the persistence service to be called from bindings to persist previous (ie offline sensor time series being reported when sensor is offline) or future time series data such as electricity prices or weather forecasts. 

A better approach would be to build this into the framework, but this seems to be a larger and quite more complex job unfortunately. See https://github.com/openhab/openhab-core/issues/844

EDIT: Someone just created a PR to handle this properly in core; https://github.com/openhab/openhab-core/pull/3000

### DIFF
--- a/bundles/org.openhab.persistence.influxdb/src/main/feature/feature.xml
+++ b/bundles/org.openhab.persistence.influxdb/src/main/feature/feature.xml
@@ -4,8 +4,9 @@
 
 	<feature name="openhab-persistence-influxdb" description="InfluxDB Persistence" version="${project.version}">
 		<feature>openhab-runtime-base</feature>
-		<bundle start-level="80">mvn:org.openhab.addons.bundles/org.openhab.persistence.influxdb/${project.version}</bundle>
-		<configfile finalname="${openhab.conf}/services/influxdb.cfg" override="false">mvn:org.openhab.addons.features.karaf/org.openhab.addons.features.karaf.openhab-addons-external/${project.version}/cfg/influxdb</configfile>
+		<bundle start-level="79">mvn:org.openhab.addons.bundles/org.openhab.persistence.influxdb/${project.version}</bundle>
+		<configfile finalname="${openhab.conf}/services/influxdb.cfg" override="false">mvn:org.openhab.addons.features.karaf/org.openhab.addons.features.karaf.openhab-addons-external/${project.version}/cfg/influxdb
+		</configfile>
 	</feature>
 
 </features>
--- a/bundles/org.openhab.persistence.influxdb/src/test/java/org/openhab/persistence/influxdb/internal/ItemToStorePointCreatorTest.java
+++ b/bundles/org.openhab.persistence.influxdb/src/test/java/org/openhab/persistence/influxdb/internal/ItemToStorePointCreatorTest.java
@@ -13,11 +13,7 @@
 package org.openhab.persistence.influxdb.internal;
 
 import static org.hamcrest.MatcherAssert.assertThat;
-import static org.hamcrest.Matchers.equalTo;
-import static org.hamcrest.Matchers.hasEntry;
-import static org.hamcrest.Matchers.hasKey;
-import static org.hamcrest.Matchers.is;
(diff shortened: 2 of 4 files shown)


bundles/org.openhab.persistence.influxdb/src/main/feature/feature.xml
```diff
@@ -4,8 +4,9 @@
 
 	<feature name="openhab-persistence-influxdb" description="InfluxDB Persistence" version="${project.version}">
 		<feature>openhab-runtime-base</feature>
-		<bundle start-level="80">mvn:org.openhab.addons.bundles/org.openhab.persistence.influxdb/${project.version}</bundle>
-		<configfile finalname="${openhab.conf}/services/influxdb.cfg" override="false">mvn:org.openhab.addons.features.karaf/org.openhab.addons.features.karaf.openhab-addons-external/${project.version}/cfg/influxdb</configfile>
+		<bundle start-level="79">mvn:org.openhab.addons.bundles/org.openhab.persistence.influxdb/${project.version}</bundle>
+		<configfile finalname="${openhab.conf}/services/influxdb.cfg" override="false">mvn:org.openhab.addons.features.karaf/org.openhab.addons.features.karaf.openhab-addons-external/${project.version}/cfg/influxdb
+		</configfile>
```
wborn, line 9: What is the reason for these changes?

bundles/org.openhab.persistence.influxdb/src/test/java/org/openhab/persistence/influxdb/internal/ItemToStorePointCreatorTest.java
```diff
@@ -13,7 +13,11 @@
 package org.openhab.persistence.influxdb.internal;
 
 import static org.hamcrest.MatcherAssert.assertThat;
-import static org.hamcrest.Matchers.*;
+import static org.hamcrest.Matchers.equalTo;
+import static org.hamcrest.Matchers.hasEntry;
+import static org.hamcrest.Matchers.hasKey;
+import static org.hamcrest.Matchers.is;
+import static org.hamcrest.Matchers.not;
```
wborn, line 20: If you don't add any tests I suggest you leave the file alone. :wink:

bundles/org.openhab.persistence.influxdb/src/main/feature/feature.xml
seime: It was to start the persistence service before the bindings so they could make use of it. But the same night another pr with changes to core appeared rendering this change obsolete (when/if accepted)

bundles/org.openhab.persistence.influxdb/src/test/java/org/openhab/persistence/influxdb/internal/ItemToStorePointCreatorTest.java
seime: Sorry, intellij changes. Should have been removed

lolodomo:
@seime : you answered to comments but you pushed no change ?

seime:
I was hoping this PR would be rendered obsolete by https://github.com/openhab/openhab-core/pull/3000, but that one appears to be stuck :(

jlaur:
@seime - would you like to pursue this PR to have a solution until core will be enhanced in this regard?

bundles/org.openhab.persistence.influxdb/src/main/feature/feature.xml
seime: Should in fact all persistence bundles start _before_ level 80?

lsiepel: Just checked, without exception all persistence bundles start at level 80.

Issue #7573, opened 2020-05-08
Replace gnu.io dependency of add-ons with serial transport


wborn:
There are still add-ons which depend on gnu.io for their serial communications. By using the serial transport we can switch between serial implementations (rxtx, javacomm) without having to change any add-on code.

These bundles still depend on gnu.io:

* [ ] org.openhab.binding.cm11a
* [ ] org.openhab.binding.digiplex
* [ ] org.openhab.binding.dscalarm
* [ ] org.openhab.binding.elerotransmitterstick
* [ ] org.openhab.binding.jeelink
* [ ] org.openhab.binding.knx
* [ ] org.openhab.binding.lgtvserial
* [ ] org.openhab.binding.lutron
* [ ] org.openhab.binding.meteostick
* [ ] org.openhab.binding.nibeheatpump
* [ ] org.openhab.binding.oceanic
* [ ] org.openhab.binding.pentair
* [ ] org.openhab.binding.phc
* [ ] org.openhab.binding.pioneeravr
* [ ] org.openhab.binding.regoheatpump
* [ ] org.openhab.binding.rme
* [ ] org.openhab.binding.rotelra1x
* [ ] org.openhab.binding.smartmeter
* [ ] org.openhab.binding.urtsi
* [ ] org.openhab.binding.velbus
* [ ] org.openhab.io.transport.modbus

lolodomo:
The rotelra1x binding is supposed to be removed in OH 3.0 as it was replaced by the rotel binding. We kept it in 2.5 to let the time to users to move to the new binding.

wborn:
Great! I've updated the list so we won't spend any time on migrating that one. :+1:

gnlpfjh:
With #6447 phc binding uses serial transport.

wborn:
Very nice! I even see you updated the POM for the import that was added in https://github.com/openhab/openhab-addons/pull/7589. :+1:

mvalla:
What is to be done in case of a transitive dependency from a library that uses nrjavaserial (like in the case of KNX) ?

bodiroga:
This should be pinned again, I think that @jsjames has unpinned it unintentionally :wink:

holgerfriedrich:
[...]

PR #12942, opened 2022-06-16
[knx] FT12: List available serial ports if port cannot be opened

holgerfriedrich:
Fixes a regression introduced by upgrading Calimero library to 2.5.

### DIFF
--- a/bundles/org.openhab.binding.knx/src/main/java/org/openhab/binding/knx/internal/client/SerialClient.java
+++ b/bundles/org.openhab.binding.knx/src/main/java/org/openhab/binding/knx/internal/client/SerialClient.java
@@ -36,6 +36,8 @@ import tuwien.auto.calimero.link.medium.TPSettings;
 @NonNullByDefault
 public class SerialClient extends AbstractKNXClient {
 
+    private static final String CALIMERO_ERROR_CANNOT_OPEN_PORT = "failed to open serial port";
+
     private final Logger logger = LoggerFactory.getLogger(SerialClient.class);
 
     private final String serialPort;
@@ -69,7 +71,8 @@ public class SerialClient extends AbstractKNXClient {
                     e);
         } catch (KNXException e) {
             final String msg = e.getMessage();
-            if ((msg != null) && (msg.startsWith("can not open serial port"))) {
+            // TODO add a test for this string match; error message might change in later version of Calimero library
+            if ((msg != null) && (msg.startsWith(CALIMERO_ERROR_CANNOT_OPEN_PORT))) {
                 StringBuilder sb = new StringBuilder("Available ports are:\n");
                 Enumeration<?> portList = CommPortIdentifier.getPortIdentifiers();
                 while (portList.hasMoreElements()) {


bundles/org.openhab.binding.knx/src/main/java/org/openhab/binding/knx/internal/client/SerialClient.java
```diff
@@ -69,7 +69,8 @@ protected KNXNetworkLink establishConnection() throws KNXException, InterruptedE
                     e);
         } catch (KNXException e) {
             final String msg = e.getMessage();
-            if ((msg != null) && (msg.startsWith("can not open serial port"))) {
+            // TODO add a test for this string match; error message might change in later version of Calimero library
+            if ((msg != null) && (msg.startsWith("failed to open serial port"))) {
```
lolodomo, line 73: You could at least define a constant?

lolodomo (changes requested):

bundles/org.openhab.binding.knx/src/main/java/org/openhab/binding/knx/internal/client/SerialClient.java
holgerfriedrich: Done as requested. This was not my code. I tried to keep the change minimal and just adapted to the new message.

bundles/org.openhab.binding.knx/src/main/java/org/openhab/binding/knx/internal/client/SerialClient.java
```diff
@@ -36,6 +36,8 @@
 @NonNullByDefault
 public class SerialClient extends AbstractKNXClient {
 
+    private final String CALIMERO_ERROR_CANNOT_OPEN_PORT = "failed to open serial port";
```
lolodomo, line 39: Sorry but `private static final` is expected.

lolodomo (changes requested):

holgerfriedrich:
@lolodomo valid comment.... done

bundles/org.openhab.binding.knx/src/main/java/org/openhab/binding/knx/internal/client/SerialClient.java
holgerfriedrich: done

lolodomo (approved):
LGTM

bundles/org.openhab.binding.knx/src/main/java/org/openhab/binding/knx/internal/client/SerialClient.java
```diff
@@ -69,7 +71,8 @@ protected KNXNetworkLink establishConnection() throws KNXException, InterruptedE
                     e);
         } catch (KNXException e) {
             final String msg = e.getMessage();
-            if ((msg != null) && (msg.startsWith("can not open serial port"))) {
+            // TODO add a test for this string match; error message might change in later version of Calimero library
+            if ((msg != null) && (msg.startsWith(CALIMERO_ERROR_CANNOT_OPEN_PORT))) {
```
jlaur, line 75: Just curious: Can you point to the code throwing this exception? What is the actual exception - raw `KNXException` or a subclass? Does it have cause within, e.g. `IOException`? If there is really not enough information to handle it appropriately besides using the message, perhaps it would be worth filing an issue (or pull request) towards the Calimero repository?

holgerfriedrich: @jlaur  I agree, this is not the perfect way to do it (but adapting the text is a quick solution for 3.3 which at least makes the feature work again).
https://github.com/calimero-project/calimero-core/blob/0442d65f39ce0a2c39ce8d4674465de62bee6d11/src/tuwien/auto/calimero/serial/LibraryAdapter.java#L168
Maybe it is; the handling of serial devices in the KNX needs to be changed anyway and due to usage of gnu.io library (#7573) and because there have been changes on the Calimero main branch in the area serial communication.

jlaur: If the exception is thrown by that line, it includes the original exception, which can be obtained by `e.getCause()`. Perhaps the original exception provides more useful information?

holgerfriedrich: Yes, good point. But I am still not convinced that this helps much in this case, as it is depending on underlying com stack implementation. For me it is ```gnu.io.NoSuchPortException```, but it could also be ```gnu.io.PortInUseException``` or ```gnu.io.UnsupportedCommOperationException```.
Borderline is: port cannot be opened. This is what the initial implementation tried to detect via the string comparison. It was broken by the update to Calimero 2.5 as the text changed.

jlaur: OK, fair enough. Fix is fine for now. If we want a more robust solution, an issue for throwing a dedicated exception (subclass of `KNXException`) could be created for the Calimero project.

PR #12946, opened 2022-06-16
[transform.javascript] Update examples in README

dilyanpalauzov:
According to my experience, after JS there must be a colon, not a parenthesis.  The `transform/` path must be skipped.

### DIFF
--- a/bundles/org.openhab.transform.javascript/README.md
+++ b/bundles/org.openhab.transform.javascript/README.md
@@ -43,8 +43,8 @@ Following example will return value `23.54` when `input` data is `214`.
 
 ### Inline script example:
 
-Normally JavaScript transformation is given by filename, e.g. `JS(transform/getValue.js)`.
-Inline script can be given by `|` character following the JavaScript, e.g. `JS(| input / 10)`.
+Normally JavaScript transformation is given by filename, e.g. `JS:getValue.js`.
+Inline script can be given by `|` character following the JavaScript, e.g. `JS:| input / 10`.
    
 ## Test JavaScript
 


lolodomo:
Anyone to validate the proposed updated syntax ?

dilyanpalauzov:
Update: the original syntax might still be correct, depending on where it is used.

When the transforms are written in the UI as channel-transformation, there must be a colon after the protocol (`JS:`, `MAP:`).  This is repeated at https://www.openhab.org/docs/configuration/things.html#defining-bridges-using-files (see `transformationPattern="JSONPATH:$.StatusFWR.Version"`).

But in sitemap files in `label` the syntax is `label="zzz [MAP(....):%s]"` .  Thus there are brackets after the transform module.

That said, the documentation for transforms shall state when to use colon after the transform and when brackets.

Or, even better, the syntax should be unified.

jlaur:
@florian-h05 - could you have a look and possibly comment?

florian-h05:
I'm not sure how to proceed with the docs of JavaScript transformation since it has been replaced by the [`SCRIPT` transformation](https://next.openhab.org/docs/configuration/transformations.html#script-transformation).

Note that the JS transformation addon has been removed, so if one wants to update the docs, the `main` branch is the wrong place.

Anyway, I had a look at the syntax:

- Item labels:
  - in `.items` & `.sitemap` files: I can remember to always have used format `JS(script.js):%s` (note that e.g. `JS(script.js):%d` also works) (I switched to the SCRIPT transformation)
  - in the UI (`stateDescription` metadata -> profile): same syntax as in files
  - in both cases, you can neither leave out the script name in the brackets nor the `:` together with the state format (means **the changes from this PR don't work**)
- Channel (link) transformations:
  - in `.items` files: `channel="..." [profile="transform:JS" function="script.js"]`

dilyanpalauzov:
I have forgotten all the details in the mean time and currently on my system do transforms only in `.things` files and I do not use JS-transforms anymore.  I believe at the time I wrote the comments, they were correct.  If it is a JavaScript or a MAP transform does not make a difference.  If you think the changes shall not be applied, I am absolutely fine with this.

jlaur:
@florian-h05 - thanks for the explanation. I had not realized this was replaced. So since the transformation doesn't exist anymore, this PR is obsoleted, so let's close it. @dilyanpalauzov - thanks for the effort to describe this though.

PR #12947, opened 2022-06-16
[gardena] Fix compiler warnings

### LINKED ISSUE #12898: [gardena] Build system issues compiler warnings
As mentioned in #11825 the build system issues some compiler warnings.
I intend to modify the binding code to fix those warnings and create a new PR.
This issue is a placeholder to track those fixes.

andrewfg:
Fixes #12898

### DIFF
--- a/bundles/org.openhab.binding.gardena/README.md
+++ b/bundles/org.openhab.binding.gardena/README.md
@@ -108,6 +108,7 @@ openhab:send WC_Valve_cmd_CloseValve.sendCommand(ON) // stop any active watering
 ```
 
 If you send a REFRESH command to the last update timestamp (no matter which thing), **ALL** items from **ALL** things are updated
+
 ```
 DateTime LastUpdate "LastUpdate [%1$td.%1$tm.%1$tY %1$tH:%1$tM]" { channel="gardena:water_control:home:myWateringComputer:common#lastUpdate_timestamp" }
 
--- a/bundles/org.openhab.binding.gardena/src/main/java/org/openhab/binding/gardena/internal/model/dto/Device.java
+++ b/bundles/org.openhab.binding.gardena/src/main/java/org/openhab/binding/gardena/internal/model/dto/Device.java
@@ -126,8 +126,7 @@ public class Device {
             valveSet = (ValveSetServiceDataItem) dataItem;
         } else if (dataItem instanceof ValveServiceDataItem) {
             String valveNumber = StringUtils.substringAfterLast(dataItem.id, ":");
-            if (valveNumber != null
-                    && (valveNumber.equals("") || valveNumber.equals("wc") || valveNumber.equals("0"))) {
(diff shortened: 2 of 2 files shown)


jlaur:
@andrewfg - I see only one compiler error which hasn't been fixed by this PR?

`Warning:  /home/runner/work/openhab-addons/openhab-addons/bundles/org.openhab.binding.gardena/src/main/java/org/openhab/binding/gardena/internal/model/dto/api/DataItem.java:[31,14] The field attributes may not have been initialized, whereas its type 'T' is a free type variable that may represent a '@NonNull' type`

Perhaps you can add some details as there seem to be a mismatch between PR title and actual result.

jlaur:
@andrewfg - I see only one compiler error which hasn't been fixed by this PR?

`Warning:  /home/runner/work/openhab-addons/openhab-addons/bundles/org.openhab.binding.gardena/src/main/java/org/openhab/binding/gardena/internal/model/dto/api/DataItem.java:[31,14] The field attributes may not have been initialized, whereas its type 'T' is a free type variable that may represent a '@NonNull' type`

Perhaps you can add some details as there seem to be a mismatch between PR title and actual result.

andrewfg:
> see only one compiler warning which hasn't been fixed by this PR

@jlaur that is correct. I would have fixed that too, but I don't know how. (It concerns a nullable furld in a dto, so it should already be implicitly nullable and nulled..)

jlaur:
> @jlaur that is correct. I would have fixed that too, but I don't know how. (It concerns a nullable furld in a dto, so it should already be implicitly nullable and nulled..)

OK, so now I follow. I missed that the lines you changed actually triggered some SAT warnings. I was looking for compiler warnings and only found the one that wasn't fixed. Will perform review now - less confused. :-)

jlaur (approved):
Thanks.

andrewfg:
> Warning: /home/runner/work/openhab-addons/openhab-addons/bundles/org.openhab.binding.gardena/src/main/java/org/openhab/binding/gardena/internal/model/dto/api/DataItem.java:[31,14] The field attributes may not have been initialized, whereas its type 'T' is a free type variable that may represent a '@NonNull' type

> It concerns a nullable **_generic_** field in a dto, so it should already be implicitly nullable and nulled..)

@jlaur your own PR today with an interesting null annotation made me think more about this case: The binding code is definitely an edge case where null annotations and null checking falls over, so I would not change the code in any way. However I could eliminate the compiler warning by adding a `@SuppressWarnings("null")` to it. It is probably a trivial issue but what do you suggest? Personally I always aim for a totally clean build if I can..

jlaur:
@andrewfg - it's a tricky case. Type _T_ can either be nullable or non-null. If we explicitly annotate it as nullable, we get into null constraint mismatch troubles. But actually it would make sense to annotate it as non-null. Now we have another problem, since `attributes` is not initialized by constructor (but later by deserialization).

I have created #12957 as a proposed solution, where `attributes` is annotated as nullable and corresponding null-checks have been made.

PR #12953, opened 2022-06-17
Append the public holiday plugin

Hinterwaeldlers:
Instead of doing the calculation, if the current day (or the day after) is a public holiday, within a rule, this addon will do this and provide the result as channels.

As result, related rules can be simplified and the need to execute the calculation, if the current day is a public holiday is removed.

The user can specify which of the holidays are relevant by name and assign the channels to an item.

Draw back: Missing public holidays have to be added within the code, but only requires a few lines to be supported.

Further information of the usage are given within the readme file: https://github.com/Hinterwaeldlers/openhab-addons/blob/publicHoliday/bundles/org.openhab.binding.publicholiday/README.md

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -257,6 +257,7 @@
 /bundles/org.openhab.binding.powermax/ @lolodomo
 /bundles/org.openhab.binding.proteusecometer/ @2chilled
 /bundles/org.openhab.binding.prowl/ @octa22
+/bundles/org.openhab.binding.publicholiday/ @Hinterwaeldlers
 /bundles/org.openhab.binding.publictransportswitzerland/ @jeremystucki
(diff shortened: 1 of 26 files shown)


fwolter:
Are you aware of Ephemeris? https://www.openhab.org/docs/configuration/actions.html#ephemeris

It covers all local holidays in Germany (e.g. Reformationstag only in Niedersachsen) and other countries and it's very easy to use in rules, also with offset.

grafik

fwolter:
I'm afraid this won't have a chance to get merged as the functionality is already present in openhab-core. Feel free to comment if you have any questions.

Hinterwaeldlers:
I'm wasn't aware of this functionality and haven't step over it while searching around. So it's fine to me to close it. thanks for pointing out to the relevant documentation

PR #12955, opened 2022-06-18
[dali] Implement DT8 (single-channel RGB & color temperature) device type

s3lph:
# [dali] Add support for DT8 (single-channel RGB & color temperature) devices.

# Description

This PR adds support for DALI DT8 devices. These are devices supporting customizable color temperature and/or single-channel (as opposed to the 3-channel solution already implemented) RGB color controls.
This PR adds two new thing-types:
- `dali.device-dt8`
- `dali.group-dt8`
These are largely identical to the already existing `dali.device` and `dali.group`, with the addition of two new channels, one for controlling RGB color, and another for controlling color temperature.

This PR also fixes some issues in DALI device addressing:
- Group addressing was broken, effectively only group 0 could be addressed.
- Uppermost bit of short (6bit) addresses was removed, so effectively only devices 0-31 could be addressed.
- Reading state from group addresses does not work in general, because multiple devices attempt to respond at once. It basically only works if all devices in the group are returning the same value, and even then it can fail, e.g. when different devices with slightly different timing characterstics are part of the same group. As a workaround, I added a config option to DALI groups which allows OpenHAB to read a single device's state and treat it as the entire group's state.

### DIFF
--- a/bundles/org.openhab.binding.dali/src/main/java/org/openhab/binding/dali/internal/handler/DaliDt8DeviceHandler.java
+++ b/bundles/org.openhab.binding.dali/src/main/java/org/openhab/binding/dali/internal/handler/DaliDt8DeviceHandler.java
@@ -181,7 +181,6 @@ public class DaliDt8DeviceHandler extends DaliDeviceHandler {
                 super.handleCommand(channelUID, command);
             }
         } catch (DaliException e) {
-            logger.error("Dali Exception", e);
             updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.COMMUNICATION_ERROR, e.getMessage());
         }
     }
--- a/bundles/org.openhab.binding.dali/src/main/java/org/openhab/binding/dali/internal/handler/DaliDt8DeviceHandler.java
+++ b/bundles/org.openhab.binding.dali/src/main/java/org/openhab/binding/dali/internal/handler/DaliDt8DeviceHandler.java
@@ -1,0 +1,187 @@
+/**
+ * Copyright (c) 2010-2022 Contributors to the openHAB project
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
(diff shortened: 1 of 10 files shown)


rs22 (approved):
Thanks for this addition and the fixes to the codebase! 

I haven't had the chance to test with my own DALI installation, but the code changes look good to me! 👍

s3lph:
@rs22 only fixed the DCO, had to force push and override my commits for this.

bundles/org.openhab.binding.dali/src/main/java/org/openhab/binding/dali/internal/handler/DaliDt8DeviceHandler.java
```diff
@@ -0,0 +1,187 @@
...
+                }).exceptionally(e -> {
+                    logger.warn("Error querying device status: {}", e.getMessage());
+                    return null;
+                });
+
+            } else {
+                super.handleCommand(channelUID, command);
+            }
+        } catch (DaliException e) {
+            logger.error("Dali Exception", e);
```
jlaur, line 183: Will this make the binding stop working? Otherwise log level should be reduced. Perhaps it can even be removed as the next line will also be logged by the framework. For reference: https://www.openhab.org/docs/developer/guidelines.html#f-logging

bundles/org.openhab.binding.dali/src/main/java/org/openhab/binding/dali/internal/handler/DaliserverBridgeHandler.java
```diff
@@ -142,6 +142,7 @@ public void dispose() {
                 } catch (Exception e) {
                     logger.warn("Unexpected exception while sending command to daliserver: {} Message: {}", frame,
                             e.getMessage());
+                    logger.error("Stacktrace", e);
```
jlaur, line 145:
```suggestion
                    logger.trace("Stacktrace", e);
```

jlaur (changes requested):
Thanks, looks really good. Only some minor comments.

bundles/org.openhab.binding.dali/src/main/java/org/openhab/binding/dali/internal/handler/DaliDt8DeviceHandler.java
s3lph: Removed.

jlaur (approved):
Thank you!

PR #12956, opened 2022-06-18
[globalcache] Fix NPE when reading reply from device

mhilbush:
If network issues occur, it's possible for `readLine` to return null, which causes a NPE and kills the command processor job. This PR prevents that from occurring.

```
2022-05-19 17:57:27.650 [WARN ] [mmon.WrappedScheduledExecutorService] - Scheduled runnable ended with an exception:
        at org.openhab.binding.globalcache.internal.handler.GlobalCacheHandler$CommandProcessor.readReplyFromDevice(GlobalCacheHandler.java:564) ~[?:?]
        at org.openhab.binding.globalcache.internal.handler.GlobalCacheHandler$CommandProcessor.run(GlobalCacheHandler.java:502) ~[?:?]
```

### DIFF
--- a/bundles/org.openhab.binding.globalcache/src/main/java/org/openhab/binding/globalcache/internal/handler/GlobalCacheHandler.java
+++ b/bundles/org.openhab.binding.globalcache/src/main/java/org/openhab/binding/globalcache/internal/handler/GlobalCacheHandler.java
@@ -130,11 +130,6 @@ public class GlobalCacheHandler extends BaseThingHandler {
 
     @Override
     public void handleCommand(ChannelUID channelUID, Command command) {
-        if (command == null) {
-            logger.warn("Command passed to handler for thing {} is null", thingID());
-            return;
-        }
-
         // Don't try to send command if the device is not online
         if (!isOnline()) {
             logger.debug("Can't handle command {} because handler for thing {} is not ONLINE", command, thingID());
@@ -538,12 +533,10 @@ public class GlobalCacheHandler extends BaseThingHandler {
          */
         private void writeCommandToDevice(RequestMessage requestMessage) throws IOException {
             logger.trace("Processor for thing {} writing command to device", thingID());
-
             if (connectionManager.getCommandOut() == null) {
                 logger.debug("Error writing to device because output stream object is null");
                 return;
             }
-
             byte[] deviceCommand = (requestMessage.getDeviceCommand() + '\r').getBytes();
             connectionManager.getCommandOut().write(deviceCommand);
             connectionManager.getCommandOut().flush();
@@ -554,14 +547,16 @@ public class GlobalCacheHandler extends BaseThingHandler {
          */
         private String readReplyFromDevice(RequestMessage requestMessage) throws IOException {
             logger.trace("Processor for thing {} reading reply from device", thingID());
-
             if (connectionManager.getCommandIn() == null) {
                 logger.debug("Error reading from device because input stream object is null");
                 return "ERROR: BufferedReader is null!";
             }
-
-            logger.trace("Processor for thing {} reading response from device", thingID());
-            return connectionManager.getCommandIn().readLine().trim();
+            String reply = connectionManager.getCommandIn().readLine();
+            if (reply == null) {
+                logger.debug("Read of reply from device returned null!");
+                return "ERROR: reply is null!";
+            }
+            return reply.trim();
         }
 
         /*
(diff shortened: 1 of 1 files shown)


bundles/org.openhab.binding.globalcache/src/main/java/org/openhab/binding/globalcache/internal/handler/GlobalCacheHandler.java
```diff
@@ -554,14 +547,16 @@ private void writeCommandToDevice(RequestMessage requestMessage) throws IOExcept
...
                 logger.debug("Error reading from device because input stream object is null");
                 return "ERROR: BufferedReader is null!";
             }
-
-            logger.trace("Processor for thing {} reading response from device", thingID());
-            return connectionManager.getCommandIn().readLine().trim();
+            String reply = connectionManager.getCommandIn().readLine();
+            if (reply == null) {
+                logger.debug("Read of reply from device returned null!");
+                return "ERROR: reply is null!";
```
jlaur, line 557: It seems that it would be more appropriate to throw `IOException` here, so the error would be handled correctly. Now the method returns successfully with reply "ERROR: reply is null!", leaving the caller no options for handling the situation gracefully.

jlaur (changes requested):
Thanks. The `NullPointerException` is definitely handled, but I have posted a comment about how it could possibly be handled more appropriately in the flow following that.

bundles/org.openhab.binding.globalcache/src/main/java/org/openhab/binding/globalcache/internal/handler/GlobalCacheHandler.java
```diff
@@ -554,14 +547,16 @@ private void writeCommandToDevice(RequestMessage requestMessage) throws IOExcept
          */
         private String readReplyFromDevice(RequestMessage requestMessage) throws IOException {
             logger.trace("Processor for thing {} reading reply from device", thingID());
-
             if (connectionManager.getCommandIn() == null) {
                 logger.debug("Error reading from device because input stream object is null");
                 return "ERROR: BufferedReader is null!";
```
jlaur, line 552: I see now that this kind of error handling was also used here, so consider if this should also be turned into something like:
`throw new IOException("BufferedReader is null");`

bundles/org.openhab.binding.globalcache/src/main/java/org/openhab/binding/globalcache/internal/handler/GlobalCacheHandler.java
mhilbush: While I agree it would make sense to do that, I was trying to follow the model that I used way back when I initially wrote the binding (it was the first binding that I wrote and I definitely would do some things differently knowing what I know now about openHAB and Java).

When the GlobalCache devices have an error, they reply with a message of the form "ERROR: some descriptive message". So I was trying to be consistent with that model. In fact if you look directly above at line 560, that's exactly how a null input stream object is handled.

bundles/org.openhab.binding.globalcache/src/main/java/org/openhab/binding/globalcache/internal/handler/GlobalCacheHandler.java
mhilbush: See my reply above. I was hoping to fix a clear error condition rather than rework a bunch of code. Not to say that it shouldn't be reworked at some point. 😉

jlaur: Yeah, I posted this comment seconds before your reply to the other comment, resolving this one. :)

bundles/org.openhab.binding.globalcache/src/main/java/org/openhab/binding/globalcache/internal/handler/GlobalCacheHandler.java
jlaur: OK, I didn't have this knowledge that the prefix "ERROR: " would be recognized and handled somewhere else. So in this case the fix is following the convention, however obscure that is. :)

jlaur (approved):
LGTM

bundles/org.openhab.binding.globalcache/src/main/java/org/openhab/binding/globalcache/internal/handler/GlobalCacheHandler.java
mhilbush: Thanks. Yeah, at some point I definitely need to come back around and spend some time on this binding. I'm pretty sure I can do what needs to be done with a lot less code. 😉

mhilbush: And I also was feeling a little time pressure with the release just around the corner.

Issue #12983, opened 2022-06-21
[Homekit] change request: support a switch  (off=0/on=1) for TargetFanState property


2xPower:
The TargetFanState  (values: 0=MANUAL, 1=AUTO) property can easily be supported by a Switch item.
Doing so would,  my my specific use case, allow the use of an existing item.
Without support for switch i need to create homekit specific items and rules so keep state synchronised.

2xPower:
a current workaround is the use of complex accesories (openhab 3.3) - and combine the fan with a switchable  accessory.

yfre:
@2xPower  good idea. thanks. a fix is committed, maybe you get it with oh3.3 already

2xPower:
@yfre thank you bro!
And again thank you for all the commits / contributions to the HomeKit implementation. I use what you created about 30 times a day.

yfre:
> @yfre thank you bro! And again thank you for all the commits / contributions to the HomeKit implementation. I use what you created about 30 times a day.

@2xPower you are welcome. 
thank you for your feedback. glad to hear. it motivates me a lot. 
but im not doing it alone. We have excellent and patient source code reviewers ensuring the quality and the community testing and making suggestions.  so, please continue with further feature requests and feedback.

PR #12985, opened 2022-06-21
[homekit] Target Fan state - add support for Switch Item 

yfre:
make use of already available generic functions for Target Fan State and by doing so, add supports for item types switch as requested here #12983  as well as support for "inverted" flag


jlaur (approved):
LGTM - thank you!

PR #12988, opened 2022-06-22
[dynamodb] Corrected instructions configuring AWS permissions

BuckligerBaron:
The documented Procedure with creating an Access Policy First with a ARN you get in Step 2 was wrong. I hope this is better to understand.
Signed-off-by: Tobias Koch <toxxic@gmx.net>

### DIFF
--- a/bundles/org.openhab.persistence.dynamodb/README.md
+++ b/bundles/org.openhab.persistence.dynamodb/README.md
@@ -48,7 +48,7 @@ Please also note possible [Free Tier](https://aws.amazon.com/free/) benefits.
 **Note:** The below policy assumes that `eu-west-1` region is used, the new table schema is used, and the default table name of `openhab` is used.
 Modify the policy accordingly if needed.
 
-**Note 2:** As a more simple alternative, one can use pre-existing policy `AmazonDynamoDBFullAccess`, although the policy grants the openHAB user wider-than-necessary permissions.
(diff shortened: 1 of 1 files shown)


BuckligerBaron:
grafik

It is documented here how to Sign Off with the Github Webeditor:
https://community.openhab.org/t/dco-check-signing-off-with-github-web-editor-explanation/83330

bundles/org.openhab.persistence.dynamodb/README.md
```diff
@@ -48,7 +48,7 @@ Please also note possible [Free Tier](https://aws.amazon.com/free/) benefits.
 **Note:** The below policy assumes that `eu-west-1` region is used, the new table schema is used, and the default table name of `openhab` is used.
 Modify the policy accordingly if needed.
 
-**Note 2:** As a more simple alternative, one can use pre-existing policy `AmazonDynamoDBFullAccess`, although the policy grants the openHAB user wider-than-necessary permissions.
+**Note2:** The Number in arn:aws:dynamodb:eu-west-1:084669220525:table/openhab must be replaced by the unique ID of your own table to grant access for your new db user. To get this ID there are two Options:The first one is that you create the table manually in the AWS Console. Once the table is created you could click on the new table and could see your unique Amazon Resource Name starting with arn: which you must replace in the JSON script.  The other is that you give your new database user the existing IAM policy `AmazonDynamoDBFullAccess` to grant him the right to create the new table. After that you could revoke the `AmazonDynamoDBFullAccess` from the openhab user for security reasons.
```
ssalonen, line 51: I would not recommend creating the table manually as you would need to carefully configure all the things by hand. (Time to live and whatnot) incorrect config will lead to all sort of issues.

The binding configures these things only when the binding is creating the table.

Can we not change the policy on this page such that it would allow creating the table, similar how AmazonDynamoDBFullAccess allows creating a table? I might have some time to look this during the next week or two.

ssalonen:
@BuckligerBaron I do not see the need for the complicated 2-phase setup ("create table manually first" or "have AmazonDynamoDBFullAccess first") you are proposing here. 

I tested locally and the addon can create the table right away, even when limited by the policy. Naturally, if the table name is not allowed by the policy, addon cannot do much -- this is expected.

I added more steps in my PR here: https://github.com/openhab/openhab-addons/pull/13000 . Can you please try those out and see if it works for you?

BuckligerBaron:
Your PR looks good. I also agree to your objections. I will close my PR because your is better.

PR #12989, opened 2022-06-22
[boschindego] Add channels for last/next cutting time

### LINKED ISSUE #12986: [boschindego] Last/next cutting time
The API provides methods for requesting last/next cutting time from the device. These timestamps could be exposed as new channels.

They should not be updated as often as normal state since each timestamp require a distinct call and information is quite static.
- Last cutting could be updated when state indicates a possible change, i.e. recently cutting.
- Next cutting will change according to calendar configuration, so it would make sense to refresh with low frequency (perhaps every few hours) in case of configuration change and after current timestamp is exceeded. However, SmartMowing should also be considered, and time could change more unpredictably.

### Last cutting
**Method:** GET https://api.indego.iot.bosch-si.com/
**Path:** /api/v1/alms/serialNumber/predictive/lastcutting

Example response:

```
{
    "last_mowed": "2022-06-21T18:14:22.425+01:00"
}
```

### Next cutting
**Method:** GET https://api.indego.iot.bosch-si.com/
**Path:** /api/v1/alms/serialNumber/predictive/nextcutting

Example response:

```
{
    "mow_next": "2022-06-22T09:30:00+01:00"
}
```

When calendar hasn't been configured yet, response body is empty and HTTP code 204 ("No Content") is returned.

See also: https://github.com/zazaz-de/iot-device-bosch-indego-controller/blob/master/PROTOCOL.md

jlaur:
Fixes #12986

Additionally:
- Add missing description for channel type **mowed**.

### DIFF
--- a/bundles/org.openhab.binding.boschindego/src/main/java/org/openhab/binding/boschindego/internal/handler/BoschIndegoHandler.java
+++ b/bundles/org.openhab.binding.boschindego/src/main/java/org/openhab/binding/boschindego/internal/handler/BoschIndegoHandler.java
@@ -64,8 +64,6 @@ public class BoschIndegoHandler extends BaseThingHandler {
     private @NonNullByDefault({}) IndegoController controller;
     private @Nullable ScheduledFuture<?> statePollFuture;
     private @Nullable ScheduledFuture<?> cuttingTimePollFuture;
-    private long stateRefreshRate;
-    private long cuttingTimeRefreshRate;
     private boolean propertiesInitialized;
     private int previousStateCode;
 
@@ -96,14 +94,12 @@ public class BoschIndegoHandler extends BaseThingHandler {
         }
 
         controller = new IndegoController(httpClient, username, password);
-        stateRefreshRate = config.refresh;
-        cuttingTimeRefreshRate = config.cuttingTimeRefresh;
 
         updateStatus(ThingStatus.UNKNOWN);
         this.statePollFuture = scheduler.scheduleWithFixedDelay(this::refreshStateWithExceptionHandling, 0,
-                stateRefreshRate, TimeUnit.SECONDS);
+                config.refresh, TimeUnit.SECONDS);
         this.cuttingTimePollFuture = scheduler.scheduleWithFixedDelay(this::refreshCuttingTimesWithExceptionHandling, 0,
-                cuttingTimeRefreshRate, TimeUnit.MINUTES);
+                config.cuttingTimeRefresh, TimeUnit.MINUTES);
     }
 
     @Override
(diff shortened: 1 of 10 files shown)


lolodomo:
@jlaur : build failed

jlaur:
@lolodomo - would this be possible for you to review? I'm piling up the next improvements, so considering whether to include them here, or try to keep it clean and separated with potentially multiple layers of PR dependencies, rebasing etc. If you would be able to review this week, I would prefer the clean approach with smaller PR's. :)

bundles/org.openhab.binding.boschindego/src/main/java/org/openhab/binding/boschindego/internal/handler/BoschIndegoHandler.java
```diff
@@ -120,6 +140,23 @@ public void handleCommand(ChannelUID channelUID, Command command) {
...
 
+    private void handleRefreshCommand(String channelId) {
+        switch (channelId) {
+            case STATE:
+            case TEXTUAL_STATE:
+            case MOWED:
+            case ERRORCODE:
+            case STATECODE:
+            case READY:
+                scheduler.submit(() -> this.refreshStateWithExceptionHandling());
```
lolodomo, line 151: I see that it was already done like that before, but can you please explain why it is done asynchronously ?

bundles/org.openhab.binding.boschindego/src/main/java/org/openhab/binding/boschindego/internal/handler/BoschIndegoHandler.java
```diff
@@ -55,16 +59,22 @@ public class BoschIndegoHandler extends BaseThingHandler {
...
     private final BoschIndegoTranslationProvider translationProvider;
+    private final TimeZoneProvider timeZoneProvider;
 
     private @NonNullByDefault({}) IndegoController controller;
-    private @Nullable ScheduledFuture<?> pollFuture;
-    private long refreshRate;
+    private @Nullable ScheduledFuture<?> statePollFuture;
+    private @Nullable ScheduledFuture<?> cuttingTimePollFuture;
+    private long stateRefreshRate;
+    private long cuttingTimeRefreshRate;
```
lolodomo, line 68: Apparently, you don't really class members for `stateRefreshRate` and `cuttingTimeRefreshRate`. Simple local variables would be enough.

lolodomo (changes requested):
Only two minor comments.

bundles/org.openhab.binding.boschindego/src/main/java/org/openhab/binding/boschindego/internal/handler/BoschIndegoHandler.java
jlaur: Actually I can't, you have a point. I was somehow assuming it had to be async, probably because I had the hdpowerview binding in mind. I will need to revisit that. Changed to send requests synchronously.

bundles/org.openhab.binding.boschindego/src/main/java/org/openhab/binding/boschindego/internal/handler/BoschIndegoHandler.java
jlaur: You are completely right - fixed.

lolodomo (approved):
LGTM

PR #12990, opened 2022-06-22
[boschindego] Fix communication error description on failed HTTP call

jlaur:
Offline/communication error description should include HTTP code, but actually was set to "The login credentials are wrong or another client is connected to your Indego account".


lolodomo (approved):
LGTM

Issue #12997, opened 2022-06-23
[wundergroundupdatereceiver] The binding adds autodiscovered channels even on things defined in files


danieldemus:
## Expected Behaviour
The channels defined in the thing file are unchangeable even when a request is received with unmapped parameters while autodiscovery is enabled.

## Current Behaviour
Any unmapped channels are added to the thing, but disappear again when openhab is restarted.

## Possible Solution
Check whether the thing is managed or not before adding channels.

## Steps to Reproduce (for Bugs)
1. Create a WundergrounUpdateReceiver thing in a things file. Include some known channels.
2. Post a request with parameters not mentioned in the things file.
3. Observe that extra channels have appeared on the thing.

PR #10105, opened 2021-02-08
[wundergroundupdatereceiver] Initial contribution

danieldemus:
Signed-off-by: Daniel Demus <daniel-github@demus.dk>

Being the owner of a weather station that is able to submit measurements to wunderground.com via the update URL, I looked for a binding to capture the values and use them locally on my openhab instance to fx. open and close windows or awnings depending on weather conditions, or just to display current weather on a weather page.

I couldn't find any thus this binding. It registers a servlet listening on the wunderground.com update URL path and gathers the submitted values into channels on the thing. Multiple wunderground accounts can be supported by configuring the station id of the thing. The station id must match the ID parameter in the requests from the weather station. Every update is also passed along in a trigger channel, which can then resend the update to wunderground.com in a trivial rule script. In addition you can also submit measurements to other weather services in whatever format they accept.

The build artifact can be found here:
https://openhab.jfrog.io/artifactory/libs-pullrequest-local/org/openhab/addons/bundles/org.openhab.binding.wundergroundupdatereceiver/3.2.0-SNAPSHOT/

PR #12998, opened 2022-06-23
[wundergroundupdatereceiver] Bugfix: Check whether the thing is managed before adding newly discovered parameters as channels.

danieldemus:
# Description
Fix for bug #12997


jlaur:
@fwolter - would it be possible for you to review this in time for 3.3, since you did the review of #10105?

fwolter:
If you just need to check if a Thing exists, `ThingRegistry` seems to be more suitable.

danieldemus:
> If you just need to check if a Thing exists, `ThingRegistry` seems to be more suitable.

I need to know whether it's configured in a file or the UI. This is how the rest api finds out whether the thing is editable. It comes from here: ThingResource.java::725. It might be an idea to add a property such as isEditable to Thing.

fwolter (approved):
Ok, LGTM

PR #13003, opened 2022-06-24
[pidcontroller] Implement previous state recovery on startup

lnagel:
This feature allows the PID controller parameters to be updated and OpenHAB to be restarted without losing the current controller state.

This is especially important for systems with a long response time. For example it might take up to a day for an underfloor heating
controller to stabilise after losing state and having to build up the integrator value from zero.

In case the controller instance has configured OpenHAB items for the integral part, derivative part and error inspectors, then we can query the previous value from the persistance layer for these items and restore them during the controller startup.

I have been using this successfully for about 4 months now during the last heating season.

### DIFF
--- a/bundles/org.openhab.automation.pidcontroller/src/main/java/org/openhab/automation/pidcontroller/internal/handler/PIDControllerTriggerHandler.java
+++ b/bundles/org.openhab.automation.pidcontroller/src/main/java/org/openhab/automation/pidcontroller/internal/handler/PIDControllerTriggerHandler.java
@@ -223,7 +223,7 @@ public class PIDControllerTriggerHandler extends BaseTriggerModuleHandler implem
 
         try {
             value = getItemValueAsNumber(itemRegistry.getItem(itemName));
-            logger.info("Item '{}' value {} recovered by PID controller", itemName, value);
+            logger.debug("Item '{}' value {} recovered by PID controller", itemName, value);
         } catch (ItemNotFoundException e) {
             throw new IllegalArgumentException("Configured item not found: " + itemName, e);
         } catch (PIDException e) {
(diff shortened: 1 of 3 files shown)


bundles/org.openhab.automation.pidcontroller/src/main/java/org/openhab/automation/pidcontroller/internal/handler/PIDControllerTriggerHandler.java
```diff
@@ -208,6 +213,26 @@ private TriggerHandlerCallback getCallback() {
...
+            throws IllegalArgumentException {
+        double value = 0.0;
+
+        if (itemName == null) {
+            return value;
+        }
+
+        try {
+            value = getItemValueAsNumber(itemRegistry.getItem(itemName));
+            logger.info("Item '{}' value {} recovered by PID controller", itemName, value);
```
fwolter, line 226: This should be debug or trace.

fwolter (changes requested):
Looks like a handy solution!

Can you add a short note to the readme, that you need to enable the inspector items to recover the values? And maybe add a reference to the core documentation how to setup persistence on Items?

bundles/org.openhab.automation.pidcontroller/src/main/java/org/openhab/automation/pidcontroller/internal/handler/PIDControllerTriggerHandler.java
lnagel: Fixed.

lnagel:
Well I fixed everything, but it seems that builds are failing now (and I have a feeling that it's related to the build failures in main).

fwolter (changes requested):

fwolter (approved):
LGTM, thanks!

lnagel:
Cheers!

PR #13009, opened 2022-06-25
[remoteopenhab] Use channelTypeI18nLocalizationService to localize

lolodomo:
channel types built dynamically


lolodomo:
Does not work as expected, I forgot to set the parameter to the label/description. Unfortunately, it is not supported by the core framework.

PR #13010, opened 2022-06-25
[squeezebox] Fix NumberFormatException when parsing status message

### LINKED ISSUE #13005: [squeezebox] Infrequently not working after receiving string for parsing as number
From time to time (about once every two weeks), the squeezebox binding stops working.
There does not seem to be a manual trigger for this.

The logs shows:
```
Exception in thread "Squeeze Server Listener" java.lang.NumberFormatException: For input string: "N/A",
	at java.base/jdk.internal.math.FloatingDecimal.readJavaFormatString(FloatingDecimal.java:2054),
	at java.base/jdk.internal.math.FloatingDecimal.parseDouble(FloatingDecimal.java:110),
	at java.base/java.lang.Double.parseDouble(Double.java:543),
	at org.openhab.binding.squeezebox.internal.handler.SqueezeBoxServerHandler$SqueezeServerListener.handleStatusMessage(SqueezeBoxServerHandler.java:699),
	at org.openhab.binding.squeezebox.internal.handler.SqueezeBoxServerHandler$SqueezeServerListener.handlePlayerUpdate(SqueezeBoxServerHandler.java:604),
	at org.openhab.binding.squeezebox.internal.handler.SqueezeBoxServerHandler$SqueezeServerListener.run(SqueezeBoxServerHandler.java:481)
```

Workaround:
Disable and reenable the squeezebox:squeezeboxserver thing.

Environment:
openhab Snapshot Build 2955, also  happening in 3.2.0 stable
Logitech media server 8.2.1 - TRUNK
both in docker containers

----

According to the logs, the crulip lies here, trying to parse "N/A" as Number:
https://github.com/openhab/openhab-addons/blob/ccbc6a8f1167ebba697c8e87eefd755767c3b69d/bundles/org.openhab.binding.squeezebox/src/main/java/org/openhab/binding/squeezebox/internal/handler/SqueezeBoxServerHandler.java#L699

There does not seem to be any exception handling in this function.

mhilbush:

Fixes #13005

### DIFF
--- a/bundles/org.openhab.binding.squeezebox/src/main/java/org/openhab/binding/squeezebox/internal/handler/SqueezeBoxServerHandler.java
+++ b/bundles/org.openhab.binding.squeezebox/src/main/java/org/openhab/binding/squeezebox/internal/handler/SqueezeBoxServerHandler.java
@@ -766,7 +766,7 @@ public class SqueezeBoxServerHandler extends BaseBridgeHandler {
                     }
                 } catch (NumberFormatException e) {
                     // Skip this key/value
-                    logger.debug("Cannot parse number in status message: value '{}', key '{}'", entry.key, entry.value);
+                    logger.debug("Cannot parse number in status message: key '{}', value '{}'", entry.key, entry.value);
                 }
             }
 
--- a/bundles/org.openhab.binding.squeezebox/src/main/java/org/openhab/binding/squeezebox/internal/handler/SqueezeBoxServerHandler.java
+++ b/bundles/org.openhab.binding.squeezebox/src/main/java/org/openhab/binding/squeezebox/internal/handler/SqueezeBoxServerHandler.java
@@ -696,7 +696,7 @@ public class SqueezeBoxServerHandler extends BaseBridgeHandler {
                         updatePlayer(listener -> listener.modeChangeEvent(mac, entry.value));
                     }
                     // Parameter Playing Time
-                    else if ("time".equals(entry.key)) {
+                    else if ("time".equals(entry.key) && !"N/A".equals(entry.value)) {
                         final int time = (int) Double.parseDouble(entry.value);
                         updatePlayer(listener -> listener.currentPlayingTimeEvent(mac, time));
                     }
--- a/bundles/org.openhab.binding.squeezebox/src/main/java/org/openhab/binding/squeezebox/internal/handler/SqueezeBoxServerHandler.java
+++ b/bundles/org.openhab.binding.squeezebox/src/main/java/org/openhab/binding/squeezebox/internal/handler/SqueezeBoxServerHandler.java
@@ -680,88 +680,93 @@ public class SqueezeBoxServerHandler extends BaseBridgeHandler {
             String artworkUrl = null;
 
             for (KeyValue entry : decodeKeyValueResponse(messageParts)) {
-                // Parameter Power
-                if ("power".equals(entry.key)) {
-                    final boolean power = "1".equals(entry.value);
-                    updatePlayer(listener -> listener.powerChangeEvent(mac, power));
-                }
-                // Parameter Volume
-                else if ("mixer volume".equals(entry.key)) {
-                    final int volume = (int) Double.parseDouble(entry.value);
-                    updatePlayer(listener -> listener.absoluteVolumeChangeEvent(mac, volume));
-                }
-                // Parameter Mode
-                else if ("mode".equals(entry.key)) {
-                    updatePlayer(listener -> listener.modeChangeEvent(mac, entry.value));
-                }
-                // Parameter Playing Time
-                else if ("time".equals(entry.key)) {
-                    final int time = (int) Double.parseDouble(entry.value);
-                    updatePlayer(listener -> listener.currentPlayingTimeEvent(mac, time));
-                }
-                // Parameter duration
-                else if ("duration".equals(entry.key)) {
-                    final int duration = (int) Double.parseDouble(entry.value);
-                    updatePlayer(listener -> listener.durationEvent(mac, duration));
-                }
-                // Parameter Playing Playlist Index
-                else if ("playlist_cur_index".equals(entry.key)) {
-                    final int index = (int) Double.parseDouble(entry.value);
-                    updatePlayer(listener -> listener.currentPlaylistIndexEvent(mac, index));
-                }
-                // Parameter Playlist Number Tracks
-                else if ("playlist_tracks".equals(entry.key)) {
-                    final int track = (int) Double.parseDouble(entry.value);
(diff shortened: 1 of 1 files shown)


bundles/org.openhab.binding.squeezebox/src/main/java/org/openhab/binding/squeezebox/internal/handler/SqueezeBoxServerHandler.java
```diff
@@ -680,88 +680,93 @@ private void handleStatusMessage(final String mac, String[] messageParts) {
...
+                    // Id for covert art (only valid when coverart is "1")
+                    else if ("coverid".equals(entry.key)) {
+                        coverid = entry.value;
+                    } else {
+                        // Added to be able to see additional status message types
+                        logger.trace("Unhandled status message type '{}' (value '{}')", entry.key, entry.value);
+                    }
+                } catch (NumberFormatException e) {
+                    // Skip this key/value
+                    logger.debug("Cannot parse number in status message: value '{}', key '{}'", entry.key, entry.value);
```
lolodomo, line 769: Looks like your parameters are reversed.

lolodomo (changes requested):

bundles/org.openhab.binding.squeezebox/src/main/java/org/openhab/binding/squeezebox/internal/handler/SqueezeBoxServerHandler.java
mhilbush: Ugh. Too early in the morning...

Good catch. Thanks!

Fixed

bundles/org.openhab.binding.squeezebox/src/main/java/org/openhab/binding/squeezebox/internal/handler/SqueezeBoxServerHandler.java
```diff
@@ -680,88 +680,93 @@ private void handleStatusMessage(final String mac, String[] messageParts) {
...
+                        final int volume = (int) Double.parseDouble(entry.value);
+                        updatePlayer(listener -> listener.absoluteVolumeChangeEvent(mac, volume));
+                    }
+                    // Parameter Mode
+                    else if ("mode".equals(entry.key)) {
+                        updatePlayer(listener -> listener.modeChangeEvent(mac, entry.value));
+                    }
+                    // Parameter Playing Time
+                    else if ("time".equals(entry.key)) {
+                        final int time = (int) Double.parseDouble(entry.value);
```
jlaur, line 700: Perhaps we should not even try to parse "N/A" here, if this is a known string (described in linked issue)? This would be faster than having an exception thrown.

jlaur (changes requested):
Also one comment from me.

bundles/org.openhab.binding.squeezebox/src/main/java/org/openhab/binding/squeezebox/internal/handler/SqueezeBoxServerHandler.java
```diff
@@ -680,88 +680,93 @@ private void handleStatusMessage(final String mac, String[] messageParts) {
...
+                    else if ("mixer volume".equals(entry.key)) {
+                        final int volume = (int) Double.parseDouble(entry.value);
+                        updatePlayer(listener -> listener.absoluteVolumeChangeEvent(mac, volume));
+                    }
+                    // Parameter Mode
+                    else if ("mode".equals(entry.key)) {
+                        updatePlayer(listener -> listener.modeChangeEvent(mac, entry.value));
+                    }
+                    // Parameter Playing Time
+                    else if ("time".equals(entry.key)) {
```
jlaur, line 699: So something like this in addition to exception be caught:
```suggestion
                    else if ("time".equals(entry.key) && !"N/A".equals(entry.value)) {
```

bundles/org.openhab.binding.squeezebox/src/main/java/org/openhab/binding/squeezebox/internal/handler/SqueezeBoxServerHandler.java
mhilbush: I thought about checking explicitly for "N/A". But, since this is the first I've heard of this issue in several years of using/maintaining this binding, I'm just not sure where it came from. So, I thought it best to be defensive in this section of code since there are multiple places where `parseDouble` can fail.

WDYT?

mhilbush: And, the net effect of a failure here is that the listener job ends, which requires a bundle restart.

bundles/org.openhab.binding.squeezebox/src/main/java/org/openhab/binding/squeezebox/internal/handler/SqueezeBoxServerHandler.java
mhilbush: Oh, I see. You're suggesting adding this **in addition** to catching the exception. Yes, I can do that.

mhilbush: Done

jlaur: I was not able to quickly find this value in the LMS source code. Probably would have to clone the repository and perform a full text search. But this might have been introduced in some recent version, and we could also learn which keys would be affected (if not only **time**). Doesn't need to be in this PR, just brain-storming. :)

mhilbush: Maybe it's not in the LMS source code. It might be returned from the content provider.

In any event, I added it. 😉

jlaur (approved):
LGTM

lolodomo (approved):
LGTM

jlaur:
@kaikreuzer - although this is not a regression in 3.3 and probably doesn't meet the criteria for being "critical", this bug fixed here would make the binding stop working because of unhandled `RuntimeException`. More over, the nature of the fix is quite low risk IMHO, so I would consider it for 3.3. WDYT?

kaikreuzer (approved):
Well, as it is a bug that makes the binding stop working, let's consider it critical and have it merged. Thanks, @mhilbush!

mhilbush:
Thanks guys. Very much appreciated!!

PR #13014, opened 2022-06-26
[netatmo] Document extensible channels

lolodomo:

### DIFF
--- a/bundles/org.openhab.binding.netatmo/README.md
+++ b/bundles/org.openhab.binding.netatmo/README.md
@@ -186,6 +186,21 @@ Based on a standard update period of 10mn by Netatmo systems - it will auto adap
 
 All these channels are read only.
 
+**Extensible channels for the main indoor module:**
+
+| Channel Type         | Item Type            | Description                       | Channel parameters               |
+|----------------------|----------------------|-----------------------------------|----------------------------------|
+| co2-measurement      | Number:Dimensionless | CO2 measurement                   | limit (MIN, MAX), period (30min, 1hour, 3hours, 1day, 1week, 1month) |
+| co2-timestamp        | DateTime             | CO2 measurement timestamp         | limit (DATE_MIN, DATE_MAX), period (1week, 1month) |
+| hum-measurement      | Number:Dimensionless | Humidity measurement              | limit (MIN, MAX), period (30min, 1hour, 3hours, 1day, 1week, 1month) |
+| hum-timestamp        | DateTime             | Humidity measurement timestamp    | limit (DATE_MIN, DATE_MAX), period (1week, 1month) |
+| noise-measurement    | Number:Dimensionless | Noise measurement                 | limit (MIN, MAX), period (30min, 1hour, 3hours, 1day, 1week, 1month) |
+| noise-timestamp      | DateTime             | Noise measurement timestamp       | limit (DATE_MIN, DATE_MAX), period (1week, 1month) |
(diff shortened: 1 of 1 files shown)


lolodomo:
@jlaur @kaikreuzer : as this is only documentation and a feature which is not yet documented, I suggest to merge this PR before 3.3 is released.

@clinique : if my interpretation of the code is correct, there is no "CO2 Measurement" extensible channel type for the Healthy Home Coach. I don't know if this is deliberate and expected, or if it is a miss.

clinique:
> @clinique : if my interpretation of the code is correct, there is no "CO2 Measurement" extensible channel type for the Healthy Home Coach. I don't know if this is deliberate and expected, or if it is a miss.

You're correct, this is a miss. (BTW there is no official documentation in API doc of measurement for healthy home coach, I discovered that it worked like weather station)

clinique (approved):

bundles/org.openhab.binding.netatmo/README.md
```diff
@@ -264,6 +299,12 @@ All these channels are read only.
 
 All these channels are read only.
 
+**Extensible channels for the rain guage:**
+
+| Channel Type         | Item Type            | Description                       | Channel parameters                                 |
+|----------------------|----------------------|-----------------------------------|----------------------------------------------------|
+| sum_rain-measurement | Number:Length        | Summing rain measurement          | period (30min, 1hour, 3hours, 1day, 1week, 1month) |
```
jlaur, line 306: Is it really **sum_rain** with underscore? Just checking, as it seems inconsistent with convention used.

lolodomo: Yes.
https://github.com/openhab/openhab-addons/blob/main/bundles/org.openhab.binding.netatmo/src/main/java/org/openhab/binding/netatmo/internal/api/data/NetatmoConstants.java#L87

lolodomo:
> > @clinique : if my interpretation of the code is correct, there is no "CO2 Measurement" extensible channel type for the Healthy Home Coach. I don't know if this is deliberate and expected, or if it is a miss.
> 
> You're correct, this is a miss. (BTW there is no official documentation in API doc of measurement for healthy home coach, I discovered that it worked like weather station)

Ok, I let you create a potential PR to fix that miss in OH 3.4.

bundles/org.openhab.binding.netatmo/README.md
jlaur: Now wondering if this is by mistake, and if we should change it to "sum-rain" for consistency? In 3.3 all channels are renamed anyway, so last chance to fix this without further breaking. @clinique - is it by accident, or by purpose that underscore is used here?

clinique: Yes, it is not very nice but by design because I use Netatmo API keyword

lolodomo:
@clinique : I don't know how you discovered this feature (extensible channel types). I am even not sure if this is really documented.
The method "withExtensibleChannelTypeIds" is used only by 2 bindings: the Netatmo binding and the MQTT generic binding.

clinique:
I think, IIRW that it has been introduced with gps tracker binding. Maybe is it used dynamically only by mqtt and Netatmo, but statically by many more (I also used it in openuv binding for the safe exposure channel)

lolodomo:
> but statically by many more (I also used it in openuv binding for the safe exposure channel)

Ah ok this is the `extensible=` in XML description.

clinique:
Correct

PR #13016, opened 2022-06-26
[gardena] Improve API rate limit handling

### LINKED ISSUE #13011: [gardena] Eliminate dangling class references on shutdown
### Background

As described in #13001 the binding could cause ClassCastException in some cases when it was stopped and restarted. These ClassCastException were eliminated as a quick fix in #13004. However as @kaikreuzer stated in https://github.com/openhab/openhab-addons/issues/13001#issuecomment-1166243331 we should look at eliminating the underlying cause(s).

### Solution (hypothesis)

There are a couple of hypotheses as to the cause..
- unfinished tasks on the scheduler that are executed after the owner things may have been disposed
- references to classes held in Maps or Sets that are not cleared on dispose (e.g. `allDevicesById` or `devicesToNotify` to give two potential examples)
- other possible reasons tbd

### LINKED ISSUE #13137: [gardena] Error 429 Limit Exceeded
The gardena account Thing sometimes goes permanently OFFLINE with the message "Error 429 Limit Exceeded".

Disabling/enabling resolves the situation.

_Expected behavior:_

The binding itself should be able to automatically do this by itself without requiring a user intervention.

andrewfg:
Closes #13011 

Draft code to potentially resolve the above issue (for discussion).

Fixes #13137

Improve code for more graceful handling of HTTP 429 errors (see https://community.openhab.org/t/gardena-error-429-limit-exceeded/137419)


bundles/org.openhab.binding.gardena/src/main/java/org/openhab/binding/gardena/internal/handler/GardenaAccountHandler.java
```diff
@@ -79,52 +90,93 @@ public void setDiscoveryService(GardenaDeviceDiscoveryService discoveryService)
...
+                        : isRepeatError ? REINITIALIZE_DELAY_MINUTES_BACK_OFF : REINITIALIZE_DELAY_SECONDS;
+
+                ScheduledFuture<?> reInitializationTask = this.reInitializationTask;
+                if (reInitializationTask == null || reInitializationTask.isDone()
+                        || (isHttp429Error != reInitializationCausedBy429)) {
+                    reInitializationTask = scheduleReinitialize(delay);
                 }
-                logger.warn("{}", ex.getMessage());
+                reInitializationCausedBy429 = isHttp429Error;
+                lastErrorTime = LocalDateTime.now();
```
jlaur, line 129: `LocalDateTime` is not well suited for this purpose as it's susceptible to clock changes. I would propose to use `Instant` instead.

bundles/org.openhab.binding.gardena/src/main/java/org/openhab/binding/gardena/internal/handler/GardenaAccountHandler.java
```diff
@@ -79,52 +90,93 @@ public void setDiscoveryService(GardenaDeviceDiscoveryService discoveryService)
...
+            }
+            reInitializationCausedBy429 = false;
+            updateStatus(ThingStatus.ONLINE);
+        } catch (GardenaException ex) {
+            logger.warn("{}", ex.getMessage());
+
+            long secondsUntilRestart;
+
+            synchronized (reInitializationCodeLock) {
+                boolean isHttp429Error = (ex.getStatus() == 429);
```
jlaur, line 116: Constant `HttpStatus.TOO_MANY_REQUESTS_429` (in package `org.eclipse.jetty.http`) could be used here.

bundles/org.openhab.binding.gardena/src/main/java/org/openhab/binding/gardena/internal/handler/GardenaAccountHandler.java
andrewfg: Ok. Done.

bundles/org.openhab.binding.gardena/src/main/java/org/openhab/binding/gardena/internal/handler/GardenaAccountHandler.java
andrewfg: Ok. Done.

andrewfg:
For testing purposes, the JAR file is here..

[org.openhab.binding.gardena-3.4.0-SNAPSHOT.jar.zip](https://github.com/openhab/openhab-addons/files/9175594/org.openhab.binding.gardena-3.4.0-SNAPSHOT.jar.zip)

bundles/org.openhab.binding.gardena/src/main/java/org/openhab/binding/gardena/internal/handler/GardenaAccountHandler.java
```diff
@@ -158,9 +256,12 @@ public Collection<Class<? extends ThingHandlerService>> getServices() {
     @Override
     public void handleCommand(ChannelUID channelUID, Command command) {
         if (RefreshType.REFRESH == command) {
+            // ****
+            // TODO: REALLY ??? should a UI refresh be causing a complete initialisation of the bridge ???
```
kaikreuzer, line 260: Pfew, no, definitely not. Btw, it is not a "UI refresh", but a command that can be sent e.g. by rules. As it is also called once for each channel during startup, this code here can potentially cause quite some hiccups during system startup, so I would tend to completely remove it, if we do not have a chance to easily provide an up-to-date value for the requested channel.

bundles/org.openhab.binding.gardena/src/main/java/org/openhab/binding/gardena/internal/handler/GardenaAccountHandler.java
```diff
@@ -51,80 +60,168 @@
...
 
+    /**
+     * Format a description text to display on the thing main UI page
+     *
+     * @param delaySeconds the delay that will be added to the current time
+     * @return the description text
+     */
+    public String uiText(long delaySeconds) {
+        return "Waiting to make automatic reconnection attempt at " + LocalDateTime.now().plusSeconds(delaySeconds)
+                .truncatedTo(ChronoUnit.SECONDS).format(DateTimeFormatter.ofPattern("HH:MM:SS (cccc)"));
```
maniac103, line 131: If this is an UI text, shouldn't it be localized?

bundles/org.openhab.binding.gardena/src/main/java/org/openhab/binding/gardena/internal/handler/GardenaAccountHandler.java
```diff
@@ -51,80 +60,168 @@
...
+    }
+
     /**
      * Initializes the GardenaSmart account.
+     * This method is called on a thread.
```
maniac103, line 136: Indeed. All methods are ;-)

(Seriously: this comment seems a bit useless?)

bundles/org.openhab.binding.gardena/src/main/java/org/openhab/binding/gardena/internal/handler/GardenaAccountHandler.java
```diff
@@ -51,80 +60,168 @@
...
+            reInitializationCausedBy429 = false;
+            updateStatus(ThingStatus.ONLINE);
+        } catch (GardenaException ex) {
+            logger.warn("{}", ex.getMessage());
+
+            long delaySecs;
+
+            synchronized (reInitializationCodeLock) {
+                Duration delay;
+                boolean isHttp429Error = (HttpStatus.TOO_MANY_REQUESTS_429 == ex.getStatus());
```
maniac103, line 160: I find `variable == constant` a bit more logical than `constant == variable`.

bundles/org.openhab.binding.gardena/src/main/java/org/openhab/binding/gardena/internal/handler/GardenaAccountHandler.java
```diff
@@ -51,80 +60,168 @@
...
+                if (isHttp429Error) {
+                    delay = REINITIALIZE_DELAY_HOURS_LIMIT_EXCEEDED;
                 } else {
-                    // otherwise reinitialize after 120 seconds
-                    scheduleReinitialize(REINITIALIZE_DELAY_SECONDS, TimeUnit.SECONDS);
+                    Instant now = Instant.now();
+                    Instant notBeforeTime = lastApiCallTime().plus(REINITIALIZE_DELAY_MINUTES_BACK_OFF)
+                            .plus(REINITIALIZE_DELAY_MINUTES_BACK_OFF);
```
maniac103, line 166: Why add the delay twice?

bundles/org.openhab.binding.gardena/src/main/java/org/openhab/binding/gardena/internal/handler/GardenaAccountHandler.java
```diff
@@ -51,80 +60,168 @@
...
+
+            updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.COMMUNICATION_ERROR, uiText(delaySecs));
+            disposeGardena();
+        }
+        lastApiCallTimeUpdate();
+    }
+
+    /**
+     * Re-initializes the GardenaSmart account.
+     * This method is called on a thread.
```
maniac103, line 191: As above

bundles/org.openhab.binding.gardena/src/main/java/org/openhab/binding/gardena/internal/handler/GardenaAccountHandler.java
```diff
@@ -51,80 +60,168 @@
...
+    private synchronized void reIninitializeGardena() {
+        if (getThing().getStatus() != ThingStatus.UNINITIALIZED) {
+            initializeGardena();
+        }
     }
 
     /**
      * Schedules a reinitialization, if Gardena smart system account is not reachable.
+     *
+     * @return pointer to the reinitialization task
```
maniac103, line 202: Java doesn't know pointers ;-) Suggestion: `reinitialization task reference`

andrewfg: Ok. Done.

bundles/org.openhab.binding.gardena/src/main/java/org/openhab/binding/gardena/internal/handler/GardenaAccountHandler.java
andrewfg: Yes. 'background'

bundles/org.openhab.binding.gardena/src/main/java/org/openhab/binding/gardena/internal/handler/GardenaAccountHandler.java
andrewfg: Yes. But I don't recall how to do that.

bundles/org.openhab.binding.gardena/src/main/java/org/openhab/binding/gardena/internal/handler/GardenaAccountHandler.java
andrewfg: Ok. Agreed.

( Its just that some people seem to prefer (say) `assertEquals(constant, variable)` or `constant.equals(variable)`. But I agree that `(variable == constant)` is better.. )

bundles/org.openhab.binding.gardena/src/main/java/org/openhab/binding/gardena/internal/handler/GardenaAccountHandler.java
andrewfg: We need to limit the API call rate to one call every 15 minutes. There are three API calls that must be spread over time, namely 1) the 'lastAPICall', 2) the call that caused this current error, and 3) the yet to-be-scheduled reconnection call. So essentially we have to schedule call 3) to not be less than 30 minutes after call 1)..

bundles/org.openhab.binding.gardena/src/main/java/org/openhab/binding/gardena/internal/handler/GardenaAccountHandler.java
andrewfg: Ok. Thanks for the confirmation. I will remove it.

bundles/org.openhab.binding.gardena/src/main/java/org/openhab/binding/gardena/internal/handler/GardenaAccountHandler.java
maniac103: Using [`TranslationProvider`](https://www.openhab.org/docs/developer/utils/i18n.html#i18n-text-provider-api). In your case, the string would be something like `gardena.delayed-reconnection-status = Waiting to make automatic reconnection attempt at {0}`, and the code to generate the string something like

```
DateTimeFormatter formatter = DateTimeFormatter.ofLocalizedTime(FormatStyle.SHORT);
String delayedTimestamp = LocalDateTiume.now().plusSeconds(delaySeconds).format(formatter);
return i18nProvider.getText(bundle, "gardena.delayed-reconnection-status", "", localeProvider.getLocale(), delayedTimestamp);
```

i18nProvider is the `TranslationProvider` instance and `localeProvider` a `LocaleProvider` instance, both of which need to be fetched as OSGi services in the `HandlerFactory`.

bundles/org.openhab.binding.gardena/src/main/java/org/openhab/binding/gardena/internal/handler/GardenaAccountHandler.java
maniac103: Whether or not 2) should count into the delay depends on the exception type: If it's a HTTP error (`ex.getStatus()` returns a positive value) that's likely the case, but in case of e.g. a network issue it's likely not the case.

My suggestion would be something like this:
```
if (ex.getStatus() > 0) {
    // HTTP error -> be safe and assume the server counts it against the request rate limit
    lastApiCallTimeUpdate();
}
notBeforeTime = lastApiCallTime().plus(REINITIALIZE_DELAY_MINUTES_BACK_OFF);
```

That way the logic would be easier to understand.

bundles/org.openhab.binding.gardena/src/main/java/org/openhab/binding/gardena/internal/handler/GardenaAccountHandler.java
andrewfg: I figured it out. Done.

bundles/org.openhab.binding.gardena/src/main/java/org/openhab/binding/gardena/internal/handler/GardenaAccountHandler.java
andrewfg: Good point. Will do.

bundles/org.openhab.binding.gardena/src/main/java/org/openhab/binding/gardena/internal/handler/GardenaAccountHandler.java
```diff
@@ -51,80 +62,191 @@
...
 
+    /**
+     * Format a localised description text to display on the thing main UI page.
+     *
+     * @param delaySeconds the delay that will be added to the current time.
+     * @return the (localised) description text.
+     */
+    private String uiText(long delaySeconds) {
+        return i18nProvider.getText(bundle, RECONNECT_MSG_KEY, RECONNECT_MSG_KEY, localeProvider.getLocale())
+                + LocalDateTime.now().plusSeconds(delaySeconds).format(DATE_TIME_FORMAT);
```
maniac103, line 156: Please use a placeholder (like in my example) to allow for rearranging the sentence; some languages might not put the time to the end of the sentence.

bundles/org.openhab.binding.gardena/src/main/java/org/openhab/binding/gardena/internal/handler/GardenaAccountHandler.java
jlaur: Usually this is to eliminate risk of `NullPointerException` if variable might be null, e.g.:
```
"Hello".equals(a); // will never throw a NPE
a.equals("Hello"); // can throw an NPE.
```

Your `assertEquals()` example is not valid, as it takes _expected_ as first argument, so you can't choose the order.

bundles/org.openhab.binding.gardena/src/main/java/org/openhab/binding/gardena/internal/handler/GardenaAccountHandler.java
```diff
@@ -51,80 +60,168 @@
...
+                }
+                reInitializationCausedBy429 = isHttp429Error;
+                delaySecs = reInitializationTask.getDelay(TimeUnit.SECONDS);
             }
-        });
+
+            updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.COMMUNICATION_ERROR, uiText(delaySecs));
+            disposeGardena();
+        }
+        lastApiCallTimeUpdate();
```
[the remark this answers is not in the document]
andrewfg: Actually it must be in both the try and the catch parts; with different conditions applied.

bundles/org.openhab.binding.gardena/src/main/java/org/openhab/binding/gardena/internal/handler/GardenaAccountHandler.java
andrewfg: Ok. Will do.

bundles/org.openhab.binding.gardena/src/main/java/org/openhab/binding/gardena/internal/handler/GardenaAccountHandler.java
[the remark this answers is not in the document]
maniac103: That's what I was heading at ... Sorry for unclear wording.

bundles/org.openhab.binding.gardena/src/main/java/org/openhab/binding/gardena/internal/handler/GardenaAccountHandler.java
```diff
@@ -51,80 +62,191 @@
...
+    private final LocaleProvider localeProvider;
 
-    public GardenaAccountHandler(Bridge bridge, HttpClientFactory httpClientFactory,
-            WebSocketFactory webSocketFactory) {
+    // re- initialisation stuff
+    private final Object reInitializationCodeLock = new Object();
+    private @Nullable ScheduledFuture<?> reInitializationTask;
+    private boolean reInitializationCausedBy429 = false;
+    private Instant lastApiCallTime = Instant.MIN;
+    private boolean lastApiCallTimeLoaded = false;
```
jlaur, line 89: Minor, but you could compare `lastApiCallTime` to `Instant.MIN` to know if loaded/initialized, thus eliminate the need for this additional variable.

bundles/org.openhab.binding.gardena/src/main/java/org/openhab/binding/gardena/internal/handler/GardenaAccountHandler.java
```diff
@@ -51,80 +62,191 @@
...
+        this.localeProvider = localeProvider;
+    }
+
+    /**
+     * Returns the time when the last api call was made. If the thing was newly created, restores the time from a
+     * property to ensure consistent behaviour over restarts.
+     *
+     * @return the time when the last api call was made.
+     */
+    private Instant lastApiCallTime() {
```
jlaur, line 107: Perhaps `getLastApiCallTime()` would be more descriptive?

bundles/org.openhab.binding.gardena/src/main/java/org/openhab/binding/gardena/internal/handler/GardenaAccountHandler.java
```diff
@@ -51,80 +62,191 @@
...
     }
 
     @Override
     public void initialize() {
         logger.debug("Initializing Gardena account '{}'", getThing().getUID().getId());
-        initializeGardena();
+        Instant now = Instant.now();
+        Instant notBeforeTime = lastApiCallTime().plus(REINITIALIZE_DELAY_MINUTES_BACK_OFF)
+                .plus(REINITIALIZE_DELAY_MINUTES_BACK_OFF);
+        if (now.isBefore(notBeforeTime)) {
```
jlaur, line 133: Readability: "is before not before time" is somehow hard for me to read. :) Consider inverting the logic, for example by renaming variable to `scheduledAfter`, then:
```suggestion
        if (now.isAfter(scheduledAfter)) {
```
(and swapping the two blocks, of course)

bundles/org.openhab.binding.gardena/src/main/java/org/openhab/binding/gardena/internal/handler/GardenaAccountHandler.java
jlaur, line 156: `LocalDateTime.now()` will use default time-zone, not time-zone configured in openHAB. Since it's just a text string for the offline status, you can leave it as is, but if you would like to display the local time for the configured time-zone, you should use [TimeZoneProvider](https://www.openhab.org/javadoc/latest/org/openhab/core/i18n/timezoneprovider). You can then use `LocalDateTime.now(timeZoneProvider.getTimeZone())`.

bundles/org.openhab.binding.gardena/src/main/java/org/openhab/binding/gardena/internal/handler/GardenaAccountHandler.java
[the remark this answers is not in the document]
andrewfg: I am adapting the api call suppression concept. 
But the change is based on your comments :)

bundles/org.openhab.binding.gardena/src/main/java/org/openhab/binding/gardena/internal/handler/GardenaAccountHandler.java
andrewfg: Field deleted. And api call suppression modified.

bundles/org.openhab.binding.gardena/src/main/java/org/openhab/binding/gardena/internal/handler/GardenaAccountHandler.java
andrewfg: Api call suppression concept modified.

bundles/org.openhab.binding.gardena/src/main/java/org/openhab/binding/gardena/internal/handler/GardenaAccountHandler.java
andrewfg: Api call suppression concept modified.

bundles/org.openhab.binding.gardena/src/main/java/org/openhab/binding/gardena/internal/handler/GardenaAccountHandler.java
andrewfg: Ok. Done.

andrewfg:
The JAR file, for testing, is here..

[org.openhab.binding.gardena-3.4.0-SNAPSHOT.jar.zip](https://github.com/openhab/openhab-addons/files/9194018/org.openhab.binding.gardena-3.4.0-SNAPSHOT.jar.zip)

bundles/org.openhab.binding.gardena/src/main/java/org/openhab/binding/gardena/internal/handler/GardenaAccountHandler.java
```diff
@@ -51,80 +64,188 @@
...
+     * Format a localised description text to display on the thing main UI page.
+     *
+     * @return the (localised) description text.
+     */
+    private String getUiText() {
+        String template = i18nProvider.getText(bundle, RECONNECT_MSG_KEY, RECONNECT_MSG_KEY,
+                localeProvider.getLocale());
+        String dateTime = LocalDateTime.ofInstant(apiCallSuppressionEnd, timeZoneProvider.getTimeZone())
+                .truncatedTo(ChronoUnit.SECONDS).format(DateTimeFormatter.ISO_LOCAL_DATE_TIME);
+        return MessageFormat.format(template, dateTime);
```
maniac103, line 175: Why don't you simply pass `dateTime` as last parameter to `getText`? The additional `format()` call is not needed.

bundles/org.openhab.binding.gardena/src/main/java/org/openhab/binding/gardena/internal/handler/GardenaAccountHandler.java
```diff
@@ -51,80 +64,188 @@
...
+    /**
+     * Format a localised description text to display on the thing main UI page.
+     *
+     * @return the (localised) description text.
+     */
+    private String getUiText() {
+        String template = i18nProvider.getText(bundle, RECONNECT_MSG_KEY, RECONNECT_MSG_KEY,
+                localeProvider.getLocale());
+        String dateTime = LocalDateTime.ofInstant(apiCallSuppressionEnd, timeZoneProvider.getTimeZone())
+                .truncatedTo(ChronoUnit.SECONDS).format(DateTimeFormatter.ISO_LOCAL_DATE_TIME);
```
maniac103, line 174: Any specific reason for not using a localized date/time format here? Additionally, one could omit the date if the suppression time span is less than 24 hours.

bundles/org.openhab.binding.gardena/src/main/java/org/openhab/binding/gardena/internal/handler/GardenaAccountHandler.java
```diff
@@ -157,11 +279,7 @@ public Collection<Class<? extends ThingHandlerService>> getServices() {
 
     @Override
     public void handleCommand(ChannelUID channelUID, Command command) {
-        if (RefreshType.REFRESH == command) {
-            logger.debug("Refreshing Gardena account '{}'", getThing().getUID().getId());
-            disposeGardena();
-            initializeGardena();
-        }
+        // nothing to do here because the thing has no channels
     }
```
maniac103, line 283: Can simply drop that method override then.

bundles/org.openhab.binding.gardena/src/main/java/org/openhab/binding/gardena/internal/handler/GardenaAccountHandler.java
```diff
@@ -202,8 +320,11 @@ public void onNewDevice(Device device) {
 
     @Override
     public void onError() {
-        updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.COMMUNICATION_ERROR, "Connection lost");
+        Duration delay = REINITIALIZE_DELAY_SECONDS;
+        synchronized (reInitializationCodeLock) {
+            scheduleReinitialize(delay);
```
maniac103, line 325: `scheduleReinitialize(REINITIALIZE_DELAY_SECONDS)`?

bundles/org.openhab.binding.gardena/src/main/java/org/openhab/binding/gardena/internal/handler/GardenaAccountHandler.java
andrewfg: I tried to do that, but I couldn't figure out a way around the compiler warning _"Null type mismatch (type annotations): required '@NonNull String' but this expression has type '@Nullable String'"_

image

bundles/org.openhab.binding.gardena/src/main/java/org/openhab/binding/gardena/internal/handler/GardenaAccountHandler.java
andrewfg: > reason for not using a localized date/time format

No reason. Which do you suggest?

> could omit the date if the suppression time span is less than 24 hours

Agreed. Will do.

bundles/org.openhab.binding.gardena/src/main/java/org/openhab/binding/gardena/internal/handler/GardenaAccountHandler.java
andrewfg: Nope. It is an abstract method in the superior class, so it must be overridden.

bundles/org.openhab.binding.gardena/src/main/java/org/openhab/binding/gardena/internal/handler/GardenaAccountHandler.java
andrewfg: Good point. I will look into that.

bundles/org.openhab.binding.gardena/src/main/java/org/openhab/binding/gardena/internal/handler/GardenaAccountHandler.java
andrewfg: So I suppose the correct syntax is as follows ... yuk!

```
private String getUiText() {
    String msg = i18nProvider.getText(bundle, RECONNECT_MSG_KEY, RECONNECT_MSG_KEY, localeProvider.getLocale(),
            LocalDateTime.ofInstant(apiCallSuppressionEnd, timeZoneProvider.getTimeZone())
                    .truncatedTo(ChronoUnit.SECONDS).format(DateTimeFormatter.ISO_LOCAL_DATE_TIME));
    return msg != null ? msg : "";
}
```

bundles/org.openhab.binding.gardena/src/main/java/org/openhab/binding/gardena/internal/handler/GardenaAccountHandler.java
maniac103: > No reason. Which do you suggest?

See my [earlier comment](https://github.com/openhab/openhab-addons/pull/13016#discussion_r928720621) ;-)

bundles/org.openhab.binding.gardena/src/main/java/org/openhab/binding/gardena/internal/handler/GardenaAccountHandler.java
maniac103: I think the most correct fix is making the return value nullable. Suggestion (including some renaming suggestions):

```
private @Nullable String getReconnectUiMessage() {
    LocalDateTime timestamp = LocalDateTime.ofInstant(apiCallSuppressionEnd, timeZoneProvider.getTimeZone())
            .truncatedTo(ChronoUnit.SECONDS);
    DateTimeFormatter formatter = [...]; // see below comment
    return i18nProvider.getText(bundle, RECONNECT_MSG_KEY, null, localeProvider.getLocale(), timestamp.fornat(formatter));
}
```

That way, the message will simply be omitted in the (impossible) case of the string resource not being found.

bundles/org.openhab.binding.gardena/src/main/java/org/openhab/binding/gardena/internal/handler/GardenaAccountHandler.java
maniac103: OK, TIL :+1:

bundles/org.openhab.binding.gardena/src/main/java/org/openhab/binding/gardena/internal/handler/GardenaAccountHandler.java
```diff
@@ -51,80 +64,188 @@
...
         this.discoveryService = discoveryService;
     }
 
+    /**
+     * Format a localised description text to display on the thing main UI page.
+     *
+     * @return the (localised) description text.
+     */
+    private String getUiText() {
+        String template = i18nProvider.getText(bundle, RECONNECT_MSG_KEY, RECONNECT_MSG_KEY,
```
jlaur, line 171: This is not needed, `updateStatus()` supports "@text/" references directly, i.e. "@text/accounthandler.waiting-to-reconnect-at-time". Example:
https://github.com/openhab/openhab-addons/blob/67006de60663fc367a61a8396f9a1b82543a7709/bundles/org.openhab.binding.danfossairunit/src/main/java/org/openhab/binding/danfossairunit/internal/handler/DanfossAirUnitHandler.java#L110-L111

maniac103: What's the 'this' in that statement though? ;-) The indirection via string constant?

jlaur: Direct usage of `LocaleProvider` and `TranslationProvider`.

bundles/org.openhab.binding.gardena/src/main/java/org/openhab/binding/gardena/internal/handler/GardenaAccountHandler.java
andrewfg: Ok. Done.

bundles/org.openhab.binding.gardena/src/main/java/org/openhab/binding/gardena/internal/handler/GardenaAccountHandler.java
andrewfg: Ok. Done.

bundles/org.openhab.binding.gardena/src/main/java/org/openhab/binding/gardena/internal/handler/GardenaAccountHandler.java
andrewfg: I could not do that because the scheduler task is processed in two different methods -- `dispose()` resp. `scheduleReinitialize()`

bundles/org.openhab.binding.gardena/src/main/java/org/openhab/binding/gardena/internal/handler/GardenaAccountHandler.java
andrewfg: Thanks for the tip. However if you look at the request from @maniac103 above, he requested to have different formatting depending on whether the call suppression expires today (show the time only) or tomorrow (show the date and the time). So I do need a separate method for this.

jlaur: I don't really understand what you mean? You have one i18n text and one parameter which you can format any way you prefer. However, this custom implementation is unneeded and overly complex when you can simply use a "@text/" reference for `updateStatus()`?

andrewfg: ^
Two things..
1. the code formatting is actually quite complex, so it would be hard to write the formatting as a "one liner"
2. updateStatus() is called in more than one part of the code, so the "hard one liner" would be duplicated.

andrewfg:
For testing purposes, the JAR file is here..

[org.openhab.binding.gardena-3.4.0-SNAPSHOT.jar.zip](https://github.com/openhab/openhab-addons/files/9199125/org.openhab.binding.gardena-3.4.0-SNAPSHOT.jar.zip)

bundles/org.openhab.binding.gardena/src/main/java/org/openhab/binding/gardena/internal/handler/GardenaAccountHandler.java
andrewfg: Latest version of the code is below.
If you want to convert that to a one liner, you are welcome to try it :)
```
private String getUiText() {
    Instant until = apiCallSuppressionUntil;
    if (until != null) {
        ZoneId zone = timeZoneProvider.getTimeZone();
        boolean isToday = LocalDate.now(zone).equals(LocalDate.ofInstant(until, zone));
        DateTimeFormatter formatter = isToday ? DateTimeFormatter.ofLocalizedTime(FormatStyle.MEDIUM)
                : DateTimeFormatter.ofLocalizedDateTime(FormatStyle.MEDIUM);
        String value = i18nProvider.getText(bundle, RECONNECT_MSG_KEY, null, localeProvider.getLocale(),
                formatter.format(ZonedDateTime.ofInstant(until, zone)));
        return value != null ? value : "";
    }
    return "";
}
```

bundles/org.openhab.binding.gardena/src/main/java/org/openhab/binding/gardena/internal/handler/GardenaAccountHandler.java
```diff
@@ -51,80 +66,189 @@
...
+    private String getUiText() {
+        Instant until = apiCallSuppressionUntil;
+        if (until != null) {
+            ZoneId zone = timeZoneProvider.getTimeZone();
+            boolean isToday = LocalDate.now(zone).equals(LocalDate.ofInstant(until, zone));
+            DateTimeFormatter formatter = isToday ? DateTimeFormatter.ofLocalizedTime(FormatStyle.MEDIUM)
+                    : DateTimeFormatter.ofLocalizedDateTime(FormatStyle.MEDIUM);
+            String value = i18nProvider.getText(bundle, RECONNECT_MSG_KEY, null, localeProvider.getLocale(),
+                    formatter.format(ZonedDateTime.ofInstant(until, zone)));
+            return value != null ? value : "";
```
maniac103, line 176: Making the return value nullable was not an option? ;-)

bundles/org.openhab.binding.gardena/src/main/java/org/openhab/binding/gardena/internal/handler/GardenaAccountHandler.java
maniac103: How does moving the synchronization to `scheduleReinitialize` affect the synchronization in `dispose()`?

bundles/org.openhab.binding.gardena/src/main/java/org/openhab/binding/gardena/internal/handler/GardenaAccountHandler.java
andrewfg: Aha. Ok. Done.

bundles/org.openhab.binding.gardena/src/main/java/org/openhab/binding/gardena/internal/handler/GardenaAccountHandler.java
```diff
@@ -51,80 +66,189 @@
...
+     */
+    private Duration apiCallSuppressionDelay() {
+        Instant now = Instant.now();
+        Instant until = apiCallSuppressionUntil;
+        return (until != null) && now.isBefore(until) ? Duration.between(now, until) : Duration.ZERO;
+    }
+
+    /**
+     * Updates the time when api call suppression ends to now() plus the given delay. If delay is zero or negative, the
+     * suppression time is nulled. Saves the value as a property to ensure consistent behaviour across restarts.
```
jlaur, line 132: Please be aware that you cannot rely on properties to be persisted:
https://www.openhab.org/docs/developer/bindings/#updating-thing-properties

I don't know the details, but as far as I remember properties are not persisted across openHAB restarts for unmanaged things. This can easily be verified by testing it.

jlaur:
@andrewfg - I have some general questions which could help assess the strategies provided by this pull request.
- Do you know the actual restrictions for using the API? What exactly triggers the 429?
- When the limit is reached, do you know for how long requests will be rejected? If not, this could be binary searched through testing, but of course it will take some time.
- During the restricted period, will new requests count and extend the period, or will they simply be ignored? If not known, this can also be verified through testing. This information could be quite important for the chosen strategy.
- Is there any payload returned with the 429 that might contain additional information?

bundles/org.openhab.binding.gardena/src/main/java/org/openhab/binding/gardena/internal/handler/GardenaAccountHandler.java
andrewfg: Yes I am aware of the limitations of this. As far as I can tell there are several 'levels' of restart -- namely 1) disable/enable the binding, 2) reload the JAR, 3) restart OH, 4) reboot the PC. There seems to be no perfect way to persist something like this across all such 'levels'. The solution that I am using does work across some of these however..

andrewfg:
> general questions which could help assess the strategies provided by this pull request.

As background, see the API specification https://developer.husqvarnagroup.cloud/apis/gardena-smart-system-api#readme

> Do you know the actual restrictions for using the API?
Yes, these are defined in the chapter 'Rate Limits'. There are actually two limits, namely average calls per 15 minutes, and maximum calls per month.

> What exactly triggers the 429?
I suppose that normally, the server would trigger a 429 if the 'Rate Limits' are exceeded. However as we have seen, at the moment the server is buggy, so it is currently returning 429 errors on an apparently random basis.

> When the limit is reached, do you know for how long requests will be rejected?
Testing indicates that (when the server is running normally) once a 429 error is given, the account is blocked for 24 hours. 

> During the restricted period, will new requests count and extend the period, or will they simply be ignored?
Yes, new failed requests do seem to extend the period. Therefore it is better to wait for the full 24 hours before retrying. (However at the moment with the server being buggy, this is also not fully predicable).

> Is there any payload returned with the 429 that might contain additional information?
No.

bundles/org.openhab.binding.gardena/src/main/java/org/openhab/binding/gardena/internal/GardenaSmartImpl.java
```diff
@@ -318,9 +340,11 @@ public void dispose() {
         }
         httpClient.destroy();
         webSocketClient.destroy();
-        locationsResponse = new LocationsResponse();
+        LocationsResponse locationsResponse = this.locationsResponse;
+        if (locationsResponse != null && locationsResponse.data != null) {
+            locationsResponse.data.clear();
+        }
```
jlaur, line 346:
```suggestion
            this.locationsResponse = null;
        }
```

bundles/org.openhab.binding.gardena/src/main/java/org/openhab/binding/gardena/internal/GardenaSmartImpl.java
```diff
@@ -121,14 +125,22 @@ public GardenaSmartImpl(String id, GardenaConfig config, GardenaSmartEventListen
 
             // initially load access token
             verifyToken();
-            locationsResponse = loadLocations();
+            try {
+                locationsResponse = loadLocations();
+            } catch (GardenaException ex) {
+                locationsResponse = null;
+                throw ex;
+            }
```
jlaur, line 133: Simplification possible:
```suggestion
            LocationsResponse locationsResponse = loadLocations();
            this.locationsResponse = locationsResponse;
```

and then delete line 136 also (cannot comment across deleted lines). Or perhaps you could simply keep it as is, but without the try/catch. This is the constructor, so we know `locationsResponse` is currently null, thus no need to explicitly set this. The exception will even be caught in line 154, then `dispose()` is called and everything is cleared.

bundles/org.openhab.binding.gardena/src/main/java/org/openhab/binding/gardena/internal/GardenaSmartImpl.java
andrewfg: Ok.

bundles/org.openhab.binding.gardena/src/main/java/org/openhab/binding/gardena/internal/GardenaSmartImpl.java
andrewfg: Yeah. The problem is that the compiler cannot foresee an exception occurring in the constructor, so it gives spurious null annotation warnings..

bundles/org.openhab.binding.gardena/README.md
```diff
@@ -115,6 +115,17 @@ DateTime LastUpdate "LastUpdate [%1$td.%1$tm.%1$tY %1$tH:%1$tM]" { channel="gard
 openhab:send LastUpdate REFRESH
 ```
 
+### Server Call Rate Limitation
+
+The Gardena server imposes call rate limits to prevent malicious use of its API.
+The limits are a) not more than one call every 15 minutes, and b) not more that 3000 calls per month.
+Normally the openHAB binding does not exceed these limits.
+But from time to time the server may nevertheless consider the limits to have been exceeded.
+In which case it reports an HTTP 429 Error (Limit Exceeded).
```
maniac103, line 124: Nit: '...have been exceeded, in which case' ... it seems a bit weird to write this as two sentences?

bundles/org.openhab.binding.gardena/src/main/java/org/openhab/binding/gardena/internal/GardenaSmartImpl.java
```diff
@@ -318,9 +335,12 @@ public void dispose() {
         }
         httpClient.destroy();
         webSocketClient.destroy();
-        locationsResponse = new LocationsResponse();
+        LocationsResponse locationsResponse = this.locationsResponse;
+        if (locationsResponse != null && locationsResponse.data != null) {
+            locationsResponse.data.clear();
+        }
```
maniac103, line 341: What's the point of clearing that data member if the whole object is discarded below anyway?

bundles/org.openhab.binding.gardena/src/main/java/org/openhab/binding/gardena/internal/handler/GardenaAccountHandler.java
```diff
@@ -51,80 +66,189 @@
...
+                TimeUnit.SECONDS);
     }
 
     @Override
     public void dispose() {
         super.dispose();
+        synchronized (reInitializationCodeLock) {
+            ScheduledFuture<?> reInitializeTask = this.reInitializationTask;
+            if (reInitializeTask != null) {
+                reInitializeTask.cancel(false);
```
jlaur, line 248: Since this is dispose, I suppose we need to interrupt the task?
```suggestion
                reInitializeTask.cancel(true);
```

bundles/org.openhab.binding.gardena/src/main/java/org/openhab/binding/gardena/internal/handler/GardenaAccountHandler.java
```diff
@@ -140,6 +264,7 @@ private void disposeGardena() {
         final GardenaSmart gardenaSmart = this.gardenaSmart;
         if (gardenaSmart != null) {
             gardenaSmart.dispose();
+            this.gardenaSmart = null;
```
jlaur, line 267: Move below the `if` scope for consistency with pattern in line 247-250?

bundles/org.openhab.binding.gardena/src/main/java/org/openhab/binding/gardena/internal/handler/GardenaAccountHandler.java
```diff
@@ -51,80 +66,189 @@
...
     }
 
     public void setDiscoveryService(GardenaDeviceDiscoveryService discoveryService) {
         this.discoveryService = discoveryService;
     }
 
+    /**
+     * Format a localised description text to display on the thing main UI page.
+     *
+     * @return the (localised) description text.
```
jlaur, line 165:
```suggestion
     * @return the localized description text or null if call suppression is not active.
```
(couldn't help americani**z**ing while at it, but now actually in doubt about this project. American English is most common for international source code though)

bundles/org.openhab.binding.gardena/src/main/java/org/openhab/binding/gardena/internal/handler/GardenaAccountHandler.java
```diff
@@ -51,80 +66,189 @@
...
+            updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.COMMUNICATION_ERROR, getUiText());
+        }
     }
 
     public void setDiscoveryService(GardenaDeviceDiscoveryService discoveryService) {
         this.discoveryService = discoveryService;
     }
 
+    /**
+     * Format a localised description text to display on the thing main UI page.
```
jlaur, line 163:
```suggestion
     * Formats a localized explanatory description text regarding active call suppression.
```

jlaur (changes requested):
Thanks! Sorry for previously posting "fly by" disconnected comments probably giving impression of active review. Now went properly through your changes and performed a full review. I have posted some quite minor comments that should be easily fixable so we can finally get this merged soon.

bundles/org.openhab.binding.gardena/README.md
```diff
@@ -115,6 +115,17 @@ DateTime LastUpdate "LastUpdate [%1$td.%1$tm.%1$tY %1$tH:%1$tM]" { channel="gard
 openhab:send LastUpdate REFRESH
 ```
 
+### Server Call Rate Limitation
+
+The Gardena server imposes call rate limits to prevent malicious use of its API.
+The limits are a) not more than one call every 15 minutes, and b) not more that 3000 calls per month.
+Normally the openHAB binding does not exceed these limits.
```
jlaur, line 122:
```suggestion
Normally the binding does not exceed these limits.
```

bundles/org.openhab.binding.gardena/README.md
```diff
@@ -115,6 +115,17 @@ DateTime LastUpdate "LastUpdate [%1$td.%1$tm.%1$tY %1$tH:%1$tM]" { channel="gard
 openhab:send LastUpdate REFRESH
 ```
 
+### Server Call Rate Limitation
+
+The Gardena server imposes call rate limits to prevent malicious use of its API.
+The limits are a) not more than one call every 15 minutes, and b) not more that 3000 calls per month.
```
jlaur, line 121:
```suggestion
The limits are a) not more than one call every 15 minutes, and b) not more that 3.000 calls per month.
```
According to the documentation: "On average one call every fifteen minutes." - perhaps it should be mentioned that this is on average?

Btw, 3.000 calls is exactly one call every 15 minutes for a month with 31 days. That's kind of tight. :) So if you actually need four calls per hour, in practice limit will be hit after just one mistake (additional call). How many calls will the binding actually need continuously and on restart? This is just out of curiosity as I'm currently working on similar optimizations for the Bosch Indego binding.

jlaur: Hmm, maybe even just format as in the documentation:
```suggestion
The limits are:

- On average one call every fifteen minutes.
- 3.000 requests per month.
```

jlaur:
@andrewfg - last, the usual request to optimize the PR title for release notes. :)
- **Eliminate dangling references:** Hardly understandable for users, I would suggest simply removing it as this is mostly technical refactoring. I know this is what started the PR, but since the PR span across two subjects now, this kind of feels secondary and something additional to the rate limit improvements?
- **Improve HTTP 429 handling:** Users having experienced the rate limits will probably recognize "HTTP 429" and connect the dots, but probably "HTTP 429" could be substituted with "rate limit".

bundles/org.openhab.binding.gardena/README.md
andrewfg: Done.

bundles/org.openhab.binding.gardena/README.md
```diff
@@ -115,6 +115,17 @@ DateTime LastUpdate "LastUpdate [%1$td.%1$tm.%1$tY %1$tH:%1$tM]" { channel="gard
...
+Normally the openHAB binding does not exceed these limits.
+But from time to time the server may nevertheless consider the limits to have been exceeded.
+In which case it reports an HTTP 429 Error (Limit Exceeded).
+If such an error occurs your Gardena account will be locked out for 24 hours.
```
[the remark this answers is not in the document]
andrewfg: Done.

bundles/org.openhab.binding.gardena/README.md
andrewfg: >  3.000 calls is exactly one call every 15 minutes for a month with 31 days

24 * 4 * 31 = 2976 .. so actually you are allowed 24 extra mistakes a month :)

In my experience, the limit will already have been triggered on the 15 minute rule, before you would ever hit the 3'000 rule. And also the lockout is lifted after 24 hours, so I am not sure how the 3'000 would every actually apply. It's a mystery..

bundles/org.openhab.binding.gardena/README.md
andrewfg: Done

bundles/org.openhab.binding.gardena/src/main/java/org/openhab/binding/gardena/internal/handler/GardenaAccountHandler.java
andrewfg: George Bernard Shaw said: "England and America are two countries separated by a common language" :)

PS the spell checker in my Eclipse IDE apparently speaks British English :)

image

andrewfg: Done.

bundles/org.openhab.binding.gardena/README.md
jlaur: You are right, 3.000 was a mistake based on Danish culture. :grinning: In US it would be 3,000. I never saw ' used as thousand separator before, so maybe just plain 3000?

You calculation is also correct, of course. I was mistakenly using Windows calculator in programmer mode from a previous session, so it applied rounding, lol.

So in practice, do you need just one call on binding start, and then all updates (including all information you are interested in) after that are through WebSocket?

bundles/org.openhab.binding.gardena/src/main/java/org/openhab/binding/gardena/internal/handler/GardenaAccountHandler.java
andrewfg: Done.

bundles/org.openhab.binding.gardena/src/main/java/org/openhab/binding/gardena/internal/handler/GardenaAccountHandler.java
andrewfg: Done.

bundles/org.openhab.binding.gardena/src/main/java/org/openhab/binding/gardena/internal/handler/GardenaAccountHandler.java
```diff
@@ -51,80 +66,189 @@
...
+        apiCallSuppressionUntil = until;
     }
 
     @Override
     public void initialize() {
         logger.debug("Initializing Gardena account '{}'", getThing().getUID().getId());
-        initializeGardena();
+        loadApiCallSuppressionUntil();
+        Duration delay = apiCallSuppressionDelay();
+        if (Duration.ZERO.equals(delay)) {
```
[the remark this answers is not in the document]
andrewfg: Done.

bundles/org.openhab.binding.gardena/src/main/java/org/openhab/binding/gardena/internal/GardenaSmartImpl.java
andrewfg: I changed the code, but your comments on inner workings of Java garbage collector would still be appreciated.

andrewfg:
>  request to optimize the PR title for release notes.

Done.

bundles/org.openhab.binding.gardena/src/main/java/org/openhab/binding/gardena/internal/handler/GardenaAccountHandler.java
jlaur: Almost forgot about this one. I was suggesting to use a "@text/" reference for `updateStatus()` as it directly supports it. I was not suggesting a one-liner or to introduce redundancy. But I'll still take the challenge and write the code to express my suggestion, will post as separate suggestion.

bundles/org.openhab.binding.gardena/src/main/java/org/openhab/binding/gardena/internal/handler/GardenaAccountHandler.java
```diff
@@ -51,80 +66,189 @@
...
+    private @Nullable String getUiText() {
+        Instant until = apiCallSuppressionUntil;
+        if (until != null) {
+            ZoneId zone = timeZoneProvider.getTimeZone();
+            boolean isToday = LocalDate.now(zone).equals(LocalDate.ofInstant(until, zone));
+            DateTimeFormatter formatter = isToday ? DateTimeFormatter.ofLocalizedTime(FormatStyle.MEDIUM)
+                    : DateTimeFormatter.ofLocalizedDateTime(FormatStyle.MEDIUM);
+            String value = i18nProvider.getText(bundle, RECONNECT_MSG_KEY, null, localeProvider.getLocale(),
+                    formatter.format(ZonedDateTime.ofInstant(until, zone)));
+            return value;
```
jlaur, line 176: This should work:
```suggestion
            return "@text/accounthandler.waiting-until-to-reconnect [\"" + formatter.format(ZonedDateTime.ofInstant(until, zone)) + "\"]";
                    );
```

bundles/org.openhab.binding.gardena/README.md
andrewfg: > just one call on binding start

Actually no, its is a bit more complicated than that..
1. On binding start, you consume one call to fetch the locations
2. On connection you consume one call to establish a WebSocket for each location
3. Thereafter all status notifications are done within the WebSocket (no more calls are made)
4. The server drops the WebSocket after 2 hours or when the Certificate Expires (24 hours), plus any drops due to network errors.
5. If a WebSocket does drop go to 2. above
6. If the Thing is dis-/re-enabled (including swapping a JAR or restarting OH) go to 1. above.
7. To issue commands (e.g. switch on/off the watering) you consume one more call to issue each command.

In practice, in stable operation, this is well within the rate limits. Even if issuing watering on/off commands.

But if you are testing the system with system restarts, JAR file reloads, and Thing dis-/re-enabling it could be much more than that; and any lock out is only cleared after 24 hours; so I needed a lot of patience during developing and testing this one :)

bundles/org.openhab.binding.gardena/src/main/java/org/openhab/binding/gardena/internal/handler/GardenaAccountHandler.java
andrewfg: So I don't need to capture the localProvider or the i18nProvider in the thing factory component, nor propagate them into the created Things? Or ??

jlaur: There would be no more usages of them, so correct. If you would like to eliminate them, please verify my suggestion. The magic happens within `updateStatus()`, so we just need to provide a properly formatted string for it.

andrewfg: It may actually be quite difficult to test your localization magic code because you would have to have a Gardena account, and it would have to be in lockout mode. ... ...

jlaur: It the variable substitution works correctly in English, I have no doubts that it will also work for other languages. This happens completely outside your code now, so as an additional advantage, there is less to test. :)

andrewfg: PS I grepped the addons folder for the following Regex and found about a dozen classes where your syntax is used. So even if we cannot test it completely, we are in good company using this code form. :)
```
(?s)"@text/.*\[\\"".*"\\"\]
```

jlaur (approved):
LGTM

bundles/org.openhab.binding.gardena/README.md
jlaur: So best case: one call per hour + one call per day. Additionally, and hard to control, two calls per restart/network interruption and one call per command. So at any given time, above 25% of the allowed calls simply because required, and commands can quickly influence the average allowed, depending on how that is calculated. Pretty strict.

PR #13018, opened 2022-06-26
GitHub Actions workflow improvements

wborn:
* Upgrade actions to latest versions
* Upgrade Maven to 3.8.6
* Use Temurin JDK instead of Zulu JDK

The Temurin JDK is preinstalled in the [Ubuntu image](https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu2004-Readme.md#java) whereas the Zulu JDK first need to be downloaded.
It should also perform better because the Temurin JDK is linked to glibc 2.12 whereas the Zulu JDK is linked to glibc 2.5.

### DIFF
--- a/.github/workflows/ci-build.yml
+++ b/.github/workflows/ci-build.yml
@@ -19,23 +19,23 @@ jobs:
       fail-fast: false
       matrix:
         java: [ '11', '17' ]
-        maven: [ '3.8.4']
+        maven: [ '3.8.6' ]
         os: [ 'ubuntu-20.04' ]
     name: Build (Java ${{ matrix.java }}, ${{ matrix.os }})
     runs-on: ${{ matrix.os }}
     steps:
       - name: Checkout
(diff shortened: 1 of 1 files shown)


J-N-K:
I‘m not sure we should switch the JDK. Since we recommend Zulu, we should also run out CI builds with Zulu.

wborn:
> Since we recommend Zulu

I think that advice also needs an update as we ship Temurin with both openHABian and the Docker containers nowadays. It's also used in the Jenkins builds.

I'd personally only use Zulu on outdated distro's with an older glibc version where installing Temurin is not an option. Temurin is also [TCK](https://en.wikipedia.org/wiki/Technology_Compatibility_Kit) tested and you can [get support for it by Azul](https://www.azul.com/blog/eclipse-temurin-new-distribution-of-openjdk/).

jlaur (approved):
Thanks!

PR #13019, opened 2022-06-26
[knx] Allow all port names for serial interfaces

holgerfriedrich:
Starting with openHAB 3.3, the framework enforces using serial
ports which are detected by the system by default. This feature
is now disabled for the KNX addon to allow using symlinks to
ports as well.
Refers-to: openhab/openhab-core#3014

### DIFF
--- a/bundles/org.openhab.binding.knx/src/main/resources/OH-INF/thing/serial.xml
+++ b/bundles/org.openhab.binding.knx/src/main/resources/OH-INF/thing/serial.xml
@@ -10,6 +10,7 @@
 		<config-description>
 			<parameter name="serialPort" type="text" required="true">
 				<context>serial-port </context>
+				<limitToOptions>false</limitToOptions>
 				<label>Serial Port</label>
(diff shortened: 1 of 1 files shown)


J-N-K (approved):

wborn (approved):
Thanks!

wborn:
The KNX binding doesn't use the OHC serial transport yet (https://github.com/openhab/openhab-addons/issues/7573). So you probably need to set `gnu.io.rxtx.SerialPorts` for it to work. Which is otherwise unneccessary when the OHC serial transport is used, because it automatically does that for you. :slightly_smiling_face:

holgerfriedrich:
@wborn this is the behaviour as in OH 3.2, right?
And yes it is still gnu IO, porting is not done yet. I cannot test, as I do not have a serial device at hand. Calimero lib has some changes in the unreleased main line in this area.

wborn:
Yes it's still the same as with OH 3.2. I just mentioned it for whenever users cannot get their udev rule based ports working with the KNX Binding.

PR #13020, opened 2022-06-26
[zoneminder] Fix permanent alarm in monitors

stage-rl:
Fixes a problem at least in Debian, version 1.36.17. Without this change all monitors are permanently reported as alarming. At least works for me.

### DIFF
--- a/bundles/org.openhab.binding.zoneminder/src/main/java/org/openhab/binding/zoneminder/internal/handler/Monitor.java
+++ b/bundles/org.openhab.binding.zoneminder/src/main/java/org/openhab/binding/zoneminder/internal/handler/Monitor.java
@@ -101,7 +101,7 @@ public class Monitor {
     }
 
     public void setState(MonitorState state) {
-        this.alarm = (state != MonitorState.IDLE && state != MonitorState.UNKNOWN);
+        this.alarm = (state != MonitorState.IDLE && state != MonitorState.UNKNOWN && state != MonitorState.PREALERT);
         this.state = state;
     }
 


mhilbush:
I'm confused. It looks like a ZM change was committed on Dec 10 2021 to add the `UNKNOWN` state as state 0, which shifted the numbering of all the remaining states.

https://github.com/ZoneMinder/zoneminder/commit/aa320cb28fda67ae8a2fe1a676cbe51781d023cf

So...
IDLE is now 1 (was 0)
PREALARM is now 2 (was 1)
ALARM is now 3 (was 2)
etc.

But I don't see which ZM release included this commit.

Has this been broken in the binding since December?

stage-rl:
Well, I'm surely not a coder and very new to github, so pls doublecheck me.

I found it here:
https://github.com/ZoneMinder/zoneminder/pull/3462/commits

So April 6th, that would make it to 1.36.15 version, and that goes to Debian testing on May 20th.

Which I believe could be around the date when the issue started.

stage-rl:
But that's in the 1.37.x Development Branch subforum.

mhilbush:
Yeah, I saw that.

That PR you referenced above was never merged as it was a badly constructed PR (it contained many other commits other than the author's intended changes).

I looked through the change log for 1.36.15 (in fact I looked through the change logs for all the releases since December), but I didn't see that commit in any of the change logs.

mhilbush:
I just pulled down the source for 1.36.19. That change is not in there. So it must be something else that's causing the issue.

Flole998:
Your commit is not correctly signed-off. That needs to be addressed before this can be merged.

mhilbush:
I opened the following ZoneMinder issue. This PR should not be merged until we understand the rationale for this potentially breaking change to the API.

https://github.com/ZoneMinder/zoneminder/issues/3584

connortechnology:
Apologies.  It's a mess.  The API should not change. However I also think we shouldn't be return integer  values for status.  For now I am going to fix that API by adjusting the value to pre 1.36.16 values.  I think in 1.37 we may shift to returning a string value.  Thoughts?

mhilbush:
@connortechnology Yes, returning string values instead of numbers would be ideal. It resolves the issue of the numbering getting inadvertently changed, as well as the string values being more descriptive. I can handle the breaking change on my end for users who are on older ZoneMinder versions that still use numbers.

Thanks for the fix!!

@stage-rl I'm closing this PR as a fix will be included in ZoneMinder 1.36.26.

stage-rl:
OK, thank you both for all the work. Agree with string statuses. Anyway I'm just a plain user affected by this discrepancy, so leaving the real work for PROs ;-)

PR #13030, opened 2022-06-27
[insteon] set network to ONLINE after driver is initialized

robnielsen:
Prior to this change the Insteon network was set to ONLINE before it should have been.

### DIFF
--- a/bundles/org.openhab.binding.insteon/src/main/java/org/openhab/binding/insteon/internal/handler/InsteonNetworkHandler.java
+++ b/bundles/org.openhab.binding.insteon/src/main/java/org/openhab/binding/insteon/internal/handler/InsteonNetworkHandler.java
@@ -108,8 +108,6 @@ public class InsteonNetworkHandler extends BaseBridgeHandler {
 
                         // wait until driver is initialized before setting network to ONLINE
                         driverInitializedJob = scheduler.scheduleWithFixedDelay(() -> {
-                            logger.debug("driver is not initialized yet");
-
                             if (insteonBinding.isDriverInitialized()) {
                                 logger.debug("driver is initialized");
 
@@ -119,9 +117,11 @@ public class InsteonNetworkHandler extends BaseBridgeHandler {
 
                                 ScheduledFuture<?> driverInitializedJob = this.driverInitializedJob;
                                 if (driverInitializedJob != null) {
-                                    driverInitializedJob.cancel(false);
+                                    driverInitializedJob.cancel(true);
                                 }
                                 this.driverInitializedJob = null;
+                            } else {
+                                logger.debug("driver is not initialized yet");
                             }
                         }, 0, DRIVER_INITIALIZED_TIME_IN_SECONDS, TimeUnit.SECONDS);
                     } else {
--- a/bundles/org.openhab.binding.insteon/src/main/java/org/openhab/binding/insteon/internal/handler/InsteonNetworkHandler.java
+++ b/bundles/org.openhab.binding.insteon/src/main/java/org/openhab/binding/insteon/internal/handler/InsteonNetworkHandler.java
@@ -80,66 +80,64 @@ public class InsteonNetworkHandler extends BaseBridgeHandler {
         logger.debug("Starting Insteon bridge");
         InsteonNetworkConfiguration config = getConfigAs(InsteonNetworkConfiguration.class);
 
-        scheduler.execute(() -> {
-            SerialPortManager serialPortManager = this.serialPortManager;
-            if (serialPortManager == null) {
-                String msg = "Initialization failed, serial port manager is null.";
-                logger.warn(msg);
+        SerialPortManager serialPortManager = this.serialPortManager;
+        if (serialPortManager == null) {
+            String msg = "Initialization failed, serial port manager is null.";
+            logger.warn(msg);
 
-                updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.CONFIGURATION_ERROR, msg);
+            updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.CONFIGURATION_ERROR, msg);
 
-                return;
-            }
-            insteonBinding = new InsteonBinding(this, config, serialPortManager, scheduler);
-            updateStatus(ThingStatus.UNKNOWN);
+            return;
+        }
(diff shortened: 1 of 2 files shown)


bundles/org.openhab.binding.insteon/src/main/java/org/openhab/binding/insteon/internal/handler/InsteonNetworkHandler.java
```diff
@@ -104,9 +106,24 @@ public void initialize() {
                             insteonBinding.logDeviceStatistics();
                         }, 0, LOG_DEVICE_STATISTICS_DELAY_IN_SECONDS, TimeUnit.SECONDS);
 
-                        insteonBinding.setIsActive(true);
+                        // wait until driver is initialized before setting network to ONLINE
+                        driverInitializedJob = scheduler.scheduleWithFixedDelay(() -> {
+                            logger.debug("driver is not initialized yet");
```
lolodomo, line 111: This log could be wrong.
You should move it in the else branch of your next if.
Or this is the message which is not appropriate.

bundles/org.openhab.binding.insteon/src/main/java/org/openhab/binding/insteon/internal/handler/InsteonNetworkHandler.java
```diff
@@ -129,6 +146,12 @@ public void initialize() {
     public void dispose() {
         logger.debug("Shutting down Insteon bridge");
 
+        ScheduledFuture<?> driverInitializedJob = this.driverInitializedJob;
+        if (driverInitializedJob != null) {
+            driverInitializedJob.cancel(false);
```
lolodomo, line 151: You should rather use true as parameter.

bundles/org.openhab.binding.insteon/src/main/java/org/openhab/binding/insteon/internal/handler/InsteonNetworkHandler.java
```diff
@@ -104,9 +106,24 @@ public void initialize() {
                             insteonBinding.logDeviceStatistics();
                         }, 0, LOG_DEVICE_STATISTICS_DELAY_IN_SECONDS, TimeUnit.SECONDS);
```
lolodomo, line 108: I checked the code of the `initialize` method with all these nested asynchronous calls. For me, the higher `scheduler.execute` should be simply removed, it will allow `initialze` setting a thing status before returning, which is what is expected by the framework.

lolodomo (changes requested):

bundles/org.openhab.binding.insteon/src/main/java/org/openhab/binding/insteon/internal/handler/InsteonNetworkHandler.java
robnielsen: moved it

bundles/org.openhab.binding.insteon/src/main/java/org/openhab/binding/insteon/internal/handler/InsteonNetworkHandler.java
robnielsen: fixed

bundles/org.openhab.binding.insteon/src/main/java/org/openhab/binding/insteon/internal/handler/InsteonNetworkHandler.java
robnielsen: OK, I removed it, but it was originally done this way because Kai asked me to do it that way.

robnielsen:
@lolodomo, I made the changes you suggested. Can you please approve and merge this PR?

bundles/org.openhab.binding.insteon/src/main/java/org/openhab/binding/insteon/internal/handler/InsteonNetworkHandler.java
```diff
@@ -78,57 +80,76 @@ public void initialize() {
...
+                        if (insteonBinding.isDriverInitialized()) {
+                            logger.debug("driver is initialized");
+
+                            insteonBinding.setIsActive(true);
+
+                            updateStatus(ThingStatus.ONLINE);
+
+                            ScheduledFuture<?> driverInitializedJob = this.driverInitializedJob;
+                            if (driverInitializedJob != null) {
+                                driverInitializedJob.cancel(true);
```
lolodomo, line 119: false should be better here to not interrupt the current running thread ...

bundles/org.openhab.binding.insteon/src/main/java/org/openhab/binding/insteon/internal/handler/InsteonNetworkHandler.java
```diff
@@ -78,57 +80,76 @@ public void initialize() {
...
+        }, SETTLE_TIME_IN_SECONDS, 1, TimeUnit.SECONDS);
     }
 
     @Override
     public void dispose() {
         logger.debug("Shutting down Insteon bridge");
 
+        ScheduledFuture<?> driverInitializedJob = this.driverInitializedJob;
+        if (driverInitializedJob != null) {
+            driverInitializedJob.cancel(false);
```
lolodomo, line 149: ... but true should be better here.

lolodomo (changes requested):

bundles/org.openhab.binding.insteon/src/main/java/org/openhab/binding/insteon/internal/handler/InsteonNetworkHandler.java
robnielsen: changed now, sorry

bundles/org.openhab.binding.insteon/src/main/java/org/openhab/binding/insteon/internal/handler/InsteonNetworkHandler.java
robnielsen: changed.

robnielsen:
@lolodomo, should be good now

lolodomo (approved):
LGTM

PR #13042, opened 2022-06-30
[googletts] Fix authorization from returning null refresh token

mhilbush:

Related to #12711 

This binding suffers from the same issue as the nest binding in that any authorization after the first one causes the Google API to return no refresh token. The net effect is that the binding will work until it needs to refresh the access token (usually after approximately 1 hour).


mhilbush:
Good catch. I had no idea all that was in the config description.

wborn:
>  I have not changed the de and fr translation files?

I think it is better to do it via Crowdin otherwise they will be overwritten the next time a string is changed. After the PR is merged Crowdin will automatically mark the changed strings in all other languages as untranslated.

mhilbush:
Ok, I think I got it this time. 🤞

wborn (approved):
Thank you!

PR #13043, opened 2022-06-30
[googlestt] Fix authorization from returning null refresh token

mhilbush:

Related to #12711 
Related to #13042

This binding suffers from the same issue as the nest and Google TTS bindings in that any authorization after the first one causes the Google API to return no refresh token. The net effect is that the binding will work until it needs to refresh the access token (usually after approximately 1 hour).


wborn (approved):
Thanks again!

PR #13046, opened 2022-07-01
[venstarthermostat] Split away and schedule settings calls (resolves bug exiting away mode)

dae1804:
My ColorTouch thermostat has a newer firmware, which, from the OpenHab logs & experimenting with the REST API directly, I've observed to reject any POSTs to the settings endpoint that include the `schedule` parameter when the thermostat is currently in AWAY mode, regardless of the value passed for `schedule`.  This had the effect of preventing the OpenHab binding from un-setting away mode.

This is likely to be the same problem encountered by other users on this forum thread: https://community.openhab.org/t/venstar-binding-not-working/123780/4

This PR splits out the method used to perform thermostat settings updates, so that _only_ the value being changed is sent to the thermostat.  By avoiding the `schedule` parameter from being included at all when changing `away`, this ensures that the thermostat will not reject attempts to change away mode from away to home.

### DIFF
--- a/bundles/org.openhab.binding.venstarthermostat/src/main/java/org/openhab/binding/venstarthermostat/internal/handler/VenstarThermostatHandler.java
+++ b/bundles/org.openhab.binding.venstarthermostat/src/main/java/org/openhab/binding/venstarthermostat/internal/handler/VenstarThermostatHandler.java
@@ -382,13 +382,42 @@ public class VenstarThermostatHandler extends ConfigStatusThingHandler {
     }
 
     private void setAwayMode(VenstarAwayMode away) {
-        VenstarScheduleMode schedule = infoData.getScheduleMode();
-        updateSettings(away, schedule);
+        // This function updates the away mode via a POST to the thermostat's local API's /settings endpoint.
+        //
+        // The /settings endpoint supports a number of additional parameters (tempunits, de/humedifier
+        // setpoints, etc). However, newer Venstar firmwares will reject any POST to /settings that
+        // contains a `schedule` parameter when the thermostat is currently in away mode.
+        //
+        // Separating the updates to change `schedule` and `away` ensures that the thermostat will not
+        // reject attempts to un-set away mode due to the presence of the `schedule` parameter.
+        Map<String, String> params = new HashMap<>();
+        params.put("away", String.valueOf(away.mode()));
+        VenstarResponse res = updateThermostat("/settings", params);
+        if (res != null) {
+            log.debug("Updated thermostat");
+            // update our local copy until the next refresh occurs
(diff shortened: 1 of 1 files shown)


digitaldan:
Thanks for the contribution!  I'll take a look at it over the weekend.

bundles/org.openhab.binding.venstarthermostat/src/main/java/org/openhab/binding/venstarthermostat/internal/handler/VenstarThermostatHandler.java
```diff
@@ -382,13 +382,11 @@ private void setFanMode(VenstarFanMode fanmode) {
     }
 
     private void setAwayMode(VenstarAwayMode away) {
-        VenstarScheduleMode schedule = infoData.getScheduleMode();
-        updateSettings(away, schedule);
+        updateAwayMode(away);
```
digitaldan, line 385: Rather then call updateAwayMode, just put the logic from it in this function, no need to have 2

bundles/org.openhab.binding.venstarthermostat/src/main/java/org/openhab/binding/venstarthermostat/internal/handler/VenstarThermostatHandler.java
```diff
@@ -382,13 +382,11 @@ private void setFanMode(VenstarFanMode fanmode) {
...
     }
 
     private void setScheduleMode(VenstarScheduleMode schedule) {
-        VenstarAwayMode away = infoData.getAwayMode();
-        updateSettings(away, schedule);
+        updateScheduleMode(schedule);
```
digitaldan, line 389: Rather then call updateScheduleMode, just put the logic from it in this function, no need to have 2

digitaldan:
Hi @dae1804 thanks for the fix!  That was a tough one to track down, i appreciate you figuring it out and submitting a PR.  I had two comments about condensing the 4 functions into 2, but other then that i tested it out and it looks good so far.

digitaldan:
Hi @dae1804 , any progress?

dae1804:
Hi @digitaldan!  Apologies for the delay; I was out of town last week, and just got back late Sunday.  I'm still digging out of backlog, but should have this turned around this week, perhaps tomorrow.

dae1804:
Hi @digitaldan!  Finally found a few minutes to get this turned around...  Methods merged, and I updated the comment as well.

digitaldan (approved):
LGTM, thanks!!!!

jlaur (approved):
Thanks! LGTM.

PR #13040, opened 2022-06-29
[boschindego] Add battery and garden channels (including map)

jlaur:
Introduce new channels:
- batteryVoltage
- batteryLevel
- lowBattery
- batteryTemperature
- gardenSize
- gardenMap

Binding is now also able to detect whether device is online, so it will be marked as offline when unreachable. This is based on **/operatingData** which will directly communicate with the device, whereas **/state** returns data even when device has been turned off for several days.

Fixes #12938
Fixes #13017

JAR: https://drive.google.com/file/d/1pWVrOl-QcKf3zeHuqizhV1VDpXecxizC/view?usp=sharing

PR #13058, opened 2022-07-02
[boschindego] Add missing specialized text for unreachable device

jlaur:
Small fix for #13040 as specialized text for unreachable device was missing, so it would should "Gateway timeout" where it should show "Device is unreachable".

### DIFF
--- a/bundles/org.openhab.binding.boschindego/src/main/java/org/openhab/binding/boschindego/internal/IndegoController.java
+++ b/bundles/org.openhab.binding.boschindego/src/main/java/org/openhab/binding/boschindego/internal/IndegoController.java
@@ -174,10 +174,11 @@ public class IndegoController {
      * @param dtoClass the DTO class to which the JSON result should be deserialized
      * @return the deserialized DTO from the JSON response
      * @throws IndegoAuthenticationException if request was rejected as unauthorized
+     * @throws IndegoUnreachableException if device cannot be reached (gateway timeout error)
      * @throws IndegoException if any communication or parsing error occurred
      */
     private <T> T getRequestWithAuthentication(String path, Class<? extends T> dtoClass)
-            throws IndegoAuthenticationException, IndegoException {
+            throws IndegoAuthenticationException, IndegoUnreachableException, IndegoException {
         if (!session.isValid()) {
             authenticate();
         }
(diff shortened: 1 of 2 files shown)


lolodomo:
To be clean, shouldn't you mention `IndegoUnreachableException` in each method that can throw this exception and also in the javadoc of these methods ?

jlaur:
> To be clean, shouldn't you mention `IndegoUnreachableException` in each method that can throw this exception and also in the javadoc of these methods ?

I have documented it on relevant paths now.

jlaur:
@BillGOH - latest version here, also including the SpotMow status code: https://drive.google.com/file/d/1caMAEgdSEAMYGJ0xgoYGp8al9gDsm5GW/view?usp=sharing

lolodomo:
What about getState, getCalendar, getWeather, ...

jlaur:
> What about getState, getCalendar, getWeather, ...

I only mentioned it on relevant paths, where it's expected to happen, i.e. `getOperatingData()`. In any case it's sufficient to catch `IndegoException` from which `IndegoUnreachableException` is derived, but when mentioned it should be considered whether `IndegoUnreachableException` should be handled differently.

It's already a bit too much IMHO, but on the other hand it doesn't hurt to expose it in Javadoc for considering more fine-grained exception handling. In standard Java libraries you will often see only the superclass documented, for example [IOException](https://docs.oracle.com/javase/7/docs/api/java/io/IOException.html). I'm not sure about the distinction here, but in specific scenarios the direct subclass is also documented. Perhaps when there is only one or a few possible types thrown? For example, it's documented that the constructor of `FileInputStream` can throw `FileNotFoundException`, but in this case I think no other IOExceptions are possible.

jlaur:
> I only mentioned it on relevant paths, where it's expected to happen, i.e. `getOperatingData()`. In any case it's sufficient to catch `IndegoException` from which `IndegoUnreachableException` is derived, but when mentioned it should be considered whether `IndegoUnreachableException` should be handled differently.

Just to add some more context: `getOperatingData()` is the only method which will directly contact the device, so only this method is subject to gateway timeout when doing so. Other methods will be handled directly by Bosch services, for example `getState()` will respond nicely even when the device has been turned off for several days. `getWeather()` has nothing specific to do with the device, except that's in context of the device's location. If other methods will unexpectedly cause gateway timeout, it should not be handled as device not being reachable.

lolodomo:
Ok, my fault, I did not notice the unreachable exception was derived from the other. So no risk to miss this exception.

lolodomo (approved):
LGTM, thank you

Issue #13041, opened 2022-06-30
[goecharger] Add API V2 support nrg datatype problem


sittner:
## Expected Behavior
Binding should parse the JSON response from go-e charger correctly.

## Current Behavior
JSON parser produced a NumberFormatException while parsing the "nrg" array.

## Possible Solution
Change the energy property (GoEStatusResponseBaseDTO.java) from Integer to Double.
Please see https://github.com/sittner/openhab-addons/commit/20fae22ad4f7a4502c80ad25cb68932746e7f099 for reference.

The scaling has to be fixed as well, but I'm not sure if this breaks compatibility with older API v2 firmware (tested only with  053.3).

## Steps to Reproduce
1. Install firmware 053.3
2. Install/Setup goecharger binding to communicate with that charger
3. NumberFormatException will appear in openhab.log

## Context

Firmware 053.3 transmits the the "nrg" array as double values. I'm not sure, if previous API v2 versions are affected as well.

Example:
```
{"alw":true,"dwo":null,"eto":49670,"amp":16,"frc":0,"car":2,"err":0,"cbl":32,"wh":16024.18,"fwv":"053.3","tma":[51.25,43.75],"nrg":[228,231,233,1,15.3,0,0,3500,0,0,0,3600,100,1,0,78],"psm":2}
```

## Your Environment
* Version used: openhab and openhab-addons 3.3.0
* go-e Charger version: HW 3, FW 053.3
* Operating System and version: Debian Bullseye (AMD64)

PR #13052, opened 2022-07-01
[goecharger] fix data type for energy (nrg)

ranzen84:
energy (nrg) need to be double for V2
(tested with version 053.4 (Beta)

Signed-off-by: Reinhard Plaim <reinhardplaim@gmail.com>

PR #13068, opened 2022-07-03
[goecharger] Change field nrg from Integer to Double

cribskip:
Probably fixing error "java.lang.NumberFormatException: Expected an int but was 16.1 at line 1 column 154 path $.nrg[4]"
Issue #13041 "[goecharger] Add API V2 support nrg datatype problem"


SamuelBrucksch:
Maybe related to https://github.com/openhab/openhab-addons/pull/13052 ?

cribskip:
@SamuelBrucksch thank you for the link - I didn't see this one.

Seems I'm a little late to the party and the issue should be resolved in the PR you mentioned (y)

SamuelBrucksch:
Thanks for trying to help out. Maybe you can help test / review the other PR then? It's always good to confirm it by someone who has the device.

PR #13071, opened 2022-07-03
[miele] Fix auto-update for stop channel and synchronize switch channel with appliance state

### LINKED ISSUE #13069: [miele] Fix stateless stop channel and synchronize switch channel from state
Channel **stop** should be changed to auto-update policy **veto** since it's a stateless channel with the purpose of triggering a stop action (for oven and hood). This will eliminate the need for configuring `autoupdate="false"` on the item.

Channel **switch** is used either for triggering start/stop actions or actually switching appliance on or off. Since it has this duality, it's by definition not stateless. However, the state is not initialized, so when an appliance is actually turned on, the channel might be set to false, depending on last manual value change. This should be fixed so that current state is used to set state correctly.

Additional notes:
- For my oven, H5581BP, the **switchOn**/**switchOff** methods doesn't actually work. Probably **switchOn** will not work for any oven due to Miele security policy, but **switchOff** could potentially work on other models. If not, there is no reason for having the channel at all. The **stop** method turns my oven off (as opposed to just stopping running program). This method as exposed separately in channel **stop**.
- The **switchOn**/**switchOff** methods are also supported by coffee machines, but I don't know if they actually work. @PaulchenP, perhaps you can verify if you can turn your coffee machine on and/or off using the switch channel?
- Dishwasher/washing machine/tumble dryer implements the switch channel as **start**/**stop** methods (as opposed to **switchOn**/**switchOff**).
- For tumble dryer I have verified that **start**/**stop** works (exposed as **switch**): **start** can be used to start immediately when already programmed to start later. **stop** will interrupt running program and start cooling phase. An additional **stop** will interrupt the cooling phase also and completely stop the machine. So states where it makes sense to update switch to off: `STATE_WAITING_TO_START` (because on will then start), `STATE_PAUSED`. And to on: `STATE_RUNNING` (because off will then stop). For all other states it doesn't really matter as trigger will not work.

jlaur:
Improve switch channels **stop** and **switch**:
- **stop:** Fix auto-update policy as this channel is stateless and used only to trigger a stop action.
- **switch:** Reflect appliance state and by that eliminate the need for manually switching on before being able to switch off, and vice versa.

Fixes #13069

### DIFF
--- a/bundles/org.openhab.binding.miele/README.md
+++ b/bundles/org.openhab.binding.miele/README.md
@@ -448,7 +448,7 @@ DateTime Oven_ElapsedTime "Elapsed time" <time>             {channel="miele:oven
 DateTime Oven_FinishTime "Remaining time" <time>            {channel="miele:oven:home:oven:finish"}
 Number:Temperature Oven_CurrentTemperature <temperature>    {channel="miele:oven:home:oven:measured"}
 Number:Temperature Oven_TargetTemperature <temperature>     {channel="miele:oven:home:oven:target"}
-Switch Oven_Stop                                            {channel="miele:oven:home:oven:stop", autoupdate="false"}
+Switch Oven_Stop                                            {channel="miele:oven:home:oven:stop"}
 
 String WashingMachine_State                                 {channel="miele:washingmachine:home:washingmachine:state"}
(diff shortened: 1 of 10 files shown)


lolodomo (approved):
LGTM

PR #13073, opened 2022-07-03
[iaqualink] Fix to work with recent API changes to the iAqualink cloud service

digitaldan:
Currently the binding throws a gson parsing error early in the poll process rendering it completely broken.


jlaur (approved):
Thanks!

PR #13076, opened 2022-07-04
[dsmr] Fix incorrect deriving of sub channel names when updating channels

Hilbrand:
Subchannels were created by appending the key. However this was done inside a loop and if multiple updates were needed they would be appended to the channel name, instead of taking the channel name each time.

Reported on https://community.openhab.org/t/no-power-failure-logs-in-dsmr/137116


jlaur (approved):
Thanks for the fix!

PR #13078, opened 2022-07-04
[openhabcloud] Makes the baseURL and item updates options advanced...

### LINKED ISSUE #12582: [openhabcloud] Move "Items to Expose" option to advanced settings
Although support for IFTTT was removed from myopenHAB.org long ago, there's still an option to expose items in MainUI under the openHab Cloud settings:

image

I assume this is to allow people running their own cloud servers to use IFTTT. Unfortunately, it causes confusion for many users who assume it's required for Alexa and Google Assistant integration (which is quite reasonable).

I suggest:

1. Having "Items to expose" only appear when you click "Show advanced".
2. Renaming it "Items to expose to IFTTT"
3. Updating the caption: "If you've set up a private cloud server with IFTTT integration, use this option to indicate which items can be controlled with IFTTT. This setting does not work with myopenHAB.org and is not required for integration with Amazon Alexa or Google Assistant."

Alternatively, we could just remove IFTTT support. However, that would be detrimental to anyone who has set up a cloud server specifically to keep IFTTT working with openHAB.

Thanks for you work!

digitaldan:
also cleans up documentation and disables sending updates to the community myopenHAB.org service.

Fixes #12582


bundles/org.openhab.io.openhabcloud/src/main/resources/OH-INF/config/config.xml
```diff
@@ -16,13 +16,16 @@
 		</parameter>
 		<parameter name="expose" type="text" required="false" multiple="true">
 			<label>Items to Expose</label>
-			<description>List of items that are made accessible to IFTTT and similar services.</description>
+			<description>List of items that are made accessible to IFTTT and similar services. Only supported on private cloud
```
jlaur, line 19: Should this be updated also to not mention IFTTT, but perhaps Amazon Alexa and Google Home applications? Also, i18n properties should be regenerated.

bundles/org.openhab.io.openhabcloud/README.md
```diff
@@ -43,10 +43,13 @@ After installing this add-on, you will find configuration options in the openHAB
...
-Also note that at present this feature is currently turned off.
-You just need to expose those items, which you want to be accessible by IFTTT.
-This setting has no affect on Alexa or Google Assistant functionality.
+By default both remote access and push notifications are enabled.  
+
+### Advanced Configuration
+
+For private hosted myopenHAB installations, the base URL can be changed to point another cloud instance.
+
+Private hosted myopenHAB installations may enable selected items in openHAB to have their state updates pushed to the cloud service for integrations with services like IFTTT.  Note that this is not supported on the community hosted myopenHAB service due to high load concerns and will have no effect if enabled with the default URL configured.  This is also not required for remote access through the cloud service to function.
```
jlaur, line 52: Can you reformat this to one sentence per line?

jlaur (changes requested):
Thanks! Two minor comments.

bundles/org.openhab.io.openhabcloud/src/main/resources/OH-INF/config/config.xml
digitaldan: It specifically only used for IFTTT and for anything else that wants to use the limited cloud api but NOT for Alexa and google, and it only works if running a private cloud.  We have disabled this support on the main myopenhab.org service through a runtime option.  This is also why it makes sense  to put behind the advanced flag, only users who can standup their own cloud service would be interested in it.

bundles/org.openhab.io.openhabcloud/README.md
digitaldan: Yes, good point

bundles/org.openhab.io.openhabcloud/src/main/resources/OH-INF/config/config.xml
digitaldan: >Also, i18n properties should be regenerated

Ahh, did not realize there were translations, I updated the main property file, is there anything else i need to do?

jlaur: No, that's exactly what was needed. :)

jlaur (approved):
Thank you!

scatterthought:
Hi folks! This PR doesn't appear to have been merged into the 3.4 release (or 4.0 snapshot). It came up today in the OH community, so I thought I'd give it a nudge in case it got lost in the shuffle. Thanks!

https://community.openhab.org/t/rest-api-via-tasker-here-permissions/143793/7

PR #13081, opened 2022-07-04
[boschindego] Optimize coexistence with mobile app/other clients

### LINKED ISSUE #13080: [boschindego] Support deauthentication
To avoid possible undesired killed sessions on binding restart when using multiple clients (e.g. openHAB and mobile app), session should be deleted on dispose. This is supported by the API: [Deauthentication](https://github.com/zazaz-de/iot-device-bosch-indego-controller/blob/master/PROTOCOL.md#deauthentication).

jlaur:
Fixes #13080


lolodomo (approved):
LGTM, thank you

PR #13096, opened 2022-07-08
[hdpowerview] Add shade RF RSSI channels for hub/repeater

### LINKED ISSUE #12928: [hdpowerview] Repeater usage information

jlaur:
The purpose of this pull request is to introduce the possibility of monitoring repeater usage, so it can be determined if shades far away from the hub are actually connecting through the expected repeater, or if the repeater might have been turned off by accident.

I have so far not been able to correlate RSSI values with signal strength, so exposing the raw RSSI values without any interconnected logic. However, for now, only a single repeater is supported - I'm not sure if one shade can be connected to, or be aware of, more than one repeater. Unfortunately I cannot test this, as I have only a single repeater.

I have decided against scheduling a job for refreshing the channels automatically, so this can only be done by sending **REFRESH** commands (e.g. from a rule). This is for the following reasons:
- These channels are advanced and highly specialized for the specific use-case described.
- The survey call is quite unreliable, so often the response payload will be empty, or the survey data itself might be missing.
- The call is quite slow (>6 seconds) - per shade.
- Even if only updating when having linked items, this would interfere with other queries and probably would have to be done during the night to not block shade control.
- A retry mechanism would be needed, but an aggressive one would interfere even more with normal processing.
- All in all, IMHO this should be implemented in rules by advanced users in a way the suits individual needs.

Resolves #12928


bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/handler/HDPowerViewShadeHandler.java
```diff
@@ -534,15 +536,32 @@ private void doRefreshShade(RefreshKind kind) {
...
+
                         shadeData = webTargets.getShade(shadeId);
                         updateSignalStrengthState(shadeData.signalStrength);
                     } else {
                         logger.info("No data from shade {} survey", shadeId);
                         /*
-                         * Setting channel to UNDEF here would be reverted on next poll, since
-                         * signal strength is part of shade response. So leaving current value,
+                         * Setting signal strength channel to UNDEF here would be reverted on next poll,
+                         * since signal strength is part of shade response. So leaving current value,
```
lolodomo, line 560: Looks like the comment is wrong as you are no more leaving the current value.

lolodomo (commented):
Just one comment

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/handler/HDPowerViewShadeHandler.java
jlaur: The **signalStrength** channel is still left with current value, but the new RSSI channels are updated to **UNDEF** on purpose. It's a bit tricky: signal strength is part of the **shade** response (which also includes positions, battery level etc.). But in order to force refresh the value, a **survey** request has to be made prior to the **shade** request. This is how the PowerView app works, as I was intercepting traffic from it.

lolodomo: Ok

lolodomo (approved):
LGTM

PR #13100, opened 2022-07-09
Hide compiler warnings for generated code

### LINKED ISSUE #12498: [tado] Excessive Maven / Eclipse warnings on build
This is not really a bug, but more of a niggle..

While working on #12470 I noticed that the build produces over four hundred [WARNING] messages. The build succeeds, and the binding runs fine, but the build log is not a pretty sight.

The build process relies on Swagger Code-Gen to parse a YAML description of the Tado API and auto generate the respective Java classes for serialization and de-serialization. And all but a handful of the warnings are due to the building of this Swagger auto generated code.

So I would like to figure out what setting one needs to make in (for example) the POM to supress (or eliminate) the warnings when building the Swagger Code-Gen auto created classes. I would eventually create another PR post #12470 to actually implement that fix.

=> Any thoughts / suggestions?

wborn:
Generated Swagger/OpenAPI code can be messy resulting in a lot of compiler warnings obscuring more important compiler warnings.

Fixes #12498


kaikreuzer (approved):
Thanks!

PR #13107, opened 2022-07-10
Update OH version in skeleton scripts

netwolfuk:
Hi. I tried to start working on a new binding on the weekend, and it wouldn't build in Eclipse, because it couldn't find the 3.3.0 SNAPSHOTS. I have never run a `mvn install` against 3.3.0-SNAPSHOT, so it make sense that the base is not in my m2 repo.

After a little bit of digging, I found that all the bundles had moved to 3.4.0-SNAPSHOT, but my binding was still on 3.3.0-SNAPSHOT because it was created by this skeleton shell script.

This PR should solve that in future for anyone else that encounters this issue.

### DIFF
--- a/bundles/create_openhab_binding_skeleton.cmd
+++ b/bundles/create_openhab_binding_skeleton.cmd
@@ -10,7 +10,7 @@ IF %ARGC% NEQ 3 (
 	exit /B 1
 )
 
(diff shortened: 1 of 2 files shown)


wborn (approved):
Thank you!

PR #12568, opened 2022-04-03
[mail] Fix UnsupportedDataTypeException

hmerk:
Originally by @J-N-K fixed in smarthomej/addons
https://github.com/smarthomej/addons/pull/254

Reported on the forum.
https://community.openhab.org/t/mail-binding-errors-sending-mail-unsupporteddatatypeexception-text-plaijn-charset-utf-8/130365
The reason is that the standard implementation uses a context class loader to get the mime type handlers and tries to load the handlers from there. Under some circumstances this is the class loader of the activation bundle which does doesn't have the com.sun.mail packages available, leading to an javax.activation.UnsupportedDataTypeException: text/plain; charset=UTF-8 exception.

Fixes #11838

Signed-off-by: Hans-Jörg Merk <github@hmerk.de>
Also Signed-off-by: Jan N. Klug <jan.n.klug@rub.de>

PR #13108, opened 2022-07-10
[mail] Fix sending HTML/Multipart mails

hmerk:
Originally fixed in smarthomej/addons by @J-N-K 
https://github.com/smarthomej/addons/pull/378
The fix in https://github.com/smarthomej/addons/pull/254 and https://github.com/openhab/openhab-addons/pull/12568 was not sufficient for multipart messages as they contain a DataHandler for each body part. The command map was only set for the outer mime part, not for the body parts.

Also Signed-off-by: Jan N. Klug [jan.n.klug@rub.de](mailto:jan.n.klug@rub.de)


kaikreuzer (approved):
Many thanks for porting this fix!

PR #13118, opened 2022-07-12
[dali] Use correct command argument to query current color temperature

s3lph:
# [dali] Use correct command argument to query current color temperature

# Description

This PR fixes a small bug in the recently introduced DALI DT8 support:

When querying the current color temperature from a DALI DT8 device (e.g. after changing it or due to a refresh command), the OpenHAB DALI binding would currently query the actual current color temperature. If this happens while the device is still fading between two temperatures, a value somewhere between the start and target values is returned. This makes OpenHAB behave somewhat erratic (sliders jumping to unexpected positions), and we've even experienced some cases (with up/down buttons instead of sliders) where the color temperature could not be changed apart from a very limited range.

This PR changes the behavior to query the target color temperature (for some reason referred to as "temporary color temperature" in DALI instead of only "color temperature") instead of the current color temperature. This fix leads to the behavior a user would expect when chaning the color temperature: OpenHAB showing the value the DALI device is fading to.


lolodomo:
@rs22 : could you please review?

rs22 (approved):

jlaur (approved):
Thanks for the fix!

jlaur:
@lolodomo, I removed the bug label since it's a regression of an enhancement for 3.4. So it doesn't need to be mentioned in the release notes.

PR #13122, opened 2022-07-13
[homekit] fix potential null pointer exceptions

yfre:
fix 2 potentail NPE

- first one happens if addon is deployed on new and not configured openhab installation. in this case "settings" can be null
- second one can happen if homekit accessory type is missing, wrong or incomplete. reported here
https://community.openhab.org/t/homekit-binding-stops-working-after-upgrading-to-oh-3-3-0/137302


jlaur (approved):
Thank you.

PR #10278, opened 2021-03-04
[bluetooth] Add support for service data

PRosenb:
Bluetooth low energy allows broadcasters to advertise not only their presence but also `Service Data`. In `Service Data`, data of a Bluetooth Service's Characteristic can be advertised. Doing so allows a central like openHAB to receive data from a Bluetooth device without the need to connect to it. This helps to reduce power consumption on the peripheral.

[Supplement to the Bluetooth Core Specification](https://www.bluetooth.org/DocMan/handlers/DownloadDoc.ashx?doc_id=282152), 1.11 SERVICE DATA, page 19

The aim of this pull request is to add basic support of `Service Data` to openHAB so that a temperature/humidity sensor broadcasting its value as `Service Data` is supported.

I'm curious what you think about this idea in general @cpmeister and also on how I suggest to implement it. Input and feedback is very welcome.

PR #13128, opened 2022-07-14
Switch to fork of gatt parser library in order to fix incompatibility with xstream

kaikreuzer:
See https://github.com/openhab/openhab-addons/pull/10278#issuecomment-1171727057

### DIFF
--- a/bundles/org.openhab.binding.bluetooth.generic/pom.xml
+++ b/bundles/org.openhab.binding.bluetooth.generic/pom.xml
@@ -22,9 +22,9 @@
       <scope>provided</scope>
     </dependency>
     <dependency>
-      <groupId>org.sputnikdev</groupId>
+      <groupId>org.openhab</groupId>
       <artifactId>bluetooth-gatt-parser</artifactId>
-      <version>1.9.4</version>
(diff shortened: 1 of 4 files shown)


kaikreuzer:
Hm, I'm not sure why SAT complains about errors in the external library.
@wborn Can this somehow be ignored?

wborn:
>  I'm not sure why SAT complains about errors in the external library.

This library is embedded into the bundle so the library classes now also end up in `target/classes/org/openhab` which are [configured](https://github.com/openhab/static-code-analysis/blob/daee6323c1ff57e5afb0c9f076185abea8bd7bf5/sat-plugin/src/main/resources/configuration/spotbugs.properties#L10) to be analyzed as Spotbugs inspects bytecode.

> Can this somehow be ignored?

You can have a look at how SAT is [configured](https://github.com/openhab/openhab-core/blob/16cf5a8ae1e737fc42b72d33e7a3edb5fb3b5a25/pom.xml#L493) in OHC to suppress some SpotBugs findings.

kaikreuzer:
Thanks @wborn, that helped!

PR #13133, opened 2022-07-16
[rotel] Fix handling of balance command feedback

lolodomo:


jlaur (approved):
Thanks for the fix!

PR #13134, opened 2022-07-16
[rotel] Properly disable bass/treble channels when bypass is enabled

lolodomo:


lolodomo:
This is covered in #13136

PR #12731, opened 2022-05-13
[ws980wifi] Initial contribution

prof2004:
Signed-off-by: Joerg Dokupil <jdokupil@web.de>

@lsiepel 

Suggested git-commands to fix the problem with DCO have not work. Git did not know the command.
Have made your requested changes from your email regarding Re: [openhab/openhab-addons] [ws980wifi] Initial contribution (PR #12674)

Have deleted the fork, created a new fork, cloned the fork, copied my development into the bundles directory, made a complete mvn clean install, made "git commit -m 'Initial contribution' -s, git pus origin ws980wifi.

Hopefully it works...

PR #13138, opened 2022-07-18
[fineoffsetweatherstation] Add support for ELV protocol

### LINKED ISSUE #12763: [fineoffsetweatherstation] Add support for ELV WS980WiFi
#12464 introduces a new generic binding supporting white-labeled weather stations manufactured by [Fine Offset](http://www.foshk.com/), but rebranded by many companies.

Recently #12731 was also published, supporting a specific German-branded weather station, [ELV WS980WiFi](https://de.elv.com/elv-wifi-wetterstation-ws980wifi-inkl-funk-aussensensor-868-mhz-app-pc-auswertesoftware-250408)

A [sharp eye](@Hilbrand) detected that actually this [WS980WiFi weather station is one of those rebranded Fine Offset weather stations](https://github.com/openhab/openhab-addons/pull/12731#discussion_r874617259), WH2350:

image

@Andy2003 has offered to try to add support for the WS980WiFi in the Fine Offset binding when receiving some response payloads from @prof2004.

Andy2003:
Add support for ELV protocol.

resolves #12763


Andy2003:
@prof2004 can you test this PR with your weather station, please: 

https://openhab.jfrog.io/artifactory/libs-pullrequest-local/org/openhab/addons/bundles/org.openhab.binding.fineoffsetweatherstation/3.4.0-SNAPSHOT/org.openhab.binding.fineoffsetweatherstation-3.4.0-SNAPSHOT.jar

prof2004:
@Andy2003 have deleted my bundle and moved your jar-file to addons folder. stopped openhab, made openhab-cli clean-cache, and have started openhab. openhab has found the binding. Search for things has no result. entered the ws980 data manually. Station was detected and thing status green. After a short period the following message appears: 16:40:33.498 [WARN ] [.core.thing.internal.ThingManagerImpl] - Initializing handler for thing 'fineoffsetweatherstation:gateway:dcc820b94f' takes more than 5000ms. This is a loop the thing is in.

Andy2003:
@prof2004 did you configure `protocol` to `ELV`?

Andy2003:
> Protocoll is 'default'. Still communication error. Modify Protokoll to ELV. Still communication error.

Is it still timeout or anything else?

Andy2003:
@prof2004 can you provide me a trace log of the plugin please?

Andy2003:
@prof2004 testet this version with his WS980 succesfully

jlaur:
> @prof2004 tested this version with his WS980 succesfully

Excellent, I'm glad to hear that you worked it out and we will now have a single binding supporting also the tricky ELV variant. Before I dive in - mostly to you, @prof2004, does this version now support all channels/features you implemented in #12731?

bundles/org.openhab.binding.fineoffsetweatherstation/src/main/java/org/openhab/binding/fineoffsetweatherstation/internal/domain/Measurands.java
```diff
@@ -0,0 +1,171 @@
...
+ * @author Andreas Berger - Initial contribution
+ */
+@NonNullByDefault
+public class Measurands {
+
+    private static final Map<Protocol, Measurands> INSTANCES = new HashMap<>();
+    private final Map<Byte, List<Parser>> parsersPerCode = new HashMap<>();
+
+    private Measurands(Protocol protocol) {
+        try (InputStream data = Measurands.class.getResourceAsStream("/measurands.csv")) {
```
jlaur, line 48: Just for my understanding: What is the benefit of parsing this CSV resource compared to previous coded solution? Second, if this approach is beneficial, would it be worth considering JSON and getting rid of **commons-csv**, or would it be too bloated?

jlaur (changes requested):
Looks really good. One small request and one question.

bundles/org.openhab.binding.fineoffsetweatherstation/src/main/java/org/openhab/binding/fineoffsetweatherstation/internal/domain/Measurands.java
Andy2003: > What is the benefit of parsing this CSV resource compared to previous coded solution?
The CSV is clearer, and the differences between the individual protocols are easier to maintain.

If this were mapped in Java, the constructor would become even more complex than it already was.
> Second, if this approach is beneficial, would it be worth considering JSON and getting rid of commons-csv, or would it be too bloated?

Is there any problem with common-csv? The possibility of editing would not be as easy with JSON as it is with CSV.

If CSV is a problem, I would then take the Java approach again and put everything into code.

jlaur: I was remembering #7722, so I believe it should be avoided. Nothing wrong with CSV though, perhaps you could even parse this in a simple way with custom code?

Andy2003: > perhaps you could even parse this in a simple way with custom code?

Sure but why should I implement a CSV-Parser, if I can use a rock solid one?

So should I remove the dependency? If so, I would turn the data back into code...

jlaur: Looking at the code, you wouldn't need to implement a full-blown CSV parser, just some simple line-parser using a separator. Before you revert anything, let's hear @openhab/add-ons-maintainers - is it correctly understood, according to #7722, that usage of org.apache.common.* is no longer allowed, including org.apache.commons.csv?

Andy2003: > Something like this could be considered: https://www.baeldung.com/java-csv-file-array#buff-reader

Yes, I know, but this solution does not support column-based lookup, which is what I use.

prof2004:
@jlaur @Andy2003 
I have used the version of the binding provided by Andy yesterday night. I have used my binding last version and Andy's binding in parallel, after automatic discovery I have created items from all channels provided by the created things, and have compared the measures of each channel (see the dirty handwritten page attached).

findings:
channel Temperature Inside is missing
some channels have different dimensions (should not matter)
the measures are with small differences (may be because of the update cycle, do not know)

The Thing in Andy's binding shows occacionally communication errors, Read Time Out


[2022-06-21 Testergebnis ELV.pdf](https://github.com/openhab/openhab-addons/files/9150563/2022-06-21.Testergebnis.ELV.pdf)[2022-06-21 Testergebnis ELV.pdf](https://github.com/openhab/openhab-addons/files/9150563/2022-06-21.Testergebnis.ELV.pdf)

prof2004:
One more finding:
The system discovers the weatherstation via Andy's binding and puts it into the inbox of things, although it is already part of the list of things. This is double discovery.

jlaur:
> The system discovers the weatherstation via Andy's binding and puts it into the inbox of things, although it is already part of the list of things.

Looks like missing representation property when creating the discovery result for the bridge thing. This is not specific to this PR, but good finding nevertheless. @Andy2003 - this could be addressed as a separate issue/PR if you prefer. At least I will not consider it blocking for this one.

bundles/org.openhab.binding.fineoffsetweatherstation/src/main/java/org/openhab/binding/fineoffsetweatherstation/internal/domain/Measurands.java
jlaur: @cweitkamp, @kaikreuzer - approaching you directly, since I'm guessing you would know best: Should we avoid usage of `org.apache.common.*` (including `org.apache.commons.csv` as used here), as mentioned in #7722? I would not like to reintroduce usage after a previous cleanup effort, but may have misunderstood intentions, so would like confirmation before blocking this PR with such requirement.

Andy2003:
> channel Temperature Inside is missing

The value is parsed correctly according to the test data you provided via https://github.com/openhab/openhab-addons/issues/12763#issuecomment-1185644270

I've added a test for this, where you can see the indoor temperature (`temperature-indoor`) as well:

https://github.com/openhab/openhab-addons/pull/13138/files#diff-381f456875c31c995e7be38ff14dfefa42511e63117e76d34ad30f9512b421d2R59-R75

---

> some channels have different dimensions (should not matter)

the only concerning difference I can see is `irradiation-uv`.
Here I'm not quite sure if my unit (21,53 W/m²) is the correct one, but the one from @prof2004  (2153 W/m²) isn't it either.

In https://en.wikipedia.org/wiki/Ultraviolet_index it states 

> This gives a weighted figure (sometimes called Diffey-weighted UV irradiance, or DUV, or erythemal dose rate) typically around 250 mW/m² in the midday summer sunlight.
For convenience, this is divided by 25 mW/m² to produce an index nominally from 0 to 11+, though ozone depletion is now resulting in higher values.

so the uv irradiation for an index 10 would be around 250 mW/m²

In [the spec](https://osswww.ecowitt.net/uploads/20210716/WN1900%20GW1000,1100%20WH2680,2650%20telenet%20v1.6.0%20.pdf)
on page 6 it states:

> #define ITEM_UV 0x16//UV (uW/m2)

which I interpreted as µW/m² (OK there is a bug since I used µW/**cm²**) but given the [example request](https://github.com/openhab/openhab-addons/issues/12763#issuecomment-1185644270), we get the following values:

- uv irradiation: 2153 µW/m² == 2.153 mW/m²
- uv-index: 5 (weather station)
- uv-index via wiki-formula: 2.153 mW/m² / 25 mW/m² = 0,08612

taking my historical values of the binding, which are stored in W/m² in my influxdb, I calculated a quotient of approximately 4 used by the weather station.

Given all these information, I would adjust the unit of the uv irradiation to mW/m² and divide the read out value by 10:

- read out value: 2153
- interpreted value: 215.3 mW/m²
- uv-index caclulated: 215.3 mW/m³ / 40 mW/m² = 5,3
- read out uv-index: 5

so the internal conversion of the weather station is not using 25 mW/m² as stated in wiki but 40 mW/m² to get the uv-index.

---

> the measures are with small differences (may be because of the update cycle, do not know)

the differences are in fast changing values like wind direction or speed. So you are right, it is b/c of the different update cycles

---

> The Thing in Andy's binding shows occacionally communication errors, Read Time Out

How often does this happen?
Which class is printing the timeout?
If it is the discovery, it is normal, since I try to discover `ELV`  and `DEFAULT` protocol, where one may not answer the request.

Andy2003:
@prof2004 please, can you test the [latest build](https://openhab.jfrog.io/artifactory/libs-pullrequest-local/org/openhab/addons/bundles/org.openhab.binding.fineoffsetweatherstation/3.4.0-SNAPSHOT/org.openhab.binding.fineoffsetweatherstation-3.4.0-SNAPSHOT.jar)?

bundles/org.openhab.binding.fineoffsetweatherstation/src/main/java/org/openhab/binding/fineoffsetweatherstation/internal/domain/Measurands.java
cweitkamp: The general idea behind the removal of dependencies is to avoid bloated code where not necessary. For me it is okay if a binding still uses `org.apache.commons.*` for special purposes if the feature is needed and no different solution can be found. e.g. the mail binding uses `commons-email`, dbquery and influx are using `commons-csv`. Maybe the versions can be aligned.

jlaur: @cweitkamp - thanks for adding some context. The two mentioned bundles use version 1.6, but we can try to upgrade them to 1.9 for alignment. @Andy2003, I'm resolving this comment since you now have a go-ahead to use **commons-csv**.

prof2004:
@Andy2003 have tested bundle org.openhab.binding.fineoffsetweatherstation:3.4.0.202207210831. There is still no channel temp. inside, gateway has still communication errors, changing between ERROR and ONLINE. Nevertheless the values of the itmes are continuing updating. Weatherstation was recreated based on this version of the bundle. After creating the thing, discovery still finds the ws980 (double discovery).

jlaur:
> There is still no channel temp. inside

The channel doesn't even exist? Can you post a screenshot of available channels?

> gateway has still communication errors, changing between ERROR and ONLINE.

Do you mean OFFLINE and ONLINE? If OFFLINE, does it also show a reason?

> Nevertheless the values of the itmes are continuing updating.

Perhaps this could be an exception thrown somewhere because of specific values not handled/parsed correctly. Can you post logs having TRACE log level enabled?

> After creating the thing, discovery still finds the ws980 (double discovery).

Was your thing created from the discovery result, or manually? Do you see a MAC address property for the thing?

prof2004:
Available Channels:
Bildschirmfoto vom 2022-07-22 16-50-27

Status Screen Shots:
Bildschirmfoto vom 2022-07-22 16-49-59
Bildschirmfoto vom 2022-07-22 16-51-30

Part of log:
16:45:56.014 [TRACE] [ery.FineOffsetGatewayDiscoveryService] - sendBroadcastPacket: send request: FFFF12000416
16:45:56.074 [TRACE] [ery.FineOffsetGatewayDiscoveryService] - Received 46 bytes response from 192.168.0.55:46000 on Port 56086
16:45:56.083 [TRACE] [ery.FineOffsetGatewayDiscoveryService] - Discovery response received: FFFF12002CE868E7871CDAC0A80037AFC81B45617379576561746865722D57494649314344412056312E362E3447
16:45:56.092 [TRACE] [ery.FineOffsetGatewayDiscoveryService] - Thread Thread[Thread-32,5,main] waiting for data on port 56086
16:45:56.098 [TRACE] [service.FineOffsetGatewayQueryService] - executeCommand(CMD_GW1000_LIVEDATA): send request: FFFF27032A
16:46:01.023 [TRACE] [ery.FineOffsetGatewayDiscoveryService] - sendBroadcastPacket: send request: FFFF120315
16:46:01.115 [TRACE] [ery.FineOffsetGatewayDiscoveryService] - found 0 measured values via protocol DEFAULT
16:46:01.128 [TRACE] [ternal.service.ELVGatewayQueryService] - executeCommand(CMD_WS980_LIVEDATA): send request: FFFF0B0006040419
16:46:01.151 [TRACE] [ternal.service.ELVGatewayQueryService] - executeCommand(CMD_WS980_LIVEDATA): received: FFFF0B00500401011302014A0300A804014A05014D062F072608253A09253A0A00E90B00000C00000E000000001000000000110000000312000001591300001011140000130715000AA8D41607DE1705A09B
16:46:01.168 [TRACE] [ery.FineOffsetGatewayDiscoveryService] - found 21 measured values via protocol ELV
16:46:01.205 [INFO ] [ig.discovery.internal.PersistentInbox] - Added new thing 'fineoffsetweatherstation:gateway:3899189127' to inbox.
16:46:01.215 [DEBUG] [ery.FineOffsetGatewayDiscoveryService] - Thing discovered 'DiscoveryResult [thingUID=fineoffsetweatherstation:gateway:3899189127, properties={macAddress=E8:68:E7:87:1C:DA, protocol=ELV, port=45000, ip=192.168.0.55}, representationProperty=macAddress, flag=NEW, label=Wetterstation, bridgeUID=null, ttl=-1, timestamp=1658501161191]'
16:46:15.216 [INFO ] [openhab.event.ItemStateChangedEvent  ] - Item 'wetterDO_TemperatureOutside' changed from 33 °C to 33.2 °C
16:46:15.231 [INFO ] [openhab.event.ItemStateChangedEvent  ] - Item 'wetterDO_DewPointTemperature' changed from 16.4 °C to 17 °C
16:46:15.242 [INFO ] [openhab.event.ItemStateChangedEvent  ] - Item 'wetterDO_WindChillTemperature' changed from 33 °C to 33.2 °C
16:46:15.249 [INFO ] [openhab.event.ItemStateChangedEvent  ] - Item 'wetterDO_HeatIndexTemperature' changed from 33.2 °C to 33.7 °C
16:46:15.259 [INFO ] [openhab.event.ItemStateChangedEvent  ] - Item 'wetterDO_HumidityOutside' changed from 37 % to 38 %
16:46:15.264 [INFO ] [openhab.event.ItemStateChangedEvent  ] - Item 'wetterDO_PressureAbsolut' changed from 95300 Pa to 95290 Pa
16:46:15.274 [INFO ] [openhab.event.ItemStateChangedEvent  ] - Item 'wetterDO_PressureRelative' changed from 95300 Pa to 95290 Pa
16:46:15.278 [INFO ] [openhab.event.ItemStateChangedEvent  ] - Item 'wetterDO_WindDirection' changed from 233 ° to 241 °
16:46:15.284 [INFO ] [openhab.event.ItemStateChangedEvent  ] - Item 'wetterDO_WindSpeed' changed from 0.6 km/h to 0 km/h
16:46:15.293 [INFO ] [openhab.event.ItemStateChangedEvent  ] - Item 'wetterDO_WindGustSpeed' changed from 1 km/h to 0 km/h
16:46:15.298 [INFO ] [openhab.event.ItemStateChangedEvent  ] - Item 'wetterDO_LightLevel' changed from 70401 lx to 69409 lx
16:46:15.304 [INFO ] [openhab.event.ItemStateChangedEvent  ] - Item 'wetterDO_UVRadiation' changed from 2032 W/m² to 2005 W/m²
16:46:50.962 [TRACE] [ternal.service.ELVGatewayQueryService] - executeCommand(CMD_READ_FIRMWARE_VERSION): send request: FFFF500353
16:46:50.969 [TRACE] [ternal.service.ELVGatewayQueryService] - executeCommand(CMD_READ_FIRMWARE_VERSION): received: FFFF501511456173795765617468657256312E362E34
16:46:56.023 [TRACE] [ternal.service.ELVGatewayQueryService] - executeCommand(CMD_WS980_LIVEDATA): send request: FFFF0B0006040419
16:46:56.064 [TRACE] [ternal.service.ELVGatewayQueryService] - executeCommand(CMD_WS980_LIVEDATA): received: FFFF0B0050040101130201500300A9040150050155062F07250825390925390A00C90B00060C000A0E000000001000000000110000000312000001591300001011140000130715000A93081607C71705AAAF
16:47:06.080 [INFO ] [openhab.event.ItemStateChangedEvent  ] - Item 'Balkon_Lichtsensor' changed from 2629.6 to 2513.12
16:47:12.132 [TRACE] [ternal.service.ELVGatewayQueryService] - executeCommand(CMD_WS980_LIVEDATA): send request: FFFF0B0006040419
16:47:12.171 [TRACE] [ternal.service.ELVGatewayQueryService] - executeCommand(CMD_WS980_LIVEDATA): received: FFFF0B0050040101140201510300AA040151050156062F07250825380925380A00DC0B00030C00050E000000001000000000110000000312000001591300001011140000130715000A99B61607CC1705713D
16:47:12.200 [INFO ] [openhab.event.ItemStateChangedEvent  ] - Item 'ELV_Aussentemperatur' changed from NULL to 33.7 °C
16:47:12.223 [INFO ] [openhab.event.ItemStateChangedEvent  ] - Item 'ELV_Taupunkt' changed from NULL to 17 °C
16:47:12.234 [INFO ] [openhab.event.ItemStateChangedEvent  ] - Item 'ELV_GefuhlteTemperatur' changed from NULL to 33.7 °C
16:47:12.249 [INFO ] [openhab.event.ItemStateChangedEvent  ] - Item 'ELV_Warmeindex' changed from NULL to 34.2 °C
16:47:12.264 [INFO ] [openhab.event.ItemStateChangedEvent  ] - Item 'ELV_Luftfeuchtigkeitinnen' changed from NULL to 47 %
16:47:12.276 [INFO ] [openhab.event.ItemStateChangedEvent  ] - Item 'ELV_Luftfeuchtigkeitaussen' changed from NULL to 37 %
16:47:12.299 [INFO ] [openhab.event.ItemStateChangedEvent  ] - Item 'ELV_AbsoluterDruck' changed from NULL to 952.8 hPa
16:47:12.309 [INFO ] [openhab.event.ItemStateChangedEvent  ] - Item 'ELV_RelativerDruck' changed from NULL to 952.8 hPa
16:47:12.319 [INFO ] [openhab.event.ItemStateChangedEvent  ] - Item 'ELV_Windrichtung' changed from NULL to 220 °
16:47:12.328 [INFO ] [openhab.event.ItemStateChangedEvent  ] - Item 'ELV_Windgeschwindigkeit' changed from NULL to 1.08 km/h
16:47:12.334 [INFO ] [openhab.event.ItemStateChangedEvent  ] - Item 'ELV_Boengeschwindigkeit' changed from NULL to 1.8 km/h
16:47:12.339 [INFO ] [openhab.event.ItemStateChangedEvent  ] - Item 'ELV_Niederschlagsrate' changed from NULL to 0 mm/h
16:47:12.346 [INFO ] [openhab.event.ItemStateChangedEvent  ] - Item 'ELV_NiederschlagsmengeHeute' changed from NULL to 0 mm
16:47:12.353 [INFO ] [openhab.event.ItemStateChangedEvent  ] - Item 'ELV_NiederschlagsmengedieseWoche' changed from NULL to 0.3 mm
16:47:12.357 [INFO ] [openhab.event.ItemStateChangedEvent  ] - Item 'ELV_NiederschlagsmengediesenMonat' changed from NULL to 34.5 mm
16:47:12.363 [INFO ] [openhab.event.ItemStateChangedEvent  ] - Item 'ELV_NiederschlagsmengediesesJahr' changed from NULL to 411.3 mm
16:47:12.369 [INFO ] [openhab.event.ItemStateChangedEvent  ] - Item 'ELV_Niederschlagsmengegesamt' changed from NULL to 487.1 mm
16:47:12.373 [INFO ] [openhab.event.ItemStateChangedEvent  ] - Item 'ELV_Lichtstarke' changed from NULL to 69471 lx
16:47:12.378 [INFO ] [openhab.event.ItemStateChangedEvent  ] - Item 'ELV_UVStrahlung' changed from NULL to 0.1996 W/m²
16:47:12.385 [INFO ] [openhab.event.ItemStateChangedEvent  ] - Item 'ELV_UVIndex' changed from NULL to 5

I think, there is no entry indicating a communication error or even a change of status.

Yes, the thing was created from the discovery result, but I have modified the naming to ELV.

prof2004:
Another finding:
system log indicates, that the binding has identified 21 measured values via protocol ELV. This should include temp. inside, otherwise it would be only 20.

jlaur:
@Andy2003 - I'm now wondering if it would make sense to handle ELV as a separate thing type instead of using a property to distinguish. Did you consider that? Or too similar?

prof2004:
Have created a thing from the automatic discovery without modifying the name. ONLINE/OFFLINE Problem is the same, also there are only 20 channels displayed. 

Next finding: Using the default name adding the thing as discoverd leads to no more discovery results. Testet this several times in different combinations. It seams, that the test of already discovered does not handle a name (here ELV) instead of the number provided as default. 

Deleting the thing is not covered by setting the item values to null or something similar. They display the last value read throug the thing before deletion of the thing. That was one of the requests, I had to fullfill in my binding.

bundles/org.openhab.binding.fineoffsetweatherstation/src/main/java/org/openhab/binding/fineoffsetweatherstation/internal/discovery/FineOffsetGatewayDiscoveryService.java
```diff
@@ -168,15 +179,39 @@ private void discovered(String ip, int port, byte[] macAddr, String name) {
...
         properties.put(FineOffsetGatewayConfiguration.PORT, port);
+        FineOffsetGatewayConfiguration config = new Configuration(properties).as(FineOffsetGatewayConfiguration.class);
+        Protocol protocol = determineProtocol(config);
+        if (protocol != null) {
+            properties.put(FineOffsetGatewayConfiguration.PROTOCOL, protocol.name());
+        }
 
         ThingUID uid = new ThingUID(THING_TYPE_GATEWAY, id);
         DiscoveryResult result = DiscoveryResultBuilder.create(uid).withProperties(properties)
+                .withRepresentationProperty(Thing.PROPERTY_MAC_ADDRESS)
```
jlaur, line 190: You might also need to add this:
```
<representation-property>macAddress</representation-property>
```
before this line:
https://github.com/openhab/openhab-addons/blob/c729b82d59473f1fac871a5b18bda35f697fcc7a/bundles/org.openhab.binding.fineoffsetweatherstation/src/main/resources/OH-INF/thing/gateway.xml#L11

jlaur: If it has any effect, the same might be needed in `sensor.xml` (property "sensor").

Andy2003:
> There is still no channel temp. inside

Which version of openhab do you use for testing? The [indoor temperature system channel was added in 3.3](https://github.com/openhab/openhab-core/pull/2866), so you need at least this version

> gateway has still communication errors, changing between ERROR and ONLINE

@prof2004 does the offline error also happens, when you disable your plugin?
I think this could happen, when multiple parallel requests are send to the gateway (discovery + normal read out).
I can try to optimize these to not do parallel requests.

> I'm now wondering if it would make sense to handle ELV as a separate thing type instead of using a property to distinguish. Did you consider that?

Yes, I was thinking about this as well, but the only difference in handling these gateways is the protocol. 
For me, using a config value seemed less overhead than implementing a new thing.

> Deleting the thing is not covered by setting the item values to null or something similar. They display the last value read throug the thing before deletion of the thing. That was one of the requests, I had to fullfill in my binding.

Where in your PR do you set item values to `zero`, I couldn`t find that part of the code.

prof2004:
look to WS980WiFiHandler.java updateWeahterData()

    private void updateWeatherData() {
        log.debug("updateWeatherData started by pollingJob");
        WS980WiFi wsObject = new WS980WiFi(host, port);
        log.debug("wsObject for refresh created with {}, {}", wsObject.getHost(), wsObject.getPort());

        if (wsObject.refreshValues()) {
            updateState(WS980WiFiBindingConstants.CHANNEL_TEMPERATURE_INSIDE,
                    new QuantityType<>(wsObject.tempInside, SIUnits.CELSIUS));
            updateState(WS980WiFiBindingConstants.CHANNEL_TEMPERATURE_OUTSIDE,
                    new QuantityType<>(wsObject.tempOutside, SIUnits.CELSIUS));
            updateState(WS980WiFiBindingConstants.CHANNEL_TEMPERATURE_DEWPOINT,
                    new QuantityType<>(wsObject.tempDewPoint, SIUnits.CELSIUS));
            updateState(WS980WiFiBindingConstants.CHANNEL_TEMPERATURE_WINDCHILL,
                    new QuantityType<>(wsObject.tempWindChill, SIUnits.CELSIUS));
            updateState(WS980WiFiBindingConstants.CHANNEL_TEMPERATURE_HEATINDEX,
                    new QuantityType<>(wsObject.heatIndex, SIUnits.CELSIUS));
            updateState(WS980WiFiBindingConstants.CHANNEL_HUMIDITY_INSIDE,
                    new QuantityType<>(wsObject.humidityInside, Units.PERCENT));
            updateState(WS980WiFiBindingConstants.CHANNEL_HUMIDITY_OUTSIDE,
                    new QuantityType<>(wsObject.humidityOutside, Units.PERCENT));
            updateState(WS980WiFiBindingConstants.CHANNEL_PRESSURE_ABSOLUT,
                    new QuantityType<>(wsObject.pressureAbsolut, SIUnits.PASCAL));
            updateState(WS980WiFiBindingConstants.CHANNEL_PRESSURE_RELATIVE,
                    new QuantityType<>(wsObject.pressureRelative, SIUnits.PASCAL));
            updateState(WS980WiFiBindingConstants.CHANNEL_WIND_DIRECTION,
                    new QuantityType<>(wsObject.windDirection, Units.DEGREE_ANGLE));
            updateState(WS980WiFiBindingConstants.CHANNEL_WINDSPEED,
                    new QuantityType<>(wsObject.windSpeed, SIUnits.KILOMETRE_PER_HOUR));
            updateState(WS980WiFiBindingConstants.CHANNEL_WINDSPEED_GUST,
                    new QuantityType<>(wsObject.windSpeedGust, SIUnits.KILOMETRE_PER_HOUR));
            updateState(WS980WiFiBindingConstants.CHANNEL_RAIN_LAST_HOUR,
                    new QuantityType<>(wsObject.rainLastHour, Units.MILLIMETRE_PER_HOUR));
            updateState(WS980WiFiBindingConstants.CHANNEL_RAIN_LAST_DAY,
                    new QuantityType<>(wsObject.rainLastDay, Units.MILLIMETRE_PER_HOUR));
            updateState(WS980WiFiBindingConstants.CHANNEL_RAIN_LAST_WEEK,
                    new QuantityType<>(wsObject.rainLastWeek, Units.MILLIMETRE_PER_HOUR));
            updateState(WS980WiFiBindingConstants.CHANNEL_RAIN_LAST_MONTH,
                    new QuantityType<>(wsObject.rainLastMonth, Units.MILLIMETRE_PER_HOUR));
            updateState(WS980WiFiBindingConstants.CHANNEL_RAIN_LAST_YEAR,
                    new QuantityType<>(wsObject.rainLastYear, Units.MILLIMETRE_PER_HOUR));
            updateState(WS980WiFiBindingConstants.CHANNEL_RAIN_TOTAL,
                    new QuantityType<>(wsObject.rainTotal, Units.MILLIMETRE_PER_HOUR));
            updateState(WS980WiFiBindingConstants.CHANNEL_LIGTH_LEVEL,
                    new QuantityType<>(wsObject.lightLevel, Units.LUX));
            updateState(WS980WiFiBindingConstants.CHANNEL_UV_RAW, new QuantityType<>(wsObject.uvRaw, Units.IRRADIANCE));
            updateState(WS980WiFiBindingConstants.CHANNEL_UV_INDEX,
                    new QuantityType<>(wsObject.uvIndex, Units.IRRADIANCE));
            log.debug("refreshValues successfully done");
        } else {
            log.debug("refreshValues stops with Error");
            getThing().getChannels().forEach(c -> updateState(c.getUID(), UnDefType.UNDEF));
        }
    }
}

else clause...
getThing().getChannels().forEach(c -> updateState(c.getUID(), UnDefType.UNDEF));

prof2004:
I am using binding
29 │ Active │  80 │ 3.4.0.202207240824    │ openHAB Add-ons :: Bundles :: Fine Offset Weather Station

There is no channel temp inside

Andy2003:
> I am using binding 29 │ Active │ 80 │ 3.4.0.202207240824 │ openHAB Add-ons :: Bundles :: Fine Offset Weather Station
> 
> There is no channel temp inside

AND 

> Which version of openhab do you use for testing? The [indoor temperature system channel was added in 3.3](https://github.com/openhab/openhab-core/pull/2866), so you need at least this version

It is not the plugin version I am referring to, but the version of OpenHab itself, which must be at least 3.3. to test this version of the binding.

prof2004:
The openhab version I use is 4.3.2 based on rasberry pi os.

If I run the system with your binding only, there is a stabil ONLINE status, even with one item assigned to a channel.
As soon, as I add my binding, your binding is alternating between ONLINE/OFFLINE (comm. error).

Andy2003:
> The openhab version I use is 4.3.2 based on rasberry pi os.

4.3.2 sounds like the Karaf version. What version of OpenHab are you using?

https://www.openhab.org/download/



> If I run the system with your binding only, there is a stabil ONLINE status, even with one item assigned to a channel.

OK, so the 2 bindings create conflicts when querying the data at the gateway. But this should not occur in normal operation with a single binding.



> 16:47:12.171 [TRACE] [ternal.service.ELVGatewayQueryService] - executeCommand(CMD_WS980_LIVEDATA): received: FFFF0B0050040101140201510300AA040151050156062F07250825380925380A00DC0B00030C00050E000000001000000000110000000312000001591300001011140000130715000A99B61607CC1705713D

This data includes the indoor temperature.
I am quite convinced that the problem with the missing channel comes from the fact that the OpenHab version you are using does not yet have the [system channel for the indoor temperature](https://github.com/openhab/openhab-core/pull/2866) implemented.

prof2004:
You are right: openhab-cli info shows Version 3.1.1. (Build)

jlaur:
> You are right: openhab-cli info shows Version 3.1.1. (Build)

Are there any more pending issues then? Can you confirm if:
- Discovery is working correctly with latest JAR?
- The ONLINE/OFFLINE problem is resolved by stopping your own binding from running simultaneously?

bundles/org.openhab.binding.fineoffsetweatherstation/src/main/java/org/openhab/binding/fineoffsetweatherstation/internal/handler/FineOffsetSensorHandler.java
```diff
@@ -62,6 +63,7 @@ public void updateSensorState(@Nullable SensorDevice sensorDevice) {
         }
         if (sensorDevice == null) {
```
jlaur, line 64: In which situation will this happen?

Andy2003: This case occurs when a sensor thing has been configured but it is not returned by the gateway.

jlaur: Ah, thanks. So this is probably mostly relevant when unpairing a sensor because the user is getting rid of it, or can it also happen as a temporary glitch? If device is permanently removed, consider marking the thing as **GONE**:
https://www.openhab.org/docs/concepts/things.html#status-details

Comment is not directly related to this PR, just noticed it because channels are now being updated to UNDEF, perhaps because of this comment: https://github.com/openhab/openhab-addons/pull/12731#discussion_r872973142

prof2004:
Have added the binding to my test server, after upgrading the test server to openhab 3.3.0. My binding is not installed, but running on the production system. I can confirm, that al 21 measures are working with this openhab release.

ONLINE/OFFLINE changes are still existing, but now it is for sure, that this is a problem from the WS980 weatherstation in answering concurrent requests. From my point of view this should be fine, because it should be very seldem to have more than one weatherstation in one network beeing requested from to bindings at the same time.

bundles/org.openhab.binding.fineoffsetweatherstation/src/main/java/org/openhab/binding/fineoffsetweatherstation/internal/handler/FineOffsetSensorHandler.java
Andy2003: > So this is probably mostly relevant when unpairing a sensor because the user is getting rid of it

exactly

> or can it also happen as a temporary glitch?

no, this case will be handled before this condition is evaluated

> If device is permanently removed, consider marking the thing as GONE

Done

jlaur (approved):
Thanks for the effort of integrating the ELV weather station also.

PR #13139, opened 2022-07-18
[fineoffsetweatherstation] fix conversion of pressure

### LINKED ISSUE #13127: [FineOffsetWeatherStation] Air pressure numbers are not correctly imported
it seems  that values for absolute pressure and relative pressure are not correct shown in openhab.

## Expected Behavior
correct numbers should be displayed in openhab 

## Current Behavior
it looks like both pressure data are multiplied by 10 ...
(a value of 1001.7 hPa from the live date page on the gateway is shown as 10017,000 hPa in openhab)

## Possible Solution
change data collection process to return correct numbers for pressure data

## Your Environment
openhab 3.3.0 / weatherstation GW1100A_V2.1.5 with sensor WH65 / FineOffsetWeatherStation Binding
screenshot_20220714_221649
screenshot_20220714_221703

Andy2003:
resolves #13127


jlaur (approved):
Thanks for the fix!

PR #13144, opened 2022-07-19
[souliss] Souliss Typical T13 is now managed as Contact so state must be open/closed type instead of on/off type.

lucacalcaterra:
Souliss Typical T13 is now managed as Contact so state must be open/closed type instead of on/off type.


jlaur:
@lucacalcaterra - can you describe what is being fixed by this PR? Also, the PR title will show up in release notes, so it need should be understandable for users.

bundles/org.openhab.binding.souliss/README.md
```diff
@@ -192,13 +187,14 @@ Thing t11 10-0 "Albero di Natale" [node=10,slot=0]
 Thing t11 11-0 "Birra"@"Soppalco" [node=11,slot=0]
 Thing t52 11-1 "Birra - Temp 1"@"Soppalco" [node=11,slot=1]
 Thing t52 11-3 "Birra - Temp 2"@"Soppalco" [node=11,slot=3]
+
+Thing t13 0-12 "Valvola Riscaldamento Zona Giorno"@"Riscaldamento" [node=0,slot=12]
```
jlaur, line 191: I see that other existing strings here are in Italian. Nevertheless, as they should be in English, please make sure that at least newly added strings are in English.

bundles/org.openhab.binding.souliss/src/main/resources/OH-INF/thing/thing-types.xml
```diff
@@ -191,9 +191,7 @@
 		</supported-bridge-type-refs>
 		<label>T13</label>
 		<description>Digital Input</description>
-
 		<channels>
-			<channel id="stateOnOff" typeId="stateOnOff"/>
```
jlaur, line 196: Since this was the only usage of channel type id `stateOnOff`, the channel type can be removed as well.

bundles/org.openhab.binding.souliss/README.md
```diff
@@ -19,19 +19,18 @@ You can use SoulissApp and the Souliss binding at the same time, and generally u
 
 ### Sketches
 
-The easiest way is start with a simple example to control an ON/OFF light (though a relay). 
+The easiest way is start with a simple example to control an ON/OFF light (though a relay).
```
jlaur, line 22:
```suggestion
The easiest way is start with a simple example to control an ON/OFF light (through a relay).
```

bundles/org.openhab.binding.souliss/src/main/java/org/openhab/binding/souliss/internal/handler/SoulissT13Handler.java
```diff
@@ -65,12 +60,6 @@ public void setState(@Nullable PrimitiveType state) {
     public void handleCommand(ChannelUID channelUID, Command command) {
         if (command instanceof RefreshType) {
             switch (channelUID.getId()) {
-                case SoulissBindingConstants.STATEONOFF_CHANNEL:
```
jlaur, line 68: I believe constant `SoulissBindingConstants.STATEONOFF_CHANNEL` can also be removed.

jlaur (changes requested):
I have added a few comments, but also have one question: It seems this PR doesn't fix anything per se, but rather removes channel `stateOnOff` as an alternative to `stateOpenClose` (supporting the same underlying logic). Is this correct? If this channel was actually working, this is a breaking change, and should be mentioned here:
https://github.com/openhab/openhab-distro/blob/221e59eab7137d4b80aa539f37cdb104939cab13/distributions/openhab/src/main/resources/bin/update.lst#L77

bundles/org.openhab.binding.souliss/README.md
lucacalcaterra: > I have added a few comments, but also have one question: It seems this PR doesn't fix anything per se, but rather removes channel `stateOnOff` as an alternative to `stateOpenClose` (supporting the same underlying logic). Is this correct? If this channel was actually working, this is a breaking change, and should be mentioned here: https://github.com/openhab/openhab-distro/blob/221e59eab7137d4b80aa539f37cdb104939cab13/distributions/openhab/src/main/resources/bin/update.lst#L77

@jlaur on the 3.3 section ? It will be released on 3.4

jlaur: No, new section 3.4 should be added after that line. :)

lucacalcaterra:
prefer to close it, may introduce new errors.
I'll open another pr with minor fixes

PR #13147, opened 2022-07-19
[miele] Allow hostname as gateway network address

### LINKED ISSUE #13145: [miele] Hostname for gateway should be accepted
See this forum post:
https://community.openhab.org/t/miele-binding-does-not-accept-local-ip-name/63350

Currently gateway has to be configured with IP address, while local hostname is not accepted. The validation for IP address should be dropped.

jlaur:
Fixes #13145

Additionally:
- Documented configuration parameters (and removed redundant example).
- Added message for bridge communication error, so that it now shows for example "java.net.UnknownHostException: mielex: Name or service not known"


kaikreuzer (approved):
lgtm, thanks!

PR #13148, opened 2022-07-19
[rotel] Fix syntax of few ASCII commands

lolodomo:
Impacted channels: track and playControl for models CD11, CD14 and RCD1572


lolodomo:
As "?" was missing, the commands to retrieve the current playback status and current track number were not working and corresponding channels were not set at Initialization. 

Regarding the command to get the DSP mode, this is different. The 3 models use the ASCII V1 protocol and so the command get_dsp_mode. The command for ASCII V2 protocol is just not yet known.

jlaur:
> > Regarding the command to get the DSP mode, this is different. The 3 models use the ASCII V1 protocol and so the command get_dsp_mode. The command for ASCII V2 protocol is just not yet known.
> 
> Are some models still using the ASCII V1 protocol? Since that enum seems not model-specific, will `DSP_MODE` then stop working for any such models?

Ah, sorry, now I get it. Last argument is for V2 protocol, and it's not known, so you removed it.

jlaur (approved):
Thanks for the fix! Can you update the PR title to mention that requesting play status and track has been fixed, as this seems like a bugfix that could actually affect users?

PR #13150, opened 2022-07-20
Fix build

J-N-K:
See https://github.com/openhab/openhab-core/pull/2981

This seems a bit awkward. But it'll work the same way.

### DIFF
--- a/bundles/org.openhab.binding.mielecloud/src/main/java/org/openhab/binding/mielecloud/internal/config/servlet/CreateBridgeServlet.java
+++ b/bundles/org.openhab.binding.mielecloud/src/main/java/org/openhab/binding/mielecloud/internal/config/servlet/CreateBridgeServlet.java
@@ -127,10 +127,11 @@ public final class CreateBridgeServlet extends AbstractRedirectionServlet {
                 .withProperty(Thing.PROPERTY_MODEL_ID, MIELE_CLOUD_BRIDGE_NAME)
                 .withProperty(MieleCloudBindingConstants.CONFIG_PARAM_LOCALE, locale)
                 .withProperty(MieleCloudBindingConstants.CONFIG_PARAM_EMAIL, email).build();
-        if (inbox.add(result)) {
-            return pairBridge(bridgeUid);
-        } else {
+        if (thingRegistry.get(bridgeUid) != null) {
             return reconfigureBridge(bridgeUid);
+        } else {
+            inbox.add(result);
+            return pairBridge(bridgeUid);
         }
     }
 
--- a/itests/org.openhab.binding.mielecloud.tests/src/main/java/org/openhab/binding/mielecloud/internal/config/servlet/CreateBridgeServletTest.java
+++ b/itests/org.openhab.binding.mielecloud.tests/src/main/java/org/openhab/binding/mielecloud/internal/config/servlet/CreateBridgeServletTest.java
@@ -25,6 +25,7 @@ import org.junit.jupiter.api.Test;
 import org.openhab.binding.mielecloud.internal.MieleCloudBindingConstants;
 import org.openhab.binding.mielecloud.internal.auth.OAuthTokenRefresher;
 import org.openhab.binding.mielecloud.internal.config.MieleCloudConfigService;
+import org.openhab.binding.mielecloud.internal.config.exception.BridgeReconfigurationFailedException;
 import org.openhab.binding.mielecloud.internal.util.AbstractConfigFlowTest;
 import org.openhab.binding.mielecloud.internal.util.MieleCloudBindingIntegrationTestConstants;
 import org.openhab.binding.mielecloud.internal.util.Website;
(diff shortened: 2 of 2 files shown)


jlaur (approved):
Thank you!

bundles/org.openhab.binding.mielecloud/src/main/java/org/openhab/binding/mielecloud/internal/config/servlet/CreateBridgeServlet.java
```diff
@@ -127,10 +127,11 @@ private Thing pairOrReconfigureBridge(String locale, ThingUID bridgeUid, String
                 .withProperty(Thing.PROPERTY_MODEL_ID, MIELE_CLOUD_BRIDGE_NAME)
                 .withProperty(MieleCloudBindingConstants.CONFIG_PARAM_LOCALE, locale)
                 .withProperty(MieleCloudBindingConstants.CONFIG_PARAM_EMAIL, email).build();
-        if (inbox.add(result)) {
-            return pairBridge(bridgeUid);
-        } else {
+        if (thingRegistry.get(bridgeUid) != null) {
             return reconfigureBridge(bridgeUid);
+        } else {
+            inbox.add(result);
```
BjoernLange, line 133: Wouldn't `inbox.approve` in `pairBridge` fail if the discovery result cannot be added instantly? Shouldn't the result of the future be awaited here and the pairing only proceed if the discovery result was successfully added to the inbox? Awaiting should be done with a reasonable timeout because the user is waiting for the success / failure page to load in the browser.

bundles/org.openhab.binding.mielecloud/src/main/java/org/openhab/binding/mielecloud/internal/config/servlet/CreateBridgeServlet.java
```diff
@@ -127,10 +127,11 @@ private Thing pairOrReconfigureBridge(String locale, ThingUID bridgeUid, String
                 .withProperty(Thing.PROPERTY_MODEL_ID, MIELE_CLOUD_BRIDGE_NAME)
                 .withProperty(MieleCloudBindingConstants.CONFIG_PARAM_LOCALE, locale)
                 .withProperty(MieleCloudBindingConstants.CONFIG_PARAM_EMAIL, email).build();
-        if (inbox.add(result)) {
-            return pairBridge(bridgeUid);
-        } else {
+        if (thingRegistry.get(bridgeUid) != null) {
```
BjoernLange, line 130: With this check in place, `reconfigureBridge` should be changed to accept the thing obtained here if it has been obtained.

itests/org.openhab.binding.mielecloud.tests/src/main/java/org/openhab/binding/mielecloud/internal/config/servlet/CreateBridgeServletTest.java
```diff
@@ -74,11 +74,10 @@ public void whenBridgeReconfigurationFailsDueToMissingBridgeThenAWarningIsShownO
         assertNotNull(createBridgeServlet);
 
         Inbox inbox = mock(Inbox.class);
-        when(inbox.add(any())).thenReturn(false);
         setPrivate(Objects.requireNonNull(createBridgeServlet), "inbox", inbox);
 
         ThingRegistry thingRegistry = mock(ThingRegistry.class);
-        when(thingRegistry.get(any())).thenReturn(null);
+        when(thingRegistry.get(any())).thenThrow(new BridgeReconfigurationFailedException(""));
```
BjoernLange, line 80: `ThingRegistry` will never throw `BridgeReconfigurationFailedException`, so this shouldn't be done in tests.

J-N-K: It just mocks the behavior before: if inbox.add returned false and thingRegistry.get returned null this exception is thrown.

bundles/org.openhab.binding.mielecloud/src/main/java/org/openhab/binding/mielecloud/internal/config/servlet/CreateBridgeServlet.java
J-N-K: Probably. Can you prepare a PR for that?

BjoernLange: Yes.

itests/org.openhab.binding.mielecloud.tests/src/main/java/org/openhab/binding/mielecloud/internal/config/servlet/CreateBridgeServletTest.java
BjoernLange: It's not about changing the test behavior. Mocking `ThingRegistry.get` this way breaks the method contract of  `ThingRegistry.get`, i.e. the code doesn't behave the same way when executing tests as when running it productively. In consequence this might lead to subtle bugs when code is changed that calls this method. It might seem fine now but can easily break in the future.

PR #13155, opened 2022-07-21
[hdpowerview] Fix excessive firmware version updates

### LINKED ISSUE #13154: [hdpowerview] Firmware versions are fetched at every poll cycle
## Expected Behavior
Firmware versions should be fetched when binding is initialized, and optionally with low frequency after that.

## Current Behavior
Version information is fetched at every poll cycle, i.e. once per minute with default settings.

## Possible Solution
Root cause is that method `updateFirmwareProperties` does not cache the firmware information instance variable `firmwareVersions`, but only in local variable of same name. The reason why the instance variable was introduced was having the ability to check cached firmware version in logic which needed to know whether on V1 or V2 hub because of #11856. However, the final fix didn't need this information after all, so it should be dropped and replaced by a timestamp instead, so it could be invalidated after some time.

## Steps to Reproduce
Enable debug logging and observe that `GET /api/fwversion` requests are made every minute:

```
2022-07-21 11:42:52.649 [DEBUG] [ternal.handler.HDPowerViewHubHandler] - Main processor firmware version received: PV Hub2.0, 2.0.1056
2022-07-21 11:42:52.660 [DEBUG] [ternal.handler.HDPowerViewHubHandler] - Radio firmware version received: 2.0.2610

```
## Context
This is a regression of #11979.

jlaur:
Firmware version properties are now only updated:
- On binding initialization.
- Once per day.
- After hub communication errors (which might be due to firmware upgrade).

Fixes #13154

### DIFF
--- a/bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/handler/HDPowerViewHubHandler.java
+++ b/bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/handler/HDPowerViewHubHandler.java
@@ -12,6 +12,8 @@
  */
 package org.openhab.binding.hdpowerview.internal.handler;
 
+import java.time.Duration;
+import java.time.Instant;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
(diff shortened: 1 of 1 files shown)


lolodomo (approved):
LGTM

PR #13101, opened 2022-07-09
[homematic] Update README.md

MeisterPiffer:
I came over this topic: https://community.openhab.org/t/refreshtypes-not-working-with-oh-3/114715/4 as I tried to follow the docs, which didn't mention rules DSL and I think this should be in the docs as I guess many people use Rules DSL. Hence adding another example.
Not sure at all, if I'm doing it the right way. This is my first try to contribute something via github. Thanks

Ok, now that I read the links below and contributing guidelines, I'm 100% sure that I did something wrong, but I'll try anyway :)

PR #13159, opened 2022-07-22
[homematic] Update README.md

MeisterPiffer:
not sure about the php though, this is from the example which is DSL as well.

### DIFF
--- a/bundles/org.openhab.binding.homematic/README.md
+++ b/bundles/org.openhab.binding.homematic/README.md
@@ -660,25 +660,26 @@ openhab:send Var_1 REFRESH
 
 In scripts:
 
- :::: tabs 
+:::: tabs
 
- ::: tab JavaScript 
+::: tab JavaScript
  
- ``` javascript 
- import org.openhab.core.types.RefreshType 
- ... 
- Var_1.sendCommand(RefreshType.REFRESH) 
- ``` 
- ::: 
+``` javascript
+import org.openhab.core.types.RefreshType
+...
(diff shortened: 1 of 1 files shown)


jlaur:
Correction to #13101.

jlaur:
@MeisterPiffer - not sure if it will be rendered differently, but can we have linefeeds before and after each code fence like in previously posted example?

MeisterPiffer:
> @MeisterPiffer - not sure if it will be rendered differently, but can we have linefeeds before and after each code fence like in previously posted example?

I now tried to copy the formatting with all extra spaces like from the example.

jlaur:
> > @MeisterPiffer - not sure if it will be rendered differently, but can we have linefeeds before and after each code fence like in previously posted example?
> 
> I now tried to copy the formatting with all extra spaces like from the example.

Sorry, but only newlines requested, no additional spaces.

MeisterPiffer:
> > > @MeisterPiffer - not sure if it will be rendered differently, but can we have linefeeds before and after each code fence like in previously posted example?
> > 
> > 
> > I now tried to copy the formatting with all extra spaces like from the example.
> 
> Sorry, but only newlines requested, no additional spaces. _Edit:_ It might help to look at the example in GitHub as raw.

ok, thanks for your patience.

jlaur:
@MeisterPiffer - now it looks good:
https://next.openhab.org/addons/bindings/homematic/#troubleshooting

:thumbsup:

PR #13166, opened 2022-07-23
[shelly] Restructuring to prepare integration of Plus/Pro support

markus7017:
This PR is just to reduce the delta before I create the PR for Plus/Pro support. In general Gen1 and Gen2 devices have a total different API so I spliced the API code in version/gen 1 and 2. This means restructuring the classes (e.h. api+coap became api+api1+api2) and therefor dummy changes in the code, e.g. import statements. This creates a lot of overhead and separating those "dummy changes" simplifies merging the actual PR.

@fwolter Do a cross check, but I thing you could just merge it


markus7017:
something went wrong with push, closing the PR and creating a new one

PR #13167, opened 2022-07-23
[boschindego] Fix communication error when smart mowing is disabled

### LINKED ISSUE #13163: [boschindego] COMMUNICATION_ERROR No content returned
Since two weeks or so I get "COMMUNICATION_ERROR No content returned" with the binding. Some values are not returned anymore (e.g. battery temperature, battery soc, battery voltage).

I am using openHAB 3.3 with openHABian, binding version is 3.4.0.202207031005. I did reinstall the binding several times.

Any ideas?

Thanks for help!

jlaur:
With smart mowing disabled and channel **nextCutting** linked, binding would go **OFFLINE** with **COMMUNICATION_ERROR**.

This happened because server in this case returns empty response with HTTP code 204 ("No Content"), which would throw `IndegoInvalidResponseException`. This is now being directly handled so that deserialization or other errors will still be handled correctly as errors.

Same fix applied for channel **lastCutting** for the specific case of a brand new mower without any cuts yet.

Tested with smart mowing enabled as well as disabled. When disabled, channel **nextCutting** will be updated to **UNDEF**.

Regression of #12989
Fixes #13163

### DIFF
--- a/bundles/org.openhab.binding.boschindego/src/main/java/org/openhab/binding/boschindego/internal/IndegoController.java
+++ b/bundles/org.openhab.binding.boschindego/src/main/java/org/openhab/binding/boschindego/internal/IndegoController.java
@@ -643,8 +643,14 @@ public class IndegoController {
      * @throws IndegoException if any communication or parsing error occurred
      */
     public @Nullable Instant getPredictiveLastCutting() throws IndegoAuthenticationException, IndegoException {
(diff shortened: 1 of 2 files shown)


jlaur:
@lolodomo - if you find time, please have a look at this one. It would be nice to have merged before first 3.4 milestone as it fixes a feature that will also be included in that milestone.

lolodomo (approved):
LGTM, thank you

PR #13168, opened 2022-07-24
[homematic] Allow default value that is less than allowed minimum

### LINKED ISSUE #13097: [homematic] HM-CC-TC Thermostat HANDLER_CONFIGURATION_PENDING due to invalid default parameters
The default configuration of things for the HM-CC-TC device seems to contain invalid parameter values.

## Expected Behavior
Configuration files (thing files) that are correctly working in OH2.5 should load correctly in OH3.x

## Current Behavior
Loading thing files with HM-CC-TC things results in them ending up in the `UNINITIALIZED (HANDLER_CONFIGURATION_PENDING)` instead of `ONLINE`.
Full message shown in the web UI is available [here](https://pastebin.com/QHDiYXXn) and hints at invalid default parameters being set when the thing is loaded.
If a HM-CC-TC thing is created via the web UI, the configuration fields mentioned in the error message are highlighted red and need to be manually corrected to a value larger than 10 for the user to be able to create a thing.

## Possible Solution
From what I can tell, this issue is very similar to #12183, which was related to another component and fixed in February in #12192. I am not familiar with the code enough to provide more pointers or fix it myself and create a PR.

Via the web UI or with a thing file, create a thing for a HM-CC-TC device, you should see that the default values are invalid.
An example thing config:
```
Bridge homematic:bridge:ccu [ gatewayAddress="192.168.xxx.xxx", callbackHost="192.168.xxx.xxx" ] {
  Thing HM-CC-TC JEQ0012660 "Name" @ "Location"
}
```

## Context
My OpenHAB instance is still running on v2.5 since I didn't have time to spend the time on upgrading.
Now I was starting to play around with this an this seems to not work right now.
Is there any way to fix this that his not creating the things via the web UI and fixing all of the invalid parameters by hand?

## Your Environment
* OpenHAB 3.3.0
* Homematic 3.3.0
* OpenHAB Docker on Debian

MHerbst:
For some Homematic devices the default values can be higher than the allowed maximum value or less than the allowed minimum value (e.g. 0 for "not set"). 

Default values that are higher than the maximum value were alreaday allowed. With this fix also default values that are less than  the minimum value are allowed.

Unfortunately, this means that the range of allowed values has to be increased.

Resolves #13097

Signed-off-by: Martin Herbst <develop@mherbst.de>

### DIFF
--- a/bundles/org.openhab.binding.homematic/src/main/java/org/openhab/binding/homematic/internal/type/HomematicTypeGeneratorImpl.java
+++ b/bundles/org.openhab.binding.homematic/src/main/java/org/openhab/binding/homematic/internal/type/HomematicTypeGeneratorImpl.java
@@ -374,12 +374,18 @@ public class HomematicTypeGeneratorImpl implements HomematicTypeGenerator {
                     if (dp.isNumberType()) {
                         Number defaultValue = (Number) dp.getDefaultValue();
                         Number maxValue = dp.getMaxValue();
-                        // some datapoints can have a default value that is greater than the maximum value
-                        if (defaultValue != null && maxValue != null
-                                && defaultValue.doubleValue() > maxValue.doubleValue()) {
-                            maxValue = defaultValue;
+                        Number minValue = dp.getMinValue();
+                        if (defaultValue != null) {
+                            // some datapoints can have a default value that is greater than the maximum value
+                            if (maxValue != null && defaultValue.doubleValue() > maxValue.doubleValue()) {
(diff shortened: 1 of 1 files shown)


jlaur:
@MHerbst - if you change "Solves #xxxxx" in PR description to "Resolves #xxxxx", then issue will be automatically linked and closed when PR is merged.

jlaur (approved):
LGTM - thanks!

MHerbst:
Thanks for the hint, I always forget the correct wording :-)

PR #13169, opened 2022-07-24
[hdpowerview] Fix tilt support for vertical slats

andrewfg:
### Background

As reported on the forum [here](https://community.openhab.org/t/hd-powerview-binding-issue-openhab-3-3-0/136921/29?u=andrewfg) one user has reported that his type 55 and type 56 shades do not support vane tilt controls.

As discussed [here](https://github.com/sander76/aio-powerview-api/issues/16#issuecomment-1193251782) it seems that some instances of these shades are incorrectly reporting capabilities:4 when they actually support capabilities:3 functionality

### Solution

This PR applies a `capabilitiesOverride()` to those shade types.


andrewfg:
For testing, the JAR file is here...

[org.openhab.binding.hdpowerview-3.4.0-SNAPSHOT.jar.zip](https://github.com/openhab/openhab-addons/files/9175448/org.openhab.binding.hdpowerview-3.4.0-SNAPSHOT.jar.zip)

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/database/ShadeCapabilitiesDatabase.java
```diff
@@ -79,9 +79,9 @@ public class ShadeCapabilitiesDatabase {
             new Type(47).capabilities(7).text("Pleated Top Down Bottom Up"),
             new Type(49).capabilities(0).text("AC Roller"),
             new Type(51).capabilities(2).text("Venetian"),
-            new Type(54).capabilities(3).text("Vertical Slats Left Stack"),
-            new Type(55).capabilities(3).text("Vertical Slats Right Stack"),
-            new Type(56).capabilities(3).text("Vertical Slats Split Stack"),
+            new Type(54).capabilities(4).text("Vertical Slats Left Stack").capabilitiesOverride(3),
+            new Type(55).capabilities(4).text("Vertical Slats Right Stack").capabilitiesOverride(3),
+            new Type(56).capabilities(4).text("Vertical Slats Split Stack").capabilitiesOverride(3),
```
jlaur, line 84: I'm now wondering: From which source did we get the specification of capabilities 3 and 4? Capabilities 4 is not referenced anymore, so the problem could in theory also be wrong definitions. For example, adding `tiltAnywhere()` and `tilt180()` to capabilities 4 (so it would be identical to capabilities 3) would also solve the problem (again, theoretically).

andrewfg: Agreed. However I would like to try, if possible, to harmonise the OH approach with that of HA, in order to avoid confusion.

> Capabilities 4 is not referenced anymore

Actually shade types 69,70,71 do reference capabilities:4, and from the description 'curtain' I would assume that they do not support tilt. Which is consistent with what capabilities:4 currently says. However I think that neither @kingy444 nor we have actually seen a real user with any of those types. However if / when such a user does appear then we could confirm or deny the suppositions.

jlaur: You are right, I missed that.

jlaur (approved):
Thanks! Do you want to await user confirmation before merging?

jlaur:
@andrewfg - consider changing PR title to something like "Fix tilt support for vertical slats" so it will be more understandable for users when reading release notes.

andrewfg:
> Do you want to await user confirmation before merging?

Yes, I think we should do that.

> consider changing PR title

Ok. Done.

andrewfg:
@jlaur the user has confirmed the proper functionality [here](https://community.openhab.org/t/hd-powerview-binding-issue-openhab-3-3-0/136921/42?u=andrewfg)

PR #13176, opened 2022-07-25
[shelly]Restructuriing to prepare Plus/Pro support integration

markus7017:
This PR is just to reduce the delta before I create the PR for Plus/Pro support. In general Gen1 and Gen2 devices have a total different API so I spliced the API code in version/gen 1 and 2. This means restructuring the classes (e.h. api+coap became api+api1+api2) and therefor dummy changes in the code, e.g. import statements. This creates a lot of overhead and separating those "dummy changes" simplifies merging the actual PR.


markus7017:
@fwolter fyi: This should be an easy PR. I does not contain functional changes, just restructuring the API layer. Next one will then include bug fixes from current hardening and another one will finally add Plus/Pro support.

fwolter (approved):

fwolter:
I removed the label to avoid a listing in the release notes.

PR #13185, opened 2022-07-28
[mielecloud] Adapt to inbox api changes

BjoernLange:
Adapt mielecloud binding to `Inbox` API changes from https://github.com/openhab/openhab-core/pull/2981 and fix potential problems from #13150.


jlaur (approved):
I have just given this a test run and it seems to work. Thanks for adapting this to the API change!

jlaur:
@BjoernLange - I'm a bit confused about the pairing process though. Specifically where client ID and client secret is stored, since this is not configured within the bridge?

BjoernLange:
Thanks, @jlaur. Client ID and client secret aren't stored, they are used to obtain access and refresh token during the pairing process and then never required again.

Issue #11686, opened 2021-12-01
[http-binding]: Inbound state is not fully implemented on the rollershutter channel type


Neodym81:
After switching from openhab 2.5 to 3.x the ability to update a Rollershutter item via Rollershutter channel does not work. The item value is set to UNDEF. For other basic types like String and Number it does work as expected.

The discussion around the issue took place in https://community.openhab.org/t/rollershutter-on-http-bindig-with-transformation/129427/16

## Expected Behavior
The current value from stateExtension in the Rollershutter channel should be transformed to the Rollershutter item. (Percentage value)

## Current Behavior
The value is always set to UNDEF

## Steps to Reproduce (for Bugs)
1. Create an HTTP URL thing with a rollershutter channel
2. configure stateExtension to point to a mock service having a static numeric value like 50.
3. Create a rollershutter item and link it to the channel of point 2

## Context
showing the percentage value of a rollershutter actor on the UI

## Your Environment
inux raspberrypi 5.10.63-v7l+
openhab 3.2.0.M4 but also present in 3.1.0

cat101:
Just to add one more piece of context: I observe the same behavior as @Neodym81. I Also tried changing the channel type to `dimmer` and in this case the state is reflected correctly on the UI (but unfortunately I cannot control it). As already mentioned this used to work fine for me on v2.5

Timtam:
Same issue here. I was able to retrieve the current position via a number channel, but I wasn't able to set my position yet, doesn't matter if through the rollershutter channel or the number channel, http binding doesn't trigger at all when trying to send a command, at least the logs don't show anything when set to log:set TRACE org.openhab.binding.http. Retrieval works, and rollershutter channel always shows UNDEF, no matter what I do.

PR #13191, opened 2022-07-29
[http] fix rollershutter return as position always undef

vich-667:
Fix [ISSUE 11686 ](https://github.com/openhab/openhab-addons/issues/11686). It worked to me on my local tryout.


lolodomo (approved):
LGTM, thank you

Issue #13182, opened 2022-07-27
[boschindego] Optimization of API calls


jlaur:
In order to:
- Reduce risk of hitting rate limits (or being banned) by Bosch.
- Avoid unneeded wake-ups from robot sleep (and display turning on in the middle of the night).
- Possibly have faster updates.

it would be beneficial to:
- Explore additional possibilities with the API
- Revisit previously implemented logic (#12938, #12986, #13017).
- Introduce new algorithms for smarter polling.

As an example, operating data is of most interest when the device is working or charging. Additionally, when this data is requested, the display on the device will turn on, i.e. the device will be woken up. On the other hand, only by performing this call will we actually reach the device, so if we would avoid calling it for several days, the device could have been stolen, and it would still appear offline.

The state request:
`GET https://api.indego.iot.bosch-si.com/api/v1/alms/xxxxxxxx/state`

has some parameters that might be of use:
- **longpoll**=true: awaits any state change
- **timeout**=&lt;timeout in seconds&gt;: timeout for long poll, when reached will return 504.
- **cached**=false: not sure what it does
- **force**=true: not sure what it does

The **longpoll** is particularly interesting as this could be a way to get faster updates with fewer calls, i.e. a kind of push mechanism. If this would be implemented as asynchronous HTTP calls using Jetty, that might be a significant improvement.

#13158 recently improved update of next planned cutting time, but this could be further improved by separating last and next cutting times. This way one of the two calls could be eliminated in some cases. For example, when time of next planned cutting is hit, next planned cutting needs an update, but last cutting does not. This will only be updated after completing a session - which does not affect next cutting. The calls are inexpensive, but nevertheless.

Last, when making changes to .items files, all items will request **REFRESH**. Not only will this cause multiple calls, but worse than that, it will cause redundant calls as some channels are updated through the same API calls.

jlaur:
I have now been able to verify that this request indeed does start position tracking:
`POST /api/v1/alms/xxxxxxx/requestPosition?count=10&interval=6`

Also, **longpoll** mentioned above has been confirmed to work in the way that it doesn't return until state has changed. When it does, the response is a bit differently structured than usual:

```
{
    "state": 258,
    "map_update_available": false,
    "charge": 24,
    "operate": 30,
    "mapsvgcache_ts": 1659036798849,
    "svg_xPos": 408,
    "svg_yPos": 680
}
```

jlaur:
[...]

PR #13192, opened 2022-07-29
[boschindego] Provide faster channel updates

jlaur:
This pull request implements a balanced way of using the API in order to provide faster updates and reduced load when fast updates are not needed. In addition, position tracking is now automatically requested when mower is active. This allows for updating the map with a frequency as high as six seconds between refreshes.

Some details:
- The most expensive call, `/operatingData` (contacting the robot), was previously called every three minutes (with default configuration). This has now been changed to every six hours when idle and every two minutes when active or charging. For a mower running one hour every day and charging 30 minutes, this **reduces** the number of calls per day from 480 to 49 while at the same time providing faster updates. It also fixes an issue with the display being always on, since this call wakes up the device.
- The most common call, `/state`, was previously called every three minutes (with default configuration). This has now been changed to every three minutes when idle, every two minutes when charging and every 30 seconds when active (in which case the result now includes updated map positions). For the same mower example, this **increases** the number of calls from 480 to 585 if the map channel is linked.
- The last/next cutting times are still updated once per hour, but additional updates have been **reduced** and at the same time been made more accurate. For example, the last cutting time will now be updated before the mower even returns to the dock.
- State update after triggering commands has been fixed to wait for new state and then start receiving updates with a frequency according to the need. Previously state would be refreshed before actually changed, so when starting a mowing session it could take up to three minutes to update state accordingly.

Related to:
- #13182
- #12938
- #12986
- #13017

### DIFF
--- a/bundles/org.openhab.binding.boschindego/README.md
+++ b/bundles/org.openhab.binding.boschindego/README.md
@@ -8,12 +8,13 @@ His [Java Library](https://github.com/zazaz-de/iot-device-bosch-indego-controlle
 
 Currently the binding supports  ***indego***  mowers as a thing type with these configuration parameters:
 
-| Parameter          | Description                                                     | Default |
-|--------------------|-----------------------------------------------------------------|---------|
-| username           | Username for the Bosch Indego account                           |         |
-| password           | Password for the Bosch Indego account                           |         |
-| refresh            | The number of seconds between refreshing device state           | 180     |
-| cuttingTimeRefresh | The number of minutes between refreshing last/next cutting time | 60      |
+| Parameter          | Description                                                       | Default |
+|--------------------|-------------------------------------------------------------------|---------|
+| username           | Username for the Bosch Indego account                             |         |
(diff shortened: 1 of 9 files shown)


jlaur:
This PR is mostly an enhancement with faster updates and especially almost real-time map position tracking.

However, the first mentioned point about **operating data** requests can be considered a fix for a regression in #12938. At that time I was not aware that this request would wake up the device, effectively keeping it awake all the time (poll frequency three minutes, time for going to sleep 10 minutes).

Therefore it would be very nice to get ready in time for first 3.4 milestone release to avoid potentially a lot of robots being contacted through the service way too many times. Potentially Bosch could take counter-measures and we could get into the same situation as the Gardena binding with HTTP error 429 ("Too Many Requests"). While working on this PR I have only experienced that a single time during some heavy testing with binding, app and Postman performing requests simultaneously. FYI @openhab/add-ons-maintainers

lolodomo (approved):
LGTM, thank ypu

PR #13203, opened 2022-08-02
[rotel] New command to toggle between stereo and bypass

lolodomo:
Concerns the models RSP-1570, RSP-1572 and RSX-1562

BYPASS as DSP mode is also removed for these models.


jlaur (approved):
LGTM, thanks.

Issue #13209, opened 2022-08-03
[openhabcloud] sleepSocketIO(long delay) halts entire io.socket thread


lsafelix75:
Hi 

This line of code in CloudClient.java causes a huge problem if another bundle is developed to use io.socket. Once this function is called in e.g openhabcloud add-on bundle due to connectivity issue, the delay function will affect io.socket  on another bundle. Please consider to use asynchronous way like runnable scheduler to prepare for delay. 

  ```
                  socket.close();
        -----> sleepSocketIO(delay);
                    socket.connect();
```

ssalonen:
Discussion on this in community forums: https://community.openhab.org/t/solved-io-socket-2-difference-socket-instances-connect-to-2-different-cloud-servers-causing-internal-routing-confusion-within-oh3/137818/3?u=ssalonen

lolodomo:
Closed by #13421

PR #13210, opened 2022-08-03
[openhabcloud] sleepSocketIO(long delay) halts entire io.socket thread

lsafelix75:
refer to https://github.com/openhab/openhab-addons/issues/13209


lolodomo:
Your PR is not compiling. Please run `mvn spotless:apply`
You also need to properly sign your PR to avoid the DCO failure.

Regarding the new thread pool you are creating, I am asking myself if you could not use the common thread pool. @kaikreuzer  (as binding maintainer), WDYT ?

bundles/org.openhab.io.openhabcloud/src/main/java/org/openhab/io/openhabcloud/internal/CloudClient.java
```diff
@@ -124,6 +125,10 @@ public class CloudClient {
     private boolean remoteAccessEnabled;
     private Set<String> exposedItems;
 
+    /*
+     * This variable is used to schedule io.socket reconnection asynchornously
```
Flole998, line 129: You misspelled "asynchronously"

Flole998 (changes requested):

PR #13211, opened 2022-08-03
Update Ffmpeg for accept special URL

siglp:
# Update Ffmpeg command
Ffmpeg command  should accept special URL (with parameters)

# Descripton

Proposal for fixing bug.

Some cameras use in rtps parameters (ex. rtsp://user:password@ipadress:554/cam/realmonitor?channel=1&amp;subtype=0). But ffmpeg will not run if whole input is not in quotes or double quotes.

Note: This example is from Amcrest Doorbell, it must be set manully as Generic IP camera, because URL in Dahua and Amcrest is not ok for this camera.

# Testing

On my own doorbell camera :-) (https://www.amazon.com/Amcrest-Doorbell-Weatherproof-Wide-Angle-AD110/dp/B07ZJS3L5Y)


siglp:
Sorry, this is not the right way. All is needed is to add double-quotes around "FFmpeg Input" value in Thing definiton.

PR #13215, opened 2022-08-04
[atlona] Fix NPE on shutdown when Thing disabled

mlobstein:
This PR fixes an NPE that occurs when the Thing is disabled and openhab calls dispose() on shutdown. Also fix a large number of warnings.

### DIFF
--- a/bundles/org.openhab.binding.atlona/src/main/java/org/openhab/binding/atlona/internal/pro3/AtlonaPro3PortocolHandler.java
+++ b/bundles/org.openhab.binding.atlona/src/main/java/org/openhab/binding/atlona/internal/pro3/AtlonaPro3PortocolHandler.java
@@ -199,7 +199,7 @@ class AtlonaPro3PortocolHandler {
         String response;
         try {
             response = callback.getResponse();
-            if (!response.equals("")) {
+            if (!"".equals(response)) {
                 logger.debug("Atlona protocol violation - didn't start with an inital empty response: '{}'", response);
             }
         } catch (Exception e) {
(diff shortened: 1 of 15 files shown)


bundles/org.openhab.binding.atlona/src/main/java/org/openhab/binding/atlona/internal/pro3/AtlonaPro3PortocolHandler.java
```diff
@@ -1088,7 +1088,7 @@ private class NormalResponseCallback implements SocketSessionListener {
 
         @Override
         public void responseReceived(String response) {
-            if (response == null || response.isEmpty()) {
+            if (response.isEmpty()) {
```
jlaur, line 1091: This class is not `@NonNullByDefault`, so I'm wondering if this change introduces risk of NPE? I have not checked usages etc., so just question.

mlobstein: It is only used once in SocketChannelSession and is wrapped with a null check there. I removed the null check here to eliminate the redundant null check warning.
socketchanelsession

jlaur: Much better not to call with null. Thanks for clarifying.

jlaur (approved):
Thanks for the fix!

PR #13219, opened 2022-08-05
[boschindego] Fix thing marked as offline when device is reachable

### LINKED ISSUE #13218: [boschindego] Thing offline after executing invalid command
## Expected Behavior
Failure to execute a command should:
- Not affect online status.
- Be logged.
- Result in response payload from PUT request being logged with TRACE level.

## Current Behavior
The binding has logic trying to prevent executing invalid commands. However, this logic is not bulletproof as it tries to mimic logic from the mower/service which is closed source.

When such an invalid command is executed, service will respond with HTTP code 500 and a payload with an error code:

```
{
    "error": 12545
}
```

This is currently handled by `handleCommand` by setting the thing **OFFLINE** with **COMMUNICATION_ERROR**.

After this, next state request might return with success, but thing will remain offline. This is because only _operatingData_ requests should affect online status, since only those actually communicate with the device.

## Possible Solution
Failure to execute commands should not affect online status, but only be logged. Additionally, TRACE logging should be added for PUT response payload.

## Steps to Reproduce
Press the manual stop button on the device while not docked. Then execute a command for returning to dock.

### LINKED ISSUE #13220: [boschindego] Thing offline after hitting rate limit restriction
HTTP code 429 ("Too Many Requests") when requesting state is not handled appropriately.

## Expected Behavior
Thing should be marked as **OFFLINE** with **COMMUNICATION_ERROR** when HTTP code 429 is encountered. On first successful request hereafter it should be marked as **ONLINE** again.

Any kind of throttling, fallback mechanisms, request postponing etc. are not in scope for this bug report.

## Current Behavior

When state request fails with code 429:
```
2022-08-05 17:46:17.649 [INFO ] [ab.event.ThingStatusInfoChangedEvent] - Thing 'boschindego:indego:lawnmower' changed from ONLINE to OFFLINE (COMMUNICATION_ERROR): The request failed with error: 429
```

Thing will not come online again until next successful _operatingState_ request. This request is scheduled for every six hours and additionally every two minutes while device is moving or charging.

## Possible Solution
Keep track of the reason why device went offline. In case of 429 when thing was previously ONLINE, we can safely go back to ONLINE on first successful request without disturbing the _operatingState_ logic.

## Steps to Reproduce
I have only experienced 429 very few times, even during testing, restarts, requests every six seconds for map position etc. I have no knowledge about which restrictions apply, so currently cannot reproduce without manipulating the request handling.

jlaur:
When sending a command which is rejected by the device, the thing would be marked as **OFFLINE** with **COMMUNICATION_ERROR** until the next _operatingData_ request which could take as long as six hours.

This pull request fixes this by:
- Logging a warning without changing thing status.
- Updating channel **errorcode** with any error code received. Please note that for now this is updated in the existing errorcode channel. After next state update, this error code will be reset.

Additionally, PUT/POST response payloads are now logged (with TRACE level) and GET response payload is now also logged in case of HTTP failure.

Fixes #13218

After temporary disruptions like 429 ("Too Many Requests"), revert back to previous "real" online status based on ability to reach the device. In other words:
- If device cannot be reached: **OFFLINE**.
- If state cannot be fetched because of error: **OFFLINE**.
- When state can be fetched again: still **OFFLINE** if previous reachability check failed (description will be updated). Otherwise **ONLINE**.
- When reachability check has been performed: **ONLINE** or **OFFLINE** (disregarding results of other requests).

Fixes #13220

### DIFF
--- a/bundles/org.openhab.binding.boschindego/src/main/java/org/openhab/binding/boschindego/internal/IndegoController.java
+++ b/bundles/org.openhab.binding.boschindego/src/main/java/org/openhab/binding/boschindego/internal/IndegoController.java
@@ -38,6 +38,7 @@ import org.openhab.binding.boschindego.internal.dto.request.SetStateRequest;
 import org.openhab.binding.boschindego.internal.dto.response.AuthenticationResponse;
 import org.openhab.binding.boschindego.internal.dto.response.DeviceCalendarResponse;
 import org.openhab.binding.boschindego.internal.dto.response.DeviceStateResponse;
+import org.openhab.binding.boschindego.internal.dto.response.ErrorResponse;
 import org.openhab.binding.boschindego.internal.dto.response.LocationWeatherResponse;
 import org.openhab.binding.boschindego.internal.dto.response.OperatingDataResponse;
 import org.openhab.binding.boschindego.internal.dto.response.PredictiveLastCuttingResponse;
@@ -236,6 +237,10 @@ public class IndegoController {
             }
             ContentResponse response = sendRequest(request);
             status = response.getStatus();
+            String jsonResponse = response.getContentAsString();
+            if (!jsonResponse.isEmpty()) {
+                logger.trace("JSON response: '{}'", jsonResponse);
+            }
             if (status == HttpStatus.UNAUTHORIZED_401) {
                 // This will currently not happen because "WWW-Authenticate" header is missing; see below.
                 throw new IndegoAuthenticationException("Context rejected");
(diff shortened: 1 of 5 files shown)


jlaur:
I just encountered 429 ("Too Many Requests") which exhibit same behavior, so will provide a fix for that also.

fwolter (approved):

jlaur:
Removing "Bug" label again, since this was a regression from #13040 which is also for first milestone 3.4 release, so the bug was never included in any release. Thus no need to include in release notes.

PR #13223, opened 2022-08-06
[rotel] Other commands added for RSP-1576, RSP-1582, RAP-1580, RDD-1580, RA-1592, RC-1590

lolodomo:


bundles/org.openhab.binding.rotel/src/main/java/org/openhab/binding/rotel/internal/communication/RotelSimuConnector.java
```diff
@@ -44,6 +44,7 @@
 public class RotelSimuConnector extends RotelConnector {
 
     private static final int STEP_TONE_LEVEL = 1;
+    private static final double STEP_DECIBEL = 0.5;
```
jlaur, line 47: Is this consistent across all models and zones and non-configurable on the receiver? For Denon/Marantz step is usually 0.5 dB for main zone and 1 dB for other zones.

lolodomo: This is not the general volume but the adjustment on each channel in a multi channel processor. +/- 0.5 dB is what is mentioned in the API. The API only provides UP and DOWN commands for each channel.
Please note that this constant is defined and only used in my simulator class, so this is not really important.
Even if a real model would use a different step (not the one defined in the Rotel API !), this will change absolutely nothing in the binding. And the resulting value is returned by the API in response to the UP/DOWN command.
These commands are available today only for 3 models: RSP-1576, RSP-1582 and RAP-1580.

lolodomo: In that PR, I just made these existing API commands available to the user, this is a minimal implementation without any display of the current value.

In the future, I could consider adding an OH channel for each of them but I miss an API command to get the current value. I could send UP and then DOWN to retrieve the current value. I just don't know if these commands are always available.

bundles/org.openhab.binding.rotel/src/main/java/org/openhab/binding/rotel/internal/handler/RotelHandler.java
```diff
@@ -150,6 +150,28 @@ public RotelHandler(Thing thing, RotelStateDescriptionOptionProvider stateDescri
...
         logger.debug("Start initializing handler for thing {}", getThing().getUID());
 
+        RotelThingConfiguration config = getConfigAs(RotelThingConfiguration.class);
+
+        protocol = RotelProtocol.HEX;
+        if (config.protocol != null && !config.protocol.isEmpty()) {
+            try {
+                protocol = RotelProtocol.getFromName(config.protocol);
+            } catch (RotelException e) {
+                // Invalid protocol name in configuration, HEX will be considered by default
```
jlaur, line 160: Perhaps this should result in OFFLINE with CONFIGURATION_ERROR? The user might have tried to configure another protocol and made a typo, and now will unexpectedly and silently be using HEX. Having a clear error shown might save some time trouble-shooting this?

jlaur: Now saw that this is moved code, so resolve comment if you prefer not to improve this in context of this PR.

jlaur (approved):
Thanks! I have added one small suggestion, but realized you didn't change this code, so completely optional.

bundles/org.openhab.binding.rotel/src/main/java/org/openhab/binding/rotel/internal/handler/RotelHandler.java
lolodomo: I will check that later but I guess it should be already checked and handled by the core framework (thing configuration validation). So the exception could probably never happened in practice. 
I will create another PR if necessary.

lolodomo:
Thank you

Issue #11508, opened 2021-11-04
[homekit] Add support for additional bridge due to 150 items limitation


IonasElate:
Homekit bridges do support a maximum of 150 items (Apple limitation). 
This number is easily reached when implementing sensors / actuators etc.

The solution would be multibrige support of the homekit addon.

jlaur:
See also #12927

PR #13226, opened 2022-08-06
[homekit] allow multiple bridge instances to break the 150 limit

ccutrer:
Adds a configuration option so that multiple bridges will be created, and metadata to assign accessories to a particular instance.


lolodomo:

bundles/org.openhab.io.homekit/README.md
```diff
@@ -115,6 +116,8 @@ org.openhab.homekit:name=openHAB
 | thermostatTargetModeAuto | Word used for activating the automatic mode of the device (if applicable). It can be overwritten at item level.                                                                                                                                                               | Auto          |
 | thermostatTargetModeOff  | Word used to set the thermostat mode of the device to off (if applicable).  It can be overwritten at item level.                                                                                                                                                             | Off           |
 | name                     | Name under which this HomeKit bridge is announced on the network. This is also the name displayed on the iOS device when searching for available bridges.                                                                               | openHAB       |
+| instances                | Defines how many bridges to expose. Necessary if you have more than 149 accessories. Accessories must be
+				assigned to additional instances via metadata. Additional bridges will use incrementing port numbers. | 1         |
```
lolodomo, line 120: Please do not return at a new line, this will probably break the table.

bundles/org.openhab.io.homekit/README.md
```diff
@@ -836,6 +839,33 @@ Number          cooler_cool_thrs           "Cooler Cool Threshold Temp [%.1f °C
...
+## Multiple Instances
+
+Homekit has a limitation of 150 accessories per bridge. The bridge itself
+counts as an accessory, so in practice it's 149. In order to overcome this
+limitation, you can instruct OpenHAB to expose multiple bridges to
+Homekit, and then manually assign specific accessories to different
+instances. You will need to manually add each additional bridge in the
+Home app, since the QR Code in settings will only be for the primary bridge;
+however the same PIN is still used. In order to assign a particular
+accessory to a different bridge, set the `instance` metadata parameter:
```
lolodomo, line 851: Our convention for maintainability is to return to a new line at the end of a sentence, not inside a sentence.
Please also fix your other addings in README.

lolodomo (changes requested):
Initial review of README

bundles/org.openhab.io.homekit/src/main/java/org/openhab/io/homekit/internal/HomekitChangeListener.java
```diff
@@ -358,35 +357,60 @@ private void createRootAccessories(Item item) {
...
+    }
+
+    private boolean itemIsForThisBridge(Item item, @Nullable Map<String, Object> configuration) {
+        // non-tagged accessories belong to the first instance
+        if (configuration == null) {
+            return (instance == 1);
+        }
+
+        final @Nullable Object value = configuration.get(HomekitTaggedItem.INSTANCE);
+        if (value == null && instance == 1) {
```
yfre, line 403: this and next if can be replaced with 
```
if (value == null) {
            return (instance == 1);
        }
```

bundles/org.openhab.io.homekit/src/main/java/org/openhab/io/homekit/internal/HomekitChangeListener.java
```diff
@@ -358,35 +357,60 @@ private void createRootAccessories(Item item) {
...
+        }
+        if (value == null) {
+            return false;
+        }
+        if (value instanceof Number) {
+            return (instance == ((Number) value).intValue());
         }
+        logger.warn("Unrecognized instance tag {} ({}) for item {}; not creating accessory.", value, value.getClass(),
+                item.getName());
+        return false;
```
yfre, line 414: would it not be better to assign wrong tagged items to first instance? (while keeping the warning)

bundles/org.openhab.io.homekit/src/main/java/org/openhab/io/homekit/internal/HomekitImpl.java
```diff
@@ -69,18 +69,20 @@
...
 
+    private final StorageService storageService;
     private final NetworkAddressService networkAddressService;
     private final ConfigurationAdmin configAdmin;
-    private final Storage<String> storage;
+    private final ItemRegistry itemRegistry;
+    private final MetadataRegistry metadataRegistry;
 
-    private HomekitAuthInfoImpl authInfo;
+    private final List<HomekitAuthInfoImpl> authInfos = new ArrayList<>();
```
yfre, line 78: probably in 99,9% of the cases these 4 new ArrayLists will have size of 1.
so, im wondering whether we should initialize them later in constructor with actual size, e.g.
... ArrayList<>(settings.instances)

not sure how much memory we save.. probably not much.
@lolodomo what do you think?

yfre:
@ccutrer thank you a lot! 
finally we get support for multiple bridges (btw, requested here https://github.com/openhab/openhab-addons/issues/11508)
in general, the design looks good and elegant to me.

only following things i was not sure about:
- homekit servers, bridges, authinfos and listeners are logically connected to each other (1 server -> 1 bridge -> 1 auth info) but they are disconnected in the code. but it is probably more question to Java-HAP that is missing some getters.

- in your proposal you create several registry listeners, one per instance. all of them are get triggered on change of an item and then checking whether they have the correct bridge instance. i was thinking whether having one listener for all bridges, that would manage all registries / bridges and add item to correct one, would not be a better solution. but again, in most of the cases it will be 1 or max just 2-3 listeners/instances. this should not make much difference. so, i would be fine with both option (one listener or multiple)

i see this PR as foundation for some more enhancements, e.g. different PIN and QR codes per bridge, support for "instance" metadate in UI, list accessories link to one bridge, clearPairing for specific bridge, etc. 

i also would like to test whether having many small bridges, e.g. with 10 accessories per bridge, would beneficial comparing to one bridge with 100 accessories. if yes, we could recommend multiple bridges by default and maybe have automated logic to distributed items across the bridges.

but this all feature enhancement

bundles/org.openhab.io.homekit/src/main/java/org/openhab/io/homekit/internal/HomekitChangeListener.java
ccutrer: I don't have strong feelings either way. I initially did it this way because of PTSD of going over the limit and _everything_ disappearing from Home on the phone, and then have to re-assign all the items to their rooms. But yeah, I think you're right it's better to have them show up _somewhere_.

bundles/org.openhab.io.homekit/src/main/java/org/openhab/io/homekit/internal/HomekitImpl.java
ccutrer: I doubt it's worth it. I've been exploring core and the rule engine a lot lately, and there are sooooo many objects allocated, this is barely any noise.

ccutrer:
> * homekit servers, bridges, authinfos and listeners are logically connected to each other (1 server -> 1 bridge -> 1 auth info) but they are disconnected in the code. but it is probably more question to Java-HAP that is missing some getters.

Yeah, I started with just the HomeKit Server, and then had to add the others in parallel. I didn't love it, but wanted to avoid scope creep refactoring all that.

> * in your proposal you create several registry listeners, one per instance. all of them are get triggered on change of an item and then checking whether they have the correct bridge instance. i was thinking whether having one listener for all bridges, that would manage all registries / bridges and add item to correct one, would not be a better solution. but again, in most of the cases it will be 1 or max just 2-3 listeners/instances. this should not make much difference. so, i would be fine with both option (one listener or multiple)

Given that every rule trigger registers as a separate listener in the event system (slightly different use case, but same general idea), I don't think registering lots of registry listeners is a big deal. Especially since it's a constant number of bridges, and not (currently) dynamic.

> i see this PR as foundation for some more enhancements, e.g. different PIN and QR codes per bridge, support for "instance" metadate in UI, list accessories link to one bridge, clearPairing for specific bridge, etc.

+1. It wasn't immediately obvious to me how to get multiple QR codes or PINs to show up in the UI, so just skipped that for now since I have it working without. The metadata UI is all in the openhab-web repo (and beyond my skillset). Extending commands to be instance aware is definitely an easy follow-on PR, along with several other ideas.

> i also would like to test whether having many small bridges, e.g. with 10 accessories per bridge, would beneficial comparing to one bridge with 100 accessories. if yes, we could recommend multiple bridges by default and maybe have automated logic to distributed items across the bridges.

I've thought about dynamically distributing items between bridges, but couldn't come up with a good way at the moment. It's just such a pain that when an accessory disappears from one bridge, and appears on another, you'd have to reconfigure its room (and favorite, status, and sometimes specific icon/type) in Home, which is a massive pain when you have hundreds of accessories. Reducing the "blast radius" to fewer accessories per bridge when you accidentally goof up a config file is also attractive. But I've got ideas in mind to mitigate that issue.

I don't love that instances are just identified by number. But having an explicit mapping of accessories to instance kept things relatively simple. That said, is everyone okay with "instance" (instead of say, "bridge")? And do we want to make it be a string identifier? Both are things that would require people to change their configuration if we change our mind down the road.

bundles/org.openhab.io.homekit/src/main/java/org/openhab/io/homekit/internal/HomekitImpl.java
yfre: > I doubt it's worth it. I've been exploring core and the rule engine a lot lately, and there are sooooo many objects allocated, this is barely any noise.

you are right, we should not try to optimize where it is not required and has very little impact on overall memory usage

yfre (approved):
LGTM from functional perspective

lolodomo (approved):
LGTM

IonasElate:
It works great!
Thank you all for your work!!

PR #13232, opened 2022-08-08
[homekit] fix window coverings based on groups of  rollershutters

ccutrer:
it was getting the state as a decimal type, not a percent type, so
it was getting 1 instead of 100.


lolodomo:

yfre (approved):
LGTM

lolodomo (approved):
Code LGTM

PR #13236, opened 2022-08-09
[homekit] allow group items of the correct type anywhere that type is needed

ccutrer:
this is a theme, and I've extracted a helper method and applied it to several
other locations as well

example, I have a double-sliding door, with two contact sensors
in the alarm system, defined as:

```
Group:Contact:AND(CLOSED,OPEN) DeckDoor_Contact "Deck Door [%s]" <door> (gExteriorDoors, gWarningVisibility_Contact, gDoor, lMainFloor, gInflux) { homekit="ContactSensor" [instance=2] }
Contact DeckNDoor_Contact "Deck Door North" <door> (gInflux, DeckDoor_Contact, gGreatContacts) { channel="dscalarm:zone:46c52f35:zone11:zone_status" }
Switch DeckNDoor_Tamper "Deck Door North Tamper [%s]" <warning> (gTamper, gWarningVisibility_Switch) { channel="dscalarm:zone:46c52f35:zone11:zone_tamper" }
Contact DeckSDoor_Contact "Deck Door South" <door> (gInflux, DeckDoor_Contact, gGreatContacts) { channel="dscalarm:zone:46c52f35:zone12:zone_status" }
Switch DeckSDoor_Tamper "Deck Door South Tamper [%s]" <warning> (gTamper, gWarningVisibility_Switch) { channel="dscalarm:zone:46c52f35:zone12:zone_tamper" }
```

This allows the Group:Contact to be used for the ContactSensor.

### DIFF
--- a/bundles/org.openhab.io.homekit/src/main/java/org/openhab/io/homekit/internal/HomekitTaggedItem.java
+++ b/bundles/org.openhab.io.homekit/src/main/java/org/openhab/io/homekit/internal/HomekitTaggedItem.java
@@ -22,6 +22,15 @@ import org.eclipse.jdt.annotation.NonNullByDefault;
 import org.eclipse.jdt.annotation.Nullable;
 import org.openhab.core.items.GroupItem;
 import org.openhab.core.items.Item;
+import org.openhab.core.library.items.DimmerItem;
+import org.openhab.core.library.items.NumberItem;
+import org.openhab.core.library.items.RollershutterItem;
+import org.openhab.core.library.items.StringItem;
+import org.openhab.core.library.items.SwitchItem;
+import org.openhab.core.library.types.DecimalType;
+import org.openhab.core.library.types.OnOffType;
+import org.openhab.core.library.types.PercentType;
+import org.openhab.core.library.types.StringType;
 import org.openhab.core.types.State;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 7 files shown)


bundles/org.openhab.io.homekit/src/main/java/org/openhab/io/homekit/internal/HomekitTaggedItem.java
```diff
@@ -145,6 +161,78 @@ public HomekitOHItemProxy getProxyItem() {
...
+     */
+    public boolean send(DecimalType command) {
+        if (getItem() instanceof GroupItem && getBaseItem() instanceof NumberItem) {
+            ((GroupItem) getItem()).send(command);
+            return true;
+        } else if (getItem() instanceof NumberItem) {
+            ((NumberItem) getItem()).send(command);
+            return true;
+        }
+        return false;
```
yfre, line 178: not sure about return's. 
as far as i saw the return values are not evaluated by the caller. it cannot do much anyway except of print a warning. maybe we could just print warning here. 

if we want caller to take care, then we should consider throwing an exception. 
not sure whether we have a openHAB guidance on usage of return values vs exceptions

same applies to other send methods.

yfre (approved):
LGTM from functional perspective

not sure only about usage of boolean returns in sends methods (see inline comment)

bundles/org.openhab.io.homekit/src/main/java/org/openhab/io/homekit/internal/HomekitTaggedItem.java
ccutrer: I see where you're coming from. My original intention was to make it so the caller could avoid the type check itself, by doing something like

1)
```
if (send(percentType)) {
   return;
} else if (send(decimalType)) {
  return;
} else {
  logger.warn(...)
}
```

(which I now see could even simplify to, even if a bit opaque):

2)
```
if (!send(percentType) && !send(decimalType)) {
  logger.warn(...);
}
```

But then I ended up not doing it because each of those cases require constructing a new object for the command based on the type, so it would be an unnecessary allocation if it's not that type. I'm open to going any direction here
 * doing 1) above
 * doing 2) above
 * returning void (assume the caller already did the type check)
 * raising an exception in the false case (assume the caller already did the type check)

yfre: im just afraid that at some point the developer implementing the caller will forget to check the return value and nothing will happen in case of wrong type. 

i just checked how such cases are handled in the OH core and GenericItem for instance just print a warning
https://github.com/openhab/openhab-core/blob/6758aacba59478938281249a6ca05265717de772/bundles/org.openhab.core/src/main/java/org/openhab/core/items/GenericItem.java#L443

i would suggest we do the same, just add warning logging before "return false" in the send methods. if the caller done correctly and checks the type before calling "send" we would never see this warning.

ccutrer:
ping @yfre

yfre (approved):
LGTM

yfre:
> ping @yfre

@ccutrer  thank you for the ping. missed the commit. thank you for adding warnings in case of wrong type 
it looks good for me now

ccutrer:
@lolodomo can we get this submitted now?

bundles/org.openhab.io.homekit/src/main/java/org/openhab/io/homekit/internal/HomekitTaggedItem.java
```diff
@@ -145,6 +161,78 @@ public HomekitOHItemProxy getProxyItem() {
...
+        } else if (getItem() instanceof SwitchItem) {
+            ((SwitchItem) getItem()).send(command);
+            return;
+        }
+        logger.warn("Received OnOffType command for item {} that doesn't support it. This is probably a bug.",
+                getName());
+    }
+
+    /**
+     * Send PercentType command to a DimmerItem or RollershutterItem (or a Group:Switch/Group:Rollershutter)
```
lolodomo, line 199: Is it really Group:Switch ?
Maybe more Group:Dimmer ?

lolodomo (changes requested):
Only a very minor comment

bundles/org.openhab.io.homekit/src/main/java/org/openhab/io/homekit/internal/HomekitTaggedItem.java
ccutrer: Absolutely right. Minor brain fart when going through all of these. PR updated.

lolodomo:
This is terrible for a reviewer to not get a delta but the full code when something is fixed. 
I haven't the courage the check again everything!

ccutrer:
> This is terrible for a reviewer to not get a delta but the full code when something is fixed.
> I haven't the courage the check again everything!

Heh, my apologies. Seems like every project/reviewer has their own preferences wrt squashing/fixup commits and rebasing/merging. My own personal preference is to rebase and squash (unless there are atomic separate commits) every time you update, but I'll definitely agree that GitHub can be difficult to review such a workflow. I've used Gerrit, Google's review tool, a lot in the past, which is much better at that workflow (but worse at branching workflows). I'm more than happy to abide by whatever a particular project's preference is, once I know about it. Sounds like that's new commits for any changes to PR. What about rebasing vs. merging?

lolodomo:
The clear recommendation here is to never squash a PR while the review process has started.

PR #13238, opened 2022-08-09
[gardena] Fixed connection tracker

gerrieg:
The connection tracker does not start after some websocket connections are disconnected. Therefore, every 2 minutes some API requests are sent to Gardena. This can lead to the message 429 limit exceeded after a few days/weeks.

### DIFF
--- a/bundles/org.openhab.binding.gardena/src/main/java/org/openhab/binding/gardena/internal/GardenaSmartWebSocket.java
+++ b/bundles/org.openhab.binding.gardena/src/main/java/org/openhab/binding/gardena/internal/GardenaSmartWebSocket.java
@@ -87,7 +87,6 @@ public class GardenaSmartWebSocket {
         final ScheduledFuture<?> connectionTracker = this.connectionTracker;
         if (connectionTracker != null) {
             connectionTracker.cancel(true);
-            this.connectionTracker = null;
         }
 
         logger.debug("Closing Gardena Webservice ({})", socketId);
@@ -123,7 +122,6 @@ public class GardenaSmartWebSocket {
         ScheduledFuture<?> connectionTracker = this.connectionTracker;
         if (connectionTracker != null && !connectionTracker.isCancelled()) {
             connectionTracker.cancel(true);
-            this.connectionTracker = null;
         }
 
         // start sending PING every two minutes
--- a/bundles/org.openhab.binding.gardena/src/main/java/org/openhab/binding/gardena/internal/GardenaSmartWebSocket.java
+++ b/bundles/org.openhab.binding.gardena/src/main/java/org/openhab/binding/gardena/internal/GardenaSmartWebSocket.java
@@ -87,6 +87,7 @@ public class GardenaSmartWebSocket {
         final ScheduledFuture<?> connectionTracker = this.connectionTracker;
         if (connectionTracker != null) {
             connectionTracker.cancel(true);
+            this.connectionTracker = null;
         }
 
         logger.debug("Closing Gardena Webservice ({})", socketId);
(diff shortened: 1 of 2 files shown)


andrewfg:
@gerrieg is this PR related to this post on the forum?

https://community.openhab.org/t/gardena-error-429-limit-exceeded/137419/62?u=andrewfg

bundles/org.openhab.binding.gardena/src/main/java/org/openhab/binding/gardena/internal/GardenaSmartWebSocket.java
```diff
@@ -121,11 +122,12 @@ public void onConnect(Session session) {
 
         ScheduledFuture<?> connectionTracker = this.connectionTracker;
         if (connectionTracker != null && !connectionTracker.isCancelled()) {
-            connectionTracker.cancel(false);
+            connectionTracker.cancel(true);
+            this.connectionTracker = null;
```
andrewfg, line 126: Because of line 130 this line (126) is not needed.

bundles/org.openhab.binding.gardena/src/main/java/org/openhab/binding/gardena/internal/GardenaSmartWebSocket.java
```diff
@@ -121,11 +122,12 @@ public void onConnect(Session session) {
...
         }
 
         // start sending PING every two minutes
-        this.connectionTracker = scheduler.scheduleWithFixedDelay(this::sendKeepAlivePing, 2, 2, TimeUnit.MINUTES);
+        this.connectionTracker = scheduler.scheduleWithFixedDelay(this::sendKeepAlivePing, 1, 2, TimeUnit.MINUTES);
```
andrewfg, line 130: Yes. Makes sense.

andrewfg (changes requested):

bundles/org.openhab.binding.gardena/src/main/java/org/openhab/binding/gardena/internal/GardenaSmartWebSocket.java
jlaur: Something is fishy. Nullifying and reassigning should make no difference for the Garbage Collector, in both cases the object is eligible for GC.

I'm wondering if actually the bug could be something else? For example, previously `cancel()` was called without interrupting the thread, so it could still be running, but marked as cancelled. On next call another thread could be spawned, while old one is still running (but without any references to it). This could happen a few times depending on how long these threads would continue running. At some point, could this lead to a problem, or is it too theoretical? Just adding some lose thoughts, hopefully they will trigger new ideas and not mislead too much. :)

andrewfg: @jlaur I share your concerns; I see that @gerrieg also changed `cancel(false)` to `cancel(true)` so that may be his actual “fix”; ?? .. but I think it is important that we understand what is really going on here; let me think about this further; I will get back tomorrow..

andrewfg: > explicitly setting it to null helps the garbage collector and the scheduled pool

I am pretty sure the above hypothesis is wrong: the binding code only nulls an external reference to the task; the task has already been submitted to the OH scheduler, and that holds its own internal reference until the task is finished; so nulling this external reference cannot reduce the reference count sufficiently to trigger the GC.

> the connection tracker will not start after some restarts.

So we need to get to the root of why this happens. (And more specifically why you see it on on your system, but I do not see it on my own system).

Maybe the scheduled task method `sendKeepAlivePing()` is blocking in some way; perhaps on a TCP/IP IO call. In that case, the change from `cancel(false)` to `cancel(true)` might already be sufficient to fix things (because it would interrupt the IO call on that respective thread). 

It also occurs to me that the `sendKeepAlivePing()` could potentially have been called on two tasks from two separate scheduled task series -- the first one having blocked in the first task series, and the second one therefore not being thread safe. So to prevent that we should (in any case) synchronize the scheduled task method i.e. `private synchronized void sendKeepAlivePing()`.

Note: if there were some IO blocking going on, then synchronize would actually make the error more obvious, but at least it would a first step towards identifing what might be going on underneath..

lolodomo:
I am ok to merge the PR as it is, except if someone still thinks `cancel` should no be called with true as parameter.
The timing is now very short if we want to include this fix in 3.4M1.
I will merge it in the middle of the afternoon if noone reacts.

lolodomo (approved):

andrewfg:
> except if someone still thinks cancel should no be called with true as parameter

No problem with that. But the task method should be synchronized..

andrewfg:
@lolodomo you need to merge #13253 as well

jlaur:
@gerrieg, @andrewfg - can you of you change the PR title so it reflects the change from user-perspective? The title will be part of the release notes, and I don't think users will understand what connection tracker is. Alternatively remove the bug label if this fix is something additional for another PR already merged into 3.4, thus already covered in release notes?

PR #13240, opened 2022-08-09
[rotel] Set model and firmware properties

lolodomo:
For all models providing these information and relying on ASCII protocol.

The binding now supports reading of status message containing variable length content.


lolodomo:
@jlaur : any chance to merge this PR, so that I can go to the next steps ?

bundles/org.openhab.binding.rotel/src/main/java/org/openhab/binding/rotel/internal/handler/RotelHandler.java
```diff
@@ -1696,9 +1696,11 @@ public void onNewMessageEvent(EventObject event) {
                 case KEY_PCUSB_CLASS:
                     logger.debug("PC-USB Audio Class is set to {}", value);
                     break;
+                case KEY_PRODUCT_TYPE:
```
jlaur, line 1699: What is the difference between `KEY_PRODUCT_TYPE` and `KEY_MODEL` (as well as `KEY_PRODUCT_VERSION` vs. `KEY_VERSION`)? Do they belong to different protocols so they cannot both be received, which would then conflict?

lolodomo: Yes, two different ASCII protocols.

bundles/org.openhab.binding.rotel/src/main/java/org/openhab/binding/rotel/internal/protocol/ascii/RotelAsciiV2ProtocolHandler.java
```diff
@@ -35,15 +36,22 @@ public class RotelAsciiV2ProtocolHandler extends RotelAbstractAsciiProtocolHandl
...
     private static final char CHAR_END_RESPONSE = '$';
 
+    private static final Set<String> KEYSET = Set.of(KEY_DISC_NAME, KEY_DISC_TYPE, KEY_TRACK_NAME, KEY_TIME, KEY_FM_RDS,
+            KEY_DAB_STATION);
+
     private final Logger logger = LoggerFactory.getLogger(RotelAsciiV2ProtocolHandler.class);
 
+    private boolean searchKey = true;
+    private boolean variableLength;
+    private boolean prevIsEndCharacter;
```
jlaur, line 46: Forgive me, but too tired right now to understand algorithm/parser below: Why are these instance variables?

bundles/org.openhab.binding.rotel/src/main/java/org/openhab/binding/rotel/internal/handler/RotelHandler.java
```diff
@@ -1944,6 +1946,13 @@ private void schedulePowerOnJob() {
                                 sendCommand(RotelCommand.SPEAKER);
                                 Thread.sleep(SLEEP_INTV);
                             }
+                            if (model != RotelModel.RAP1580 && model != RotelModel.RSP1576
+                                    && model != RotelModel.RSP1582) {
```
jlaur, line 1950: What is special for these three devices - no support for getting model/version? If this is the case, do you know that all other receivers support this - and if not, what will be the impact of sending and waiting for these commands?

lolodomo: Yes, the API documentation for these models do not mention these commands. It could be a simple miss in the documentation but I prefer to respect the Rotel documentation.
Of course, I checked the documentation for all other models and these commands are there.
I have no idea what could be the result of sending these commands if not supported. Maybe they will be simply ignored or maybe an error could be returned, I don't know. That's why I prefer to not send the command.

bundles/org.openhab.binding.rotel/src/main/java/org/openhab/binding/rotel/internal/protocol/ascii/RotelAsciiV2ProtocolHandler.java
lolodomo: In case the reading thread handles the response in multiple "blocs".

bundles/org.openhab.binding.rotel/src/main/java/org/openhab/binding/rotel/internal/protocol/ascii/RotelAsciiV2ProtocolHandler.java
```diff
@@ -116,6 +124,44 @@ public byte[] buildCommandMessage(RotelCommand cmd, @Nullable Integer value) thr
...
+                handleIncomingMessage(getDataBuffer());
+                resetDataBuffer();
+                searchKey = true;
+                variableLength = false;
+                prevIsEndCharacter = false;
+            } else {
+                prevIsEndCharacter = inDataBuffer[i] == CHAR_END_RESPONSE;
+            }
+        }
+    }
```
jlaur, line 163: As far as I can understand from briefly looking at this method, it supports "variable length" in addition to the V1 parser. Would it be feasible to support the V1 algorithm by providing some option? Then the two parsers could be merged and redundancy eliminated.

bundles/org.openhab.binding.rotel/src/main/java/org/openhab/binding/rotel/internal/protocol/ascii/RotelAsciiV2ProtocolHandler.java
jlaur: So the order is guaranteed and there is no need to "reset" when reaching the end of the full response?

jlaur:
@lolodomo - I posted a few single comments to speed up the review. EOF now. :)

bundles/org.openhab.binding.rotel/src/main/java/org/openhab/binding/rotel/internal/protocol/ascii/RotelAsciiV2ProtocolHandler.java
lolodomo: In fact, the variable length implementation is totally different in the two versions.
In V1, this will look like: key=7,my text
In V2, this will look like! key=my text$$
For fixed length, the V1 syntax is: key=value!
and the V2 syntax is: key=value$

bundles/org.openhab.binding.rotel/src/main/java/org/openhab/binding/rotel/internal/protocol/ascii/RotelAsciiV2ProtocolHandler.java
lolodomo: There is a reset done in the code (when calling `resetDataBuffer()`) to start with the handling of a new response.

jlaur (approved):
LGTM - thanks!

lolodomo:
Thank you a lot @jlaur

PR #13241, opened 2022-08-10
[sonos] Ignore TRANSITIONING device state when updating channels

lolodomo:


lolodomo:
Here is a jar for testing (renamed the file with extension jar first):
[org.openhab.binding.sonos-3.4.0-SNAPSHOT.zip](https://github.com/openhab/openhab-addons/files/9302829/org.openhab.binding.sonos-3.4.0-SNAPSHOT.zip)

jlaur (approved):
Thanks for the fix!

ecser99:
The issue is still open on version 3.4.3

Retested the  issue described here today step by step on version 3.4.3. with the same result (the error or unexpected behaviour is still there).
https://community.openhab.org/t/inconsistent-state-reported-by-sonos-state-channels/137983/6

After step 5. both my players  report the status in OpenHAB: "stopped" (unexpected behaviour).
Exepected behaviour: the player, which is still playing should continue to report a PLAYING state in OpenHAB.
Additional info: The player, which is actually playing in step 5.  gets the correct status "playing" in OpenHAB later, when a new item starts playing in the actual playlist. 
I use SONOS S2 players, firmware updated.

PR #13242, opened 2022-08-10
[jdbc] Upgrade MySQL JDBC driver to 8.0.30

jlaur:
See
- https://dev.mysql.com/doc/relnotes/connector-j/8.0/en/
- https://mvnrepository.com/artifact/mysql/mysql-connector-java/8.0.30

Tested by installing compiled bundle as well as mysql-connector into addons directory:
```
openhabian@openhabian:/usr/share/openhab/addons $ ll *mysql* *jdbc*
-rw-r--r-- 1 openhabian openhab 2.4M Jul  1 05:18 mysql-connector-java.jar
-rw-r--r-- 1 openhabian openhab 296K Aug 10 22:26 org.openhab.persistence.jdbc-3.4.0-SNAPSHOT.jar
```

Then checking in database that items are persisted after the upgrade.

### DIFF
--- a/bundles/org.openhab.persistence.jdbc/README.md
+++ b/bundles/org.openhab.persistence.jdbc/README.md
@@ -13,7 +13,7 @@ The following databases are currently supported and tested:
 | [H2](https://www.h2database.com/)            | [h2-1.4.191.jar](https://mvnrepository.com/artifact/com.h2database/h2) |
 | [HSQLDB](http://hsqldb.org/)                 | [hsqldb-2.3.3.jar](https://mvnrepository.com/artifact/org.hsqldb/hsqldb) |
 | [MariaDB](https://mariadb.org/)              | [mariadb-java-client-1.4.6.jar](https://mvnrepository.com/artifact/org.mariadb.jdbc/mariadb-java-client) |
-| [MySQL](https://www.mysql.com/)              | [mysql-connector-java-5.1.39.jar](https://mvnrepository.com/artifact/mysql/mysql-connector-java) |
+| [MySQL](https://www.mysql.com/)              | [mysql-connector-java-8.0.30.jar](https://mvnrepository.com/artifact/mysql/mysql-connector-java) |
 | [PostgreSQL](https://www.postgresql.org/)    | [postgresql-42.4.1.jar](https://mvnrepository.com/artifact/org.postgresql/postgresql) |
 | [SQLite](https://www.sqlite.org/)            | [sqlite-jdbc-3.16.1.jar](https://mvnrepository.com/artifact/org.xerial/sqlite-jdbc) |
 | [TimescaleDB](https://www.timescale.com/)    | [postgresql-42.4.1.jar](https://mvnrepository.com/artifact/org.postgresql/postgresql) |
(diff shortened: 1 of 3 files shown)


jlaur:
Investigating rule errors that started to show:
`2022-08-10 23:15:26.875 [ERROR] [internal.handler.ScriptActionHandler] - Script execution of rule with UID 'danfoss-7' failed: class java.time.LocalDateTime cannot be cast to class java.sql.Timestamp (java.time.LocalDateTime is in module java.base of loader 'bootstrap'; java.sql.Timestamp is in module java.sql of loader 'platform') in danfoss`

It might be related to the way I installed the bundle, so will now try to install 8.0.28 the same way.

jlaur:
> It might be related to the way I installed the bundle, so will now try to install 8.0.28 the same way.

It happens also when using version from main branch with connector 8.0.28 installed the same way. So something is wrong with the way I'm testing. Good news is that so far I have not found any issues (besides this one), bad news is that I have not been able to fully test this upgrade. Items are persisted when updated and restored on startup - that much I was able to verify.

jlaur:
@kaikreuzer - should mysql-connector-java also be uploaded to jfrog for this upgrade to work?

kaikreuzer:
@jlaur I guess so, yes, and I just uploaded it.
What is weird, though, is the fact that 8.0.28 wasn't uploaded before, I can only see version 8.0.15.

kaikreuzer (approved):
I do not think your errors are related to this upgrade.
Moving two digits higher on the patch level should hopefully not have any negative impact, so I'd suggest to merge.

jlaur:
@kaikreuzer - unfortunately I'm now seeing this after upgrading to M2:

`2022-09-11 22:35:02.679 [ERROR] [internal.handler.ScriptActionHandler] - Script execution of rule with UID 'danfoss-7' failed: class java.time.LocalDateTime cannot be cast to class java.sql.Timestamp (java.time.LocalDateTime is in module java.base of loader 'bootstrap'; java.sql.Timestamp is in module java.sql of loader 'platform') in danfoss`

I'll need some time to understand this problem. Any ideas would be much appreciated.

kaikreuzer:
@jlaur I just had a look in the code and the only place where there is a cast to `Timestamp` seems to be 
https://github.com/openhab/openhab-addons/blob/ddd25316d9c59e3e6016715d807b90c4e3242e58/bundles/org.openhab.persistence.jdbc/src/main/java/org/openhab/persistence/jdbc/db/JdbcBaseDAO.java#L566-L572

Maybe the lib changed to now return `LocalDateTime` instead of `Timestamp`? Possibly, adding another if statement in the method above for dealing with `LocalDateTime` instances could fix the issue?

jlaur:
Thank you @kaikreuzer for pointing to this place in the code, it looks quite related indeed. I'll try to fix and test this in the evening. I also found this article: https://dev.mysql.com/blog-archive/support-for-date-time-types-in-connector-j-8-0/ which perhaps provides some more context. The POM upgrade was 8.0.28 to 8.0.30, but actually the feature.xml was at 8.0.22, so before 8.0.23.

jlaur:
@kaikreuzer - you hit the nail on the head. Thanks a lot! I have created #13382 for fixing this problem.

kaikreuzer:
Excellent - glad I could help!

PR #13245, opened 2022-08-11
[sonos] Unit tests added for OPML request parsing

lolodomo:

### DIFF
--- a/bundles/org.openhab.binding.sonos/src/main/java/org/openhab/binding/sonos/internal/handler/SonosMediaInformation.java
+++ b/bundles/org.openhab.binding.sonos/src/main/java/org/openhab/binding/sonos/internal/handler/SonosMediaInformation.java
@@ -12,7 +12,6 @@
  */
 package org.openhab.binding.sonos.internal.handler;
 
-import java.io.IOException;
 import java.util.ArrayList;
 import java.util.List;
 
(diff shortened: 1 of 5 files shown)


bundles/org.openhab.binding.sonos/pom.xml
```diff
@@ -14,4 +14,13 @@
...
   <name>openHAB Add-ons :: Bundles :: Sonos Binding</name>
 
+  <dependencies>
+    <dependency>
+      <groupId>org.mockito</groupId>
+      <artifactId>mockito-inline</artifactId>
+      <version>3.6.28</version>
+      <scope>test</scope>
+    </dependency>
+  </dependencies>
```
jlaur, line 24: We should not use **mockito-inline**, please see https://github.com/openhab/openhab-addons/pull/12061#issuecomment-1016226346.

jlaur (changes requested):
For now just quick comment about usage of mockito-inline.

bundles/org.openhab.binding.sonos/pom.xml
lolodomo: How can I mock a static method without that (HttpUtil.executeUrl)?

lolodomo: I will rather move the HTTP code outside the method I want to test.

lolodomo: Changed.

jlaur: Even better.

jlaur (approved):
LGTM

PR #13251, opened 2022-08-12
[Konnected] Adding base URL to config to allow for manually adding thing

haavar:
This PR will allow for things to be manually added, and for the address of the konnected panel to be updated.

Before this change there is no configuration for the address of the konnected panel. If a panel is manually added the thing is not functional, and there is a null pointer exception on initialization.

I renamed the fields `ipAddress` and `host` to `baseUrl` for clarity. The field contains a URL, and not an IP address. I'm not sure if this breaks backwards compatibility, but recreation of the konnected things is anyway needed for openhab version 3.4.0.

### DIFF
--- a/bundles/org.openhab.binding.konnected/src/main/java/org/openhab/binding/konnected/internal/KonnectedBindingConstants.java
+++ b/bundles/org.openhab.binding.konnected/src/main/java/org/openhab/binding/konnected/internal/KonnectedBindingConstants.java
@@ -36,11 +36,11 @@ public class KonnectedBindingConstants {
     public static final ThingTypeUID THING_TYPE_PROMODULE = new ThingTypeUID(BINDING_ID, PRO_MODULE);
 
     // Thing config properties
-    public static final String BASE_URL = "base_url";
+    public static final String BASE_URL = "baseUrl";
     public static final String MAC_ADDR = "macAddress";
-    public static final String REQUEST_TIMEOUT = "request_timeout";
-    public static final String RETRY_COUNT = "retry_count";
-    public static final String CALLBACK_URI = "callback_uri";
+    public static final String REQUEST_TIMEOUT = "requestTimeout";
+    public static final String RETRY_COUNT = "retryCount";
+    public static final String CALLBACK_URL = "callbackUrl";
 
     // ESP8266_ZONE_TO_PIN map, this maps a zone to a pin for ESP8266 based devices
     // Source: https://help.konnected.io/support/solutions/articles/32000026808-zone-to-gpio-pin-mapping
--- a/bundles/org.openhab.binding.konnected/src/main/java/org/openhab/binding/konnected/internal/KonnectedBindingConstants.java
+++ b/bundles/org.openhab.binding.konnected/src/main/java/org/openhab/binding/konnected/internal/KonnectedBindingConstants.java
@@ -36,7 +36,7 @@ public class KonnectedBindingConstants {
     public static final ThingTypeUID THING_TYPE_PROMODULE = new ThingTypeUID(BINDING_ID, PRO_MODULE);
 
     // Thing config properties
-    public static final String HOST = "ipAddress";
+    public static final String BASE_URL = "base_url";
     public static final String MAC_ADDR = "macAddress";
     public static final String REQUEST_TIMEOUT = "request_timeout";
     public static final String RETRY_COUNT = "retry_count";
--- a/bundles/org.openhab.binding.konnected/src/main/java/org/openhab/binding/konnected/internal/handler/KonnectedHandler.java
+++ b/bundles/org.openhab.binding.konnected/src/main/java/org/openhab/binding/konnected/internal/handler/KonnectedHandler.java
@@ -56,7 +56,7 @@ public class KonnectedHandler extends BaseThingHandler {
     private final String konnectedServletPath;
     private final KonnectedHTTPUtils http = new KonnectedHTTPUtils(30);
     private String callbackIpAddress = null;
-    private String baseUri;
+    private String baseUrl;
     private final Gson gson = new GsonBuilder().create();
     private int retryCount;
     private final String thingID;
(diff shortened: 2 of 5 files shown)


bundles/org.openhab.binding.konnected/src/main/java/org/openhab/binding/konnected/internal/handler/KonnectedHandler.java
```diff
@@ -288,7 +289,9 @@ public void initialize() {
         } catch (ConfigValidationException e) {
             updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.CONFIGURATION_ERROR, e.getMessage());
         }
-        this.moduleIpAddress = this.getThing().getProperties().get(HOST).toString();
+        if (baseUri == null) {
+            this.baseUri = this.getThing().getProperties().get(BASE_URL);
+        }
```
lolodomo, line 294: New code to be removed, this is not a property but a configuration parameter.
You are getting the parameter and you set the member `baseUrl` in method `checkConfiguration`.
What you could/should do is to check in `checkConfiguration` that the config parameter is not null and if it is, you should trigger a `ConfigValidationException` exception

bundles/org.openhab.binding.konnected/src/main/java/org/openhab/binding/konnected/internal/handler/KonnectedHandler.java
```diff
@@ -177,6 +177,7 @@ private void checkConfiguration() throws ConfigValidationException {
         Configuration testConfig = this.getConfig();
         String testRetryCount = testConfig.get(RETRY_COUNT).toString();
         String testRequestTimeout = testConfig.get(REQUEST_TIMEOUT).toString();
+        baseUri = testConfig.get(BASE_URL) != null ? testConfig.get(BASE_URL).toString() : null;
```
lolodomo, line 180: Cjheck here that the value is not null and if it is then trigger a `ConfigValidationException` exception

bundles/org.openhab.binding.konnected/src/main/java/org/openhab/binding/konnected/internal/handler/KonnectedHandler.java
```diff
@@ -56,7 +56,7 @@ public class KonnectedHandler extends BaseThingHandler {
     private final String konnectedServletPath;
     private final KonnectedHTTPUtils http = new KonnectedHTTPUtils(30);
     private String callbackIpAddress = null;
-    private String moduleIpAddress;
+    private String baseUri;
```
lolodomo, line 59: To avoid compilation warnings, you could initialize it with an empty string.

lolodomo (changes requested):

bundles/org.openhab.binding.konnected/src/main/java/org/openhab/binding/konnected/internal/handler/KonnectedHandler.java
lolodomo: With the new config validation by the core framework done before initializing the thing handler, I am not sure this is even necessary. I mean, as you defined the parameter as required, I believe your thing handler will not be initialized if the config parameter is not set. Maybe you can check that. As a result, you would get always a non null value here and you have no check to add or exception to trigger.

bundles/org.openhab.binding.konnected/src/main/java/org/openhab/binding/konnected/internal/handler/KonnectedHandler.java
lolodomo: As finally mentioned in another comment, you probably do not need to add any check.

Just remove these useless lines ad `baseUri` is set with a non null value in `checkConfiguration`.

bundles/org.openhab.binding.konnected/src/main/java/org/openhab/binding/konnected/internal/KonnectedBindingConstants.java
```diff
@@ -36,7 +36,7 @@ public class KonnectedBindingConstants {
     public static final ThingTypeUID THING_TYPE_PROMODULE = new ThingTypeUID(BINDING_ID, PRO_MODULE);
 
     // Thing config properties
-    public static final String HOST = "ipAddress";
+    public static final String BASE_URL = "base_url";
```
jlaur, line 39: Since backwards compatibility is already broken, could this be **camelCase** like ipAddress was and macAddress still is?

lolodomo: I have the same remark until I see that others are not camelCase !

jlaur: This could be the opportunity for renaming them too. :)

bundles/org.openhab.binding.konnected/src/main/java/org/openhab/binding/konnected/internal/handler/KonnectedHandler.java
haavar: I'm not getting any warnings for this. I prefer to have it be uninitialized and have any bugs related to it not set be obvious.

bundles/org.openhab.binding.konnected/src/main/java/org/openhab/binding/konnected/internal/handler/KonnectedHandler.java
haavar: Thanks for the feedback. I did not realize the properties from the discovery would carry over to the configuration. I simplified the code a bit.

bundles/org.openhab.binding.konnected/src/main/java/org/openhab/binding/konnected/internal/handler/KonnectedHandler.java
haavar: Thanks for the feedback. Updated.

bundles/org.openhab.binding.konnected/src/main/java/org/openhab/binding/konnected/internal/KonnectedBindingConstants.java
haavar: I renamed all config parameters to camel case for consistency.

lolodomo (approved):
LGTM, thank you

lolodomo:
You already prepared an alert for the release notes, fine. Maybe you can adjust a little the description to be more in line with this new PR ?

jlaur:
> You already prepared an alert for the release notes, fine. Maybe you can adjust a little the description to be more in line with this new PR ?

Also the README should be updated. At least things example where ipAddress is still mentioned.

PR #13255, opened 2022-08-12
[transform.jruby] new Ruby transformation

ccutrer:
mostly based off the JS one, but had to do some finagling with
bindings to get it to work right with cached scripts. I also
took out the profile stuff for now - it can be added back in in
another PR.

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -381,6 +381,7 @@
 /bundles/org.openhab.transform.exec/ @openhab/add-ons-maintainers
 /bundles/org.openhab.transform.javascript/ @openhab/add-ons-maintainers
 /bundles/org.openhab.transform.jinja/ @jochen314
+/bundles/org.openhab.transform.jruby/ @boc-tothefuture
 /bundles/org.openhab.transform.jsonpath/ @clinique
 /bundles/org.openhab.transform.map/ @openhab/add-ons-maintainers
 /bundles/org.openhab.transform.regex/ @openhab/add-ons-maintainers
--- a/bom/openhab-addons/pom.xml
+++ b/bom/openhab-addons/pom.xml
@@ -1908,6 +1908,11 @@
     </dependency>
     <dependency>
       <groupId>org.openhab.addons.bundles</groupId>
+      <artifactId>org.openhab.transform.jruby</artifactId>
+      <version>${project.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>org.openhab.addons.bundles</groupId>
       <artifactId>org.openhab.transform.jsonpath</artifactId>
       <version>${project.version}</version>
     </dependency>
--- a/bundles/org.openhab.transform.jruby/NOTICE
+++ b/bundles/org.openhab.transform.jruby/NOTICE
@@ -1,0 +1,13 @@
+This content is produced and maintained by the openHAB project.
+
+* Project home: https://www.openhab.org
+
+== Declared Project Licenses
+
+This program and the accompanying materials are made available under the terms
+of the Eclipse Public License 2.0 which is available at
+https://www.eclipse.org/legal/epl-2.0/.
+
+== Source Code
+
+https://github.com/openhab/openhab-addons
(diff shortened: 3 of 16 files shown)


ccutrer:
cc @boc-tothefuture @jimtng @lolodomo

J-N-K:
Why do you need that? Core supports a SCRIPT transformation that can handle all scripting languages.

ccutrer:
well that's good to know. except AFAICT it's not documented anywhere. it took me quite a bit of googling to even find it - "openhab script transformation" just gives you the JS one. I had to search for ScriptTransformationService to find the actual class.

I'm also worried that it won't work for ruby. I had to do the bindings finagling because when you pre-compile a script with JRuby, it's not seeing the current value of the variables at execution time, only from compilation time. I'll play with this some, though.

J-N-K:
Then we should work on fixing this instead of introducing a new add-on. Why is it that the values are not updated? Can you provide an example script that shows this behavior (I know nothing about ruby, so it's hard for me to test that on my own)?

ccutrer:
I agree. When I didn't know SCRIPT existed, I figured it would be easy enough to work around idiosyncrasies of JRuby. But it'd definitely be better to be able to use the generic transform.

The problem has to do with bindings, pre-compilation, and persistence of variables. It seems that when pre-compiling a ruby script, local variables are captured at compile time, not execution time. I worked around this by using instance variables (effectively globals, but ruby has a special naming convention for determining the scope of a variable). So when the script is pre-compiled, `input` in your script will throw an error because it's not defined when you run it. The simplest workaround would be to not pre-compile the ruby. The other work around is to use instance variables (and additionally I had to construct a particular execution context to ensure all other local variables are cleared between executions). I still need to test with SCRIPT to ensure it behaves the same, since it sets up the local variables slightly differently. Then I'll dig into JRuby to see if it's considered a (fixable) bug on that side (I've done a few PRs over there too).

Tangentially - any suggestion on where to document the SCRIPT transform? It seems to have gotten lost because it's in core, and not its own bundle. So maybe just the main documentation for transformations? And finally, should we be deprecating the JS transform in favor of SCRIPT? I'm not sure if SCRIPT implements the profile transformation as well, like JS does.

J-N-K:
The profile part is still WIP (see https://github.com/openhab/openhab-core/pull/2872), unfortunately it is stalled.

From my understanding compiled scripts should still have access to the variables (global scope/engine scope). Probably a solution for https://github.com/jruby/jruby/issues/5876 is what is needed here. Preventing compilation would also be possible but comes with performance penalties.

ccutrer:
I was able to find sufficient workarounds for jruby to get this to work. For posterity, the steps are:

 * Install the jrubyscripting add-on, which will register jruby as a scripting language with openhab's infrastructure.
 * Create a ruby file in `transform` directory _named_ `mytransform.script` (it's important that the extension is `script`, not `rb`)
 * In your ruby file, before doing any other variable assignments, do `input ||= nil`; this works around https://github.com/jruby/jruby/issues/5876 (and doing it first works around another bug that I haven't nailed down yet).
 * For your formatting definition, do `SCRIPT(rb:mytransform.script):%s`. The `rb` is how jruby registers itself.

An example transform:

`compass.script`
```ruby
# frozen_string_literal: true

input ||= nil
directions = %w[N NE E SE S SW W NW N].freeze

if input.nil? || input == "NULL" || input == "UNDEF"
  "-"
else
  cardinal = directions[(input.to_f / 45).round]
  "#{cardinal} (#{input.to_i}°)"
end
```

`weather.items
```
Number:Angle Exterior_WindDirection "Wind Direction [SCRIPT(rb:compass.script):%s]" <wind>
```

J-N-K:
Impressive. Can you add that to the jrubyscripting add-on documentation? I believe that language specific things should go there.

ccutrer:
Yup, I'm planning on it.

ccutrer:
https://github.com/openhab/openhab-addons/pull/13286

PR #13259, opened 2022-08-13
[rollershutterposition] Initial contribution

jsjames:
# Rollershutter Position emulation

# Description

This transform will emulate setting of absolute position for rollershutters which only support Up/Down/Stop commands.  By specifying the time it takes for the rollershutter to fully extend or close, the transform will estimate the time it will take for the rollershutter to move to an absolute position.

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -399,8 +399,8 @@
 /bundles/org.openhab.transform.jsonpath/ @clinique
 /bundles/org.openhab.transform.map/ @openhab/add-ons-maintainers
 /bundles/org.openhab.transform.regex/ @openhab/add-ons-maintainers
-/bundles/org.openhab.transform.scale/ @clinique
 /bundles/org.openhab.transform.rollershutterposition/ @jsjames
+/bundles/org.openhab.transform.scale/ @clinique
 /bundles/org.openhab.transform.xpath/ @openhab/add-ons-maintainers
 /bundles/org.openhab.transform.xslt/ @openhab/add-ons-maintainers
 /bundles/org.openhab.voice.googlestt/ @GiviMAD
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -400,6 +400,7 @@
 /bundles/org.openhab.transform.map/ @openhab/add-ons-maintainers
 /bundles/org.openhab.transform.regex/ @openhab/add-ons-maintainers
 /bundles/org.openhab.transform.scale/ @clinique
+/bundles/org.openhab.transform.rollershutterposition/ @jsjames
 /bundles/org.openhab.transform.xpath/ @openhab/add-ons-maintainers
 /bundles/org.openhab.transform.xslt/ @openhab/add-ons-maintainers
 /bundles/org.openhab.voice.googlestt/ @GiviMAD
--- a/bundles/org.openhab.transform.rollershutterposition/pom.xml
+++ b/bundles/org.openhab.transform.rollershutterposition/pom.xml
@@ -1,0 +1,17 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+  xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd">
+
+  <modelVersion>4.0.0</modelVersion>
+
+  <parent>
+    <groupId>org.openhab.addons.bundles</groupId>
+    <artifactId>org.openhab.addons.reactor.bundles</artifactId>
+    <version>3.4.0-SNAPSHOT</version>
+  </parent>
+
+  <artifactId>org.openhab.transform.rollershutterposition</artifactId>
+
+  <name>openHAB Add-ons :: Bundles :: Transformation Service :: Roller Shutter Position</name>
+
+</project>
--- a/bundles/org.openhab.transform.rollershutterposition/pom.xml
+++ b/bundles/org.openhab.transform.rollershutterposition/pom.xml
@@ -7,7 +7,7 @@
   <parent>
     <groupId>org.openhab.addons.bundles</groupId>
     <artifactId>org.openhab.addons.reactor.bundles</artifactId>
-    <version>3.4.0-SNAPSHOT</version>
+    <version>4.0.0-SNAPSHOT</version>
   </parent>
 
   <artifactId>org.openhab.transform.rollershutterposition</artifactId>
--- a/bundles/org.openhab.transform.rollershutterposition/src/main/java/org/openhab/transform/rollershutterposition/internal/RollerShutterPositionConstants.java
+++ b/bundles/org.openhab.transform.rollershutterposition/src/main/java/org/openhab/transform/rollershutterposition/internal/RollerShutterPositionConstants.java
@@ -27,10 +27,11 @@ public class RollerShutterPositionConstants {
     // Profile Type UID
     public static final ProfileTypeUID PROFILE_TYPE_UID = new ProfileTypeUID("rollershutter", "position");
 
-    // Paramaters
+    // Parameters
     public static final String UPTIME_PARAM = "uptime";
     public static final String DOWNTIME_PARAM = "downtime";
     public static final String PRECISION_PARAM = "precision";
 
     public static final int POSITION_UPDATE_PERIOD = 800; // fixed delay in milliseconds
+    public static final int DEFAULT_PRECISION = 5;
 }
--- a/bundles/org.openhab.transform.rollershutterposition/src/main/java/org/openhab/transform/rollershutterposition/internal/RollerShutterPositionConstants.java
+++ b/bundles/org.openhab.transform.rollershutterposition/src/main/java/org/openhab/transform/rollershutterposition/internal/RollerShutterPositionConstants.java
@@ -1,0 +1,36 @@
+/**
+ * Copyright (c) 2010-2022 Contributors to the openHAB project
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package org.openhab.transform.rollershutterposition.internal;
+
+import org.eclipse.jdt.annotation.NonNullByDefault;
+import org.openhab.core.thing.profiles.ProfileTypeUID;
+
+/**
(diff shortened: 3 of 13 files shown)


bundles/org.openhab.transform.rollershutterposition/pom.xml
```diff
@@ -0,0 +1,17 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+  xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd">
+
+  <modelVersion>4.0.0</modelVersion>
+
+  <parent>
+    <groupId>org.openhab.addons.bundles</groupId>
+    <artifactId>org.openhab.addons.reactor.bundles</artifactId>
+    <version>3.4.0-SNAPSHOT</version>
```
wborn, line 10:
```suggestion
    <version>4.0.0-SNAPSHOT</version>
```

bundles/org.openhab.transform.rollershutterposition/src/main/java/org/openhab/transform/rollershutterposition/internal/RollerShutterPositionConstants.java
```diff
@@ -0,0 +1,37 @@
...
+ * The {@link RollerShutterPositionConstants } class to define transform constants
+ * used across the whole binding.
+ *
+ * @author Jeff James - Initial contribution
+ */
+@NonNullByDefault
+public class RollerShutterPositionConstants {
+
+    // Profile Type UID
+    public static final ProfileTypeUID PROFILE_TYPE_UID = new ProfileTypeUID("rollershutter", "position");
```
jlaur, line 28: Other transformations seems to use `TransformationService.TRANSFORM_PROFILE_SCOPE` as first parameter (scope) and the transformation name in uppercase as second parameter (id) - in this case it would be ROLLERSHUTTERPOSITION.

jlaur (changes requested):
Thank you. Some minor feedback provided.

CODEOWNERS
```diff
@@ -400,6 +400,7 @@
 /bundles/org.openhab.transform.map/ @openhab/add-ons-maintainers
 /bundles/org.openhab.transform.regex/ @openhab/add-ons-maintainers
 /bundles/org.openhab.transform.scale/ @clinique
+/bundles/org.openhab.transform.rollershutterposition/ @jsjames
```
jlaur, line 403: This should be in alphabetical order.

bundles/org.openhab.transform.rollershutterposition/src/main/java/org/openhab/transform/rollershutterposition/internal/RollerShutterPositionProfileFactory.java
```diff
@@ -0,0 +1,58 @@
...
+import org.openhab.core.thing.profiles.ProfileContext;
+import org.openhab.core.thing.profiles.ProfileFactory;
+import org.openhab.core.thing.profiles.ProfileType;
+import org.openhab.core.thing.profiles.ProfileTypeBuilder;
+import org.openhab.core.thing.profiles.ProfileTypeProvider;
+import org.openhab.core.thing.profiles.ProfileTypeUID;
+import org.osgi.service.component.annotations.Component;
+
+/**
+ * {@link RollerShutterPositionProfileFactory } Factory to create the profile
```
jlaur, line 35:
```suggestion
 * {@link RollerShutterPositionProfileFactory} to create the profile
```

bundles/org.openhab.transform.rollershutterposition/src/main/java/org/openhab/transform/rollershutterposition/internal/config/RollerShutterPositionConfig.java
```diff
@@ -0,0 +1,30 @@
...
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package org.openhab.transform.rollershutterposition.internal.config;
+
+import static org.openhab.transform.rollershutterposition.internal.RollerShutterPositionConstants.DEFAULT_PRECISION;
+
+import org.eclipse.jdt.annotation.NonNullByDefault;
+
+/**
+ * The {@link RollerShutterPositionConfig } class contains the parameters for RollerShutterPosition
```
jlaur, line 20:
```suggestion
 * The {@link RollerShutterPositionConfig} class contains the parameters for RollerShutterPosition
```

bundles/org.openhab.transform.rollershutterposition/src/main/resources/OH-INF/i18n/rollershutter.properties
```diff
@@ -0,0 +1,6 @@
+profile.config.rollershutter.position.downtime.label = Down Time
+profile.config.rollershutter.position.downtime.description = Time it takes for roller shutter to extend the full length (in seconds). Defaults to Up Time if not specified.
+profile.config.rollershutter.position.precision.label = Precision
+profile.config.rollershutter.position.precision.description = Minimum movement (in percent) that can be requested. If the requested change is less than this amount, no action will be taken. This may be required for systems where there is a lag in the stop command and consequently it is not possible for fine control of movement. (default = 1)
```
jlaur, line 4: Please regenerate this file as some of the strings changed (e.g. default is now 5). See https://www.openhab.org/docs/developer/utils/i18n.html#generating-i18n-properties-file

bundles/pom.xml
```diff
@@ -40,6 +40,7 @@
     <module>org.openhab.transform.map</module>
     <module>org.openhab.transform.regex</module>
     <module>org.openhab.transform.scale</module>
+    <module>org.openhab.transform.rollershutterposition</module>
```
jlaur, line 43: This should be in alphabetical order.

jlaur (changes requested):
A few last comments.

jlaur (approved):
Thank you for this transformation add-on. LGTM.

jlaur:
@jsjames - can you create a follow-up PR adding the file `src/main/resources/OH-INF/addon/addon.xml`? I missed this during review, and only now discovered that when also started to work on a transformation.

jlaur:
@jsjames - gentle ping. 🙂

jsjames:
Will definitely get to this this weekend. Just traveling now. On Mar 8, 2023, at 11:57 PM, Jacob Laursen ***@***.***> wrote:﻿
@jsjames - gentle ping. 🙂

jlaur:
> Will definitely get to this this weekend. Just traveling now.

That's okay. I have created #14559 trying to squeeze it into the M1 build. You might want to have a look - I copied name and description from pom/readme.

jsjames:
> Is there a possibility to add functionality for blinds that can open/close (angle of the blades)?

What item type are you using to control the open/close - a Contact or Switch?  I think this would be possible and could look at this with you.

> So depending on the wanted open/close state, the shutter has to move in the opposite direction for a short time at the end.

I don't follow what you mean here?

> Also some shutters like mine have a feedback for full open and fully closed. It would be nice if it could be used to adjust the times automatically, and/or to recalibrate.

I think that would be possible as well.  Let me know if you want to work on this together?

Openhabber:
Wow, thank you!

I'm talking about this type of shutter:
https://www.storen.ch/fileadmin/_processed_/b/1/csm_EC80_mit_Profilfuehrung_gross_HRes_RGB.tif_da7f47f1cf.jpg

I'm using the elero binding, but it should be adaptable to other bindings as well 

The shutter is controlled via 0..100%, and responds also with 0..100. it's kinda awkward as it ignores most commands and also only reports 0, 50, and 100.

But generally speaking, it is controlled via UP, DOWN, and STOP.
Physically, when issuing an UP command, the blades first tilt open and then the shutter goes up.
If you issue a down command, the blades first close and then the shutter goes down.

So if you go down, but want the blades open, you have to move it down to the desired position, STOP, and then go UP again for a second.

The elero binding even has separate commands that do that for you (OPEN and CLOSE).

They are really common here (not generally elero, but the type of blinds that have tilting blades).

I'm unsure how this could be added in a general way to this function (feedback fully open, feedback fully closed, and the tilting command after reaching the target position)

jsjames:
In looking at the elerotransmitterstick binding, I don't see how it supports the OPEN/CLOSE commands, it looks like there is only a rollershutter type channel and a read-only status channel.  Am I looking at the right binding?

If I understand how the shade works, the basic function of the transform should work as-is if you set the uptime/downtime to equal the total time it takes for the blind to tilt open/close + move the shade fully up or down.  The angle of the blades would just be controlled based on the last few percentages (i.e. 98%, 99%, 100%) - i'm not sure what percent of the tilt time is to the total time of tilt time down + close time so this may not give the granularity you want.  But the limit here is how fast openhab + transmitter will send a UP command and then a quick STOP command once the position is reached is likely going to limit the level of accuracy you can achieve.

An enhancement could be made to the transform so that if the value being set is close to the extreme (i.e. 98% closed), then it could recalibrate by sending DOWN command to get to 100% and then send UP command followed by STOP.

Regarding looking at the status to recalibrate - the transform will recalibrate automatically to the full up / down position even without feedback from the handler.  If a 0% or 100% command is issued, the transform will an UP or DOWN command to the handler without a stop.

Openhabber:
You're right. It does not support close and open commands.
I actually do that from rules so I can issue open and close through a button press (up, wait, stop and down, wait, stop).

As for the percentages, you're right that this should work.
Also the recalibration should also be ok like you described it.

In a perfect world, I would imagine the following:
-being able to optionally define a TILT_OPEN_TIME and TILT_CLOSE_TIME (in the 0.5..3s range typically)
-being able to issue a TILT_OPEN and TILT_CLOSE command in addition to up, down, stop. If the shutter is currently moving, those commands have immediate effect and stop the shutter. It also updates the switch described in the following point.
-an optional second control of type switch that controls the wanted state for TILT. If I change it, it has the same effect as TILT_OPEN/TILT_CLOSE but does not stop a moving shutter (it will be issued after moving is done).
If the shutter is done moving through the percentage control or even full up/down commands, it will be adjusted according to TILT if it did move in the other direction.

Im also willing to test if I can get the transform running in OH3.4? I sadly can't upgrade to 4.0 just for testing at the moment.

jsjames:
There is as version of the transform compiled for openhab3 here: https://drive.google.com/file/d/1b-3cpLOFW2T51h_2tWvq1zagcwvSpVVs/view?usp=share_link which you can try.

Since this is a transform, I don't believe it can accept different types of commands vs. what the item type can support - so a Rollershutter item type would only support UP, DOWN, STOP and percent.   If you are thinking to combine multiple items to the handler via the transform, I don't believe that would be supported either.  I wonder if the best way to handle this is with scripts?

Openhabber:
Thanks, I guess then it won't make sense to adjust anything, as you would need to control the tilt independently from the opening. I will have to go via scripts then I guess :) wouldve just been nice to have.

Thanks anyway!

PR #13271, opened 2022-08-16
[velux] Vane position channel is visible when required

### LINKED ISSUE #13260: [velux] Create vane position channel dynamically
### Background

In PR #12618 support was added for a vane position channel on venetian blind type roller shutters. In that PR the extra channel is created statically in the things xml description file. This means that the channel is always present on all types of type roller shutters, even if the roller shutter does not actually support vanes. This is rather ugly and may lead to user confusion.

### Proposed Solution

The vane position should not be created statically in the things xml description file. Instead the channel should be created dynamically, only on roller shutter types that do actually support vanes.

andrewfg:
### Background

In #12618 we added support for a vane position channel on venetian style blinds. And in that PR the respective `vanePosition` channel was created statically for ALL types of shutters and blinds regardless of whether the actual shutter/blind does physically support vanes. This could be confusing for users who have shutters / blinds that do not support vanes.

### Solution

In this PR the binding queries each shutter/blind in the hub, and if that device actually supports vanes, then the binding will create a `vanePosition` channel dynamically.

Resolves #13260

### Acknowledgments

PS many thanks to @pacive, @splatch, @lolodomo and @jlaur for support via the community forum..

https://community.openhab.org/t/dynamically-created-channel-not-showing-up-in-main-ui/138150

### DIFF
--- a/bundles/org.openhab.binding.velux/src/main/java/org/openhab/binding/velux/internal/handler/VeluxBridgeHandler.java
+++ b/bundles/org.openhab.binding.velux/src/main/java/org/openhab/binding/velux/internal/handler/VeluxBridgeHandler.java
@@ -466,7 +466,7 @@ public class VeluxBridgeHandler extends ExtendedBaseBridgeHandler implements Vel
             }
         }
 
-        initializeVanePositionChannels();
+        updateDynamicChannels();
 
         veluxBridgeConfiguration.hasChanged = false;
         logger.debug("Velux veluxBridge is online, now.");
@@ -964,11 +964,11 @@ public class VeluxBridgeHandler extends ExtendedBaseBridgeHandler implements Vel
     /**
      * Ask all things in the hub to initialise their dynamic vane position channel if they support it.
      */
-    private void initializeVanePositionChannels() {
+    private void updateDynamicChannels() {
         getThing().getThings().stream().forEach(thing -> {
             ThingHandler thingHandler = thing.getHandler();
             if (thingHandler instanceof VeluxHandler) {
-                ((VeluxHandler) thingHandler).initializeVanePositionChannel(this);
+                ((VeluxHandler) thingHandler).updateDynamicChannels(this);
             }
         });
     }
--- a/bundles/org.openhab.binding.velux/src/main/java/org/openhab/binding/velux/internal/handler/VeluxBridgeHandler.java
+++ b/bundles/org.openhab.binding.velux/src/main/java/org/openhab/binding/velux/internal/handler/VeluxBridgeHandler.java
@@ -1004,8 +1004,6 @@ public class VeluxBridgeHandler extends ExtendedBaseBridgeHandler implements Vel
         final boolean vaneChannelExisting = channels.stream().anyMatch(vaneChannelFilter);
         final boolean vaneChannelRequired = product.supportsVanePosition();
 
-        logger.warn("channelExisting:{} channelRequired:{}", vaneChannelExisting, vaneChannelRequired);
-
         if (vaneChannelExisting == vaneChannelRequired) {
             // no change needed
             return;
--- a/bundles/org.openhab.binding.velux/src/main/java/org/openhab/binding/velux/internal/handler/VeluxBridgeHandler.java
+++ b/bundles/org.openhab.binding.velux/src/main/java/org/openhab/binding/velux/internal/handler/VeluxBridgeHandler.java
@@ -1002,7 +1002,7 @@ public class VeluxBridgeHandler extends ExtendedBaseBridgeHandler implements Vel
 
         // current and required state of the vane channel
         final boolean vaneChannelExisting = channels.stream().anyMatch(vaneChannelFilter);
-        final boolean vaneChannelRequired = !product.supportsVanePosition();
+        final boolean vaneChannelRequired = product.supportsVanePosition();
 
         logger.warn("channelExisting:{} channelRequired:{}", vaneChannelExisting, vaneChannelRequired);
 
--- a/bundles/org.openhab.binding.velux/src/main/java/org/openhab/binding/velux/internal/handler/VeluxBridgeHandler.java
+++ b/bundles/org.openhab.binding.velux/src/main/java/org/openhab/binding/velux/internal/handler/VeluxBridgeHandler.java
@@ -1021,8 +1021,8 @@ public class VeluxBridgeHandler extends ExtendedBaseBridgeHandler implements Vel
 
         if (vaneChannelRequired) {
             // build the vane channel
-            Channel newChannel = ChannelBuilder.create(vaneChannelUID).withAcceptedItemType(CoreItemFactory.DIMMER)
-                    .withKind(ChannelKind.STATE)
+            Channel newChannel = ChannelBuilder.create(vaneChannelUID).withType(VeluxBindingConstants.CHANNEL_TYPE_VANE)
+                    .withKind(ChannelKind.STATE).withAcceptedItemType(CoreItemFactory.DIMMER)
                     .withDescription(localization.getText("channel-type.velux.vanePosition.description"))
                     .withLabel(localization.getText("channel-type.velux.vanePosition.label")).build();
 
--- a/bundles/org.openhab.binding.velux/src/main/java/org/openhab/binding/velux/internal/handler/VeluxBridgeHandler.java
+++ b/bundles/org.openhab.binding.velux/src/main/java/org/openhab/binding/velux/internal/handler/VeluxBridgeHandler.java
@@ -12,7 +12,9 @@
  */
 package org.openhab.binding.velux.internal.handler;
 
+import java.util.ArrayList;
 import java.util.Collection;
+import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Set;
@@ -21,6 +23,7 @@ import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.TimeUnit;
+import java.util.function.Predicate;
 
 import org.eclipse.jdt.annotation.NonNullByDefault;
 import org.eclipse.jdt.annotation.Nullable;
@@ -61,15 +64,21 @@ import org.openhab.binding.velux.internal.things.VeluxProductPosition.PositionTy
 import org.openhab.binding.velux.internal.utils.Localization;
 import org.openhab.core.common.AbstractUID;
 import org.openhab.core.common.NamedThreadFactory;
+import org.openhab.core.library.CoreItemFactory;
 import org.openhab.core.library.types.DecimalType;
 import org.openhab.core.library.types.OnOffType;
 import org.openhab.core.library.types.PercentType;
 import org.openhab.core.thing.Bridge;
+import org.openhab.core.thing.Channel;
 import org.openhab.core.thing.ChannelUID;
+import org.openhab.core.thing.Thing;
 import org.openhab.core.thing.ThingStatus;
 import org.openhab.core.thing.ThingStatusDetail;
 import org.openhab.core.thing.ThingTypeUID;
+import org.openhab.core.thing.binding.ThingHandler;
 import org.openhab.core.thing.binding.ThingHandlerService;
+import org.openhab.core.thing.binding.builder.ChannelBuilder;
+import org.openhab.core.thing.type.ChannelKind;
 import org.openhab.core.types.Command;
 import org.openhab.core.types.RefreshType;
 import org.openhab.core.types.State;
@@ -464,6 +473,8 @@ public class VeluxBridgeHandler extends ExtendedBaseBridgeHandler implements Vel
                 logger.warn("Activation of House-Status-Monitoring failed (might lead to a lack of status updates).");
             }
         }
+
+        initializeVanePositionChannels();
 
         veluxBridgeConfiguration.hasChanged = false;
         logger.debug("Velux veluxBridge is online, now.");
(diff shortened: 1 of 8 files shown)


bundles/org.openhab.binding.velux/src/main/resources/OH-INF/thing/rollershutter.xml
```diff
@@ -8,7 +8,7 @@
 	<!-- -->
 	<!-- Velux rollershutter Binding -->
 	<!-- -->
-	<thing-type id="rollershutter">
+	<thing-type id="rollershutter" extensible="position, limitMinimum, limitMaximum, vanePosition">
```
splatch, line 11: This statement means that you might have multiple channels of given types. I am not sure if that's what you expect, for sure roller shutter can't have two positions at the same time.

andrewfg: I suspected as much.. 

I think `extensible` is not very well specified -- it should probably have `minInstances` / `maxInstances` attributes for each of the named channel-types. Or something like that..

In this case we need..
- position => minInstances:1, maxInstances:1
- limitMinimum => minInstances:1, maxInstances:1
- limitMaximum => minInstances:1, maxInstances:1
- vanePosition => minInstances:0, maxInstances:1

andrewfg: PS for this case I will implement whatever the OH maintainers instruct me to do..
- if you want an `extensible` attribute, please tell me the exact attribute value
- if you do not want an `extensible` attribute, please say so

:)

pacive: If you want the users to be able to add channels themselves via the UI, you should have it, otherwise no.

andrewfg: Ok. I just deleted it..

gs4711 (approved):

pacive (approved):
Looks good

bundles/org.openhab.binding.velux/src/main/java/org/openhab/binding/velux/internal/handler/VeluxBridgeHandler.java
```diff
@@ -957,4 +968,82 @@ public ProductBridgeIndex getProductBridgeIndex(String thingName) {
...
+
+        ThingHandler handler = thing.getHandler();
+        if (!(handler instanceof VeluxHandler)) {
+            // handler is of the wrong wrong type
+            return;
+        }
+        VeluxHandler veluxHandler = (VeluxHandler) handler;
+
+        // make a mutable copy of the original immutable channel list
+        channels = new ArrayList<>(channels);
```
lolodomo, line 1020: Why not moving most of this code in a new method in the thing handler? 
This would avoid overriding methods visibility.

bundles/org.openhab.binding.velux/src/main/java/org/openhab/binding/velux/internal/handler/VeluxBridgeHandler.java
```diff
@@ -957,4 +968,82 @@ public ProductBridgeIndex getProductBridgeIndex(String thingName) {
...
+        final boolean vaneChannelRequired = product.supportsVanePosition();
+
+        if (vaneChannelExisting == vaneChannelRequired) {
+            // no change needed
+            return;
+        }
+
+        ThingHandler handler = thing.getHandler();
+        if (!(handler instanceof VeluxHandler)) {
+            // handler is of the wrong wrong type
```
lolodomo, line 1014: Wrong is duplicated

lolodomo (changes requested):

bundles/org.openhab.binding.velux/src/main/java/org/openhab/binding/velux/internal/handler/VeluxBridgeHandler.java
andrewfg: Makes sense. Will do

bundles/org.openhab.binding.velux/src/main/java/org/openhab/binding/velux/internal/handler/VeluxBridgeHandler.java
andrewfg: Done.

bundles/org.openhab.binding.velux/src/main/java/org/openhab/binding/velux/internal/handler/VeluxBridgeHandler.java
andrewfg: Done.

andrewfg:
@lolodomo please hold; in the handler calling `updateThing()` causes its `thingUpdated()` to be called, which in turn calls the `dispose()` and `initialize()` methods; so there may be a risk of recursion; in fact I don't see this, but let me just check it again..

andrewfg:
Converted this to draft until the way forward for other similar PRs has been resolved.

bundles/org.openhab.binding.velux/src/main/java/org/openhab/binding/velux/internal/handler/VeluxHandler.java
```diff
@@ -125,4 +129,37 @@ public void handleConfigurationUpdate(Map<String, Object> configurationParameter
...
+            return;
+        }
+
+        String id = VeluxBindingConstants.CHANNEL_VANE_POSITION;
+        ChannelUID uid = new ChannelUID(thing.getUID(), id);
+        Thing2VeluxActuator actuator = new Thing2VeluxActuator(bridgeHandler, uid);
+        VeluxProduct product = bridgeHandler.existingProducts().get(actuator.getProductBridgeIndex());
+
+        if (product.equals(VeluxProduct.UNKNOWN)) {
+            throw new IllegalStateException("initializeVanePositionChannel(): Product unknown in the bridge");
```
lolodomo, line 152: Minor: maybe you would like to change "initializeVanePositionChannel()" into "updateDynamicChannels()" ?

bundles/org.openhab.binding.velux/src/main/java/org/openhab/binding/velux/internal/handler/VeluxHandler.java
```diff
@@ -125,4 +129,37 @@ public void handleConfigurationUpdate(Map<String, Object> configurationParameter
...
+        }
+
+        Channel channel = thing.getChannel(id);
+        boolean required = product.supportsVanePosition();
+
+        if (!required && channel != null) {
+            logger.debug("Removing unsupported channel for {}: {}", thing.getUID(), id);
+            scheduler.submit(() -> updateThing(editThing().withoutChannels(channel).build()));
```
lolodomo, line 160: For my curiosity, why is it run in a separate thread ?

bundles/org.openhab.binding.velux/src/main/resources/OH-INF/thing/channels.xml
```diff
@@ -132,6 +132,13 @@
 
 	<!-- Channel Type - Generic Things -->
 
+	<channel-type id="windowPosition">
+		<item-type>Rollershutter</item-type>
+		<label>@text/channel-type.velux.windowPosition.label</label>
+		<description>@text/channel-type.velux.windowPosition.description</description>
+		<category>Window</category>
+	</channel-type>
```
lolodomo, line 140: You are not setting a min/max state like for the "position" type ?

bundles/org.openhab.binding.velux/src/main/resources/OH-INF/thing/window.xml
```diff
@@ -16,7 +16,7 @@
 		<description>@text/thing-type.velux.window.description</description>
 		<category>Window</category>
 		<channels>
-			<channel id="position" typeId="position"></channel>
+			<channel id="position" typeId="windowPosition"></channel>
```
lolodomo, line 19: This becomes a breaking change for existing things created with UI.

bundles/org.openhab.binding.velux/src/main/java/org/openhab/binding/velux/internal/handler/VeluxHandler.java
```diff
@@ -125,4 +129,37 @@ public void handleConfigurationUpdate(Map<String, Object> configurationParameter
...
+        } else if (required && channel == null) {
+            logger.warn("Thing {} does not have a '{}' channel => please reinitialize it", thing.getUID(), id);
```
lolodomo, line 162: Is it really "reinitialize" or rather "recreate" ?

If I correctly understand, there is no dynamic channel created by the binding. This is more a static channel which is removed if not expected. Is it correct ?

lolodomo (changes requested):

bundles/org.openhab.binding.velux/src/main/java/org/openhab/binding/velux/internal/handler/VeluxHandler.java
andrewfg: Oops. Done.

bundles/org.openhab.binding.velux/src/main/java/org/openhab/binding/velux/internal/handler/VeluxHandler.java
andrewfg: > This is more a static channel which is removed if not expected. Is it correct ?

Correct. So indeed 'recreate' is better.

andrewfg: Done.

bundles/org.openhab.binding.velux/src/main/resources/OH-INF/thing/window.xml
andrewfg: I don't want to make a breaking change! The only reason why I created a new channel type is because it should have a different label and description text. So I will revert adding this new channel type and instead just override the texts for the channel.

bundles/org.openhab.binding.velux/src/main/resources/OH-INF/thing/channels.xml
andrewfg: Actually the position type does NOT have min/max. Whereas the vanePosition type DOES. But really I think that all three of them should have min/max. => WDYT?

bundles/org.openhab.binding.velux/src/main/java/org/openhab/binding/velux/internal/handler/VeluxHandler.java
andrewfg: Umm. When the bridge goes online, the bridge handler iterates over each of its things in a for loop with a call to this method, and I was a bit concerned about things updating themselves while the bridge is still within the process of updating _**itself**_, so I put it on a thread just in case. But maybe I am worrying about nothing?

jlaur: Seems similar to: https://github.com/openhab/openhab-addons/pull/13275#discussion_r947239063

andrewfg: It is :)  ..
..  the only difference is that this time the method is called within the Bridge BridgeParamsUpdated method iteratively over all shades; so there are more bridges and things to possibly get entangled with each other..

lolodomo: @jlaur: is it ok for you ?

bundles/org.openhab.binding.velux/src/main/resources/OH-INF/thing/channels.xml
lolodomo: My feeling is that it would be better.

bundles/org.openhab.binding.velux/src/main/resources/OH-INF/thing/window.xml
lolodomo: I tend to agree with you. That being said, I let you evaluate the balance between a better "matching" and a breaking change.
Of course, we can accept a breaking change, especially if it is documented in the release notes.

bundles/org.openhab.binding.velux/src/main/java/org/openhab/binding/velux/internal/handler/VeluxHandler.java
andrewfg: Ok. I have just tested it without the thread, and I confirm that there is no loop created.. So I will delete the submit() call.

bundles/org.openhab.binding.velux/src/main/resources/OH-INF/thing/channels.xml
andrewfg: Ok. Added.

andrewfg: PS to be consistent I added it now to all position channels

bundles/org.openhab.binding.velux/src/main/resources/OH-INF/thing/window.xml
andrewfg: Ok. I think we should change it. The previous category setting was wrong, but in earlier versions of OH was relatively unimportant. However as you know, the OH v3.x UI places much more emphasis on the correct semantic categorisation of things and channels within a house, so I think we should correct this error.

PS the limitMinimum / limitMaximum channels were also in category 'blinds', so to be consistent I have removed the category, since those channels are equally applicable to 'windows' and 'blinds' things.

andrewfg:
@lolodomo I added an alert [#1414](https://github.com/openhab/openhab-distro/pull/1414) to the release notes concerning the category correction of window position channels.

bundles/org.openhab.binding.velux/src/main/resources/OH-INF/i18n/velux.properties
```diff
@@ -133,6 +133,8 @@ channel-type.velux.scenes.description = Scenes which are configured on the Bridg
 channel-type.velux.check.label = Check of configuration
 channel-type.velux.check.description = Result of the check of current item configuration.
 #
+channel-type.velux.windowPosition.label = Position
```
lolodomo, line 136: Maybe "Window Position" would be better ?

lolodomo: I am going to merge to have your PR included in coming milestone.
You could change that in a next PR if you like.

lolodomo (approved):
LGTM

bundles/org.openhab.binding.velux/src/main/resources/OH-INF/i18n/velux.properties
andrewfg: Done.

PR #13272, opened 2022-08-16
[tado] Fix AC target temperature reading and writing bugs

### LINKED ISSUE #13252: [tado] Air Conditioning: Thing values not updating
I have a Tado Smart AC Control in my house which is properly working.

## Expected Behavior
The states/values in the Things get updated from the subscribed Channels. When I set a trigger on HVAC, the AC Control does it's stuff and I would expect the power state of the Things to change. I would also expect a change after a timer expires and the AC turns off.

## Current Behavior
After setting up the Bridge, Channels and Things, the values are initially updated, but then no longer until a restart. After the restart, the values are fetched once again. Triggers do not affect other Channels, they are not updated.

1. Setup a "Tado Home" Thing
2. Setup a "Tado Air Condition" Thing with "Tado Home" as Bridge and Zone-Id=2
3. Setup a Thing for e.g. temperature directly from the Channel
4. Watch for channel inputs or trigger HVAC state

## Your Environment
* Version used: openHAB 3.3.0
* Hardware: RaspberryPi 4 8GB with PoE+ Hat
* Operating System and version: Ubuntu 22.04 Server for RasPi
* Other bindings installed: MQTT Broker, Netatmo

Is the issue related to this community post?
https://community.openhab.org/t/several-items-from-different-bindings-stop-updating-after-a-while/47863

Here's a screenshot of my Overview page.

image

andrewfg:
### Background

As described in #13252, on certain types of AC units, when they are in certain operating modes, they do not support a target temperature value. Specifically there are two bugs..
1. When reading from such a device in such a mode, the returned target temperature is `null`, and this causes an NPE when trying to update the respective target temperature channel's state.
2. When writing to such a device in such a mode, if a target temperature is set, the tado server rejects the value with an HTTP 422 error.

### Solution

There are two fixes..
1. If reading a `null` target temperature value, the respective channel's state is now set to 'undefined'.
2. If attempting to write a target temperature value, the binding now first checks a) if target temperature is supported, and b) if the new value is within the min/max temperature range, and if not a warning message is logged.

Fixes #13252 


dandjo (approved):
Works as expected.

`2022-08-16 17:06:22.823 [WARN ] [r.AirConditioningZoneSettingsBuilder] - Your a/c unit does not support 'Target Temperature:23.0' when in state 'AcMode:FAN', (supported values: [none]).`

lolodomo (approved):
LGTM

PR #13263, opened 2022-08-14
[automower] Adapt login due to Husqvarna API change

crnjan:
The Automower Connect API Authentication does not work anymore as they moved to a new authentication method (AppKey and AppSecret) - adopted. 

Fix #12980

Signed-off-by: Boris Krivonog [boris.krivonog@inova.si](mailto:boris.krivonog@inova.si)

PR #13274, opened 2022-08-16
[automower] Refresh token if it's expired

crnjan:
Automower binding stops updating after ~24h because access token expires after 86399s and token was not checked for expiration.


jlaur:
Regression of #13263

crnjan:
Yes, seems as before token refresh was not required ...

jlaur (approved):
Thanks!

PR #13275, opened 2022-08-16
[hdpowerview] Channels are visible depending on shade capabilities

### LINKED ISSUE #13273: [hdpowerview] Create channels dynamically depending on shade capabilities
### Background

As suggested in this [community forum thread](https://community.openhab.org/t/dynamically-created-channel-not-showing-up-in-main-ui/138150/3?u=andrewfg) we should perhaps create shade channels dynamically depending of the shade capabilities.

### Proposed Solution

The `secondary` and `vane` channels should be created dynamically depending on whether the shade reports support for a secondary vertical sliding panel (e.g. top down bottom up) or a shade with vane tilt capability.

### Open Issue

Input requested from @jlaur on the following..

In addition to the `secondary` and `vane` channels, there may be some other channels which may or may not be supported on different types of shades (e.g. `lowBattery` `batteryLevel` `batteryVoltage` on mains powered shades). So before finalising this issue and creating a respective PR we need to decide if such channels shall be included.

andrewfg:
### Background

Resolves #13273

As described in #13273 we should create shade channels dynamically depending of the shade capabilities.

### Solution

The secondary and vane channels are created dynamically depending on whether the shade reports support for a secondary vertical sliding panel (e.g. top down bottom up) or a shade with vane tilt capability.


andrewfg:
@jlaur this is a first commit based on my experience with the Velux binding; it needs rigorous testing from my side; and I would also recommend rigorous testing by you too :)

bundles/org.openhab.binding.hdpowerview/README.md
```diff
@@ -92,8 +92,8 @@ All of these channels appear in the binding, but only those which have a physica
...
-| secondary      | Rollershutter            | The vertical position of the secondary rail (if any). Its function is similar to the `position` channel above -- but see [next chapter](#Roller-Shutter-Up/Down-Position-vs.-Open/Close-State). |
-| vane           | Dimmer                   | The degree of opening of the slats or vanes. Setting this to a non-zero value will first move the shade `position` fully down, since the slats or vanes can only have a defined state if the shade is in its down position -- see [Interdependency between Channel positions](#Interdependency-between-Channel-positions). |
+| secondary      | Rollershutter            | The vertical position of the secondary rail (if any). Its function is similar to the `position` channel above -- but see [next chapter](#Roller-Shutter-Up/Down-Position-vs.-Open/Close-State). (optional) |
```
jlaur, line 95: I would revert this change since this is a channel, not a parameter, so it's hard to follow what "optional channel" actually means. The description already mentions "if any".

bundles/org.openhab.binding.hdpowerview/README.md
```diff
@@ -92,8 +92,8 @@ All of these channels appear in the binding, but only those which have a physica
...
+| vane           | Dimmer                   | The degree of opening of the slats or vanes. Setting this to a non-zero value will first move the shade `position` fully down, since the slats or vanes can only have a defined state if the shade is in its down position -- see [Interdependency between Channel positions](#Interdependency-between-Channel-positions). (optional) |
```
jlaur, line 96: Same here.

bundles/org.openhab.binding.hdpowerview/README.md
```diff
@@ -104,6 +104,10 @@ All of these channels appear in the binding, but only those which have a physica
 
 Please note that RSSI values will only be updated upon manual request by a `REFRESH` command (e.g. in a rule).
 
+The channels `secondary` and `vane` only provide useful data if the shade physically supports such channels.
+The binding queries the shade capabilities if they would support such channels, and if necessary it creates those channels automatically.
+The remainin channels in the above table always provide useful data, so they are always created.
```
jlaur, line 109: I would either remove this or simplify to just mention that the channels only exist for shade things when supported. However, it's kind of already mentioned in the channel descriptions by "(if any)" so any description here is quite verbose.

Mentioning "usefulness of data" seems too complicated/misleading - either you can read and write values (for example get position and set position) or you can't (if not supported).

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/database/ShadeCapabilitiesDatabase.java
```diff
@@ -289,6 +289,24 @@ public boolean supportsTilt180() {
...
+        }
+
+        /**
+         * Check if the Capabilities class instance uses a secondary channel.
+         *
+         * @return true if the shade supports a secondary channel.
+         */
+        public boolean usesSecondaryChannel() {
+            return supportsSecondary || supportsSecondaryOverlapped;
+        }
```
jlaur, line 309: IMHO the channel logic belongs in the handler, not in the database class. The database class should simply expose attributes/capabilities. Alternatively the methods could be renamed to not mention "Channel", so it would be more abstract and only about having vane/secondary and not about whether or not to provide a channel for that.

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/handler/HDPowerViewShadeHandler.java
```diff
@@ -131,6 +142,7 @@ public void dispose() {
         }
         refreshBatteryLevelFuture = null;
         capabilities = null;
+        dynamicChannelsInitialized = false;
```
jlaur, line 145: Seems unneeded.

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/handler/HDPowerViewShadeHandler.java
```diff
@@ -250,6 +262,8 @@ private void handleShadeCommand(String channelId, Command command, HDPowerViewWe
     protected void onReceiveUpdate(ShadeData shadeData) {
         updateStatus(ThingStatus.ONLINE);
         updateCapabilities(shadeData);
+        // note: updateDynamicChannels() requires updateCapabilities() to have been called first
```
jlaur, line 265: So maybe call it from that method?

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/handler/HDPowerViewShadeHandler.java
```diff
@@ -265,12 +279,13 @@ private void updateCapabilities(ShadeData shade) {
...
             return;
         }
-        Capabilities capabilities = db.getCapabilities(shade.type, shade.capabilities);
+        Capabilities capabilities = DB.getCapabilities(shade.type, shade.capabilities);
         if (capabilities.getValue() < 0) {
             logger.debug("Unable to set capabilities for shade {}", shade.id);
             return;
         }
         logger.debug("Caching capabilities {} for shade {}", capabilities.getValue(), shade.id);
+        dynamicChannelsInitialized = false;
```
jlaur, line 288: Continuing from previous comment: Call `updateDynamicChannels()` here?

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/handler/HDPowerViewShadeHandler.java
```diff
@@ -596,4 +611,100 @@ private void doRefreshShade(RefreshKind kind) {
...
+            dynamicChannelsInitialized = dynamicChannelsInitializer();
+        }
+    }
+
+    /**
+     * Helper to do the actual work of initializing the dynamic channels if the respective device supports them.
+     *
+     * @return true if all channels were successfully initialised.
+     */
+    private boolean dynamicChannelsInitializer() {
```
jlaur, line 629: Would it be possible to create a builder class for some of this logic and keep this handler a bit more clean (or at least smaller). See other builders in subdirectory "builders". As an example, see method `updateSceneChannels()` how the footprint was reduced in `HDPowerViewHubHandler` with most logic put into the builder class (which is then also covered by tests).

bundles/org.openhab.binding.hdpowerview/src/main/resources/OH-INF/thing/shade.xml
```diff
@@ -13,11 +13,14 @@
...
-			</channel>
-			<channel id="vane" typeId="shade-vane"/>
+
+			<!-- these channels now created dynamically
+
+				<channel id="secondary" typeId="shade-secondary"/>
+				<channel id="vane" typeId="shade-vane"/>
+
+			-->
+
```
jlaur, line 23: Can this be removed?

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/handler/HDPowerViewShadeHandler.java
```diff
@@ -83,17 +91,20 @@ private enum RefreshKind {
...
+    private static final ShadeCapabilitiesDatabase DB = new ShadeCapabilitiesDatabase();
 
     private @Nullable ScheduledFuture<?> refreshPositionFuture = null;
     private @Nullable ScheduledFuture<?> refreshSignalFuture = null;
     private @Nullable ScheduledFuture<?> refreshBatteryLevelFuture = null;
     private @Nullable Capabilities capabilities;
     private int shadeId;
     private boolean isDisposing;
+    private boolean dynamicChannelsInitialized;
+    private final HDPowerViewTranslationProvider translater;
```
jlaur, line 103: For clarity:
```suggestion
    private final HDPowerViewTranslationProvider translationProvider;
```

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/handler/HDPowerViewShadeHandler.java
```diff
@@ -596,4 +611,100 @@ private void doRefreshShade(RefreshKind kind) {
...
+            // build the vane channel
+            Channel vaneChannel = ChannelBuilder
+                    .create(new ChannelUID(thing.getUID(), HDPowerViewBindingConstants.CHANNEL_SHADE_VANE))
+                    .withType(HDPowerViewBindingConstants.CHANNEL_TYPE_VANE).withKind(ChannelKind.STATE)
+                    .withAcceptedItemType(CoreItemFactory.DIMMER)
+                    .withDescription(translater.getText("channel-type.hdpowerview.shade-vane.description"))
+                    .withLabel(translater.getText("channel-type.hdpowerview.shade-vane.label")).build();
+
+            // add the vane channel
+            error |= !channels.add(vaneChannel);
```
jlaur, line 674: If `error` is `true` after this, why continue (everything will be dropped in line 699 anyway)?

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/handler/HDPowerViewShadeHandler.java
```diff
@@ -596,4 +611,100 @@ private void doRefreshShade(RefreshKind kind) {
...
+        // note: this is an immutable list
+        List<Channel> channels = thing.getChannels();
+        final Stream<Channel> channelStream = channels.stream();
+
+        // predicate to filter the vane channel
+        final Predicate<Channel> vaneChannelFilter = c -> HDPowerViewBindingConstants.CHANNEL_TYPE_VANE
+                .equals(c.getChannelTypeUID());
+
+        // predicate to filter the secondary channel
+        final Predicate<Channel> secChannelFilter = c -> HDPowerViewBindingConstants.CHANNEL_TYPE_SECONDARY
```
jlaur, line 645: For clarity:
```suggestion
        final Predicate<Channel> secondaryChannelFilter = c -> HDPowerViewBindingConstants.CHANNEL_TYPE_SECONDARY
```

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/handler/HDPowerViewShadeHandler.java
```diff
@@ -596,4 +611,100 @@ private void doRefreshShade(RefreshKind kind) {
...
+            // add the secondary channel
+            error |= !channels.add(secChannel);
+        } else {
+            // remove the secondary channel
+            error |= !channels.removeIf(secChannelFilter);
+        }
+
+        if (error) {
+            // something went wrong
+            return false;
```
jlaur, line 699: What will happen after this? The actual cause is interesting, because if this can actually happen, we could potentially get stuck in a loop trying again and again to construct these channels each time the shade is updated? I would assume it would fail consistently unless we have a concurrency issue.

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/handler/HDPowerViewShadeHandler.java
```diff
@@ -596,4 +611,100 @@ private void doRefreshShade(RefreshKind kind) {
...
+        }
+
+        /*
+         * If we got this far then we will update the thing by building an identical copy of the target thing, except
+         * that it uses the newly modified channel list instead of the original one.
+         */
+        final Thing thing = editThing().withChannels(channels).build();
+        scheduler.submit(() -> updateThing(thing));
```
jlaur, line 707: Why do you need the scheduler for this?

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/handler/HDPowerViewShadeHandler.java
```diff
@@ -596,4 +611,100 @@ private void doRefreshShade(RefreshKind kind) {
...
+            // remove the vane channel
+            error |= !channels.removeIf(vaneChannelFilter);
+        }
+
+        if (secChannelRequired) {
+            // build the secondary channel
+            Channel secChannel = ChannelBuilder
+                    .create(new ChannelUID(thing.getUID(),
+                            HDPowerViewBindingConstants.CHANNEL_SHADE_SECONDARY_POSITION))
+                    .withType(HDPowerViewBindingConstants.CHANNEL_TYPE_SECONDARY).withKind(ChannelKind.STATE)
```
jlaur, line 685: Why not use channel type `shade-position` like the static channel did? We now have redundant channel types only different in label and description, and this is set explicitly anyway.

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/handler/HDPowerViewShadeHandler.java
```diff
@@ -596,4 +611,100 @@ private void doRefreshShade(RefreshKind kind) {
...
+                    .withAcceptedItemType(CoreItemFactory.DIMMER)
+                    .withDescription(translater.getText("channel-type.hdpowerview.shade-secondary.description"))
+                    .withLabel(translater.getText("channel-type.hdpowerview.shade-secondary.label")).build();
+
+            // add the secondary channel
+            error |= !channels.add(secChannel);
```
jlaur, line 691: I haven't tested this, but I seem to remember that channels are presented in the order of being added. I'm afraid this means that `secondary` will now be at the bottom of the list, not right after `position`?

jlaur (changes requested):
Thanks! First review round completed. I have not performed any tests yet, but would like to do that at some point before merging.

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/database/ShadeCapabilitiesDatabase.java
andrewfg: I originally had the code in the calling class, but I moved it to the database to make it bit faster. However I will move it back again..

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/handler/HDPowerViewShadeHandler.java
andrewfg: Probably is..

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/handler/HDPowerViewShadeHandler.java
andrewfg: Ok.

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/handler/HDPowerViewShadeHandler.java
andrewfg: Ok.

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/handler/HDPowerViewShadeHandler.java
andrewfg: Yeah.

bundles/org.openhab.binding.hdpowerview/src/main/resources/OH-INF/thing/shade.xml
andrewfg: Yeah.

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/handler/HDPowerViewShadeHandler.java
andrewfg: Ok

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/handler/HDPowerViewShadeHandler.java
andrewfg: Ok

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/handler/HDPowerViewShadeHandler.java
andrewfg: I think a) it will never happen, b) if it did happen then it would do so consistently, but c) we would not be be stuck in a loop, since updateCapabilities() will have only called us one time...   ...

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/handler/HDPowerViewShadeHandler.java
andrewfg: I just had a feeling that a class updating parts of itself should not be done directly within itself. But I think `poll()` is on a scheduled task so probably we can eliminate the scheduler here..

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/handler/HDPowerViewShadeHandler.java
andrewfg: Makes sense.

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/handler/HDPowerViewShadeHandler.java
andrewfg: Makes sense.

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/handler/HDPowerViewShadeHandler.java
andrewfg: Unfortunately it does not move 'most of the logic' to the builder. But it is nevertheless done.

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/handler/HDPowerViewShadeHandler.java
andrewfg: You may be right. The 'position' channel is statically created, so I replaced `add(channel)`  by `add(1, channel)`..

bundles/org.openhab.binding.hdpowerview/README.md
andrewfg: I will do this once the functionality has been frozen.

bundles/org.openhab.binding.hdpowerview/README.md
andrewfg: Ditto

bundles/org.openhab.binding.hdpowerview/README.md
andrewfg: Ditto

andrewfg:
@jlaur I have made the following changes..
1. In general adopted your preliminary review suggestions. Except the read me which I will do last of all.
2. Moved the channel creation code to a separate `ShadeChannelBuilder` class.
3. Created a JUnit test class `ShadeChannelBuilderTest`.
4. Learned from the tests that one capabilities variant has no primary, so the `position` channel needs also to be dynamic.
5. Learned from the tests that some potential error cases cannot actually occur, so deleted the respective error handling code.

bundles/org.openhab.binding.hdpowerview/README.md
andrewfg: Done

andrewfg:
@jlaur please hold; in the handler calling 'updateThing()` causes its 'thingUpdated()` to be called, which in turn calls the `dispose()` and `initialize()` methods; so there may be a risk of recursion; in fact I don't see this, but let me just check it again..

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/handler/HDPowerViewShadeHandler.java
jlaur: Reopening this since unresolved. Proposition was to have only one channel type for shade position which can be used for both channels - like previously. They only differ in name, so there is no reason for duplication.

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/handler/HDPowerViewShadeHandler.java
jlaur: Reopening as I still don't see any reason for this.

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/builders/ShadeChannelBuilder.java
```diff
@@ -0,0 +1,204 @@
...
+import org.openhab.core.thing.type.ChannelKind;
+import org.openhab.core.thing.type.ChannelTypeUID;
+
+/**
+ * The {@link ShadeChannelBuilder} class creates vane and secondary position channels.
+ *
+ * @author Andrew Fiddian-Green - Initial contribution
+ */
+@NonNullByDefault
+public class ShadeChannelBuilder {
```
jlaur, line 34: Maybe I mislead you, but when suggesting to provide a channel builder, intention was to provide another builder following the same pattern as the existing builders. Existing builders:
- Each builder builds only one channel.
- Each builder is a subclass of `BaseChannelBuilder`.
- Each builder contains a static method `create` returning a builder instance.
- Builders are used by things but are completely decoupled from things (i.e. contain no references to handlers).
- With all of the above, these builders are quite focused and simple (`SceneChannelBuilder` being the simplest and `AutomationChannelBuilder` containing most logic).

I'll let you consider this before starting a review of the current implementation.

jlaur (changes requested):
A few more comments. Full review and test still pending, but in meantime would be interesting to see your test results (including positions of the now dynamic channels in Main UI).

bundles/org.openhab.binding.hdpowerview/src/test/java/org/openhab/binding/hdpowerview/ShadeChannelBuilderTest.java
```diff
@@ -0,0 +1,415 @@
...
+     * Helper method to run tests on the given thing instance, with the given shade capabilities id, and with the given
+     * expected result
+     *
+     * @param thing test article (may be a full or an empty thing)
+     * @param capbilitiesId target capabilities
+     * @param expectedDirty expected dirty outcome of tests
+     * @param expectedChannelCount expected resulting number of channels
+     * @return the resulting channel list
+     */
+    private List<Channel> testListAddRemoveChannelsHelper(Thing thing, int capbilitiesId, Expect expectedDirty,
```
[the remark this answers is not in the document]
andrewfg: Ok. Done.

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/handler/HDPowerViewShadeHandler.java
andrewfg: Calling the framework's updateThing() method causes the framework to call the handler's thingUpdated() method, which in turn calls the handler's dispose() and initialize methods in turn. So in other words the handler instance is launching a series of actions that can have a significant impact on its own state. Therefore I feel more comfortable to definitively decouple any possible risk of the handler recursing method calls onto itself, by starting the avalanche separately from the task scheduler rather than from itself.

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/handler/HDPowerViewShadeHandler.java
andrewfg: I want to follow a consistent approach towards the creation of the three dynamic channels (position, vane and secondary). And the problem with your approach is due to the internationalisation, where one would be creating the position and vane channels using the internationalisation properties key for the channel-type and the secondary channel using the internationalisation properties key for the thing-channel. This is messy. So I want to create all three things, and internationalize them using the exact same algorithm.

PS your statement _"They only differ in name"_ is in fact wrong -- in fact they differ in name, label and description.

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/builders/ShadeChannelBuilder.java
andrewfg: Your idea was basically excellent because with a builder one can use the **_exact same algorithm_** for creating and adding resp. removing each of the three possible dynamic things.

I looked at your BaseChannelBuilder, HOWEVER I could not use it because your builder is designed to create MULTIPLE instances of the SAME channel type, hosted in a BRIDGE, and WITH the channels in a channel group. Whereas in this case we need to create SINGLE instances of MULTIPLE channel types, hosted in THING, and WITHOUT a channel group.

In short it was easier to create a dedicated new builder class than to 'munge' (aka re-purpose) your class.

To respond to your comments ..
- Each builder builds only one channel => yes
- Each builder is a subclass of BaseChannelBuilder => no (see above)
- Each builder contains a static method create returning a builder instance => Huh? I honestly don't see any difference between calling 'new construstor()' versus calling static create() which calls 'new construstor()' ..
=> Builders are used by things but are completely decoupled from things (i.e. contain no references to handlers) => yes
- .. builders are quite focused and simple => yes (and my class is not extended from BaseChannelBuilder..)

andrewfg:
>  would be interesting to see your test results (including positions of the now dynamic channels in Main UI).

The updateDynamicChannels() method creates a list of channel builders; then it adds channel builder instances to the list in a specific order (vane, secondary, position); then it iterates over the list of channel builders, and if the respective builder produces a valid channel, then that channel is added to the thing's channel list via add(0, channel). So since position is (if required) created last in the channel builder list, it will appear first in the thing's channel list.

See screenshot below. This is for one of my shades where position and vane are required and therefore dynamically added, whereas secondary is not required and therefore not added..

image

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/handler/HDPowerViewShadeHandler.java
jlaur: Can you point me to the code which calls `dispose()`/`initialize()`? It doesn't sound right that the handler should need to be completely reinitialized because the model is updated. That would have high risk of becoming an infinite loop when `initialize()` would then again cause the model to be updated.

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/handler/HDPowerViewShadeHandler.java
jlaur: You are of course right, label and description - I'm considering them a pair. :-) There is no such thing as name, but of course they also have their own id. My argument was that from a **design**/model perspective it's in fact messy to duplicate channel types. It seems your reasoning for doing so was **implementation** issues. Knowing this, I'll have another look and see if I can come up with something that would be clean and still avoid the redundancy.

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/handler/HDPowerViewShadeHandler.java
andrewfg: Look into the BaseThingHandler code.

Snag_195de5

 In updateThing() at line 456 it calls 'callback.thingUpdated()' and if you trace through the framework code, that eventually ends up calling 'BaseThingHandler.thingUpdate()' at line 152 ..

Snag_1bb32f

andrewfg: > That would have high risk of becoming an infinite loop when initialize() would then again cause the model to be updated.

Exactly.  !!!

andrewfg:
@jlaur I am using basically the same approach to create channels dynamically in the tado binding, so you may want to look at that code in parallel to this one => #13301 

{note: in the tado case the channel builder differs slightly from the HDPowerview case, because the channel names and the channel-type names are identical in the tado binding, and that means one can eliminate a couple of intermediary steps}.

jlaur:
@andrewfg - I'm wondering if you considered and/or tried this approach suggested in the [forum thread](https://community.openhab.org/t/dynamically-created-channel-not-showing-up-in-main-ui/138150/8?u=laursen)? It might simplify the implementation a lot and could even be faster runtime also.

andrewfg:
> I'm wondering if you considered and/or tried this approach suggested in the [forum thread](https://community.openhab.org/t/dynamically-created-channel-not-showing-up-in-main-ui/138150/8?u=laursen)?

Of course I have considered that!!! ..

.. it was after all myself who created that thread; and based on the feedback I created my original PR; and based on the reviews the PR got to the current state.

> could even be faster runtime also

IMHO this is completely irrelevant. The updateDynamicChannels() method is created exactly once when the thing handler is created and the shade capabilities is first loaded. After that the shade capabilities is cached so updateDynamicChannels() is never called again.

PS this caching mechanism probably explains why updateThing() -> thingUpdated() -> dispose() -> initialize() does not cause the infinite loop that you feared.

> might simplify the implementation 

I am happy!!

andrewfg:
@jlaur honestly I don't understand why you are so worried about the complexity etc. I have done this in the last few days with following three bindings where dynamic channels are required..
- Velux requires => 1 dynamic channel => one instance of the channel building is hard coded directly in the handler
- HD Powerview => 3 dynamic channels => channel building done by the handler using three instances of a standard channel builder => simpler than writing the same hard code in the handler three times..
- Tado => 12 dynamic channels => channel building done by the handler using twelve instances of a standard channel builder => simpler than writing the same hard code in the handler twelve times..

@jlaur may I therefore kindly suggest that you look at #13301 before you finalise your review of this one?

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/handler/HDPowerViewShadeHandler.java
jlaur: `updateThing` (which you've shown above) calls this method in ThingHandlerCallback:

https://github.com/openhab/openhab-core/blob/8f745b0d3d5d263cffcd50b2a98989486af5415f/bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/internal/ThingManagerImpl.java#L261-L289

This class implements `ThingManager`.

This is not the same method as in `ThingHandler`. So our `dispose()`/`initialize()` methods are not called.

jlaur:
@andrewfg - I'm sorry to have offended you. I made the mistake of sneak-peaking and pre-reviewing the PR before actually having the time and focus to conduct a proper full review, as you should expect. That unfortunately lead to some standalone comments which shouldn't have been made before asking some important overall questions. When I finally did that today, I might have already mislead you by ideas that may not even be relevant. I'm truly sorry about this.

However, one question remains:

>> I'm wondering if you considered and/or tried this approach suggested in the [forum thread](https://community.openhab.org/t/dynamically-created-channel-not-showing-up-in-main-ui/138150/8?u=laursen)?

> Of course I have considered that!!! ..

What was the result of this consideration?

I have just tried it, and I don't really see any significant downsides with this approach? 

In last line of `updateCapabilities`, add:
```java
        removeUnsupportedChannels(capabilities);
```

Then add this method:
```java
    private void removeUnsupportedChannels(Capabilities capabilities) {
        List<Channel> removeChannels = new ArrayList<Channel>(3);
        Channel channel;
        if (!capabilities.supportsPrimary()) {
            channel = thing.getChannel(CHANNEL_SHADE_POSITION);
            if (channel != null) {
                removeChannels.add(channel);
            }
        }
        if (!capabilities.supportsSecondary() && !capabilities.supportsSecondaryOverlapped()) {
            channel = thing.getChannel(CHANNEL_SHADE_SECONDARY_POSITION);
            if (channel != null) {
                removeChannels.add(channel);
            }
        }
        if (!capabilities.supportsTiltAnywhere() && !capabilities.supportsTiltOnClosed()) {
            channel = thing.getChannel(CHANNEL_SHADE_VANE);
            if (channel != null) {
                removeChannels.add(channel);
            }
        }
        if (!removeChannels.isEmpty()) {
            if (logger.isDebugEnabled()) {
                StringJoiner joiner = new StringJoiner(", ");
                removeChannels.forEach(c -> joiner.add(c.getUID().getId()));
                logger.debug("Removing unsupported channels for {}: {}", shadeId, joiner.toString());
            }
            updateThing(editThing().withoutChannels(removeChannels).build());
        }
    }
```

I tested it, and it works as expected:

```
2022-08-21 21:39:14.789 [DEBUG] [ternal.handler.HDPowerViewHubHandler] - Received data for 11 shades
2022-08-21 21:39:14.792 [DEBUG] [ternal.handler.HDPowerViewHubHandler] - Updating shade '24927'
2022-08-21 21:39:14.794 [DEBUG] [rnal.handler.HDPowerViewShadeHandler] - Caching capabilities 0 for shade 24927
2022-08-21 21:39:14.797 [DEBUG] [rnal.handler.HDPowerViewShadeHandler] - Removing unsupported channels for 24927: secondary, vane
2022-08-21 21:39:14.812 [DEBUG] [ternal.handler.HDPowerViewHubHandler] - Updating shade '5868'
2022-08-21 21:39:14.815 [DEBUG] [rnal.handler.HDPowerViewShadeHandler] - Caching capabilities 0 for shade 5868
2022-08-21 21:39:14.817 [DEBUG] [rnal.handler.HDPowerViewShadeHandler] - Removing unsupported channels for 5868: secondary, vane
2022-08-21 21:39:14.833 [DEBUG] [ternal.handler.HDPowerViewHubHandler] - Updating shade '34896'
2022-08-21 21:39:14.835 [DEBUG] [rnal.handler.HDPowerViewShadeHandler] - Caching capabilities 1 for shade 34896
2022-08-21 21:39:14.837 [DEBUG] [rnal.handler.HDPowerViewShadeHandler] - Removing unsupported channels for 34896: secondary
2022-08-21 21:39:14.854 [DEBUG] [ternal.handler.HDPowerViewHubHandler] - Updating shade '42949'
2022-08-21 21:39:14.856 [DEBUG] [rnal.handler.HDPowerViewShadeHandler] - Caching capabilities 1 for shade 42949
2022-08-21 21:39:14.858 [DEBUG] [rnal.handler.HDPowerViewShadeHandler] - Removing unsupported channels for 42949: secondary
2022-08-21 21:39:14.874 [DEBUG] [ternal.handler.HDPowerViewHubHandler] - Updating shade '23137'
2022-08-21 21:39:14.877 [DEBUG] [rnal.handler.HDPowerViewShadeHandler] - Caching capabilities 0 for shade 23137
2022-08-21 21:39:14.880 [DEBUG] [rnal.handler.HDPowerViewShadeHandler] - Removing unsupported channels for 23137: secondary, vane
2022-08-21 21:39:14.895 [DEBUG] [ternal.handler.HDPowerViewHubHandler] - Updating shade '32544'
2022-08-21 21:39:14.898 [DEBUG] [rnal.handler.HDPowerViewShadeHandler] - Caching capabilities 1 for shade 32544
2022-08-21 21:39:14.900 [DEBUG] [rnal.handler.HDPowerViewShadeHandler] - Removing unsupported channels for 32544: secondary
2022-08-21 21:39:14.916 [DEBUG] [ternal.handler.HDPowerViewHubHandler] - Updating shade '38196'
2022-08-21 21:39:14.919 [DEBUG] [rnal.handler.HDPowerViewShadeHandler] - Caching capabilities 1 for shade 38196
2022-08-21 21:39:14.921 [DEBUG] [rnal.handler.HDPowerViewShadeHandler] - Removing unsupported channels for 38196: secondary
2022-08-21 21:39:14.937 [DEBUG] [ternal.handler.HDPowerViewHubHandler] - Updating shade '36137'
2022-08-21 21:39:14.940 [DEBUG] [rnal.handler.HDPowerViewShadeHandler] - Caching capabilities 7 for shade 36137
2022-08-21 21:39:14.942 [DEBUG] [rnal.handler.HDPowerViewShadeHandler] - Removing unsupported channels for 36137: vane
2022-08-21 21:39:14.959 [DEBUG] [ternal.handler.HDPowerViewHubHandler] - Updating shade '46532'
2022-08-21 21:39:14.962 [DEBUG] [rnal.handler.HDPowerViewShadeHandler] - Caching capabilities 1 for shade 46532
2022-08-21 21:39:14.964 [DEBUG] [rnal.handler.HDPowerViewShadeHandler] - Removing unsupported channels for 46532: secondary
2022-08-21 21:39:14.980 [DEBUG] [ternal.handler.HDPowerViewHubHandler] - Updating shade '24076'
2022-08-21 21:39:14.982 [DEBUG] [rnal.handler.HDPowerViewShadeHandler] - Caching capabilities 0 for shade 24076
2022-08-21 21:39:14.985 [DEBUG] [rnal.handler.HDPowerViewShadeHandler] - Removing unsupported channels for 24076: secondary, vane
2022-08-21 21:39:15.001 [DEBUG] [ternal.handler.HDPowerViewHubHandler] - Updating shade '58120'
2022-08-21 21:39:15.003 [DEBUG] [rnal.handler.HDPowerViewShadeHandler] - Caching capabilities 1 for shade 58120
2022-08-21 21:39:15.005 [DEBUG] [rnal.handler.HDPowerViewShadeHandler] - Removing unsupported channels for 58120: secondary
```

> honestly I don't understand why you are so worried about the complexity etc.

When I see 737 lines added and 28 removed in order to get rid of three channels, I *have* to ask the question: Are there any other ways/is it really needed?

So let me, respectfully, try to evaluate the two approaches (not comparing any code, just approach of removing channels declared in XML vs. dynamically creating the channels (or omitting to do so).

Removal - pros:
- Implementation is very simple.
- Runtime robustness, missing capabilities or anything like that will result in the channels being preserved.
- Channels are easily maintainable in XML and I18N is automatically supported.
- Channel order from XML is respected without any custom logic.

Removal - cons:
- Removal is persisted, so if capabilities database would have a change for an existing managed thing, the thing might need to be recreated.

Dynamic creation - pros:
- Channels can be recreated on binding restart in case of changes in capabilities database.

Dynamic creation - cons:
- Many code lines to write, review and maintain. This also possesses higher risk of bugs.
- Reinventing the wheel: Channels are not truly dynamic, only the decision whether to include them or not is taken runtime. The definitions are static, and yet I18N etc. is implemented.
- Channel order is implemented in logic, thus fragmented because XML and code is mixed. One needs to be aware of both when adding new channels.

This is what I could come up with with my current knowledge, so please let me know if I missed anything regarding the desired functionality to be implemented, and also add your arguments so we can decide which approach would be most suitable to solve the problem at hand.

andrewfg:
@jlaur I was thinking more about (what I think) you are saying. Are you suggesting that instead of having an empty thing and adding channels dynamically, that you prefer instead to have one with all channels created statically, and then delete them dynamically? If so, then we would not need any channel builders at all. And my PRs could/should be essentially totally rewritten. Or?

bundles/org.openhab.binding.hdpowerview/README.md
```diff
@@ -91,9 +91,9 @@ All of these channels appear in the binding, but only those which have a physica
...
+| position       | Rollershutter            | The vertical position of the shade's rail (if any). -- See [next chapter](#Roller-Shutter-Up/Down-Position-vs.-Open/Close-State). Up/Down commands will move the rail completely up or completely down. Percentage commands will move the rail to an intermediate position. Stop commands will halt any current movement of the rail. |
+| secondary      | Rollershutter            | The vertical position of the secondary rail (if any). Its function is similar to the `position` channel above. -- But see [next chapter](#Roller-Shutter-Up/Down-Position-vs.-Open/Close-State). |
+| vane           | Dimmer                   | The degree of opening of the slats or vanes (if any). On some shade types, setting this to a non-zero value might first move the shade `position` fully down, since the slats or vanes can only have a defined state if the shade is in its down position. -- See [Interdependency between Channel positions](#Interdependency-between-Channel-positions). |
```
jlaur, line 96:
```suggestion
| vane           | Dimmer                   | The degree of opening of the slats or vanes (if any). On some shade types, setting this to a non-zero value might first move the shade `position` fully down, since the slats or vanes can only have a defined state if the shade is in its down position. See [Interdependency between Channel positions](#Interdependency-between-Channel-positions). |
```

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/handler/HDPowerViewShadeHandler.java
```diff
@@ -621,7 +621,7 @@ private void removeListProcessChannel(List<Channel> removeList, String channelId
      * Remove previously statically created channels if the shade does not support them.
      */
     private void updateDynamicChannels(Capabilities capabilities) {
-        List<Channel> removeList = new ArrayList<Channel>(3);
+        List<Channel> removeList = new ArrayList<>();
```
jlaur, line 624: I would suggest to change this back to initialCapacity = 3. Otherwise 10 will be allocated, which is way more than the maximum of 3 needed.

andrewfg: >  way more 

Wow! I dare note even think about the number of nano seconds that may be spared.. ;)

Seriously though: I changed it to the default because I had copy pasted the code (with your 3) to the tado binding, which requires 12, so that was was therefore wrong. And in such cases I think the preferred practice it to initialise lists with default settings and let Java sort things out automatically.

andrewfg:
@jlaur apropos using `new ArrayList<>(3)` versus `new ArrayList<>()` and your assertion that the former is faster than the latter: personally I suspected that with small numbers of list elements the extra cost of specifically calling the constructor with a capacity argument might outweigh any benefit gained from allocating less memory in the data array; so I thought it would be interesting to actually test it..

Using the following code, I got the results below. This was running the test on my Windows PC, inside Eclipse debugger, in debug mode (running in debug mode forces Java not to do compiler optimisations, since in non debug mode Java code optimisation eliminates the entire inner loop).

**_=> Ergo `new ArrayList<>()` is TWO times faster than `new ArrayList<>(3)`_**

### Code
```
@Test
public void testArrayList() {
    for (int j = 0; j < 20; j++) {
        Instant start = Instant.now();
        for (int i = 0; i < 100000000; i++) {
            List<Channel> x = new ArrayList<>(3); // resp. new ArrayList<>()
        }
        Instant end = Instant.now();
        System.out.println("Duration: " + Duration.between(start, end).toMillis());
    }
}
```

### Result for List<Channel> x = new ArrayList<>();
```
Duration: 642
Duration: 460
Duration: 437
Duration: 454
Duration: 495
Duration: 423
Duration: 463
Duration: 454
Duration: 417
Duration: 601
Duration: 432
Duration: 439
Duration: 463
Duration: 439
Duration: 485
Duration: 479
Duration: 423
Duration: 432
Duration: 470
Duration: 463
```

### Result for List<Channel> x = new ArrayList<>(3)
```
Duration: 1482
Duration: 1103
Duration: 1109
Duration: 1135
Duration: 1034
Duration: 1086
Duration: 1034
Duration: 1103
Duration: 1041
Duration: 1034
Duration: 1034
Duration: 1110
Duration: 1097
Duration: 1103
Duration: 1124
Duration: 1003
Duration: 1088
Duration: 1050
Duration: 1072
Duration: 1065
```

jlaur:
> apropos using `new ArrayList<>(3)` versus `new ArrayList<>()` and your assertion that the former is faster than the latter:

I never claimed that.

> personally I suspected that with small numbers of list elements the extra cost of specifically calling the constructor with a capacity argument might outweigh any benefit gained from allocating less memory in the data array;

It depends if you're most likely to add something to the list or not. If you will not add anything, it will use less memory and be faster to use default constructor, since it won't expand to capacity of 10 until first element is added. Otherwise it _should_ be cheaper to immediately reserve the capacity needed to avoid memory reallocation and copy upon expansion.

> so I thought it would be interesting to actually test it..

Agreed, it sure beats feelings and assumptions. :-)

> Using the following code, I got the results below. This was running the test on my Windows PC, inside Eclipse debugger, in debug mode (running in debug mode forces Java not to do compiler optimisations, since in non debug mode Java code optimisation eliminates the entire inner loop).

The test is not valid because:
- It doesn't take into account the probability of anything being added after initial allocation. If nothing is added, it will use less memory. If something will be added, it will allocate more memory, but only after adding first element. So the effect is delayed, which is not taken into account.
- In debug mode without JVM optimizations we won't measure anything comparable to the real situation we are interested in.

The probability of at least one element being added is high, as it seems only capabilities 9 will preserve all channels. All my 11 shades will have at least one channel removed.

I tried to perform some tests using https://www.jdoodle.com/online-java-compiler/, but unfortunately the only thing I seemed able to prove (unless I made some mistake and/or false assumptions about the optimizer - which is quite likely) is that the difference in speed is insignificant (default vs. 3 vs. 10, although default seems consistently faster than 10). I have not analyzed difference in memory usage. I tricked the optimizer by:

```java
long counter = 0;
List<Integer> x = new ArrayList<>();
x.add(integerValue);
if (x.size() > 0) {
    counter++;
}
// ...
System.out.println("Counter verification: " + counter);
```

> And in such cases I think the preferred practice it to initialise lists with default settings

You think? While _you_ would tend to use default constructor, _I_ would tend to provide _initialCapacity_ (when known) because of less memory used. Call it a habit, if you will. Whether it's the preferred practice (by who?), I don't know. I just found this also: https://stackoverflow.com/questions/57774825/initialcapacity-in-arraylist

> and let Java sort things out automatically.

Well, it has no way of knowing your intentions, that's why the additional constructor is provided, so you can give it a clue and gain better performance when dealing with large lists, and in general use only the needed amount of memory when you know how many items you are going to add. For very small lists like in our case the tests now performed seems to conclude that doesn't matter much, and the additional memory used is of course small. But at least now we learned, and I'm sure the knowledge gained can be useful in other situations. :-)

Btw, actually `ArrayList` might not even be the best data structure for this case. `LinkedList` might have been a better match, but wouldn't have left any room for initialCapacity discussions and tests. ;-)

jlaur (approved):
LGTM

andrewfg:
@jlaur thanks for merging.

.. apropos ArrayList, I guess you and I are debating over +/- 5 nanoseconds time versus +/- 28 bytes of memory; ... IMHO not the biggest worry in my life..

PR #13278, opened 2022-08-16
Adapt to AccessTokenResponse API change

### LINKED ISSUE #13277: Adapt to OAuth API changes
This issue is a placeholder for fixing breaking changes as consequence of openhab/openhab-core#3065

Current Javadoc: https://next.openhab.org/javadoc/latest/org/openhab/core/auth/client/oauth2/accesstokenresponse

jlaur:
Fix breaking change from openhab/openhab-core#3066

Fixes #13277


bundles/org.openhab.binding.homeconnect/src/main/java/org/openhab/binding/homeconnect/internal/client/HttpHelper.java
```diff
@@ -106,11 +108,13 @@ public static String getAuthorizationHeader(OAuthClientService oAuthClientServic
                     String lastToken = lastAccessToken;
                     if (lastToken == null) {
                         LoggerFactory.getLogger(HttpHelper.class).debug("The used access token was created at {}",
-                                accessTokenResponse.getCreatedOn().format(DateTimeFormatter.ISO_LOCAL_DATE_TIME));
+                                LocalDateTime.ofInstant(accessTokenResponse.getCreatedOn(), ZoneId.systemDefault())
```
lolodomo, line 111: I generally prefer using the timezone setup by the user in OH as setting rather than the OS timezone.

jlaur: This is the minimum change needed for the API change without changing any binding specific logic. Previously the usage of `LocalDateTime` also relied on system's timezone, not `TimeZoneProvider`. Also, this is only logging, so a better change would be simply dropping conversion to local datetime and use default `Instant` formatting. This would however also change logic, so should not be done in context of this 1:1 API compatiblity fix PR.

lolodomo: You're right  it was previously also based on system timezone.

mhilbush (approved):
LGTM for Ecobee

jlaur:
@openhab/add-ons-maintainers, @lolodomo - openhab/openhab-core#3066 is now approved. It will be merged after this PR has also been approved, since it will then be ready for being merged also.

wborn (approved):
LGTM!

lolodomo:
@jlaur : can you have a look to the forum duscussion? Exception when upgrading from 3.2 to 3.4 snapshot.

PR #13283, opened 2022-08-17
[fineoffsetweatherstation] Add support for the new GW2001 weather station

Andy2003:
This PR is to add support for the new [GW2001 / WH90 sensor](http://www.foshk.com/Wifi_Weather_Station/GW2001.html).

related discussion: https://community.openhab.org/t/fine-offset-weather-station-binding-beta-and-discussion/134167/66


bundles/org.openhab.binding.fineoffsetweatherstation/src/main/java/org/openhab/binding/fineoffsetweatherstation/internal/domain/Measurand.java
```diff
@@ -0,0 +1,443 @@
...
+        private final MeasureType measureType;
+
+        public ParserCustomization(ParserCustomizationType type, MeasureType measureType) {
+            this.type = type;
+            this.measureType = measureType;
+        }
+
+        public ParserCustomizationType getType() {
+            return type;
+        }
```
jlaur, line 390: This method is not used (resulting in a compiler warning).

jlaur (changes requested):
Thanks! This is really high quality code, and you even make it seem so easy (which it's not). Only two minor findings, one of which was missed on a previous occasion.

Andy2003:
Before merching this PR, please [wait for feedback from Bill](https://community.openhab.org/t/fine-offset-weather-station-binding-beta-and-discussion/134167/85?u=andy2003), if the changes works as expected.

jlaur (approved):
LGTM

bundles/org.openhab.binding.fineoffsetweatherstation/src/main/resources/OH-INF/i18n/fineoffsetweatherstation.properties
```diff
@@ -139,7 +141,7 @@ thing-type.fineoffsetweatherstation.gateway.channel.temperature-soil-channel-16.
 thing-type.fineoffsetweatherstation.gateway.channel.moisture-soil-channel-1.label = Soil Moisture Channel 1
 thing-type.fineoffsetweatherstation.gateway.channel.moisture-soil-channel-2.label = Soil Moisture Channel 2
 thing-type.fineoffsetweatherstation.gateway.channel.moisture-soil-channel-3.label = Soil Moisture Channel 3
-thing-type.fineoffsetweatherstation.gateway.channel.moisture-soil-channel-4.label = soil Moisture Channel 4
+thing-type.fineoffsetweatherstation.gateway.channel.moisture-soil-channel-4.label = Soil Moisture Channel 4
```
jlaur, line 144: Now I got curious, so will post two comments which are not related to this PR, but triggered from this change:
- Since these channels are dynamically created and not statically referenced, there is no reason for this duplication. The number of the channel could be provided by placeholder variable.
- Also, since dynamically created, it might be wise to change namespace to not potentially conflict with framework namespace for statically declared channels. For example by using "dynamic-channel" as prefix rather than "thing-type.fineoffsetweatherstation.gateway.channel". This would also clearly indicate that they are in fact dynamic and used by custom I18N implementation, not by the framework.

Just sharing these observations, which can be considered for possible future improvements.

Andy2003:
@jlaur OK, [I've got positive feedback](https://community.openhab.org/t/fine-offset-weather-station-binding-discussion/134167/111?u=andy2003), that the binding is working as  expected.

jlaur:
@Andy2003 - let's get it merged then. Thank you!

PR #13292, opened 2022-08-19
[systeminfo] Add CPU load channel, update dependencies

mherwege:

The systeminfo binding currently provides average CPU load (expressed as average number of jobs queued) over 1, 5 and 15 minute intervals with respective channels. These measures are not available on Windows. There used to be a % CPU load channel, which was removed a while ago as the underlying oshi library changed.
It is still possible to get dynamic CPU load from current versions of the oshi library and have that available on all platforms (including Windows). That is implemented in this pull request.

This PR also updates the dependency to the latest version of the [oshi libary (6.2.2)](https://github.com/oshi/oshi/blob/master/CHANGELOG.md). This version should also detect and support Apple M2 chips (not tested).


lsiepel (approved):
LGTM

bundles/org.openhab.binding.systeminfo/src/main/resources/OH-INF/thing/channels.xml
```diff
@@ -295,6 +296,14 @@
 		<config-description-ref uri="channel-type:systeminfo:highpriority_process"/>
 	</channel-type>
 
+	<channel-type id="cpuLoad">
+		<item-type>Number:Dimensionless</item-type>
```
lolodomo, line 300: You are setting `Number:Dimensionless` but you are building a DecimalType state to update the channel state.
So `Number` looks sufficient.

bundles/org.openhab.binding.systeminfo/src/main/java/org/openhab/binding/systeminfo/internal/handler/SysteminfoHandler.java
```diff
@@ -318,6 +318,13 @@ private State getInfoForChannel(ChannelUID channelUID) {
                 case CHANNEL_SENSORS_FAN_SPEED:
                     state = systeminfo.getSensorsFanSpeed(deviceIndex);
                     break;
+                case CHANNEL_CPU_LOAD:
+                    DecimalType load = systeminfo.getSystemCpuLoad();
+                    if (load != null) {
+                        load = new DecimalType(load.doubleValue() * 100);
```
lolodomo, line 324: Why not using `PercentType` which is the appropriate state for a percentage state value ?

bundles/org.openhab.binding.systeminfo/README.md
```diff
@@ -104,6 +104,7 @@ The binding introduces the following channels:
 
 | Channel ID         | Channel Description                                              | Supported item type | Default priority | Advanced |
 |--------------------|------------------------------------------------------------------|---------------------|------------------|----------|
+| load               | CPU Load in %                                                    | Number:Dimensionless| High             | False    |
```
lolodomo, line 107: See my other comments, I am not sure that `Number:Dimensionless` is more appropriate than `Number`.

lolodomo (changes requested):

bundles/org.openhab.binding.systeminfo/src/main/java/org/openhab/binding/systeminfo/internal/handler/SysteminfoHandler.java
mherwege: @lolodomo I think this would be the right thing to do.
However, I did the same thing as for the existing usedHeapPercent channel, which was the last channel added and uses Number:Dimensionless with an underlying DecimalType.
But there are also a number of older channels that represent a percentage, with just Number as the type.
Converting them all to PercentType and Number:Dimensionless would be a breaking change. Adding yet another different combination doesn't feel right.

mherwege: @lolodomo What is your view on this?

lolodomo (approved):
LGTM

lolodomo:
I am not sure if the changed item type from Number to Number:Dimensionless for the channel type load_process is or not a breaking change.

mherwege:
@lolodomo Thank you.

> I am not sure if the changed item type from Number to Number:Dimensionless for the channel type load_process is or not a breaking change.

I think it is backward compatible, so nt a breaking change. I tested and was able to link a plain Number item to a Number:Dimensionless channel.

PR #13293, opened 2022-08-19
[knx] Improve localization

holgerfriedrich:
- introduce localization of error messages
    - add new strings for common exceptions
    - provide helper functions for translation
    - add test cases

### DIFF
--- a/bundles/org.openhab.binding.knx/src/main/java/org/openhab/binding/knx/internal/handler/IPBridgeThingHandler.java
+++ b/bundles/org.openhab.binding.knx/src/main/java/org/openhab/binding/knx/internal/handler/IPBridgeThingHandler.java
@@ -23,6 +23,7 @@ import org.openhab.binding.knx.internal.client.IPClient;
 import org.openhab.binding.knx.internal.client.KNXClient;
 import org.openhab.binding.knx.internal.client.NoOpClient;
 import org.openhab.binding.knx.internal.config.IPBridgeConfiguration;
+import org.openhab.binding.knx.internal.i18n.KNXTranslationProvider;
 import org.openhab.core.net.NetworkAddressService;
 import org.openhab.core.thing.Bridge;
 import org.openhab.core.thing.ThingStatus;
@@ -87,11 +88,11 @@ public class IPBridgeThingHandler extends KNXBridgeBaseThingHandler {
         } catch (KnxSecureException e) {
             logger.debug("{}, {}", thing.getUID(), e.toString());
 
-            String message = e.getLocalizedMessage();
-            if (message == null) {
-                message = e.getClass().getSimpleName();
-            }
-            updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.CONFIGURATION_ERROR, "KNX security: " + message);
+            Throwable cause = e.getCause();
+            if (cause == null)
+                cause = e;
+            updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.CONFIGURATION_ERROR,
+                    KNXTranslationProvider.I18N.getLocalizedException(cause));
             return;
         }
 
@@ -120,7 +121,7 @@ public class IPBridgeThingHandler extends KNXBridgeBaseThingHandler {
             if (!securityAvailable) {
                 logger.warn("Bridge {} missing security configuration for secure tunnel", thing.getUID());
                 updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.CONFIGURATION_ERROR,
-                        "Security configuration missing for secure tunnel");
+                        "@text/error.knx-secure-tunnel-config-missing");
                 return;
             }
             boolean tunnelOk = ((secureTunnel.user > 0) && (secureTunnel.devKey.length == 16)
@@ -128,7 +129,7 @@ public class IPBridgeThingHandler extends KNXBridgeBaseThingHandler {
             if (!tunnelOk) {
                 logger.warn("Bridge {} incomplete security configuration for secure tunnel", thing.getUID());
                 updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.CONFIGURATION_ERROR,
-                        "Security configuration for secure tunnel is incomplete");
+                        "@text/error.knx-secure-tunnel-config-incomplete");
                 return;
             }
 
@@ -150,22 +151,21 @@ public class IPBridgeThingHandler extends KNXBridgeBaseThingHandler {
             if (!securityAvailable) {
                 logger.warn("Bridge {} missing security configuration for secure routing", thing.getUID());
                 updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.CONFIGURATION_ERROR,
-                        "Security configuration missing for secure routing");
+                        "@text/error.knx-secure-routing-config-missing");
                 return;
             }
             if (secureRouting.backboneGroupKey.length != 16) {
                 // failed to read shared backbone group key from config
-                logger.warn("Bridge {} missing security configuration for secure routing", thing.getUID());
+                logger.warn("Bridge {} invalid security configuration for secure routing", thing.getUID());
                 updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.CONFIGURATION_ERROR,
-                        "backboneGroupKey required for secure routing; please configure");
+                        "@text/error.knx-secure-routing-backbonegroupkey-invalid");
                 return;
             }
             logger.debug("KNX secure routing needs a few seconds to establish connection");
         } else {
             logger.debug("Bridge {} unknown connection type", thing.getUID());
-            updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.CONFIGURATION_ERROR, MessageFormat.format(
-                    "Unknown IP connection type {0}. Known types are either 'TUNNEL', 'ROUTER', 'SECURETUNNEL', or 'SECUREROUTER'",
-                    connectionTypeString));
+            updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.CONFIGURATION_ERROR,
+                    MessageFormat.format("@text/knx-unknown-ip-connection-type", connectionTypeString));
             return;
         }
 
--- a/bundles/org.openhab.binding.knx/src/main/java/org/openhab/binding/knx/internal/handler/IPBridgeThingHandler.java
+++ b/bundles/org.openhab.binding.knx/src/main/java/org/openhab/binding/knx/internal/handler/IPBridgeThingHandler.java
@@ -89,8 +89,9 @@ public class IPBridgeThingHandler extends KNXBridgeBaseThingHandler {
             logger.debug("{}, {}", thing.getUID(), e.toString());
 
             Throwable cause = e.getCause();
-            if (cause == null)
+            if (cause == null) {
                 cause = e;
+            }
             updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.CONFIGURATION_ERROR,
                     KNXTranslationProvider.I18N.getLocalizedException(cause));
             return;
--- a/bundles/org.openhab.binding.knx/src/main/java/org/openhab/binding/knx/internal/i18n/KNXTranslationProvider.java
+++ b/bundles/org.openhab.binding.knx/src/main/java/org/openhab/binding/knx/internal/i18n/KNXTranslationProvider.java
@@ -25,7 +25,7 @@ import org.osgi.framework.FrameworkUtil;
 /**
  * This class provides translations. It is a helper class for i18n / localization efforts.
  *
- * @implNote It is implemented as a static singelton, enforced by the single-element enum pattern.
+ * @implNote It is implemented as a static singleton, enforced by the single-element enum pattern.
  * @apiNote @set() must be called to provide tanslation service, otherwise all functions will return untranslated text.
  *          Thread safety is ensured.
  * @author Holger Friedrich - Initial contribution
@@ -80,10 +80,7 @@ public enum KNXTranslationProvider {
      * @return localized message in form <description (translated)> (<class name>, <e.getLocalizedMessage (not
      *         translated)>), empty string for null. May possibly change in further releases.
      */
-    public String getLocalizedException(final @Nullable Throwable e) {
-        if (e == null) {
-            return "";
-        }
+    public String getLocalizedException(final Throwable e) {
         StringBuffer res = new StringBuffer();
         final String exName = e.getClass().getSimpleName();
         final String key = "exception." + exName;
--- a/bundles/org.openhab.binding.knx/src/main/java/org/openhab/binding/knx/internal/i18n/KNXTranslationProvider.java
+++ b/bundles/org.openhab.binding.knx/src/main/java/org/openhab/binding/knx/internal/i18n/KNXTranslationProvider.java
@@ -62,11 +62,12 @@ public enum KNXTranslationProvider {
             // locale cannot be cached, as getLocale() will return different result once locale is changed by user
             final Locale locale = (localeProvider != null) ? localeProvider.getLocale() : Locale.getDefault();
             final String res = translationProvider.getText(bundle, text, text, locale, arguments);
-            if (res != null)
+            if (res != null) {
                 return res;
+            }
         }
         // translating not possible, we still have the original text without any subsititutions
-        if ((arguments == null) || (arguments.length == 0)) {
+        if (arguments == null || arguments.length == 0) {
             return text;
         }
         // else execute pattern substitution in untranslated text
--- a/bundles/org.openhab.binding.knx/src/main/java/org/openhab/binding/knx/internal/i18n/KNXTranslationProvider.java
+++ b/bundles/org.openhab.binding.knx/src/main/java/org/openhab/binding/knx/internal/i18n/KNXTranslationProvider.java
@@ -1,0 +1,130 @@
+/**
+ * Copyright (c) 2010-2022 Contributors to the openHAB project
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package org.openhab.binding.knx.internal.i18n;
+
+import java.text.MessageFormat;
+import java.util.Locale;
+
+import org.eclipse.jdt.annotation.NonNullByDefault;
+import org.eclipse.jdt.annotation.Nullable;
+import org.openhab.core.i18n.LocaleProvider;
+import org.openhab.core.i18n.TranslationProvider;
+import org.osgi.framework.Bundle;
+import org.osgi.framework.FrameworkUtil;
+
+/**
+ * This class provides translations. It is a helper class for i18n / localization efforts.
+ *
+ * @implNote It is implemented as a static singelton, enforced by the single-element enum pattern.
+ * @apiNote @set() must be called to provide tanslation service, otherwise all functions will return untranslated text.
+ *          Thread safety is ensured.
+ * @author Holger Friedrich - Initial contribution
+ *
+ */
+@NonNullByDefault
+public enum KNXTranslationProvider {
+    I18N;
+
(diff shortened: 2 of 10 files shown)


holgerfriedrich:
This is still work in progress. It adds German translations to the KNX binding.

I would appreciate any feedback on the general approach how to add the translations.

1. property files: German translations are based on latest Crowdin export. Is it a valid approach to copy the export to get a start for a new language?
2. property files: I have  added a few lines which are not part of the offical binding yet. If I got it correctly, I should get this into knx.propeties. Once it is merged on master, it will show up on Crowdin and can be translated. When will the automatic PR process start for a language? What is the trigger for the automatic PR process? Who approves the changes on Crowdin?
3. property files: Will the automatic PR process wipe out all unknown lines?
4. property files: Any hints regarding namespace for additional strings/error messages?
5. LocaleProvider.getlocale() -> will it always return the same locale or does this change dynamically?

Thanks!

holgerfriedrich:
@lolodomo could you give some input to the questions above or recommend whom to ask? Thanks!

jlaur:
> I would appreciate any feedback on the general approach how to add the translations.

This page might also shed some light: https://www.openhab.org/docs/developer/utils/i18n.html#managing-translations

> 1. property files: German translations are based on latest Crowdin export. Is it a valid approach to copy the export to get a start for a new language?

I don't understand this question, can you elaborate?

> 2. property files: I have  added a few lines which are not part of the offical binding yet. If I got it correctly, I should get this into knx.propeties. Once it is merged on master, it will show up on Crowdin and can be translated.

Yes.

> When will the automatic PR process start for a language?

At least once per day.

> What is the trigger for the automatic PR process?

I believe it's a time trigger, but it can also be triggered manually by admins.

> Who approves the changes on Crowdin?

Preferably proof-readers, but I'm not sure we have many yet. So actually, this question I cannot answer - @lolodomo, can you?

> 3. property files: Will the automatic PR process wipe out all unknown lines?

Yes.

> 4. property files: Any hints regarding namespace for additional strings/error messages?

Each binding has its own namespace, so you don't need to use any prefixes to secure uniqueness. Best advice for custom texts would be to try to avoid potential conflicts with framework-generated strings/prefixes.

> 5. LocaleProvider.getlocale() -> will it always return the same locale or does this change dynamically?

It will change dynamically when user changes regional settings.

holgerfriedrich:
@jlaur thanks for the feedback.
I think I will add a test and then set this PR to ready for review the next days.

> > I would appreciate any feedback on the general approach how to add the translations.

> > 1. property files: German translations are based on latest Crowdin export. Is it a valid approach to copy the export to get a start for a new language?
> 
> I don't understand this question, can you elaborate?
>

This was about how to introduce a new language to a binding. Is it intended to download the property file for the new language from Crowdin and add it to the binding - and after this the bot takes over for further updates? Or should I just add empty files like ``knx_de.properties`` and the bot will take over as well?
 

> > Who approves the changes on Crowdin?
> 
> Preferably proof-readers, but I'm not sure we have many yet. So actually, this question I cannot answer - @lolodomo, can you?
> 

That would be interesting. We seem to have unapproved translations for KNX (de). According to the docs the bot will not include them, correct?

jlaur:
> > > 1. property files: German translations are based on latest Crowdin export. Is it a valid approach to copy the export to get a start for a new language?
> > 
> > 
> > I don't understand this question, can you elaborate?
> 
> This was about how to introduce a new language to a binding. Is it intended to download the property file for the new language from Crowdin and add it to the binding - and after this the bot takes over for further updates? Or should I just add empty files like `knx_de.properties` and the bot will take over as well?

Neither. New languages are automatically provided from Crowdin.

> > > Who approves the changes on Crowdin?
> > 
> > Preferably proof-readers, but I'm not sure we have many yet. So actually, this question I cannot answer - @lolodomo, can you?
> 
> That would be interesting. We seem to have unapproved translations for KNX (de). According to the docs the bot will not include them, correct?

Correct.

holgerfriedrich:
> > This was about how to introduce a new language to a binding. Is it intended to download the property file for the new language from Crowdin and add it to the binding - and after this the bot takes over for further updates? Or should I just add empty files like `knx_de.properties` and the bot will take over as well?
> 
> Neither. New languages are automatically provided from Crowdin.

@jlaur For me it is still not clear what actually triggers a new language file to be added in the first place. For example, there is a ``knx.properties``, but no file ``knx_de.properties`` in ``resources/OH-INF/i18n`` yet. Though there are translations available at Crowdin, the bot does not create a PR. I suppose this is because ``knx_de.properties`` does not exist.

jlaur:
> For me it is still not clear what actually triggers a new language file to be added in the first place. For example, there is a `knx.properties`, but no file `knx_de.properties` in `resources/OH-INF/i18n` yet. Though there are translations available at Crowdin, the bot does not create a PR. I suppose this is because `knx_de.properties` does not exist.

The important sentence from the documentation is this one: "Please note that all texts must be translated and approved in order to be included." So for a language to be included in PR's created from Crowdin, 100% of strings must be translated, and 100% must be approved. So it is not because `knx_de.properties` does not exist, but because at least one of those criteria are not fulfilled.

In German only 82% are approved. @cweitkamp, would you be able to have a look at the pending knx translations and/or possible shed some more light on the approval process?

cweitkamp:
> In German only 82% are approved. @cweitkamp, would you be able to have a look at the pending knx translations and/or possible shed some more light on the approval process?

Done.

jlaur:
@holgerfriedrich - the German translation is now merged: #13448.

holgerfriedrich:
thanks, @cweitkamp !
@jlaur I added test cases and removed ``knx_de.properties`` from the commit as it is now already merged. I will introduce new strings via CrowdIn once this PR is merged. I took over you mock classes from hdpowerview.

holgerfriedrich:
@kaikreuzer could you please have a look? I'd Like to close this before getting deeper into the data secure topic again... Thanks!

bundles/org.openhab.binding.knx/src/test/java/org/openhab/binding/knx/internal/i18n/KNXTranslationProviderTest.java
```diff
@@ -0,0 +1,118 @@
...
+        // assertNull(KNXTranslationProvider.I18N.get(null));
+
+        // initial state, should not crash and preferrably return original strings (w. pattern substitution)
+        assertEquals(UNKNOWN, KNXTranslationProvider.I18N.get(UNKNOWN));
+        assertEquals(UNKNOWN, KNXTranslationProvider.I18N.get(UNKNOWN, 5));
+        assertEquals(UNKNOWN_NULL, KNXTranslationProvider.I18N.get(UNKNOWN_PATTERN, null, null));
+        assertEquals(UNKNOWN_FIVE, KNXTranslationProvider.I18N.get(UNKNOWN_PATTERN, 5));
+        // this would cause a compiler warning as it cannot be casted to Object[]
+        // assertNotNull(KNXTranslationProvider.I18N.get(UNKNOWN, null));
+        // assertNotNull(KNXTranslationProvider.I18N.get(UNKNOWN_PATTERN, null));
```
jlaur, line 55: Can this be omitted?

bundles/org.openhab.binding.knx/src/test/java/org/openhab/binding/knx/internal/i18n/KNXTranslationProviderTest.java
```diff
@@ -0,0 +1,118 @@
...
+    @Test
+    public void testGetLocalizedException() {
+        // initial state, should not crash
+        KNXTranslationProvider.I18N.setProvider(null, null);
+
+        final Exception e = new KNXException("error 1");
+        final Exception se = new KNXLinkClosedException("connection closed", e);
+        assertNotNull(KNXTranslationProvider.I18N.getLocalizedException(e));
+        assertNotNull(KNXTranslationProvider.I18N.getLocalizedException(se));
+        // assertEquals("Exception", KNXTranslationProvider.I18N.getLocalizedException(e));
```
jlaur, line 107: Can this be omitted?

bundles/org.openhab.binding.knx/src/test/java/org/openhab/binding/knx/internal/i18n/MockedTranslationProvider.java
```diff
@@ -0,0 +1,68 @@
...
+            entry("dynamic-channel.automation.hour-minute", "{0}hr {1}m"),
+            entry("dynamic-channel.automation.at-sunrise", "At sunrise"),
+            entry("dynamic-channel.automation.before-sunrise", "{0} before sunrise"),
+            entry("dynamic-channel.automation.after-sunrise", "{0} after sunrise"),
+            entry("dynamic-channel.automation.at-sunset", "At sunset"),
+            entry("dynamic-channel.automation.before-sunset", "{0} before sunset"),
+            entry("dynamic-channel.automation.after-sunset", "{0} after sunset"),
+            entry("dynamic-channel.automation.weekdays", "Weekdays"),
+            entry("dynamic-channel.automation.weekends", "Weekends"),
+            entry("dynamic-channel.automation.all-days", "All days"),
```
jlaur, line 50: These are copied from hdpowerview and are probably useless here.

bundles/org.openhab.binding.knx/src/main/java/org/openhab/binding/knx/internal/i18n/KNXTranslationProvider.java
```diff
@@ -0,0 +1,130 @@
...
+ * This class provides translations. It is a helper class for i18n / localization efforts.
+ *
+ * @implNote It is implemented as a static singelton, enforced by the single-element enum pattern.
+ * @apiNote @set() must be called to provide tanslation service, otherwise all functions will return untranslated text.
+ *          Thread safety is ensured.
+ * @author Holger Friedrich - Initial contribution
+ *
+ */
+@NonNullByDefault
+public enum KNXTranslationProvider {
```
jlaur, line 35: Why is this implemented as **enum**? You could consider implementing it as a **class** and inject `LocaleProvider` and `TranslationProvider` in constructor. This way you would also avoid some nullness issues. See for example:
https://github.com/openhab/openhab-addons/blob/6e4f78458d143ced35485241ab1ec87c0990dcc9/bundles/org.openhab.binding.miele/src/main/java/org/openhab/binding/miele/internal/MieleTranslationProvider.java#L29-L60

bundles/org.openhab.binding.knx/src/main/java/org/openhab/binding/knx/internal/i18n/KNXTranslationProvider.java
```diff
@@ -0,0 +1,130 @@
...
+     * get exception in user readable (and possibly localized) form
+     *
+     * @param e any exception
+     * @return localized message in form <description (translated)> (<class name>, <e.getLocalizedMessage (not
+     *         translated)>), empty string for null. May possibly change in further releases.
+     */
+    public String getLocalizedException(final @Nullable Throwable e) {
+        if (e == null) {
+            return "";
+        }
```
jlaur, line 86: This seems like potentially hiding a bug, consider making parameter `@NonNull`.

bundles/org.openhab.binding.knx/src/main/java/org/openhab/binding/knx/internal/i18n/KNXTranslationProvider.java
```diff
@@ -0,0 +1,130 @@
...
+     * @param arguments any optional arguments, will be inserted
+     * @return translated text with subsitutions if translationprovide is set and provides a translation, otherwise
+     *         returns original text with substitutions
+     */
+    public String get(final String text, @Nullable Object @Nullable... arguments) {
+        // ensure thread safety: calls to set(..) should not lead to race condition
+        final TranslationProvider translationProvider = this.translationProvider;
+        final LocaleProvider localeProvider = this.localeProvider;
+        if (translationProvider != null) {
+            // localeProvider might be null, but if not, getLoacle will return NonNull Locale
```
jlaur, line 61:
```suggestion
            // localeProvider might be null, but if not, getLocale will return NonNull Locale
```

jlaur (changes requested):
Thanks! I've added some feedback. Perhaps @kaikreuzer would like to have a second look.

bundles/org.openhab.binding.knx/src/test/java/org/openhab/binding/knx/internal/i18n/KNXTranslationProviderTest.java
holgerfriedrich: I activated this test and added a SuppressWarnings to the test case.... Maybe better to check this edge cases as well.

bundles/org.openhab.binding.knx/src/main/java/org/openhab/binding/knx/internal/i18n/KNXTranslationProvider.java
holgerfriedrich: I wanted to implement the TranslationProvider as a static singleton. This makes it easy to use from any class within the binding. Injecting LocaleProvider and TranslationProvider is done in KNXHandlerFactory, which hands it over to the static TranslationProvider instance.

I tried implement it differently before by propagating LocaleProvider and TranslationProvider all the way down from KNXHandlerFactory to all classes which require the translation service.... This was quite a mess and the static singleton approach looks much clearer.

>  * @implNote It is implemented as a static singelton, enforced by the single-element enum pattern.

Implementing a singleton using an enum is actually a quite nice pattern from the boot "Effective Java". It may look a bit strange when reading the code the first time :-)

holgerfriedrich:
@jlaur thanks for the review! I think I addressed all of your comments. I left one conversation open, explaining why I used the enum to implement the singleton.

kaikreuzer:
> @kaikreuzer could you please have a look?

@jlaur is the much better reviewer for this, so he can merge right away once he is done.

bundles/org.openhab.binding.knx/src/main/java/org/openhab/binding/knx/internal/handler/IPBridgeThingHandler.java
```diff
@@ -87,11 +88,11 @@ public void initializeLater() {
...
             logger.debug("{}, {}", thing.getUID(), e.toString());
 
-            String message = e.getLocalizedMessage();
-            if (message == null) {
-                message = e.getClass().getSimpleName();
-            }
-            updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.CONFIGURATION_ERROR, "KNX security: " + message);
+            Throwable cause = e.getCause();
+            if (cause == null)
+                cause = e;
```
jlaur, line 93:
```suggestion
            if (cause == null)
            {
                cause = e;
            }
```

bundles/org.openhab.binding.knx/src/main/java/org/openhab/binding/knx/internal/i18n/KNXTranslationProvider.java
```diff
@@ -0,0 +1,127 @@
...
+        // ensure thread safety: calls to set(..) should not lead to race condition
+        final TranslationProvider translationProvider = this.translationProvider;
+        final LocaleProvider localeProvider = this.localeProvider;
+        if (translationProvider != null) {
+            // localeProvider might be null, but if not, getLocale will return NonNull Locale
+            // locale cannot be cached, as getLocale() will return different result once locale is changed by user
+            final Locale locale = (localeProvider != null) ? localeProvider.getLocale() : Locale.getDefault();
+            final String res = translationProvider.getText(bundle, text, text, locale, arguments);
+            if (res != null)
+                return res;
```
jlaur, line 66:
```suggestion
            if (res != null)
            {
                return res;
            }
```

bundles/org.openhab.binding.knx/src/main/java/org/openhab/binding/knx/internal/i18n/KNXTranslationProvider.java
```diff
@@ -0,0 +1,127 @@
...
+        }
+        // translating not possible, we still have the original text without any subsititutions
+        if ((arguments == null) || (arguments.length == 0)) {
```
jlaur, line 69: Redundant parenthesis:
```suggestion
        if (arguments == null || arguments.length == 0) {
```

bundles/org.openhab.binding.knx/src/test/java/org/openhab/binding/knx/internal/i18n/KNXTranslationProviderTest.java
```diff
@@ -0,0 +1,115 @@
...
+    static final String KNOWN_DAY_PATTERN_KEY = "dynamic-channel.automation.after-sunset";
+    static final String KNOWN_DAY_PATTERN_VALUE = "{0} after sunset";
+    static final String KNOWN_DAY_FIVE_VALUE = "5 after sunset";
+    static final String KNOWN_DAY_ORIG_VALUE = "null after sunset";
+    static final String KNOWN_DAY_NULL_VALUE = "null after sunset";
+
+    @Test
+    public void testGetBeforeInit() {
+        // NonNull, compilation error
+        // assertNull(KNXTranslationProvider.I18N.get(null));
```
jlaur, line 45: Can this be removed?

bundles/org.openhab.binding.knx/src/test/java/org/openhab/binding/knx/internal/i18n/KNXTranslationProviderTest.java
```diff
@@ -0,0 +1,115 @@
...
+    static final String UNKNOWN_PATTERN = "unknown text {0}";
+    static final String UNKNOWN_FIVE = "unknown text 5";
+    static final String UNKNOWN_NULL = "unknown text null";
+    static final String KNOWN_DAY_KEY = "dynamic-channel.automation.weekdays";
+    static final String KNOWN_DAY_VALUE = "Weekdays";
+    static final String KNOWN_DAY_PATTERN_KEY = "dynamic-channel.automation.after-sunset";
+    static final String KNOWN_DAY_PATTERN_VALUE = "{0} after sunset";
+    static final String KNOWN_DAY_FIVE_VALUE = "5 after sunset";
+    static final String KNOWN_DAY_ORIG_VALUE = "null after sunset";
+    static final String KNOWN_DAY_NULL_VALUE = "null after sunset";
```
jlaur, line 40: I'm wondering if it would make sense with some strings from the knx binding here? The mentioned keys do not exist in as properties, so it seems quite artificial.

bundles/org.openhab.binding.knx/src/test/java/org/openhab/binding/knx/internal/i18n/MockedTranslationProvider.java
```diff
@@ -0,0 +1,55 @@
...
+ * Translation provider for unit tests.
+ *
+ * @author Jacob Laursen - Initial contribution
+ * @author Holger Friedrich - Imported from hdpowerview, adapted
+ */
+@NonNullByDefault
+public class MockedTranslationProvider implements TranslationProvider {
+
+    private static final Map<String, String> TEXTS = Map.ofEntries(
+            entry("dynamic-channel.automation.after-sunset", "{0} after sunset"),
```
jlaur, line 36: Again, seems quite artificial here. Perhaps `error.knx-unknown-ip-connection-type` would be a better choice for an example with substitution?

bundles/org.openhab.binding.knx/src/main/java/org/openhab/binding/knx/internal/i18n/KNXTranslationProvider.java
jlaur: OK.

jlaur (changes requested):
I've added some minor comments. Let's merge after you had a chance to have a look. :-)

holgerfriedrich:
> I've added some minor comments. Let's merge after you had a chance to have a look. :-)

@jlaur, thanks. Finally, you convinced me to exchanged the strings for the unit test :-)

jlaur (approved):
LGTM, thanks!

holgerfriedrich:
@cweitkamp maybe you could approve the translations for the new strings introduced in this PR once again? I have uploaded German translations to Crowdin. Thanks!

cweitkamp:
> @cweitkamp maybe you could approve the translations for the new strings introduced in this PR once again?

Done.

PR #13295, opened 2022-08-20
[Konnected] Setting auto discovered callback URL in configuration

haavar:
Setting the auto discovered callback URL in the thing configuration. This will expose the user to the discovered URL, instead of an empty form field. 

In my case the auto discovered callback URL is incorrect, and I had to dig into code / debug logs to troubleshoot. If I was exposed to the discovered callback URL in the thing settings, I may have found the issue earlier.

The auto discovered callback URL is incorrect when using reverse proxy or TLS.

Change also includes general cleanup of callback URL handling and method / variable naming.


jlaur (approved):
LGTM - thanks!

PR #13296, opened 2022-08-20
[transformation.jinja] Update jinjava to 2.6.0 from 2.5.7

ssalonen:
Updated jinjava library used by jinja transformation to 2.6.0 version (released 2021-10), from 2.5.7 (released 2021-04)

Changes compared to 2.5.7 version, from https://github.com/HubSpot/jinjava/releases

```
2021-10-29 Version 2.6.0 ([Maven Central](https://search.maven.org/artifact/com.hubspot.jinjava/jinjava/2.6.0/jar))
Create interface for object truth values
Catch concurrent modification in for loop
Add Originating Exception Message For A TemplateSyntaxException
Throw a template error when attempting to divide by zero
Make unixtimestamp behave the same as System.currentTimeMillis()
handle null argument in range function
Track Current Processed Node In The Context
Add Base 64 encode and decode filters

2021-09-03 Version 2.5.10 ([Maven Central](https://search.maven.org/artifact/com.hubspot.jinjava/jinjava/2.5.10/jar))
Make LazyExpression memoization disable-able
Add new MapELResolver with type coercion to support accessing enum keys
Add methods to remove error from interpreter,
get the last error,
and remove the last error
Pass value of throwInterpreterErrors to child contexts
Support Assignment Blocks with Set tags
Handle spaces better in for loop expressions
Support "not in"
Set propertyResolved after evaluating the AbstractCallableMethod
Limit infinite evaluation from recursive extends tags
Fix striptags to clean HTML instead of parsing
Various PRs for eager execution to support two-phase rendering.

2021-05-20 Version 2.5.8 ([Maven Central]Various PRs for eager execution to support two-phase rendering.
Add rangeLimit to JinjavaConfig
Add namespace functionality
Fix capitalize and title filters
```

Signed-off-by: Sami Salonen <ssalonen@gmail.com>

### DIFF
--- a/bundles/org.openhab.transform.jinja/pom.xml
+++ b/bundles/org.openhab.transform.jinja/pom.xml
@@ -25,7 +25,7 @@
     <dependency>
       <groupId>com.hubspot.jinjava</groupId>
       <artifactId>jinjava</artifactId>
-      <version>2.5.7</version>
+      <version>2.6.0</version>
       <scope>compile</scope>
     </dependency>
     <dependency>


jlaur (approved):
LGTM - thanks!

PR #13299, opened 2022-08-20
[lutron] Add mDNS discovery for RadioRA 3 processor

bobadair:
Adds mDNS discovery support for the new RadioRA 3 Processor.


wborn (approved):
Thanks!

PR #13301, opened 2022-08-21
[tado] Channels are visible depending on device type and capabilities

### LINKED ISSUE #13290: [tado] Create channels dynamically depending on device capabilities
### Background

Analog to #13271 and #13275 

Currently a tado zone thing can be either a central heating or an a/c device. And currently all things are created statically with the superset of all channels for all possible functions of heating and a/c devices.

So for example 'fanSpeed' and 'fanSwing' channels are statically created for a heating radiator device. Which makes no sense, and could be confusing for users.

### Proposed Solution

Depending on the device 'type' and its published 'capabilities', the appropriate channels will be created dynamically.

andrewfg:
### Background

Fixes #13290 

Analog to https://github.com/openhab/openhab-addons/pull/13271 and https://github.com/openhab/openhab-addons/pull/13275

Currently a tado zone thing can be either a central heating controller, an a/c device controller (with different levels of functionality depending on the machine's make and model), or a domestic hot water controller. And currently all things are created statically with the superset of all channels for all possible functions of heating and a/c devices. So for example 'fanSpeed' and 'fanSwing' channels are statically created for a heating radiator device. Which makes no sense, and could be confusing for users.

### Solution

In this PR, depending on the device 'type' and its published 'capabilities', the appropriate channels are created dynamically.

### Details of Changes

This PR makes the following changes and additions..
1. Adds a class `CapabilitiesSupport` that checks which type of channels are needed in a thing that is to be built around the given 'capabilities' argument and the (optional) 'zone' argument. It iterates over each of the capabilities argument's mode specific sub-capabilities to determine the maximum super set of all sub-capabilities. And it checks the power capabilities of the optional zone argument too. It sets boolean fields true or false depending whether the respective channel may or may not be needed.
2. The `TadoZoneHandler` (thing handler) class utilizes the information gathered by the `CapabilitiesSupport` to delete previously statically created channels that are not required.
3. Among other things `CapabilitiesSupport` checks if a zone controller is mains or battery powered so that if it is mains powered the respective low battery alarm channel can also be deleted. This therefore required a change in the `TadoBatteryChecker` class to support detection of mains / battery support. We also take opportunitiy in this PR to refactor the getBatteryLowAlarm() method, and use an `Instant` object rather than a `Date` object.
4. A JUnit test class has been added to confirm the `CapabilitiesSupport` class functionality.
5. The order of the channels has been changed in the thing-type xml description so that channels are displayed in the UI in a more function oriented manner (e.g. heating channels are grouped, before cooling channels, before secondary channels, etc.)
6. Also a category element has been added for channels that previously did not have such a category defined.
7. The wording of the `acPower` channel has been changed to remove potential user confusion over the meaning of "acPower" - it DOES mean "Air-conditioning Power" (i.e. whether the air-conditioning is running or not), and it does NOT mean "Alternating Current Power" (i.e. whether the unit itself is turned on or off) (for example the overall unit can be 'on' but in heating mode, in which case `heatingPower` is on (non zero value), and `acPower` is off..


andrewfg:
Below are some screenshots of three different Things with channels created dynamically..

### A. Thermostatic Radiator Valve controller device

image

### B. Multi function Air-Conditioning controller device

image

### C. Basic function Air-Conditioning controller device

Snag_2f7664

andrewfg:
@lolodomo / @jlaur this PR is analogous to what I did in recent PRs for 'velux' and 'hdpowerview' bindings to hide channels that are not required for a given device. It is a little more complex since the previous PRs concerned 1..3 channels whereas this one handles 8 or 9. To help understanding I added a "Details of Changes" section to the PR description above to explain more what is going on.

CODEOWNERS
```diff
@@ -105,7 +105,7 @@
 /bundles/org.openhab.binding.fronius/ @trokohl
 /bundles/org.openhab.binding.fsinternetradio/ @paphko
 /bundles/org.openhab.binding.ftpupload/ @paulianttila
-/bundles/org.openhab.binding.gardena/ @gerrieg
+/bundles/org.openhab.binding.gardena/ @gerrieg @andrewfg
```
lolodomo, line 108: Please move in separates PRs your assignation as maintainer for other bindings than Velux.

bundles/org.openhab.binding.tado/src/main/resources/OH-INF/thing/thing-types.xml
```diff
@@ -197,18 +193,21 @@
...
 		<label>Swing</label>
 		<description>State of AC swing (only if supported by AC)</description>
+		<category>flow</category>
 	</channel-type>
 
 	<channel-type id="light">
 		<item-type>Switch</item-type>
 		<label>Light</label>
 		<description>State of control panel light (only if supported by AC)</description>
+		<category>light</category>
```
lolodomo, line 203: The practice is to have an uppercase letter as first letter for the category.
If you can change that to be homogenous with all the other bindings.

lolodomo (changes requested):

CODEOWNERS
andrewfg: Ok. Done.

bundles/org.openhab.binding.tado/src/main/resources/OH-INF/thing/thing-types.xml
andrewfg: Ok. Done.

lolodomo (approved):
LGTM, thank you

PR #13302, opened 2022-08-21
[arcam] Initial contribution

joepadmiraal:
# [Arcam] new binding

Initial contribution of the [Arcam](https://www.arcam.co.uk/) binding.

# Description

This PR contains a new binding for the Arcam audio devices.
Currently the AVR and SA series of devices is supported.
It allows users to control and read status information from their Arcam amplifier.

I have tried to make the binding generic in order to prevent code duplication.
For this I have added a device specific class in the `devices` directory.
Unfortunately this is needed as each device has different capabilities and protocol nuances.
The protocol itself is a TCP based protocol and has different commands for the a single functionality, based on which device model you are communicating with.
The connection code contains retries etc. so temporarily connectivity issues will not break the binding. 

Each device also has a ChannelTypeProvider which allows the binding to have device specific OptionChannels, for example showing only the inputs that are actually available on that specific device.

Discovery is done with UPnP.

# Testing
This binding has been tested with the Arcam SA30

The jar can be downloaded [here](https://openhab.jfrog.io/artifactory/libs-pullrequest-local/org/openhab/addons/bundles/org.openhab.binding.arcam/3.4.0-SNAPSHOT/org.openhab.binding.arcam-3.4.0-SNAPSHOT.jar).

### DIFF
--- a/bundles/org.openhab.binding.arcam/pom.xml
+++ b/bundles/org.openhab.binding.arcam/pom.xml
@@ -1,0 +1,17 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns="http://maven.apache.org/POM/4.0.0"
+  xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd">
+
+  <modelVersion>4.0.0</modelVersion>
+
+  <parent>
+    <groupId>org.openhab.addons.bundles</groupId>
+    <artifactId>org.openhab.addons.reactor.bundles</artifactId>
+    <version>3.4.0-SNAPSHOT</version>
+  </parent>
+
+  <artifactId>org.openhab.binding.arcam</artifactId>
+
+  <name>openHAB Add-ons :: Bundles :: Arcam Binding</name>
+
+</project>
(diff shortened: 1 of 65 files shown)


bundles/org.openhab.binding.arcam/pom.xml
```diff
@@ -0,0 +1,17 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns="http://maven.apache.org/POM/4.0.0"
+  xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd">
+
+  <modelVersion>4.0.0</modelVersion>
+
+  <parent>
+    <groupId>org.openhab.addons.bundles</groupId>
+    <artifactId>org.openhab.addons.reactor.bundles</artifactId>
+    <version>3.4.0-SNAPSHOT</version>
```
wborn, line 10:
```suggestion
    <version>4.0.0-SNAPSHOT</version>
```

bundles/org.openhab.binding.arcam/src/main/resources/OH-INF/binding/binding.xml
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
lsiepel, line 1: This needs to be adapted to the new `addon.xml`, the PR to adapt the developer docs is not merged yet, so if you need more details, the corresponding issue is here: https://github.com/openhab/openhab-core/issues/2058
But you can also update your branch and look at another binding's `addon.xml` and adapt it to this initial contribution.

lsiepel:
@joepadmiraal do you still want to proceed with this pr?

joepadmiraal:
Hi @lsiepel 

I am not actively using OpenHAB anymore and I think not a lot of users are using this addon.
So I think we should drop this PR unless you can see more than a few users are relying on it.

lsiepel:
> Hi @lsiepel
> 
> I am not actively using OpenHAB anymore and I think not a lot of users are using this addon. So I think we should drop this PR unless you can see more than a few users are relying on it.

Thanks for the reply. As this binding has not been part of openHAB yet, I don’t think there will be any active users. If you abandon this PR i think it is better to close this. If ever needed it can be tracked and code can be re-used.

PR #13303, opened 2022-08-21
[mqtt] Add missing broker connection configuration parameters

wborn:
Allows to configure the following broker connection configuration parameters:

* Protocol: TCP (default), WEBSOCKETS
* MQTT Version: V3 (default), V5
* Hostname Validation: true (default), false

The defaults are chosen to remain backwards compatible.

Depends on https://github.com/openhab/openhab-core/pull/3068


jlaur (approved):
LGTM - thanks!

PR #13304, opened 2022-08-22
Improvement: Added Parameter nukiId

spacemanspiff2007:
Parameter is missing in the docs


lsiepel (changes requested):
Thanks for improving the docs. It looks like deviceType is also a parameter if i look ath the .things example.  Could you add that too? Maybe also add a bit more explenation to what the nukeId parameter does.

spacemanspiff2007:
Like that?

lsiepel (approved):

lsiepel:
> Like that?

👍

janvyb:
Replaced by https://github.com/openhab/openhab-addons/pull/13331, which adds more documentation and also fixes warning for textual items.

Issue #11391, opened 2021-10-16
[ipcamera] Add channel to show adv event and object detection data


jimtng:
@Skinah Continuing on #9129, it would be useful to get the raw data provided by the camera on smart motion detection events. I have Dahua cameras with SMD (Smart motion detection) and IVS (Intelligent Video Surveillance), Facial detection, etc.

Currently the ipcamera binding can receive events from SMD and IVS, but it only signals the start and end of such events. Dahua sends more details in the `data` field which is currently discarded / not used by the binding.

I propose adding a special channel called `lastRawData` that captures this extra data field. Dahua sends such data in a JSON format, but other brands might use a different format. It would be up to the rule writer to parse the json string and interpret this data.

Some examples of data sent by Dahua, retrieved using:
```
$ curl -s  --digest --user 'xxx:xxx' 'http://x.x.x.x:80/cgi-bin/eventManager.cgi?action=attach&codes=[All]'
```

```
Code=SmartMotionHuman;action=Start;index=0;data={
   "RegionName" : [ "Region1" ],
   "WindowId" : [ 0 ],
   "object" : [
      {
         "HumamID" : 542,
         "Rect" : [ 2072, 1808, 3440, 8184 ]
      }
   ]
}

Code=RtspSessionDisconnect;action=Start;index=0;data={
   "Device" : "192.168.x.x",
   "StreamType" : "Main",
   "UserAgent" : ""
}

Code=VideoMotion;action=Start;index=0;data={
   "Id" : [ 0 ],
   "RegionName" : [ "Region1" ],
   "SmartMotionEnable" : true
}

Code=CrossLineDetection;action=Start;index=0;data={
   "Class" : "Normal",
   "CountInGroup" : 1,
   "DetectLine" : [
      [ 2694, 2301 ],
      [ 8009, 6056 ]
   ],
   "Direction" : "RightToLeft",
   "EventID" : 10025,
   "GroupID" : 0,
   "Name" : "IVS-2",
   "Object" : {
      "Action" : "Appear",
      "BelongID" : 0,
      "BoundingBox" : [ 4592, 1280, 5552, 5648 ],
      "Center" : [ 5072, 3464 ],
      "Confidence" : 0,
      "ObjectID" : 542,
      "ObjectType" : "Unknown",
      "RelativeID" : 0,
      "Speed" : 0
   },
   "PTS" : 42955211710.0,
   "RuleID" : 5,
   "Track" : [],
   "UTC" : 1634296535,
[...]

PR #13313, opened 2022-08-24
[ipcamera] Add new channels for Dahua API-based doorphones

mesetka:
This commit is adding some new possibilities for Dahua API-based doorphones
exit button state
door contact state
accepted access card channel
door unlock channel
motion detection sensitivity level

And a few spaces cleaning for README.md.

Signed-off-by: mesetka <yarkiyluch@gmail.com>


bundles/org.openhab.binding.ipcamera/src/main/java/org/openhab/binding/ipcamera/internal/DahuaHandler.java
```diff
@@ -169,18 +169,100 @@ private void processEvent(String content) {
...
                     } else {
                         ipCameraHandler.setChannelState(CHANNEL_EXTERNAL_ALARM_INPUT2, OnOffType.ON);
+                        ipCameraHandler.logger.debug("External alarm Dahua event, content={}", content);
                     }
                 } else if ("Stop".equals(action)) {
                     if (content.contains("index=0")) {
                         ipCameraHandler.setChannelState(CHANNEL_EXTERNAL_ALARM_INPUT, OnOffType.OFF);
+                    } else if (content.contains("index=3")) {
+                        ipCameraHandler.setChannelState(CHANNEL_EXIT_BUTTON, OnOffType.OFF);
+                        ipCameraHandler.logger.debug("Dahua exit button released");
```
Skinah, line 262: Remove the logging line as it gives no extra information that is not already logged in the event.log when the channel state changes. As a general rule, never log when you're changing the state of a channel as it taxes the machine running openHAB for zero benefit.

bundles/org.openhab.binding.ipcamera/src/main/java/org/openhab/binding/ipcamera/internal/DahuaHandler.java
```diff
@@ -169,18 +169,100 @@ private void processEvent(String content) {
...
+            case "AccessControl":
+                if ("Pulse".equals(action)) {
+                    if (content.contains("\"Method\" : 1")) {
+                        startIndex = content.indexOf("CardNo", startIndex) + 11;
+                        if (startIndex > 0) {
+                            endIndex = content.indexOf(",", startIndex) - 1;
+                            String cardNo = content.substring(startIndex, endIndex);
+                            ipCameraHandler.setChannelState(CHANNEL_ACCEPTED_CARD_NUMBER, new StringType(cardNo));
+                            ipCameraHandler.setChannelState(CHANNEL_DOOR_UNLOCK, OnOffType.ON);
+                            ipCameraHandler.logger.debug("Door opened by accepted access card number: {}", cardNo);
```
Skinah, line 181: Remove the logging as the same info is already sent to event.log

bundles/org.openhab.binding.ipcamera/src/main/java/org/openhab/binding/ipcamera/internal/DahuaHandler.java
```diff
@@ -169,18 +169,100 @@ private void processEvent(String content) {
...
+                        ipCameraHandler.logger.debug("Camera Magnetic lock fail warning gone");
+                } else {
+                    ipCameraHandler.logger.debug("Unrecognised Alarm Dahua event, content={}", content);
+                }
+                break;
+            case "DoorStatus":
+                if ("Pulse".equals(action)) {
+                    if (content.contains("\"Relay\" : true")) {
+                        ipCameraHandler.setChannelState(CHANNEL_DOOR_UNLOCK, OnOffType.OFF);
+                        ipCameraHandler.logger.debug("Dahua door lock unlocked");
```
Skinah, line 234: Remove the logging as the same info is already sent to event.log

bundles/org.openhab.binding.ipcamera/src/main/java/org/openhab/binding/ipcamera/internal/DahuaHandler.java
```diff
@@ -169,18 +169,100 @@ private void processEvent(String content) {
...
+                    } else if (content.contains("\"Status\" : \"Close\"")) {
+                        ipCameraHandler.setChannelState(CHANNEL_DOOR_CONTACT, OnOffType.ON);
+                        ipCameraHandler.logger.debug("Dahua door contact closed");
```
Skinah, line 237: Remove the logging as the same info is already sent to event.log

bundles/org.openhab.binding.ipcamera/src/main/java/org/openhab/binding/ipcamera/internal/DahuaHandler.java
```diff
@@ -169,18 +169,100 @@ private void processEvent(String content) {
...
+                    } else if (content.contains("\"Status\" : \"Open\"")) {
+                        ipCameraHandler.setChannelState(CHANNEL_DOOR_CONTACT, OnOffType.OFF);
+                        ipCameraHandler.logger.debug("Dahua door contact opened");
```
Skinah, line 240: Remove the logging as the same info is already sent to event.log

bundles/org.openhab.binding.ipcamera/src/main/java/org/openhab/binding/ipcamera/internal/DahuaHandler.java
```diff
@@ -169,18 +169,100 @@ private void processEvent(String content) {
...
+                    }
+                }
+                break;
             case "AlarmLocal":
                 if ("Start".equals(action)) {
                     if (content.contains("index=0")) {
                         ipCameraHandler.setChannelState(CHANNEL_EXTERNAL_ALARM_INPUT, OnOffType.ON);
+                    } else if (content.contains("index=3")) {
+                        ipCameraHandler.setChannelState(CHANNEL_EXIT_BUTTON, OnOffType.ON);
+                        ipCameraHandler.logger.debug("Dahua exit button pressed");
```
Skinah, line 252: Remove the logging as the same info is already sent to event.log

bundles/org.openhab.binding.ipcamera/src/main/java/org/openhab/binding/ipcamera/internal/DahuaHandler.java
```diff
@@ -169,18 +169,100 @@ private void processEvent(String content) {
...
                     } else {
                         ipCameraHandler.setChannelState(CHANNEL_EXTERNAL_ALARM_INPUT2, OnOffType.ON);
+                        ipCameraHandler.logger.debug("External alarm Dahua event, content={}", content);
```
Skinah, line 255: Remove the logging IMHO I do not see the content as being helpful for a user and the event.log already has the channel changing state logged.

bundles/org.openhab.binding.ipcamera/src/main/java/org/openhab/binding/ipcamera/internal/DahuaHandler.java
```diff
@@ -169,18 +169,100 @@ private void processEvent(String content) {
...
                     }
                 } else if ("Stop".equals(action)) {
                     if (content.contains("index=0")) {
                         ipCameraHandler.setChannelState(CHANNEL_EXTERNAL_ALARM_INPUT, OnOffType.OFF);
+                    } else if (content.contains("index=3")) {
+                        ipCameraHandler.setChannelState(CHANNEL_EXIT_BUTTON, OnOffType.OFF);
+                        ipCameraHandler.logger.debug("Dahua exit button released");
                     } else {
                         ipCameraHandler.setChannelState(CHANNEL_EXTERNAL_ALARM_INPUT2, OnOffType.OFF);
+                        ipCameraHandler.logger.debug("External alarm Dahua event, content={}", content);
```
Skinah, line 265: Same here

bundles/org.openhab.binding.ipcamera/src/main/java/org/openhab/binding/ipcamera/internal/DahuaHandler.java
mesetka: As far as I remember this one could be rather interesting - there could be more indexes for different cameras/doorphones for different situations, so maybe this should be kept, just with loglevel trace maybe.

lolodomo:
The PR is badly signed.

Maybe you can check @Skinah if your review comments were all considered.

bundles/org.openhab.binding.ipcamera/src/main/java/org/openhab/binding/ipcamera/internal/DahuaHandler.java
Skinah: If it could be useful then happy to keep this one.

bundles/org.openhab.binding.ipcamera/src/main/java/org/openhab/binding/ipcamera/internal/DahuaHandler.java
```diff
@@ -352,6 +437,16 @@ public void handleCommand(ChannelUID channelUID, Command command) {
                             .sendHttpGET("/cgi-bin/configManager.cgi?action=setConfig&MotionDetect[0].Enable=false");
                 }
                 return;
+            case CHANNEL_MOTION_DETECTION_LEVEL:
+                int motionlevel = Math.round(Float.valueOf(command.toString()));
```
Skinah, line 441: This channel is a DIMMER so this line of code will fail if ON or OFF commands are sent which are valid for this channel type. You could use something like the following `if (command instanceof DecimalType)`, but a better way would be not to turn into a String and then parse and round it if a less costly approach can be used.

Skinah:
Thanks for the contribution. Before being happy with this I would like to discuss the extra channels and if they are needed, or if the lastEventData channel can be used? Please read this for the background... https://github.com/openhab/openhab-addons/issues/11391

The change to using it is pretty easy..

ipCameraHandler.motionDetected(CHANNEL_DOOR_UNLOCK);
ipCameraHandler.noMotionDetected(CHANNEL_DOOR_UNLOCK);


The decision was made not to create hundreds of channels to expose the extra event data, but to offer the data in a single channel. The number of the card in my opinion falls into this category. If every piece of advanced data from an event was filtered into separate channels, this binding would have 200+ channels and be a pain to navigate and find the right channel.

We should discuss the above point FIRST and then I would like to go over the code and the readme a little more carefully, but it looks pretty good already.

mesetka:
> Thanks for the contribution. Before being happy with this I would like to discuss the extra channels and if they are needed, or if the lastEventData channel can be used? Please read this for the background... #11391
> 
> The change to using it is pretty easy..
> 
> ipCameraHandler.motionDetected(CHANNEL_DOOR_UNLOCK); ipCameraHandler.noMotionDetected(CHANNEL_DOOR_UNLOCK);
> 
> The decision was made not to create hundreds of channels to expose the extra event data, but to offer the data in a single channel. The number of the card in my opinion falls into this category. If every piece of advanced data from an event was filtered into separate channels, this binding would have 200+ channels and be a pain to navigate and find the right channel.
> 
> We should discuss the above point FIRST and then I would like to go over the code and the readme a little more carefully, but it looks pretty good already.

Skinah:
I do agree with your comments, and I am looking for your feedback to discuss this with an open mind. Separate channels are what I prefer for the reasons you have listed, I also do not like a user scrolling through 200 channels that their model of camera does not support just to find the one which does do something... The other option would be to REMOVE the channels if the binding can ask the camera if they are supported or not with an API call? Is it possible to make an API call to see if the camera supports each channel, or is it possible to ask if the camera is a SIP capable model and then bulk remove them if the camera does not? If the approach to remove the channels were made, then you leave all the coding the way it is and only add a small amount, which could be done in this PR or a future one.

Once this is decided then I can review the readme and go over the code 1 more time.

mesetka:
> I do agree with your comments, and I am looking for your feedback to discuss this with an open mind. Separate channels are what I prefer for the reasons you have listed, I also do not like a user scrolling through 200 channels that their model of camera does not support just to find the one which does do something... The other option would be to REMOVE the channels if the binding can ask the camera if they are supported or not with an API call? Is it possible to make an API call to see if the camera supports each channel, or is it possible to ask if the camera is a SIP capable model and then bulk remove them if the camera does not? If the approach to remove the channels were made, then you leave all the coding the way it is and only add a small amount, which could be done in this PR or a future one.
> 
> Once this is decided then I can review the readme and go over the code 1 more time.
Check could be done by API call like `http://%3Cserver%3E/cgi-bin/magicBox.cgi?action=getDeviceClass`, for VTO the answer is `class=VTO`.
But I couldn't find docs about adding/deleting channels, just a few posts on community forum, including your own.

bundles/org.openhab.binding.ipcamera/README.md
```diff
@@ -264,6 +264,11 @@ The channels are kept consistent as much as possible from brand to brand to make
 | `tooDarkAlarm` | Switch (read only) | ONVIF cameras only will reflect the status of the ONVIF event of the same name. |
 | `pollImage` | Switch | This control can be used to manually start and stop using your CPU to create snapshots from a RTSP source. If you have a snapshot URL setup in the binding, only then can this control can be used to update the Image channel. |
 | `zoom` | Dimmer | Works with ONVIF cameras that can be moved. |
+| `acceptedCardNumber` | String (read only) | This channel shows the last Pass Card Number, authorized by Dahua VTO to open door. This could be used by Openhab's presence detection and security monitoring. Cards could be programmed into VTO without VTH with a help of Dahua SmartPSS soft <https://www.youtube.com/watch?v=qzR_ILT7j-0>. After that cards appear in VTO's web interface. The channel doesn't show rejected/unauthorised cards. |
```
Skinah, line 267:
```suggestion
| `acceptedCardNumber` | String (read only) | This channel shows the last accepted access card number that opened the door. The channel doesn't show rejected/unauthorized cards. |
```
1. You spelled openHAB wrongly as we are picky about the capitals in our docs.
2. The channel will get re-used by other brands most likely in the future so removing mention of brands and keeping this short I think is a better idea. 
3. I probably would prefer to call the channel `lastAcceptedCard` as I would have to check how other brands that have a similar feature work. I suspect some may use a card hash ID, some may use a users name, and some may use a index number.

bundles/org.openhab.binding.ipcamera/README.md
```diff
@@ -264,6 +264,11 @@ The channels are kept consistent as much as possible from brand to brand to make
...
+| `doorUnlock` | Switch | This channel could reflect Dahua VTO door lock state and at the same time send commands to door lock. Note that under some conditions Dahua doesn't send "lock off" message, so it's better to add expiration timer to corresponding item. |
+| `doorContact` | Switch (read only) | Reflects Dahua VTO door open/closed contact state.
+| `exitButton` | Switch (read only) | Reflects Dahua VTO exit button state. This could be used to check for exit button's long clicks/double clicks, so the button could control other gates connected to Openhab, or outdoor lights.
```
Skinah, line 270: openHAB not Openhab

bundles/org.openhab.binding.ipcamera/src/main/resources/OH-INF/thing/thing-types.xml
```diff
@@ -2672,6 +2677,44 @@
...
+		<description>Exit button pressed/released</description>
+		<category>Lock</category>
+		<state readOnly="true"/>
+	</channel-type>
+
+	<channel-type id="motionDetectionLevel" advanced="true">
+		<item-type>Dimmer</item-type>
+		<label>Motion Detection Level</label>
+		<description>By moving this control you should be able to change how sensitive the camera's motion detection is to
+			movements.</description>
```
Skinah, line 2714:
```suggestion
		<description>Change how sensitive the camera's motion detection is to
			movement.</description>
```
Shorten to make it more concise.

Skinah:
Ok that is everything I can contribute to the review. As a summary:

1. I will look at adding the deleting of channels in the future and use your suggested url and method to determine if the camera is VTO capable or not.
2. I may need to do a quick look at other brands API documentation and see if they have ID cards and if they return a number or user name if you wish to keep as is, I think it is worth changing it... This is in regards to making the `acceptedCardNumber` channel more generic and re-usable by removing the NUMBER from the channel name.
3. Address the changes suggested in the review comments.

Thanks for the PR.

mesetka:
> 1. I will look at adding the deleting of channels in the future and use your suggested url and method to determine if the camera is VTO capable or not.
> 2. I may need to do a quick look at other brands API documentation and see if they have ID cards and if they return a number or user name if you wish to keep as is, I think it is worth changing it... This is in regards to making the `acceptedCardNumber` channel more generic and re-usable by removing the NUMBER from the channel name.
> 3. Address the changes suggested in the review comments.

1. Maybe this should go farther than that - if you'll check [miio binding](https://github.com/openhab/openhab-addons/blob/main/bundles/org.openhab.binding.miio/src/main/java/org/openhab/binding/miio/internal/handler/MiIoBasicHandler.java), you can see, that Mr. [Marcel](https://github.com/marcelrv) is creating all channels dynamically. That's the only possible way for him due to all the variety of devices his binding should work with, maybe his approach is suitable for this binding's advanced channels.
2. There is a nuance in this channel, I use it to process card number by VTO itself instead of getting openHAB server into this, things do work a bit faster this way, however there is another message about mifare cards:
`Code=DoorCard;action=Pulse;index=0;data={
   "Check" : [ 0, 0 ],
   "Number" : "MIFARE CARD NUMBER",
   "Type" : 0
}
`
This one requires card number processing on openHAB itself, it's a bit slower to react on cards, but it's more universal - no need to program cards into VTO, some cards could open door only at defined time. It's for you to choose, which one must be implemented, or maybe both of them.

bundles/org.openhab.binding.ipcamera/src/main/java/org/openhab/binding/ipcamera/internal/DahuaHandler.java
mesetka: Tests required to set actually usable Motion Detection Level limits. I've found out for my VTO that I do need levels from 1 to 4, and level 4 is extremely sensitive in my case. But my VTO is placed outdoors, and objects max distance from camera is something like 10 meters. Could you test those settings with your camera? If you'll find out the same results, item's type could be changed to type Number, limited from 1 to 4.

Skinah: Sorry I do not own a VTO camera, is that a requirement to test?

bundles/org.openhab.binding.ipcamera/src/main/java/org/openhab/binding/ipcamera/internal/DahuaHandler.java
```diff
@@ -352,6 +437,16 @@ public void handleCommand(ChannelUID channelUID, Command command) {
...
+
+                if (motionlevel == 0) {
+                    ipCameraHandler.sendHttpGET("/cgi-bin/configManager.cgi?action=setConfig&MotionDetect[0].Level=1");
+                } else {
+                    ipCameraHandler.sendHttpGET(
+                            "/cgi-bin/configManager.cgi?action=setConfig&MotionDetect[0].Level=" + motionlevel);
+                }
```
Skinah, line 448:
```suggestion
if (command instanceof PercentType) {
if (PercentType.ZERO.equals(command)) {
ipCameraHandler.sendHttpGET("/cgi-bin/configManager.cgi?action=setConfig&MotionDetect[0].Level=1");
}else{
ipCameraHandler.sendHttpGET(
                            "/cgi-bin/configManager.cgi?action=setConfig&MotionDetect[0].Level=" + ((PercentType) command).intValue());
}
}
```

Please test and use this code instead of yours. This is much more CPU efficient and also will not create issues if someone does not send a slider value to the channel. PercentType is a BigDecimal internally, so it is very simple to get it as an Integer. If someone was to send a String command, my guess is your code would fail with an error of some kind.

bundles/org.openhab.binding.ipcamera/src/main/java/org/openhab/binding/ipcamera/internal/DahuaHandler.java
```diff
@@ -169,18 +169,94 @@ private void processEvent(String content) {
...
+                    if (content.contains("\"Method\" : 1")) {
+                        startIndex = content.indexOf("CardNo", startIndex) + 11;
+                        if (startIndex > 0) {
+                            endIndex = content.indexOf(",", startIndex) - 1;
+                            String cardNo = content.substring(startIndex, endIndex);
+                            ipCameraHandler.setChannelState(CHANNEL_ACCEPTED_CARD_NUMBER, new StringType(cardNo));
+                            ipCameraHandler.setChannelState(CHANNEL_DOOR_UNLOCK, OnOffType.ON);
+                        }
+                    }
+                    if (content.contains("\"Method\" : 5")) {
```
Skinah, line 183: Should this be ELSE IF instead of IF? If it matches method1 do we need to keep checking?

bundles/org.openhab.binding.ipcamera/src/main/java/org/openhab/binding/ipcamera/internal/DahuaHandler.java
```diff
@@ -169,18 +169,94 @@ private void processEvent(String content) {
...
+                        ipCameraHandler.setChannelState(CHANNEL_DOOR_UNLOCK, OnOffType.ON);
+                        ipCameraHandler.logger.debug("Door opened from button");
+                    }
+                    if (content.contains("\"Method\" : 4")) {
```
Skinah, line 187: Should this be ELSE IF instead of IF? If it matches method1 or Method 5, do we need to keep checking?

bundles/org.openhab.binding.ipcamera/src/main/java/org/openhab/binding/ipcamera/internal/DahuaHandler.java
```diff
@@ -169,18 +169,94 @@ private void processEvent(String content) {
...
+                }
+                break;
+            case "DoorCard":
+                if ("Pulse".equals(action)) {
+                    if (content.contains("\"Number\"")) {
+                        startIndex = content.indexOf("Number", startIndex) + 11;
+                        if (startIndex > 0) {
+                            endIndex = content.indexOf(",", startIndex) - 1;
+                            String cardNo = content.substring(startIndex, endIndex);
+                            ipCameraHandler.logger.debug("Door recognized an access card, {}", cardNo);
```
Skinah, line 202: Is this a bug? The code takes the time to get cardNo, then does nothing with it except to use it in the logs?

bundles/org.openhab.binding.ipcamera/src/main/java/org/openhab/binding/ipcamera/internal/DahuaHandler.java
```diff
@@ -169,18 +169,94 @@ private void processEvent(String content) {
...
+                        }
+                    }
+                } else {
+                    ipCameraHandler.logger.debug("Unrecognised Access control Dahua event, content={}", content);
+                }
+                break;
+            case "SIPRegisterResult":
+                if ("Pulse".equals(action)) {
+                    if (content.contains("\"Success\" : true"))
+                        ipCameraHandler.logger.debug("Camera register at SIP success");
```
Skinah, line 212: As above, it appears it only logs this and performs nothing but a debug log?

lolodomo:
@mesetka : any feedback?

bundles/org.openhab.binding.ipcamera/src/main/java/org/openhab/binding/ipcamera/internal/DahuaHandler.java
mesetka: Why not, if it's only for debug log - this could be implemented later as a channel when we'll find a way to hide channels. This one is rather useful for those, who want to add this channel to security monitoring - it's obvious that if the lock failed (door contact opened while the lock is closed), than it could be an intruder, who broke the lock.

bundles/org.openhab.binding.ipcamera/src/main/java/org/openhab/binding/ipcamera/internal/DahuaHandler.java
mesetka: Again - that`s a debug log, invisible in general, and as i mentioned [above](https://github.com/openhab/openhab-addons/pull/13313#issuecomment-1273310379), that's an alternative way to deal with access cards. It could be added later as a channel for those, who want openHAB to deal with cards.

bundles/org.openhab.binding.ipcamera/src/main/java/org/openhab/binding/ipcamera/internal/DahuaHandler.java
mesetka: Oh yeah, fixed.

Skinah: It is not showing as changed in this PR, can you push the updated code or tell me if I am wrong, thanks.

bundles/org.openhab.binding.ipcamera/src/main/java/org/openhab/binding/ipcamera/internal/DahuaHandler.java
Skinah: The general rule is to not use the logs as a debugger, so my opinion is that the binding should be as low impact on a system as possible if there is nothing to be gained. Since this only runs when a card is scanned and not regularly, I do not feel strongly about it, so please let me know if you wish to leave it as is as I have already reviewed it and you have already written it.... In future, please finish a feature fully as it increases lines of code that need to be maintained and more work for reviewers which we need more of.
The load on the system is not just logging, its searching strings, creating sub strings and checking what level the logger is set to as it gives no benefit in return, only a load which is why I have commented on it.

bundles/org.openhab.binding.ipcamera/src/main/java/org/openhab/binding/ipcamera/internal/DahuaHandler.java
Skinah: For reasons explained above. I do not feel strongly about it, so if you want to add these features soon, then I am happy to leave as is since it only causes a load when a card is scanned and not a regular re-occurring polling type load.

bundles/org.openhab.binding.ipcamera/src/main/java/org/openhab/binding/ipcamera/internal/DahuaHandler.java
Skinah, line 441:
```suggestion
if (command instanceof PercentType) {
int motionlevel = ((PercentType) command).toBigDecimal().intValue();
}                
```
More efficient and should prevent the ON and OFF commands from causing issues.

bundles/org.openhab.binding.ipcamera/src/main/java/org/openhab/binding/ipcamera/internal/DahuaHandler.java
mesetka: Not pushed, as I'm a bit stuck with other issue - what if we'll just organize additional channels to make channels easier to use for binding's user?
`		
<channel-group-type id="doorphoneProperties">
			<label>Doorphone Properties</label>
			<description>This is a group of channels relating to doorphone properties</description>
			<category>Door</category>
			<channels>
				<channel id="acceptedCardNumber" typeId="acceptedCardNumber"/>
				<channel id="doorUnlock" typeId="doorUnlock"/>
				<channel id="doorContact" typeId="doorContact"/>
				<channel id="exitButton" typeId="exitButton"/>
				<channel id="motionDetectionLevel" typeId="motionDetectionLevel"/>
			</channels>
		</channel-group-type>
`
This way I will just implement all VTO's additional channels. There will be a plenty of them, but they will be at the bottom and categorized.

lolodomo:
A conflict with README has to be resolved.

@Skinah : any chance you have time to review the last changes quickly ?

bundles/org.openhab.binding.ipcamera/src/main/resources/OH-INF/thing/thing-types.xml
jlaur: Setting back to unresolved since neither commented or changed.

lsiepel:
Please also run `mvn i18n:generate-default-translations`

jlaur:
> Please also run `mvn i18n:generate-default-translations`

And please also provide update instructions for the added channels - see https://github.com/openhab/openhab-docs/blob/088e0929d467d23014af613c1b3ced389c8ab0b9/developers/bindings/thing-xml.md#updating-thing-types

Skinah:
@lolodomo All my comments seem to be addressed. Have been away a lot recently.

lsiepel:
@mesetka are you able to continue on this PR and fix the i18n file and upgrade instructions?

Skinah:
It is now more then a year ago since the last comment or changes made to the PR. Please feel free to remove the tags when you do so someone starts looking at this again. Would be great to have this completed when you have the time.

lsiepel:
As this PR was stale, i tried to revive it.
- Solved the conflicts
- Fixed (two) very minor readme table markup issues
- Added thing type instructions

I think this is in a mergable state. Would be nice of @Skinah kan review.

lsiepel:
Fixed another merge conflict.

Skinah:
LGTM

lolodomo:
I am going to review this PR but please fix the conflicting file.

bundles/org.openhab.binding.ipcamera/src/main/java/org/openhab/binding/ipcamera/internal/DahuaHandler.java
```diff
@@ -172,18 +206,85 @@ private void processEvent(String content) {
...
+                        }
+                    }
+                } else {
+                    ipCameraHandler.logger.debug("Unrecognised Access control Dahua event, content={}", content);
+                }
+                break;
+            case "ProfileAlarmTransmit":
+                if ("Start".equals(action)) {
+                    if (content.contains("DoorMagnetism"))
+                        ipCameraHandler.setChannelState(CHANNEL_MAGNETIC_LOCK_WARNING, OnOffType.ON);
```
lolodomo, line 249: Please add {} around the block: if (condition) { code ]

bundles/org.openhab.binding.ipcamera/src/main/java/org/openhab/binding/ipcamera/internal/DahuaHandler.java
```diff
@@ -172,18 +206,85 @@ private void processEvent(String content) {
...
+                } else if ("Stop".equals(action)) {
+                    if (content.contains("DoorMagnetism"))
+                        ipCameraHandler.setChannelState(CHANNEL_MAGNETIC_LOCK_WARNING, OnOffType.OFF);
```
lolodomo, line 252: Please add {} around the block: if (condition) { code ]

bundles/org.openhab.binding.ipcamera/src/main/java/org/openhab/binding/ipcamera/internal/DahuaHandler.java
```diff
@@ -223,6 +325,12 @@ private void processSettings(String content) {
             ipCameraHandler.setChannelState(CHANNEL_ENABLE_MOTION_ALARM, OnOffType.OFF);
         }
 
+        // Handle MotionDetectLevel alarm
+        if (content.contains("table.MotionDetect[0].Level=")) {
+            String value = ipCameraHandler.returnValueFromString(content, "table.MotionDetect[0].Level=");
+            ipCameraHandler.setChannelState(CHANNEL_MOTION_DETECTION_LEVEL, PercentType.valueOf(value));
```
lolodomo, line 331: Look strange to use a PercentType while tjos channel type expects a number amongst 0, 1, 2, 3, 4, 5 or 6.
What do I miss ?

lolodomo (changes requested):

wborn (commented):
The default translations are also missing.

bundles/org.openhab.binding.ipcamera/src/main/resources/OH-INF/thing/thing-types.xml
```diff
@@ -3041,6 +3049,78 @@
...
+
+	<channel-type id="doorUnlock" advanced="true">
+		<item-type>Switch</item-type>
+		<label>Door Unlock</label>
+		<description>Lock/unlock.</description>
+		<category>Lock</category>
+	</channel-type>
+
+	<channel-type id="doorContact" advanced="true">
+		<item-type>Switch</item-type>
```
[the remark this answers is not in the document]
Skinah: Switch icons dynamically update too, but I agree if it's a door state it would be better so the icon by default is a door and not a switch.

bundles/org.openhab.binding.ipcamera/README.md
```diff
@@ -280,6 +280,14 @@ The channels are kept consistent as much as possible from brand to brand to make
...
 | `zoom`                       | Dimmer | RW | Works with ONVIF cameras that can be moved. |
+| `acceptedCardNumber` | String | R | This channel shows the last accepted access card number that opened the door. The channel doesn't show rejected/unauthorized cards. |
+| `unacceptedCardNumber` | String | R | This channel shows the last unaccepted access card number that was read. |
+| `doorUnlock` | Switch | RW | This channel could reflect door lock state and at the same time send commands to door lock. Note that under some conditions doorphone doesn't send "lock off" message, so it's better to add expiration timer to corresponding item. |
+| `doorContact` | Switch | R | Reflects door open/closed contact state. |
+| `exitButton` | Switch | R | Reflects exit button state. This could be used to check for exit button's long clicks/double clicks, so the button could control other gates connected to openHAB, or outdoor lights. |
+| `exitButtonEnable` | Switch | RW | This channel could be used to disable the exit button to provide additional security at night or when noone is home. |
```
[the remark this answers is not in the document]
lolodomo: In case you believe exitButtonEnable is a more appropriate name, then you have to use it with this name in other files.

bundles/org.openhab.binding.ipcamera/src/main/resources/OH-INF/thing/thing-types.xml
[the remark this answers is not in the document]
lolodomo: Door is already set as category for this channel type.
Contact is the item type matching perfectly this channel, it is to show if something is open or closed.

bundles/org.openhab.binding.ipcamera/src/main/java/org/openhab/binding/ipcamera/internal/DahuaHandler.java
lolodomo: I would expect a DecimalType with value 0, 1, .... or 6.

bundles/org.openhab.binding.ipcamera/README.md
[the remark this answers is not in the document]
lsiepel: Done

bundles/org.openhab.binding.ipcamera/src/main/resources/OH-INF/update/instructions.xml
```diff
@@ -34,6 +34,32 @@
...
+				<type>ipcamera:doorUnlock</type>
+			</add-channel>
+			<add-channel id="doorContact">
+				<type>ipcamera:doorContact</type>
+			</add-channel>
+			<add-channel id="exitButton">
+				<type>ipcamera:exitButton</type>
+			</add-channel>
+			<add-channel id="exitButtonEnable">
+				<type>ipcamera:exitButtonEnable</type>
```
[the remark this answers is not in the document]
lsiepel: Done

bundles/org.openhab.binding.ipcamera/src/main/resources/OH-INF/thing/thing-types.xml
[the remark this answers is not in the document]
lsiepel: Changed to `Contact` seems more appropiate

bundles/org.openhab.binding.ipcamera/src/main/java/org/openhab/binding/ipcamera/internal/DahuaHandler.java
lsiepel: Done

bundles/org.openhab.binding.ipcamera/src/main/java/org/openhab/binding/ipcamera/internal/DahuaHandler.java
lsiepel: Done

bundles/org.openhab.binding.ipcamera/src/main/java/org/openhab/binding/ipcamera/internal/DahuaHandler.java
lsiepel: Agree and changed.

bundles/org.openhab.binding.ipcamera/src/main/java/org/openhab/binding/ipcamera/internal/DahuaHandler.java
```diff
@@ -172,18 +206,87 @@ private void processEvent(String content) {
...
+                }
+                break;
+            case "DoorStatus":
+                if ("Pulse".equals(action)) {
+                    if (content.contains("\"Relay\" : true")) {
+                        ipCameraHandler.setChannelState(CHANNEL_DOOR_UNLOCK, OnOffType.OFF);
+                    } else if (content.contains("\"Status\" : \"Close\"")) {
+                        ipCameraHandler.setChannelState(CHANNEL_DOOR_CONTACT, OnOffType.ON);
+                    } else if (content.contains("\"Status\" : \"Open\"")) {
+                        ipCameraHandler.setChannelState(CHANNEL_DOOR_CONTACT, OnOffType.OFF);
```
lolodomo, line 266: Now that the item is of type Contact, you have to use the proper state.

lolodomo: @lsiepel : did you see that comment ?

lsiepel: Yes, nice timing, i was actually looking at it at this very moment.

lsiepel: Done

bundles/org.openhab.binding.ipcamera/src/main/java/org/openhab/binding/ipcamera/internal/DahuaHandler.java
```diff
@@ -172,18 +207,87 @@ private void processEvent(String content) {
...
+                        ipCameraHandler.setChannelState(CHANNEL_MAGNETIC_LOCK_WARNING, OnOffType.OFF);
+                    }
+                } else {
+                    ipCameraHandler.logger.debug("Unrecognised Alarm Dahua event, content={}", content);
+                }
+                break;
+            case "DoorStatus":
+                if ("Pulse".equals(action)) {
+                    if (content.contains("\"Relay\" : true")) {
+                        ipCameraHandler.setChannelState(CHANNEL_DOOR_UNLOCK, OpenClosedType.CLOSED);
```
lolodomo, line 263: Take care, CHANNEL_DOOR_UNLOCK is of type Switch.
But maybe you would like to use CHANNEL_DOOR_CONTACT here instead?

lsiepel: Thanks ofr saving me here. I was too quickly to adjust and adjusted this line too while it should have remained the same.

lolodomo (approved):
LGTM, thank you

PR #13324, opened 2022-08-26
[hdpowerview] Battery channels visible only when required

### LINKED ISSUE #13321: [hdpowerview] Show battery channels only when required
### Background

Currently shade things have two channels `batteryLevel` and `lowBattery` which are always visible, even if the shade is connected to mains power. 

In the PowerView App there is a Shade 'Power Option' which can have the following values..
- Battery Wand
- Rechargeable Battery Wand
- Hardwired Power Supply

And the `ShadeData` json element `batteryKind` reports with the following values..
1. when Shade 'Power Option' is 'Hardwired Power Supply'
2. when Shade 'Power Option' is 'Battery Wand'
3. when Shade 'Power Option' is 'Rechargeable Battery Wand'

### Proposed Enhancement

When `batteryKind` == 1 the two channels `batteryLevel` and `lowBattery` should not be visible.

andrewfg:
As explained in #13321 the batteryKind field in ShadeData indicates if the battery channels are required.

This PR makes the battery channels visible (or not) depending on the value of the batteryKind field.

Resolves #13321

### DIFF
--- a/bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/handler/HDPowerViewShadeHandler.java
+++ b/bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/handler/HDPowerViewShadeHandler.java
@@ -84,6 +84,8 @@ public class HDPowerViewShadeHandler extends AbstractHubbedThingHandler {
     private static final String DETECTED_TILT_ANYWHERE = "tiltAnywhereDetected";
     private static final ShadeCapabilitiesDatabase DB = new ShadeCapabilitiesDatabase();
 
+    private static final int HARDWIRED_POWER_SUPPLY = 1;
+
     private final Map<String, String> detectedCapabilities = new HashMap<>();
     private final Logger logger = LoggerFactory.getLogger(HDPowerViewShadeHandler.class);
 
@@ -274,7 +276,7 @@ public class HDPowerViewShadeHandler extends AbstractHubbedThingHandler {
         logger.debug("Caching capabilities {} for shade {}", capabilities.getValue(), shade.id);
         this.capabilities = capabilities;
 
-        updateDynamicChannels(capabilities);
+        updateDynamicChannels(capabilities, shade);
     }
 
     private Capabilities getCapabilitiesOrDefault() {
@@ -619,8 +621,10 @@ public class HDPowerViewShadeHandler extends AbstractHubbedThingHandler {
 
     /**
      * Remove previously statically created channels if the shade does not support them.
+     *
+     * @param shade
      */
-    private void updateDynamicChannels(Capabilities capabilities) {
+    private void updateDynamicChannels(Capabilities capabilities, ShadeData shade) {
         List<Channel> removeList = new ArrayList<>();
 
         removeListProcessChannel(removeList, CHANNEL_SHADE_POSITION, capabilities.supportsPrimary());
@@ -630,6 +634,18 @@ public class HDPowerViewShadeHandler extends AbstractHubbedThingHandler {
 
         removeListProcessChannel(removeList, CHANNEL_SHADE_VANE,
                 capabilities.supportsTiltAnywhere() || capabilities.supportsTiltOnClosed());
+
+        boolean batteryChannelsRequired;
+        try {
+            String batteryKind = shade.batteryKind;
+            batteryChannelsRequired = batteryKind == null || Integer.parseInt(batteryKind) != HARDWIRED_POWER_SUPPLY;
+        } catch (NumberFormatException e) {
+            batteryChannelsRequired = true;
+        }
+
+        removeListProcessChannel(removeList, CHANNEL_SHADE_BATTERY_LEVEL, batteryChannelsRequired);
+        removeListProcessChannel(removeList, CHANNEL_SHADE_BATTERY_VOLTAGE, batteryChannelsRequired);
+        removeListProcessChannel(removeList, CHANNEL_SHADE_LOW_BATTERY, batteryChannelsRequired);
 
         if (!removeList.isEmpty()) {
             if (logger.isDebugEnabled()) {
--- a/bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/handler/HDPowerViewShadeHandler.java
+++ b/bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/handler/HDPowerViewShadeHandler.java
@@ -622,7 +622,8 @@ public class HDPowerViewShadeHandler extends AbstractHubbedThingHandler {
     /**
      * Remove previously statically created channels if the shade does not support them.
      *
-     * @param shade
+     * @param capabilities the capabilities of the shade.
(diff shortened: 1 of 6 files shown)


bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/handler/HDPowerViewShadeHandler.java
```diff
@@ -84,6 +84,8 @@ private enum RefreshKind {
     private static final String DETECTED_TILT_ANYWHERE = "tiltAnywhereDetected";
     private static final ShadeCapabilitiesDatabase DB = new ShadeCapabilitiesDatabase();
 
+    private static final int HARDWIRED_POWER_SUPPLY = 1;
```
jlaur, line 87: This could be moved to `ShadeData` for encapsulation. Additionally, in that class a new method could be added that would either return an enum for the batteryKind (which would need to be introduced) or simply a boolean, e.g. `isHardwiredPowerSupply()`. This way data and interpretation of that data would be closely together.

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/handler/HDPowerViewShadeHandler.java
```diff
@@ -631,6 +636,18 @@ private void updateDynamicChannels(Capabilities capabilities) {
...
+        boolean batteryChannelsRequired;
+        try {
+            String batteryKind = shade.batteryKind;
+            batteryChannelsRequired = batteryKind == null || Integer.parseInt(batteryKind) != HARDWIRED_POWER_SUPPLY;
+        } catch (NumberFormatException e) {
+            batteryChannelsRequired = true;
+        }
+
+        removeListProcessChannel(removeList, CHANNEL_SHADE_BATTERY_LEVEL, batteryChannelsRequired);
+        removeListProcessChannel(removeList, CHANNEL_SHADE_BATTERY_VOLTAGE, batteryChannelsRequired);
```
jlaur, line 648: The voltage channel should not be removed as it works just as well for fixed power supplies as it does for batteries, and it can be used for verifying that the power supply (with voltage drop for long cables) is correctly set up, and for monitoring the voltage even after initial installation. Personally I have items bound to all voltage channels, also for fixed power supply.

The same is actually true for the other two channels, although they are less useful and accurate. The only problem I see is that the user now need to configure the battery kind in the PowerView app, which was previously not needed. If you want to go ahead with this change, I would therefore suggest to clearly mention this requirement in the README.

jlaur (changes requested):
Thanks. I have provided feedback for the voltage channel which should be kept. Additionally a suggestion for refactoring, which you can choose to do or not, and finally a request to update the README with the new requirement for correct configuration of battery kind.

This last point could lead to reconsideration if removing the channels is actually worth doing because it could lead to confusion when they are missing, because the user didn't pay attention to configuring the battery kind in the app. This was never important before. However, if clearly mentioned, I'm fine with it.

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/handler/HDPowerViewShadeHandler.java
andrewfg: Yes I think this is a new feature in the PowerView App, and probably in the motor firmware. When I bought my shades, I was told that rechargeable batteries were not supported. But now apparently they are..

So I will modify the ReadMe to explain that. And I will retain the battery voltage channel; although it should probably be rephrased as power supply voltage..

andrewfg: PS I think the default setting in the App is batteryKind:2 so configuring something else, and thus losing the battery channels, is an act of knowing..

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/handler/HDPowerViewShadeHandler.java
andrewfg: Yes I agree.

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/handler/HDPowerViewShadeHandler.java
jlaur: I agree, I actually woke up this morning realizing that probably default is not fixed power supply. I have no memory about the default though, so I hope you are right. Default could also be null? Anyway, I think all that needs to be explained, assuming that default is not hardwired, is a short text for the two channels that they are only present when power source is configured as battery.

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/handler/HDPowerViewShadeHandler.java
```diff
@@ -619,8 +621,11 @@ private void removeListProcessChannel(List<Channel> removeList, String channelId
 
     /**
      * Remove previously statically created channels if the shade does not support them.
```
jlaur, line 623: Since having this Javadoc, to be correct, this line would now need to be rephrased slightly. The power/battery channels are supported for all power source configurations, so they are removed when **irrelevant**, not only when **unsupported**.
```suggestion
     * Remove previously statically created channels if the shade does not support them or they are not relevant.
```

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/handler/HDPowerViewShadeHandler.java
andrewfg: In old JSON captured from my hub two years ago the element had a string value `"batteryKind":"unassigned"`, whereas in new JSON captured from my hub today it has a numeric value `"batteryKind":2` .. and I did not expressly change anything in the app, so I assume these (both old and new JSON) were/are indeed the default values.

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/handler/HDPowerViewShadeHandler.java
andrewfg: Ok.

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/handler/HDPowerViewShadeHandler.java
andrewfg: I have implemented an enum

jlaur (approved):
LGTM

PR #13326, opened 2022-08-27
[nobohub] Add more component types with serial numbers.

austvik:
Minor improvement which gives more properties to detected things, which
should make a little easier for users with these devices to identify them.

Information from https://www.hjemmeautomasjon.no/forums/topic/267-nob%C3%B8-energy-control/?&page=7


jlaur (approved):
Thanks!

PR #13329, opened 2022-08-28
[Konnected] Fixed null pointer exception on state update with missing state

haavar:
I was hitting a race condition when adding zones. The status seems to be missing from the panel's json response right after new configuration is pushed to the panel and the panel restarts. The was causing a NullPointerException.

### DIFF
--- a/bundles/org.openhab.binding.konnected/src/main/java/org/openhab/binding/konnected/internal/handler/KonnectedHandler.java
+++ b/bundles/org.openhab.binding.konnected/src/main/java/org/openhab/binding/konnected/internal/handler/KonnectedHandler.java
@@ -137,9 +137,12 @@ public class KonnectedHandler extends BaseThingHandler {
                 // check if the itemType has been defined for the zone received
                 // check the itemType of the Zone, if Contact, send the State if Temp send Temp, etc.
                 if (channelType.contains(CHANNEL_SWITCH) || channelType.contains(CHANNEL_ACTUATOR)) {
-                    OnOffType onOffType = event.getState().equalsIgnoreCase(getOnState(channel)) ? OnOffType.ON
-                            : OnOffType.OFF;
-                    updateState(channelId, onOffType);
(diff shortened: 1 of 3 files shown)


bundles/org.openhab.binding.konnected/src/main/java/org/openhab/binding/konnected/internal/handler/KonnectedHandler.java
```diff
@@ -137,9 +137,12 @@ public void handleWebHookEvent(KonnectedModuleGson event) {
                 // check if the itemType has been defined for the zone received
                 // check the itemType of the Zone, if Contact, send the State if Temp send Temp, etc.
                 if (channelType.contains(CHANNEL_SWITCH) || channelType.contains(CHANNEL_ACTUATOR)) {
-                    OnOffType onOffType = event.getState().equalsIgnoreCase(getOnState(channel)) ? OnOffType.ON
-                            : OnOffType.OFF;
-                    updateState(channelId, onOffType);
+                    logger.debug("The event state is: {}", event.getState());
```
jlaur, line 140: A local variable could be introduced for containing event state so that `getState()` wouldn't be called three times. This would also prevent a situation where state could change between calls.

jlaur (changes requested):
Thanks! One minor comment to consider.

bundles/org.openhab.binding.konnected/src/main/java/org/openhab/binding/konnected/internal/handler/KonnectedHandler.java
haavar: Sounds good. The value state can not change, but it's probably a good practice to do it that way.

I also changed the type of onvalue to integer in the config. It was annoying that it was not typed properly. The Konnected panel sends a numeric value.

jlaur (approved):
LGTM

PR #13333, opened 2022-08-29
[Konnected] Removed defaults for zone configuration parameter from the channels

haavar:
Bugfix for issue when adding channels with the default zone configured. The thing get into an invalid config state when adding a channel with the default zone selected. The only way to add channel for the default zone is to add a channel for a different zone, then go back and edit it to the default value.

The cause is that the zone parameter was marked as required and had a default.

The default value did not make sense to me, so I removed it.


jlaur:
@haavar - should this now be mentioned in the README? Since now required, I would assume it would need to be included in full example as bare minimum? https://next.openhab.org/addons/bindings/konnected/#full-example

haavar:
> @haavar - should this now be mentioned in the README? Since now required, I would assume it would need to be included in full example as bare minimum? https://next.openhab.org/addons/bindings/konnected/#full-example

I read though the readme, and I did not see anything that would need to be updated because of this change. I did however notice that the example config used the wrong property name for the zone. I updated that.

jlaur (approved):
Thanks!

PR #13337, opened 2022-08-31
[tesla] Options to control polling frequency and sleep

billfor:
Signed-off-by: bill <git@billforsyth.net>

# [ tesla ] Options to control polling frequency and sleep mode 

# Description

These are just some changes to the polling that have been useful to me
All of the changes in this PR have no effect on the binding unless the new options are set by the user.
The new options are only exposed for the Model 3 but I would expect them to work in the others (I can only test on mine).

The current binding looks at geographic position to determine inactivity and does not consider other states to infer that the fast/slow state jobs
should be continued (user in vehicle, dog mode, climate keep, camp, sentry etc...), albeit at a slower rate. All those should indicate the car is not going offline and 
the binding should continue to poll so that it does not miss any updates.

The current binding only gives 5 minutes before going into backoff mode to allow the vehicle to sleep. Often I take a short trip to the store and I come back to the car after 10 minutes and the
binding does not know what is happening because the car has been  stationary for more than 5 minutes. This 5 minute constant has been broken out into a binding parameter in case
the user wants to adjust it. In the past I used the car position returned by the binding to trigger automation to disable and enable my alarm 
so I'de prefer a better probability of having reliable information from it. 

In summary here are the specific additions:
- allow user to specify a threshold in seconds for inactivity when away from the house (homelink unavailable) and the resulting polling backoff that promotes sleeping
  The default is 5 minutes (current hardcoded setting). It will always be 5 minutes if the car is at home (homelink available) no matter what.
  This could give the user more time without missing updates if they take the car on short trips to the store.
  The parameter could better be named "Let me go to store for X minutes without missing updates when I get back"
- add a new flag to allow the binding to keep polling under the certain condtions described below. Defaults to false (current behavior)
- add "is_user_present" from the vehicle state and optionally allow binding to continue the polling at a slower rate if the vehicle is occupied and not in gear.
- if any "center_display_state" is not zero (modes: camp, dog, keep, sentry) optionally continue the polling at a slower rate because the car is not going to sleep
- add sentry state to the binding and make a decision to continue slower polling if this is on
- add new hardcoded timeout of 60 minutes for all the new "continue fastpoll" options above except camp/dog/climatekeep, after which the behavior is the same as before. 
  The assumption being that the user is going to come back to the car when in one of these modes.
- allow user to optionally use the Drive State to determine inactivity rather than the location. Maybe this provides no added benefit but it works, and is a different 
  approach than waiting for the gps to stabilize. Drivestate comesfrom the API and is used by other software to back off.
  In my case letting it use GPS results in the following occasionally:
```
28-Aug-2022 21:55:51.474 [DEBUG] [binding.tesla.internal.handler.TeslaVehicleHandler] - Calling parseAndUpdate from main Vehicle loop.
28-Aug-2022 21:55:51.476 [DEBUG] [binding.tesla.internal.handler.TeslaVehicleHandler] - Vehicle state is online
28-Aug-2022 21:55:51.477 [DEBUG] [binding.tesla.internal.handler.TeslaVehicleHandler] - Car is at home. Movemement threshold defaults to 5 min.
28-Aug-2022 21:55:51.477 [DEBUG] [binding.tesla.internal.handler.TeslaVehicleHandler] - Car has moved recently and can not sleep
28-Aug-2022 21:55:51.479 [DEBUG] [binding.tesla.internal.handler.TeslaVehicleHandler] - allowQuery set. Polling for charge, climate, gui, and mobile.

```
   But the car has not moved. It has been sitting there. In my garage perhaps the gps is not as reliable and it bounces around too much.
   In any event this new flag/parameter breaks nothing and defaults to gps so there is no change unless somebody goes out of their way to set it.
   I think it is at least worth making available as an option to expirement with. Other web sites that use the API use this method to backoff.

I also commented out the parseAndUpdate() under queryVehicleAndUpdate() because I don't understand what it was doing. queryVehicle()  calls parseAndUpdate(),
   so it seemed to be doing duplicate polling. This is probably my mistake but I didn't understand it correctly. Nevertheless commenting it out hasn't created an issue
   for me.

Thing example (everthhing default to current binding behavior. Inactivity 5 minutes, no drive state used for inactivity, and no console modes used to keep polling.)
`Thing tesla:model3:car1 "Tesla Model 3" (tesla:account:bill)  [ vin="XXXX" ]`

Thing example (the way I drive):
T`hing tesla:model3:car1 "Tesla Model 3" (tesla:account:bill)  [ vin="XXXX" , inactivity=25, useDriveState=true, useAdvancedStatesForPolling=true ]`

# Testing

I can only test this on the model3. It probably works on other models. I can add the parameters to the other models if you want.


billfor:
@kgoderis let me know if you don't want this or if the work in 13116 are going to make merging it difficult. I will just continue to maintain as patches.

In any event, the delay settings should be adjustable for short trips and some better intelligence around polling:
see https://github.com/alandtse/tesla/wiki/Polling-policy for methods.

lolodomo:
> @kgoderis let me know if you don't want this or if the work in 13116 are going to make merging it difficult. I will just continue to maintain as patches.

@kgoderis : your feedback would be welcome.

By the way, this PR has now conflicted files, so @billfor  you will have to fix them, in case this PR is not abandoned.

kaikreuzer:
@billfor I only saw this PR now (through @lolodomo's comment) - If @kgoderis does not find the time to test/review/comment, I'm happy to do so! The additions all seem to make sense. Could you please rebase your PR, so that I can test it on the latest code base?

kgoderis:
@kaikreuzer pls go ahead Sent from my iPhoneOn 2 Oct 2022, at 12:58, Kai Kreuzer ***@***.***> wrote:﻿
@billfor I only saw this PR now (through @lolodomo's comment) - If @kgoderis does not find the time to test/review/comment, I'm happy to do so! The additions all seem to make sense. Could you please rebase your PR, so that I can test it on the latest code base?

billfor:
@kaikreuzer I will rebase it to main. Do you want me to expose these new options for models other than the "3". I can only test the 3 myself but I don't have any reason to believe it wouldn't work on the others.

kaikreuzer:
@billfor Yes, we should add it to all models as I also think it will behave the same for all - after all, we are talking only to a cloud service and not to the cars directly.
But it is ok if you leave it for the model 3 for the start - in case I have review comments, you then would only have to apply them once and copy the final result for the other models afterwards.

billfor:
Updated and conflicts resolved.

bundles/org.openhab.binding.tesla/src/main/java/org/openhab/binding/tesla/internal/handler/TeslaVehicleHandler.java
```diff
@@ -150,6 +159,19 @@ public void initialize() {
...
+        useDriveState = useDriveStateParam == null ? false : useDriveStateParam;
+        Boolean useAdvancedStatesParam = (boolean) getConfig().get(TeslaBindingConstants.CONFIG_USEDADVANCEDSTATES);
+        useAdvancedStates = useAdvancedStatesParam == null ? false : useAdvancedStatesParam;
+
+        logger.debug("Inactivity period set to {} minutes for thing {}.", inactivity, getThing().getUID());
+        logger.debug("Use drivestate for inactivity setting is {} for thing {}", useDriveState, getThing().getUID());
+        logger.debug("Use advanced modes is {} for thing {}", useAdvancedStatesParam, getThing().getUID());
+
+        // the streaming API seems to be broken - let's keep the code, if it comes back one day
+        // enableEvents = (boolean) getConfig().get(TeslaBindingConstants.CONFIG_ENABLEEVENTS);
```
kaikreuzer, line 174: This is meanwhile enabled - I guess this change here is not intended?

bundles/org.openhab.binding.tesla/src/main/java/org/openhab/binding/tesla/internal/handler/TeslaVehicleHandler.java
```diff
@@ -762,9 +848,13 @@ protected Vehicle queryVehicle() {
...
-        if (vehicle != null) {
-            parseAndUpdate("queryVehicle", null, vehicleJSON);
-        }
+        /*
+         * This seems to be called by queryVehicle and is redundant?
+         * if (vehicle != null) {
+         * logger.debug("Calling parseAndUpdate from queryVehicleAndUpdate");
+         * parseAndUpdate("queryVehicle", null, vehicleJSON);
+         * }
+         */
```
kaikreuzer, line 857: Let's assume it is redundant and remove this comment again, please.

bundles/org.openhab.binding.tesla/src/main/resources/OH-INF/thing/model3.xml
```diff
@@ -145,6 +145,23 @@
 				<label>Enable Events</label>
 				<description>Enable the event stream for the vehicle</description>
 			</parameter>
+			<parameter name="inactivity" type="integer" min="5" required="false">
+				<label>Inactivity interval</label>
```
kaikreuzer, line 149:
```suggestion
				<label>Inactivity Interval</label>
```

bundles/org.openhab.binding.tesla/src/main/resources/OH-INF/thing/model3.xml
```diff
@@ -145,6 +145,23 @@
...
+				<description>The inactivity period in minutes, after which the binding stops for 20 minutes to let the car sleep.</description>
+				<default>5</default>
+			</parameter>
+			<parameter name="useDriveState" type="boolean" required="false">
+				<default>false</default>
+				<label>Use Drive State for inactivity</label>
```
kaikreuzer, line 155:
```suggestion
				<label>Use Drive State for Inactivity</label>
```

bundles/org.openhab.binding.tesla/src/main/resources/OH-INF/thing/model3.xml
```diff
@@ -145,6 +145,23 @@
...
+				<description>Use the drive state instead of location to determine vehicle inactivity.</description>
+			</parameter>
+			<parameter name="useAdvancedStatesForPolling" type="boolean" required="false">
+				<default>false</default>
+				<label>Use Camp, Climate Keep, Dog, Sentry mode, and occupancy information</label>
```
kaikreuzer, line 160:
```suggestion
				<label>Use Camp, Climate Keep, Dog, Sentry mode, and Occupancy Information</label>
```

kaikreuzer, line 160: This is fairly long for a label text - do you have any good idea for a shorter, yet descriptive label text?

kaikreuzer (commented):
Thank you for these additions!
Code looks good and it also seems to work.

One suggestion: As we have quite many config parameter by now, I would vote to mark a few of them as "advanced", so that it does not clutter the thing configuration.
I'd suggest to make all the new parameters "advanced" as well as the "enableEvents" and "allowWakeup" params, wdyt?

billfor:
Thank you for all those!  I have made all the changes and extended the new parameters to the other models. I made allowWakeup advanced but was not sure if you wanted allowWakeupForCommands to be advanced also, so I left it alone.

kaikreuzer:
Many thanks, looks good like this!
I was just almost pressing the merge button when I noticed that we still need to update the documentation as well.
Could you be so kind and document these new config parameters in the README file, so that they end up in the official docs?

billfor:
Sure. I didn't write the enableEvents.   @kgoderis  did you want me to add it? Also let me know if you want me to update the demo.items and demo.sitemap, since the one I use seems to have more stuff in it (attached).
 
[tesla.items.txt](https://github.com/openhab/openhab-addons/files/9718555/tesla.items.txt)
[tesla.sitemap.txt](https://github.com/openhab/openhab-addons/files/9718556/tesla.sitemap.txt)

kaikreuzer:
Would be great if you could add all the missing parameters (I think I also missed to add `allowWakeupForCommands` at the time to the docs 🙄). Possibly it is time to move from prosa text to a small table to describe the params.

And yes: Feel free to also enhance the examples - this is imho always helpful for users!

billfor:
I fixed the math for the expiration of the access token and added additional logging around tokens in general, in case it might help troubleshoot those folks that have been complaining that the binding times out after refreshing the token. At least the logs will make more sense.

theater:
I started playing with the binding few days back and I figured that the constants could be moved as bridge settings. Good that you're already onto it and I see also you will cover some new use cases. Great addition to the binding. Thanks a lot !

billfor:
The math change that I buried in the last commit could be important. The existing logic does not calculate the expiration time on the access token properly. In some cases it cause the binding to refresh the access token more frequently that is required. With several people complaining about the refresh not working, it would be good to give this fix to them to try, but I don't see that the CI pipeline builds store the build for the PR anywhere, so I guess they can't test until this is  merged. I agree maybe everything could be put into the bridge as long as you weren't expecting that to control a non-car like solar walls, etc...

kaikreuzer (approved):
Thanks for your work @billfor!
Let's quickly merge before you push further fixes that are completely unrelated to the title of this PR. 😄  (Note that it is better for traceability and release note generation to have small independent PRs that address one thing each.)

PR #13338, opened 2022-08-31
[pulseaudio] Apply real disconnection when needed

dalgwen:
Bug description: 
When the pulseaudio server is shutdown and relaunched, the socket is dead. But the audiosink connection doesn't properly reset the next time it is needed. The isIdle boolean prevents this disconnection/reconnection by checking if it's in use before effectively disconnecting (and of course it's in use, because we are trying to use it when we see that the connection is broken...)
Additionnally, isIdle is not in a try catch block for the pulseaudio sink : so exception within an operation doesn't free the boolean.

Actually, isIdle seems not even relevant : we should always honnor the disconnection request, even if an operation is running (disconnection requests occur when we get I/O exception, or when shutting down the bridge/thing).

This PR remove the isIdle boolean.

+Little bug fix on volume parsing: in some case (audio source), a volume request doesn't respond with a space after the comma separating left/right channel, so I change the volume pattern.

### DIFF
--- a/bundles/org.openhab.binding.pulseaudio/src/main/java/org/openhab/binding/pulseaudio/internal/PulseAudioAudioSource.java
+++ b/bundles/org.openhab.binding.pulseaudio/src/main/java/org/openhab/binding/pulseaudio/internal/PulseAudioAudioSource.java
@@ -23,7 +23,6 @@ import java.util.Set;
 import java.util.concurrent.ConcurrentLinkedQueue;
 import java.util.concurrent.Future;
 import java.util.concurrent.ScheduledExecutorService;
-import java.util.function.Consumer;
 
 import org.eclipse.jdt.annotation.NonNullByDefault;
 import org.eclipse.jdt.annotation.Nullable;
@@ -94,13 +93,9 @@ public class PulseAudioAudioSource extends PulseaudioSimpleProtocolStream implem
                     };
                     registerPipe(pipeOutput);
                     // get raw audio from the pulse audio socket
-                    return new PulseAudioStream(sourceFormat, pipeInput, (idle) -> {
-                        if (idle) {
-                            scheduleDisconnect();
-                        } else {
-                            // ensure pipe is writing
-                            startPipeWrite();
-                        }
+                    return new PulseAudioStream(sourceFormat, pipeInput, () -> {
+                        // ensure pipe is writing
+                        startPipeWrite();
                     });
                 } catch (IOException e) {
                     disconnect(); // disconnect to force clear connection in case of socket not cleanly shutdown
@@ -121,14 +116,15 @@ public class PulseAudioAudioSource extends PulseaudioSimpleProtocolStream implem
             }
         } catch (IOException e) {
             throw new AudioException(e);
-        } finally {
-            scheduleDisconnect();
         }
         throw new AudioException("Unable to create input stream");
     }
 
     private synchronized void registerPipe(PipedOutputStream pipeOutput) {
-        this.pipeOutputs.add(pipeOutput);
+        boolean isAdded = this.pipeOutputs.add(pipeOutput);
+        if (isAdded) {
+            addClientCount();
+        }
         startPipeWrite();
     }
 
@@ -186,7 +182,10 @@ public class PulseAudioAudioSource extends PulseaudioSimpleProtocolStream implem
     }
 
     private synchronized void unregisterPipe(PipedOutputStream pipeOutput) {
-        this.pipeOutputs.remove(pipeOutput);
+        boolean isRemoved = this.pipeOutputs.remove(pipeOutput);
+        if (isRemoved) {
+            minusClientCount();
+        }
         try {
             Thread.sleep(0);
         } catch (InterruptedException ignored) {
@@ -238,13 +237,13 @@ public class PulseAudioAudioSource extends PulseaudioSimpleProtocolStream implem
         private final Logger logger = LoggerFactory.getLogger(PulseAudioAudioSource.class);
         private final AudioFormat format;
         private final InputStream input;
-        private final Consumer<Boolean> setIdle;
+        private final Runnable activity;
         private boolean closed = false;
 
-        public PulseAudioStream(AudioFormat format, InputStream input, Consumer<Boolean> setIdle) {
+        public PulseAudioStream(AudioFormat format, InputStream input, Runnable activity) {
             this.input = input;
             this.format = format;
-            this.setIdle = setIdle;
+            this.activity = activity;
         }
 
         @Override
@@ -277,14 +276,13 @@ public class PulseAudioAudioSource extends PulseaudioSimpleProtocolStream implem
             if (closed) {
                 throw new IOException("Stream is closed");
             }
-            setIdle.accept(false);
+            activity.run();
             return input.read(b, off, len);
         }
 
         @Override
         public void close() throws IOException {
             closed = true;
-            setIdle.accept(true);
             input.close();
         }
     };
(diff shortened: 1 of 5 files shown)


GiviMAD:
> The "isidle" logic attempts to prevent closing the connection for a while to speed up the response time when you are using it frequently (for example when using it with the dialog processor implementation), it was not implemented by me but doesn't seems like something we want to remove.
> I think it will affect the performance of the binding for people which is already using it, as you will be tearing up and down the pulseaudio "Simple TCP Stream" protocol port on each use.

Have reviewed the code and I was wrong, it just close the socket, do not instruct the pulseaudio server to close the "Simple TCP Stream" protocol port, but still the next usage will require to check if this port is available plus reopening the connection so I don't think is a bad idea to keep the "isidle" logic if we can fix the issue in another way.

dalgwen:
Thank you for checking this PR !
You point me to something I forgot (case 3 below)

So, if I want to sum up, as far as I understand it, the isIdle boolean is : 
- set to false when we want to use the socket
- set to true when it is finished
- Its only purpose is to be checked against, when something tries to disconnect (false --> do not close, true --> ok to close)

So it prevents the socket from being disconnect, when we are playing or recording a sound.

The situations, when we want to disconnect, are : 
1- getting an IO exception --> we should force close it now, even if isIdle is false. Currently the boolean prevents the closure and the socket is in a broken state until the scheduled time out closure (which in some case could never happen, if the timeout is set to -1)
2- disabling the sink / source / bridge / bundle / whatever --> we should force close it, even if isIdle is false. Currently it is not closed: if we try to disable the thing when a sound is playing, the socket is kept open (possibly forever if the timeout is set to -1)
3- the idle timeout is expired --> Indeed, if isIdle in false, we should NOT close it. **In fact we should NOT even try to. The idle timeout and the associated scheduledDisconnection should be canceled/postponed the exact moment we start to use the socket.**
 
I think we should implement it this way: canceling the scheduledDisconnection when the socket is opened/used, instead of not closing it at the scheduledDisconnection timeout expiration if the socket is in use.
If we do it like this, we don't need the isIdle boolean, and its side effect bug with the other close cases are gone.

I will add a commit to this PR to take the third point in consideration.

lolodomo:
> I will add a commit to this PR to take the third point in consideration.

So, PR is not yet ready for review & merge ?

dalgwen:
This PR is now ready for a second review.

After the judicious comment from @GiviMAD , I complete my initial proposal with a rewrited system for the idle -> disconnection functionnality.

Instead of simply deleting the isIdle variable (which as I said in the OP was faulty and prevented in some case a needed healing disconnection) I now propose using a count of the clients using the connection, and giving it the responsibility of scheduling optional disconnection.
The isIdle variable was not thread safe, and a thread safe count seams a better logic system. It works like this : 
- When the count is down to 0 (i.e. when the last client finish its operation), a future disconnection is scheduled.
- When the count is up (a client initiate an operation), then the future disconnection is canceled.
AFAIK, it ensures that :
- no disconnection can happen if the source/sink is still in use
- a disconnection is always scheduled after all operations are done.

Alternatively, the connectIfNeeded method can also schedule a disconnection. Indeed, using solely the end of an operation to schedule a disconnection is not sufficient because the handler also initiate a dumb connection at startup, without operation, with the goal to check connectivity.

I tested this solution with a sink (with a TTS) and with a source (using rustpotter KS and volk STT) and it works.

Additionnaly, I made a little enhancement to avoid using a synchronized method for a method called very often.

bundles/org.openhab.binding.pulseaudio/src/main/java/org/openhab/binding/pulseaudio/internal/PulseAudioAudioSource.java
```diff
@@ -95,14 +93,9 @@ public void close() throws IOException {
...
                     // get raw audio from the pulse audio socket
-                    return new PulseAudioStream(sourceFormat, pipeInput, (idle) -> {
-                        setIdle(idle);
-                        if (idle) {
-                            scheduleDisconnect();
-                        } else {
-                            // ensure pipe is writing
-                            startPipeWrite();
-                        }
+                    return new PulseAudioStream(sourceFormat, pipeInput, () -> {
```
GiviMAD, line 96: I do not understand this change. Why is the consumed value ignored in the new implementation?

bundles/org.openhab.binding.pulseaudio/src/main/java/org/openhab/binding/pulseaudio/internal/PulseAudioAudioSource.java
```diff
@@ -243,13 +248,13 @@ static class PulseAudioStream extends AudioStream {
         private final Logger logger = LoggerFactory.getLogger(PulseAudioAudioSource.class);
         private final AudioFormat format;
         private final InputStream input;
-        private final Consumer<Boolean> setIdle;
+        private final Runnable activity;
```
GiviMAD, line 251: Why is this activity required? I can't realize what it does.

GiviMAD:
Hey @dalgwen, I let you a couple of comments, but it seems like a great improvement, thank you!

bundles/org.openhab.binding.pulseaudio/src/main/java/org/openhab/binding/pulseaudio/internal/PulseAudioAudioSource.java
GiviMAD: After reviewing the previous code I think it should be:
```
if (!idle) {
     // ensure pipe is writing
     startPipeWrite();
}
```

dalgwen: I change the Consumer<Boolean> to a Runnable : 
The Consumer<Boolean> is, as I understand it, a sort of on/off switch that the PulseAudioStream can call to activate or stop the  process of copying data from the source.
In the new method I try to use, I want to centralize the place where we can cause a (smart) disconnection. So, I want the PulseAudioStream to only decrement the counter.
And it seems that the "off" switch part of the Consumer is not necessary, because it was called from the PulseAudioStream.close() method, and this method calls the close() method of the PipedInputStream, which in turn can act as the central method to decrement the counter properly by calling unregisterPipe().

And by removing the need for an "off" switch, remains only the need for a "on" switch, i.e. an activity function. Which is why I change the Consumer<Boolean> to a Runnable called "activity".
I think it should be possible to also delete this "activity" runnable, because starting the startPipeWrite method once should be sufficient, but I thought "better safe than sorry".

You propose : 
`if (!idle) {
     // ensure pipe is writing
     startPipeWrite();
}`
I would like to remove the global isIdle boolean as it introduced side effect I explained in the open post. A count of active client should be safer.
And the activity Runnable is now only called from the read() method of the PulseAudioStream : so when we call it, we never are in an idle state. So the test seems not needed. Or there is something I didn't think of, which is totally possible ?

Thank for your implication with my bug.

bundles/org.openhab.binding.pulseaudio/src/main/java/org/openhab/binding/pulseaudio/internal/PulseAudioAudioSource.java
dalgwen: (see previous comment)

bundles/org.openhab.binding.pulseaudio/src/main/java/org/openhab/binding/pulseaudio/internal/PulseAudioAudioSource.java
GiviMAD: Thanks for the explanation, I was missing the consumer change to Runnable, now I see it clear.

GiviMAD (approved):
LGTM!
Thanks for the changes. I'm using this binding on a daily basis to connect to a speaker at my living room for using the voice assistant capabilities and I think I have faced the issue couple of times, so really appreciated!

lolodomo (approved):
Code LGTM, I have not tried to understand the detailed stuff with "idle" but as @GiviMAD already approved the changes, that is fine for me.

PR #13345, opened 2022-09-02
[nuvo] Fix MPS4 flag and online status reporting

mlobstein:
Fix two minor issues:

1. When thing is provisioned from the UI, the port config item was populated with a default value. This caused the MPS4 flag to be set erroneously even when using serial port connections.
2. For serial port connections, the thing was always showing as ONLINE even when the serial port was not actually connected to the amplifier.

### DIFF
--- a/bundles/org.openhab.binding.nuvo/src/main/java/org/openhab/binding/nuvo/internal/handler/NuvoHandler.java
+++ b/bundles/org.openhab.binding.nuvo/src/main/java/org/openhab/binding/nuvo/internal/handler/NuvoHandler.java
@@ -182,19 +182,18 @@ public class NuvoHandler extends BaseThingHandler implements NuvoMessageEventLis
             return;
         }
 
-        if (serialPort != null) {
+        if (serialPort != null && !serialPort.isEmpty()) {
             connector = new NuvoSerialConnector(serialPortManager, serialPort, uid);
         } else if (port != null) {
             connector = new NuvoIpConnector(host, port, uid);
+            this.isMps4 = (port.intValue() == MPS4_PORT);
+            if (this.isMps4) {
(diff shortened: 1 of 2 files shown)


bundles/org.openhab.binding.nuvo/src/main/java/org/openhab/binding/nuvo/internal/handler/NuvoHandler.java
```diff
@@ -598,21 +597,15 @@ private void scheduleReconnectJob() {
...
                     }
                 } else {
-                    error = "Reconnection failed";
-                }
-                if (error != null) {
-                    updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.COMMUNICATION_ERROR, error);
+                    updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.COMMUNICATION_ERROR, "Reconnection failed");
                     closeConnection();
-                } else {
-                    updateStatus(ThingStatus.ONLINE, ThingStatusDetail.NONE, this.versionString);
```
lolodomo, line 615: You want now to set the thing status ONLINE only when something is received through the connection ?
In practice, will a message be received systematically quickly after the connection ?

mlobstein: Yes, for serial port connections, pollStatus() will cause messages to be received almost immediately. For MPS4, a ping message will arrive within a few seconds.

lolodomo: Ok

lolodomo (approved):
LGTM

PR #13354, opened 2022-09-04
[Konnected] Using default onvalue if configured is null

haavar:
During testing I found that sometimes configuration for onvalue can be null. 

Maybe it's a bad assumption, but I assumed that the configuration would be populated with the default value set in thing-types.xml. That does not seem to be the case, and the application need to fall back to the default value.


bundles/org.openhab.binding.konnected/src/main/java/org/openhab/binding/konnected/internal/handler/KonnectedHandler.java
```diff
@@ -526,6 +527,7 @@ private void sendSetSwitchState(String thingId, String payloadString) throws Kon
     }
 
     private int getOnState(Channel channel) {
-        return ((Number) channel.getConfiguration().get(CHANNEL_ONVALUE)).intValue();
+        Number configuredOnValue = (Number) (channel.getConfiguration().get(CHANNEL_ONVALUE));
```
jlaur, line 530: If you would like to consolidate default values for this configuration, you could consider providing a DTO class and use `getConfiguration().as(ClassName.class)` here. Example with some default values:
https://github.com/openhab/openhab-addons/blob/5a11f1420ed201a1dcca24e82251917c2d9d72f0/bundles/org.openhab.binding.boschindego/src/main/java/org/openhab/binding/boschindego/internal/config/BoschIndegoConfiguration.java#L27-L28

And usage:
https://github.com/openhab/openhab-addons/blob/5a11f1420ed201a1dcca24e82251917c2d9d72f0/bundles/org.openhab.binding.boschindego/src/main/java/org/openhab/binding/boschindego/internal/handler/BoschIndegoHandler.java#L113-L115

haavar: Thanks, I looked at it again and it does seem to clean up the code by using a DTC class for the channel config.

bundles/org.openhab.binding.konnected/src/main/resources/OH-INF/thing/thing-types.xml
```diff
@@ -246,15 +246,15 @@
...
 			</parameter>
-			<parameter name="tempsensorType" type="boolean">
+			<parameter name="dht22" type="boolean">
 				<label>DHT22</label>
 				<description>Is the sensor a dht22 or a ds18b20? Set to true for dht22 sensor</description>
 			</parameter>
-			<parameter name="pollinterval" type="integer">
+			<parameter name="pollInterval" type="integer">
 				<label>Poll Interval</label>
 				<description>The interval in minutes to poll the sensor.</description>
```
jlaur, line 255: For consistency with default value in code:
```suggestion
				<description>The interval in minutes to poll the sensor.</description>
				<default>3</default>
```

bundles/org.openhab.binding.konnected/src/main/resources/OH-INF/thing/thing-types.xml
```diff
@@ -67,15 +67,15 @@
...
```
jlaur, line 76: For consistency with default value in code:
```suggestion
				<description>The interval in minutes to poll the sensor.</description>
				<default>3</default>
```

bundles/org.openhab.binding.konnected/src/main/java/org/openhab/binding/konnected/internal/ZoneConfiguration.java
```diff
@@ -0,0 +1,29 @@
...
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package org.openhab.binding.konnected.internal;
+
+/**
+ * The {@link ZoneConfiguration} class contains the configuration parameters for the zone.
+ *
+ * @author Haavar Valeur
+ */
+public class ZoneConfiguration {
```
jlaur, line 20: Please annotate with `@NonNullByDefault`.

bundles/org.openhab.binding.konnected/src/main/java/org/openhab/binding/konnected/internal/ZoneConfiguration.java
```diff
@@ -0,0 +1,29 @@
...
+    public String zone;
+    public boolean dht22;
+    public String ds18b20Address;
+    public int pollInterval = 3;
+    public Integer times;
+    public Integer momentary;
+    public Integer pause;
```
jlaur, line 27: Why `Integer` for these three variables as opposed to simple types `int` or `long` (depending on size requirements)?

jlaur (changes requested):
Thanks! I have added a few comments. Also checked openhab/openhab-distro#1412, and it seems like existing compatibility note also covers these breaking changes.

bundles/org.openhab.binding.konnected/src/main/java/org/openhab/binding/konnected/internal/ZoneConfiguration.java
haavar: They are nullable, and I think we need to tell the difference between 0 and null.

bundles/org.openhab.binding.konnected/src/main/java/org/openhab/binding/konnected/internal/ZoneConfiguration.java
haavar: Most of these are nullable, but I guess if that's the convention I could do that.

jlaur (approved):
LGTM

PR #13355, opened 2022-09-04
[hdpowerview] Implementation of Generation 3 API

andrewfg:
### Background
See https://github.com/openhab/openhab-addons/issues/12678

This is the second of (probably) three Pull Requests for integrating Hunter Douglas Generation 3 hubs into the binding (which currently only supports Generation 1 & 2 hubs).

This PR depends on #13352 being merged first.

### Content of this Pull Request

PR #13352 provides the basic architecture and base abstract "hub facing" DTO classes for supporting hubs of different generations; and it also provides the respective implementation classes for Generation 1/2 hubs.

This PR adds the "hub facing" class implementations for Generation 3 hubs.

A further PR is foreseen after this, in order to make adjustments to optional channels, and update the read me, (among possibly other things).

### Specifics of Changes

The following classes have been added as Generation 3 implementations of the "hub facing" DTO classes that were previously defined in the previous PR..

- HDPowerViewWebTargetsV3 (implementation)
- ShadeDataV3 (implementation)
- ShadePositionV3 (implementation)
- SceneV3 (implementation)
- ScheduledEventV3 (implementation)

The following additions have been made as follows..
- An 'InfoV3' DTO class has been added to support the gateway info function (new to Generation 3)
- Two DTO classes 'SseSceneV3' and 'SseShadeV3' have been added to support SSE  (new to Generation 3)
- HDPowerViewWebTargetsV3 includes SSE support and calls the SseSinkV3 interface (below)
- The interface for SSE call-backs has been defined in the SseSinkV3 interface class
- Junit test class added for the Generation 3 DTOs

The following changes have been made in the following existing classes as follows..
- The bridge handler has been extended to support the following..
1. Pings the hub to check if it is Generation 1/2 or 3, and then automatically uses a HDPowerViewWebTargetsV1 or HDPowerViewWebTargetsV3 web targets as required.
2. Adds support for SSE call-backs by implementing the SseSinkV3 interface
3. Uses less frequent polling when subscribed to SSE; although it still polls at the 'refresh' interval in case the SSE connection may have been lost; (and in that case it automatically reconnects to the SSE server)
4. Adds a version check method for forking generation specific functionality
- The shade handler has been extended to support the following..
1. Adds a version check method for forking generation specific functionality

### Java Package Structure

In cases where the classes have been split into base and implementation classes, the base classes are in the package where the original combined classes was located, and the split V1 implementation classes are now in a ._v1 sub package, and the V3 implementation classes are in a ._v3 sub package.

Signed-off-by: Andrew Fiddian-Green [software@whitebear.ch](mailto:software@whitebear.ch)


andrewfg:
For testing purposes the Jar file is here..

[org.openhab.binding.hdpowerview-3.4.0-SNAPSHOT.jar.zip](https://github.com/openhab/openhab-addons/files/9546958/org.openhab.binding.hdpowerview-3.4.0-SNAPSHOT.jar.zip)

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/gen3/dto/SceneEvent3.java
```diff
@@ -0,0 +1,36 @@
...
+/**
+ * Scene SSE event object as supplied an HD PowerView hub of Generation 3.
+ *
+ * @author Andrew Fiddian-Green - Initial contribution
+ */
+@NonNullByDefault
+public class SceneEvent3 {
+    private int id;
+    // private @NonNullByDefault({}) String evt;
+    // private @NonNullByDefault({}) String isoDate;
```
jlaur, line 26: Can this be removed?

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/gen3/dto/ShadeEvent3.java
```diff
@@ -0,0 +1,38 @@
...
+ * Shade SSE event object as supplied an HD PowerView hub of Generation 3.
+ *
+ * @author Andrew Fiddian-Green - Initial contribution
+ */
+@NonNullByDefault
+public class ShadeEvent3 {
+    private int id;
+    // private @NonNullByDefault({}) String evt;
+    // private @NonNullByDefault({}) String isoDate;
+    // private @NonNullByDefault({}) String bleName;
```
jlaur, line 27: Can this be removed?

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/gen3/dto/ShadeEvent3.java
```diff
@@ -0,0 +1,38 @@
...
+    private @NonNullByDefault({}) ShadePosition3 currentPositions;
+    // private @NonNullByDefault({}) ShadePosition3 targetPositions;
```
jlaur, line 29: Can this be removed?

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/gen3/dto/ShadePosition3.java
```diff
@@ -0,0 +1,86 @@
...
+ * The position of a single shade, as returned by an HD PowerView hub of Generation 3
+ *
+ * @author Andrew Fiddian-Green - Initial contribution
+ */
+@NonNullByDefault
+public class ShadePosition3 {
+    private @NonNullByDefault({}) Double primary;
+    private @NonNullByDefault({}) Double secondary;
+    private @NonNullByDefault({}) Double tilt;
+    // private @NonNullByDefault({}) Double velocity;
```
jlaur, line 31: Can this be removed?

bundles/org.openhab.binding.hdpowerview/src/main/resources/OH-INF/thing/bridge.xml
```diff
@@ -47,4 +47,33 @@
 		</config-description>
 	</bridge-type>
 
+	<bridge-type id="gen3">
+		<label>PowerView Gen3 Hub</label>
```
jlaur, line 51:
```suggestion
		<label>PowerView Gen3 Gateway</label>
```

bundles/org.openhab.binding.hdpowerview/src/main/resources/OH-INF/thing/bridge.xml
```diff
@@ -47,4 +47,33 @@
...
+		<description>Hunter Douglas (Luxaflex) PowerView Hub Generation 3</description>
```
jlaur, line 52:
```suggestion
		<description>Hunter Douglas (Luxaflex) PowerView Generation 3 Gateway/Gateway Pro</description>
```

bundles/org.openhab.binding.hdpowerview/src/main/resources/OH-INF/thing/bridge.xml
```diff
@@ -47,4 +47,33 @@
...
+
+		<channel-groups>
+			<channel-group id="scenes" typeId="scenes"/>
+		</channel-groups>
+
+		<properties>
+			<property name="vendor">Hunter Douglas (Luxaflex)</property>
+			<property name="modelId">PowerView Hub</property>
```
jlaur, line 60:
```suggestion
			<property name="modelId">PowerView Gateway</property>
```

bundles/org.openhab.binding.hdpowerview/src/main/resources/OH-INF/thing/bridge.xml
```diff
@@ -47,4 +47,33 @@
...
+
+		<config-description>
+			<parameter name="host" type="text" required="true">
+				<label>Host</label>
+				<description>The Host address of the PowerView Hub</description>
+				<context>network-address</context>
+			</parameter>
+			<parameter name="hardRefresh" type="integer" required="false">
+				<label>Hard Refresh Interval</label>
+				<description>The number of minutes between hard refreshes from the PowerView Hub</description>
```
jlaur, line 73:
```suggestion
				<description>The number of minutes between hard refreshes from the PowerView Gateway</description>
```

bundles/org.openhab.binding.hdpowerview/src/main/resources/OH-INF/thing/bridge.xml
```diff
@@ -47,4 +47,33 @@
 		</config-description>
 	</bridge-type>
 
+	<bridge-type id="gen3">
```
jlaur, line 50: Since Gen 3 is a whole line of new products, probably we should name the bridge according to its type rather than generation alone. I would suggest either:
```suggestion
	<bridge-type id="gateway">
```

or:
```suggestion
	<bridge-type id="gateway3">
```
(which would be consistent with "shade3")

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/discovery/HDPowerViewHubDiscoveryParticipant.java
```diff
@@ -43,7 +43,7 @@ public class HDPowerViewHubDiscoveryParticipant implements MDNSDiscoveryParticip
 
     private final Logger logger = LoggerFactory.getLogger(HDPowerViewHubDiscoveryParticipant.class);
 
-    private static final Pattern VALID_IP_V4_ADDRESS = Pattern
+    public static final Pattern VALID_IP_V4_ADDRESS = Pattern
```
jlaur, line 46: Consider extracting this to some common class so that `HDPowerViewDeviceDiscoveryService3` won't have to reference this class. Perhaps even `HDPowerViewBindingConstants`?

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/database/ShadeCapabilitiesDatabase.java
```diff
@@ -68,8 +68,12 @@ public class ShadeCapabilitiesDatabase {
             new Type( 7).capabilities(6).text("Top Down"),
             new Type( 8).capabilities(7).text("Duette Top Down Bottom Up"),
             new Type( 9).capabilities(7).text("Duette DuoLite Top Down Bottom Up"),
+            new Type(10).capabilities(0).text("Duette/Applause Skylift"),
             new Type(18).capabilities(1).text("Pirouette"),
             new Type(23).capabilities(1).text("Silhouette"),
+            new Type(26).capabilities(3).text("Skyline Panel, Left Stack"),
+            new Type(27).capabilities(3).text("Skyline Panel, Right Stack"),
+            new Type(28).capabilities(3).text("Skyline Panel, Split Stack"),
```
jlaur, line 76: Rebase needed as this was already merged in #13574?

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/gen3/discovery/HDPowerViewDeviceDiscoveryService3.java
```diff
@@ -0,0 +1,111 @@
...
+ * @author Andrew Fiddian-Green - Initial contribution
+ */
+@NonNullByDefault
+public class HDPowerViewDeviceDiscoveryService3 extends AbstractDiscoveryService {
+
+    private final Logger logger = LoggerFactory.getLogger(HDPowerViewDeviceDiscoveryService3.class);
+    private final HDPowerViewHubHandler3 hub;
+    private final Runnable scanner;
+    private @Nullable ScheduledFuture<?> backgroundFuture;
+    private final ShadeCapabilitiesDatabase db = new ShadeCapabilitiesDatabase();
```
jlaur, line 46: Unused.

jlaur (changes requested):
Awesome, @andrewfg, this looks really good and clean to me - good job! I already added a few initial comments, but will need more time for going into the nitty gritty details.

How progressed is it, did you have a chance to test it with actual Gen 3 equipment through VPN?

Regarding the structure:
- I believe you can remove the "3" suffix from all DTO classes as there are no namespace conflicts.
- I'm in doubt whether I like all Gen 3 classes being in a completely separate tree as opposed to existing alongside the old classes. See below.

With the new hierarchy a lot of classes seems "hidden". So I would consider instead of:

- api/responses/shade
- discovery/HDPowerViewDeviceDiscoveryService
- handler/HDPowerViewHubHandler
- gen3/discovery/HDPowerViewDeviceDiscoveryService3
- gen3/handler/HDPowerViewHubHandler3

to have it flattened like this:

- dto/responses/Scenes
- dto/responses/Shade
- dto/gen3/Shade
- dto/gen3/Scene
- discovery/HDPowerViewDeviceDiscoveryService
- discovery/HDPowerViewDeviceDiscoveryService3
- handler/HDPowerViewHubHandler
handler/HDPowerViewHubHandler3

andrewfg:
> How progressed is it, did you have a chance to test it with actual Gen 3 equipment through VPN?

No. Unfortunately @vves seems to have other priorities at the moment. => ???

jlk:
@andrewfg if it helps, I have a new gen3 install that I'm willing to alpha test with. 😃

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/database/ShadeCapabilitiesDatabase.java
andrewfg: Done.

andrewfg:
> I have a new gen3 install that I'm willing to alpha test with

@jlaur for me my next step will be to confirm that Gen1/2 is not broken.

jlaur:
> @andrewfg if it helps, I have a new gen3 install that I'm willing to alpha test with. 😃

That's good to know, thanks! However, I'm afraid that the API is not finalized/published yet, so I don't think it will be possible to test with your firmware just yet. @andrewfg - for transparency, perhaps the PR should be marked as draft for now?

andrewfg:
> you can remove the "3" suffix from all DTO classes as there are no namespace conflicts.
> I'm in doubt whether I like all Gen 3 classes being in a completely separate tree as opposed to existing alongside the old classes. 

@jlaur before I rename any classes, and/or move them to other folders, I would like to make 100% sure we are in agreement on the approach. So can you please confirm the following from your side?
1. Can you confirm that in general you do not want to have the new classes in own gen3 specific folders?
2. Can you confirm that in general you do not want to have the new classes named by any gen 3 specific suffix (e.g. '3' suffix)?
3. Notwithstanding the above, apropos your point about DTO classes with '3' suffix; there are I think are two class name conflicts -- namely the `Scene` class embedded in the `Scenes` class, and the `ShadePosition` class; so in these cases the '3' suffix is probably inevitable. Do you agree to that exception to point 2 above?
4. For the handlers, discovery, and webTargets classes, so far I used the nomenclature 'HDPowerViewBlahBlah' => HDPowerViewBlahBlah3', which I think is inelegant, so I propose to name the gen3 versions of those classes more plainly as e.g. GatewayBridgeHandler, GatewayWebTargets, GatewayDiscoveryParticipant, GatewayShadeDiscoveryService, and GatewayShade3Handler, => Is that OK with you?
5. For info: there is another DTO class 'ScheduledEvent3' which potentially has a name conflict. However I think that we should probably delete that class because, unlike in gen1/2 such ScheduledEvents (aka Automations) the respective resources are read only in gen3, so there are no channels created in the gateway bridge thing for such resources, because there is actually nothing that OH could do with / do to, such resources. => Do you agree with this deletion?

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/gen3/dto/SceneEvent3.java
andrewfg: I think we could certainly delete the 'isoDate' field as I can see no potential purpose for it in OH.

By contrast the 'evt' field might or might not be useful -- it depends on the testing -- this field contains a message like 'motion-started' or 'motion-completed', which we currently do not need, since for both such events we consume the 'current position' from the respective event. However testing might potentially show that 'evt' might potentially have some relevance in reality. So I suggest to leave it the _ad interim_.

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/gen3/dto/ShadeEvent3.java
andrewfg: Ditto, as above..

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/gen3/dto/ShadeEvent3.java
andrewfg: Ditto, as above for 'bleName' and 'isoDate'. But as mentioned above we currently only consume currentPosition, but potentially we might find that targetPositions might potentially be useful..

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/gen3/dto/ShadePosition3.java
andrewfg: Yes. I think so.

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/gen3/discovery/HDPowerViewDeviceDiscoveryService3.java
andrewfg: Done.

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/discovery/HDPowerViewHubDiscoveryParticipant.java
andrewfg: Done.

bundles/org.openhab.binding.hdpowerview/src/main/resources/OH-INF/thing/bridge.xml
andrewfg: Yes.

bundles/org.openhab.binding.hdpowerview/src/main/resources/OH-INF/thing/bridge.xml
andrewfg: Done.

bundles/org.openhab.binding.hdpowerview/src/main/resources/OH-INF/thing/bridge.xml
andrewfg: Done.

bundles/org.openhab.binding.hdpowerview/src/main/resources/OH-INF/thing/bridge.xml
andrewfg: Done.

bundles/org.openhab.binding.hdpowerview/src/main/resources/OH-INF/thing/bridge.xml
andrewfg: Done.

jlaur:
> 1. Can you confirm that in general you do not want to have the new classes in own gen3 specific folders?

In general I don't think a total segregation with two complete separate hierarchies will be useful. That would be almost like having two bindings in one, but actually I do think there will be some overlap and dependencies. The exception would be the DTO classes, which I think should be kept separate. Two reasons for that:
1. They represent different API's with different serialization/deserialization of different JSON structures, as previously discussed.
2. There are simply many of them, and they could also be structured in subfolders like we have for Gen 1/2: request, response, etc. where applicable.

> 2. Can you confirm that in general you do not want to have the new classes named by any gen 3 specific suffix (e.g. '3' suffix)?

In general, yes.

> 3. Notwithstanding the above, apropos your point about DTO classes with '3' suffix; there are I think are two class name conflicts -- namely the `Scene` class embedded in the `Scenes` class, and the `ShadePosition` class; so in these cases the '3' suffix is probably inevitable. Do you agree to that exception to point 2 above?

Not for this case. First, I don't think any Gen3 DTO class should include other DTO classes from Gen 1/2 and vice versa. This would break the separation. And also, in specific scenarios where we need to reference both, this shouldn't be a problem: We can just reference them by fully qualified namespaces. As an example, this might be needed in the console class being introduced with #13615.

> 4. For the handlers, discovery, and webTargets classes, so far I used the nomenclature 'HDPowerViewBlahBlah' => HDPowerViewBlahBlah3', which I think is inelegant, so I propose to name the gen3 versions of those classes more plainly as e.g. GatewayBridgeHandler, GatewayWebTargets, GatewayDiscoveryParticipant, GatewayShadeDiscoveryService, and GatewayShadeHandler, => Is that OK with you?

Yes, that seems reasonable. In a way we are lucky that Hub is now Gateway, as it resolves naming conflicts for us. :-)

> 5. For info: there is another DTO class 'ScheduledEvent3' which potentially has a name conflict. However I think that we should probably delete that class because, unlike in gen1/2 such ScheduledEvents (aka Automations) the respective resources are read only in gen3, so there are no channels created in the gateway bridge thing for such resources, because there is actually nothing that OH could do with / do to, such resources. => Do you agree with this deletion?

Yes. I suggested to Wes to add possibility in the API for enabling/disabling automations, but don't know the status. This is very useful for automating the automations. :-) See #11637. I would certainly miss this. Probably it would appear in some new form with new JSON structures, so we'll see.

andrewfg:
@jlaur many thanks for the advice; I have renamed and moved the classes as you suggest; and for the DTO's it was indeed possible to use the same classes in gen 3  as in gen 1/2 by means of fully qualified references as you proposed.

In this last commit, I also extended your new console class, and did some extra work on the ReadMe..

andrewfg:
@jlaur the merge conflicts from your PR #13630 are now hopefully resolved..

andrewfg:
> next step will be to confirm that Gen1/2 is not broken

The code has been running on my operative system fine for about a week now; obviously it is using the Gen 1/2 part of the code; but at least that confirms that nothing got broken..

@jlaur did you hear anything from Wes? Shall we send him a reminder, or do you think that would be impolite?

andrewfg:
[Powerview Generation 3 API.pdf](https://github.com/openhab/openhab-addons/files/10049952/Powerview.Generation.3.API.pdf)

andrewfg:
@jlaur I am wondering how we should proceed with this PR: As you know we had a conversation with the developer at Hunter Douglas who was offering some test hardware. But that has not materialised yet. So do you think we should remind him again? Or should we try to recruit some users on the OH forum to act as independent test partners? => WDYT?

jlaur:
> I am wondering how we should proceed with this PR: As you know we had a conversation with the developer at Hunter Douglas who was offering some test hardware. But that has not materialised yet. So do you think we should remind him again?

I haven't had much time lately, sorry for that. I'm probably not up-to-date, but from the last communication I seem to remember that the API was still not finalized, even if it was pushed as a firmware upgrade to the Gen 3 gateways?

I have not forgotten about this PR, but at this point I would probably prioritize your Hue PR since it has a **much** larger user base, i.e. all existing V2 bridge users. The HD Gen 3 system is still quite new, and I would suspect very few openHAB users having adopted it, so I think there is no rush. As long as we have zero known users (including ourselves), we cannot test anything.

Without any hardware to even test with (gateway + shade), my personal motivation for investing much time into this is quite low. With 11 Gen 2 shades in my house (out of 14 in total, the last 3 are quite old non-motorized Luxaflex), and since they cannot be upgraded, realistically I won't have Gen 3 shades for years without any donations. Before then we might even see Gen 4 with Thread/Matter support. 😉

vves:
@jlaur, @andrewfg Sorry to be slow getting back - I am working on the Gen3 Matter architecture & implementation plan. ;)  I have sourced a Pro Gateway and two shades that I'll get out to you.  See email.

andrewfg:
@jlaur just so you know: @vves has been in contact with me by email, and he is kindly arranging to send me Gen3 hub and motors for testing. I will keep you up to date on the outcome..

andrewfg:
> I have a new gen3 install that I'm willing to alpha test with

@jlk I asked HD for a 'freebie' hub from HD so I can test the Gen 3 code, but they seem to have difficulties in providing one; so perhaps I can indeed take you up on your kind offer to be an alpha tester? Are you still interested? And if so, can you let me know if you are on OH 3.4.x or on OH 4.x so I can provide you with the necessary binding JAR?

jlk:
Hey @andrewfg - I've moved on from OpenHab, but happy to spin up an instance and test out the integration for a week or two.

Actually - still had the container so just spun it back up: OpenHab 3.3.0 Build 2818.

andrewfg:
@jlaur just for info, many thanks to @vves we finally got some real hardware to test on; there were a few fixes necessary, but they were relatively minor, and it seems now to be working just fine :)

andrewfg:
image

image

image

bundles/org.openhab.binding.hdpowerview/README.md
```diff
@@ -226,17 +255,24 @@ For single shades the refresh takes the item's channel into consideration:
...
 
 ```java
+// generation 1/2
 Bridge hdpowerview:hub:home "Luxaflex Hub" @ "Living Room" [host="192.168.1.123"] {
     Thing shade s50150 "Living Room Shade" @ "Living Room" [id="50150"]
     Thing repeater r16384 "Bedroom Repeater" @ "Bedroom" [id="16384"]
 }
+
+// generation 3
+Bridge hdpowerview:gateway:home "Luxaflex Hub" @ "Living Room" [host="192.168.1.123"] {
```
jlaur, line 265:
```suggestion
Bridge hdpowerview:gateway:home "Luxaflex Gateway" @ "Living Room" [host="192.168.1.123"] {
```

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/GatewayWebTargets.java
```diff
@@ -0,0 +1,523 @@
...
+        /*
+         * Hunter Douglas keeps a statistical count of systems (e.g. openHAB, Home Assistant, Amazon etc.) that are
+         * using their Generation 3 REST API to connect to their gateways. So we are asked to register with the gateway
+         * on startup. => So do not change the 'openhab.org' tag below !!
+         */
+        register = home + "integration/openhab.org";
+    }
+
+    /**
+     * Issue a command a activate a scene.
```
jlaur, line 134:
```suggestion
     * Issue a command to activate a scene.
```

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/GatewayWebTargets.java
```diff
@@ -0,0 +1,523 @@
...
+    /**
+     * Issue a calibrate command to a shade.
+     *
+     * @param shadeId the shade to be calibrated.
+     * @throws HubProcessingException if any error occurs.
+     */
+    public void calibrateShade(int shadeId) throws HubProcessingException {
+        String json = jsonParser.toJson(new ShadeMotion(ShadeMotion.Type.CALIBRATE));
+        invoke(HttpMethod.PUT, String.format(shadeMotion, shadeId), null, json);
+    }
```
jlaur, line 152: According to the README, the `CALIBRATE` command is not supported for Gen 3?

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/GatewayWebTargets.java
```diff
@@ -0,0 +1,523 @@
...
+     *
+     * @return the list of scenes.
+     * @throws HubProcessingException if any error occurs.
+     */
+    public List<Scene> getScenes() throws HubProcessingException {
+        String json = invoke(HttpMethod.GET, scenes, null, null);
+        try {
+            return List.of(jsonParser.fromJson(json, Scene[].class));
+        } catch (NullPointerException e) {
+            throw new HubProcessingException("getScenes() missing response");
```
jlaur, line 205: You should not catch `NullPointerException`, but instead avoid it.

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/GatewayWebTargets.java
```diff
@@ -0,0 +1,523 @@
...
+     *
+     * @return the list of shades.
+     * @throws HubProcessingException if any error occurs.
+     */
+    public List<Shade> getShades() throws HubProcessingException {
+        String json = invoke(HttpMethod.GET, shades, null, null);
+        try {
+            return List.of(jsonParser.fromJson(json, Shade[].class));
+        } catch (NullPointerException e) {
+            throw new HubProcessingException("getScenes() missing response");
```
jlaur, line 242: You should not catch `NullPointerException`, but instead avoid it.

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/GatewayWebTargets.java
```diff
@@ -0,0 +1,523 @@
...
+        }
+    }
+
+    /**
+     * Close the SSE links.
+     */
+    private synchronized void sseClose() {
+        logger.debug("sseClose() called");
+        ScheduledFuture<?> task = sseQuietCheck;
+        if (task != null && !task.isCancelled()) {
```
jlaur, line 423: No need to check if cancelled.

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/GatewayWebTargets.java
```diff
@@ -0,0 +1,523 @@
...
+
+    /**
+     * Initialize the web targets
+     *
+     * @param httpClient the HTTP client (the binding)
+     * @param ipAddress the IP address of the server (the hub)
+     */
+    public GatewayWebTargets(GatewayBridgeHandler hubHandler, HttpClient httpClient, ClientBuilder clientBuilder,
+            SseEventSourceFactory eventSourceFactory, String ipAddress) {
+        this.hostName = ipAddress;
```
jlaur, line 106: For consistency, consider whether to use hostname or ipAddress.

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/GatewayWebTargets.java
```diff
@@ -0,0 +1,523 @@
...
+        if (!HTTP_OK_CODES.contains(statusCode)) {
+            logger.warn("invoke() HTTP status:{}, reason:{}", statusCode, response.getReason());
+            throw new HubProcessingException(String.format("HTTP %d error", statusCode));
+        }
+        String jsonResponse = response.getContentAsString();
+        if (logger.isTraceEnabled()) {
+            logger.trace("invoke() response JSON:{}", jsonResponse);
+        }
+        if ((HttpStatus.OK_200 == statusCode) && ((jsonResponse == null) || (jsonResponse.isEmpty()))) {
+            logger.warn("invoke() hub returned no content");
```
jlaur, line 296: Any reason to log warning in addition to throwing exception - will exception also be logged?

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/handler/GatewayBridgeHandler.java
```diff
@@ -0,0 +1,325 @@
...
+        /*
+         * Normally the thing's position state is updated by SSE. However we must do a refresh once on start up in order
+         * to get the initial state. Also the other properties (battery, signal strength etc.) are not updated by SSE.
+         * Furthermore we need to do periodic refreshes just in case the SSE connection may have been lost. So we
+         * schedule the refresh at the 'hardRefresh' interval.
+         */
+        ScheduledFuture<?> refreshTask = this.refreshTask;
+        if (refreshTask != null) {
+            refreshTask.cancel(false);
+        }
```
jlaur, line 219: This it not needed as already done by `dispose`. The framework will not call `initialize` two times without calling dispose in between.

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/handler/GatewayBridgeHandler.java
```diff
@@ -0,0 +1,325 @@
...
+        updateStatus(ThingStatus.UNKNOWN);
+    }
+
+    /**
+     * Method that is called when a scene changes state.
+     *
+     * @param scene the one that changed.
+     */
+    public void onSceneEvent(Scene scene) {
+        // TODO perhaps we should trigger an OH core event here ??
```
jlaur, line 230: Not sure what you mean here?

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/handler/ShadeThingHandler.java
```diff
@@ -0,0 +1,339 @@
...
+                        break;
+                    }
+                    throw new IllegalArgumentException(INVALID_COMMAND);
+
+                case CHANNEL_SHADE_COMMAND:
+                    if (command instanceof StringType) {
+                        if (COMMAND_IDENTIFY.equals(((StringType) command).toString())) {
+                            webTargets.jogShade(shadeId);
+                            break;
+                        } else if (COMMAND_CALIBRATE.equals(((StringType) command).toString())) {
```
jlaur, line 173: The README says:
- On Generation 1/2 hubs, the valid values for the `command` channel are: `CALIBRATE`, `IDENTIFY`.
On Generation 3 gateways, the only valid value is: `IDENTIFY`.

bundles/org.openhab.binding.hdpowerview/src/main/resources/OH-INF/thing/channels.xml
```diff
@@ -8,13 +8,15 @@
 		<item-type>Rollershutter</item-type>
 		<label>Position</label>
 		<description>The vertical position of the shade</description>
-		<category>blinds</category>
+		<category>Blinds</category>
+		<state min="0" max="100" pattern="%.1f %%"/>
```
jlaur, line 12: Update instructions are needed for the corresponding channel definition to be updated for managed things. See https://next.openhab.org/docs/developer/bindings/thing-xml.html#updating-thing-types

jlaur (changes requested):
Thanks for your contribution. The code looks good, I have added only a few comments.

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/console/HDPowerViewCommandExtension.java
```diff
@@ -87,13 +90,28 @@ public void execute(String[] args, Console console) {
...
+                    console.println("Error retrieving ID's: " + e.getMessage());
+                }
             }
         }
     }
 
     @Override
     public List<String> getUsages() {
-        return Arrays.asList(buildCommandUsage(SHOW_IDS, "list all shades and repeaters"));
+        return Arrays.asList(buildCommandUsage(SHOW_IDS, "list all shades and eventually repeaters"));
```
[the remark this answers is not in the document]
andrewfg: Removed

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/GatewayWebTargets.java
andrewfg: I think you need to address your comment to the author of the Java List class .. below is the JavaDoc for the `List.of()` method..
```
   /**
     * Returns an unmodifiable list containing an arbitrary number of elements.
     * See <a href="#unmodifiable">Unmodifiable Lists</a> for details.
     *
     * @apiNote
     * This method also accepts a single array as an argument. The element type of
     * the resulting list will be the component type of the array, and the size of
     * the list will be equal to the length of the array. To create a list with
     * a single element that is an array, do the following:
     *
     * <pre>{@code
     *     String[] array = ... ;
     *     List<String[]> list = List.<String[]>of(array);
     * }</pre>
     *
     * This will cause the {@link List#of(Object) List.of(E)} method
     * to be invoked instead.
     *
     * @param <E> the {@code List}'s element type
     * @param elements the elements to be contained in the list
     * @return a {@code List} containing the specified elements
     * @throws NullPointerException if an element is {@code null} or if the array is {@code null}
     *
     * @since 9
     */
    @SafeVarargs
    @SuppressWarnings("varargs")
    static <E> List<E> of(E... elements) {
```

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/GatewayWebTargets.java
andrewfg: Ditto.

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/GatewayWebTargets.java
andrewfg: No reason. Exceptions are logged in the handler.

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/GatewayWebTargets.java
jlaur: That documents exactly how you can avoid it, which is my point.

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/discovery/GatewayDiscoveryParticipant.java
```diff
@@ -0,0 +1,98 @@
...
+
+    @Override
+    public @Nullable DiscoveryResult createResult(ServiceInfo service) {
+        for (String host : service.getHostAddresses()) {
+            if (VALID_IP_V4_ADDRESS.matcher(host).matches()) {
+                ThingUID thingUID = new ThingUID(THING_TYPE_GATEWAY, host.replace('.', '_'));
+                DiscoveryResult hub = DiscoveryResultBuilder.create(thingUID)
+                        .withProperty(HDPowerViewHubConfiguration.HOST, host)
+                        .withRepresentationProperty(HDPowerViewHubConfiguration.HOST)
+                        .withLabel(i18nProvider.getText(FrameworkUtil.getBundle(getClass()), LABEL_KEY, LABEL_KEY,
```
jlaur, line 69: This could be simplified a bit - see #14577 - and this for the parameter part:
https://github.com/openhab/openhab-addons/blob/289aa0b017987fe7a30b6437ee094f490dc1c8f0/bundles/org.openhab.binding.danfossairunit/src/main/java/org/openhab/binding/danfossairunit/internal/handler/DanfossAirUnitHandler.java#L107-L108

So it would be something like should be possible:
```
.withLabel("@text/discovery.gateway.label [\"" + host + "\"]")
```

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/GatewayWebTargets.java
```diff
@@ -0,0 +1,450 @@
...
+     * Get the list of shades.
+     *
+     * @return the list of shades.
+     * @throws HubProcessingException if any error occurs.
+     */
+    public List<Shade> getShades() throws HubProcessingException {
+        String json = invoke(HttpMethod.GET, shades, null, null);
+        try {
+            Shade[] shades = jsonParser.fromJson(json, Shade[].class);
+            if (shades == null || Arrays.stream(shades).anyMatch(Objects::isNull)) {
```
jlaur, line 230: `fromJson` returns only null if provided with null or empty json. This can happen only for HTTP code 204 as `invoke` will otherwise throw in this case:
```java
        if ((HttpStatus.OK_200 == statusCode) && ((jsonResponse == null) || (jsonResponse.isEmpty()))) {
            throw new HubProcessingException("Missing response entity");
        }
```

But since you're throwing anyway here, I'm wondering why you want `invoke` to allow returning null or empty json for 204 rather than just throwing also in this case? It seems to just transfer this problem to the caller instead of dealing directly with it. Otherwise could be be simplified to:
```suggestion
            if (Arrays.stream(shades).anyMatch(Objects::isNull)) {
```

I have not checked if objects can be null, but it wouldn't make much sense, I guess at least empty objects should be returned.

jlaur: Or return `List.of()` here instead of throwing - if 204 is expected and should be interpreted as "no shades" rather than an error.

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/discovery/GatewayDiscoveryParticipant.java
andrewfg: > .withLabel("@text/discovery.gateway.label [\"" + host + "\"]")

@jlaur where is it documented that OH would properly parse this?

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/GatewayWebTargets.java
andrewfg: > can happen only for HTTP code 204

Ignore this special case. The gateway only returns 204 for the integration registration..

image

andrewfg: ^
I will revert the change and let the NPE (which as you imagine) would never happen, simply escalate (if it would ever so do)..

jlaur: Wouldn't the registration then go wrong? Perhaps 204 should only be allowed for `PUT`, but otherwise throw?

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/discovery/GatewayDiscoveryParticipant.java
jlaur: I don't know. But I just tested with these changes to be sure:

image

Result:

image

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/GatewayWebTargets.java
jlaur: Just to be clear: I think in addition to reverting the checks, this would make the error handling safer (at the end of `invoke`):
```java
        if (method == HttpMethod.GET && (jsonResponse == null || jsonResponse.isEmpty())) {
            throw new HubProcessingException("Missing response entity");
        }
```

andrewfg: Ok, although I will remove the 'jsonResponse == null' case which cannot occur because 'response.getContentAsString()' cannot return 'null'.

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/discovery/GatewayDiscoveryParticipant.java
andrewfg: Ok. Done

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/GatewayWebTargets.java
jlaur: It's still there, so redundant logging, and even logged as warning here?

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/GatewayWebTargets.java
```diff
@@ -0,0 +1,440 @@
...
+            response = request.send();
+        } catch (InterruptedException e) {
+            Thread.currentThread().interrupt();
+            throw new HubProcessingException(String.format("%s: \"%s\"", e.getClass().getName(), e.getMessage()));
+        } catch (TimeoutException | ExecutionException e) {
+            throw new HubProcessingException(String.format("%s: \"%s\"", e.getClass().getName(), e.getMessage()));
+        }
+        int statusCode = response.getStatus();
+        if (!HTTP_OK_CODES.contains(statusCode)) {
+            logger.warn("invoke() HTTP status:{}, reason:{}", statusCode, response.getReason());
```
jlaur, line 269: Redundant logging:
```suggestion
```

bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/GatewayWebTargets.java
jlaur: Sorry, you removed it, but I saw it also in another line. I have removed it there also.

jlaur (approved):
LGTM

PR #13359, opened 2022-09-06
[dsmr] Made additional key for smarty meter an option

Hilbrand:
This adds support for smart meters that require 2 keys.


jlaur (approved):
Thanks!

jlaur:
@Hilbrand - can you update the README to include this new parameter, before merging?

Hilbrand:
@jlaur done. Added to readme

@lsiepel Originally, support for the Luxembourg smart meter was added first. They use the name `smarty`, and this is also in the name of the bridge. Turns out meter in Austria use the same authentication as the Luxembourg smart meters. Hence the word `smarty` is used instead of `smart` as this change was added to the `smartyBridge`.

PR #13360, opened 2022-09-07
[Konnected] Taking configured onValue into account when sending status to actuator

haavar:
Fixing a bug where the configured onValue for the channel was ignored when sending a status to the panel. The onValue configuration was only considered when receiving a status from the panel. Honoring the onValue configuration is important when using the standard relay boards recommended by Konnected that are low triggered.

### DIFF
--- a/bundles/org.openhab.binding.konnected/src/main/java/org/openhab/binding/konnected/internal/handler/KonnectedHandler.java
+++ b/bundles/org.openhab.binding.konnected/src/main/java/org/openhab/binding/konnected/internal/handler/KonnectedHandler.java
@@ -92,10 +92,7 @@ public class KonnectedHandler extends BaseThingHandler {
             if (channelType.contains(CHANNEL_SWITCH)) {
                 logger.debug("A command was sent to a sensor type so we are ignoring the command");
             } else {
-                int sendCommand = (OnOffType.OFF.compareTo((OnOffType) command));
-                logger.debug("The command being sent to zone {} for channel:{}  is {}", zone, channelUID.getAsString(),
-                        sendCommand);
-                sendActuatorCommand(sendCommand, zone, channelUID);
+                sendActuatorCommand((OnOffType) command, zone, channelUID);
             }
         } else if (command instanceof RefreshType) {
             // check to see if handler has been initialized before attempting to get state of pin, else wait one minute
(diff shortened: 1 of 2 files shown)


bundles/org.openhab.binding.konnected/src/main/java/org/openhab/binding/konnected/internal/handler/KonnectedHandler.java
```diff
@@ -404,34 +401,28 @@ private String updateKonnectedModule() throws KonnectedHttpRetryExceeded {
...
-
                 payload.setZone(thingID, zone);
 
-                // check to see if this is an On Command type, if so add the momentary, pause, times to the payload if
-                // they exist on the configuration.
-                if (scommand == zoneConfig.onValue) {
+                if (command == OnOffType.ON) {
+                    payload.setState(zoneConfig.onValue);
                     payload.setTimes(zoneConfig.times);
-                    logger.debug("The times configuration was set to: {} for channelID: {}.", zoneConfig.times,
```
haavar, line 426: The serialized payload is logged below, and provides more accurate information than the individual debug lines.

jlaur:
> can you review this PR when you get a chance?

Sure.

> I have a few other changes queued, and I want to avoid large PRs and merge conflicts.

That's a good approach! And it will usually also result in faster reviews.

> Let me know if I can clarify anything.

Perhaps you could mention this **onValue** in the README as currently it seems undocumented? That would help me also understanding the logic behind it.

haavar:
Thanks. I did not mean to nag, I just saw you approve my other (newer) PR, and wanted to see if there was any issues with this.

Added a paragraph about the onValue in the readme. The name for this parameter may not be the best. It's really for controlling whether whether the on value in openhab corresponds to high or low signal on the panel.

I used this setting when connecting a low level triggered relay board. That means that the relay turns on when the signal is set to 0V (ground / low). Since the onValue setting was not taken into account, I had to remember that for that zone ON was turning the relay OFF and vise versa. It got pretty confusing when writing rules using that zone. After fixing the onValue settings, the rules became much easier to understand.

bundles/org.openhab.binding.konnected/src/main/java/org/openhab/binding/konnected/internal/handler/KonnectedHandler.java
```diff
@@ -92,10 +92,7 @@ public void handleCommand(ChannelUID channelUID, Command command) {
             if (channelType.contains(CHANNEL_SWITCH)) {
                 logger.debug("A command was sent to a sensor type so we are ignoring the command");
             } else {
-                int sendCommand = (OnOffType.OFF.compareTo((OnOffType) command));
-                logger.debug("The command being sent to zone {} for channel:{}  is {}", zone, channelUID.getAsString(),
```
volfan6415, line 96: any reason you are taking these logging statements out?  I get not wanting to clutter up the logs.  But i think maybe we keep them in but up the logger to logger.trace? 

Just so that they are still available for future debugging?

jlaur:
@haavar - I will let @volfan6415 as code owner respond to the open comment/review the PR before merging.

bundles/org.openhab.binding.konnected/src/main/java/org/openhab/binding/konnected/internal/handler/KonnectedHandler.java
volfan6415: ok that makes sense.

volfan6415 (approved):

jlaur (approved):
LGTM

PR #13362, opened 2022-09-07
[unifi] Fix portoverride to not remove any other data

Hilbrand:
When a user has configured additional settings on a PoE port, like name. These settings where lost when changing the PoEPort status in openHAB. This was because in the binding only some information of the override was stored and when writing the new state this information would have been send too. In this change the object to store the override has been replaced by a plain json object. Therefor we don't have to know what is in it and all information is kept.


jlaur:
Confirmation of test: https://community.openhab.org/t/unifi-binding-beta/131156/87

jlaur (approved):
LGTM - thanks!

PR #13364, opened 2022-09-08
[upnpcontrol] Prepare dynamic channels and status messages for translation

mherwege:

This PR externalises all dynamic channel names and labels and all thing status messages, so they can be translated.


bundles/org.openhab.binding.upnpcontrol/src/main/java/org/openhab/binding/upnpcontrol/internal/handler/UpnpServerHandler.java
```diff
@@ -165,8 +165,8 @@ public void dispose() {
     protected void initJob() {
         synchronized (jobLock) {
             if (!upnpIOService.isRegistered(this)) {
-                updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.COMMUNICATION_ERROR,
-                        "UPnP device with UDN " + getUDN() + " not yet registered");
+                String msg = String.format("@text/offline.device-not-registed [ \"%s\" ]", getUDN());
```
jlaur, line 168: Doesn't match actual key:
```suggestion
                String msg = String.format("@text/offline.device-not-registered [ \"%s\" ]", getUDN());
```

jlaur (changes requested):
Thanks! Spotted only one key which seems to mismatch.

bundles/org.openhab.binding.upnpcontrol/src/main/java/org/openhab/binding/upnpcontrol/internal/handler/UpnpServerHandler.java
mherwege: Many thanks and fixed. There was a second place with the mismatch for the same key.

jlaur (approved):
LGTM

PR #13376, opened 2022-09-11
[gpio] Set UP value on startup when inverted

### LINKED ISSUE #11296: [gpio] Invalid state of output channel after reboot
**My configuration:**

Openhabian: Raspbian GNU/Linux 10 (buster)
Raspberry Pi 3 Model B Rev 1.2
Openhab: 3.1.0

Thing gpio:pigpio-remote:local "GPIO localhost" [host="localhost", port=8888] {
    Channels:
        Type pigpio-digital-output : output-14 [ gpioId=14, invert=true] 
}

Switch RelaySwitch {channel="gpio:pigpio-remote:local:output-14" }

**Problem 1:**
After reboot is state of GPIO14 set to LOW and that means relay is switched ON. I would like to have it switched OFF.

GPIO 14: level=0 fsel=1 func=OUTPUT

**Proposed solution:** Add possibility to define initial value.

**Problem 2:**
My RelaySwitch item reposts invalid OFF state. GPIO14 is set to LOW and this state should be inverted which means it should be set to ON.

**Proposed solution:** Read GPIO state once GPIO thing is initialized.

**Temporary solution:** 
Initializing GPIO pins via /boot/config.txt - https://www.raspberrypi.org/documentation/computers/config_txt.html

gpio=14=op,dh

msteiger:
# Description

By default, the GPIO output pins are OFF. Using the "invert" flag, the behavior is changed so that the default state is ON. This works fine unless the service is restarted. On startup, the GPIO values are not set to ON despite the inverted flag being set.

This PR fixes that problem by setting the value in the constructor.

Maybe @nils-bauer or @SloCompTech could review and/or test a bit?

This might also be a fix for issue #11296.

### DIFF
--- a/bundles/org.openhab.binding.gpio/src/main/java/org/openhab/binding/gpio/internal/handler/PigpioDigitalOutputHandler.java
+++ b/bundles/org.openhab.binding.gpio/src/main/java/org/openhab/binding/gpio/internal/handler/PigpioDigitalOutputHandler.java
@@ -63,6 +63,10 @@ public class PigpioDigitalOutputHandler implements ChannelHandler {
             throw new NoGpioIdException();
         }
         this.gpio = new GPIO(jPigpio, gpioId, JPigpio.PI_OUTPUT);
+        if (configuration.invert) {
+            updateStatus.accept(OnOffType.ON);
+            gpio.setValue(true);
+        }
     }
 
     @Override


bundles/org.openhab.binding.gpio/src/main/java/org/openhab/binding/gpio/internal/handler/PigpioDigitalOutputHandler.java
```diff
@@ -63,6 +63,10 @@ public PigpioDigitalOutputHandler(GPIOOutputConfiguration configuration, JPigpio
             throw new NoGpioIdException();
         }
         this.gpio = new GPIO(jPigpio, gpioId, JPigpio.PI_OUTPUT);
+        if (configuration.invert) {
+            updateStatus.accept(OnOffType.ON);
```
SloCompTech, line 67: Should this line be added ? Because inverted means that pin should be by default in state 1, but this means OFF not ON, because it is inverted .

msteiger: Good point, I'm not sure on this on either. Maybe not setting `updateStatus` would be the correct thing to do. What do you think @nils-bauer ?

msteiger:
Could someone review this, please?

lolodomo:
> Could someone review this, please?

As I can see, this was already reviewed and there is still an open point. We need the feedback form the binding maintainer @nils-bauer . I have myself not the response to the opened question.

nils-bauer:
Hi all together,
sorry for the late response, I have a lot on my plate right now (university and work...).
@jeremyrumpf is also working on an update for this addon too (see [forum](https://community.openhab.org/t/gpio-new-gpio-binding-for-oh3/113943/81), I haven't tested it yet but it could be  relevant here.
In order to give you a meaningful review, I need to look into it a bit deeper and test it in my lab setup. Sadly, I won't have the time to do this before the end of October.

msteiger:
Any chance you could review this soon?

nils-bauer:
Hopefully end of October/start of November.

jeremyrumpf:
Hi Nils,

Over the last few weeks I have been working on the documentation and
cleanup of the work I did. I really want to get it submitted as it fixes,
in my opinion, the most glaring issue with the gpio binding. Hopefully I
can get you a pull request before you sit down and test it.

Regards,
Jeremy

msteiger:
Hey guys, any updates on this one? It would be great to have this issue fixed.

Maybe even better to fully replace the implementation with that of @jeremyrumpf but that seems to be a bigger project.

jeremyrumpf:
The pull request has been submitted, awaiting review:

https://github.com/openhab/openhab-addons/pull/13643

You can run the updated plugin if you are willing to manually install it.
If you need assistance, let me know. A build is at:

https://github.com/jeremyrumpf/jeremyrumpf/tree/main/files

The updated documentation is here:

https://github.com/jeremyrumpf/openhab-addons/tree/main/bundles/org.openhab.binding.gpio

msteiger:
Looks good to me! There seems to be a merge conflict which needs to be resolved before it can be integrated.

jeremyrumpf:
I will say learning the continuous integration and workflows has been pretty fun. I believe i have the conflicts resolved against 4.0.0-SNAPSHOT.

bundles/org.openhab.binding.gpio/src/main/java/org/openhab/binding/gpio/internal/handler/PigpioDigitalOutputHandler.java
lsiepel: AS 6 months have passsed, is this properly addressed ?

msteiger: I was hoping that this entire PR could be replaced with #13643 and still keep my fingers crossed that it will be merged over the next days.

PR #13382, opened 2022-09-12
[jdbc] Fix date cast exception

jlaur:
After upgrading mysql-connector to 8.0.30 this exception was thrown: `class java.time.LocalDateTime cannot be cast to class java.sql.Timestamp`

The problem was introduced with 8.0.23 - more details here:
https://dev.mysql.com/blog-archive/support-for-date-time-types-in-connector-j-8-0/

Regression of #13242


kaikreuzer (approved):
lgtm, thanks for the quick fix!

PR #13391, opened 2022-09-14
[vosk] Upgrade sdk and handle UnsatisfiedLinkError exceptions

GiviMAD:
Update Vosk sdk version.
Also handle UnsatisfiedLinkError exceptions, I was facing those on recent versions of the docker based Debian container as libatomic lib is not present any more and the exception was not handled properly.


lolodomo (approved):
LGTM

PR #13393, opened 2022-09-15
[miele] Migrate start channel to full DateTime channel and add end channel

### LINKED ISSUE #13316: [miele] Time inconsistencies and problems
I'm currently working on some rule logic for calculating the cheapest period for running my dishwasher. The goal is to make it finish at latest by the programmed end time, but with the possibility of triggering an earlier start based on electricity prices for each hour interval.

During this I found some inconsistencies in the time-based channels which I can work around, but which I think should be reviewed and could be candidates for improvements.

# The channels

In general the channels are of type `DateTime`, since there is no native type for containing a duration. First observations from this fact:
- Initialization is done by applying duration as epoch milliseconds to local time with time-zone "GMT+0". For example, if programmed start time is 6 hours from now, this would be stored as "1970-01-01T06:00+01:00" for CET, which equals 1970-01-01T05:00:00Z.
- Probably duration of more than 24 hours cannot be expressed correctly.

## start

- **Label:** Start Time
- **Description:** Programmed start time of the program

The actual value is the duration from now and until the program will start. This means that this channel will be updated every minute until the program starts. This doesn't seem consistent with the label or description.

In a rule, to get an `Instant` for when the program actually starts, some conversions/calculations are needed:

```javascript
val DateTimeType startTimeState = (Dishwasher_StartTime.state as DateTimeType)
val ZonedDateTime startTimeZoned = startTimeState.getZonedDateTime()
val programmedStart = now.plusSeconds(startTimeZoned.getHour() * 60 * 60 + startTimeZoned.getMinute() * 60).toInstant()
```

Because of time-zone issues, epoch seconds cannot be used directly as there will be an offset from local time-zone.

Proposal:
- Change to represent the actual start time, i.e. date and time. This would change the channel from being a countdown and instead contain full information about when the program will start. That of course also implies that the channel will no longer be updated each minute while in state **waiting to start**.
- Update state to **UNDEF** when no longer in state **programmed**/**waiting to start**/**running**.

## finish

- **Label:** Finish Time
- **Description:** Time to finish the program running on the appliance

The actual value is the duration from now and until the program will end. This means that this channel will be updated every minute until the program ends. This doesn't seem consistent with the label or description.

While the program is running, the could be interpreted as "Remaining Time", but until the program starts it seems somewhat strange to have this value count down.

Additionally there is a Miele bug when triggering start ahead of the programmed start time: Initially the original finish time is kept, and during the program this will catch up with the actual predicted finish time. Example - programmed to finish 07:00, but started ~01:00 using the **switch** channel:

```
2022-09-07 00:59:00.852 [INFO ] [openhab.event.ItemStateChangedEvent ] - Item 'Dishwasher_FinishTime' changed from 1970-01-01T06:02:00.000+0100 to 1970-01-01T06:01:00.000+0100
2022-09-07 01:00:01.537 [INFO ] [openhab.event.ItemStateChangedEvent ] - Item 'Dishwasher_FinishTime' changed from 1970-01-01T06:01:00.000+0100 to 1970-01-01T06:00:00.000+0100
2022-09-07 01:01:00.530 [INFO ] [openhab.event.ItemStateChangedEvent ] - Item 'Dishwasher_FinishTime' changed from 1970-01-01T06:00:00.000+0100 to 1970-01-01T05:59:00.000+0100
2022-09-07 01:01:01.234 [INFO ] [openhab.event.ItemStateChangedEvent ] - Item 'Dishwasher_FinishTime' changed from 1970-01-01T05:59:00.000+0100 to 1970-01-01T05:58:00.000+0100
2022-09-07 01:02:38.557 [INFO ] [openhab.event.ItemStateChangedEvent ] - Item 'Dishwasher_FinishTime' changed from 1970-01-01T05:58:00.000+0100 to 1970-01-01T05:56:00.000+0100
2022-09-07 01:02:47.626 [INFO ] [openhab.event.ItemStateChangedEvent ] - Item 'Dishwasher_FinishTime' changed from 1970-01-01T05:56:00.000+0100 to 1970-01-01T05:49:00.000+0100
2022-09-07 01:02:59.041 [INFO ] [openhab.event.ItemStateChangedEvent ] - Item 'Dishwasher_FinishTime' changed from 1970-01-01T05:49:00.000+0100 to 1970-01-01T05:48:00.000+0100
2022-09-07 01:03:00.857 [INFO ] [openhab.event.ItemStateChangedEvent ] - Item 'Dishwasher_FinishTime' changed from 1970-01-01T05:48:00.000+0100 to 1970-01-01T05:47:00.000+0100
...
2022-09-07 04:50:49.838 [INFO ] [openhab.event.ItemStateChangedEvent ] - Item 'Dishwasher_FinishTime' changed from 1970-01-01T00:03:00.000+0100 to 1970-01-01T00:01:00.000+0100
2022-09-07 04:53:14.648 [INFO ] [openhab.event.ItemStateChangedEvent ] - Item 'Dishwasher_FinishTime' changed from 1970-01-01T00:01:00.000+0100 to 1970-01-01T00:00:00.000+0100
```

Proposals:
- Change to represent the actual end/finish time, i.e. date and time. This way the value should only change when the appliance reevaluates end time based on the running program.
- Additionally a channel "remaining" could be introduced containing remaining duration of running program.
- Consider renaming to "End Time" for clarity.
- Update state to **UNDEF** when no longer in state **programmed**/**waiting to start**/**running**.

## duration

- **Label:** Duration
- **Description:** Duration of the program running on the appliance

The actual value is:
- Total duration of program until the program starts.
- Remaining duration when program is running (like **finish**), i.e. countdown.

Proposals: None, since actual total duration is probably not available anymore when the program is running. Perhaps it should change to UNDEF at this point to not overlap with remaining, if this is introduced.

## elapsed

- **Label:** Elapsed Time
- **Description:** Time elapsed in the program running on the appliance

Proposals: None.

# Implementation notes

Start Time at 17:39 (local time):

```json
                {
                    "Name": "startTime",
                    "Value": 561,
                    "Polling": 0,
                    "Metadata": {
                        "unit": "min",
                        "Filter": "true",
                        "description": "Property name for the selected start time of the appliance.",
                        "LocalizedID": "Start Time",
                        "LocalizedValue": "3:00",
                        "access": "RWE"
                    }
                },
```

Value is start time in minutes from now. However, this is relative to the appliance's (or gateway's) clock. Therefore, using it to calculate absolute time is susceptible to subtle inaccuracies and undesired updates. Example - programmed to start 06:00:

```
2022-09-13 20:08:08.023 [INFO ] [openhab.event.ItemStateChangedEvent ] - Item 'Dishwasher_StartTime' changed from 2022-09-14T06:00:00.000+0200 to 2022-09-14T06:01:00.000+0200
2022-09-13 20:09:15.585 [INFO ] [openhab.event.ItemStateChangedEvent ] - Item 'Dishwasher_StartTime' changed from 2022-09-14T06:01:00.000+0200 to 2022-09-14T06:02:00.000+0200
2022-09-13 20:09:52.959 [INFO ] [openhab.event.ItemStateChangedEvent ] - Item 'Dishwasher_StartTime' changed from 2022-09-14T06:02:00.000+0200 to 2022-09-14T06:00:00.000+0200
```

Corresponding three payload snips:

openHAB timestamp: 2022-09-13 20:08:07.750
```
{
	"Name": "startTime",
	"Value": 593,
	"Polling": 0,
	"Metadata": {
		"unit": "min",
		"Filter": "true",
		"description": "Property name for the selected start time of the appliance.",
		"LocalizedID": "Start Time",
		"LocalizedValue": "6:01",
		"access": "RWE"
	}
}
```

openHAB timestamp: 2022-09-13 20:09:15.508
```
{
	"Name": "startTime",
	"Value": 593,
	"Polling": 0,
	"Metadata": {
		"unit": "min",
		"Filter": "true",
		"description": "Property name for the selected start time of the appliance.",
		"LocalizedID": "Start Time",
		"LocalizedValue": "6:02",
		"access": "RWE"
	}
}
```

openHAB timestamp: 2022-09-13 20:09:52.944
Multicast packet: `type=7&id=xxxxxxxxxxxx&property=startTime&value=591`

As seen, it actually wouldn't help much to use **LocalizedValue** as Miele@home itself has problems calculating a stable value because of internal implementation problems.

In order to avoid frequent updates with these small differences, some algorithm would be needed.

jlaur:
Change **start** channel to contain date and time for start of program rather than countdown. This is a breaking change as the contents of the `DateTimeType` has been redefined. However, the channel was marked as advanced and carried the misleading description "Programmed start time of the program". Now, then channel will actually comply with this description.

Introduce new **end** channel for containing date and time for end of program. This could potentially replace the channel **finish** which has the same countdown behavior as the previous **start** channel. However, the channel has been left for compatibility reasons as it arguably could have been more useful than **start**, since it can be considered as "elapsed time" similarly to **duration**.

As **start** and **end** are no longer countdowns, this would in theory allow them to be updated only once when an appliance has been programmed, rather than once per minute. However, since the API provides only offsets from current time, this tends to be unreliable:
- The provided values are calculated from a different clock (either appliance or XGW 3000 clock) than the system running openHAB.
- Network delays translates directly into fluctuations. Especially Wi-Fi appliances seems to be very slow in their communication, I'm not sure if there is a dependency towards Miele Cloud for those.

To mitigate this problem a moving average algorithm has been implemented, which will smoothen out these short-term fluctuations. A real-life example where first timestamp is a raw calculation with `now` truncated to minutes. Second timestamp is corrected by moving average, and third is the result, truncated to minutes:
```
2022-09-19 11:23:07.446 Update state of miele:dishwasher:home:dishwasher:end from 1177 -> '2022-09-20T05:00:00Z' -> '2022-09-20T07:00+02:00[Europe/Copenhagen]' to '2022-09-20T07:00+02:00[Europe/Copenhagen]'
2022-09-19 11:23:13.120 Update state of miele:dishwasher:home:dishwasher:end from 1177 -> '2022-09-20T05:00:00Z' -> '2022-09-20T07:00+02:00[Europe/Copenhagen]' to '2022-09-20T07:00+02:00[Europe/Copenhagen]'
2022-09-19 11:23:33.440 Update state of miele:dishwasher:home:dishwasher:end from 1177 -> '2022-09-20T05:00:00Z' -> '2022-09-20T07:00+02:00[Europe/Copenhagen]' to '2022-09-20T07:00+02:00[Europe/Copenhagen]'
2022-09-19 11:23:49.170 Update state of miele:dishwasher:home:dishwasher:end from 1177 -> '2022-09-20T05:00:00Z' -> '2022-09-20T07:00+02:00[Europe/Copenhagen]' to '2022-09-20T07:00+02:00[Europe/Copenhagen]'
2022-09-19 11:24:09.558 Update state of miele:dishwasher:home:dishwasher:end from 1177 -> '2022-09-20T05:01:00Z' -> '2022-09-20T07:00+02:00[Europe/Copenhagen]' to '2022-09-20T07:00+02:00[Europe/Copenhagen]'
2022-09-19 11:24:27.663 Update state of miele:dishwasher:home:dishwasher:end from 1177 -> '2022-09-20T05:01:00Z' -> '2022-09-20T07:00:13+02:00[Europe/Copenhagen]' to '2022-09-20T07:00+02:00[Europe/Copenhagen]'
2022-09-19 11:24:47.830 Update state of miele:dishwasher:home:dishwasher:end from 1177 -> '2022-09-20T05:01:00Z' -> '2022-09-20T07:00:22+02:00[Europe/Copenhagen]' to '2022-09-20T07:00+02:00[Europe/Copenhagen]'
2022-09-19 11:25:03.275 Update state of miele:dishwasher:home:dishwasher:end from 1177 -> '2022-09-20T05:02:00Z' -> '2022-09-20T07:00:27+02:00[Europe/Copenhagen]' to '2022-09-20T07:00+02:00[Europe/Copenhagen]'
2022-09-19 11:25:10.369 Update state of miele:dishwasher:home:dishwasher:end from 1175 -> '2022-09-20T05:00:00Z' -> '2022-09-20T07:00:32+02:00[Europe/Copenhagen]' to '2022-09-20T07:00+02:00[Europe/Copenhagen]'
2022-09-19 11:25:18.840 Update state of miele:dishwasher:home:dishwasher:end from 1175 -> '2022-09-20T05:00:00Z' -> '2022-09-20T07:00:30+02:00[Europe/Copenhagen]' to '2022-09-20T07:00+02:00[Europe/Copenhagen]'
2022-09-19 11:25:38.764 Update state of miele:dishwasher:home:dishwasher:end from 1175 -> '2022-09-20T05:00:00Z' -> '2022-09-20T07:00:26+02:00[Europe/Copenhagen]' to '2022-09-20T07:00+02:00[Europe/Copenhagen]'
2022-09-19 11:25:57.273 Update state of miele:dishwasher:home:dishwasher:end from 1175 -> '2022-09-20T05:00:00Z' -> '2022-09-20T07:00:23+02:00[Europe/Copenhagen]' to '2022-09-20T07:00+02:00[Europe/Copenhagen]'
2022-09-19 11:26:17.645 Update state of miele:dishwasher:home:dishwasher:end from 1175 -> '2022-09-20T05:01:00Z' -> '2022-09-20T07:00:21+02:00[Europe/Copenhagen]' to '2022-09-20T07:00+02:00[Europe/Copenhagen]'
2022-09-19 11:26:33.209 Update state of miele:dishwasher:home:dishwasher:end from 1175 -> '2022-09-20T05:01:00Z' -> '2022-09-20T07:00:24+02:00[Europe/Copenhagen]' to '2022-09-20T07:00+02:00[Europe/Copenhagen]'
2022-09-19 11:26:39.007 Update state of miele:dishwasher:home:dishwasher:end from 1174 -> '2022-09-20T05:00:00Z' -> '2022-09-20T07:00:24+02:00[Europe/Copenhagen]' to '2022-09-20T07:00+02:00[Europe/Copenhagen]'
```

Having these stable channels can be used for example to implement a rule calculating the energy price for each interval until the program is scheduled to end, and then trigger the program to start earlier. I have this implemented already and will provide it as an example in the forums for inspiration.

Fixes #13316


kgoderis:
I am not sure I understand why you move this code out of the ChannelSelectors. IMO it breaks the design of the binding

jlaur:
@kgoderis - two main reasons:
- **State and context:** I need `TimeZoneProvider` and `TimeStabilizer` for calculation of the channel values. I can of course pass all of this, but it seems I would need to do that for all `getState()` implementation in all channel selectors?
- **Code duplication:** Code is duplicated into each channel selector instead of being shared.

In past work I also struggled with these channel selectors and the 1:1 relationship between properties and channels. For extended properties there is one property for multiple channels, so this didn't fit into this design. Also, when introducing raw (numeric) channels in addition to text channels, there was a similar problem because numeric and text values are contained in the same property.

I'm very open to suggestions, I just didn't find a way yet to solve these problems within the boundaries of the original design, or a way of extending it.

J-N-K:
If I‘m not mistaken you change the meaning of the `finish` channel from a time difference to a point in time, correct? This is a breaking change and might cause big confusion for the user.

jlaur:
> If I‘m not mistaken you change the meaning of the `finish` channel from a time difference to a point in time, correct? This is a breaking change and might cause big confusion for the user.

Correct. This was discussed here: https://github.com/openhab/openhab-addons/issues/13316#issuecomment-1244409824. The channel probably wasn't of much use previously, since there is also a **duration** channel which is duration of the program until it's started, after which it will become the remaining time. The **finish** channel was remaining time from now and until the end of the program. Example:
- Now: 15:00
- Start: 3:00
- Duration: 1:00
- Finish: 4:00

After the changes:
- Now: 15:00
- Start: 18:00
- Duration: 1:00
- Finish: 19:00

I doubt there are many users who would be affected by this breaking change, so IMHO it worth to do this cleanup rather than preserve the channels as they were and introduce new ones.

jlaur:
@J-N-K - on afterthought I have decided to preserve the **finish** channel for backwards compatibility and instead introduce a new **end** channel. The **start** channel redefinition is still a breaking change though - I have updated the description with the following reasoning:

> Change **start** channel to contain date and time for start of program rather than countdown. This is a breaking change as the contents of the `DateTimeType` has been redefined. However, the channel was marked as advanced and carried the misleading description "Programmed start time of the program". Now, then channel will actually comply with this description.

I will create a PR towards openhab-distro to mention this in the upgrade notes.

bundles/org.openhab.binding.miele/src/main/java/org/openhab/binding/miele/internal/TimeStabilizer.java
```diff
@@ -0,0 +1,104 @@
...
+        if (cache.isEmpty()) {
+            cache.add(new Item(value, now, now));
+            return value;
+        }
+
+        @Nullable
+        Item first = cache.getFirst();
+        @Nullable
+        Item last = cache.getLast();
+        last.end = now;
```
lolodomo, line 71: Maybe you should check that `first` and `last` are not null before using them ?

bundles/org.openhab.binding.miele/src/main/java/org/openhab/binding/miele/internal/handler/DishwasherChannelSelector.java
```diff
@@ -73,20 +73,8 @@ public State getState(String s, @Nullable DeviceMetaData dmd, MieleTranslationPr
...
-            try {
-                date.setTime(Long.valueOf(s) * 60000);
-            } catch (Exception e) {
-                date.setTime(0);
-            }
-            return getState(dateFormatter.format(date));
-        }
-    },
+    START_TIME("", START_CHANNEL_ID, DateTimeType.class, false, false),
+    END_TIME("", END_CHANNEL_ID, DateTimeType.class, false, false),
```
lolodomo, line 77: Just to be sure, it is expected to have the first parameter as an empty string ?
And not START_TIME_PROPERTY_NAME  for the first one ?

lolodomo (changes requested):

bundles/org.openhab.binding.miele/src/main/java/org/openhab/binding/miele/internal/TimeStabilizer.java
jlaur: It is checked a few lines above by isEmpty.

bundles/org.openhab.binding.miele/src/main/java/org/openhab/binding/miele/internal/handler/DishwasherChannelSelector.java
jlaur: Yes, it i s overridden by the handler, so channel selector is not used.

bundles/org.openhab.binding.miele/src/main/java/org/openhab/binding/miele/internal/TimeStabilizer.java
lolodomo: That's right.

lolodomo (approved):
LGTM, thank you

lolodomo:
@jlaur : please prepare a comment for the release notes about the breaking changes.

jlaur:
@lolodomo - see openhab/openhab-distro#1416.

PR #13394, opened 2022-09-16
[radiothermostat] Ignore updates if thermostat data is invalid

mlobstein:
It could be that my thermostat is beginning to fail since it has started randomly returning -1 instead of the actual temperature. This change will cause a polling update with invalid data to be ignored so it will not update the channels. As a result, the temperature graphs will no longer look like this:

temp dropout

### DIFF
--- a/bundles/org.openhab.binding.radiothermostat/src/main/java/org/openhab/binding/radiothermostat/internal/handler/RadioThermostatHandler.java
+++ b/bundles/org.openhab.binding.radiothermostat/src/main/java/org/openhab/binding/radiothermostat/internal/handler/RadioThermostatHandler.java
@@ -372,14 +372,14 @@ public class RadioThermostatHandler extends BaseThingHandler implements RadioThe
                 case DEFAULT_RESOURCE:
                     rthermData.setThermostatData(gson.fromJson(evtVal, RadioThermostatTstatDTO.class));
                     // if thermostat returned -1 for temperature, skip this update
-                    if (rthermData.getThermostatData().getTemperature() > 0) {
+                    if (rthermData.getThermostatData().getTemperature() >= 0) {
                         updateAllChannels();
                     }
                     break;
                 case HUMIDITY_RESOURCE:
                     RadioThermostatHumidityDTO dto = gson.fromJson(evtVal, RadioThermostatHumidityDTO.class);
                     // if thermostat returned -1 for humidity, skip this update
-                    if (dto != null && dto.getHumidity() > 0) {
+                    if (dto != null && dto.getHumidity() >= 0) {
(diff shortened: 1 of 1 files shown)


bundles/org.openhab.binding.radiothermostat/src/main/java/org/openhab/binding/radiothermostat/internal/handler/RadioThermostatHandler.java
```diff
@@ -371,14 +371,18 @@ public void onNewMessageEvent(RadioThermostatEvent event) {
             switch (evtKey) {
                 case DEFAULT_RESOURCE:
                     rthermData.setThermostatData(gson.fromJson(evtVal, RadioThermostatTstatDTO.class));
-                    updateAllChannels();
+                    // if thermostat returned -1 for temperature, skip this update
+                    if (rthermData.getThermostatData().getTemperature() > 0) {
```
lolodomo, line 375: Maybe rather >= 0

bundles/org.openhab.binding.radiothermostat/src/main/java/org/openhab/binding/radiothermostat/internal/handler/RadioThermostatHandler.java
```diff
@@ -371,14 +371,18 @@ public void onNewMessageEvent(RadioThermostatEvent event) {
...
+                        updateAllChannels();
+                    }
                     break;
                 case HUMIDITY_RESOURCE:
                     RadioThermostatHumidityDTO dto = gson.fromJson(evtVal, RadioThermostatHumidityDTO.class);
-                    if (dto != null) {
+                    // if thermostat returned -1 for humidity, skip this update
+                    if (dto != null && dto.getHumidity() > 0) {
```
lolodomo, line 382: Maybe rather >= 0

lolodomo (changes requested):

bundles/org.openhab.binding.radiothermostat/src/main/java/org/openhab/binding/radiothermostat/internal/handler/RadioThermostatHandler.java
mlobstein: ok, fixed

bundles/org.openhab.binding.radiothermostat/src/main/java/org/openhab/binding/radiothermostat/internal/handler/RadioThermostatHandler.java
mlobstein: fixed

lolodomo (approved):
LGTM

PR #13397, opened 2022-09-16
[mqtt.homeassistant] support availability_templates

ccutrer:


antroids:
LGTM
Could you please check if there is possible to add some unit tests without overriding all these mocked handlers?
The Availability tests could be added to any component, i.e. SensorTests, maybe along with the additional helper method at AbstractComponentTests.

ccutrer:
Test added.

bundles/org.openhab.binding.mqtt.generic/src/main/java/org/openhab/binding/mqtt/generic/AvailabilityTracker.java
```diff
@@ -33,6 +34,21 @@ public interface AvailabilityTracker {
      */
     public void addAvailabilityTopic(String availability_topic, String payload_available, String payload_not_available);
 
+    /**
+     * Adds an availability topic to determine the availability of a device.
+     * <p>
+     * Availability topics are usually set by the device as LWT.
+     *
+     * @param availability_topic
```
lolodomo, line 42: If you add the javadoc, can you please do it fully and so set a description to each parameter.

lolodomo (changes requested):

bundles/org.openhab.binding.mqtt.generic/src/main/java/org/openhab/binding/mqtt/generic/AvailabilityTracker.java
ccutrer: Done

lolodomo (approved):
LGTM

lolodomo:
@antroids : could you please have a look?

bundles/org.openhab.binding.mqtt.homeassistant/src/main/java/org/openhab/binding/mqtt/homeassistant/internal/config/dto/Availability.java
```diff
@@ -25,6 +25,8 @@ public class Availability {
     protected String payloadAvailable = "online";
     @SerializedName("payload_not_available")
     protected String payloadNotAvailable = "offline";
+    @SerializedName("value_template")
+    protected String valueTemplate;
```
antroids, line 29: Should be Nullable according to documentation https://www.home-assistant.io/integrations/sensor.mqtt/#value_template

bundles/org.openhab.binding.mqtt.homeassistant/src/test/java/org/openhab/binding/mqtt/homeassistant/internal/component/SensorTests.java
```diff
@@ -77,6 +77,9 @@ public void test() throws InterruptedException {
                 .build().getPattern(), is("%s %unit%"));
 
         waitForAssert(() -> assertState(component, Sensor.SENSOR_CHANNEL_ID, UnDefType.UNDEF), 10000, 200);
+
+        publishMessage("zigbee2mqtt/bridge/state", "{ \"state\": \"offline\" }");
```
antroids, line 81: It seems that thing should be offline after receiving the message.
Probably the status tracker is async or replaced by mock in the abstract test.

antroids (changes requested):

bundles/org.openhab.binding.mqtt.homeassistant/src/main/java/org/openhab/binding/mqtt/homeassistant/internal/config/dto/Availability.java
ccutrer: ah, I was following payloadAvailable not being Nullable, but it makes sense that they're not, because they have a default value.

bundles/org.openhab.binding.mqtt.homeassistant/src/test/java/org/openhab/binding/mqtt/homeassistant/internal/component/SensorTests.java
ccutrer: 🤦  copy/paste error. good catch. and yeah, I had to wrap it in waitForAssert to get it to pass.

ccutrer: it ended up a little more complicated than that. Since the ThingHandlerCallback is a mock, it wasn't actually propagating the OFFLINE status to the Thing. Now fixed.

bundles/org.openhab.binding.mqtt.homeassistant/src/main/java/org/openhab/binding/mqtt/homeassistant/internal/config/dto/Availability.java
antroids: Could you please add Nullable to the getter too.
DTOs are excluded from static code analysis tool and they don't have NonNullByDefault, but it will be useful to know which fields can be null by API documentation.

bundles/org.openhab.binding.mqtt.homeassistant/src/test/java/org/openhab/binding/mqtt/homeassistant/internal/component/SensorTests.java
antroids: Great!
You could make an cosmetic improvement by replacing inline Answer class implementation by Java lambda: `doAnswer(invocation -> {...})...`

bundles/org.openhab.binding.mqtt.homeassistant/src/main/java/org/openhab/binding/mqtt/homeassistant/internal/config/dto/Availability.java
ccutrer: good catch

bundles/org.openhab.binding.mqtt.homeassistant/src/test/java/org/openhab/binding/mqtt/homeassistant/internal/component/SensorTests.java
ccutrer: thanks! java definitely isn't my primary language, so it's always good to get little tips like this to continue to improve in it

antroids (approved):
LGTM

PR #13398, opened 2022-09-16
[mqtt.homeassistant] sensors with a state_class are numeric

ccutrer:
see reference in code comment, but a measurement sensor is assumed to be numeric, even if it doesn't have a unit

### DIFF
--- a/bundles/org.openhab.binding.mqtt.homeassistant/src/main/java/org/openhab/binding/mqtt/homeassistant/internal/component/Sensor.java
+++ b/bundles/org.openhab.binding.mqtt.homeassistant/src/main/java/org/openhab/binding/mqtt/homeassistant/internal/component/Sensor.java
@@ -49,6 +49,8 @@ public class Sensor extends AbstractComponent<Sensor.ChannelConfiguration> {
         protected @Nullable String unitOfMeasurement;
         @SerializedName("device_class")
         protected @Nullable String deviceClass;
+        @SerializedName("state_class")
+        protected @Nullable String stateClass;
         @SerializedName("force_update")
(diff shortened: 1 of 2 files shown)


bundles/org.openhab.binding.mqtt.homeassistant/src/main/java/org/openhab/binding/mqtt/homeassistant/internal/component/Sensor.java
```diff
@@ -70,9 +72,14 @@ public Sensor(ComponentFactory.ComponentConfiguration componentConfiguration) {
 
         Value value;
         String uom = channelConfiguration.unitOfMeasurement;
+        String sc = channelConfiguration.stateClass;
 
         if (uom != null && !uom.isBlank()) {
             value = new NumberValue(null, null, null, UnitUtils.parseUnit(uom));
+        } else if (sc != null && !sc.isBlank()) {
```
lolodomo, line 79: No need to check for "None"?

ccutrer: No. `None` in the HASS docs isn't a literal string, but the Python `None`, which is equivalent to `null` in Java and JSON.

lolodomo: Ok

lolodomo (approved):
LGTM, thank you

PR #13406, opened 2022-09-18
[velux] Bridge status shows offline when communication fails

### LINKED ISSUE #13383: [velux] Bridge Thing remains ONLINE even though connection can't be established
## Expected Behavior
If the connection to the VELUX KLF can't be established, the thing should transition into the appropriate status.

## Current Behavior
I do get error messages about the KLF device refusing connections:

```
11:16:50.704 [INFO ] [ux.internal.bridge.slip.io.Connection] - io() on 192.168.178.204: Exception occurred during I/O: raised a non-recoverable error during connection setup: Connection refused (Connection refused).
```

But even though this is logged, the thing does not show anything is wrong:

image

(Please disregard the thing ID, it still has an old IP address and I was too lazy to fix it 😄) 

1.
2.

## Your Environment

* Version used: (e.g., openHAB and add-on versions)
* Operating System and version (desktop or mobile, Windows 10, Raspbian Buster, ...):
openHAB 3.2.0 Release
using stock addons, so also 3.2.0

andrewfg:
Resolves #13383

### Background

As reported in issue #13383 the bridge thing does not indicate offline status when when the communication fails.

### Solution

In this PR the bridge thing status will change to offline when the communication fails, and return to online when it is restored.

### DIFF
--- a/bundles/org.openhab.binding.velux/src/main/java/org/openhab/binding/velux/internal/handler/VeluxBridgeHandler.java
+++ b/bundles/org.openhab.binding.velux/src/main/java/org/openhab/binding/velux/internal/handler/VeluxBridgeHandler.java
@@ -291,7 +291,7 @@ public class VeluxBridgeHandler extends ExtendedBaseBridgeHandler implements Vel
          * When a binding call to the hub fails with a communication error, it will retry the call for a maximum of
          * veluxBridgeConfiguration.retries times, where the interval between retry attempts increases on each attempt
          * calculated as veluxBridgeConfiguration.refreshMSecs * 2^retry (i.e. 1, 2, 4, 8, 16, 32 etc.) so a complete
-         * retry series takes (veluxBridgeConfiguration.refreshMSecs * ((2^veluxBridgeConfiguration.retries + 1) - 1)
+         * retry series takes (veluxBridgeConfiguration.refreshMSecs * ((2^(veluxBridgeConfiguration.retries + 1)) - 1)
          * milliseconds. So we have to let this full retry series to have been tried (and failed), before we consider
          * the thing to be actually offline.
          */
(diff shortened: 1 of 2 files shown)


bundles/org.openhab.binding.velux/src/main/java/org/openhab/binding/velux/internal/handler/VeluxBridgeHandler.java
```diff
@@ -283,6 +287,9 @@ public void initialize() {
         logger.trace("initialize(): initialize bridge configuration parameters.");
         veluxBridgeConfiguration = new VeluxBinding(getConfigAs(VeluxBridgeConfiguration.class)).checked();
 
+        offlineDelay = Duration.ofMillis((((long) Math.pow(2, veluxBridgeConfiguration.retries) * 2) - 1)
```
lolodomo, line 290: Maybe you can find an even more complex calculation? lol
To be serious, maybe you could add a comment to explain this calculation?

lolodomo:
Your approach is relatively unusual. 
You don't want to adjust the thing status at each command?

bundles/org.openhab.binding.velux/src/main/java/org/openhab/binding/velux/internal/handler/VeluxBridgeHandler.java
andrewfg: >  find an even more complex calculation? lol

I have added a comment to the code.

andrewfg:
> Your approach is relatively unusual.
> You don't want to adjust the thing status at each command?

I agree with you that it is rather unusual (I did not write this part of the code). However you may have missed the peculiarity that the `handleCommandCommsJob()` method is called both from the core `handleCommand()` (for handling commands from the core), and also from the polling scheduler (with the `RefreshType` command argument to refresh the channels). This means that the status will indeed be checked on each command received from the core, but it will ALSO be checked with each poller refresh "command" from the scheduler. And it is really the latter that we are interested in for this case.

lolodomo (approved):
LGTM, thank you

PR #13409, opened 2022-09-18
ZoneMinder - Change alarm state to only when monitor is in actual alarm state

rossbuggins:
Currently seeing a bug in Zoneminder where all monitors are always in state 1 - PREALARM, instead on going to idle. Would like to discuss why the current alarm logic is OpenHab is currently as it is (ie Alarm whenever not IDLE or UNKNOWN), rather than just looking for state Alarm.

As it is currently, I'm seeing all ZoneAlarm cameras showing as always alarming in OpenHab. (Was fine on ZoneMinder 1.34.x, have just recently upgraded to v1.36.25). - See issue raised with ZM  https://github.com/ZoneMinder/zoneminder/issues/3595

### DIFF
--- a/bundles/org.openhab.binding.zoneminder/src/main/java/org/openhab/binding/zoneminder/internal/handler/Monitor.java
+++ b/bundles/org.openhab.binding.zoneminder/src/main/java/org/openhab/binding/zoneminder/internal/handler/Monitor.java
@@ -101,7 +101,7 @@ public class Monitor {
     }
 
     public void setState(MonitorState state) {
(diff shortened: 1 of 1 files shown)


rossbuggins:
@mhilbush

rossbuggins:
Just seen that ZM have fixed the bug in 1.36.26 with the renumbering of state changes. https://github.com/ZoneMinder/zoneminder/issues/3584
https://community.openhab.org/t/new-zoneminder-binding-for-zoneminder-versions-1-34-0/95112/416

PR #11049, opened 2021-07-22
[broadlink] Initial contribution

themillhousegroup:
A binding that can communicate with the Broadlink family of WiFi-addressable devices, currently comprising:

**A1** environmental sensor
**RMx** Remote IR blasters
**SPx** WiFi-controlled power switches
**MPx** WiFi-controlled multi-outlet power strips
Development of this binding was begun some four (!) years ago by one Cato Sognen, who unfortunately seems to have stopped development a few months after posting about it here:
https://community.openhab.org/t/broadlink-binding-for-rmx-a1-spx-and-mp-any-interest/22768

I reverse-engineered the binding's JAR file and gradually built up an understanding of its operation, fixing a few bugs and adding support for more devices along the way, including bringing it up through the 2.x and 3.x upgrade paths.

There has been significant interest and help provided by interested parties, as evidenced by the enormity of the above thread! It would be fantastic to get this binding into the official release to reward their patience :-)

Thanks also to @Skinah who provided invaluable guidance in previous PRs for this binding.
closes #322

Signed-off-by: John Marshall john@themillhousegroup.com

PR #13412, opened 2022-09-18
[broadlinkthermostat] Aesthetic rename and add RM Mini 

GiviMAD:

This is a proposal for renaming the broadlinkthermostat binding into broadlink also add a new device type "RM Mini" which is a universal controller infrared device.
Instead of adding a new binding I thought it was better to rename the current one, as the package relies in a library which supports other devices, as indicated in the readme.
Let me know in case is better to create a new binding.
To keep the autor aware @flo-02-mu.


GiviMAD:
CI error seems unrelated.

lolodomo:
> Let me know in case is better to create a new binding.

I have no definitive opinion on the two possible options.
I would be in favour of your proposal but of course this will be a breaking change for any user who previously installed this binding. At the same time, this is probably not a very used binding ?

@openhab/add-ons-maintainers : WDYT ?

fwolter:
I think "broadlink" would be a better name, too. We discussed this partially in #11049. But I don't see much benefit of renaming this now and breaking all installations. I'd suggest to rename it with the OH 4 release (whenever that will be) and add the new device in the existing binding.

GiviMAD:
@lolodomo, @fwolter thanks for the feedback. I will revert the package id while keeping other changes that will not break the binding.
I will put this on hold a couple of days while finishing another contribution.

bundles/org.openhab.binding.broadlinkthermostat/src/main/java/org/openhab/binding/broadlinkthermostat/internal/handler/RMUniversalRemoteHandler.java
```diff
@@ -0,0 +1,154 @@
...
+        } catch (Exception e) {
+            logger.warn("IOException while running channel {}", channelUID);
+        }
+    }
+
+    private void handleLearningCommand() throws IOException {
+        RM2Device rm2Device = this.rm2Device;
+        if (rm2Device != null) {
+            try {
+                rm2Device.enterLearning();
```
lolodomo, line 111: There is no way to exit properly from learning mode ?

bundles/org.openhab.binding.broadlinkthermostat/src/main/java/org/openhab/binding/broadlinkthermostat/internal/handler/RMUniversalRemoteHandler.java
```diff
@@ -0,0 +1,154 @@
...
+        }
+    }
+
+    @Override
+    public void handleCommand(ChannelUID channelUID, Command command) {
+        logger.debug("Command: {}", command.toFullString());
+        authenticate(false);
+
+        if (command == RefreshType.REFRESH) {
+            logger.warn("Nothing to refresh on thing {}", thing.getUID());
```
lolodomo, line 84: Rather a DEBUG level.

bundles/org.openhab.binding.broadlinkthermostat/src/main/java/org/openhab/binding/broadlinkthermostat/internal/handler/RMUniversalRemoteHandler.java
```diff
@@ -0,0 +1,154 @@
...
+            return;
+        }
+        String channelId = channelUID.getIdWithoutGroup();
+        try {
+            switch (channelId) {
+                case BroadlinkBindingConstants.LEARNING_MODE:
+                    handleLearningCommand();
```
lolodomo, line 91: It looks like only ON command is supported ?
In that case, you should call handleCommand only if command is ON.

lolodomo (changes requested):

lolodomo:
As mentioned in one of my review comments, you introduced a breaking change (by renaming a thing type) and I think it can be avoided, in a simple way.

bundles/org.openhab.binding.broadlinkthermostat/src/main/java/org/openhab/binding/broadlinkthermostat/internal/handler/RMUniversalRemoteHandler.java
```diff
@@ -0,0 +1,154 @@
...
+                case BroadlinkBindingConstants.SAVE_LEARNED:
+                    handleSaveLearned(command);
+                    break;
+                case BroadlinkBindingConstants.SEND_LEARNED:
+                    handleSendLearned(command);
+                    break;
+                default:
+                    logger.warn("Channel {} does not supported by thing {}", channelId, thing.getUID());
+            }
+        } catch (Exception e) {
```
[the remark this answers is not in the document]
GiviMAD: No, it's required due to the underling library throwing Exception direct instances.

bundles/org.openhab.binding.broadlinkthermostat/src/main/java/org/openhab/binding/broadlinkthermostat/internal/handler/RMUniversalRemoteHandler.java
```diff
@@ -0,0 +1,154 @@
...
+            } catch (IOException e) {
+                updateState(BroadlinkBindingConstants.LEARNING_MODE, OFF);
+                throw e;
+            }
+        } else {
+            logger.warn("Device not initialized");
+        }
+    }
+
+    private void handleSaveLearned(Command command) throws Exception {
```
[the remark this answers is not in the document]
GiviMAD: No, the method checkData from the library thows Exception, only option I see will be catching it inside the method and throw a different one.

bundles/org.openhab.binding.broadlinkthermostat/src/main/java/org/openhab/binding/broadlinkthermostat/internal/handler/RMUniversalRemoteHandler.java
GiviMAD: It's exited when check data method is called.
One option to implement it will be to call check data and ignore the result when and off command is send to startlearning.
Should I added it?

bundles/org.openhab.binding.broadlinkthermostat/src/main/java/org/openhab/binding/broadlinkthermostat/internal/BroadlinkHandlerFactory.java
```diff
@@ -0,0 +1,83 @@
...
+        ThingTypeUID thingTypeUID = thing.getThingTypeUID();
+
+        if (BroadlinkBindingConstants.FLOUREON_THERMOSTAT_THING_TYPE.equals(thingTypeUID)
+                || BroadlinkBindingConstants.HYSEN_THERMOSTAT_THING_TYPE.equals(thingTypeUID)) {
+            return new FloureonThermostatHandler(thing);
+        }
+        if (BroadlinkBindingConstants.RM_UNIVERSAL_REMOTE_THING_TYPE.equals(thingTypeUID)) {
+            return new RMUniversalRemoteHandler(thing);
+        }
+        return null;
```
lolodomo, line 81: What for the 2 unknown thing types ?
As they are supported, you must create an handler for them. If not, the things will remain uninitialized. Maybe it is what you expect ?

bundles/org.openhab.binding.broadlinkthermostat/src/main/java/org/openhab/binding/broadlinkthermostat/internal/handler/RMUniversalRemoteHandler.java
lolodomo: I let you decide.
At least now you added a check to handle only ON command, so it looks now better to me.

lolodomo:
There is now no more breaking changes that I can identify, that is a good point.

bundles/org.openhab.binding.broadlinkthermostat/src/main/java/org/openhab/binding/broadlinkthermostat/internal/BroadlinkHandlerFactory.java
lolodomo: But I see no real interest to create things if they are not handled by the binding.

bundles/org.openhab.binding.broadlinkthermostat/src/main/java/org/openhab/binding/broadlinkthermostat/internal/handler/RMUniversalRemoteHandler.java
GiviMAD: I'll keep it like this then, I don't think it will be used.

bundles/org.openhab.binding.broadlinkthermostat/src/main/java/org/openhab/binding/broadlinkthermostat/internal/BroadlinkHandlerFactory.java
GiviMAD: Yes, that is from the original implementation, it was also strange for me.
I think it just to show you there are other devices compatibles with the protocol but not supported, I think that by returning null thing remains uninitialized.
Not sure what to do, I can completely remove the unknown thing types.

lolodomo: Yes, I think it would be better to completely remove the unknown thing types.

lolodomo (approved):
LGTM, thank you

PR #13416, opened 2022-09-19
[tibber] Add channel for tomorrows prices and timestamps as JSON array

seime:
Allows users to build rules based on tomorrow's electricity cost.

Should be rewritten if https://github.com/openhab/openhab-core/pull/3000 is accepted in one or another form some time in the future. Until then, this PR helps.

### DIFF
--- a/bundles/org.openhab.binding.tibber/README.md
+++ b/bundles/org.openhab.binding.tibber/README.md
@@ -32,6 +32,7 @@ Tibber Default:
 | Hourly Consumption | Hourly Consumption (last/previous hour) | True      |
 | Hourly From        | Timestamp (hourly from)                 | True      |
 | Hourly To          | Timestamp (hourly to)                   | True      |
+| Tomorrow           | JSON array of tomorrow's prices         | True      |
 
 Tibber Pulse (optional):
 
@@ -139,4 +140,5 @@ Number:Power               TibberAPILivePowerProduction          "Live Power Pro
 Number:Power               TibberAPILiveMinPowerproduction       "Min Power Production [%.0f W]"                  {channel="tibber:tibberapi:7cfae492:live_minPowerproduction"}
 Number:Power               TibberAPILiveMaxPowerproduction       "Max Power Production [%.0f W]"                  {channel="tibber:tibberapi:7cfae492:live_maxPowerproduction"}
 Number:Energy              TibberAPILiveAccumulatedProduction    "Accumulated Production [%.2f kWh]"         {channel="tibber:tibberapi:7cfae492:live_accumulatedProduction"}
+String                     TibberAPITomorrowPrices               "Price per hour tomorrow JSON array"        {channel="tibber:tibberapi:7cfae492:tomorrow"}
 ```
--- a/bundles/org.openhab.binding.tibber/README.md
+++ b/bundles/org.openhab.binding.tibber/README.md
@@ -19,20 +19,20 @@ The channels (i.e. measurements) associated with the Binding:
 
 Tibber Default:
 
-| Channel ID         | Description                             | Read-only |
-|--------------------|-----------------------------------------|-----------|
-| Current Total      | Current Total Price (energy + tax)      | True      |
-| Starts At          | Current Price Timestamp                 | True      |
-| Current Level      | Current Price Level                     | True      |
-| Daily Cost         | Daily Cost (last/previous day)          | True      |
-| Daily Consumption  | Daily Consumption (last/previous day)   | True      |
-| Daily From         | Timestamp (daily from)                  | True      |
-| Daily To           | Timestamp (daily to)                    | True      |
-| Hourly Cost        | Hourly Cost (last/previous hour)        | True      |
-| Hourly Consumption | Hourly Consumption (last/previous hour) | True      |
-| Hourly From        | Timestamp (hourly from)                 | True      |
-| Hourly To          | Timestamp (hourly to)                   | True      |
-| Tomorrow           | JSON array of tomorrow's prices         | True      |
+| Channel ID         | Description                                             | Read-only |
+|--------------------|---------------------------------------------------------|-----------|
+| Current Total      | Current Total Price (energy + tax)                      | True      |
+| Starts At          | Current Price Timestamp                                 | True      |
+| Current Level      | Current Price Level                                     | True      |
+| Daily Cost         | Daily Cost (last/previous day)                          | True      |
+| Daily Consumption  | Daily Consumption (last/previous day)                   | True      |
+| Daily From         | Timestamp (daily from)                                  | True      |
+| Daily To           | Timestamp (daily to)                                    | True      |
+| Hourly Cost        | Hourly Cost (last/previous hour)                        | True      |
+| Hourly Consumption | Hourly Consumption (last/previous hour)                 | True      |
+| Hourly From        | Timestamp (hourly from)                                 | True      |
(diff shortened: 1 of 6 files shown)


kjoglum (approved):
Looks good to me.

bundles/org.openhab.binding.tibber/src/main/resources/OH-INF/thing/thing-types.xml
```diff
@@ -115,4 +116,10 @@
 		<description>Accumulated Production since Midnight</description>
 		<state pattern="%.3f %unit%"></state>
 	</channel-type>
+	<channel-type id="tomorrow">
+		<item-type>String</item-type>
+		<label>Prices for tomorrow as a JSON array</label>
+		<description>JSON array of tuples startsAt,total, ie {["startsAt": "2022-09-10T00:00:00+02:00", "total": 5.332},
```
jlaur, line 122:
```suggestion
		<description>JSON array of tuples startsAt,total, e.g. {["startsAt": "2022-09-10T00:00:00+02:00", "total": 5.332},
```

bundles/org.openhab.binding.tibber/src/main/resources/OH-INF/thing/thing-types.xml
```diff
@@ -115,4 +116,10 @@
 		<description>Accumulated Production since Midnight</description>
 		<state pattern="%.3f %unit%"></state>
 	</channel-type>
+	<channel-type id="tomorrow">
```
jlaur, line 119: Consider naming as quite generic, perhaps `tomorrow_prices`? Will it always be limited to tomorrow's prices, not just "future" prices, which could also include prices for today as well as possibly prices after midnight tomorrow?

bundles/org.openhab.binding.tibber/README.md
```diff
@@ -32,6 +32,7 @@ Tibber Default:
 | Hourly Consumption | Hourly Consumption (last/previous hour) | True      |
 | Hourly From        | Timestamp (hourly from)                 | True      |
 | Hourly To          | Timestamp (hourly to)                   | True      |
+| Tomorrow           | JSON array of tomorrow's prices         | True      |
```
jlaur, line 35: Consider if it's worth documenting in more details. Currently the channel type description contains an example which is not documented here.

jlaur (changes requested):
Thanks! I've added some very minor comments.

bundles/org.openhab.binding.tibber/src/main/resources/OH-INF/thing/thing-types.xml
seime: Updated

bundles/org.openhab.binding.tibber/README.md
seime: Added full example

bundles/org.openhab.binding.tibber/src/main/resources/OH-INF/thing/thing-types.xml
seime: Changed to `tomorrow_prices`. 

Always is tomorrows prices, no values provided (empty array) from midnight until around 13 o'clock when Nordpool releases tomorrows prices.

Do not Nordpool will change their policy any time soon.

jlaur: And the prices for remaining part of today are not relevant? I'm asking because I've just been doing something similar, and in my use-case current price and all future prices were relevant: https://community.openhab.org/t/dishwasher-price-calculation-automation/139207

seime: One could of course add a separate channel for todays prices and/or all known future prices combined as well, but I think that is out of scope for this PR :)

jlaur: Valid point. :) However, currently PR description says "Allows users to build rules based on **future** electricity cost." So it seems there is a gap, thus I'm wondering about the use-case for this, since some hours are simply missing:
- Now: 20:58
- 20:00-21:00 Missing
- 21:00-22:00 Missing
- 22:00-23:00 Missing
- 23:00-00:00 Missing
- 00:00-01:00 Available
- etc.

So if next PR would introduce future prices as opposed to the current subset (tomorrow's prices), probably the channel would need to be renamed, which would be a breaking change.

So I guess it's relevant to have this discussion now, but if the use-case for this PR doesn't care about future prices from now until midnight, it's fine as-is.

jlaur: If it was to be considered, probably the simplest would be one channel: **future_prices**, which would simply contain all of them, as they have both date and time.

seime: I changed the word used in the PR description.

Don't get me wrong, I _do_ see use for both `todays_prices` as well as `future_prices` where the latter would be a _combination_ of both.

I've been tinkering with the idea of creating a standalone automation bundle along the lines of what you and others (me including) do to calculate start/end time of appliances with rules. It could download prices from a stable source (or read from an existing price item - see https://github.com/openhab/openhab-core/pull/3000) and be based on parameters such as

- required runtime
- continous or split periods
- spacing in time since last run (avoid the appliance ie running from 22-00 and then again from 00-02)
- must start within X hours and/or after time of day
- must complete within X hours and/or time of day

.. and others. Could be configured like channels for the astro binding. 

But for now I also have some rules to do this, and this PR enables those rules. Nothing more :)

bundles/org.openhab.binding.tibber/src/main/resources/OH-INF/thing/thing-types.xml
```diff
@@ -115,4 +116,10 @@
 		<description>Accumulated Production since Midnight</description>
 		<state pattern="%.3f %unit%"></state>
 	</channel-type>
+	<channel-type id="tomorrow_prices">
```
jlaur, line 119: Consider if this channel should be marked as advanced, since it contains a JSON string that would need to be parsed by a rule:
```suggestion
	<channel-type id="tomorrow_prices" advanced="true">
```

bundles/org.openhab.binding.tibber/src/main/resources/OH-INF/thing/thing-types.xml
jlaur: I have had similar thoughts, but didn't yet find justification for an add-on, and also ended up just parsing JSON in rules. We could create an issue though and gather/discuss ideas that could eventually grow into some kind of addon.

Regarding future vs. tomorrow, let's keep it scoped as your prefer, since it fulfills your use-case, and probably also others.

bundles/org.openhab.binding.tibber/src/main/resources/OH-INF/thing/thing-types.xml
jlaur: @seime - this is the only remaining comment to be considered before merging, now wondering if you might have missed it?

seime: I most certainly have, will fix tonight :)

jlaur (approved):
LGTM

PR #13421, opened 2022-09-20
[openhabcloud] sleepSocketIO(long delay) halts entire io.socket thread #13210

lsafelix75:
this change is to address the following issue

https://github.com/openhab/openhab-addons/issues/13209

### DIFF
--- a/bundles/org.openhab.io.openhabcloud/src/main/java/org/openhab/io/openhabcloud/internal/CloudClient.java
+++ b/bundles/org.openhab.io.openhabcloud/src/main/java/org/openhab/io/openhabcloud/internal/CloudClient.java
@@ -141,7 +141,8 @@ public class CloudClient {
      * Delay reconnect scheduler pool
      * 
      */
-    protected final ScheduledExecutorService scheduler = ThreadPoolManager.getScheduledPool("DelayConnect");
+    protected final ScheduledExecutorService scheduler = ThreadPoolManager
+            .getScheduledPool(ThreadPoolManager.THREAD_POOL_NAME_COMMON);
 
     /**
      * Constructor of CloudClient
--- a/bundles/org.openhab.io.openhabcloud/src/main/java/org/openhab/io/openhabcloud/internal/CloudClient.java
+++ b/bundles/org.openhab.io.openhabcloud/src/main/java/org/openhab/io/openhabcloud/internal/CloudClient.java
@@ -24,6 +24,8 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ScheduledExecutorService;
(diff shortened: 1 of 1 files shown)


bundles/org.openhab.io.openhabcloud/src/main/java/org/openhab/io/openhabcloud/internal/CloudClient.java
```diff
@@ -135,6 +137,12 @@ public class CloudClient {
      */
     private final Backoff reconnectBackoff = new Backoff();
 
+    /*
+     * Delay reconnect scheduler pool
+     * 
+     */
+    protected final ScheduledExecutorService scheduler = ThreadPoolManager.getScheduledPool("DelayConnect");
```
lolodomo, line 144: Is there a need to create a new thread pool or can the common thread pool be used ?
How many threads do we expect to be started in parallel ? In case of several, I can understand the need of a dedicated thread pool.

lolodomo: @kaikreuzer : WDYT ?

lolodomo: Maybe @digitaldan ?

digitaldan: >Maybe @digitaldan ?

 I'm on a plane with limited access to the interwebs, so I'm not able to quickly run through the code as normal.  I wanted to check if calling connect ends up running the client on that thread until disconnected or if the socket client used its own thread.  It looks like `io.socket.client.Socket` does do something with its own thread management ?  So i am assuming calling connect here should return rather quickly.   If true, i think that part looks good here.

>How many threads do we expect to be started in parallel ?

Only one, and it should not happen that frequently, and if my assumption above is correct, it would be a short lived thread.  I agree a common thread pool would make sense if available.

lolodomo: So you could rather use `ThreadPoolManager.THREAD_POOL_NAME_COMMON` as thread pool name.

As this binding is probably used by almost all openHAB users, we have to be sure about any change.

lolodomo: @lsafelix75 : you are welcome to discuss here with us and explain your arguments.

lolodomo (approved):
LGTM, I hope this will not have any bad impact.

digitaldan:
I'm pretty sure this is a safe change and good solution. I'll run this on my home system and trigger a few disconnects and watch for any potential issues.

PR #13426, opened 2022-09-21
[mqtt.generic] fix range of RGB values

ccutrer:
Range is 0..255, not 0..250.

rgb -> hsv -> rgb still isn't perfect, but it's better. In particular, I found this when using HSBType.BLUE in a test, and it was coming out as 0,0,250 in RGB. It now comes out as a proper 0,0,255.

### DIFF
--- a/bundles/org.openhab.binding.mqtt.generic/src/main/java/org/openhab/binding/mqtt/generic/values/ColorValue.java
+++ b/bundles/org.openhab.binding.mqtt.generic/src/main/java/org/openhab/binding/mqtt/generic/values/ColorValue.java
@@ -46,6 +46,8 @@ import org.slf4j.LoggerFactory;
  */
 @NonNullByDefault
 public class ColorValue extends Value {
+    private static BigDecimal factor = new BigDecimal("2.55"); // string to not lose precision
+
     private final Logger logger = LoggerFactory.getLogger(ColorValue.class);
 
     private final ColorMode colorMode;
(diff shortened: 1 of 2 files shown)


bundles/org.openhab.binding.mqtt.generic/src/main/java/org/openhab/binding/mqtt/generic/values/ColorValue.java
```diff
@@ -120,7 +120,7 @@ public void update(Command command) throws IllegalArgumentException {
         }
     }
 
-    private static BigDecimal factor = new BigDecimal(2.5);
+    private static BigDecimal factor = new BigDecimal("2.55"); // string to not lose precision
```
lolodomo, line 123: Would `new BigDecimal(2.55)` not work ? I am not sure to understand the need for a  string.

Could you move up this constant at the beginning of the class please.

lolodomo (changes requested):

bundles/org.openhab.binding.mqtt.generic/src/main/java/org/openhab/binding/mqtt/generic/values/ColorValue.java
ccutrer: Because `new BigDecimal(2.55).toString()` yields `2.54999999999999982236431605997495353221893310546875`. In other words, `new BigDecimal(2.55).multiply(new BigDecimal(100)).intValue()` yields `254`, not `255`.

Yes, I'll move it to the top.

lolodomo (approved):

PR #13427, opened 2022-09-21
[touchwand] Fix thermostat target temperature command 

roieg:
### [bug] Fix thermostat target temperature command 
Command value was send to the hub as stringed QuantityType and not as a stringed integer

### DIFF
--- a/bundles/org.openhab.binding.touchwand/src/main/java/org/openhab/binding/touchwand/internal/TouchWandThermostatHandler.java
+++ b/bundles/org.openhab.binding.touchwand/src/main/java/org/openhab/binding/touchwand/internal/TouchWandThermostatHandler.java
@@ -62,9 +62,8 @@ public class TouchWandThermostatHandler extends TouchWandBaseUnitHandler {
             }
             if (command instanceof QuantityType) {
                 QuantityType<?> value = (QuantityType<?>) command;
-                int targetTemperature = value.intValue();
-                touchWandBridgeHandler.touchWandClient.cmdThermostatTargetTemperature(unitId,
-                        String.valueOf(targetTemperature));
+                String targetTemperature = String.valueOf(value.intValue());
+                touchWandBridgeHandler.touchWandClient.cmdThermostatTargetTemperature(unitId, targetTemperature);
                 return;
             }
 
--- a/bundles/org.openhab.binding.touchwand/src/main/java/org/openhab/binding/touchwand/internal/TouchWandThermostatHandler.java
+++ b/bundles/org.openhab.binding.touchwand/src/main/java/org/openhab/binding/touchwand/internal/TouchWandThermostatHandler.java
@@ -58,28 +58,35 @@ public class TouchWandThermostatHandler extends TouchWandBaseUnitHandler {
         if (touchWandBridgeHandler != null) {
             if (command instanceof OnOffType) {
                 touchWandBridgeHandler.touchWandClient.cmdThermostatOnOff(unitId, (OnOffType) command);
-            } else {
-                String sCommand = command.toString();
-                switch (sCommand) {
-                    case "cool":
-                    case "heat":
-                    case "fan":
-                    case "auto":
-                    case "dry":
-                        touchWandBridgeHandler.touchWandClient.cmdThermostatMode(unitId, sCommand);
-                        break;
-                    case "low":
-                    case "medium":
-                    case "high":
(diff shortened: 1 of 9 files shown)


bundles/org.openhab.binding.touchwand/src/main/java/org/openhab/binding/touchwand/internal/TouchWandThermostatHandler.java
```diff
@@ -58,28 +58,34 @@ void touchWandUnitHandleCommand(Command command) {
...
-                        break;
-                    default:
-                        touchWandBridgeHandler.touchWandClient.cmdThermostatTargetTemperature(unitId, sCommand);
-                        break;
-                }
+                return;
+            }
+            if (command instanceof QuantityType) {
+                QuantityType<?> value = (QuantityType<?>) command;
+                String targetTemperature = String.valueOf(value.intValue());
```
jlaur, line 65: It seems that you are simply stripping the unit. Which unit type is expected here, °C or °F? Perhaps you need to convert to target unit before applying the value, something like:

```
final QuantityType<?> value = ((QuantityType<?>) command).toUnit(SIUnits.CELSIUS);
```

bundles/org.openhab.binding.touchwand/src/main/java/org/openhab/binding/touchwand/internal/TouchWandThermostatHandler.java
```diff
@@ -58,28 +58,34 @@ void touchWandUnitHandleCommand(Command command) {
...
-                    case "fanAuto":
-                        touchWandBridgeHandler.touchWandClient.cmdThermostatFanLevel(unitId, "auto");
-                        break;
-                    default:
-                        touchWandBridgeHandler.touchWandClient.cmdThermostatTargetTemperature(unitId, sCommand);
-                        break;
-                }
+                return;
+            }
+            if (command instanceof QuantityType) {
```
jlaur, line 63: Consider if method should be extended to include the channel also. It seems that currently the channel is assumed from the type of command only. So let's say the someone would try to set **roomTemperature**, wouldn't this silently be accepted as a command for setting **targetTemperature**?

Additionally, for channel **targetTemperature** it would probably also make sense to support DecimalType/StringType using default unit?

jlaur (changes requested):
Thank you. I have added some comments regarding assumptions made.

bundles/org.openhab.binding.touchwand/src/main/java/org/openhab/binding/touchwand/internal/TouchWandThermostatHandler.java
roieg: good idea , I converted to C and then apply the value

bundles/org.openhab.binding.touchwand/src/main/java/org/openhab/binding/touchwand/internal/TouchWandThermostatHandler.java
roieg: it do make sense , I will think how I do it and implement across the Things handlers 
Since it is a bit more risky I will do it in a separate PR in the near future 
At the moment I want to make sure the current bug is resolved

roieg: I have checked temperatureItem.sendCommand("21") and temperatureItem.sendCommand(21) and both are working well with the current implementation

bundles/org.openhab.binding.touchwand/src/main/java/org/openhab/binding/touchwand/internal/TouchWandWebSockets.java
```diff
@@ -59,6 +59,7 @@ public class TouchWandWebSockets {
 
     private WebSocketClient client;
     private String controllerAddress;
+    private int port;
```
jlaur, line 62: Left-over?

roieg: yep , fixed it

jlaur (approved):
LGTM, thanks!

jlaur:
@roieg - can you update PR description and remove the port related stuff?

roieg:
> @roieg - can you update PR description and remove the port related stuff?

Done

PR #13432, opened 2022-09-23
[remoteopenhab] Optimize filtering of ALIVE events

lolodomo:

### DIFF
--- a/bundles/org.openhab.binding.remoteopenhab/src/main/java/org/openhab/binding/remoteopenhab/internal/rest/RemoteopenhabRestClient.java
+++ b/bundles/org.openhab.binding.remoteopenhab/src/main/java/org/openhab/binding/remoteopenhab/internal/rest/RemoteopenhabRestClient.java
@@ -401,9 +401,7 @@ public class RemoteopenhabRestClient {
             listeners.forEach(listener -> listener.onConnected());
         }
         if (!"message".equals(name)) {
-            if (!"alive".equals(name)) {
-                logger.debug("Received unhandled event with name '{}' and data '{}'", name, data);
-            }
+            // Ignore silently all events which are not "message" events. This includes the "alive" events.
             return;
         }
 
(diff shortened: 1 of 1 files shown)


bundles/org.openhab.binding.remoteopenhab/src/main/java/org/openhab/binding/remoteopenhab/internal/rest/RemoteopenhabRestClient.java
```diff
@@ -400,17 +400,15 @@ private void onEvent(InboundSseEvent inboundEvent) {
             connected = true;
             listeners.forEach(listener -> listener.onConnected());
         }
+        if (!"message".equals(name)) {
+            if (!"alive".equals(name)) {
```
digitaldan, line 404: Any reason you can't silently ignore anything other than "message"?  Seems like SSE types the client doesn't know about does not want to react too should just be silently discarded.

lolodomo: It was to detect potential new messages types not yet supported by the binding. 
But yes, I could rather ignore them silently.

lolodomo: Done

J-N-K:
I belege this does not depend on the core PR but should be merged before, correct?

lolodomo:
> I believe this does not depend on the core PR but should be merged before, correct?

As the current alive event has "event" as name, I believe you are right.

lolodomo:
So, to be clear, this PR is working with and without the core PR changing the alive event.

lolodomo:
Can this very small PR be merged please? 
@jlaur or @fwolter

jlaur (approved):
LGTM

PR #13436, opened 2022-09-24
[jellyfin] Add URL path support

SR-Lut3t1um:
### [Jellyfin] Added URL path support.
**Changes:**
**Fix:** Now using KTOR instead of building the url by simple string concad. This fixes the https and port 443 bug that existed prier to this change. (https://example.com would look like: https://example.com:443)
**Added:** Support for jellyfins Base URL option. This allows the addon to be used on installations were jellyfin is configured to use a base path.


GiviMAD:
Hi, thank you for the changes!
Can you add the new parameter also to the readme on the "Server Thing Configuration" table?
Also the DCO check is not passing, I think is required in order to get this merged.

GiviMAD:
Sorry, I closed this by error.

SR-Lut3t1um:
Alright should work now :)

GiviMAD (approved):
LGTM! Thanks for the changes.

lolodomo (approved):
LGTM

lolodomo:
@GiviMAD : why did you use ktor for this binding ? Wasn't it possible to use the recommended HTTP client for openHAB (Jetty) ?
Typically, I am almost sure that the new code for the method `getServerUrl` could also be done easily with Jetty stuff.

GiviMAD:
@lolodomo it's required by the jellyfin sdk.

PR #13437, opened 2022-09-24
[SMA Energy Meter] Add additional values for Home manager 2.0

### LINKED ISSUE #11497: [smaenergymeter] Weird PowerIn values with newer firmwares
Since around two weeks I have this kind of graph (displaying Energy in out)
grafik

In my DB  I see some weird values
grafik

Before it worked for years perfectly. All systems/IT were already rebooted but behavior keeps weird.
I am still on OH 2.5 because 3.x is not working for me.

lrep:
# [SMA Energy Meter] Add additional values for Home manager 2.0

# Description
The SMA Energy Meter binding only supported a limited amount of values available in the SMA Home manger 2.0. After the change it is also possible to show additional values is voltage and current in OH.

It fixes the following issues:
- Fixes #11497 
- Fixes problems with negative serial numbers of devices.

# Testing

The compiled bundle is available under: https://drive.google.com/file/d/1wuszmXzC9MvUh-oBeQl8F2Tin__q0zF9/view?usp=sharing


lolodomo:
Moving to more generic channel types is not the recommended way to do and by changing existing channel types, you break the compatibility (for users who defined their things in UI).

PR #13442, opened 2022-09-25
Added a second code owner

andrewfg:
Adding myself as a second code owner for several bindings.


lolodomo:
Why not just adding a line for velux binding?

jlaur:
> Why not just adding a line for velux binding?

@andrewfg - or change all four bindings in same PR. @lolodomo commented because it was changed for a different binding in context of a specific binding PR. But just changing code owners doesn't have to be in different PR's.

andrewfg:
> Why not just adding a line for gardena binding?

Sorry for any confusion, but I was trying to avoid and risk of a merge conflict between the various PR's.

> or change all four bindings in same PR

Ok. I will do that.

lolodomo (approved):
There were already several contributions by @andrewfg for these 3 bindings.

PR #13451, opened 2022-09-26
[jrubyscripting] clean up assignment of class objects

ccutrer:
pass the actual class objects through, and post-process in Ruby, to avoid the rescue nil -- and by extension, any issues with visibility into private objects from JRuby.


ccutrer:
cc @jimtng @boc-tothefuture

jimtng:
I've tested this change and it appears to be working.

boc-tothefuture (approved):
LGTM

jlaur (approved):
Thanks!

PR #13455, opened 2022-09-28
[unifi] Detect all PoE ports, and set PoE thing offline if no data could be found

Hilbrand:
*  Fix bug to detect PoE ports when first port is not PoE port. The binding assumed either all ports or no ports were PoE, and asssumed if port 0 was not PoE none was PoE. However, some switches have ports starting at port 5 to be PoE. Therefor changed code to just test each port if it is PoE.
* Set PoE thing offline if no data could be found . This would better reflect the PoE thing status if there is a problem with either the data from the api or a configuration problem (like invalid port number).

### DIFF
--- a/bundles/org.openhab.binding.unifi/src/main/java/org/openhab/binding/unifi/internal/handler/UniFiPoePortThingHandler.java
+++ b/bundles/org.openhab.binding.unifi/src/main/java/org/openhab/binding/unifi/internal/handler/UniFiPoePortThingHandler.java
@@ -137,9 +137,6 @@ public class UniFiPoePortThingHandler
 
     private State setOfflineOnNoPoEPortData() {
         if (getThing().getStatus() != ThingStatus.OFFLINE) {
-            logger.debug(
-                    "No PoE port information for thing '{}' could be found in the UniFi API data for port '{}'. Setting thing off line.",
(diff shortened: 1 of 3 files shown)


bundles/org.openhab.binding.unifi/src/main/java/org/openhab/binding/unifi/internal/handler/UniFiPoePortThingHandler.java
```diff
@@ -132,6 +135,17 @@ protected State getChannelState(final Map<Integer, UniFiPortTuple> ports, final
         return state;
     }
 
+    private State setOfflineOnNoPoEPortData() {
+        if (getThing().getStatus() != ThingStatus.OFFLINE) {
+            logger.debug(
+                    "No PoE port information for thing '{}' could be found in the UniFi API data for port '{}'. Setting thing off line.",
+                    getThing().getUID(), config.getPortNumber());
```
jlaur, line 142: This seems unneeded as framework will already log the status change. Instead, consider adding port number to the description if this is useful.

jlaur (changes requested):
Thanks! Only one minor comment.

jlaur (approved):
LGTM

PR #13466, opened 2022-09-29
[mqtt.homeassistant] ensure availability topics get subscribed

ccutrer:
HASS registers availability topics before calling start(), so the AbstractMQTTThingHandler was never subscribing/starting the availability channel(s). So do so in start() of the base class.

I checked other implementations, and either they already handle re-registering availabilityTopics in their start() (GenericMQTTThingHandler), or they don't use availabilityTopics at all from the base class and manage it themselves (Homie).

Note that this shows up as newly-added HASS things not having a problem (because the components aren't discovered until after the ThingHandler is started), but if you restart OpenHAB or disable/enable the thing, the channels (and components) are cached, thus how availabilityTopics are known before starting.


lolodomo (approved):
LGTM

lolodomo:
Waiting for a validation by @antroids

lolodomo:
A new milestone is in preparation, let's go for a merge of that PR even without @antroids feedback.

PR #12847, opened 2022-05-29
[openwebnet] add support for Basic Scenarios (WHO=0)

mvalla:
This PR adds support for Basic Scenarios (WHO=0) to the OpenWebNet binding.
With this enhancement it will be possible to detect scenario activation events to trigger rules in openHAB: this way openHAB becomes a very powerful scenario manager activated by physical BTicino scenario panels. 
This will solve #11378.

Issue #11378, opened 2021-10-13
[openwebnet] support for WHO=0 Scenarios


RubenFuser:
hello, i'm working with openhab 3 with openwebnet binding installed but with my surprise WHO=0 scenarios are not supported.
is it possible to add the support for this?

thanks a lot

mvalla:
are you interested in activating a scenario from OH ?  or being notified in OH when a scenario gets activated in your system?

RubenFuser:
The idea is that when rhe physical button is pressed,it calls a scenario/rule in openhab.

mvalla:
it would be useful then if you configure one of the physical buttons you want to use in the BTicino system, activate official OpenWebNet Client monitor mode and then copy here which messages are sent on the BUS

RubenFuser:
if i press the scenario #1 i have this: *0*1*95##
if i press the scenario #2 i have this: *0*2*95##
if i press the scenario #3 i have this: *0*3*95##
if i press the scenario #4 i have this: *0*4*95##

thanks

mvalla:
Can you use code fences? Just to be sure of the format of the message. See the < > button when editing the message

RubenFuser:
yes you're right, was wrong formatted:

`*0*2*95##`

this is now right.

mvalla:
@RubenFuser to check if configured scenarios panels can be discovered, can you try this command again with the BTicino client?
(then paste here the image of the output):
`*#0*0##`

RubenFuser:
![image](https://user-images.githubusercontent.com/92428083/170880650-0bc1c858-0da3-4d0b-9b6c-615979af5365.png)

this is output. i have set on one scenario button, to open all rollershutters, so you'll see open command for 5 rollershutters and stop command for rollershutters.

thanks

mvalla:
ok, so this test shows that Scenarios WHO=0 cannot be discovered automatically, at least not with the command: `*#0*0##`.
They must be configured manually or using Discovery by Activation as it's explained in the binding readme.

RubenFuser:
So just to recap, i can do auto discover by pressing the physical button. Do i need to update binding with a special release? Which version?

mvalla:
look at the messages/threads on the openHAB community forum

RubenFuser:
if i press the scenario physical button number 2, the command that i see is `0*2*95##`
i've put the binding in scanning mode, but if i press the physical button of the scenario 2, it isn't detected by the binding.
[...]

PR #13473, opened 2022-10-01
[openwebnet] add support for basic Scenarios (WHO=0)

mvalla:
This PR adds support for basic Scenarios (WHO=0) to the OpenWebNet binding.
With this enhancement it will be possible to detect scenario activation events on the BUS to trigger rules in openHAB: openHAB becomes a very powerful scenario manager activated by physical BTicino scenario modules and buttons.
This will solve https://github.com/openhab/openhab-addons/issues/11378.

(This PR continues from #12847 on a new branch)


lolodomo (changes requested):
Only few minor comments, mainly for i18n properties.
Remember you  can simply run "mvn i18n:generate-default-translations" to update it

lolodomo:
There might be a bug in the i18n generation tool as I have the feeling the options for a trigger channel type are not considered.
@wborn @cweitkamp for information

mvalla:
> There might be a bug in the i18n generation tool as I have the feeling the options for a trigger channel type are not considered.

@lolodomo what should I do with this trigger options default translation ? Indeed they are not automatically generated by the "generate-default-translations" command. 
Shall I add them manually? If yes, what is the correct syntax?

lolodomo:
I have the strange feeling internalization of trigger channels is not yet implemented in the core framework. 
So maybe, you can ignore my comments.

lolodomo:
I add in my TODO list to add internationalization of trigger options in the core framework. 
So forget my comments about trigger events.

wborn:
Doesn't look like any other binding has translations for event options. I don't know if they are used/translated for UIs. If so, I would expect properties for such translations to be similar to the state options, e.g.:

`channel-type.openwebnet.scenarioEvent.event.option.SCENARIO_01 = Scenario 1`

lolodomo:
@mvalla : you marked my comments as resolved without pushing any new code.

mvalla:
ok I added manullay default translations for event options . 
Do not know how they render on UI considering also that these are trigger events and not states. I guess if you do not correct the auto-gen script these will be overwritten if the script is called again, right?

I also made other requested changes. So for me this can be merged now.

lolodomo:
> ok I added manullay default translations for event options.

Sorry, I was probably not clear enough. This is something not yet implemented in the core framework. Including and forcing to translate them makes no sense now. So please keep them but comment them.

mvalla:
Ok, now I have commented the event options translations.
For me this PR is ready to be merged.

lolodomo (approved):
LGTM, thank you

PR #13480, opened 2022-10-02
[SunSpec] Add missing Inverter Status, Add Channel for Vendor-Specific Status

Blizzard26:
This PR does two things:
* It adds missing Inverter Status according to SunSpec Specification (see e.g., https://github.com/sunspec/models/blob/master/json/model_101.json#L307)
* It adds a Channel for Vendor Specific Status code. The technical part was already present (e.g., parsing of the vendor specific status code), but it was not exposed as channel so far.

Moreover, I fixed two old TODOs which where pointing to an already merge PR.


mrbig (approved):
Thank you for your pull request. It looks all good to me, and it is a nice thing you have also fixed some old todos.

lolodomo:
I don't find the declaration of the new channel in the thing definition.

Blizzard26:
@lolodomo Good catch. Not sure where that got lost (probably when I my other changes :( ). Will fix it right away.

lolodomo (approved):
LGTM, thank you

PR #13425, opened 2022-09-21
[jdbc] Fix date cast exception

jlaur:
After upgrading mysql-connector to 8.0.30 this exception was thrown twice during boot:

```
2022-09-21 15:23:29.985 [ERROR] [nal.common.AbstractInvocationHandler] - An error occurred while calling method 'QueryablePersistenceService.query()' on 'org.openhab.persistence.jdbc.internal.JdbcPersistenceService@b1af48': class java.time.LocalDateTime cannot be cast to class java.sql.Timestamp (java.time.LocalDateTime is in module java.base of loader 'bootstrap'; java.sql.Timestamp is in module java.sql of loader 'platform')
java.lang.ClassCastException: class java.time.LocalDateTime cannot be cast to class java.sql.Timestamp (java.time.LocalDateTime is in module java.base of loader 'bootstrap'; java.sql.Timestamp is in module java.sql of loader 'platform')
	at org.openhab.persistence.jdbc.db.JdbcBaseDAO.objectAsLong(JdbcBaseDAO.java:587) ~[?:?]
	at org.openhab.persistence.jdbc.db.JdbcBaseDAO.objectAsState(JdbcBaseDAO.java:553) ~[?:?]
	at org.openhab.persistence.jdbc.db.JdbcBaseDAO.lambda$0(JdbcBaseDAO.java:364) ~[?:?]
	at java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:195) ~[?:?]
	at java.util.ArrayList$ArrayListSpliterator.forEachRemaining(ArrayList.java:1655) ~[?:?]
	at java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:484) ~[?:?]
	at java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:474) ~[?:?]
	at java.util.stream.ReduceOps$ReduceOp.evaluateSequential(ReduceOps.java:913) ~[?:?]
	at java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234) ~[?:?]
	at java.util.stream.ReferencePipeline.collect(ReferencePipeline.java:578) ~[?:?]
	at org.openhab.persistence.jdbc.db.JdbcBaseDAO.doGetHistItemFilterQuery(JdbcBaseDAO.java:365) ~[?:?]
	at org.openhab.persistence.jdbc.internal.JdbcMapper.getHistItemFilterQuery(JdbcMapper.java:179) ~[?:?]
[...]

PR #13483, opened 2022-10-03
[jdbc] Consolidate and optimize datetime conversions

jlaur:
Minor refactoring:
- Optimize datetime conversions by avoiding multiple steps, for example `LocalDateTime` -> `Instant` -> epoch -> `Instant` -> `ZonedDateTime`. This will now be converted directly from `LocalDateTime` to `ZonedDateTime`.
- Eliminate redundant method after this.
- Introduce tests for all supported types.

Based on findings after #13382 and #13425.


lolodomo (approved):
LGTM, thank you

PR #13484, opened 2022-10-03
[homekit] persist all known accessories, to prevent loss of homekit information

ccutrer:
See the readme for more details, but basically this keeps track of every accessory we've ever created, and if it no longer exists, presents a dummy accessory instead. If the accessory comes back, nothing is lost in the Home app; if you meant to prune it permanently, you have to run a console command.


ccutrer:
I'd love to be able to expose "prune dummy accessories" in the UI as a button somehow, but I don't know how. Any suggestions?

yfre:
@ccutrer this is mega. this will solve the main headache and root cause of many dissatisfactions with the addon. 
thank you. 
i will take a look on the PR and also think about exposure to mainUI.

ccutrer:
So I've thought through this more, and I think this behavior should be configurable, and default to off. If you primarily use the UI to create items, the "oops I had a syntax error and everything disappeared for a few minutes" or "oops my script that generates items had a bug and everything disappeared" and various other flavors aren't really an issue for you. If you're _not_ using the UI, you should be familiar with console and able to run the command to prune when you need to.

ccutrer:
ok, I've updated it, also restoring the old startDelay logic if useDummyAccessories isn't enabled.

ccutrer:
Hmm... I'm still seeing all items on a bridge get reset sometimes :(. Will need to dig more.

yfre:
> Hmm... I'm still seeing all items on a bridge get reset sometimes :(. Will need to dig more.

the complete reset behavior of home app is still strange to me. why home app needs to delete accessories instead of showing them as "not available" or so.. or have a warning, explicit delete button or so

ccutrer:
Dependent on https://github.com/hap-java/HAP-Java/pull/167. @yfre: it'd be great if you tested this locally now. it seems to be performing a lot better at the moment for me, but having more "runtime" on it would be good.

ccutrer:
@yfre : this is ready for review. I've had one instance of everything disappearing in the past week, but I suspect that because of issues from multiple instances of the bundle running inside openhab at once. Regardless this is already far more stable than what's currently on main.

bundles/org.openhab.io.homekit/src/main/java/org/openhab/io/homekit/internal/HomekitImpl.java
```diff
@@ -183,6 +183,11 @@ private HomekitRoot startBridge(HomekitServer homekitServer, HomekitAuthInfoImpl
         bridges.add(bridge);
         bridge.setConfigurationIndex(changeListener.getConfigurationRevision());
         bridge.refreshAuthInfo();
+        if (settings.useDummyAccessories) {
```
yfre, line 186: just as suggestion, what about to add a condition to the "IF" below instead, e.g.

if (!settings.useDummyAccessorie && currentAccessoryCount < lastAccessoryCount)

it will add 2 more, potentially unnecessary calls for get currentAccessoryCount  and lastAccessoryCount but imo it makes code easier to understand. eitherwise i have 3 times bridge.start in that function.

bundles/org.openhab.io.homekit/src/main/java/org/openhab/io/homekit/internal/accessories/DummyHomekitAccessory.java
```diff
@@ -0,0 +1,153 @@
...
+
+import io.github.hapjava.accessories.HomekitAccessory;
+import io.github.hapjava.characteristics.Characteristic;
+import io.github.hapjava.services.Service;
+
+/**
+ * Implements a dummy placeholder accessory for when configuration is missing
+ *
+ * @author Cody Cutrer - Initial contribution
+ */
```
yfre, line 36: please annotate the class with 
@NonNullByDefault

and add Nullable where required
https://www.openhab.org/docs/developer/guidelines.html#null-annotations

bundles/org.openhab.io.homekit/src/main/java/org/openhab/io/homekit/internal/accessories/DummyHomekitAccessory.java
```diff
@@ -0,0 +1,153 @@
...
+public class DummyHomekitAccessory implements HomekitAccessory {
```
yfre, line 37: in general, im not sure whether this class belongs to Java-HAP or to openHAB.. the code is pretty low-level and is similar to Java-HAP stuff. but dummy accessory is OH specific.
im fine with both (as still unsure what is the best place)

bundles/org.openhab.io.homekit/src/main/java/org/openhab/io/homekit/internal/accessories/AbstractHomekitAccessoryImpl.java
```diff
@@ -349,4 +354,58 @@ protected BooleanItemReader createBooleanReader(HomekitCharacteristicType charac
...
+
+    private JsonObjectBuilder serviceToJson(Service service) {
+        var serviceBuilder = Json.createObjectBuilder();
+        serviceBuilder.add("type", service.getType());
+        var characteristics = Json.createArrayBuilder();
+
+        service.getCharacteristics().stream().sorted((l, r) -> l.getClass().getName().compareTo(r.getClass().getName()))
+                .forEach(c -> {
+                    try {
+                        var cJson = c.toJson(0).get();
```
yfre, line 378: toJson(0)
does it make iid in HAP json always 0? or do i miss something?

bundles/org.openhab.io.homekit/src/main/java/org/openhab/io/homekit/internal/HomekitChangeListener.java
```diff
@@ -407,4 +480,54 @@ private boolean itemIsForThisBridge(Item item, @Nullable Map<String, Object> con
...
+            }
+        }
+
+        toRemove.forEach(k -> knownAccessories.remove(k));
+        return !toRemove.isEmpty();
+    }
+
+    private void checkForDummyAccessories() {
+        var dummyAccessories = accessoryRegistry.getAllAccessories().values().stream()
+                .filter(a -> a instanceof DummyHomekitAccessory).toArray(HomekitAccessory[]::new);
```
yfre, line 514: would not be forEach better here and we print warning to each dummy accessory?
e.g.
```

accessoryRegistry.getAllAccessories().values().stream()
                .filter(a -> a instanceof DummyHomekitAccessory).forEach(dummyAccessory -> logger.warn(
                            "{} has been replaced with a dummy. See https://www.openhab.org/addons/integrations/homekit/#dummy-accessories for more information.",
                            dummyAccessory.getSerialNumber().get())
```

bundles/org.openhab.io.homekit/src/main/java/org/openhab/io/homekit/internal/HomekitChangeListener.java
```diff
@@ -221,25 +249,64 @@ private Optional<Item> getItemOptional(String name) {
...
+        checkForDummyAccessories();
+        logger.trace("Making new configuration revision {}", newRevision);
         storage.put(REVISION_CONFIG, "" + newRevision);
-        storage.put(ACCESSORY_COUNT, "" + lastAccessoryCount);
+        storage.put(KNOWN_ACCESSORIES, knownAccessories);
+    }
+
+    public synchronized void pruneDummyAccessories() {
+        boolean removed = false;
+        for (HomekitAccessory accessory : accessoryRegistry.getAllAccessories().values()
```
yfre, line 261: i would use a filter and for each here as well.
e.g.
`
accessoryRegistry.getAllAccessories().values().filter((a -> a instanceof DummyHomekitAccessory).forEach(accessory -> {
                try { ...
`

bundles/org.openhab.io.homekit/src/main/java/org/openhab/io/homekit/internal/HomekitChangeListener.java
```diff
@@ -130,35 +138,55 @@ public void updated(final Metadata oldMetadata, final Metadata newMetadata) {
...
+                changed = true;
+            }
+        }
+        // order of this conditional is important - checkMissingAccessories has side effects that need to always happen
+        if (checkMissingAccessories() || changed) {
+            makeNewConfigurationRevision();
+        } else {
+            logger.info("Created {} HomeKit items in instance {} (no change from prior configuration).",
+                    accessoryRegistry.getAllAccessories().size(), instance);
+            checkForDummyAccessories();
```
yfre, line 157: we need to execute check only if useDummy is true?
i..e
if (settings.useDummyAccessories) checkForDummyAccessories();

yfre (changes requested):

lolodomo:
What is the current status of that PR ?
@yfre : are all your comments now "resolved" ? Do you approve all the changes or not ?

ccutrer:
I found an issue that this PR causes. There's a PR in HAP-Java that's required first to fix that. @yfre just reviewed it a few days ago, and I haven't had time to address his comments yet. So waiting on me.

lsiepel:
> I'd love to be able to expose "prune dummy accessories" in the UI as a button somehow, but I don't know how. Any suggestions?

Maybe have a look at the zwave binding as that has buttons like ‘heal device’ I see some similarity in use case.  Only difference is you might want to have this button on binding level as removes items do not show up in the ui.

yfre:
> > I'd love to be able to expose "prune dummy accessories" in the UI as a button somehow, but I don't know how. Any suggestions?
> 
> Maybe have a look at the zwave binding as that has buttons like ‘heal device’ I see some similarity in use case. Only difference is you might want to have this button on binding level as removes items do not show up in the ui.

@lsiepel  thank you for the suggestion, i will have a look on zwave ui implementation.

bundles/org.openhab.io.homekit/src/main/java/org/openhab/io/homekit/internal/HomekitChangeListener.java
```diff
@@ -96,7 +104,7 @@ public void added(final Metadata metadata) {
                     try {
                         markDirty(itemRegistry.getItem(uid.getItemName()));
                     } catch (ItemNotFoundException e) {
-                        logger.debug("Could not find item for metadata {}", metadata);
+                        logger.trace("Could not find item for metadata {}", metadata);
```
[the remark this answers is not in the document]
yfre: thanks for explanation. you are right, it is not unexpected to have a metadata without item. agree, keep it as trace.

bundles/org.openhab.io.homekit/src/main/java/org/openhab/io/homekit/internal/accessories/DummyHomekitAccessory.java
yfre: yes, let keep it here. but we see that we add more and more low-level HAP things to the openHAB then maybe we should either merge Java-HAP to OH completely and move low level HAP things to Java-HAP

bundles/org.openhab.io.homekit/src/main/java/org/openhab/io/homekit/internal/accessories/AbstractHomekitAccessoryImpl.java
yfre: understood. thank you for the explanation. if JSON is not used directly as HAP payload, then it should be fine.

bundles/org.openhab.io.homekit/src/main/java/org/openhab/io/homekit/internal/HomekitChangeListener.java
yfre: right. missed that. thanks

ccutrer:
I tweaked how the logging works about dummy accessories so that it only logs what has changed, instead of the current status (and also includes logging when a dummy goes back to be properly backed). The build is now passing with the updated HAP-Java. 🤞 we can get this in soon.

yfre (approved):

ccutrer:
@lolodomo : did you want to review this again yourself before merging? I've got a few more HomeKit PRs I want to push up as soon as this is merged.

bundles/org.openhab.io.homekit/README.md
```diff
@@ -208,6 +210,22 @@ Switch light1 "Light 1" (gLight) {homekit="Lighting.OnState"}
 Switch light2 "Light 2" (gLight) {homekit="Lighting.OnState"}
 ```
 
+## Dummy Accessories
+
+OpenHAB is a highly dynamic system, and prone to occasional misconfigurations where items can't be loaded for various reasons, especially if you're using something besides the UI to manage your items.
+This is a problem for Homekit because if the bridge makes a connection, but accessories are missing, then the Homekit database will simply remove that accessory.
+When the accessory does come back (i.e. because you corrected a syntax error in an .items file, or OpenHAB completes booting), all customization of that accessory will be lost - the room assignment, customized  name, custom icon, status/home screen/favorite preferences, etc.
+In order to work around this, the Homekit addon can create dummy accessories for any accessory it has previously published to Homekit.
+To enable this behavior, turn on the `createDummyAccessories` setting.
```
lolodomo, line 219: Is it createXXX or useXXX?

bundles/org.openhab.io.homekit/README.md
```diff
@@ -208,6 +210,22 @@ Switch light1 "Light 1" (gLight) {homekit="Lighting.OnState"}
...
+OpenHAB will then simply present a non-interactive accessory for any that are missing.
+The OpenHAB log will also contain information whenever a dummy accessory is created.
+If the item backing the accessory is later re-created, everything will sync back up and nothing will be lost.
+You can also run the console command `openhab:homekit listDummyAccessories` to see which items are missing.
+Apple devices may or may not show "Not Responding" for some or all accessories when there are dummy accessories, since it won't be updating.
```
lolodomo, line 224: updating or updated?

bundles/org.openhab.io.homekit/src/main/java/org/openhab/io/homekit/internal/accessories/DummyHomekitAccessory.java
```diff
@@ -0,0 +1,164 @@
...
+import io.github.hapjava.accessories.HomekitAccessory;
+import io.github.hapjava.characteristics.Characteristic;
+import io.github.hapjava.services.Service;
+
+/**
+ * Implements a dummy placeholder accessory for when configuration is missing
+ *
+ * @author Cody Cutrer - Initial contribution
+ */
+@NonNullByDefault({})
```
lolodomo, line 37: Why not just "@NonNullByDefault"?

lolodomo (changes requested):
Review part 1 of 2

bundles/org.openhab.io.homekit/src/main/java/org/openhab/io/homekit/internal/accessories/DummyHomekitAccessory.java
ccutrer: It implements an interface from HAP-Java that is not `@NonNullByDefault`, and I was unable to convince the compiler to be okay with using just `@NonNullByDefault` via judicious use of `@Nullable`. I'm still not an expert on how adding `({})` works, but I _think_ it makes it `@NonNullByDefault` for "new" methods, and inherit the nullability for overridden methods, ensuring they're compatible if you're writing the signature the same way.

ccutrer:
🤦 I apologize. I accidentally pushed my "integration" branch to this PR, which includes several follow-on PRs that don't belong here. I've removed those commits.

lolodomo:
> I apologize. I accidentally pushed my "integration" branch to this PR, which includes several follow-on PRs that don't belong here. I've removed those commits.

The new big size dissuaded me from reviewing again last weekend.
Now, with the reduced size, I will do it probably this evening.

PR #13489, opened 2022-10-04
[netatmo] Buffer last event requests per child modules

### LINKED ISSUE #13358: [Netatmo] Rest call failed: statusCode=MAXIMUM_USAGE_REACHED
I have the following Netatmo devices in my OH3.3 system:
- 6 Netatmo cameras (5x Presence, 1x Welcome),
- Video Doorbell and
- Netatmo Weather Station (1x indoor, 1x outdoor, wind, rain) 

Reconnect Interval for Netatmo Account is set to 600.

There are warnings in the log saying I have reached the maximum user usage:
```
2022-09-06 09:38:51.368 [WARN ] [al.handler.capability.HomeCapability] - Error gettting Home informations : Rest call failed: statusCode=MAXIMUM_USAGE_REACHED, message=User usage reached
2022-09-06 09:38:51.440 [WARN ] [andler.capability.SecurityCapability] - Error retrieving last events of camera '70:ee:50:6f:xx:xx' : Rest call failed: statusCode=MAXIMUM_USAGE_REACHED, message=User usage reached
[same message for the other 5 cameras]
2022-09-06 09:45:21.369 [WARN ] [handler.capability.WeatherCapability] - Error retrieving weather data '70:ee:50:17:yy:yy' : Rest call failed: statusCode=MAXIMUM_USAGE_REACHED, message=User usage reached
```
Regardless of the warnings, the Binding/Things/Channel/Items work properly.

Are there any parameters to limit the number or the interval of Rest calls so that the maximum usage limit will not be reached?

Many thanks for your help!

Kind regards,
Joachim

----------------------------------------
openHAB 3.3.0 Release Build
Netatmo binding 3.3.0
Raspberry / Linux/5.10.103-v7l+ (arm)
Java Raspbian 11.0.16

clinique:
This is targetted to decrease the number of requests transmitted to Netatmo API. Resolves #13358


bundles/org.openhab.binding.netatmo/src/main/java/org/openhab/binding/netatmo/internal/handler/capability/SecurityCapability.java
```diff
@@ -110,7 +116,55 @@ protected void updateHomeEvent(HomeEvent homeEvent) {
                 });
     }
 
-    public Collection<HomeEvent> getDeviceEvents(String cameraId, String deviceType) {
+    @Override
+    protected List<NAObject> updateReadings(SecurityApi api) {
```
lolodomo, line 120: Just one question: are we sure that this method will be always called before any call to `updateReadings` in CameraCapability and PersonCapability ?

clinique: Should be but not guaranteed 100% but this should not be an issue because when requested event is not in the buffer it is grabbed especially for the device. Do you see a potential issue here ?

lolodomo: The issue I see is that the event for person / camera could then be not the most recent, rather the last present in the cache.

lolodomo: Yes, it will work in case the home events are always polled before the camera and the person are updated.
In case the camera/person refresh is (more or less) independent from the home, you could get provide an outdated event.
That was my origin question.

lolodomo:
The cache for persons is really a great idea as for some persons the new events can be very very rare. So it will be requested one time.

clinique:
> The cache for persons is really a great idea as for some persons the new events can be very very rare.

Indeed, in my case it does not change that much as I only validate persons things of the family living in the house but if you create all things for all persons recognized, it can grow a lot.

bundles/org.openhab.binding.netatmo/src/main/java/org/openhab/binding/netatmo/internal/handler/capability/SecurityCapability.java
clinique: Ok, I get your question, but I don't know how to address it. You're correct, until next home refresh, we're suspect to provide outdated events, but then I don't see how to avoid emitting a new request. It should not be the case when using webhooks, but apart from that...your ideas are welcome.

lolodomo: I do not remember exactly how you manage that. Don't you trigger the refresh of the home, the camera devices and the persons at the same frequency, more or less at the same time. Maybe even in parallel ?
If that is correct, one solution, if doable, would be to force the home refresh being finished before refreshing the camera devices and persons.

clinique: According to CommonInterface.java : 

```
    default List<NAObject> updateReadings() {
        List<NAObject> result = new ArrayList<>();
        getCapabilities().values().forEach(cap -> result.addAll(cap.updateReadings()));
        getActiveChildren().forEach(child -> result.addAll(child.updateReadings()));
        return result;
    }
```
Home will be updated (Api requests sent) in sequence just before childs (modules, cameras, persons). So I do not think this should generate outdated data in some milliseconds between the two events.

lolodomo: Good

lolodomo:
A conflict was introduced with the merge of another of your PRs. Can you please fix it?

clinique:
It's weird because when I click on "Resolve conflicts" it show 0 conflict on SecurityApi.java.
I'm going to make a force push.

lolodomo (approved):
LGTM, thank you

lolodomo:
@clinique : I probably merged too fast. To not loose camera / person events, your request to get all recent events should not retrieved a fixed number of events but rather all events since the last request. Does the API allow that? 
How many events are you requesting?

lolodomo:
The new request should at least contain the most recent event of the previous request to be sure to loose no event.

clinique:
> @clinique : I probably merged too fast. To not loose camera / person events, your request to get all recent events should not retrieved a fixed number of events but rather all events since the last request. Does the API allow that? How many events are you requesting?

Unfortunately not. There is only an option that allow retrieved x events before a given event.
That would mean that we need to setup a recursive loop : 

get current list of last 30 events
retrieve 30 events before the last of the list until retrieved events are after timestamp of previous known last event.

Do you agree ?

clinique:
Thinking of it twice, I see the issue : 
- an event is in the buffer for a given module but not any presented in the list of 30 => we could miss some new event.

PR #12042, opened 2022-01-14
[nuvo] Add NuvoNet source communication capabilities

mlobstein:
Implement NuvoNet source communication that allows for deeper interaction with the physical keypads to include custom menus, custom favorite lists and album art display. The user can configure any/all of the amplifier's 6 sources for use by openHAB. This will allow them to integrate an external source such as a Chromecast to the Nuvo keypads with full metadata and transport controls being passed through openHAB.

The custom menu feature allows the user to create a menu structure that will be available on the physical keypad.  By using rules, it is possible to execute an action on any other openHAB item as a result of selecting a menu item on the physical keypad.

A connection through an MPS4 music server to the Nuvo amplifier must be used in order to utilize these new features. All previous binding functionality available using a serial port connection is unchanged.

PR #13498, opened 2022-10-07
[nuvo] Add Image channel for album art

mlobstein:
This is a follow up to #12042. In the previous PR, the ability for album art to be loaded into the physical keypads was added. In this PR, an Image channel was added so that the album art can also be displayed in an openHAB UI widget.

### DIFF
--- a/bundles/org.openhab.binding.nuvo/src/main/java/org/openhab/binding/nuvo/internal/handler/NuvoHandler.java
+++ b/bundles/org.openhab.binding.nuvo/src/main/java/org/openhab/binding/nuvo/internal/handler/NuvoHandler.java
@@ -224,6 +224,13 @@ public class NuvoHandler extends BaseThingHandler implements NuvoMessageEventLis
             return;
         }
 
+        nuvoNetSrcMap.put("1", config.nuvoNetSrc1);
+        nuvoNetSrcMap.put("2", config.nuvoNetSrc2);
+        nuvoNetSrcMap.put("3", config.nuvoNetSrc3);
+        nuvoNetSrcMap.put("4", config.nuvoNetSrc4);
+        nuvoNetSrcMap.put("5", config.nuvoNetSrc5);
+        nuvoNetSrcMap.put("6", config.nuvoNetSrc6);
+
         if (this.isMps4) {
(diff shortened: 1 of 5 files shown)


bundles/org.openhab.binding.nuvo/src/main/java/org/openhab/binding/nuvo/internal/handler/NuvoHandler.java
```diff
@@ -1229,13 +1236,25 @@ private void cancelClockSyncJob() {
...
+
+    /**
+     * Update the state of a channel (overloaded method to handle album_art channel)
+     *
+     * @param target the channel group
+     * @param channelType the channel group item
+     * @param value the value to be updated
+     * @param bytes the byte[] to load into the Image channel
+     */
+    private void updateChannelState(NuvoEnum target, String channelType, String value, byte[] bytes) {
```
lolodomo, line 1257: I think it will be cleaner to have a nullable final parameter (`bytes`), rather than using NO_ART with a comparison `NO_ART == bytes`.
`byte[] @Nullable bytes`

lolodomo (changes requested):
Just one comment/proposal

bundles/org.openhab.binding.nuvo/src/main/java/org/openhab/binding/nuvo/internal/handler/NuvoHandler.java
mlobstein: I tried doing this but ran into this:
>  The nullness annotation @Nullable is not applicable for the primitive type byte
I have refactored to remove the comparison. The original path to setting the channel to UNDEF will be used to clear the image.

lolodomo:
Has the new part of the PR added after the review in link with the original PR? 
If not, why not create a separate PR with am appropriate description. This will also avoid delaying the merge of the original PR.

mlobstein:
> Has the new part of the PR added after the review a link with the original PR? If not, why not create a separate PR with an appropriate description. This will also avoid delaying the merge of the original PR.

Not directly.  I have reverted the those changes and will raise a new PR for that. You can continue with the review of this PR.

lolodomo (approved):
LGTM, thank you

PR #13504, opened 2022-10-07
[homekit] allow RelativeHumidity as an optional characteristic on Thermostat

ccutrer:
also support QuantityType as a valid state type for double characteristics (such as RelativeHumidity).


lolodomo:
@yfre  : maube you can make a first review

yfre (approved):
LGTM

yfre:
@ccutrer indeed, according to homekit spec, thermostat supports humidity as optional characteristic. not sure why i missed it. thanks for fixing it.

lolodomo (approved):
LGTM

PR #13505, opened 2022-10-07
[netatmo] Ensure all events are retrieved since they are buffered

clinique:
Following the merge of #13489, @lolodomo identified a potential edge case that could lead to miss some events.
Discussed [here](https://github.com/openhab/openhab-addons/pull/13489#issuecomment-1271441821)

### DIFF
--- a/bundles/org.openhab.binding.netatmo/src/main/java/org/openhab/binding/netatmo/internal/api/SecurityApi.java
+++ b/bundles/org.openhab.binding.netatmo/src/main/java/org/openhab/binding/netatmo/internal/api/SecurityApi.java
@@ -15,7 +15,10 @@ package org.openhab.binding.netatmo.internal.api;
 import static org.openhab.binding.netatmo.internal.api.data.NetatmoConstants.*;
 
 import java.net.URI;
-import java.util.Collection;
+import java.time.ZonedDateTime;
+import java.util.Comparator;
+import java.util.List;
+import java.util.stream.Collectors;
 
 import javax.ws.rs.core.UriBuilder;
 
@@ -66,7 +69,7 @@ public class SecurityApi extends RestManager {
         return true;
     }
 
-    private Collection<HomeEvent> getEvents(@Nullable Object... params) throws NetatmoException {
+    private List<HomeEvent> getEvents(@Nullable Object... params) throws NetatmoException {
         UriBuilder uriBuilder = getApiUriBuilder(SUB_PATH_GET_EVENTS, params);
         BodyResponse<Home> body = get(uriBuilder, NAEventsDataResponse.class).getBody();
         if (body != null) {
@@ -78,16 +81,23 @@ public class SecurityApi extends RestManager {
         throw new NetatmoException("home should not be null");
     }
 
-    public Collection<HomeEvent> getHomeEvents(String homeId) throws NetatmoException {
-        return getEvents(PARAM_HOME_ID, homeId);
+    public List<HomeEvent> getHomeEvents(String homeId, @Nullable ZonedDateTime oldestKnown) throws NetatmoException {
+        List<HomeEvent> events = getEvents(PARAM_HOME_ID, homeId);
+        if (oldestKnown != null) { // we have to rewind to get the latest event just after oldestKnown
+            HomeEvent oldestRetrieved = events.get(events.size() - 1);
+            while (oldestRetrieved.getTime().isAfter(oldestKnown)) {
+                events.addAll(getEvents(PARAM_HOME_ID, homeId, PARAM_EVENT_ID, oldestRetrieved.getId()));
+                oldestRetrieved = events.get(events.size() - 1);
+            }
+        }
+        return events.stream().sorted(Comparator.comparing(HomeEvent::getTime).reversed()).collect(Collectors.toList());
     }
 
-    public Collection<HomeEvent> getPersonEvents(String homeId, String personId) throws NetatmoException {
+    public List<HomeEvent> getPersonEvents(String homeId, String personId) throws NetatmoException {
         return getEvents(PARAM_HOME_ID, homeId, PARAM_PERSON_ID, personId, PARAM_OFFSET, 1);
     }
 
-    public Collection<HomeEvent> getDeviceEvents(String homeId, String deviceId, String deviceType)
-            throws NetatmoException {
+    public List<HomeEvent> getDeviceEvents(String homeId, String deviceId, String deviceType) throws NetatmoException {
         return getEvents(PARAM_HOME_ID, homeId, PARAM_DEVICE_ID, deviceId, PARAM_DEVICES_TYPE, deviceType);
     }
 
--- a/bundles/org.openhab.binding.netatmo/src/main/java/org/openhab/binding/netatmo/internal/api/SecurityApi.java
+++ b/bundles/org.openhab.binding.netatmo/src/main/java/org/openhab/binding/netatmo/internal/api/SecurityApi.java
@@ -81,18 +81,18 @@ public class SecurityApi extends RestManager {
         throw new NetatmoException("home should not be null");
     }
 
-    public List<HomeEvent> getHomeEvents(String homeId, @Nullable ZonedDateTime oldestKnown) throws NetatmoException {
+    public List<HomeEvent> getHomeEvents(String homeId, @Nullable ZonedDateTime freshestEventTime) throws NetatmoException {
         List<HomeEvent> events = getEvents(PARAM_HOME_ID, homeId);
 
         // we have to rewind to the latest event just after oldestKnown
         HomeEvent oldestRetrieved = events.get(events.size() - 1);
-        while (oldestKnown != null && oldestRetrieved.getTime().isAfter(oldestKnown)) {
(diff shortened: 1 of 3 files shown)


bundles/org.openhab.binding.netatmo/src/main/java/org/openhab/binding/netatmo/internal/api/SecurityApi.java
```diff
@@ -78,16 +81,26 @@ private Collection<HomeEvent> getEvents(@Nullable Object... params) throws Netat
         throw new NetatmoException("home should not be null");
     }
 
-    public Collection<HomeEvent> getHomeEvents(String homeId) throws NetatmoException {
-        return getEvents(PARAM_HOME_ID, homeId);
+    public List<HomeEvent> getHomeEvents(String homeId, @Nullable ZonedDateTime oldestKnown) throws NetatmoException {
+        List<HomeEvent> events = getEvents(PARAM_HOME_ID, homeId);
+
+        // we have to rewind to the latest event just after oldestKnown
+        HomeEvent oldestRetrieved = events.get(events.size() - 1);
```
lolodomo, line 88: Is the API documented that returned events will be sorted form the most recent to the last recent ?

bundles/org.openhab.binding.netatmo/src/main/java/org/openhab/binding/netatmo/internal/api/SecurityApi.java
```diff
@@ -78,16 +81,26 @@ private Collection<HomeEvent> getEvents(@Nullable Object... params) throws Netat
...
+        while (oldestKnown != null && oldestRetrieved.getTime().isAfter(oldestKnown)) {
+            events.addAll(getEvents(PARAM_HOME_ID, homeId, PARAM_EVENT_ID, oldestRetrieved.getId()));
```
lolodomo, line 90: Passing an event id will request all events before this one ?

bundles/org.openhab.binding.netatmo/src/main/java/org/openhab/binding/netatmo/internal/handler/capability/SecurityCapability.java
```diff
@@ -133,7 +140,10 @@ protected List<NAObject> updateReadings(SecurityApi api) {
                         eventBuffer.put(personId, event);
                     }
                 }
-            });
+                if (oldestKnown == null || event.getTime().isBefore(oldestKnown)) {
```
lolodomo, line 143: We want to save the most recent event retrieved, so for me you should use `isAfter` rather than `isBefore` ?

bundles/org.openhab.binding.netatmo/src/main/java/org/openhab/binding/netatmo/internal/api/SecurityApi.java
```diff
@@ -66,7 +69,7 @@ public boolean addwebhook(URI uri) throws NetatmoException {
         return true;
     }
 
-    private Collection<HomeEvent> getEvents(@Nullable Object... params) throws NetatmoException {
+    private List<HomeEvent> getEvents(@Nullable Object... params) throws NetatmoException {
         UriBuilder uriBuilder = getApiUriBuilder(SUB_PATH_GET_EVENTS, params);
         BodyResponse<Home> body = get(uriBuilder, NAEventsDataResponse.class).getBody();
```
lolodomo, line 74: You were talking about 30 events returned. Is it a number fixed by the API ?

lolodomo (changes requested):

bundles/org.openhab.binding.netatmo/src/main/java/org/openhab/binding/netatmo/internal/api/SecurityApi.java
clinique: Yes, events are returned from the most recent to the oldest.

bundles/org.openhab.binding.netatmo/src/main/java/org/openhab/binding/netatmo/internal/api/SecurityApi.java
clinique: Exactly.

bundles/org.openhab.binding.netatmo/src/main/java/org/openhab/binding/netatmo/internal/api/SecurityApi.java
clinique: Yes, by default they are paged by batches of 30 (this can be changed using the `size`parameter

lolodomo: We have to avoid the recursion as much as possible. Do you think 30 is enough to cover 95% of the normal usages (including the case you own several cameras) or do you think we should increase it?

clinique: It is hard to tell. The number of events presented between two polls depends highly upon setup size and the number of devices emitting events. We could implement a kind of recursion counter that would increase page size dynamically. Maybe in a second step ?

bundles/org.openhab.binding.netatmo/src/main/java/org/openhab/binding/netatmo/internal/handler/capability/SecurityCapability.java
clinique: No, I want to identify the oldest event.

bundles/org.openhab.binding.netatmo/src/main/java/org/openhab/binding/netatmo/internal/api/SecurityApi.java
lolodomo: Yes, can be improved later

bundles/org.openhab.binding.netatmo/src/main/java/org/openhab/binding/netatmo/internal/handler/capability/SecurityCapability.java
lolodomo: I don't understand at all in that case, we don't want to retrieve all the events since the origin at each poll, we just want all those after the last already retrieved.

lolodomo: I don't understand at all in that case. 
What you want is to keep trace is the last event retrieved in order to search those most recent than this one at next poll.

clinique: Yes, maybe you're correct. Let me rethink of this. I may have made a mistake here while building the logic.

bundles/org.openhab.binding.netatmo/src/main/java/org/openhab/binding/netatmo/internal/api/SecurityApi.java
```diff
@@ -78,16 +81,26 @@ private Collection<HomeEvent> getEvents(@Nullable Object... params) throws Netat
...
 
-    public Collection<HomeEvent> getHomeEvents(String homeId) throws NetatmoException {
-        return getEvents(PARAM_HOME_ID, homeId);
+    public List<HomeEvent> getHomeEvents(String homeId, @Nullable ZonedDateTime freshestEventTime) throws NetatmoException {
+        List<HomeEvent> events = getEvents(PARAM_HOME_ID, homeId);
+
+        // we have to rewind to the latest event just after oldestKnown
+        HomeEvent oldestRetrieved = events.get(events.size() - 1);
+        while (freshestEventTime != null && oldestRetrieved.getTime().isAfter(freshestEventTime)) {
+            events.addAll(getEvents(PARAM_HOME_ID, homeId, PARAM_EVENT_ID, oldestRetrieved.getId()));
```
lolodomo, line 90: Last question: this event (oldestRetrieved) will then be twice in the final list. No bad impact on the binding behaviour ?

lolodomo (changes requested):
Final question

bundles/org.openhab.binding.netatmo/src/main/java/org/openhab/binding/netatmo/internal/api/SecurityApi.java
clinique: I will not because the API provides the batch of events just before the one requested.

lolodomo: Perfect

lolodomo:
Please run spotless, the file SecurityApi.java broke the build.

lolodomo:
> Please run spotless, the file SecurityApi.java broke the build.

I just did it for you

lolodomo (approved):
LGTM, thank you

lolodomo:
I removed the "Bug" tag as it is a fix for an enhancement merged recently (yesterday I believe). This will avoid an entry in the release notes.

clinique:
Ok, thanks for this @lolodomo .

PR #13507, opened 2022-10-07
[homekit] fix BooleanItemReader to work with DimmerItems

ccutrer:
DimmerItems have a PercentType state, which is easily convertible to OnOffType, but is not inherited from it. So take that into account.

i.e. a variable speed fan can use a single DimmerItem for both ActiveStatus and RotationSpeed.


lolodomo:

bundles/org.openhab.io.homekit/README.md
```diff
@@ -676,7 +676,7 @@ Support for this is planned for the future release of openHAB HomeKit binding.
...
-|                      | ActiveStatus                |                              | Switch                        | Accessory current working status. A value of "ON"/"OPEN" indicates that the accessory is active and is functioning without any errors.                                                                                                                                                                                                              |
+|                      | ActiveStatus                |                              | Switch, Dimmer                | Accessory current working status. A value of "ON"/"OPEN" indicates that the accessory is active and is functioning without any errors.                                                                                                                                                                                                              |
```
yfre, line 679: please add Dimmer as supported item to ActiveStatus of other characteristics. if they use the same booleanitemreader. 
at least for HeaterCooler and Valve i could think of an use cases.

yfre:
@ccutrer interesting case and an good fix. as BooleanItemReader is used by other accessories, please add Dimmer as supported item type to the docu for them

yfre (approved):
LGTM

ccutrer:
force-pushed to avoid a conflict on the CurrentSlatState line that fixed a spacing issue. I must have fixed it in another one off my commits this week.

ccutrer:
... and now rebased and resolved a conflict on the BatteryChargingStatus line before the Filter accessory.

PR #13508, opened 2022-10-07
[homekit] allow configuring maxValue on VOCDensity

ccutrer:
the default of 1000 is quite low in reality. tested and confirmed working with iOS 16


lolodomo:

yfre:
@ccutrer good idea.  but why you have not added the support for step? 
not sure whether there is an use case for step, but Homekit and Java-HAP supports custom step and for us is just one line.
so, if it is ok for you, please add support for custom step

bundles/org.openhab.io.homekit/src/main/java/org/openhab/io/homekit/internal/accessories/HomekitCharacteristicFactory.java
```diff
@@ -800,6 +800,11 @@ private static PM10DensityCharacteristic createPM10DensityCharacteristic(final H
     private static VOCDensityCharacteristic createVOCDensityCharacteristic(final HomekitTaggedItem taggedItem,
             HomekitAccessoryUpdater updater) {
         return new VOCDensityCharacteristic(
+                taggedItem.getConfigurationAsDouble(HomekitTaggedItem.MIN_VALUE,
+                        VOCDensityCharacteristic.DEFAULT_MIN_VALUE),
+                taggedItem.getConfigurationAsDouble(HomekitTaggedItem.MAX_VALUE,
+                        VOCDensityCharacteristic.DEFAULT_MAX_VALUE),
+                1.0,
```
yfre, line 807: please add support for step

taggedItem.getConfigurationAsDouble(HomekitTaggedItem.STEP,
                        VOCDensityCharacteristic.DEFAULT_STEP),

ccutrer: sure. I didn't bother with step, because it's a read only attribute and AFAIK step should only affect UIs for _writing_ things, to make it more helpful. but maybe it does respect it for something else, and I did min -- which one would think should always be zero -- so might as well. Though I could see setting min as something higher than zero if your sensor has a floor of how low it can detect. For example, my CO2 sensors have a floor of 400ppm.

yfre (approved):
LGTM

lolodomo (approved):

PR #13510, opened 2022-10-07
[homekit] get min/max/step values from state description if possible

ccutrer:
this helps to auto-configure if the binding is providing the necessary info

metadata config still overrides


lolodomo:

yfre (approved):
LGTM

yfre:
@ccutrer thank you
good idea. we should reduce the need of double configuration. if we can get something from semantic model we should use it. 
good, that one can overwrite the model settings for homekit.

lolodomo (approved):
LGTM

PR #13521, opened 2022-10-08
[caddx] Correct status initialization of the things after a restart

jossuar:
Signed-off-by: Georgios Moutsos <georgios.moutsos@gmail.com>

After a binding restart the things could not be found in the Hashmap because the keys were BigDecimal.
The keys were added in the hasmap with a decimal place e.g.: 1.0
BigDecimal.equals checks for the value and the precision. The result is that 1 <> 1.0 and the Hashmap lookups failed.


wborn (approved):
Thanks!

Issue #13522, opened 2022-10-09
[gce] Dynamic creation of channel prevents usage of .things file


clinique:
## Expected Behavior
Thing should be created without error message, even if it is declared in a things file.

## Current Behavior
Declaring the IPX thing in a things file conflicts with dynamic creation of channels because the creation does not take in account already existing channels.

## Possible Solution
Take in account existing channels when defining dynamically created once.

## Steps to Reproduce (for Bugs)
1. Create a things file : 
```
Thing gce:ipx800v3:ipx "GCE IPX800" @ "node0" [hostname="192.168.0.244", portNumber=9870] 
{
	Channels:
		Type relay : relay#1 [       // Actionneur porte de garage
			pulse=true    
		]
}
```
2. Save the things file
3. Look at openhab.log : 
```
2022-10-09 09:33:35.534 [ERROR] [nal.common.AbstractInvocationHandler] - An error occurred while calling method 'ThingHandler.initialize()' on 'org.openhab.binding.gce.internal.handler.Ipx800v3Handler@620bb655': Duplicate channels gce:ipx800v3:ipx:relay#1
java.lang.IllegalArgumentException: Duplicate channels gce:ipx800v3:ipx:relay#1
	at org.openhab.core.thing.util.ThingHelper.ensureUniqueChannels(ThingHelper.java:152) ~[?:?]
	at org.openhab.core.thing.util.ThingHelper.ensureUniqueChannels(ThingHelper.java:128) ~[?:?]
	at org.openhab.core.thing.binding.builder.ThingBuilder.withChannels(ThingBuilder.java:86) ~[?:?]
	at org.openhab.binding.gce.internal.handler.Ipx800v3Handler.discoverAttributes(Ipx800v3Handler.java:174) ~[?:?]
	at org.openhab.binding.gce.internal.handler.Ipx800v3Handler.initialize(Ipx800v3Handler.java:124) ~[?:?]
	at jdk.internal.reflect.GeneratedMethodAccessor87.invoke(Unknown Source) ~[?:?]
	at jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43) ~[?:?]
	at java.lang.reflect.Method.invoke(Method.java:566) ~[?:?]
	at org.openhab.core.internal.common.AbstractInvocationHandler.invokeDirect(AbstractInvocationHandler.java:154) [bundleFile:?]
[...]

PR #13526, opened 2022-10-09
[gce] Avoid conflicts with things file defined

clinique:
Avoiding dynamic creation of channels when channels already exists.
Plus some small code improvements.
Solves #13522 

### DIFF
--- a/bundles/org.openhab.binding.gce/src/main/java/org/openhab/binding/gce/internal/handler/Ipx800v3Handler.java
+++ b/bundles/org.openhab.binding.gce/src/main/java/org/openhab/binding/gce/internal/handler/Ipx800v3Handler.java
@@ -101,7 +101,8 @@ public class Ipx800v3Handler extends BaseThingHandler implements Ipx800EventList
         @Override
         public void run() {
             PortData currentData = portDatas.get(port);
-            if (currentData != null && currentData.getValue() == 1 && currentData.getTimestamp() == referenceTime) {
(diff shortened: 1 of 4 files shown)


bundles/org.openhab.binding.gce/src/main/java/org/openhab/binding/gce/internal/handler/Ipx800v3Handler.java
```diff
@@ -125,54 +126,51 @@ public void initialize() {
...
-        PortDefinition.asStream().forEach(portDefinition -> {
-            int nbElements = statusFile.getMaxNumberofNodeType(portDefinition);
-            for (int i = 0; i < nbElements; i++) {
-                createChannels(portDefinition, i, channels);
-            }
-        });
-
-        thingBuilder.withChannels(channels);
-        updateThing(thingBuilder.build());
+        if (channels.isEmpty()) {
```
lolodomo, line 163: This is IMHO the wrong approach. You should create each dynamic channel but only after checking it does not yet exist.

lolodomo (changes requested):

bundles/org.openhab.binding.gce/src/main/java/org/openhab/binding/gce/internal/handler/Ipx800v3Handler.java
clinique: You're correct, I've enhanced the approach.

PR #13527, opened 2022-10-09
[easee] Fix number conversion error

alexf2015:
This bug was reported by the community.
https://community.openhab.org/t/easee-binding/135492/58

The binding tries to parse a decimal value as integer but fails as it should be parsed as double instead.


lolodomo (approved):
LGTM

PR #13529, opened 2022-10-09
[actiontemplatehli] Simplify usage and readme

GiviMAD:
I have have rewritten some parts of this service, to simplify its usage, as I think I overcomplicate things too much on the initial contribution.

I have removed a lot of NLP terminology and non intuitive explanations from the README while updating it to match all the other changes introduced.

I have remove all the NLP terminology from the placeholders configuration names, as it was not user friendly. Also the configuration files for the placeholder values and transformations are all in json format instead of xml, and the transformations files (now called mappedValues of the placeholder) can be reused between read and write actions because the keys define the 'user readable' value and the values the equivalent 'item state'.

Also have improved the template system:
* You can made tokens optional on the template by adding '?' to the end. Also I've removed the configuration option for made optional all tokens matching a certain tag, as I think it's more controlable and intuitive this way.
* Now you can choose to match a specific token by tag or lemma by prefixing it, so you can do thinks like "how \<lemma>be the $itemLabel". So I have removed the action type "lemmas".

Only think new that I have added is a new type for the action called "dice", to use a dice's coefficient comparison instead of the default one (equality).

Sorry for the PR been so big, I think the changes were needed as the initial version miss a lot of usability and the options for the placeholders were hard to configure, also the readme itself. I'll try not to made a change so big again because I understand reviewing this is a lot of work.

Thank you for your time!

### DIFF
--- a/bundles/org.openhab.voice.actiontemplatehli/README.md
+++ b/bundles/org.openhab.voice.actiontemplatehli/README.md
@@ -1,25 +1,16 @@
 # Action Template Interpreter
 
-A human language interpreter implementation powered by OpenNLP.
-This is an attempt to provide you with a template system to match text commands to specific items and read its state or send command to them.
-For doing this the interpreter takes advantage of some nlp techniques.
+This interpreter does nothing on his own.
+It provides you with a template system to match text commands to items and read its states or send commands to them.
 
-The (Apache OpenNLP)[https://opennlp.apache.org] library is a machine learning based toolkit for the processing of natural language text.
+When reading from an item you can transform its state to a human-readable text and template it into your defined response.
+
+When writing to an item you can extract parts of the text and transform those in a valid item command.
+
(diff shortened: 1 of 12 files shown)


GiviMAD:
Fixed build errors and regenerated the translations.

GiviMAD:
I needed to add another commit because I detected some errors on the scoring mechanism.
I've added some unit tests for those situations.

lolodomo:
@GiviMAD : why did you close this PR ? You thing the change is not more a good idea ?

GiviMAD:
Hi @lolodomo, I'd moved this pr version to the marketplace.
I think that a couple of things can be improved, and I think I'll remove the dice comparison as it's difficult to understand, and make a little UI to help the phrase configuration, so this is as easy to use as possible, because based on the little feedback I got I think is too complex for users to be interested on it, but I hope with some changes it gets more attractive.

PR #13535, opened 2022-10-10
[nuvo] Ignore NuvoNet events for sources not managed by openHAB

mlobstein:
Address bug from MR that added NuvoNet source capability to OH (#12042). Fixes errors that occur if the binding processes NuvoNet events intended for NuvoNet sources not handled by openHAB.


lolodomo (approved):
LGTM

Issue #13423, opened 2022-09-20
[vosk] Missing native dependency


lolodomo:
@GiviMAD : after installing the snapshot 3089 (generated yesterday), I can see this warning at OH startup:
```
20:24:37.125 [WARN ] [voice.voskstt.internal.VoskSTTService] - Missing native dependency: Error looking up function 'vosk_recognizer_set_partial_words': /home/xxx/.cache/JNA/temp/jna2618161390205275901.tmp: undefined symbol: vosk_recognizer_set_partial_words
```
Any idea what is wrong ?

GiviMAD:
I have this same error after updating the library, and I found a comment on the vosk issue tracker that said it's caused by not doing a clean install after the dependency upgrade. So I clean install the binding and the problem got solved.
Is maybe the CI using the cache of the previous build? If so it's maybe needed to clean the cache for that module.

lolodomo:
I removed the `.cache` directory and restarted OH. Still the same warning.

lolodomo:
Could it be my vosk model being outdated ?
I have not downloaded a new ZIP file since a long time.
I am using this file vosk-model-small-fr-0.22.zip

GiviMAD:
> I removed the `.cache` directory and restarted OH. Still the same warning.

Sorry, I think I explained myself wrong, I was talking about the maven cache.

So what happened to me was:
I build the binding doing mvn install -pl org.openhab.binding.voskstt and that build gave that error, after that I rebuild with mvn clean install -pl org.openhab.binding.voskstt and that jar worked for me perfectly.
So I think the issue happened  on the build, not related to your local installation.

GiviMAD:
This is the issue where I found the solution https://github.com/alphacep/vosk-api/issues/574

lolodomo:
Ah ok. 
I am using the jar file from the kar file built last night. We should be sure that this one is working. 
I have not played with vosk. I guess that it doesn't work at all when we have this warning?

GiviMAD:
No it doesn't work at all when that warning is shown.

GiviMAD:
I'm not sure where that kar is built, I think that by Jenkins. In that case maybe we need someone with access to invalidate/delete de build cache for the Addons-repo.

lolodomo:
Maybe @wborn can have an idea.

wborn:
[...]

PR #13536, opened 2022-10-10
[voskstt] fix sdk update and disable internal info logs

GiviMAD:
Fixes https://github.com/openhab/openhab-addons/issues/13423.

It was caused by me when updating the sdk. 
As a summary of what happened, support for arm and arm64 (meaning vosk binaries) is not included in the vosk maven package, so last time I grabbed them from the GitHub release page of the project. This time I forgot to updated those files, causing the problem. 

Also changes the vosk internal log level to suppress unwanted info messages.

Sorry for taking so long to realize what the problem was.


lolodomo (approved):
LGTM, thank you

lolodomo:
Of course, if you have fixed that yesterday just before the build of the new milestone, that would have been better ;)
Checking how many users are complaining will be a good measure of this binding audience ;)

PR #13539, opened 2022-10-12
[unifi] Fix PoE port mode for 24V

Hilbrand:
The value UniFi expects is actually pasv24. Tested with different UniFi controller versions.
(Added myself as codeowner to the UniFi binding, have been adding features lately).


bundles/org.openhab.binding.unifi/src/main/resources/OH-INF/i18n/unifi_nl.properties
```diff
@@ -82,7 +82,7 @@ channel-type.unifi.poeMode.label = PoE-modus
 channel-type.unifi.poeMode.description = De PoE-modus waarin de poort zich bevindt.
 channel-type.unifi.poeMode.state.option.off = Uit
 channel-type.unifi.poeMode.state.option.auto = Automatisch
-channel-type.unifi.poeMode.state.option.24v = 24V
```
jlaur, line 85: This file should be omitted, and strings should be updated in Crowdin after PR has been merged.

jlaur (changes requested):
Thank you! Only a minor comment for following Crowdin workflow.

jlaur (approved):
LGTM

PR #13545, opened 2022-10-13
[gce] File based items are not updated

clinique:
The linkChannel and unlinkChannel methods are not called for file based items linked to channels.
This lead the binding to be unreactive to input channels.

Some code enhancement :
Check if channel is linked before sending updates.
Removed as much usage to @NonNullByDefault({}) and @Nullable than possible.


lolodomo (approved):
LGTM

PR #13549, opened 2022-10-14
[benqprojector] Log invalid number when expecting numeric response from projector

mlobstein:
Fixes very minor issue by catching the NumberFormatException that is thrown if queryInt() receives a non-numeric response from the projector. I have seen this very rarely, but when it occurs (due to noisy serial line, cosmic rays, buggy projector firmware, etc.) a warning is logged by updateChannelState(). With the change the NumberFormatException will be caught and logged to the debug log instead.


lolodomo (approved):
LGTM

PR #13550, opened 2022-10-14
[epsonprojector] Log invalid number when expecting numeric response from projector

mlobstein:
Fixes same issue as #13549 but in the epsonprojector binding.


lolodomo (approved):
LGTM

PR #13556, opened 2022-10-16
[voskstt] Fix load on linux arm

GiviMAD:
Try to fix error on armv7.


lolodomo:
After an update of the bundle, a clean of `.cache/JNA` folder and a restart of openHAB, the exception is no more present, the bundle is active and I have this warning:
```
09:10:31.660 [WARN ] [voice.voskstt.internal.VoskSTTService] - Missing native dependency: /home/xxx/.cache/JNA/temp/jna511182918146398935.tmp: undefined symbol: __atomic_load_8
```

GiviMAD:
This last version works fine for me, only requirement is have libatomic installed.

lolodomo:
Please update README to mention the prerequisite and how to install the missing library.

GiviMAD:
> Please update README to mention the prerequisite and how to install the missing library.

Ok I'll add a "supported platforms" sectione with this information.

bundles/org.openhab.voice.voskstt/src/main/java/org/openhab/voice/voskstt/internal/VoskSTTService.java
```diff
@@ -96,7 +92,18 @@ public VoskSTTService(@Reference LocaleService localeService) {
 
     @Activate
     protected void activate(Map<String, Object> config) {
-        configChange(config);
+        try {
+            String osName = System.getProperty("os.name", "generic").toLowerCase();
+            String osArch = System.getProperty("os.arch", "").toLowerCase();
+            if (osName.contains("linux") && osArch.equals("arm") || osArch.equals("armv7l")) {
```
lolodomo, line 98: I believe parenthesis are missing around the 2 last conditions.

lolodomo (changes requested):
No need to update the library finally?

GiviMAD:
> No need to update the library finally?

No I checked it agains and outdated branch, the binary was correctly downloaded. Once I merge master to this branch that change was the same, also the call to 'LibVosk.setLogLevel(LogLevel.WARNINGS);' was missed in this branch, that is why it failed different earlier. The problem was always the missing libatomic library. Sorry for the whole confusion.

lolodomo (approved):
LGTM

lolodomo:
After installing snapshot 3143, I can tell you that I have no warning/error in logs at startup.
I did not install libatomic.
That being said, I can't tell you if the service is working or not (not tested).

GiviMAD:
> After installing snapshot 3143, I can tell you that I have no warning/error in logs at startup.
> I did not install libatomic.
> That being said, I can't tell you if the service is working or not (not tested).

Thank you for the confirmation, in my tests I enabled the model load and it works without errors, so I think it will work well.
Probably libatomic is already installed in raspbian, you can check if the library is present using the path:
```
ls /usr/lib/arm-linux-gnueabihf/libatomic.so.1
```

lolodomo:
Confirmed, the file already exists in my system (RPI3).

PR #13560, opened 2022-10-16
[MAP] Adds fallback to original value when input not found

### LINKED ISSUE #10092: [MAP transformation] Pass on missing string
Hello,

I translate incoming strings with the MAP transformation in another language. When I have in the .map-File no translation for an incoming string, the string is empty on OpenHAB and I have the following message in the log:
Could not transform 'xxx' with the file 'de.map' : Target value not found in map for 'xxx'.
The originate string would be better than none at all. I have two ideas to fix the issue:

Idea 1:
In the map-file add the default-entry with %s and resolve the placeholder. Example:
`=%s`

Idea 2:
In the "Incoming Value Transformations"-section choose a fallback.
Example: `MAP:de.map:%s` instead of `MAP:de.map`

Thanks

clinique:
Resolves #10092

### DIFF
--- a/bundles/org.openhab.transform.map/README.md
+++ b/bundles/org.openhab.transform.map/README.md
@@ -8,7 +8,7 @@ The file format is documented [here](https://docs.oracle.com/javase/8/docs/api/j
 To organize the various transformations one might use subfolders.
 
 A default value can be provided if no matching entry is found by using "=value" syntax. 
-sDefining this default value using `%s` would then return the non transformed input string.
+Defining this default value using `%s` would then return the non transformed input string.
 
 
 ## Example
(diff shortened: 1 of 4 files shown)


bundles/org.openhab.transform.map/README.md
```diff
@@ -7,7 +7,9 @@ This file should be in property syntax, i.e. simple lines with "key=value" pairs
 The file format is documented [here](https://docs.oracle.com/javase/8/docs/api/java/util/Properties.html#load-java.io.Reader-).
 To organize the various transformations one might use subfolders.
 
-A default value can be provided if no matching entry is found by using "=value" syntax
+A default value can be provided if no matching entry is found by using "=value" syntax. 
+Defining this default value using `%s` would then return the non transformed input string.
```
lolodomo, line 11: %s looks a little technical for a final user and not intuitive at all?

lolodomo: Maybe `_source_` ?

jlaur:
@clinique, if you change "Solves" to "Resolves" in the PR description, the issue will be automatically linked and closed upon PR completion.

bundles/org.openhab.transform.map/README.md
clinique: Ok, good idea. Let's do that.

lolodomo (approved):
LGTM

lolodomo:
I see no real risk with that change, the probability that someone defined `_source_` is very low I believe.

Issue #13524, opened 2022-10-09
[systeminfo] SysteminfoOSGiTest unstable


wborn:
This test failed in: https://github.com/openhab/openhab-addons/actions/runs/3213563417/jobs/5253323286

```
TEST org.openhab.binding.systeminfo.test.SysteminfoOSGiTest#assertMediumPriorityChannelIsUpdated() <<< ERROR: 
Expected: is <UNDEF>
     but: was <NULL>
java.lang.AssertionError: 
Expected: is <UNDEF>
     but: was <NULL>
	at org.hamcrest.MatcherAssert.assertThat(MatcherAssert.java:20)
	at org.hamcrest.MatcherAssert.assertThat(MatcherAssert.java:6)
	at org.openhab.binding.systeminfo.test.SysteminfoOSGiTest.lambda$4(SysteminfoOSGiTest.java:267)
	at org.openhab.core.test.java.JavaTest.waitForAssert(JavaTest.java:245)
	at org.openhab.core.test.java.JavaTest.waitForAssert(JavaTest.java:213)
	at org.openhab.core.test.java.JavaTest.waitForAssert(JavaTest.java:177)
	at org.openhab.binding.systeminfo.test.SysteminfoOSGiTest.assertItemState(SysteminfoOSGiTest.java:265)
	at org.openhab.binding.systeminfo.test.SysteminfoOSGiTest.assertMediumPriorityChannelIsUpdated(SysteminfoOSGiTest.java:336)
```

This test failed in: https://github.com/openhab/openhab-addons/actions/runs/3195031635/jobs/5215227925

```
TEST org.openhab.binding.systeminfo.test.SysteminfoOSGiTest#assertChannelCpuDescriptionIsUpdated() <<< ERROR: 
Expected: is <Mocked Cpu Descr>
     but: was <NULL>
java.lang.AssertionError: 
Expected: is <Mocked Cpu Descr>
     but: was <NULL>
	at org.hamcrest.MatcherAssert.assertThat(MatcherAssert.java:20)
	at org.hamcrest.MatcherAssert.assertThat(MatcherAssert.java:6)
	at org.openhab.binding.systeminfo.test.SysteminfoOSGiTest.lambda$4(SysteminfoOSGiTest.java:267)
	at org.openhab.core.test.java.JavaTest.waitForAssert(JavaTest.java:245)
	at org.openhab.core.test.java.JavaTest.waitForAssert(JavaTest.java:213)
[...]

PR #13562, opened 2022-10-17
[systeminfo] dynamic channels

mherwege:
This PR enhances the systeminfo binding to expose all devices on the system when multiple devices of the same type exist. So far, the binding only exposed the first of such devices.

- These extra devices are exposed through dynamically created channels in dynamically created device groups. This will happen - for storage, drive, display, battery and network device groups. The new device groups will have an index appended (starting from 0). If no extra devices are detected, only the channel group without index will exist.
- Extra fanspeed channels in the sensors group may also get created if the system has multiple fans.
- Addition of a process channel group for the current process that is running OH. This makes monitoring the OH process easier, as the PID would change with a restart.


lolodomo:
Looks like the integration tests are not working.

mherwege:
> Looks like the integration tests are not working.

I noticed and tried understanding why. I have made some adjustments, which solved part of the issues. But I am left with 3 tests that consistently fail with no changes to the code for the functionality in these tests. I can't figure it out.
I have disabled them for now. The test set runs successfully on my local machine with these 3 disabled. But I would like to get some help to understand what's going on here.

mherwege:
> Do you think you could have fixed #13524 ?

I sure looks similar to what I saw. Let's hope it has that effect.

I noticed the following line in the test setup code sometimes returned null, probably because the service was not registered in time.
```
systeminfoHandlerFactory = getService(ThingHandlerFactory.class, SysteminfoHandlerFactory.class);
```
I put a `waitForAssert` around it, and I have not seen the problem since. I did similar things in other parts of the code as well. It least is should show a more relevant assert failure when it fails again.

bundles/org.openhab.binding.systeminfo/src/main/resources/OH-INF/thing/channels.xml
```diff
@@ -144,6 +156,14 @@
 		<config-description-ref uri="channel-type:systeminfo:mediumpriority"/>
 	</channel-type>
 
+	<channel-type id="path">
+		<item-type>String</item-type>
+		<label>Path</label>
+		<description>The full path</description>
```
lolodomo, line 162: I assume it should be present/added in the default translation file.

bundles/org.openhab.binding.systeminfo/src/main/resources/OH-INF/thing/channels.xml
```diff
@@ -107,7 +107,7 @@
 		<channels>
 			<channel id="name" typeId="name"/>
 			<channel id="description" typeId="description"/>
-			<channel id="load" typeId="cpuLoad"/>
+			<channel id="load" typeId="load"/>
```
lolodomo, line 110: Breaking change. Is it necessary to rename this channel type ?

bundles/org.openhab.binding.systeminfo/src/main/resources/OH-INF/thing/channels.xml
```diff
@@ -288,20 +308,20 @@
 		<config-description-ref uri="channel-type:systeminfo:mediumpriority"/>
 	</channel-type>
 
-	<channel-type id="load_process">
+	<channel-type id="load">
```
lolodomo, line 311: Entries in the translation properties should be updated: cpuLoad => load

lolodomo (changes requested):
Partial review (remains to review the thing handler and the unit tests).

You renamed a thing type which breaks the compatibility with already created things in UI.
I have the feeling that this renaming was not mandatory ?

lolodomo:
At my current review step, I am still not sure to understand why you need to create dynamically different thing types.
My feeling is that you can have still one unique thing type and when you create a thing from this type, you add dynamically new groups.
So it looks like you are creating a thing type for each thing and I do not really understand why.

jaywiseman1971:
"Addition of a process channel group for the current process that is running OH. This makes monitoring the OH process easier, as the PID would change with a restart." 

I'm interested in this piece above.  Have a recently posted to understand how to get the current binding to work as such w/o anybody replying.

https://community.openhab.org/t/oh-3-2-systeminfo-binding-process-pi4/140406

Best, Jay

mherwege:
> At my current review step, I am still not sure to understand why you need to create dynamically different thing types.
> My feeling is that you can have still one unique thing type and when you create a thing from this type, you add dynamically new groups.
> So it looks like you are creating a thing type for each thingroupsg and I do not really understand why.

It was my understanding you can add channels dynamically, but not channel groups. There are no methods for that. One needs to create a new thing type. If I would create the same thing type, I would run into a conflict, or cannot use the static definition at all anymore. I would be very happy to be proven wrong about this. I believe there are very few bindings creating channel groups (from the top of my head: mqtt).
I don’t have access to my PC at the moment, and won’t have time to look much onto this until next week.

bundles/org.openhab.binding.systeminfo/src/main/resources/OH-INF/thing/channels.xml
mherwege: It would indeed be a breaking change, but the channel only got introduced invthe previous binding. And when calling it load, it can be used in multiple channel groups, cpuLoad does not fit. It is more consistent with the other channels used in multiple groups.

lolodomo: When you write "the channel only got introduced in the previous binding", you mean when and in what OH version ?

lolodomo:
> It was my understanding you can add channels dynamically, but not channel groups. There are no methods for that. One needs to create a new thing type.

It looks like you are right. I don't know why we have this restriction in the core framework.

bundles/org.openhab.binding.systeminfo/src/main/resources/OH-INF/thing/channels.xml
mherwege: I mean, in the previous PR, and that is in the current milestone, but not release version.

lolodomo: Ok, I found the PR, I merged it in September 11. So only users who installed recent 3.4 milestones could be impacted. I remove the tag.

lolodomo (changes requested):
Review of changes in the thing handler

lolodomo:
End of review.

lolodomo:
@mherwege : any chance you update the default translations, fix the typo and answer to the remaining question before Sunday?

bundles/org.openhab.binding.systeminfo/src/main/java/org/openhab/binding/systeminfo/internal/handler/SysteminfoHandler.java
```diff
@@ -91,31 +103,59 @@ public class SysteminfoHandler extends BaseThingHandler {
...
     public void initialize() {
+        logger.trace("Initializing thing {} with thing type {}", thing.getUID().getId(),
+                thing.getThingTypeUID().getId());
+        restoreChannelsConfig(); // After a thing type change, previous channel configs will have been stored, and will
+                                 // be restored here.
         if (instantiateSysteminfoLibrary() && isConfigurationValid() && updateProperties()) {
-            groupChannelsByPriority();
-            scheduleUpdates();
-            updateStatus(ThingStatus.ONLINE);
+            if (!addDynamicChannels()) { // If there are new channel groups, the thing will get recreated with a new
```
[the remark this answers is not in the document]
mherwege: Indeed, I missed that. I don't think it can happen, as the computer thing type is the default one from the xml definition, but I added a thing status update to cope with it if it would.

mherwege:
@lolodomo I did the adjustments. Thank you for the review.

bundles/org.openhab.binding.systeminfo/src/main/java/org/openhab/binding/systeminfo/internal/handler/SysteminfoHandler.java
```diff
@@ -241,10 +241,14 @@ private boolean addDynamicChannels() {
...
             ThingTypeUID thingTypeUID = new ThingTypeUID(BINDING_ID, THING_TYPE_COMPUTER_ID + idExtString);
             if (thingTypeProvider.updateThingType(thingTypeUID, newChannelGroups)) {
-                logger.trace("Channel groups where added, changing the thing type");
+                logger.trace("Channel groups were added, changing the thing type");
                 changeThingType(thingTypeUID, thing.getConfiguration());
+                return true;
+            } else {
+                updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.HANDLER_INITIALIZING_ERROR,
+                        "@text/offline.cannot-initialize");
+                return false;
```
lolodomo, line 250: If you return false, this will trigger the additional code in `initialize` that finally sets the status to ONLINE.
So either you have to return true there or you return false but do not set the status to OFFLINE there.
Something looks not coherent.

lolodomo (changes requested):

bundles/org.openhab.binding.systeminfo/src/main/java/org/openhab/binding/systeminfo/internal/handler/SysteminfoHandler.java
mherwege: @lolodomo Indeed, didn't think it through well enough. I think the new commit should fix that.

lolodomo (approved):
LGTM, thank you

lolodomo:
@mherwege : could you please have a look to the above community topic, it looks like your PR broke the binding... for at least one user.

mherwege:
> @mherwege : could you please have a look to the above community topic, it looks like your PR broke the binding... for at least one user.

@lolodomo First analysis hints it may be linked to a file based thing definition. I don’t immediately see a way to fix that, and I am short in time at the moment to further analyse. I will do so later. Would you know if there is any way to detect if the definition was file based? If so, I could disable dynamic channels if one uses file thing definitions.

PR #13566, opened 2022-10-18
[homekit] improve bundle deactivation time

ccutrer:
If you have many instances, it can take a while. So stop all the instances in parallel. Also, fix a race condition where the update debouncer might get called again after being stopped, because the change listener was deregistered _after_ the debouncer was stopped.


lolodomo (approved):
LGTM

lolodomo:
@yfre : please comment if you like before I merge.

yfre (approved):
LGTM

jlaur (approved):
Thank you!

PR #13575, opened 2022-10-20
[mqtt] support mired<->Kelvin unit conversion

ccutrer:
Just have to use the newer toInvertibleUnit method that allows the dimension to change.


lolodomo (approved):
LGTM

PR #13582, opened 2022-10-22
[jsscripting] Fix multi-thread access

### LINKED ISSUE #13022: [jsscripting] Multi threaded access is not allowed for language(s) js.
<a alt="bounty" href="https://www.bountysource.com/issues/112778339"></a>

Using openHAB 3.2 in two `/etc/openhab/automation/js/*.js` files I have:
```javascript
class R {
    constructor(z) {
        this.item = items.getItem(z)
    }
    get isOn() {
        return this.item.state == "ON"
    }
    turn(state) {
        this.item.sendCommand(state)
    }
}

const ri = new R('i'), rz = new R('z')

rules.JSRule({
    name: 'ABC',
    description: 'ABC',
    triggers: [
        triggers.ItemStateChangeTrigger('A', null, "0 W", "Y1"),
        triggers.ItemStateChangeTrigger('B', null, "0 W", "Y2")
    ], execute: (data) => {
            console.error('M', ri.isOn, rz.isOn)
            ri.turn('OFF')
            rz.turn('OFF')
            setTimeout(() => console.error('N', ri.isOn, rz.isOn), 1000)
            setTimeout(() => console.error('P', ri.isOn, rz.isOn), 2000)
    }
})
```
Running the script logs:
```
2022-06-26 16:56:25.132 [INFO ] [org.openhab.automation.script       ] - M true true
2022-06-26 16:56:28.550 [INFO ] [org.openhab.automation.script       ] - N false false
2022-06-26 16:56:28.504 [WARN ] [ore.internal.scheduler.SchedulerImpl] - Scheduled job failed and stopped                               
java.lang.IllegalStateException: Multi threaded access requested by thread Thread[OH-scheduler-22,5,main] but is not allowed for language(s) js.
        at com.oracle.truffle.polyglot.PolyglotEngineException.illegalState(PolyglotEngineException.java:129) ~[bundleFile:?] 
        at com.oracle.truffle.polyglot.PolyglotContextImpl.throwDeniedThreadAccess(PolyglotContextImpl.java:940) ~[bundleFile:?]
        at com.oracle.truffle.polyglot.PolyglotContextImpl.checkAllThreadAccesses(PolyglotContextImpl.java:799) ~[bundleFile:?]
        at com.oracle.truffle.polyglot.PolyglotContextImpl.enterThreadChanged(PolyglotContextImpl.java:629) ~[bundleFile:?]
        at com.oracle.truffle.polyglot.PolyglotEngineImpl.enterCached(PolyglotEngineImpl.java:1885) ~[bundleFile:?]
        at com.oracle.truffle.polyglot.HostToGuestRootNode.execute(HostToGuestRootNode.java:112) ~[bundleFile:?]
        at com.oracle.truffle.api.impl.DefaultCallTarget.callDirectOrIndirect(DefaultCallTarget.java:85) ~[bundleFile:?]
        at com.oracle.truffle.api.impl.DefaultCallTarget.call(DefaultCallTarget.java:102) ~[bundleFile:?]        
        at com.oracle.truffle.polyglot.PolyglotFunctionProxyHandler.invoke(PolyglotFunctionProxyHandler.java:154) ~[bundleFile:?]
        at com.sun.proxy.$Proxy701.apply(Unknown Source) ~[?:?]
        at org.openhab.core.model.script.actions.ScriptExecution.lambda$1(ScriptExecution.java:100) ~[bundleFile:?]
        at org.openhab.core.internal.scheduler.SchedulerImpl.lambda$12(SchedulerImpl.java:184) ~[?:?]
        at org.openhab.core.internal.scheduler.SchedulerImpl.lambda$1(SchedulerImpl.java:87) ~[?:?]
Caused by: com.oracle.truffle.api.TruffleStackTrace$LazyStackTrace
```
(the above is logged twice, since I have in two distinct .js files nearly identical code).  The second exception starts with `2022-06-26 16:56:28.522 [WARN ] [ore.internal.scheduler.SchedulerImpl] - Scheduled job failed and stopped` and is thus called after the `console.log('N')` code.

* The timestamps do not strictly increase in the `/var/log/openhab/openhab.log` file.
* There shall be no `Multi threaded access requested by thread Thread[OH-scheduler-22,5,main] but is not allowed for language(s) js.` exception.  JavaScript is single threaded, setTimeout() does not necessary add a new thread.
* The error message does not suggest which script/JSRule caused the exception.

florian-h05:
Fixes #13022.

## Description

This PR fixes the problem with scheduled jobs/timers failing because they try to access the context at the same time it is used by something different.

Furthermore, GraalJS is updated to a newer version.

## Testing

Please see the issue #13022 for testing.

To verify the threadsafety, use the following script in the `/automation/js` folder:

<details>
<summary>Test script</summary>

```javascript

const Thread = Java.type('java.lang.Thread');

// Create two rules with the extact same trigger to test whether rules are still threadsafe

const Thread = Java.type('java.lang.Thread');

// Create two rules with the extact same trigger to test whether rules are still threadsafe
rules.JSRule({
  name: 'Test Rule 1',
  triggers: triggers.ItemCommandTrigger('test_switch'),
  execute: (event) => {
    console.info('Rule 1 ran');
    // Make the rule run take longer to additionally test threadsafety
    Thread.sleep(5000);
  }
});

rules.JSRule({
  name: 'Test Rule 2',
  triggers: triggers.ItemCommandTrigger('test_switch'),
  execute: (event) => {
    console.info('Rule 2 ran');
    let myVar = 'Hello world';
    // Create 100 timers for the same timeout to ensure that timers/scheduled are now threadsafe
    for (let i = 0; i < 100; i++) {
      setTimeout(() => {
        console.info('setTimeout ' + i + ' completed successfully with message: ' + myVar);
        Thread.sleep(500);
      }, 100, myVar);
    }
    // Ensure that the rule is still running when the timers expire
    Thread.sleep(5000);
   // Test whether mutating passed in vars still works
    myVar = 'Hello mutation';
  }
});
```
</details>


florian-h05:
@jpg0 @tarag
This is my solution for #13022.

bundles/org.openhab.automation.jsscripting/src/main/java/org/openhab/automation/jsscripting/internal/OpenhabGraalJSScriptEngine.java
```diff
@@ -82,6 +86,9 @@
     public OpenhabGraalJSScriptEngine(@Nullable String injectionCode) {
         super(null); // delegate depends on fields not yet initialised, so we cannot set it immediately
         this.globalScript = GLOBAL_REQUIRE + (injectionCode != null ? injectionCode : "");
+        this.lock = new Object();
```
jpg0, line 89: You may want to initialise this at the declaration in the same line

jpg0:
This part of the PR lgtm.

My feeling on the other part in core though is really that the synchronisation doesn't belong in there. Instead I believe that the callback passed to the timer should hold the synchronisation logic and wrap the real callback. Pushing it into the core means that you lose the ability to fine tune the locking - for example, allow other lock types like readwriterlocks, wait timeouts, latching etc. TBH this would make your implementation easier as it wouldn't require changes in core.

florian-h05:
> My feeling on the other part in core though is really that the synchronisation doesn't belong in there. Instead I believe that the callback passed to the timer should hold the synchronisation logic and wrap the real callback.

Okay, thank you very much for the idea, that‘s something I actually did not think of — I will definitely try this.

florian-h05:
> Instead I believe that the callback passed to the timer should hold the synchronisation logic and wrap the real callback.

Just wrapping the callback that is passed to `ScriptExecution.createTimer` does not work, instead the wrapping needs to be applied around the callback that is passed to these methods. Therefore I need to modify either core or implement a replace for `createTimer` etc. - I am currently working at the replace.

jpg0:
> Just wrapping the callback that is passed to ScriptExecution.createTimer does not work

Why? I cannot understand why it would not work. You should wrap the original callback in a new callback that synchronises on the lock before executing the original callback, and pass the new one to createTimer.

florian-h05:
`createTimer` looks like the following:

```java
    public static Timer createTimer(@Nullable String identifier, ZonedDateTime instant, Procedure0 closure, @Nullable Object lock) {
        Scheduler scheduler = ScriptServiceUtil.getScheduler();

        return new TimerImpl(scheduler, instant, () -> {
            closure.apply(arg1);
        }, identifier);
    }
```

It takes a `org.eclipse.xtext.xbase.lib.Procedures.Procedure0` that can be applied calling `apply()`.
Wrapping the `Procedure0.apply()` call in synchronized and passing this to the TimerImpl (TimerImpl takes a `ScheduledFuture` for the callback) works, but how can I wrap the callback and pass it as a Procedure to createTimer?

florian-h05:
@jpg0 
See my latest commit, now the fix doesn't depend on core by implementing a replacement for the timer creation methods of `ScriptExecution` and wrapping there.

But there are two things I don't like:
- I need the `TimerImpl` class from oh-core, but it's package internal and therefore I can not access it. To get a working PR, I just copied this class to the addon, which is not the good way.
- I am using `org.eclipse.xtext.xbase.lib` which is also used in the `ScriptExecution` class of core, but as I have no idea if or how I could access the one from core, I added xtest as dependency to the addon. For this, I have an issue open: https://github.com/openhab/openhab-core/issues/3128.

jpg0:
@florian-h05 the way to wrap would be something like this:

```
createTimer(... new Procedure0() { 
   override void apply() {
      synchronized(lock) {
        originalCallback.apply();
      }
  }
})

```

I agree that it's not ideal having to rely on `Procedure0`. _Really_ that timer class should be accepting a `Runnable` rather than a `Procedure0`.

florian-h05:
> I agree that it's not ideal having to rely on Procedure0

I opened an issue at core and Kai answered, addons should not use xtext (even core shouldnt use it) and therefore we can not use `Procedure0`. I am not sure if there is any way to wrap the callback without having access to `Procedure0`.

jpg0:
> I opened an issue at core

I saw and commented :-) there should be _some_ way of doing it

florian-h05:
> I saw and commented :-)

I saw, thanks!

> there should be some way of doing it

Yes. FYI, I am close to have a solution that is not relying on any core change and not relying on xtext.

florian-h05:
@jpg0 
FYI: I am going on a trip tomorrow (24th Oct) until friday (28th Oct), so my availability will be very limited.

florian-h05:
@jpg0 
From my side, this is ready to merge.
Would be nice if you could review it.

A few points to say:
- This doesn’t depend on xtext.
- This doesn’t depend on any core change, so it can be a quick fix.
- I copied the `TimerImpl` class from core as I need it in `ThreadsafeTimers`, but it is not accessible from outside the package.
- I’m not sure if there are multithreading problems occurring in relation with cancelling a timer, if that is so, we benefit from our own `TimerImpl` class (that is copied from core).
- I upgraded GraalJS from 21.3.0 to 22.3.0
- I will provide a PR to the JavaScript library to incorporate the fix from this PR (so that the library is using it).

jpg0:
Ok I had another look at the openHAB core classes. I assumed that the scheduling was handled by openHAB specific classes, however the scheduler is the standard Java one. I guess what this means is that you've got a lot of code that is extra and not really needed. You don't need a Timer at all, all you need to do it submit the `Runnable` with `scheduler.schedule`. This would remove most of `ThreadsafeTimers`, if not all of it.

As for the actual logic, I think it's all fine. Personally I'd prefer that you remove the Timer classes as it's just unnecessary complexity, but I don't have permission to approve the PR, so I think that you'll need to get @digitaldan to make the call.

florian-h05:
> Personally I'd prefer that you remove the Timer classes as it's just unnecessary complexity

I was also thinking a bit about them, but we can’t remove them easily because many scripts are relying on the Timer API that has some functionality over the Java scheduler (e.g. `reschedule`) and removing it would therefore break existing installations.

> I don't have permission to approve the PR

I wonder why you are no member of @openhab and the contributors team, AFAIK you normally should be because you contributed this addon.
@kaikreuzer Can you invite @jpg0 to the openhab org and the contributors team?

> so I think that you'll need to get @digitaldan to make the call.

@digitaldan @jlaur @kaikreuzer 
May any of you approve? Please have a look at the previous comment from @jpg0.

jpg0:
> but we can’t remove them easily because many scripts are relying on the Timer API

I am not suggesting removing them entirely, but that your implementation does not need to use them. For example `setTimeout` does not require rescheduling functionality.

florian-h05:
> For example setTimeout does not require rescheduling functionality.

Looking at the official JS implementation yes this is true, but see https://www.openhab.org/addons/automation/jsscripting/#timers. The docs claim that `setTimeout` and `setTimer` both return a `Timer`, so changing the implementation of `setTimeout` could also break some installations.

jpg0:
@florian-h05 Personally I regard that implementation (and hence associated docs) as a mistake:

- It means that JS written in openHAB may be not compatible with standard JS, so (for example) if we do get the option to move to standard JS (such as Graal implement `setTimeout`) then it will break JS code
- It exposes Timers to openHAB users, when it seems from @kaikreuzer like they are planned to be removed (TimerImpl is the only implementation of Timer afaics)

Saying that, I understand the reluctance to revisit that decision here, so I am fine either way.

tarag:
Thanks for working on this fix @florian-h05 , hopefully it can be merged in the next milestone.

florian-h05:
@jpg0
Your strong arguments convinced me to rethink the current Timer implementation in the addon.

I would like to get this PR merged and then discuss the implementation in https://github.com/openhab/openhab-addons/issues/13594.

bundles/org.openhab.automation.jsscripting/src/main/java/org/openhab/automation/jsscripting/internal/threading/ThreadsafeTimers.java
```diff
@@ -0,0 +1,137 @@
...
+import org.openhab.core.model.script.actions.Timer;
+import org.openhab.core.scheduler.ScheduledCompletableFuture;
+import org.openhab.core.scheduler.Scheduler;
+import org.openhab.core.scheduler.SchedulerRunnable;
+
+/**
+ * A replacement for the timer functionality of {@link org.openhab.core.model.script.actions.ScriptExecution
+ * ScriptExecution} which controls multithreaded execution access to the single-threaded GraalJS Contexts.
+ *
+ * @author Florian Hotze
```
kaikreuzer, line 30:
```suggestion
 * @author Florian Hotze - Initial contribution
```

bundles/org.openhab.automation.jsscripting/src/main/java/org/openhab/automation/jsscripting/internal/threading/ThreadsafeTimers.java
```diff
@@ -0,0 +1,137 @@
...
+import org.eclipse.jdt.annotation.Nullable;
+import org.openhab.core.model.script.ScriptServiceUtil;
+import org.openhab.core.model.script.actions.Timer;
+import org.openhab.core.scheduler.ScheduledCompletableFuture;
+import org.openhab.core.scheduler.Scheduler;
+import org.openhab.core.scheduler.SchedulerRunnable;
+
+/**
+ * A replacement for the timer functionality of {@link org.openhab.core.model.script.actions.ScriptExecution
+ * ScriptExecution} which controls multithreaded execution access to the single-threaded GraalJS Contexts.
```
kaikreuzer, line 28:
```suggestion
 * ScriptExecution} which controls multithreaded execution access to the single-threaded GraalJS contexts.
```

bundles/org.openhab.automation.jsscripting/src/main/java/org/openhab/automation/jsscripting/internal/fs/watch/JSScriptFileWatcher.java
```diff
@@ -31,7 +32,7 @@
  */
 public class JSScriptFileWatcher extends ScriptFileWatcher {
     private static final String FILE_DIRECTORY = "automation" + File.separator + "js";
-    private static final String IGNORE_DIR_NAME = "node_modules";
+    // private static final String IGNORE_DIR_NAME = "node_modules";
```
kaikreuzer, line 35: Can this line be removed?

kaikreuzer (changes requested):
Some small review comments

bundles/org.openhab.automation.jsscripting/src/main/java/org/openhab/automation/jsscripting/internal/fs/watch/JSScriptFileWatcher.java
florian-h05: Yes

tarag:
> If you want to, I can provide you an artifact so you don't have to wait for the next milestone and verify yourself that the fix works.

Thanks but I don't think I can easily deploy this where I need it the most. Were you able to check the concern about deadlock situation I discussed ?

florian-h05:
> Were you able to check the concern about deadlock situation I discussed ?

I haven't tested your code snippet from the issue, but the code snippet should work fine:

```javascript

setTimeout(() => {
    items.getItem('myItem').postUpdate('ON')
   ... more stuff ...
}, 10000)

rules.JSRule({
    ...
    triggers: [
        global.triggers.ItemStateChange('myItem'),
    ],
...
})
```

I'll try why there is no deadlock:

- Set timeout expires in 10 seconds. The time it expires, the access to the JS context is locked and the update is posted. After the code inside `setTimeout` finished, the context is unlocked.
- When the Item receives an update before or after `setTimeout` runs, the rule is run right now. While the rule is run, the context is locked.
- If the rule and `setTimeout` try to run at the exact same time, one has to wait until the other one executed its code.

florian-h05:
@kaikreuzer I addressed your review.

kaikreuzer (approved):

florian-h05:
@tarag 
PR has been merged, so the fix will be published with the next milestone.

Regarding the 200€ bounty you placed: Please contact me via mail florianh_dev@icloud.com so that we can select a payment method (e.g. PayPal). 
Feel free to test the fix before paying!

Issue #13586, opened 2022-10-23
[hue] Binding broken until you update the new protocol setting to "http" (bridge V1)


lolodomo:
I just installed snapshot 3143 containing the PR #11842 merged yesterday.
The binding does no more initialize properly:
```
09:32:18.362 [ERROR] [connection.HueTlsTrustManagerProvider] - An unexpected exception occurred - returning a TrustAllTrustManager: java.net.ConnectException: Connexion refusée (Connection refused)
org.openhab.core.io.net.http.PEMTrustManager$CertificateInstantiationException: java.net.ConnectException: Connexion refusée (Connection refused)
        at org.openhab.core.io.net.http.PEMTrustManager.getPEMCertificateFromServer(PEMTrustManager.java:220) ~[bundleFile:?]
        at org.openhab.core.io.net.http.PEMTrustManager.getInstanceFromServer(PEMTrustManager.java:131) ~[bundleFile:?]
        at org.openhab.core.io.net.http.PEMTrustManager.getInstanceFromServer(PEMTrustManager.java:116) ~[bundleFile:?]
        at org.openhab.binding.hue.internal.connection.HueTlsTrustManagerProvider.getTrustManager(HueTlsTrustManagerProvider.java:63) [bundleFile:?]
        at org.openhab.core.io.net.http.internal.ExtensibleTrustManagerImpl.addTlsTrustManagerProvider(ExtensibleTrustManagerImpl.java:220) [bundleFile:?]
        at jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method) ~[?:?]
        at jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62) ~[?:?]
        at jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43) ~[?:?]
        at java.lang.reflect.Method.invoke(Method.java:566) ~[?:?]
        at org.apache.felix.scr.impl.inject.methods.BaseMethod.invokeMethod(BaseMethod.java:244) [bundleFile:?]
        at org.apache.felix.scr.impl.inject.methods.BaseMethod.access$500(BaseMethod.java:41) [bundleFile:?]
        at org.apache.felix.scr.impl.inject.methods.BaseMethod$Resolved.invoke(BaseMethod.java:685) [bundleFile:?]
[...]

PR #13592, opened 2022-10-23
[hue] Reinitialize bridge properties after configuration change

jlaur:
After saving bridge configuration and reinitializing the bridge, all dynamic properties (serial number, firmware version etc.) were lost:
image

Related to #13586


lolodomo (approved):
LGTN

PR #13587, opened 2022-10-23
[netatmo] Console command completion

lolodomo:
Signed-off-by: Laurent Garnier <lg.hc@free.fr>

Depends on openhab/openhab-core#3111

PR #13601, opened 2022-10-24
[meater] Add console extension for showing IDs

### LINKED ISSUE #13599: [meater] Console command for identifying probe IDs
In order to support file-based configuration without the need for prior UI configuration using discovery, a console command could be added. See #13555 for a similar solution for Netatmo.

jlaur:
Example session:
```
openhab> openhab:meater showIds
API bridge: MEATER Cloud REST API
    - ID: aaa (ambient temperature: 22.1)
    - ID: bbb (ambient temperature: 22.1)
    - ID: ccc (ambient temperature: 21.9)
    - ID: ddd (ambient temperature: 21.9)
```

Fixes #13599
Depends on openhab/openhab-core#3111

### DIFF
--- a/bundles/org.openhab.binding.meater/README.md
+++ b/bundles/org.openhab.binding.meater/README.md
@@ -18,14 +18,18 @@ This binding supports the following thing types:
 
 ## Discovery
 
-The preferred way of adding MEATER probe(s) since the probe IDs are not easily found.
+The binding supports auto-discovery of all MEATER probe(s) belonging to the configured cloud API account.
 
 **NOTE**: For The Original MEATER and MEATER Plus you need to have your MEATER app running and the MEATER probe(s) must connected to the cloud (out of the charger box) before you start the discovery.
 
-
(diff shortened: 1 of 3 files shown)


lolodomo:
Look at #13587 to add command completion.

jannegpriv (approved):
Nice addition!

jlaur:
@lolodomo - I needed latest snapshot to test this, which I just now did.

Small side note: When implementing `ConsoleCommandCompleter`, the console extension is no longer compatible with older core versions, for example M3. So we could either merge this and your PR's now, or we could wait until just before M4 will be built. However, since it only breaks the console extension functionality, I guess maybe not that important? It's kind of obvious, but realized it when I tried to test #13587 on my production system running M3.

lolodomo (approved):
LGTM, thank you

lolodomo:
> So we could either merge this and your PR's now, or we could wait until just before M4 will be built.

No need to wait. The snapshot built everyday is coherent. Using recent binding builds in an old OH version should be done only by developers who know what they are doing and understand the risk to have a binding not working fully as it should.

PR #13604, opened 2022-10-25
[mqtt] set the proper unit in the state description

ccutrer:
So that other pieces of openhab can know what unit it's going to be, without it having a value yet. Importantly, any necessary conversion that need to be applied to the other portion of the state description - min, max, and step.


lolodomo (approved):
LGTM

PR #13606, opened 2022-10-26
[ipcamera] Fix multiple WARNs when HIK camera does not support alarm inputs

Skinah:
This fixes a bug reported on the forum here:
https://community.openhab.org/t/ipcamera-new-ip-camera-binding/42771/2796

It was caused by the lowPriorityRequests having only 1 element. When the camera did not support any alarm inputs (common for a camera to have none) the binding would remove the request, then later on as lowPriorityRequests was empty it would fetch the list again which added it back in. This caused an endless WARN messages in the logs.

Jar can be downloaded from:


lolodomo (approved):
LGTM

PR #13618, opened 2022-10-29
[lgwebos] Actions: sendButton updated, sendRCButton removed, sendKeyboard added

### LINKED ISSUE #13600: [lgwebos] cannot send ENTER via sendButton
## Expected Behavior
I want to activate Multi View with a few button presses.
But the last one (ENTER) does nothing.

## Current Behavior
```
val actions = getActions("lgwebos","lgwebos:WebOSTV:tv1")
TV1_Application.sendCommand("com.webos.app.multiviewsettings")
Thread::sleep(2000)
actions.sendButton("DOWN")
Thread::sleep(2000)
actions.sendButton("ENTER")
```

The Multi View settings open, the cursor goes one item down, but then nothing happenes on screen.

```
2022-10-24 21:32:22.479 [DEBUG] [ebos.internal.handler.LGWebOSHandler] - handleCommand(lgwebos:WebOSTV:tv1:appLauncher,com.webos.app.multiviewsettings)
2022-10-24 21:32:22.516 [DEBUG] [.lgwebos.internal.BaseChannelHandler] - LauncherApplication received: org.openhab.binding.lgwebos.internal.handler.core.LaunchSession@98f68cd.
2022-10-24 21:32:24.481 [DEBUG] [nternal.handler.LGWebOSTVMouseSocket] - Connecting to: ws://192.168.1.90:3000/resources/c9168d0722xxxxxxx810a909d90/netinput.pointer.sock
2022-10-24 21:32:24.505 [DEBUG] [nternal.handler.LGWebOSTVMouseSocket] - WebSocket Connected to: /192.168.1.90
2022-10-24 21:32:24.505 [DEBUG] [nternal.handler.LGWebOSTVMouseSocket] - Message [out]: type:button
name:DOWN


2022-10-24 21:32:24.506 [DEBUG] [nternal.handler.LGWebOSTVMouseSocket] - WebSocket Closed - Code: 1006, Reason: Failed to open local endpoint
```

The same with `OK` instead of `ENTER`.
It looks like `OK` is not supported.

```
2022-10-24 21:34:43.580 [DEBUG] [ebos.internal.handler.LGWebOSHandler] - handleCommand(lgwebos:WebOSTV:tv1:appLauncher,com.webos.app.multiviewsettings)
2022-10-24 21:34:43.606 [DEBUG] [.lgwebos.internal.BaseChannelHandler] - LauncherApplication received: org.openhab.binding.lgwebos.internal.handler.core.LaunchSession@342babc0.
2022-10-24 21:34:45.581 [DEBUG] [nternal.handler.LGWebOSTVMouseSocket] - Connecting to: ws://192.168.1.90:3000/resources/c9168d07223f2b2b5590afca2f86f9810a909d90/netinput.pointer.sock
2022-10-24 21:34:45.603 [DEBUG] [nternal.handler.LGWebOSTVMouseSocket] - WebSocket Connected to: /192.168.1.90
2022-10-24 21:34:45.603 [DEBUG] [nternal.handler.LGWebOSTVMouseSocket] - Message [out]: type:button
name:DOWN


2022-10-24 21:34:45.603 [DEBUG] [nternal.handler.LGWebOSTVMouseSocket] - WebSocket Closed - Code: 1006, Reason: Failed to open local endpoint
2022-10-24 21:34:47.581 [DEBUG] [nternal.handler.LGWebOSTVMouseSocket] - Connecting to: ws://192.168.1.90:3000/resources/c9168d07223f2b2b5590afca2f86f9810a909d90/netinput.pointer.sock
2022-10-24 21:34:47.602 [DEBUG] [nternal.handler.LGWebOSTVMouseSocket] - WebSocket Connected to: /192.168.1.90
2022-10-24 21:34:47.603 [DEBUG] [nternal.handler.LGWebOSTVMouseSocket] - Message [out]: type:click


2022-10-24 21:34:47.603 [DEBUG] [nternal.handler.LGWebOSTVMouseSocket] - WebSocket Closed - Code: 1006, Reason: Failed to open local endpoint
```

## Possible Solution
n/a

see [Current Behavior]

## Context
see [## Expected Behavior]

## Your Environment
* openHAB 3.2.0
* Docker
* LG OLED77C29LD
* LG software version: 3.20.70
* webOS: 7.2.0-32

lolodomo:
Fix #13600


lolodomo:
For compatibility reasons, I did not change the handling of the original ENTER value, I imagine it is working with certain TVs (in particular with the TV the original author was using when implementing sendButton action). So I added the support of value ENTER_ALT which leads in fact to a similar behaviour as sending ENTER to the rcButton channel.

bundles/org.openhab.binding.lgwebos/README.md
```diff
@@ -314,9 +314,9 @@ Sends a button press event to a WebOS device.
 
 Parameters:
 
-| Name    | Description                                                            |
-|---------|------------------------------------------------------------------------|
-| button  | Can be one of UP, DOWN, LEFT, RIGHT, BACK, DELETE, ENTER, HOME, or OK  |
+| Name    | Description                                                                            |
+|---------|----------------------------------------------------------------------------------------|
+| button  | Can be one of UP, DOWN, LEFT, RIGHT, BACK, EXIT, DELETE, ENTER, ENTER_ALT, HOME, or OK |
```
jlaur, line 319: If possible, I think we should document the difference between `ENTER `and `ENTER_ALT`. As I understand it, in technical terms, `ENTER` will use a special URL in context of a remote keyboard subscription:
https://github.com/openhab/openhab-addons/blob/99e309e991beadaae37cee09b03c711f88c7f562/bundles/org.openhab.binding.lgwebos/src/main/java/org/openhab/binding/lgwebos/internal/handler/LGWebOSTVKeyboardInput.java#L97-L99

whereas `ENTER_ALT` will use a different (mouse) API?

If we know when each of these should be used, we could document that. Even if we only know that `ENTER` works for older TV's without knowing exactly what that means.

It could also be different purposes? I could only find this in some documentation:

```
void sendEnter () Send enter key to the current text field.
Related capabilities:
• TextInputControl.Send.Enter

void sendDelete () Send delete event to the current text field.
Related capabilities:
• TextInputControl.Send.Delete
```

So looks like intention is to send ENTER to a focused text field. Is that the same as the new ENTER_ALT? Or is it a more "global" kind of SELECT/ENTER?

I have an LG webOS TV also, so could maybe do some tests tomorrow. I know, more questions than answers. But I think we need a bit more understanding to provide documentation explaining this.

lolodomo: Unfortunately, I have no answers to your questions. 
@sprehn as author of the binding could probably explain what he implemented but I have the feeling he is no more participating here.

lolodomo: What I can tell you is that the keyboard API is mainly used by the sendText binding action which allows to send/display a message on the TV. I have no idea what is exactly the purpose of ENTER and DELETE in that context, even if I can imagine this might be to validate or erase a previously filled text. I don't understand what could be the usage in the context of the binding. Maybe to erase the popup on TV ?

The mouse API is apparently the API to navigate in the TV system like you do with your TV remote control.

Maybe the way ENTER is handled in the sendButton binding action has simply no usage for us ?

But it can be interesting to check the ENTER value with the sendButton action when using the sendText action. I do not remember if the popup disappears automatically after a delay.

lolodomo: Oups, confustion between sendText and showToast ! So in fact, I have no use case in mind.
Maybe the keyboard API is when the OS is asking to fill a text ? Like a username or a password ?
Maybe ENTER as value for sendButton action should have been named ENTER_TEXT.
Worst I believe sendButton should have not handled the keyboard API at all, the action description in the documentation is "Sends a button press event to a WebOS device".
So maybe I should just replace the old handling of ENTER by the new ONE.

lolodomo: I am going to rename ENTER_ALT into ENTER_NAV and mention that this matches the remote control button in the center of the navigation group.

lolodomo: Done

jlaur: Just quick status from my side: I was able to do something with ENTER and DELETE when on-screen keyboard is displayed with focus in the text field:
- Sending ENTER will remove the keyboard.
- Sending DELETE will delete the last character.

I'm not sure yet if they have any other effects, but at least they do something.

lolodomo: Ok, I can add a comment in the documentation for DELETE and update the one for ENTER.

lolodomo: Done

lolodomo: @jlaur : there is also the OK value, I don't know what is its usage, it does nothing when I send it while an item is selected at screen.

bundles/org.openhab.binding.lgwebos/src/main/java/org/openhab/binding/lgwebos/internal/handler/LGWebOSTVKeyboardInput.java
```diff
@@ -120,6 +123,7 @@ private void sendData() {
             payload.addProperty("text", sb.toString());
             payload.addProperty("replace", 0);
         }
+        logger.debug("Message [out]: {} {}", uri, payload.toString());
```
jlaur, line 126: This is redundant:
```
2022-10-30 21:08:43.430 [DEBUG] [ernal.handler.LGWebOSTVKeyboardInput] - Message [out]: ssap://com.webos.service.ime/sendEnterKey null
2022-10-30 21:08:43.439 [TRACE] [bos.internal.handler.LGWebOSTVSocket] - Message [out]: {"type":"request","id":8,"uri":"ssap://com.webos.service.ime/sendEnterKey","payload":null}
```

bundles/org.openhab.binding.lgwebos/src/main/java/org/openhab/binding/lgwebos/internal/action/LGWebOSActions.java
```diff
@@ -88,8 +88,10 @@ private enum Button {
         LEFT,
         RIGHT,
         BACK,
+        EXIT,
         DELETE,
         ENTER,
+        ENTER_NAV,
```
jlaur, line 94: Perhaps this one should be named `ENTER` and the old one renamed to `ENTER_KEYBOARD` or something like that? The old one seems much less useful.

Actually I'm now wondering if instead we should preserve only the actions `DELETE` and `ENTER`, as all the other ones which maps to mouse actions are duplicated functionality since the same can be accomplished with channel `rcButton`. Of course the documentation should be updated to explain this. It seems like we now have a naming conflict when trying to fit two different things into one which doesn't need to be fitted as functionality is redundant?

Or maybe I'm missing something, but both solutions seems to end up in same place, `executeMouse()`?

Another approach could be separating this into two actions, `sendButton` and `sendKey` (or similar). The last one would accept only `DELETE` and `ENTER`, and first one could take all the others - without any naming conflicts. We could even consider skipping the validation, or adding an optional parameter for that, so that it would be possible to discover new buttons and use them.

bundles/org.openhab.binding.lgwebos/src/main/java/org/openhab/binding/lgwebos/internal/action/LGWebOSActions.java
```diff
@@ -200,12 +202,18 @@ public void sendButton(
...
+                    break;
                 case DELETE:
                     getConnectedSocket().ifPresent(control -> control.sendDelete());
                     break;
                 case ENTER:
                     getConnectedSocket().ifPresent(control -> control.sendEnter());
                     break;
+                case ENTER_NAV:
+                    getConnectedSocket().ifPresent(control -> control.executeMouse(s -> s.button(ButtonType.ENTER)));
+                    break;
```
jlaur, line 216: Not your code, but in line 218 below `"HOME"` could be replaced by `ButtonType.HOME`.

jlaur (changes requested):
Sorry, another iteration - I hope you are not in a rush. :-) The more I think of this, the more I'm convinced something is wrong, and we should now consider the best approach and avoid the risk of introducing technical debt which would require more breaking changes later if we want to get rid of it again.

In my biggest comment for ENTER_NAV I'm now leaning towards the last suggestion I made, but I'm curious to hear your thoughts.

bundles/org.openhab.binding.lgwebos/src/main/java/org/openhab/binding/lgwebos/internal/handler/LGWebOSTVKeyboardInput.java
lolodomo: This is the equivalent of the DEBUG log you got when the mouse API is used. I just added it for consistency and to have something logged when only DEBUG level is set.

bundles/org.openhab.binding.lgwebos/src/main/java/org/openhab/binding/lgwebos/internal/action/LGWebOSActions.java
lolodomo: My intention was to break nothing.

> Another approach could be separating this into two actions, sendButton and sendKey (or similar). The last one would accept only DELETE and ENTER, and first one could take all the others - without any naming conflicts. We could even consider skipping the validation, or adding an optional parameter for that, so that it would be possible to discover new buttons and use them.

Now if we decide to break something, your proposal looks good to me.

lolodomo: I now discover another **undocumented** action named `sendRCButton` which only uses the mouse API.
In fact, `sendButton` is an extension of `sendRCButton` including keyboard ENTER + DELETE and OK.

The action that is the equivalent of the rcButton channel is in fact `sendRCButton` except that the action is restricted to a limited choice of 8 predefined buttons while you can send any command value to the channel.

So `sendButton` could now do what `sendRCButton` is doing but without the restriction, and keeping the special case of OK. `sendRCButton` could be removed (was not documented). sendKey could be added to cover the keyboard API (ENTER+DELETE).

lolodomo: Done

bundles/org.openhab.binding.lgwebos/src/main/java/org/openhab/binding/lgwebos/internal/action/LGWebOSActions.java
lolodomo: I removed the restriction on values.

bundles/org.openhab.binding.lgwebos/src/main/java/org/openhab/binding/lgwebos/internal/handler/LGWebOSTVKeyboardInput.java
jlaur: Hmm. So maybe it should be conditional of `!logger.isTraceEnabled()` to avoid redundant logging? Although it would be kind of messy as the logging for **DEBUG** and **TRACE** would be handled in different places without any obvious correlation. Depending on state, the message could still be skipped after this:
https://github.com/openhab/openhab-addons/blob/e528564bb6f401e297649a3cfe2ad868bbdfcb8e/bundles/org.openhab.binding.lgwebos/src/main/java/org/openhab/binding/lgwebos/internal/handler/LGWebOSTVSocket.java#L331-L355

And inner logging:
https://github.com/openhab/openhab-addons/blob/e528564bb6f401e297649a3cfe2ad868bbdfcb8e/bundles/org.openhab.binding.lgwebos/src/main/java/org/openhab/binding/lgwebos/internal/handler/LGWebOSTVSocket.java#L374-L390

Do we really want to log payload when **TRACE** is not enabled? I believe these paths are pretty covered with appropriate logging, so maybe we should revisit mouse socket logging instead?

lolodomo: I have undone the adding of logs in that class.

jlaur (approved):
LGTM - thanks!

PR #13620, opened 2022-10-29
[netatmo] No restriction on max value for rain quantity

### LINKED ISSUE #13619: [netatmo] Max of 150 mm is inappropriate for sum_rain measurement channel
Each measurement channel has a min and a max value.
For sum-rain, the max value is set to 150 mm.
No reason to define such max value.
In practice, you can have a cumulated rain exceeding 150 mm.
@clinique for information.

Issue was exposed in the community forum.

lolodomo:
Fix #13619

### DIFF
--- a/bundles/org.openhab.binding.netatmo/src/main/java/org/openhab/binding/netatmo/internal/api/data/NetatmoConstants.java
+++ b/bundles/org.openhab.binding.netatmo/src/main/java/org/openhab/binding/netatmo/internal/api/data/NetatmoConstants.java
@@ -85,7 +85,7 @@ public class NetatmoConstants {
         CO2(0, 5000, 50, Units.PARTS_PER_MILLION, "co2", "measure", true),
         NOISE(35, 120, 1, Units.DECIBEL, "noise", "measure", true),
         RAIN_QUANTITY(0, Double.MAX_VALUE, 0.1, MILLI(SIUnits.METRE), "sum_rain", "sum_rain", false),
-        RAIN_INTENSITY(0, Double.MAX_VALUE, 0.1, Units.MILLIMETRE_PER_HOUR, "", "", false),
+        RAIN_INTENSITY(0, 150, 0.1, Units.MILLIMETRE_PER_HOUR, "", "", false),
         WIND_SPEED(0, 160, 1.8, SIUnits.KILOMETRE_PER_HOUR, "", "", false),
         WIND_ANGLE(0, 360, 5, Units.DEGREE_ANGLE, "", "", false),
         HUMIDITY(0, 100, 3, Units.PERCENT, "hum", "measure", true);
--- a/bundles/org.openhab.binding.netatmo/src/main/java/org/openhab/binding/netatmo/internal/api/data/NetatmoConstants.java
+++ b/bundles/org.openhab.binding.netatmo/src/main/java/org/openhab/binding/netatmo/internal/api/data/NetatmoConstants.java
@@ -84,8 +84,8 @@ public class NetatmoConstants {
         PRESSURE(260, 1260, 0.1, HECTO(SIUnits.PASCAL), "pressure", "measure", true),
         CO2(0, 5000, 50, Units.PARTS_PER_MILLION, "co2", "measure", true),
         NOISE(35, 120, 1, Units.DECIBEL, "noise", "measure", true),
-        RAIN_QUANTITY(0, 150, 0.1, MILLI(SIUnits.METRE), "sum_rain", "sum_rain", false),
-        RAIN_INTENSITY(0, 150, 0.1, Units.MILLIMETRE_PER_HOUR, "", "", false),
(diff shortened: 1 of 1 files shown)


bundles/org.openhab.binding.netatmo/src/main/java/org/openhab/binding/netatmo/internal/api/data/NetatmoConstants.java
```diff
@@ -84,8 +84,8 @@ public enum MeasureClass {
         PRESSURE(260, 1260, 0.1, HECTO(SIUnits.PASCAL), "pressure", "measure", true),
         CO2(0, 5000, 50, Units.PARTS_PER_MILLION, "co2", "measure", true),
         NOISE(35, 120, 1, Units.DECIBEL, "noise", "measure", true),
-        RAIN_QUANTITY(0, 150, 0.1, MILLI(SIUnits.METRE), "sum_rain", "sum_rain", false),
-        RAIN_INTENSITY(0, 150, 0.1, Units.MILLIMETRE_PER_HOUR, "", "", false),
+        RAIN_QUANTITY(0, Double.MAX_VALUE, 0.1, MILLI(SIUnits.METRE), "sum_rain", "sum_rain", false),
+        RAIN_INTENSITY(0, Double.MAX_VALUE, 0.1, Units.MILLIMETRE_PER_HOUR, "", "", false),
```
jlaur, line 88: I'm curious: Where did these values come from? Purpose is some form of sanity checks or limitation or possible values according to physics?

I'm wondering if other measurements should also be adapted. For example, inside temperature could be below 0°C if an inside module would be used outside. And outside could also be below -40 °C in some parts of the world. And if it's 65 °C outside, I believe it could possibly also be more than 50 °C inside? We could limit temperatures to not be below 89.2 °C (lowest natural ever recorded) if just sanity checks or -273.15 °C according to physics.

For CO2 ppm it could easily be higher than 5000. Exposure to levels exceeding 40000 ppm may lead to serious oxygen deprivation resulting in permanent brain damage, coma, even death. So it would be nice if values between 5000 and 40000 could be supported, so users could use this to be warned before that. :-)

Point/question is simply where the values come from - Netatmo API description, for example? Or and other measurements should also be adapted.

lolodomo: The problem is especially for rain quantity as a user can request the sum of rain for one month. Of course 150 mm for a month is clearly not a valid max limit.
For other measurements, the channel is not a sum but the min or max value on a period. If @clinique used the valid range that a Netatmo device can deliver, this should be OK.
But in all cases, @clinique should confirm where are coming these min/max values from.

jlaur: Just for the record, awaiting feedback from @clinique which could make it possible to settle not only RAIN_QUANTITY/RAIN_INTENSITY, but perhaps also some of the other strange-looking boundaries.

clinique: The values are physical restrictions of sensors as given by Netatmo specifications. Agree that max can raise problem when used in queries.

lolodomo: I believe the only measurement that is a sum is the RAIN_QUANTITY. All other measurements are only min or max.
@clinique : do you confirm ?
For RAIN_INTENSITY, maybe we could keep 150 mm as max ?

clinique: > I believe the only measurement that is a sum is the RAIN_QUANTITY. All other measurements are only min or max.
> @clinique : do you confirm ?
> For RAIN_INTENSITY, maybe we could keep 150 mm as max ?

Yes, I do

lolodomo:
@jlaur : can you merge please ?

bundles/org.openhab.binding.netatmo/src/main/java/org/openhab/binding/netatmo/internal/api/data/NetatmoConstants.java
jlaur: So should `RAIN_INTENSITY` be kept as 150?

lolodomo: I have undone the change for rain intensity.
So only rain quantity is now changed.

jlaur (approved):
LGTM

PR #13623, opened 2022-10-30
[jsscripting] Reimplement timer polyfills to conform standard JS

### LINKED ISSUE #13594: [jsscripting] Discuss the implementation of timers
## Current situation

The `setTimeout` and `setInterval` polyfill methods returns an openHAB `Timer` which **does not conform** to standard JS.

## The problem

1. If we get the option to move to standard JS (when GraalJS implements those methods and we don't need the polyfills) this will break existing JS code.
2. Timers might be removed from openHAB at sometime:
> It exposes Timers to openHAB users, when it seems from @kaikreuzer like they are planned to be removed

@jpg0 
Continuing the discussion from https://github.com/openhab/openhab-addons/pull/13582.

florian-h05:
Fixes #13594.

## Description

This PR reimplements the polyfills for the standard JS timers (`setTimeout` and `setInterval`) in the Java layer of the binding to conform standard JS.

This brings several improvements:
- Usage of `TimerImpl` from core is no longer required; instead the scheduler is accessed directly
- Mapping of ScheduledFutures to unique Integers to conform standard JS that returns a `timeoutId` or `intervalId` when the according funtion is run. FYI: Standard JS shares the same pool of integers for both, the binding does therefore the same.
- Scheduled jobs created by the polyfills are named by using the loggername, the type (`timeout` or `interval`) and the id (an Integer).
- All scheduled jobs created by the polyfills are automatically cancelled when the rule is unloaded to avoid having unmanagable scheduled jobs running endless (until restart of openHAB).

## Breaking Changes

- The polyfills do not return an openHAB Timer anymore, instead they return an interger id as standard JS does.
- Other ways of scheduling timers (`createTimer`, `createTimerWithArgument`) are not recommended to be used anymore as they are not threadsafe. But it is still possible to access them via the `ScriptExecution` actions from core.
- I propose to log a warning when someone tries to access those timer creation methods via the `ScriptExecution` action from the openhab-js library.

Therefore, the polyfills for the standard JS timer methods should be the only way taken for creating timers in JS Scripting!

## Testing

```javascript

const Thread = Java.type('java.lang.Thread');

rules.JSRule({
  name: 'Test Rule',
  triggers: triggers.ItemCommandTrigger('test_switch'),
  execute: (event) => {
    console.info('Rule 2 ran');
    // console.loggerName = 'org.openhab.automation.js.1123';
    let myVar = 'Hello world';

    // Create timers for the same timeout to ensure that timers/scheduled jobs are threadsafe
    for (let i = 0; i < 15; i++) {
      setTimeout(() => {
        console.info('setTimeout ' + i + ' completed successfully with message: ' + myVar);
        Thread.sleep(500);
      }, 2000, myVar);
    }

    let id;
    if (true) {
      // Create interval to verify functionality of setInterval
      id = setInterval(() => {
        console.info('setInterval completed successfully');
      }, 100, myVar, myVar);
    }

    if (true) {
      // Make a scheduled job fail to check whether naming jobs works
      setTimeout(() => {
        console.info(uninizializedVar);
      }, 1000);
    }

    if (true) {
      // Cancel interval with a timeout by using the intervalId
      setTimeout(() => {
        console.info('Attempting to clear setInterval...');
        clearTimeout(id);
      }, 5000, myVar, myVar);
    }

    myVar = 'Hello mutation';
  }
});

```

### DIFF
--- a/bundles/org.openhab.automation.jsscripting/src/main/java/org/openhab/automation/jsscripting/internal/OpenhabGraalJSScriptEngine.java
+++ b/bundles/org.openhab.automation.jsscripting/src/main/java/org/openhab/automation/jsscripting/internal/OpenhabGraalJSScriptEngine.java
@@ -69,6 +69,8 @@ public class OpenhabGraalJSScriptEngine
     // final CommonJS search path for our library
     private static final Path NODE_DIR = Paths.get("node_modules");
 
+    // shared lock object for synchronization of multi-thread access
+    public final Object lock = new Object();
     private final JSRuntimeFeatures jsRuntimeFeatures = new JSRuntimeFeatures(lock);
 
     // these fields start as null because they are populated on first use
@@ -216,6 +218,14 @@ public class OpenhabGraalJSScriptEngine
             eval(globalScript);
         } catch (ScriptException e) {
             LOGGER.error("Could not inject global script", e);
+        }
+    }
+
+    @Override
+    public Object invokeFunction(String s, Object... objects) throws ScriptException, NoSuchMethodException {
+        // Synchronize multi-thread access to avoid exceptions when reloading a script file while the script is running
+        synchronized (lock) {
+            return super.invokeFunction(s, objects);
         }
     }
 
--- a/bundles/org.openhab.automation.jsscripting/src/main/java/org/openhab/automation/jsscripting/internal/OpenhabGraalJSScriptEngine.java
+++ b/bundles/org.openhab.automation.jsscripting/src/main/java/org/openhab/automation/jsscripting/internal/OpenhabGraalJSScriptEngine.java
@@ -69,8 +69,7 @@ public class OpenhabGraalJSScriptEngine
     // final CommonJS search path for our library
     private static final Path NODE_DIR = Paths.get("node_modules");
 
-    // shared lock object for synchronization of multi-thread access
-    private final Object lock = new Object();
+    private final ThreadsafeTimers threadsafeTimers = new ThreadsafeTimers(lock);
 
     // these fields start as null because they are populated on first use
     private String engineIdentifier;
@@ -209,7 +208,7 @@ public class OpenhabGraalJSScriptEngine
         delegate.getBindings(ScriptContext.ENGINE_SCOPE).put(REQUIRE_WRAPPER_NAME, wrapRequireFn);
         // Injections into the JS runtime
         delegate.put("require", wrapRequireFn.apply((Function<Object[], Object>) delegate.get("require")));
-        delegate.put("ThreadsafeTimers", new ThreadsafeTimers(lock));
+        delegate.put("ThreadsafeTimers", threadsafeTimers);
 
         initialized = true;
 
--- a/bundles/org.openhab.automation.jsscripting/src/main/java/org/openhab/automation/jsscripting/internal/OpenhabGraalJSScriptEngine.java
+++ b/bundles/org.openhab.automation.jsscripting/src/main/java/org/openhab/automation/jsscripting/internal/OpenhabGraalJSScriptEngine.java
@@ -57,8 +57,8 @@ import com.oracle.truffle.js.scriptengine.GraalJSScriptEngine;
  *
  * @author Jonathan Gilbert - Initial contribution
  * @author Dan Cunningham - Script injections
- * @author Florian Hotze - Create lock object for multi-thread synchronization
- * @author Florian Hotze - Inject the {@link JSRuntimeFeatures} into the JS context
+ * @author Florian Hotze - Create lock object for multi-thread synchronization; Inject the {@link JSRuntimeFeatures}
+ *         into the JS context
  */
 public class OpenhabGraalJSScriptEngine
         extends InvocationInterceptingScriptEngineWithInvocableAndAutoCloseable<GraalJSScriptEngine> {
--- a/bundles/org.openhab.automation.jsscripting/src/main/java/org/openhab/automation/jsscripting/internal/OpenhabGraalJSScriptEngine.java
+++ b/bundles/org.openhab.automation.jsscripting/src/main/java/org/openhab/automation/jsscripting/internal/OpenhabGraalJSScriptEngine.java
@@ -46,7 +46,6 @@ import org.openhab.automation.jsscripting.internal.fs.PrefixedSeekableByteChanne
 import org.openhab.automation.jsscripting.internal.fs.ReadOnlySeekableByteArrayChannel;
 import org.openhab.automation.jsscripting.internal.fs.watch.JSDependencyTracker;
 import org.openhab.automation.jsscripting.internal.scriptengine.InvocationInterceptingScriptEngineWithInvocableAndAutoCloseable;
-import org.openhab.automation.jsscripting.internal.threading.ThreadsafeTimers;
 import org.openhab.core.automation.module.script.ScriptExtensionAccessor;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -59,6 +58,7 @@ import com.oracle.truffle.js.scriptengine.GraalJSScriptEngine;
  * @author Jonathan Gilbert - Initial contribution
  * @author Dan Cunningham - Script injections
  * @author Florian Hotze - Create lock object for multi-thread synchronization
+ * @author Florian Hotze - Inject the {@link JSRuntimeFeatures} into the JS context
  */
 public class OpenhabGraalJSScriptEngine
         extends InvocationInterceptingScriptEngineWithInvocableAndAutoCloseable<GraalJSScriptEngine> {
@@ -69,7 +69,7 @@ public class OpenhabGraalJSScriptEngine
     // final CommonJS search path for our library
     private static final Path NODE_DIR = Paths.get("node_modules");
 
-    private final ThreadsafeTimers threadsafeTimers = new ThreadsafeTimers(lock);
+    private final JSRuntimeFeatures jsRuntimeFeatures = new JSRuntimeFeatures(lock);
 
     // these fields start as null because they are populated on first use
     private String engineIdentifier;
@@ -208,7 +208,7 @@ public class OpenhabGraalJSScriptEngine
         delegate.getBindings(ScriptContext.ENGINE_SCOPE).put(REQUIRE_WRAPPER_NAME, wrapRequireFn);
         // Injections into the JS runtime
         delegate.put("require", wrapRequireFn.apply((Function<Object[], Object>) delegate.get("require")));
-        delegate.put("ThreadsafeTimers", threadsafeTimers);
+        jsRuntimeFeatures.getFeatures().forEach((key, obj) -> delegate.put(key, obj));
 
         initialized = true;
 
@@ -221,7 +221,7 @@ public class OpenhabGraalJSScriptEngine
 
     @Override
     public void close() {
-        threadsafeTimers.clearAll();
+        jsRuntimeFeatures.close();
     }
 
     /**
--- a/bundles/org.openhab.automation.jsscripting/src/main/java/org/openhab/automation/jsscripting/internal/OpenhabGraalJSScriptEngine.java
+++ b/bundles/org.openhab.automation.jsscripting/src/main/java/org/openhab/automation/jsscripting/internal/OpenhabGraalJSScriptEngine.java
@@ -70,7 +70,7 @@ public class OpenhabGraalJSScriptEngine
     private static final Path NODE_DIR = Paths.get("node_modules");
 
     // shared lock object for synchronization of multi-thread access
-    public final Object lock = new Object();
+    private final Object lock = new Object();
     private final JSRuntimeFeatures jsRuntimeFeatures = new JSRuntimeFeatures(lock);
 
     // these fields start as null because they are populated on first use
--- a/bundles/org.openhab.automation.jsscripting/src/main/java/org/openhab/automation/jsscripting/internal/OpenhabGraalJSScriptEngine.java
+++ b/bundles/org.openhab.automation.jsscripting/src/main/java/org/openhab/automation/jsscripting/internal/OpenhabGraalJSScriptEngine.java
@@ -219,6 +219,11 @@ public class OpenhabGraalJSScriptEngine
         }
     }
 
+    @Override
+    public void close() {
+        threadsafeTimers.clearAll();
+    }
+
     /**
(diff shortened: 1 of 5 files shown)


florian-h05:
@jpg0 @digitaldan 
This is my reimplementation of the timer polyfills.

florian-h05:
## To Do after Merge

Update the docs at openhab-js repo:

- [ ] Mention that the polyfills now behave like standard JS
- [ ] Remove the openHAB Timers from the docs
- [ ] Recommend to use `cancelAll` in the unloaded hook
- [ ] Warn that changing the logger name can lead to messages not displayed in the logs due to missing logger configuration
- [ ] Remove the docs for the raw Java timer creation methods

Update the openhab-js codebase:

- [ ] Remove access to the raw Java timer creation methods

jpg0:
@florian-h05 I would suggest that the better way to inject the ThreadsafeTimers object is the same way that is used for `SharedCache`, and everything in https://github.com/florian-h05/openhab-addons/tree/jsscripting-reimplement-timers/bundles/org.openhab.automation.jsscripting/src/main/java/org/openhab/automation/jsscripting/internal/scope. By default these are injected at module paths, but you can simply remap them in the globals JS script. They have the ability to be unloaded automatically. You may need to pass the lock object differently, I have not looked into that. I do believe that it's worth it though, requiring custom JS when using setTimeout isn't good.

Oh and one other thing - ThreadsafeTimers is not thread safe itself (by concurrent accessors). I'm guessing that this is probably ok, but wanted to make sure you are expecting this.

florian-h05:
> Oh and one other thing - ThreadsafeTimers is not thread safe itself (by concurrent accessors). I'm guessing that this is probably ok, but wanted to make sure you are expecting this.

I’m not sure what you mean by that. Could you help me a little bit?

jpg0:
> Is it still possible then to have its own instance for each script engine?

Yes, this is the same as per-script. You can use `scriptIdentifier` to determine which script it belongs to.

> I’m not sure what you mean by that. Could you help me a little bit?

I mean that if two or more threads concurrently access an instance of ThreadsafeTimers, then it may fail. If access is always done by a single thread it should be fine.

bundles/org.openhab.automation.jsscripting/src/main/java/org/openhab/automation/jsscripting/internal/threading/ThreadsafeTimers.java
```diff
@@ -13,125 +13,179 @@
...
 
-    public Timer createTimer(@Nullable String identifier, ZonedDateTime instant, Runnable callable) {
-        Scheduler scheduler = ScriptServiceUtil.getScheduler();
+    /**
+     * Creates a new and unused timerId.
+     *
+     * @return an unused timerId
+     */
+    private Integer createNewId() {
+        Integer id = idSchedulerMapping.size() + 1;
```
jpg0, line 56: Should we be reusing ids? What does JS do? I ask in case it always guarantees unused ids, in which case scripts could depend on this and reuse may break them.

bundles/org.openhab.automation.jsscripting/src/main/java/org/openhab/automation/jsscripting/internal/threading/ThreadsafeTimers.java
```diff
@@ -13,125 +13,179 @@
...
     }
 
-    public Timer createTimer(@Nullable String identifier, ZonedDateTime instant, Runnable callable) {
-        Scheduler scheduler = ScriptServiceUtil.getScheduler();
+    /**
+     * Creates a new and unused timerId.
+     *
+     * @return an unused timerId
+     */
+    private Integer createNewId() {
```
jpg0, line 55: Why is this `Integer` rather than `int`? Presumably it should never be `null`?

bundles/org.openhab.automation.jsscripting/src/main/java/org/openhab/automation/jsscripting/internal/threading/ThreadsafeTimers.java
florian-h05: I just had a look at the mdn web docs, we should not:

> It is guaranteed that a timeoutID value will never be reused by a subsequent call to setTimeout() or setInterval() on the same object (a window or a worker). However, different objects use separate pools of IDs.

bundles/org.openhab.automation.jsscripting/src/main/java/org/openhab/automation/jsscripting/internal/threading/ThreadsafeTimers.java
florian-h05: I think I have no explanation for that choice.

florian-h05:
@jpg0 
I managed to automatically cancel all scheduled jobs when the engine is unloaded without injecting `ThreadsafeTimers` the same way as `SharedCache`. In my opinion, choosing the ScriptExtension way of injection is much more complicated (and requires more code) and I also need to manage the `ThreadsafeTimer` instances using the `scriptIdentifier`.

I also synchronized another entrypoint to the context using the lock object; this entrypoint was accessed when the script file was reloaded.

> I mean that if two or more threads concurrently access an instance of ThreadsafeTimers, then it may fail.

Got it, thanks. Until now, I've never seen a case where this happened.

jpg0:
The problem with the current mechanism (rather than how SharedCache is done), is that each new piece of functionality that we need added to the environment is a custom job of explicitly modifying the creation of each environment. This is exactly the problem that the alternative mechanism solves - it is a framework for allowing new code to add new things to the environment without modifying the environment creation itself.

jpg0:
You are very unlikely to see it happen because it requires precise timing. I do believe that it's unsafe though, especially as recurring jobs are rescheduled on new threads. What it means is that very occasionally something will fail with a ConcurrentModificationException.


It's not hard to protect against however, I'm happy to add some suggested changes to fix it.

florian-h05:
> It's not hard to protect against however, I'm happy to add some suggested changes to fix it.

Thanks for the help. I'll wait with pushing new commits until you are finished.

florian-h05:
@jpg0 
Please wait before adding suggested changes, I am working on a implemenation using the ScriptExtension.

florian-h05:
@jpg0 

> The problem with the current mechanism (rather than how SharedCache is done), is that each new piece of functionality that we need added to the environment is a custom job of explicitly modifying the creation of each environment.

I was thinking a bit about this fact, and I see your point but the timer polyfills are core functionality that is also provided by NodeJS, while the `SharedCache` is an openHAB specific extension to the JS runtime.

Given that, I would regard the current implementation as okay.

jpg0:
You're right about the scope of these things, so it's not really clear that it belongs in the same place as `SharedCache`. It does still feel nasty to put `ThreadsafeTimers` _directly_ into context creation though. How about just adding one layer of abstraction? Create a `JSRuntimeFeatures` (or whatever) class, which provides a map of things to be put into context creation. This class can just collect them from a list of other classes - currently just `ThreadsafeTimers`. This means that when we add more JS features it's clear where to add them and it will only be in a single place.

florian-h05:
@jpg0 
> How about just adding one layer of abstraction? Create a JSRuntimeFeatures (or whatever) class

Done, see my latest commit.

> Regarding thread safely: you are very unlikely to see it happen because it requires precise timing. I do believe that it's unsafe though, especially as recurring jobs are rescheduled on new threads. What it means is that very occasionally something will fail with a ConcurrentModificationException.
>
> It's not hard to protect against however, I'm happy to add some suggested changes to fix it.

This is the only missing thing, from my side this is ready for merging.

florian-h05:
@jpg0 
I don’t want to stress you, but I’d like to get the changes from this PR into the next milestone (which should be releases this weekend?).

jpg0:
@florian-h05 Did you see the suggestions I added to the review?

florian-h05:
@jpg0 
No, unfortunately not. All I can see from your review are the two comments from 4 days ago about the data type and the uniqueness of the ids.

jpg0:
@florian-h05 Possibly they got saved as comments. Here is an example of one: https://github.com/openhab/openhab-addons/pull/13623#discussion_r1010864603

jpg0 (changes requested):

jpg0:
@florian-h05 give it a go now - the comments were 'pending' and I submitted a review and I think they should now appear

bundles/org.openhab.automation.jsscripting/src/main/java/org/openhab/automation/jsscripting/internal/threading/ThreadsafeTimers.java
```diff
@@ -13,125 +13,179 @@
...
+     * @param delay time in milliseconds that the timer should delay in between executions of the callback
+     * @param callback function to run
+     */
+    private void createLoopingFuture(Integer id, Long delay, Runnable callback) {
+        ScheduledCompletableFuture<Object> future = scheduler.schedule(() -> {
             synchronized (lock) {
-                callable.run();
+                callback.run();
+                if (idSchedulerMapping.get(id) != null)
+                    createLoopingFuture(id, delay, callback);
```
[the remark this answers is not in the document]
florian-h05: > This indicates that this class should probably allow concurrent access to it's methods and should hence be made threadsafe for this.

Is it enough to add the `synchronized` modifier to selected methods?

florian-h05:
@jpg0 
I've addressed your review, please re-review.
See my answer regarding the rescheduling of the job.

bundles/org.openhab.automation.jsscripting/src/main/java/org/openhab/automation/jsscripting/internal/OpenhabGraalJSScriptEngine.java
```diff
@@ -70,7 +70,8 @@
     private static final Path NODE_DIR = Paths.get("node_modules");
 
     // shared lock object for synchronization of multi-thread access
-    private final Object lock = new Object();
+    public final Object lock = new Object();
```
jpg0, line 73: Why public? Shouldn't this remain private?

jpg0:
Other than the minor comment above, it looks good. Happy with the scheduler change too.

As before, you'll have to get someone else such as @digitaldan to actually approve & merge as I don't have permission.

bundles/org.openhab.automation.jsscripting/src/main/java/org/openhab/automation/jsscripting/internal/OpenhabGraalJSScriptEngine.java
florian-h05: Yes.
Seems like code completion did not what I expected it to do.

florian-h05:
@jlaur Can you merge?

Please see the last commit if jpg0 for review.

bundles/org.openhab.automation.jsscripting/src/main/java/org/openhab/automation/jsscripting/internal/threading/ThreadsafeTimers.java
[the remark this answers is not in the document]
jlaur: @jpg0, @florian-h05 - can this comment be resolved?

florian-h05:
@jlaur 
I addressed your review.

> The PR is tagged as a breaking change,

Yes, I've added this tag as the change of the implementation of the polyfills for `setTimeout` and `setInterval` could possibly break existing installations. But this is only the case if the user's scripts expect that `setTimeout` and `setInterval` return an openHAB Timer class (see https://github.com/openhab/openhab-core/blob/18d063e272de14f4657ae925270f0d524f056fa1/bundles/org.openhab.core.model.script/src/org/openhab/core/model/script/internal/actions/TimerImpl.java#L31); this behaviour is not the standard JS behaviour.

> how can we ensure that users are warned about this, 

By adding a breaking change note to the release notes and adding an alert to https://github.com/openhab/openhab-distro/blob/main/distributions/openhab/src/main/resources/bin/update.lst.
he case, that 

> how can we ensure that users are warned about this, 

By adding a breaking change note to the release notes and adding an alert to https://github.com/openhab/openhab-distro/blob/main/distributions/openhab/src/main/resources/bin/update.lst.

> and which changes will they need to do?

For the rare case that a user relies on the polyfills returning an openHAB Timer, he has to update his code to e.g. cancel the timer by running `cancelTimeout` or `cancelInterval` on the returned numeric id. To say it in some other way, the user has to adjust his usage of `setTimeout` and `setInterval` so that his code would run on NodeJS or in the browser.

> Perhaps @kaikreuzer could also have a quick look since also involved in the previous iteration.

If you like to @jlaur, no problem. I'm just not sure if that's required since Kai onyl left a few short comments (the same sort that your comments are) and merged the PR, the changes itself were also guided and reviewed by jpg0.

jlaur:
> I propose to log a warning when someone tries to access those timer creation methods via the ScriptExecution action from the openhab-js library.

@florian-h05 - I assume this would be a PR to be merged into openhab-js after this PR is merged?

florian-h05:
> this small breaking change, which to me seems like is for the better as otherwise we would block ourselves from moving forward, and it would only get worse.

FYI also see https://github.com/openhab/openhab-addons/issues/13594#issuecomment-1289241781 for what @digitaldan said about reimplementing the polyfills.

> Let's not rush it too much, but still aim for the upcoming milestone. If no one else will provide feedback today or tomorrow, I will merge.

👍

> I assume this would be a PR to be merged into openhab-js after this PR is merged?

Yes, extractly. I'll prepare a PR in openhab-js to complete the tasks I mentioned at the beginning of this thead. The new release also needs to go into the milestone, so I'll ping you @jlaur for merging.

bundles/org.openhab.automation.jsscripting/src/main/java/org/openhab/automation/jsscripting/internal/threading/ThreadsafeTimers.java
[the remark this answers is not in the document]
florian-h05: Yes, this can be resolved:

jpg0:

> Happy with the scheduler change too.

kaikreuzer:
> If you like to @jlaur, no problem. I'm just not sure if that's required since Kai onyl left a few short comments

I agree, no need for an explicit approval by me. So @jlaur, if you are happy with the changes, please merge!

jlaur (approved):
LGTM

florian-h05:
@jlaur 
openhab-js PR is now open, see https://github.com/openhab/openhab-js/pull/169.

PR #13629, opened 2022-10-31
[paradoxalarm] Fix issue that the config of paradox bridge handler gets overwritten when there are more than one bridge

### LINKED ISSUE #13628: [paradoxalarm] When using two different bridges the same IP address gets sent to the communicator
This is a copy paste from the following community issue: [https://community.openhab.org/t/paradox-evo-binding/58026/58]

As far as I can see, the problem is that both comunicators gets the same IP address and one is timing out…

See those lines:

2022-10-31 15:20:03.894 [DEBUG] [l.handlers.ParadoxIP150BridgeHandler] - Phase2 - Creating communicator for panel EVO192

2022-10-31 15:20:03.897 [DEBUG] [.communication.ParadoxBuilderFactory] - Creating new builder for Paradox EVO192 system

2022-10-31 15:20:04.082 [DEBUG] [l.communication.AbstractCommunicator] - IP Address=10.10.6.10, TCP Port=10000

2022-10-31 15:20:04.099 [DEBUG] [al.communication.GenericCommunicator] - Use encryption=false

2022-10-31 15:20:04.113 [DEBUG] [ternal.communication.EvoCommunicator] - PanelType=EVO192, Max Partitions=1, Max Zones=16

2022-10-31 15:20:04.184 [DEBUG] [l.handlers.ParadoxIP150BridgeHandler] - Phase2 - Creating communicator for panel EVO192

2022-10-31 15:20:04.185 [DEBUG] [.communication.ParadoxBuilderFactory] - Creating new builder for Paradox EVO192 system

2022-10-31 15:20:04.188 [DEBUG] [l.communication.AbstractCommunicator] - IP Address=10.10.6.10, TCP Port=10000

The second IP should be 192.168.100.2 not 10.10.6.10 again.

openhab things are set like this:

In paradox_mcs.things file:

Bridge paradoxalarm:ip150:mcs "Paradox IP150 MCS" [refresh=3, panelType="EVO192", ip150Password="paradox", pcPassword="0000", ipAddress="10.10.6.10", port=10000, reconnectWaitTime=10, maxPartitions=1, maxZones=16 ] {}

In paradox_vb.things file:

Bridge paradoxalarm:ip150:vb "Paradox IP150 Vulcana" [refresh=10, panelType="EVO192", ip150Password="paradox", pcPassword="0000", ipAddress="192.168.100.2", port=10000, reconnectWaitTime=30, maxPartitions=1, maxZones=16 ] {}

theater:
* Change static config field in the main bridge handler to non-static (fixes #13628)

Signed-off-by: Konstantin Polihronov <polychronov@gmail.com>

### DIFF
--- a/bundles/org.openhab.binding.paradoxalarm/src/main/java/org/openhab/binding/paradoxalarm/internal/handlers/ParadoxIP150BridgeHandler.java
+++ b/bundles/org.openhab.binding.paradoxalarm/src/main/java/org/openhab/binding/paradoxalarm/internal/handlers/ParadoxIP150BridgeHandler.java
@@ -70,7 +70,7 @@ public class ParadoxIP150BridgeHandler extends BaseBridgeHandler
 
     private IParadoxCommunicator communicator;
 
-    private static ParadoxIP150BridgeConfiguration config;
+    private ParadoxIP150BridgeConfiguration config;
     private @Nullable ScheduledFuture<?> refreshCacheUpdateSchedule;
 
     private long timeStamp = 0;


lolodomo (approved):
Good catch.

fiftyz:
But still does not work... 
Here is the log:
[paradox.log](https://github.com/openhab/openhab-addons/files/9910176/paradox.log)

fiftyz:
I think that the problem is from both bridges getting the same panel with two communicators.

image

image

But I don't know how to solve it... :nerd_face:

theater:
You're right. The panel is using singleton pattern, therefore no option to have second one. I have to think how to redesign this but it will be definitely more effort than the previous fix. Thanks for the hint. Can you create another bug for that as this one is already merged so we keep clear tracking of the changes? Assign it to me and I'll have a look when I have time.

PR #13631, opened 2022-10-31
[boschshc] Support obtaining battery states

david-pace:
This change adds support for obtaining the battery state for the following devices:

* Motion Detector
* Thermostat
* Twinguard
* Wall Thermostat
* Window/Door Contact

The following changes were made:

* Add channel type definition for battery levels
* Add battery level channels to Motion Detector, Thermostat, Twinguard, Wall Thermostat and Window/Door Contact
* Add constant for battery level channel in BoschSHCBindingConstants
* Implement abstract handler and service for battery-powered devices
* Let appropriate devices inherit the abstract implementation
* Add missing super calls in initializeServices() methods
* Rename existing getServiceURL() to getServiceStateURL() in HTTP client
* Add methods to retrieve service states without the suffix "/state" in the URL
* Rename DeviceStatusUpdate to DeviceServiceData
* Let DeviceServiceData extend BoschSHCServiceState
* Extend DeviceServiceData DTO with model for faults
* Enhance bridge handler: handle updates without state sub-objects, extract methods to enhance readability
* Add unit tests for all affected devices
* Minor code enhancements
* Update documentation

### DIFF
--- a/bundles/org.openhab.binding.boschshc/README.md
+++ b/bundles/org.openhab.binding.boschshc/README.md
@@ -49,7 +49,7 @@ A compact smart plug with energy monitoring capabilities.
 | power-consumption  | Number:Power  | &#9744;  | Current power consumption (W) of the device.     |
 | energy-consumption | Number:Energy | &#9744;  | Cumulated energy consumption (Wh) of the device. |
 
-### Twinguard smoke detector
+### Twinguard Smoke Detector
 
 The Twinguard smoke detector warns you in case of fire and constantly monitors the air.
 
@@ -65,6 +65,8 @@ The Twinguard smoke detector warns you in case of fire and constantly monitors t
 | purity-rating      | String               | &#9744;  | Rating of current measured purity.                                                                |
 | air-description    | String               | &#9744;  | Overall description of the air quality.                                                           |
 | combined-rating    | String               | &#9744;  | Combined rating of the air quality.                                                               |
+| battery-level      | Number               | &#9744;  | Current battery level percentage as integer number. Bosch-specific battery levels are mapped to numbers as follows: `OK`: 100, `LOW_BATTERY`: 10, `CRITICAL_LOW`: 1, `CRITICALLY_LOW_BATTERY`: 1, `NOT_AVAILABLE`: `UNDEF`. |
+| low-battery        | Switch               | &#9744;  | Indicates whether the battery is low (`ON`) or OK (`OFF`). |
 
 ### Door/Window Contact
 
@@ -75,6 +77,8 @@ Detects open windows and doors.
 | Channel Type ID | Item Type | Writable | Description                  |
 | --------------- | --------- | :------: | ---------------------------- |
 | contact         | Contact   | &#9744;  | Contact state of the device. |
+| battery-level   | Number    | &#9744;  | Current battery level percentage as integer number. Bosch-specific battery levels are mapped to numbers as follows: `OK`: 100, `LOW_BATTERY`: 10, `CRITICAL_LOW`: 1, `CRITICALLY_LOW_BATTERY`: 1, `NOT_AVAILABLE`: `UNDEF`. |
+| low-battery     | Switch    | &#9744;  | Indicates whether the battery is low (`ON`) or OK (`OFF`). |
 
 ### Motion Detector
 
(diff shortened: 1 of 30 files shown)


jlaur:
@david-pace - have you considered mapping the possible values to a number so that system channel `system.battery-level` could be used? This would be more similar to the way other bindings implement battery level channels, and would add default functionality like showing appropriate battery icons depending on the current level.

See for example #11206 where I had a similar problem of being limited to only a few possible values. A channel for type `system.low-battery` could also be implemented for `LOW_BATTERY` and below.

david-pace:
Thanks for the hint @jlaur, I was not aware of the system channel type. I will adapt my code accordingly :+1:

jlaur:
> Thanks for the hint @jlaur, I was not aware of the system channel type. I will adapt my code accordingly 👍

The hard part will probably be finding some reasonable values to map. I see you have only four possible levels: `OK`, `LOW_BATTERY`, `CRITICAL_LOW`, `CRITICALLY_LOW_BATTERY` and `NOT_AVAILABLE` where three of them are actually low battery.

`NOT_AVAILABLE` can be mapped to `UNDEF`. I'm wondering what is the difference between `CRITICAL_LOW` and `CRITICALLY_LOW_BATTERY`?

A `lowBattery` channel where `OK` is false, and the others are true is almost enough, but only having this, you would lose the difference between "low" and "critically low". Perhaps OK -> 100, Low -> 10, Critical -> 1. If you create a sitemap with such item having `<battery>` as category, you should be able to see the different icons depending on set value.

david-pace:
There is an official [API documentation](https://apidocs.bosch-smarthome.com/local/) from the manufacturer, but unfortunately many services are not documented, `BatteryLevel` being one of them.

My source for the possible states is actually this Python API: https://github.com/tschamm/boschshcpy/blob/master/boschshcpy/services_impl.py#L580

So far I was only able to provoke the states `OK` and `LOW_BATTERY` with the hardware, but I might experiment more in order to find out if there is a difference between the other two states.

My first attempt for a mapping was:

- `OK` -> 100
- `LOW_BATTERY` -> 50
- `CRITICAL_LOW` -> 30
- `CRITICALLY_LOW_BATTERY` -> 10

but of course this can be discussed.

I'm using `DecimalType` to set the value in the channel, is this correct? I tried `PercentType` before but this displayed `1%` when I set a value of `100`. Can you give me a hint how to set the value `UNDEF`? I could not find any constants/ways to set this.

I wonder if it might be enough to just use a `system.low-battery` channel with a boolean state as you proposed. @GerdZanker @coeing do you have any preferences here and if you prefer an integer channel, which mapping would you propose?

jlaur:
> Can you give me a hint how to set the value `UNDEF`?

```java
updateState(CHANNEL_BATTERY_LEVEL, UnDefType.UNDEF);
```

GerdZanker:
> I wonder if it might be enough to just use a `system.low-battery` channel with a boolean state as you proposed. @GerdZanker @coeing do you have any preferences here and if you prefer an integer channel, which mapping would you propose?

I have no preference here, regarding the bool state or an integer. But I can suggest to start with a simple solid implementation first and later extend it - if you have a first simple mapping in place use it.

GerdZanker (approved):
Great work @david-pace.
Your code is working fine and all my batteries are OK, even after about two years of Twinguard operation.

david-pace:
Thanks for testing, but please do not merge yet. I haven't pushed a change using a standard openHAB channel yet.

jlaur:
> Thanks for testing, but please do not merge yet. I haven't pushed a change using a standard openHAB channel yet.

Yes, I'm awaiting that. :-) To be sure that no maintainer can merge by accident, you could temporarily change the PR state back to draft.

GerdZanker (approved):

jlaur:
@david-pace - all looks good to me. You could consider also adding a `low-battery` channel for convenience. This can be useful to trigger alarms, so users don't have to deal with concrete battery level.

jlaur (approved):
Thanks!

david-pace:
The latest version includes a `system.low-battery` channel as well. Now users can decide whether they want tu use the more fine-grained integer channel or a simple boolean channel :+1:

Thank you for your reviews and suggestions 🙂

PR #13636, opened 2022-11-01
[netatmo] eventBuffer in SecurityCapability should not be static

lolodomo:
Also fix few findings by code analysis

### DIFF
--- a/bundles/org.openhab.binding.netatmo/src/main/java/org/openhab/binding/netatmo/internal/handler/capability/SecurityCapability.java
+++ b/bundles/org.openhab.binding.netatmo/src/main/java/org/openhab/binding/netatmo/internal/handler/capability/SecurityCapability.java
@@ -47,7 +47,7 @@ import org.slf4j.LoggerFactory;
 class SecurityCapability extends RestCapability<SecurityApi> {
     private final Logger logger = LoggerFactory.getLogger(SecurityCapability.class);
 
-    private static final Map<String, HomeEvent> eventBuffer = new HashMap<>();
+    private final Map<String, HomeEvent> eventBuffer = new HashMap<>();
     private @Nullable ZonedDateTime freshestEventTime;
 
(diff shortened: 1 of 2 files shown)


bundles/org.openhab.binding.netatmo/src/main/java/org/openhab/binding/netatmo/internal/handler/capability/SecurityCapability.java
```diff
@@ -47,7 +47,7 @@
 class SecurityCapability extends RestCapability<SecurityApi> {
     private final Logger logger = LoggerFactory.getLogger(SecurityCapability.class);
 
-    private static final Map<String, HomeEvent> eventBuffer = new HashMap<>();
+    private static final Map<String, HomeEvent> EVENT_BUFFER = new HashMap<>();
```
lolodomo, line 50: @clinique : I have a BIG doubt if it must be static or not ?

clinique: No, it should not be static.

lolodomo: Ok, fixed.

jlaur (approved):
Thanks! Would it be possible to assess what could have been the consequence of this bug? Then we might be able to find possible connected issues and modify PR title to be user-oriented.

clinique:
We could imagine a user having multiple houses, in this case, the event buffer would have mixed events coming from both houses, leading to a potential loss of events.

jlaur:
@lolodomo - can you either remove "Bug" label for being excluded from the release notes, or adapt the title according to the possible user impact described by @clinique?

lolodomo:
Title updated

PR #13641, opened 2022-11-01
[paradoxalarm] Handle multiple panels

### LINKED ISSUE #13640: [paradoxalarm] Bindng cannot handle multiple panels
Paradox Alarm Binding cannot handle multiple bridges / panels

_Originally posted by @theater in https://github.com/openhab/openhab-addons/issues/13629#issuecomment-1298701812_

silviuchingaru:
Fixes #13640

Handle multiple panels (fixes #13640)

With this patch worked with 2 pannels connected but should work with unlimited number of panels.

P.S.: @theater I asigned to myself, you did enough already! :wink:

### DIFF
--- a/bundles/org.openhab.binding.paradoxalarm/src/main/java/org/openhab/binding/paradoxalarm/internal/communication/CommunicationState.java
+++ b/bundles/org.openhab.binding.paradoxalarm/src/main/java/org/openhab/binding/paradoxalarm/internal/communication/CommunicationState.java
@@ -22,7 +22,6 @@ import org.openhab.binding.paradoxalarm.internal.communication.messages.HeaderMe
 import org.openhab.binding.paradoxalarm.internal.communication.messages.IPPacket;
 import org.openhab.binding.paradoxalarm.internal.communication.messages.IpMessagesConstants;
 import org.openhab.binding.paradoxalarm.internal.communication.messages.ParadoxIPPacket;
-import org.openhab.binding.paradoxalarm.internal.model.ParadoxPanel;
 import org.openhab.binding.paradoxalarm.internal.util.ParadoxUtil;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -368,7 +367,6 @@ public enum CommunicationState implements IResponseReceiver {
             if (communicator != null) {
                 communicator.close();
             }
-            ParadoxPanel.getInstance().dispose();
         }
     };
 
--- a/bundles/org.openhab.binding.paradoxalarm/src/main/java/org/openhab/binding/paradoxalarm/internal/model/Partition.java
+++ b/bundles/org.openhab.binding.paradoxalarm/src/main/java/org/openhab/binding/paradoxalarm/internal/model/Partition.java
@@ -12,7 +12,6 @@
  */
 package org.openhab.binding.paradoxalarm.internal.model;
 
-import org.openhab.binding.paradoxalarm.internal.communication.IParadoxCommunicator;
 import org.openhab.binding.paradoxalarm.internal.communication.PartitionCommandRequest;
 import org.openhab.binding.paradoxalarm.internal.communication.RequestType;
 import org.openhab.binding.paradoxalarm.internal.communication.messages.CommandPayload;
@@ -34,11 +33,9 @@ public class Partition extends Entity implements Commandable {
     private final Logger logger = LoggerFactory.getLogger(Partition.class);
 
     private PartitionState state = new PartitionState();
-    private IParadoxCommunicator communicator;
 
-    public Partition(IParadoxCommunicator com, int id, String label) {
-        super(id, label);
-        communicator = com;
+    public Partition(ParadoxPanel panel, int id, String label) {
+        super(panel, id, label);
     }
 
     public PartitionState getState() {
@@ -64,6 +61,6 @@ public class Partition extends Entity implements Commandable {
         ParadoxIPPacket packet = new ParadoxIPPacket(payload.getBytes())
                 .setMessageType(HeaderMessageType.SERIAL_PASSTHRU_REQUEST);
         PartitionCommandRequest request = new PartitionCommandRequest(RequestType.PARTITION_COMMAND, packet, null);
-        communicator.submitRequest(request);
+        getPanel().getCommunicator().submitRequest(request);
     }
 }
--- a/bundles/org.openhab.binding.paradoxalarm/src/main/java/org/openhab/binding/paradoxalarm/internal/model/Partition.java
+++ b/bundles/org.openhab.binding.paradoxalarm/src/main/java/org/openhab/binding/paradoxalarm/internal/model/Partition.java
@@ -34,9 +34,11 @@ public class Partition extends Entity implements Commandable {
     private final Logger logger = LoggerFactory.getLogger(Partition.class);
 
     private PartitionState state = new PartitionState();
+    private IParadoxCommunicator communicator;
 
-    public Partition(int id, String label) {
+    public Partition(IParadoxCommunicator com, int id, String label) {
         super(id, label);
+        communicator = com;
     }
 
     public PartitionState getState() {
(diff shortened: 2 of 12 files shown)


theater:
The overall fix seems legit to me, i.e. instantiate the panel in the bridge(s). Is it all that was needed to fix the multiple bridges issue? I'll have a deeper look in the code later when I have more time.
Thanks for handling the issue! Good job!

fiftyz:
Thank you for your answer and quick review!
Yes, it was  all that was needed to fix the multiple bridges issue. The problem was with static assignment of ParadoxPanel class which has actualy all the panel info, partitions, zones and so on. Because was staticaly asigned, the ParadoxPanel was the same for both bridges so even if the bridge, by using different comunicator, tried to update the second panel, it received information or at least panel info, partitions and zone mappings of the first connected panel. Maybe information received was correct but probably was mapped to zones of the first connected panel, I did not test this but as far as I've seen in debugging, this way worked after the patch #13629.
I'll wait for your review and I'll further test it to see if everything works ok.
I'll try to add also a third EVO192 panel and a fourth SP7000 to my setup but I need to make the parser first for last one. I'll open a different issue for SP7000 support once I have something working and maybe,  if you have time, you will guide me a little if I'll get stuck... 😉

theater:
Hi, 
I hope we don't get with other design flaws during your SP7000 implementation but if we do, I guess we need to overcome them.
I made the binding with the idea of supporting other panels but I cannot guarantee that everything is OK (like the static singleton Panel here...)
I'll try to help as much as possible for the parser but the hex-stuff is not my best. :(
Will try to find time to review the code today. Does not look much.

Cheers,
K.

bundles/org.openhab.binding.paradoxalarm/src/main/java/org/openhab/binding/paradoxalarm/internal/model/Partition.java
```diff
@@ -34,9 +34,11 @@ public class Partition extends Entity implements Commandable {
     private final Logger logger = LoggerFactory.getLogger(Partition.class);
 
     private PartitionState state = new PartitionState();
+    private IParadoxCommunicator communicator;
 
-    public Partition(int id, String label) {
+    public Partition(IParadoxCommunicator com, int id, String label) {
```
theater, line 39: Adding communicator to the model object partition loads unnecessary the Partition class which is meant to work as a model bean storing data from the caches.
It was not very well designed from the beginning by me and relied on the static retrieval of the single panel before.
My proposal is to remove completely the Commandable interface and move the handleCommand logic to the ParadoxPartitionHandler where you only need to retrieve the partition ID from the model object and pass it to the communicator command.
What do you think?

bundles/org.openhab.binding.paradoxalarm/src/main/java/org/openhab/binding/paradoxalarm/internal/communication/CommunicationState.java
```diff
@@ -368,7 +367,6 @@ protected void runPhase(IParadoxInitialLoginCommunicator communicator, Object...
             if (communicator != null) {
                 communicator.close();
             }
-            ParadoxPanel.getInstance().dispose();
```
theater, line 371: Verify that if the communicator wents offline and then online the lack of disposal of internal collections (partitions and zones) will not make any side effect when reconnecting. You can test that with the RESET or LOGOUT functionality.
If that's fine, we can get rid of the dispose() method completely in the ParadoxPanel class.

theater (changes requested):
Hi,
all in all everything looks good. I've added one comment that will maybe improve a little bit the wrong design where the Partition class from the model requires dependency to the Panel/Communicator.
Please have a look and let me know what do you think. It's open for discussion.

Cheers,
Konstantin

bundles/org.openhab.binding.paradoxalarm/src/main/java/org/openhab/binding/paradoxalarm/internal/communication/CommunicationState.java
fiftyz: I think the best way to dispose the panel is in ParadoxIP150BridgeHandler.dispose() which calls CommunicationState.logout(communicator) and, according to comment:
```
// For some reason after sending logout packet the connection gets reset from the other end
// currently workaround is to run directly offline phase, i.e. close socket from our end
```
I'll do some tests.

bundles/org.openhab.binding.paradoxalarm/src/main/java/org/openhab/binding/paradoxalarm/internal/model/Partition.java
fiftyz: Actually, I think that partition should have it's Panel reference, Panel has the Communicator and the command should be sent to it's communicator in handleCommand(). I'm not seeing why _adding communicator to the model object partition loads unnecessary the Partition class_. Can you please what you mean by loads the Partition class?! It is only a pointer to communicator right now but I think it should be to actually to panel.
We could change also the interface to accept a ParadoxPanel or Communicator parameter.

bundles/org.openhab.binding.paradoxalarm/src/main/java/org/openhab/binding/paradoxalarm/internal/communication/CommunicationState.java
fiftyz: I've done more tests and I discovered that without  ParadoxPanel.dispose(), if you disable and enable bridge thing, bridge won't get ONLINE again but if we dispose the Panel in ParadoxIP150BridgeHandler.dispose() as mentioned above, everything works fine.

I've pushed the change.

bundles/org.openhab.binding.paradoxalarm/src/main/java/org/openhab/binding/paradoxalarm/internal/model/Partition.java
theater: In my opinion this is unnecessary dependency to the communicator / panel. The hierarchy should be Panel which consists of partitions/zones/other entities (doors?). I don't even remember why I did it like this... probably had an idea to command also doors (not supported) / zones (nothing to control there AFAIK) and every object to know by itself what commands to send to the panel/communicator. Currently I see it as additional dependency from the model (data) to the communicator (the action). The only benefit I see from this now (few years after the initial implementation) is that you can use this from he Main() class as a stand alone application for testing and maybe if someone adds future support for other entities, so you can just imlement internally the handleCommand method for each entity (door for example). What benefits do you see of this implementation instead of moving it in the handler logic?

theater (approved):

bundles/org.openhab.binding.paradoxalarm/src/main/java/org/openhab/binding/paradoxalarm/internal/model/Partition.java
fiftyz: So, would you agree to pass a refference of Panel to Entities? I think is the best approach in a object oriented programming like Java.

theater: Yes. I have approved the PR.

fiftyz: I saw that but there are two more things to add before this is ready to be Merged:
1. Panel disposal on ParadoxIP150BridgeHandler.dispose() - without this, Bridge won't came up online after disable / enable.
2. Every Entity should have access to it's Panel and Communicator via Panel so the constructor of Entity I propose to be modified to something like:
```
public Entity(ParadoxPanel panel, int id, String label) {
    this.panel = panel;
    this.id = id;
    this.label = label.trim();
    logger.debug("Creating entity with label [{}] and ID [{}] for panel [{}]", label, id, panel);
}
```
Do you agree with this proposal?

theater: Yep. I saw the panel dispose and agree with that. Also if you see fit the panel reference in the sub-entity, let's leave it as proposed in the constructor. Looks OK to me.

theater: I'm just not sure if my review is enough for merge. Before it was expected that someone from reviewers has to approve it.

theater:
OK I guess now we have to wait for a check by some of the maintainers for the OpenHAB guidelines. I cannot merge your request unfortunately but from semantics perspective in the binding context the PR is OK.

lolodomo (approved):
LGTM, thank you

silviuchingaru:
Thank you also!

PR #13642, opened 2022-11-01
[mqtt.generic] Update README.md

andreacioni:
`color_mode` is not recognized as a channel parameter, instead, `colorMode` is


jlaur (approved):
Thanks!

jlaur:
For reference:
https://github.com/openhab/openhab-addons/blob/e24d43aecbcfc5af551c8e9087d9be6992e5c69a/bundles/org.openhab.binding.mqtt.generic/src/main/resources/OH-INF/config/color-channel-config.xml#L15-L24

PR #13479, opened 2022-10-02
[tellstick] Avoid updates duplication after communication errors

lolodomo:
Fix #13453

Do not register the same device handler many times as listener in the bridge handler
Unregister the device handler from the bridge handler when disposing device handler

HTTP timeout set to 15s
Remove the retry mechanism related to the timeout
Check HTTP status code
Fix discovery service unregistration
Log statistics about request/refresh durations and number of timeouts/errors
Change logging in case of exception
Also change few logs level (remove usage of logger.error)
Execute one refresh at bridge initialization and not 2
Small enhancement of the bridge/things status management
implement discovery service unregistration
Fix few code analysis findings

Signed-off-by: Laurent Garnier <lg.hc@free.fr>

PR #13652, opened 2022-11-03
[verisure] Avoid updates duplication after communication errors

lolodomo:
Avoid registering several times the same device status listener.
It is called by the thing handler each time the bridge status changed to ONLINE.

This bug was reported by @ulfwin while discussing a similar problem within the Tellstick binding: PR #13479 

### DIFF
--- a/bundles/org.openhab.binding.verisure/src/main/java/org/openhab/binding/verisure/internal/VerisureSession.java
+++ b/bundles/org.openhab.binding.verisure/src/main/java/org/openhab/binding/verisure/internal/VerisureSession.java
@@ -27,8 +27,8 @@ import java.util.Collection;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
-import java.util.concurrent.CopyOnWriteArrayList;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.TimeoutException;
 
@@ -82,7 +82,7 @@ public class VerisureSession {
     private final Map<String, VerisureThingHandler<?>> verisureHandlers = new ConcurrentHashMap<>();
     private final Logger logger = LoggerFactory.getLogger(VerisureSession.class);
     private final Gson gson = new Gson();
-    private final List<DeviceStatusListener<VerisureThingDTO>> deviceStatusListeners = new CopyOnWriteArrayList<>();
+    private final Set<DeviceStatusListener<VerisureThingDTO>> deviceStatusListeners = ConcurrentHashMap.newKeySet();
     private final Map<BigDecimal, VerisureInstallation> verisureInstallations = new ConcurrentHashMap<>();
     private static final List<String> APISERVERLIST = Arrays.asList("https://m-api01.verisure.com",
             "https://m-api02.verisure.com");


lolodomo:
@jlaur: it would be cool if you can review this PR before the build of the next milestone as it fixes a relatively critical bug.

bundles/org.openhab.binding.verisure/src/main/java/org/openhab/binding/verisure/internal/VerisureSession.java
```diff
@@ -155,7 +155,8 @@ public boolean unregisterDeviceStatusListener(
 
     @SuppressWarnings("unchecked")
     public boolean registerDeviceStatusListener(DeviceStatusListener<? extends VerisureThingDTO> deviceStatusListener) {
-        return deviceStatusListeners.add((DeviceStatusListener<VerisureThingDTO>) deviceStatusListener);
+        return deviceStatusListeners.contains(deviceStatusListener) ? false
```
jlaur, line 158: Perhaps it would be better to store listeners in a `Set`? This way they would be distinct by contract and lookup would be faster:
```java
Set<DeviceStatusListener<VerisureThingDTO>> deviceStatusListeners = ConcurrentHashMap.newKeySet()
```

jannegpriv: I like the idea of using a Set.

lolodomo: @jannegpriv: would you like to propose your own PR and test it? 
I hav myself no way to test this binding.

jlaur: @lolodomo - I believe you will literally just have to replace the declaration with the suggestion above and revert the changes in `registerDeviceStatusListener`. I would consider this change at least as safe as the one above, but in any case it would be preferable to have it tested, of course.

lolodomo: Done

ulfwin: I'd be happy to test, if you want!

jannegpriv:
I have built and tested the change and it has been running in my production environmnt since yesterday evening and I have not seen any issues.

jannegpriv (approved):
LGTM!

jlaur (approved):
LGTM

jlaur:
> I have built and tested the change and it has been running in my production environment since yesterday evening and I have not seen any issues.

Just for curiosity/confirmation: You tested with `Set`, which was proposed yesterday, but only committed today?

jannegpriv:
> > I have built and tested the change and it has been running in my production environment since yesterday evening and I have not seen any issues.
> 
> Just for curiosity/confirmation: You tested with `Set`, which was proposed yesterday, but only committed today?

I built locally yesterday with proposed solution and tested that jar in my production environment. 
I did not do any commit today, not sure what you mean?

jlaur:
> I built locally yesterday with proposed solution and tested that jar in my production environment.

👍

> I did not do any commit today, not sure what you mean?

@lolodomo committed the proposed fix today, so was not sure exactly what you tested since yesterday. Got it now, thanks for testing it.

lolodomo:
@jlaur is just asking if your test was with the PR from yesterday or from this morning. Apparently, your test was not with the final proposal from this morning.

ulfwin:
This is great, thanks a lot all of you for your help!!

Would it be possible to get a link to a .jar so I can start using it immediately?

jannegpriv:
I should have clarified that  I meant `Set` solution.

lolodomo:
> Would it be possible to get a link to a .jar so I can start using it immediately?

A new 3.4 milestone is normally scheduled for today.

PR #13653, opened 2022-11-04
[homeconnect] Fix login for simulator environment

bruestel:
The simulated Home Connect environment cannot currently be used (https://developer.home-connect.com/simulator). The test environment always expects a valid "redirect URI", although according to the documentation it is optional.

I reported the bug a few months ago but unfortunately didn't get a reply. That's why I created this pull request.


lolodomo (approved):
LGTM

PR #13658, opened 2022-11-05
[nuvo] Add zone actions for rules

mlobstein:
Add a system channel that reports the most recent keypad button press and which zone it was sent from. This allows rules to take actions based on a particular zone. For example, it is now possible to make a rule that turns off all other zones except the caller zone. A sample rule demonstrating this was provided.

It was necessary to refactor NuvoMessageEvent to contain both the Zone Id and Source Id of the update messages instead of the previously ambiguous 'key'. Some variable names were also clarified to designate them as part of enhanced NuvoNet functionality vs. the generic serial port api functionality.


bundles/org.openhab.binding.nuvo/README.md
```diff
@@ -473,12 +475,57 @@ A complete XML string for the desired menu is then stored in the `menuXmlSrcN` c
...
     <item>menu3 y</item>
 </topmenu>
+<topmenu text="Turn off other zones"/>
 ```
 
 When a menu item is selected, the text of the topmenu item and sub menu item (if applicable) will be sent to the button channel in a pipe delimited format.
 For example, when item `menu1 b` is selected, the text `Top menu 1|menu1 b` will be sent to the button channel.
 When the item `Top menu 2` is selected the text sent to the button channel will simply be `Top menu 2` since this menu item does not have any sub menu items.
 
+### Rule to trigger an action based which keypad zone where a button was pressed or menu item selected
```
lolodomo, line 485: "based on which"

bundles/org.openhab.binding.nuvo/README.md
```diff
@@ -473,12 +475,57 @@ A complete XML string for the desired menu is then stored in the `menuXmlSrcN` c
...
+
+By using the `system#button_press` channel it is possible to trigger an action based on which keypad zone was used to send the action.
+This channel appends the zone number and a comma before the button action or menu item selection.  
+
+For example if the Play/Pause button is pressed on Zone 7, the channel will display: `7,PLAYPAUSE`  
+Also if a menu item from a custom menu was selected, ie: `Top menu 1` on Zone 5, the channel will display: `5,Top menu 1`  
+
+The functionality can be utilized to create very powerful rules as demontrated below. The following rule triggered from a menu item turns off all zones except for the zone that triggered the rule.
```
lolodomo, line 493: utilized or used ?
I have a doubt if "utilize" is correct in English ?

bundles/org.openhab.binding.nuvo/README.md
```diff
@@ -473,12 +475,57 @@ A complete XML string for the desired menu is then stored in the `menuXmlSrcN` c
...
+
+nuvo-turn-off-all-but-caller.rules:
+
+```
+rule "Turn off all zones except caller zone"
+when
+    Item nuvo_system_button_press received update
+then
+    var callerZone = nuvo_system_button_press.state.toString().split(",").get(0)
+    var button = nuvo_system_button_press.state.toString().split(",").get(1)
```
lolodomo, line 503: It is always more reliable to use `newState` rather than `nuvo_system_button_press.state` in that context.

lolodomo (changes requested):

bundles/org.openhab.binding.nuvo/src/main/resources/OH-INF/thing/channels.xml
```diff
@@ -342,6 +342,7 @@
 			<channel id="allmute" typeId="system.mute"/>
 			<channel id="page" typeId="page"/>
 			<channel id="sendcmd" typeId="sendcmd"/>
+			<channel id="button_press" typeId="button_press"/>
```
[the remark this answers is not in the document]
mlobstein: I renamed the new system channel to 'buttonpress' in all places that it is referenced.

bundles/org.openhab.binding.nuvo/README.md
mlobstein: good catch

bundles/org.openhab.binding.nuvo/README.md
mlobstein: 'utilized' is OK for English, but 'used' is simpler so I have changed it.

bundles/org.openhab.binding.nuvo/README.md
mlobstein: updated and tested as working.

lolodomo (approved):
LGTM, thank you

PR #13665, opened 2022-11-06
[groheondus] Upgrade dependencies

wborn:
* Upgrade commons-text to 1.10.0 (prevents CVE-2022-42889)
* Upgrade commons-lang3 to 3.12.0
* Remove commons-text, wrap from feature because it is embedded into the bundle


jlaur (approved):
Thanks!

PR #13667, opened 2022-11-06
[squeezebox] Fix bridge initialization when parameter `quoteList` is not configured

### LINKED ISSUE #13666: [squeezebox] Unable to initialize bridge
Not sure when this problem started, only noticed after upgrading to 3.3M4:
`Thing 'squeezebox:squeezeboxserver:server' changed from UNINITIALIZED (HANDLER_MISSING_ERROR) to UNINITIALIZED (HANDLER_CONFIGURATION_PENDING): {quoteList=The parameter is required.}`

I have no configuration for channel `favoritesList` and no items linked to it.

It might go all the way back to openhab/openhab-core#2691, although I would probably have noticed sooner. Perhaps it could be related to openhab/openhab-core#3024.

## Expected Behavior
Bridge should be able to initialize without any configuration for `favoritesList`.

## Current Behavior
Bridge cannot initialize.

## Possible Solution
It seems like parameter `quoteList` for channel `favoritesList` is incorrectly marked as **required**, although it has a default value:

```xml
	<channel-type id="favoritesList">
		<item-type>String</item-type>
		<label>Favorites List</label>
		<description>Comma-separated list of favorites of form favoriteId=favoriteName</description>
		<state readOnly="true" pattern="%s"></state>
		<config-description>
			<parameter name="quoteList" type="boolean" required="true">
				<label>Quote Favorites</label>
				<description>Wrap the right hand side of the favorites in quotes</description>
				<default>false</default>
			</parameter>
		</config-description>
	</channel-type>
```

## Steps to Reproduce
Add bridge and wait for initialization.

jlaur:
Fixes #13666

### DIFF
--- a/bundles/org.openhab.binding.squeezebox/src/main/resources/OH-INF/thing/thing-types.xml
+++ b/bundles/org.openhab.binding.squeezebox/src/main/resources/OH-INF/thing/thing-types.xml
@@ -121,7 +121,7 @@
 		<description>Comma-separated list of favorites of form favoriteId=favoriteName</description>
 		<state readOnly="true" pattern="%s"></state>
 		<config-description>
-			<parameter name="quoteList" type="boolean" required="true">
+			<parameter name="quoteList" type="boolean">
 				<label>Quote Favorites</label>
 				<description>Wrap the right hand side of the favorites in quotes</description>
 				<default>false</default>


mhilbush (approved):
LGTM

Thanks for the fix!

lolodomo (approved):
LGTM, thank you

PR #13669, opened 2022-11-06
[kodi] Fix bridge initialization when parameter `group` is not configured

### LINKED ISSUE #13668: [kodi] Unable to initialize bridge
Not sure when this problem started, only noticed after upgrading to 3.3M4:
`Thing 'kodi:kodi:home' changed from UNINITIALIZED (HANDLER_MISSING_ERROR) to UNINITIALIZED (HANDLER_CONFIGURATION_PENDING): {group=The parameter is required.}`

I have no configuration for channel `pvr-channel` and no items linked to it.

It might go all the way back to openhab/openhab-core#2691, although I would probably have noticed sooner. Perhaps it could be related to openhab/openhab-core#3024.

## Expected Behavior
Bridge should be able to initialize without any configuration for `group`.

## Current Behavior
Bridge cannot initialize.

## Possible Solution
It seems like parameter `group` for channel `pcvr-channel` is incorrectly marked as **required**, although it has a default value:

```xml
	<config-description uri="channel-type:kodi:pvr-channel">
		<parameter name="group" type="text" required="true">
			<label>PVR Channel Group</label>
			<description>The PVR channel group name used to identify the channel id.</description>
			<default>All channels</default>
		</parameter>
	</config-description>
```

## Steps to Reproduce
Add bridge and wait for initialization.

jlaur:
Fixes #13668

### DIFF
--- a/bundles/org.openhab.binding.kodi/src/main/resources/OH-INF/config/config.xml
+++ b/bundles/org.openhab.binding.kodi/src/main/resources/OH-INF/config/config.xml
@@ -64,7 +64,7 @@
 	</config-description>
 
 	<config-description uri="channel-type:kodi:pvr-channel">
-		<parameter name="group" type="text" required="true">
+		<parameter name="group" type="text">
 			<label>PVR Channel Group</label>
 			<description>The PVR channel group name used to identify the channel id.</description>
 			<default>All channels</default>


cweitkamp (approved):
LGTM. Thanks.

PR #13677, opened 2022-11-07
[tapocontrol] Fix configuration error

wildcs:
* fixed configuration error
* improved response logging and error handling

This PR fixes error reading configuration with integer values.
Configuration handling revised
Also tracing of sent and received http request was enhanced, because some unkown communication errors will occur sometimes. 

It has been discussed at [https://community.openhab.org/t/tapocontrol-control-tapo-smart-wifi-devices-with-openhab/118389/216](https://community.openhab.org/t/tapocontrol-control-tapo-smart-wifi-devices-with-openhab/118389/216).

A compiled jar is available at [https://github.com/wildcs/oh3_tapocontrol/tree/a82bd00a65c9fbcabbfa377258ebdffd3b649032](url)

### DIFF
--- a/bundles/org.openhab.binding.tapocontrol/src/main/java/org/openhab/binding/tapocontrol/internal/api/TapoDeviceHttpApi.java
+++ b/bundles/org.openhab.binding.tapocontrol/src/main/java/org/openhab/binding/tapocontrol/internal/api/TapoDeviceHttpApi.java
@@ -313,6 +313,8 @@ public class TapoDeviceHttpApi {
      */
     protected void sendAsyncRequest(String url, String payload, String command) {
         logger.trace("({}) sendAsncRequest to '{}' with cookie '{}'", uid, url, this.cookie);
+        logger.trace("({}) command/payload: '{}''{}'", uid, command, payload);
+
         try {
             Request httpRequest = bridge.getHttpClient().newRequest(url).method(HttpMethod.POST.toString());
 
@@ -344,22 +346,27 @@ public class TapoDeviceHttpApi {
                     } else {
                         /* request succesfull */
                         String rBody = getContentAsString();
-                        rBody = decryptResponse(rBody);
-                        logger.trace("({}) requestCompleted '{}'", uid, rBody);
-                        /* handle result */
-                        switch (command) {
-                            case DEVICE_CMD_SETINFO:
-                                handleSuccessResponse(rBody);
-                                break;
-                            case DEVICE_CMD_GETINFO:
-                                handleDeviceResult(rBody);
-                                break;
-                            case DEVICE_CMD_GETENERGY:
-                                handleEnergyResult(rBody);
-                                break;
-                            case DEVICE_CMD_CUSTOM:
-                                handleCustomResponse(rBody);
-                                break;
+                        logger.trace("({}) receivedRespose '{}'", uid, rBody);
+                        if (!hasErrorCode(rBody)) {
+                            rBody = decryptResponse(rBody);
+                            logger.trace("({}) decryptedResponse '{}'", uid, rBody);
+                            /* handle result */
+                            switch (command) {
+                                case DEVICE_CMD_SETINFO:
+                                    handleSuccessResponse(rBody);
+                                    break;
+                                case DEVICE_CMD_GETINFO:
+                                    handleDeviceResult(rBody);
+                                    break;
+                                case DEVICE_CMD_GETENERGY:
+                                    handleEnergyResult(rBody);
+                                    break;
+                                case DEVICE_CMD_CUSTOM:
+                                    handleCustomResponse(rBody);
+                                    break;
+                            }
+                        } else {
+                            getErrorCode(rBody);
                         }
                     }
                 }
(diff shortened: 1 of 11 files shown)


bundles/org.openhab.binding.tapocontrol/src/main/java/org/openhab/binding/tapocontrol/internal/api/TapoDeviceHttpApi.java
```diff
@@ -411,6 +418,26 @@ protected Integer getErrorCode(String responseBody) {
...
+     */
+    protected Boolean hasErrorCode(String responseBody) {
+        try {
+            JsonObject jsonObject = gson.fromJson(responseBody, JsonObject.class);
+            /* get errocode (0=success) */
+            Integer errorCode = jsonObjectToInt(jsonObject, "error_code", ERR_JSON_DECODE_FAIL);
+            if (errorCode > 0) {
+                return true;
+            }
+        } catch (Exception e) {
```
lolodomo, line 435: Please catch more precise exceptions, those that are thrown by gson.fromJson and jsonObjectToInt

bundles/org.openhab.binding.tapocontrol/src/main/java/org/openhab/binding/tapocontrol/internal/structures/TapoBridgeConfiguration.java
```diff
@@ -46,31 +40,4 @@ public final class TapoBridgeConfiguration {
     public Boolean udpDiscoveryEnabled = false;
     public Integer cloudReconnectIntervalM = CONFIG_CLOUD_FIXED_INTERVAL;
     public Integer discoveryIntervalM = 30;
```
lolodomo, line 42: Please update the default value of discoveryIntervalM to 60 as it is the default value set in the XML description.
And as the name of the parameter in the XML description is "discoveryInterval", you class member must have the same name. So you should remove the final "M".
And you can rather use int and boolean, rather than Boolean and Integer, at least for cloudDiscovery and discoveryInterval which are real config parameters with always a value.

udpDiscoveryEnabled and cloudReconnectIntervalM are not defined in the XML description, so it is a little surprising to find them as field in the configuration object. I have not checked how they are used by the binding. So I cannot be sure if boolean/int would be more appropriate and if constants should be rather used.

bundles/org.openhab.binding.tapocontrol/src/main/java/org/openhab/binding/tapocontrol/internal/structures/TapoDeviceConfiguration.java
```diff
@@ -26,38 +22,11 @@
...
 public final class TapoDeviceConfiguration {
-    private final Logger logger = LoggerFactory.getLogger(TapoDeviceConfiguration.class);
-
     /* THING CONFIGUTATION PROPERTYS */
     public static final String CONFIG_DEVICE_IP = "ipAddress";
     public static final String CONFIG_UPDATE_INTERVAL = "pollingInterval";
 
     /* thing configuration parameter. */
     public String ipAddress = "";
     public Integer pollingInterval = 30;
```
lolodomo, line 31: You could rather use int.

lolodomo (changes requested):

bundles/org.openhab.binding.tapocontrol/src/main/java/org/openhab/binding/tapocontrol/internal/structures/TapoBridgeConfiguration.java
lolodomo: This let me think that behaviour of config parameter discoveryInterval was even not tested. Please test that the value you setup in UI for your thing is properly propagated in the binding.

wildcs: You're right thx. Fixed and testet it now

bundles/org.openhab.binding.tapocontrol/src/main/java/org/openhab/binding/tapocontrol/internal/api/TapoDeviceHttpApi.java
wildcs: Just need value of Json-Field "errorCode" if exists. sometimes string is a decryptet non json text. Now im checking if it's json before

bundles/org.openhab.binding.tapocontrol/src/main/java/org/openhab/binding/tapocontrol/internal/structures/TapoBridgeConfiguration.java
wildcs: Now it really should be pushed. Everything looked clear already.

lolodomo (changes requested):
Additionally to my following comments, I am not sure why you are removing unit tests ?

lolodomo (changes requested):
Let's go for a last round.

PS: I will not investigate too much in your test units but I have a strange feeling that your new discovery class for tests is the same as the real class.

bundles/org.openhab.binding.tapocontrol/src/test/java/org/openhab/binding/tapocontrol/internal/structures/TapoBridgeConfiguration.java
```diff
@@ -26,52 +22,21 @@
...
-            cloudDiscoveryEnabled = Boolean.parseBoolean(config.get(CONFIG_DISCOVERY_CLOUD).toString());
-            udpDiscoveryEnabled = Boolean.parseBoolean(config.get(CONFIG_DISCOVERY_UDP).toString());
-            cloudReconnectIntervalM = Integer.valueOf(config.get(CONFIG_CLOUD_UPDATE_INTERVAL).toString());
-            discoveryIntervalM = Integer.valueOf(config.get(CONFIG_DISCOVERY_INTERVAL).toString());
-        } catch (Exception e) {
-            logger.warn("{} error reading configuration: '{}'", bridge.getUID(), e.getMessage());
-        }
-    }
+    public boolean cloudDiscovery = false;
+    public boolean udpDiscovery = false;
```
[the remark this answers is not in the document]
wildcs: i tried to integrate device discovery by udp but i didn't complete it until now.

PR #13680, opened 2022-11-08
[tibber] Updates for breaking API changes

### LINKED ISSUE #13648: [tibber] Breaking change in Tibber API websockets
Hi!

I would like to thank you for contributing to the community of Tibber users! We've added you to the list of people we respect! https://developer.tibber.com/respect

In order to improve the API resilience we are making a breaking change for websocket communication. Please see details at [developer site](https://developer.tibber.com/). Would you be able to adjust the implementation accordingly?

Also, it looks like you are not setting user-agent header value. A recommended format would be to include both the platform version and driver version for example like this: OpenHAB/1.2.3 tibber_driver_version/1.2.3'

We're trying to collect best practices [here](https://developer.tibber.com/docs/guides/calling-api)

Let us know if in need of any assistance!

Br,
Toni
Tibber

kjoglum:
Closes #13648 

Signed-off-by: kjoglum <stiankj@online.no>

### DIFF
--- a/bundles/org.openhab.binding.tibber/src/main/java/org/openhab/binding/tibber/internal/handler/TibberHandler.java
+++ b/bundles/org.openhab.binding.tibber/src/main/java/org/openhab/binding/tibber/internal/handler/TibberHandler.java
@@ -26,6 +26,8 @@ import java.util.concurrent.TimeUnit;
 
 import org.eclipse.jdt.annotation.NonNullByDefault;
 import org.eclipse.jdt.annotation.Nullable;
+import org.eclipse.jetty.client.HttpClient;
+import org.eclipse.jetty.http.HttpHeader;
 import org.eclipse.jetty.util.ssl.SslContextFactory;
 import org.eclipse.jetty.websocket.api.Session;
 import org.eclipse.jetty.websocket.api.annotations.OnWebSocketClose;
@@ -50,6 +52,7 @@ import org.openhab.core.thing.ThingStatusInfo;
 import org.openhab.core.thing.binding.BaseThingHandler;
 import org.openhab.core.types.Command;
 import org.openhab.core.types.RefreshType;
+import org.osgi.framework.FrameworkUtil;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -76,6 +79,8 @@ public class TibberHandler extends BaseThingHandler {
     private @Nullable ScheduledFuture<?> pollingJob;
     private @Nullable Future<?> sessionFuture;
     private String rtEnabled = "false";
+    private @Nullable String subscriptionURL;
+    private @Nullable String versionString;
 
     public TibberHandler(Thing thing) {
         super(thing);
@@ -85,6 +90,9 @@ public class TibberHandler extends BaseThingHandler {
     public void initialize() {
         updateStatus(ThingStatus.UNKNOWN);
         tibberConfig = getConfigAs(TibberConfiguration.class);
+
+        versionString = FrameworkUtil.getBundle(this.getClass()).getVersion().toString();
+        logger.debug("Binding version: {}", versionString);
 
         getTibberParameters();
         startRefresh(tibberConfig.getRefresh());
@@ -104,7 +112,9 @@ public class TibberHandler extends BaseThingHandler {
         try {
             httpHeader.put("cache-control", "no-cache");
             httpHeader.put("content-type", JSON_CONTENT_TYPE);
-            httpHeader.put("Authorization", "Bearer " + tibberConfig.getToken());
+            httpHeader.put(HttpHeader.USER_AGENT.asString(),
+                    "openHAB/Tibber " + versionString + " Tibber driver " + TIBBER_DRIVER);
+            httpHeader.put(HttpHeader.AUTHORIZATION.asString(), "Bearer " + tibberConfig.getToken());
 
             TibberPriceConsumptionHandler tibberQuery = new TibberPriceConsumptionHandler();
             InputStream connectionStream = tibberQuery.connectionInputStream(tibberConfig.getHomeid());
@@ -124,6 +134,15 @@ public class TibberHandler extends BaseThingHandler {
 
                 if ("true".equals(rtEnabled)) {
                     logger.debug("Pulse associated with HomeId: Live stream will be started");
+
+                    InputStream wsURL = tibberQuery.getWebsocketUrl();
+                    String wsResponse = HttpUtil.executeUrl("POST", BASE_URL, httpHeader, wsURL, null, REQUEST_TIMEOUT);
+
+                    JsonObject wsobject = (JsonObject) JsonParser.parseString(wsResponse);
+                    subscriptionURL = wsobject.getAsJsonObject("data").getAsJsonObject("viewer")
+                            .get("websocketSubscriptionUrl").toString().replaceAll("^\"|\"$", "");
+                    logger.debug("Subscribing to: {}", subscriptionURL);
+
                     open();
                 } else {
                     logger.debug("No Pulse associated with HomeId: No live stream will be started");
@@ -318,7 +337,7 @@ public class TibberHandler extends BaseThingHandler {
             sslContextFactory.setTrustAll(true);
             sslContextFactory.setEndpointIdentificationAlgorithm(null);
 
-            client = new WebSocketClient(sslContextFactory);
+            client = new WebSocketClient(new HttpClient(sslContextFactory));
             client.setMaxIdleTimeout(30 * 1000);
             this.client = client;
 
@@ -330,8 +349,10 @@ public class TibberHandler extends BaseThingHandler {
             }
 
             ClientUpgradeRequest newRequest = new ClientUpgradeRequest();
-            newRequest.setHeader("Authorization", "Bearer " + tibberConfig.getToken());
-            newRequest.setSubProtocols("graphql-subscriptions");
+            newRequest.setHeader(HttpHeader.USER_AGENT.asString(),
+                    "openHAB/Tibber " + versionString + " Tibber driver " + TIBBER_DRIVER);
+            newRequest.setHeader(HttpHeader.AUTHORIZATION.asString(), "Bearer " + tibberConfig.getToken());
+            newRequest.setSubProtocols("graphql-transport-ws");
 
             try {
                 logger.debug("Starting Websocket connection");
@@ -341,7 +362,7 @@ public class TibberHandler extends BaseThingHandler {
             }
             try {
                 logger.debug("Connecting Websocket connection");
-                sessionFuture = client.connect(socket, new URI(SUBSCRIPTION_URL), newRequest);
+                sessionFuture = client.connect(socket, new URI(subscriptionURL), newRequest);
                 try {
                     Thread.sleep(10 * 1000);
                 } catch (InterruptedException e) {
@@ -412,7 +433,8 @@ public class TibberHandler extends BaseThingHandler {
         public void onConnect(Session wssession) {
             TibberHandler.this.session = wssession;
             TibberWebSocketListener socket = TibberHandler.this.socket;
-            String connection = "{\"type\":\"connection_init\", \"payload\":\"token=" + tibberConfig.getToken() + "\"}";
+            String connection = "{\"type\":\"connection_init\", \"payload\":{\"token\":\"" + tibberConfig.getToken()
+                    + "\"}}";
             try {
                 if (socket != null) {
                     logger.debug("Sending websocket connect message");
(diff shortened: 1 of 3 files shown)


bundles/org.openhab.binding.tibber/src/main/java/org/openhab/binding/tibber/internal/handler/TibberHandler.java
```diff
@@ -104,6 +105,7 @@ public void getTibberParameters() {
         try {
             httpHeader.put("cache-control", "no-cache");
             httpHeader.put("content-type", JSON_CONTENT_TYPE);
+            httpHeader.put("user-agent", "openHAB/3.x com.tibber/1.8.3");
```
Flole998, line 108: Is there no way of getting the openHAB/binding version dynamically at runtime? This would need touching at major version upgrades aswell as on every change of this binding, that is not ideal.

bundles/org.openhab.binding.tibber/src/main/java/org/openhab/binding/tibber/internal/handler/TibberHandler.java
```diff
@@ -330,8 +341,9 @@ public void open() {
             }
 
             ClientUpgradeRequest newRequest = new ClientUpgradeRequest();
+            newRequest.setHeader("user-agent", "openHAB/3.4 com.tibber/1.8.3");
```
Flole998, line 344: Same as above, maybe even turn this into a separate method so there's only a single change needed and not multiple values all over the place...

bundles/org.openhab.binding.tibber/src/main/java/org/openhab/binding/tibber/internal/handler/TibberHandler.java
```diff
@@ -412,7 +424,8 @@ public class TibberWebSocketListener {
         public void onConnect(Session wssession) {
             TibberHandler.this.session = wssession;
             TibberWebSocketListener socket = TibberHandler.this.socket;
-            String connection = "{\"type\":\"connection_init\", \"payload\":\"token=" + tibberConfig.getToken() + "\"}";
+            String connection = "{\"type\":\"connection_init\", \"payload\":{\"token\":\"" + tibberConfig.getToken()
```
Flole998, line 427: This looks wild. Any reason to not use GSON here to build the json object and then run toString on it? If the token contains special chars this would break the JSON.

bundles/org.openhab.binding.tibber/src/main/java/org/openhab/binding/tibber/internal/handler/TibberHandler.java
```diff
@@ -530,7 +543,7 @@ private void sendMessage(String message) throws IOException {
         }
 
         public void startSubscription() {
-            String query = "{\"id\":\"1\",\"type\":\"start\",\"payload\":{\"variables\":{},\"extensions\":{},\"operationName\":null,\"query\":\"subscription {\\n liveMeasurement(homeId:\\\""
+            String query = "{\"id\":\"1\",\"type\":\"subscribe\",\"payload\":{\"variables\":{},\"extensions\":{},\"operationName\":null,\"query\":\"subscription {\\n liveMeasurement(homeId:\\\""
```
Flole998, line 546: Same as above, this looks even more wild and I am not sure if this can be built with GSON.

bundles/org.openhab.binding.tibber/src/main/java/org/openhab/binding/tibber/internal/handler/TibberHandler.java
kjoglum: Suggestion how to would be appreciated.

Flole998: You can learn from some of the other bindings, a quick search for "user-agent" leads to this:
https://github.com/openhab/openhab-addons/blob/cdff5a06e7e9b719b9d3bc484c52012eeed30ba2/bundles/org.openhab.binding.tankerkoenig/src/main/java/org/openhab/binding/tankerkoenig/internal/handler/WebserviceHandler.java#L89

I assume you only need to find a way to get the openHAB version there aswell.... Maybe you'll find some example aswell.

bundles/org.openhab.binding.tibber/src/main/java/org/openhab/binding/tibber/internal/handler/TibberHandler.java
kjoglum: Following [Tibber .NET SDK](https://github.com/tibber/Tibber.SDK.NET/commit/4beb5e691b866229f7b4adced64708f0cedada15#diff-43fedf3288bf7eb7404444803d406886bd517a4ed366f890a641fbaec11df5a4R199) it is the messages required for connection initalization and subscription, and it is confirmed to work successfully.

Flole998: If it's valid JSON it would be better for readability and maintainability to build the JSON using GSON. That's just my opinion, maybe the maintainers who review this have a different opinion on that

kjoglum: I see your point wrt code "tidiness", but struggle to see why a sudden change in code would be required, potentially also breaking working code, as the "as is" code has been verified working over years. Also taking into account how the API actually is set up.

bundles/org.openhab.binding.tibber/src/main/java/org/openhab/binding/tibber/internal/handler/TibberHandler.java
kjoglum: Added dynamic bundle version for "user-agent".

bundles/org.openhab.binding.tibber/src/main/java/org/openhab/binding/tibber/internal/handler/TibberHandler.java
kjoglum: Added dynamic bundle version for "user-agent".

bundles/org.openhab.binding.tibber/src/main/java/org/openhab/binding/tibber/internal/handler/TibberHandler.java
```diff
@@ -101,9 +104,12 @@ public void handleCommand(ChannelUID channelUID, Command command) {
 
     public void getTibberParameters() {
         String response = "";
+        versionString = FrameworkUtil.getBundle(this.getClass()).getVersion().toString();
```
jlaur, line 107: Can this be initialized in `initialize()`? This would make method `open()` independent of this method in terms of order of initialization.

bundles/org.openhab.binding.tibber/src/main/java/org/openhab/binding/tibber/internal/handler/TibberHandler.java
```diff
@@ -101,9 +104,12 @@ public void handleCommand(ChannelUID channelUID, Command command) {
...
+        logger.debug("Binding version: {}", versionString);
         try {
             httpHeader.put("cache-control", "no-cache");
             httpHeader.put("content-type", JSON_CONTENT_TYPE);
+            httpHeader.put("user-agent", "openHAB/Tibber " + versionString + " com.tibber/1.8.3");
```
jlaur, line 112: Can version number "1.8.3" be extracted to a constant? What is the origin of this - is it API version?

bundles/org.openhab.binding.tibber/src/main/java/org/openhab/binding/tibber/internal/handler/TibberHandler.java
```diff
@@ -330,8 +345,9 @@ public void open() {
             }
```
jlaur, line 345: Optional, but now that you are at it and presumably testing the binding also: Would you consider getting rid of deprecated WebSocketClient constructor above in line 336? I believe it could be rewritten to:

```java
client = new WebSocketClient(new HttpClient(sslContextFactory));
```

bundles/org.openhab.binding.tibber/src/main/java/org/openhab/binding/tibber/internal/handler/TibberHandler.java
```diff
@@ -330,8 +345,9 @@ public void open() {
...
 
             ClientUpgradeRequest newRequest = new ClientUpgradeRequest();
+            newRequest.setHeader("user-agent", "openHAB/Tibber " + versionString + " com.tibber/1.8.3");
             newRequest.setHeader("Authorization", "Bearer " + tibberConfig.getToken());
```
jlaur, line 349: Consider:
```suggestion
            newRequest.setHeader(HttpHeader.USER_AGENT.asString(), "openHAB/Tibber " + versionString + " com.tibber/1.8.3");
            newRequest.setHeader(HttpHeader.AUTHORIZATION.asString(), "Bearer " + tibberConfig.getToken());
```

jlaur (changes requested):
Thank you for taking care of this! I have added a few minor suggestions.

bundles/org.openhab.binding.tibber/src/main/java/org/openhab/binding/tibber/internal/handler/TibberHandler.java
kjoglum: Good point, moved to `initialize()`.

bundles/org.openhab.binding.tibber/src/main/java/org/openhab/binding/tibber/internal/handler/TibberHandler.java
kjoglum: Yes, it is the API version. No known way to retrieve the API version dynamically, whereas the only reference is the Tibber Driver version (`com.tibber/1.8.3`) referred to in the API breaking changes documentation. So, my guess is that the Tibber Driver version will remain such until a new/potential breaking change will happen.

bundles/org.openhab.binding.tibber/src/main/java/org/openhab/binding/tibber/internal/handler/TibberHandler.java
kjoglum: Corrected according to suggestion, and tested/confirmed ok.

bundles/org.openhab.binding.tibber/src/main/java/org/openhab/binding/tibber/internal/handler/TibberHandler.java
kjoglum: Updated.

bundles/org.openhab.binding.tibber/src/main/java/org/openhab/binding/tibber/internal/handler/TibberHandler.java
```diff
@@ -104,6 +112,7 @@ public void getTibberParameters() {
         try {
             httpHeader.put("cache-control", "no-cache");
             httpHeader.put("content-type", JSON_CONTENT_TYPE);
+            httpHeader.put("user-agent", "openHAB/Tibber " + versionString + " com.tibber/1.8.3");
             httpHeader.put("Authorization", "Bearer " + tibberConfig.getToken());
```
jlaur, line 116: Sorry, missed this place:
```suggestion
            httpHeader.put(HttpHeader.USER_AGENT.asString(), "openHAB/Tibber " + versionString + " com.tibber/1.8.3");
            httpHeader.put(HttpHeader.AUTHORIZATION.asString(), "Bearer " + tibberConfig.getToken());
```

bundles/org.openhab.binding.tibber/src/main/java/org/openhab/binding/tibber/internal/handler/TibberHandler.java
jlaur: OK. Still I would extract it as constant to be reused both places so they will not grow different over time.

kjoglum: Moved to binding constants.

bundles/org.openhab.binding.tibber/src/main/java/org/openhab/binding/tibber/internal/handler/TibberHandler.java
jlaur: I agree with @Flole998's point that it would be safer to let GSON serialize this into JSON, but since it's not new code, this refactoring is not needed for this PR.

jlaur (approved):
LGTM - thanks!

PR #13685, opened 2022-11-09
[regoheatpump] fixed polling stops problem if there is an unhandled runtime exception

crnjan:
If for some reason HP response fails to properly parse, i.e. `java.lang.NumberFormatException: For input string: "##"` exception is not handled and scheduler is not re-triggered (polling stops). Also fixed build warnings.

Signed-off-by: Boris Krivonog [boris.krivonog@inova.si](mailto:boris.krivonog@inova.si)


lolodomo (approved):
LGTM, thank you

PR #13688, opened 2022-11-09
[monopriceaudio] Update README

mlobstein:
This PR updates the README to clarify the supported thing types. It also adds a link to the OH Marketplace version of the binding that provides support for additional thing types. The intent is to inform users who may not otherwise know about the marketplace version and hopefully get some feedback about the beta functionality.

### DIFF
--- a/bundles/org.openhab.binding.monopriceaudio/README.md
+++ b/bundles/org.openhab.binding.monopriceaudio/README.md
@@ -18,8 +18,12 @@ Or you can connect it for example to a Raspberry Pi and use [ser2net Linux tool]
 
 ## Supported Things
 
-There is exactly one supported thing type, which represents the amplifier controller.
-It has the `amplifier` id.
+Monoprice 10761 & 39261 and Dayton Audio DAX66 Amplifiers use the `amplifier` thing id. Up to 18 zones with 3 linked amps, 6 source inputs.
+Note: Compatible clones (including 4 zone versions) from McLELLAND, Factor, Soundavo, etc. should work as well.
+
+Support for additional amplifier types from Monoprice, Dayton and Xantech is currently available via the
+[OH Marketplace beta version of the binding](https://community.openhab.org/t/monoprice-dayton-audio-xantech-whole-house-audio-binding-beta/128924).
+*Please report any successes or failures using the beta version of the binding so that the issues can be addressed and the beta code merged into the official version.*
 
 ## Discovery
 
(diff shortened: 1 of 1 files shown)


bundles/org.openhab.binding.monopriceaudio/README.md
```diff
@@ -18,8 +18,12 @@ Or you can connect it for example to a Raspberry Pi and use [ser2net Linux tool]
...
 ## Supported Things
 
-There is exactly one supported thing type, which represents the amplifier controller.
-It has the `amplifier` id.
+Monoprice 10761 & 39261 and Dayton Audio DAX66 Amplifiers use the `amplifier` thing id. Up to 18 zones with 3 linked amps, 6 source inputs.
+Note: Compatible clones (including 4 zone versions) from McLELLAND, Factor, Soundavo, etc. should work as well.
+
+Support for additional amplifier types from Monoprice, Dayton and Xantech is currently available via the
+[OH Marketplace beta version of the binding](https://community.openhab.org/t/monoprice-dayton-audio-xantech-whole-house-audio-binding-beta/128924).
+*Please report any successes or failures using the beta version of the binding so that the issues can be addressed and the beta code merged into the official version.*
```
lolodomo, line 26: Looks really inappropriate !
The official binding will not reference non official changes !
The good process is to provide a PR to integrate this new feature in the "official" binding. This PR will of course integrate any expected change in README.
In case you expect a feedback from a beta change, please ask it in the community forum.

lolodomo:
And you are of course very welcome to submit your changes in that repo through a new PR.

bundles/org.openhab.binding.monopriceaudio/README.md
mlobstein: I did not expect such a strong reaction. This was not an attempt to promote the marketplace ahead of the official version. The beta version was to be referenced due to the following reasons:

1. I do not own any of the `things` that the beta version adds support for; The coding was done 'blind' with no testing whatsoever.
2. There has been no community feedback in the year since the beta version was published.

Via the official binding documentation, I had hoped to raise awareness of the beta functionality to new/potential users who may only scan the documentation when deciding whether or not to choose openHAB as their home automation solution. Since this is not allowed, I have removed the reference.

lolodomo: I have myself, like others, push binding code for things I am not owning myself. For example, the Rotel binding is supporting many different Rotel models, I have myself only one particular model. So the binding integrates features I was never able to test myself with a real device.

So, if you are confident in your implementation (full implementation), I think you can already create a PR. In that case, you have to be clear in README that the binding was only tested with model XXX even if it should also work for models YYY.

lolodomo (approved):

Issue #13282, opened 2022-08-17
[iCloud] Bridge-Thing generates Communication Error


RalphSester:
The Brigde-Thing (iCloud-Account) does not go online.

It gets the state: COMMUNICATION_ERROR with the following details information:

`java.util.concurrent.ExecutionException: org.eclipse.jetty.client.HttpResponseException: HTTP protocol violation: Authentication challenge without WWW-Authenticate header
`

The binding is not usable completly at this time.

## Current Behavior
The Bridge-Thing does not come online.
Restart of binding or OH doesn't help.

## Possible Solution
It seems, that apple has changed something in their api.

## Steps to Reproduce (for Bugs)
1. Just start the binding.
2. See the error

## Environment

* Version used: OpenHAB 3.3 / 3.4
* Operating System: Windows Server 2016

BigGeorgeTx:
This is related to this https://github.com/openhab/openhab-addons/issues/13231 previously reported issue, but it appears to have gotten worse.  At least for me, A disable/enable cycle, restart of the binding, or restart of openHAB no longer fixes the issue.  From the comments in the openHAB community, https://community.openhab.org/t/icloud-binding-communication-error/122131/336?u=biggeorgetx, this is impacting many users.

JensHoRi:
Maybe this information helps:
https://github.com/jetty-project/jetty-reactive-httpclient/issues/33

If I read it right (I' no developer), looks like apple changed something in authentication and is not compliant to specifications anymore.
Under that link, they had same issue and discussing a workaround. Looks like doing the authentication handling in code directly could be a temporary good solution. But I'm no developer, maybe I'm wrong. Just try to help

BigGeorgeTx:
This issue returned today for many users. https://community.openhab.org/t/icloud-binding-communication-error/122131/327?u=biggeorgetx

RalphSester:
It seems, that the iCloud-Binding has no developer since a couple of months.
Nobody seems to care, that the binding is completly unuseable.....

if this problem can't be fixed soon,  it would be better to remove it from OH 3.4 !!

@digitaldan: Do you have some ideas?

martinvw:
I'm not very active currently as a maintainer, and will most likely not be so in the near future because of a planned move, but fixing such issues in an open-source project is not the sole responsibility of the binding maintainers, everyone is allowed, and should make a PR to fix this issue if they are capable to do so.

[...]

PR #12824, opened 2022-05-27
[hdpowerview] Fix output to console from tests and eliminate warnings

jlaur:
This pull request contains two test improvements, each committed separately:
- Avoid logging to console from test run (*). This was made possible by openhab/openhab-core#2982. Previous attempt is described [here](https://github.com/openhab/openhab-addons/pull/12061#issuecomment-1014879948), but required **mockito-inline** for static mocks. The console output was originally introduced by #11853.
- Fix warnings about redundant null checks. This might have been a regression from #12049.

*: The output looked like this:
```
[INFO] Running org.openhab.binding.hdpowerview.AutomationChannelBuilderTest
[main] WARN  o.o.b.h.i.b.AutomationChannelBuilder - Scheduled event '1' references scene '1', but no scenes are loaded
[main] WARN  o.o.b.h.i.b.AutomationChannelBuilder - Scheduled event '1' references scene collection '2', but no scene collections are loaded
```

PR #13691, opened 2022-11-10
[icloud] Rework authentication to reflect changes in iCloud API

maihacke:
# [icloud] Rework authentication to reflect changes in iCloud API

# Description

Apple changed the iCloud API which prevents the existing binding from working (https://github.com/openhab/openhab-addons/issues/13282). This PR reworks the authentication mechanism to make the binding work again.

# Testing

Your pull request will automatically be built and available under the following folder:
https://openhab.jfrog.io/ui/native/libs-pullrequest-local/org/openhab/addons/bundles/

It is a good practice to add a URL to your built JAR in this pull request description,
so it is easier for the community to test your Add-on.
If your pull request contains a new binding, it will likely take some time
before it is reviewed and processed by maintainers.
That said, consider submitting your Add-on in the Marketplace:
https://community.openhab.org/c/marketplace/69

Don't forget to submit a thread about your Add-on in the openHAB community:
https://community.openhab.org/c/add-ons 
-->


maihacke:
Nevertheless this PR could use some more testing, first results are promising (see https://community.openhab.org/t/icloud-binding-communication-error/122131/424).
I would be glad for some early review.

pgfeller:
Hi @maihacke,

thank you for the contribution! Unfortunately I can not do the review of the code -- I do not use apple devices anymore and it is long ago that I worked on this plugin (or a Java project in general).

Therefore it does not make sense that I'm listed as code owner anymore. I hope the plugin will find a new owner.

with kind regards,
Patrik

maihacke:
@jlaur : I saw you merged https://github.com/openhab/openhab-addons/pull/13766. Feature freeze for 3.4 is on 11.12. Is there anything I can do to have this PR merged

bundles/org.openhab.binding.icloud/src/main/java/org/openhab/binding/icloud/internal/handler/ICloudAccountBridgeHandler.java
```diff
@@ -88,85 +101,281 @@ public void handleCommand(ChannelUID channelUID, Command command) {
...
+
         super.handleRemoval();
     }
 
     @Override
     public void dispose() {
-        if (refreshJob != null) {
-            refreshJob.cancel(true);
+        if (this.refreshJob != null) {
+            this.refreshJob.cancel(true);
```
jlaur, line 261: Why this change? Probably there was already a null annotation warning, which could be fixed like this:
```suggestion
        ScheduledFuture<?> refreshJob = this.refreshJob;
        if (refreshJob != null) {
            refreshJob.cancel(true);
            this.refreshJob = null;
```

bundles/org.openhab.binding.icloud/src/main/java/org/openhab/binding/icloud/internal/handler/ICloudAccountBridgeHandler.java
```diff
@@ -88,85 +101,281 @@ public void handleCommand(ChannelUID channelUID, Command command) {
...
+    public void findMyDevice(String deviceId) throws IOException, InterruptedException {
+        callApiWithRetryAndExceptionHandling(() -> {
+            this.iCloudService.getDevices().playSound(deviceId);
+            return null;
+        });
     }
 
     public void registerListener(ICloudDeviceInformationListener listener) {
-        deviceInformationListeners.add(listener);
+
```
jlaur, line 274: Why was an empty line added here?

bundles/org.openhab.binding.icloud/src/main/java/org/openhab/binding/icloud/internal/handler/ICloudAccountBridgeHandler.java
```diff
@@ -88,85 +101,281 @@ public void handleCommand(ChannelUID channelUID, Command command) {
...
+        this.deviceInformationListeners.add(listener);
     }
 
     public void unregisterListener(ICloudDeviceInformationListener listener) {
-        deviceInformationListeners.remove(listener);
+
```
jlaur, line 279: Why was an empty line added here?

bundles/org.openhab.binding.icloud/src/main/java/org/openhab/binding/icloud/internal/handler/ICloudAccountBridgeHandler.java
```diff
@@ -88,85 +101,281 @@ public void handleCommand(ChannelUID channelUID, Command command) {
...
-                return;
+                        "Please provide 2-FA code in thing configuration.");
+                return false;
+            }
+
+            if (!this.iCloudService.isTrustedSession()) {
+                logger.debug("Trying to establish session trust.");
+                success = this.iCloudService.trustSession();
+                if (!success) {
+                    logger.warn("iCloud trust session failed.");
```
jlaur, line 343: This logging is not needed as the framework already logs the status change below.

bundles/org.openhab.binding.icloud/src/main/java/org/openhab/binding/icloud/internal/handler/ICloudAccountBridgeHandler.java
```diff
@@ -88,85 +101,281 @@ public void handleCommand(ChannelUID channelUID, Command command) {
...
-            logger.debug("iCloud bridge handler started.");
-        } catch (URISyntaxException e) {
-            logger.debug("Something went wrong while constructing the connection object", e);
+            authState = AuthState.AUTHENTICATED;
+            updateStatus(ThingStatus.ONLINE);
+            logger.debug("iCloud bridge handler authenticated.");
+            return true;
+
+        } catch (IOException | InterruptedException e) {
+            logger.warn("iCloud authentication caused exception.", e);
```
jlaur, line 355: This could be reduced to **debug** or **trace** level (to have full stack trace), since the framework will already log the exception message from the status change below.

bundles/org.openhab.binding.icloud/src/main/java/org/openhab/binding/icloud/internal/handler/ICloudDeviceHandler.java
```diff
@@ -56,18 +57,22 @@
 @NonNullByDefault
 public class ICloudDeviceHandler extends BaseThingHandler implements ICloudDeviceInformationListener {
     private final Logger logger = LoggerFactory.getLogger(ICloudDeviceHandler.class);
+
     private @Nullable String deviceId;
+
     private @Nullable ICloudAccountBridgeHandler icloudAccount;
 
     public ICloudDeviceHandler(Thing thing) {
+
```
jlaur, line 66: Why was this empty line added?

bundles/org.openhab.binding.icloud/src/main/java/org/openhab/binding/icloud/internal/handler/ICloudDeviceHandler.java
```diff
@@ -91,35 +96,37 @@ public void deviceInformationUpdate(List<ICloudDeviceInformation> deviceInformat
...
 
-    private void refreshData() {
-        ICloudAccountBridgeHandler bridge = getIcloudAccount();
         if (bridge != null) {
-            bridge.refreshData();
+            ((ICloudAccountBridgeHandler) bridge.getHandler()).registerListener(this);
+        }
+
+        if (bridge == null || bridge.getStatus() == ThingStatus.UNINITIALIZED) {
+            updateStatus(ThingStatus.UNKNOWN, ThingStatusDetail.BRIDGE_UNINITIALIZED);
```
jlaur, line 111: If bridge is null, this would be a configuration error, i.e. OFFLINE + CONFIGURATION_ERROR. If bridge is uninitialized it will not try to initialize child things, so this doesn't need to be handled explicitly.

bundles/org.openhab.binding.icloud/src/main/java/org/openhab/binding/icloud/internal/handler/ICloudDeviceHandler.java
```diff
@@ -91,35 +96,37 @@ public void deviceInformationUpdate(List<ICloudDeviceInformation> deviceInformat
...
+        } else if (bridge.getStatus() == ThingStatus.OFFLINE) {
+            updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.BRIDGE_OFFLINE);
```
jlaur, line 113: This should happen automatically already.

bundles/org.openhab.binding.icloud/src/main/java/org/openhab/binding/icloud/internal/handler/ICloudDeviceHandler.java
```diff
@@ -147,14 +154,16 @@ public void handleCommand(ChannelUID channelUID, Command command) {
 
     @Override
     public void dispose() {
-        ICloudAccountBridgeHandler bridge = getIcloudAccount();
+
```
jlaur, line 157: Why was this empty line added?

bundles/org.openhab.binding.icloud/src/main/java/org/openhab/binding/icloud/internal/handler/ICloudDeviceHandler.java
```diff
@@ -147,14 +154,16 @@ public void handleCommand(ChannelUID channelUID, Command command) {
...
+        Bridge bridge = getBridge();
         if (bridge != null) {
-            bridge.unregisterListener(this);
+            ((ICloudAccountBridgeHandler) bridge.getHandler()).unregisterListener(this);
```
jlaur, line 160: You could probably even do this:
```java
ThingHandler bridgeHandler = bridge.getHandler();
if (bridgeHandler instanceof ICloudAccountBridgeHandler)
{
    ((ICloudAccountBridgeHandler) bridgeHandler).unregisterListener(this);
}
```

to have both null and type check at the same time.

bundles/org.openhab.binding.icloud/src/main/java/org/openhab/binding/icloud/internal/handler/ICloudDeviceHandler.java
```diff
@@ -169,23 +178,25 @@ private void updateLocationRelatedStates(ICloudDeviceInformation deviceInformati
 
     private @Nullable ICloudDeviceInformation getDeviceInformationRecord(
             List<ICloudDeviceInformation> deviceInformationList) {
-        logger.debug("Device: [{}]", deviceId);
+
```
jlaur, line 181: Unneeded empty line.

bundles/org.openhab.binding.icloud/src/main/java/org/openhab/binding/icloud/internal/handler/ICloudDeviceHandler.java
```diff
@@ -169,23 +178,25 @@ private void updateLocationRelatedStates(ICloudDeviceInformation deviceInformati
...
+        this.logger.debug("Device: [{}]", this.deviceId);
 
         for (ICloudDeviceInformation deviceInformationRecord : deviceInformationList) {
             String currentId = deviceInformationRecord.getId();
 
-            logger.debug("Current data element: [id = {}]", currentId);
+            this.logger.debug("Current data element: [id = {}]", currentId);
```
jlaur, line 187: Unneeded change.

bundles/org.openhab.binding.icloud/src/main/java/org/openhab/binding/icloud/internal/handler/ICloudDeviceHandler.java
```diff
@@ -169,23 +178,25 @@ private void updateLocationRelatedStates(ICloudDeviceInformation deviceInformati
...
 
-            if (currentId != null && currentId.equals(deviceId)) {
+            if (currentId != null && currentId.equals(this.deviceId)) {
                 return deviceInformationRecord;
             }
         }
 
-        logger.debug("Unable to find device data.");
+        this.logger.debug("Unable to find device data.");
```
jlaur, line 194: Unneeded change.

bundles/org.openhab.binding.icloud/src/main/java/org/openhab/binding/icloud/internal/handler/ICloudDeviceHandler.java
```diff
@@ -169,23 +178,25 @@ private void updateLocationRelatedStates(ICloudDeviceInformation deviceInformati
...
         return null;
     }
 
     private State getLastLocationUpdateDateTimeState(ICloudDeviceInformation deviceInformationRecord) {
+
```
jlaur, line 199: Unneeded empty line.

bundles/org.openhab.binding.icloud/src/main/java/org/openhab/binding/icloud/internal/handler/ICloudDeviceHandler.java
```diff
@@ -169,23 +178,25 @@ private void updateLocationRelatedStates(ICloudDeviceInformation deviceInformati
...
+        this.logger.debug("Device: [{}]", this.deviceId);
 
         for (ICloudDeviceInformation deviceInformationRecord : deviceInformationList) {
             String currentId = deviceInformationRecord.getId();
 
-            logger.debug("Current data element: [id = {}]", currentId);
+            this.logger.debug("Current data element: [id = {}]", currentId);
 
-            if (currentId != null && currentId.equals(deviceId)) {
+            if (currentId != null && currentId.equals(this.deviceId)) {
```
jlaur, line 189: Unneeded change.

bundles/org.openhab.binding.icloud/src/main/java/org/openhab/binding/icloud/internal/utilities/CustomCookieStore.java
```diff
@@ -0,0 +1,111 @@
...
+import org.slf4j.LoggerFactory;
+
+import com.google.gson.Gson;
+import com.google.gson.GsonBuilder;
+
+/**
+ *
+ * This class implements a customized {@link CookieStore}. Its purpose is to add hyphens at the beginning and end of
+ * each cookie value which is required by Apple iCloud API.
+ *
```
jlaur, line 32: Very minor: Remove this line.

bundles/org.openhab.binding.icloud/src/main/java/org/openhab/binding/icloud/internal/utilities/CustomCookieStore.java
```diff
@@ -0,0 +1,111 @@
...
+ */
+public class CustomCookieStore implements CookieStore {
+
+    private CookieStore cookieStore;
+
+    private final Gson gson = new GsonBuilder().create();
+
+    private Storage<String> stateStorage;
+
+    private final static Logger LOGGER = LoggerFactory.getLogger(CustomCookieStore.class);
```
jlaur, line 44: Logger should be non-static.

bundles/org.openhab.binding.icloud/src/main/java/org/openhab/binding/icloud/internal/utilities/CustomCookieStore.java
```diff
@@ -0,0 +1,111 @@
...
+
+    /**
+     * The constructor.
+     *
+     */
+    public CustomCookieStore() {
+
```
jlaur, line 51: Unneeded empty line.

bundles/org.openhab.binding.icloud/src/main/java/org/openhab/binding/icloud/internal/utilities/CustomCookieStore.java
```diff
@@ -0,0 +1,111 @@
...
+        this.cookieStore = new CookieManager().getCookieStore();
+    }
+
+    @Override
+    public void add(URI uri, HttpCookie cookie) {
+
```
jlaur, line 57: Unneeded empty line.

bundles/org.openhab.binding.icloud/src/main/java/org/openhab/binding/icloud/internal/utilities/CustomCookieStore.java
```diff
@@ -0,0 +1,111 @@
...
+        this.cookieStore.add(uri, cookie);
+    }
+
+    @Override
+    public List<HttpCookie> get(URI uri) {
+
```
jlaur, line 63: Unneeded empty line.

bundles/org.openhab.binding.icloud/src/main/java/org/openhab/binding/icloud/internal/utilities/CustomCookieStore.java
```diff
@@ -0,0 +1,111 @@
...
+        List<HttpCookie> result = this.cookieStore.get(uri);
+        filterCookies(result);
+        return result;
+    }
+
+    @Override
+    public List<HttpCookie> getCookies() {
+
```
jlaur, line 71: Unneeded empty line.

bundles/org.openhab.binding.icloud/src/main/java/org/openhab/binding/icloud/internal/utilities/CustomCookieStore.java
```diff
@@ -0,0 +1,111 @@
...
+        List<HttpCookie> result = this.cookieStore.getCookies();
+        filterCookies(result);
+        return result;
+    }
+
+    @Override
+    public List<URI> getURIs() {
+
```
jlaur, line 79: Unneeded empty line.

bundles/org.openhab.binding.icloud/src/main/java/org/openhab/binding/icloud/internal/utilities/CustomCookieStore.java
```diff
@@ -0,0 +1,111 @@
...
+        return this.cookieStore.getURIs();
+    }
+
+    @Override
+    public boolean remove(URI uri, HttpCookie cookie) {
+
```
jlaur, line 85: Unneeded empty line.

bundles/org.openhab.binding.icloud/src/main/java/org/openhab/binding/icloud/internal/utilities/CustomCookieStore.java
```diff
@@ -0,0 +1,111 @@
...
+        return this.cookieStore.remove(uri, cookie);
+    }
+
+    @Override
+    public boolean removeAll() {
+
```
jlaur, line 91: Unneeded empty line.

bundles/org.openhab.binding.icloud/src/main/java/org/openhab/binding/icloud/internal/utilities/CustomCookieStore.java
```diff
@@ -0,0 +1,111 @@
...
+        return this.cookieStore.removeAll();
+    }
+
+    /**
+     * Add quotes add beginning and end of all cookie values
+     *
+     * @param cookieList Current cookies. This list is modified in-place.
+     */
+    private void filterCookies(List<HttpCookie> cookieList) {
+
```
jlaur, line 101: Unneeded empty line.

bundles/org.openhab.binding.icloud/src/main/java/org/openhab/binding/icloud/internal/utilities/CustomCookieStore.java
```diff
@@ -0,0 +1,111 @@
...
+/**
+ *
+ * This class implements a customized {@link CookieStore}. Its purpose is to add hyphens at the beginning and end of
+ * each cookie value which is required by Apple iCloud API.
+ *
+ *
+ * @author Simon Spielmann Initial contribution
+ */
+public class CustomCookieStore implements CookieStore {
```
jlaur, line 36: Please annotate with `@NonNullByDefault`.

bundles/org.openhab.binding.icloud/src/main/java/org/openhab/binding/icloud/internal/utilities/ListUtil.java
```diff
@@ -0,0 +1,55 @@
...
+import java.util.Iterator;
+import java.util.List;
+
+/**
+ * This class implements util methods for list handling.
+ *
+ * @author Simon Spielmann - Initial contribution
+ *
+ */
+public abstract class ListUtil {
```
jlaur, line 25: Please annotate with `@NonNullByDefault`.

bundles/org.openhab.binding.icloud/src/main/java/org/openhab/binding/icloud/internal/utilities/Pair.java
```diff
@@ -0,0 +1,63 @@
...
+
+/**
+ *
+ * Implementation of simple pair. Used mainly for HTTP header handling.
+ *
+ * @author Simon Spielmann Initial contribution.
+ * @param <K> Type of first element
+ * @param <V> Type of second element
+ */
+public class Pair<K, V> {
```
jlaur, line 23: Please annotate with `@NonNullByDefault`.

bundles/org.openhab.binding.icloud/src/main/java/org/openhab/binding/icloud/internal/utilities/Pair.java
```diff
@@ -0,0 +1,63 @@
...
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package org.openhab.binding.icloud.internal.utilities;
+
+/**
+ *
+ * Implementation of simple pair. Used mainly for HTTP header handling.
+ *
+ * @author Simon Spielmann Initial contribution.
```
jlaur, line 19:
```suggestion
 * @author Simon Spielmann - Initial contribution.
```

bundles/org.openhab.binding.icloud/src/main/java/org/openhab/binding/icloud/internal/utilities/Pair.java
```diff
@@ -0,0 +1,63 @@
...
+
+    /**
+     * Create pair with key and value. Both of type {@link String}.
+     *
+     * @param key Key
+     * @param value Value
+     * @return Pair with given key and value
+     */
+    public static Pair<String, String> of(String key, String value) {
+
```
jlaur, line 37: Unneeded empty line.

bundles/org.openhab.binding.icloud/src/main/java/org/openhab/binding/icloud/internal/utilities/Pair.java
```diff
@@ -0,0 +1,63 @@
...
+        Pair<String, String> p = new Pair<>();
+        p.key = key;
+        p.value = value;
+        return p;
+    }
+
+    @Override
+    public String toString() {
+
```
jlaur, line 46: Unneeded empty line.

bundles/org.openhab.binding.icloud/src/main/resources/OH-INF/i18n/iCloud_de.properties
```diff
@@ -10,6 +10,8 @@ icloud.account-thing.parameter.apple-id.label=Apple-ID
 icloud.account-thing.parameter.apple-id.description=Apple-ID (E-Mail Adresse), um Zugriff zur iCloud zu erhalten.
 icloud.account-thing.parameter.password.label=Passwort
 icloud.account-thing.parameter.password.description=Passwort der Apple-ID, um Zugriff zur iCloud zu erhalten.
+icloud.account-thing.parameter.code.label=Code
+icloud.account-thing.parameter.code.description=Code für die 2-Faktor-Authentifizierung des Accounts.
```
jlaur, line 14: Please remove these changes, translations should be updated through Crowdin. For reference, see https://www.openhab.org/docs/developer/utils/i18n.html#managing-translations

bundles/org.openhab.binding.icloud/src/test/java/org/openhab/binding/icloud/TestICloud.java
```diff
@@ -0,0 +1,71 @@
...
+import org.openhab.binding.icloud.internal.json.response.ICloudAccountDataResponse;
+import org.openhab.core.storage.json.internal.JsonStorage;
+
+/**
+ *
+ * Class to test/experiment with iCloud api.
+ *
+ * @author Simon Spielmann
+ */
+public class TestICloud {
```
jlaur, line 35: Please annotate with `@NonNullByDefault`.

bundles/org.openhab.binding.icloud/src/test/java/org/openhab/binding/icloud/TestICloud.java
```diff
@@ -0,0 +1,71 @@
...
+    public void testAuth() throws IOException, InterruptedException {
+
+        File jsonStorageFile = new File(System.getProperty("user.home"), "openhab.json");
+        System.out.println(jsonStorageFile.toString());
+
+        JsonStorage<String> stateStorage = new JsonStorage<String>(jsonStorageFile, TestICloud.class.getClassLoader(),
+                0, 1000, 1000, List.of());
+
+        ICloudService service = new ICloudService(this.E_MAIL, this.PW, stateStorage);
+        // ICloudService service = new ICloudService(this.E_MAIL, this.PW, new VolatileStorage<>());
```
jlaur, line 52: Can this line be removed?

bundles/org.openhab.binding.icloud/src/test/resources/simplelogger.properties
```diff
@@ -0,0 +1,34 @@
...
+# Defaults to true.
+#org.slf4j.simpleLogger.showThreadName=true
+
+# Set to true if you want the Logger instance name to be included in output messages.
+# Defaults to true.
+#org.slf4j.simpleLogger.showLogName=true
+
+# Set to true if you want the last component of the name to be included in output messages.
+# Defaults to false.
+#org.slf4j.simpleLogger.showShortLogName=false
```
jlaur, line 34: Was this file added by accident?

jlaur (changes requested):
Many thanks for taking care of this problem!

I have added some feedback. There are lots of very similar comments, so don't be scared of the amount. Most are related to null annotation and some empty lines and addition of redundant `this.` references for no apparent reason. Perhaps your editor/IDE made some of those changes on your behalf?

bundles/org.openhab.binding.icloud/src/main/java/org/openhab/binding/icloud/internal/ICloudApiResponseException.java
```diff
@@ -0,0 +1,43 @@
...
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package org.openhab.binding.icloud.internal;
+
+/**
+ *
+ * Exception for errors during calls of the iCloud API.
```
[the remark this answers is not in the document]
maihacke: Normally I avoid using checked exception. But this is a endless discussion in java space. If it is more common to use checked exception in openHAB for this types for errors (e.g. expected exceptions?!), I would change it. If it is more my personal opinion I would like to leave it as it is.

bundles/org.openhab.binding.icloud/src/test/resources/simplelogger.properties
maihacke: I wanted to have configurable log-output during local test execution. If it does not harm I like to keep it there.

maihacke:
Thanks for your detailed review. I did most of the request changes. Some I left for discussion, so please check.

maihacke:
Regarding the redundant .this reference, I personally like it very much for clarity. Is there a general rule to not do that in openHAB? My impression from the echo binding was that it is quite common also in the openHab code base to use it.

bundles/org.openhab.binding.icloud/src/main/java/org/openhab/binding/icloud/internal/ICloudApiResponseException.java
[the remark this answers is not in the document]
jlaur: Well, a `RuntimeException` is usually considered a programming error and is fatal. In this case we should expect API errors from time to time and must always handle these to avoid that the binding will stop working. So for this specific case, I would definitely recommend a checked exception.

I now double-checked our review checklist, and actually this is mentioned as a requirement: "Don't throw `RuntimeException` on expected errors" - for reference: https://github.com/openhab/openhab-addons/wiki/Review-Checklist.

jlaur:
> Thanks for your detailed review. I did most of the request changes. Some I left for discussion, so please check.

Is it possible you missed the comments hidden by GitHub? I see only a few fixes in the latest three commits, and the comments still being unresolved without comments?

jlaur:
> Regarding the redundant .this reference, I personally like it very much for clarity. Is there a general rule to not do that in openHAB? My impression from the echo binding was that it is quite common also in the openHab code base to use it.

Generally in most openHAB code I've seen it's avoided when redundant, but I don't think there is a rule currently. Personally I therefore prefer to avoid it for consistency across the codebase.

If a rule would be introduced I would probably also prefer explicit `this.` and explicit class name for static methods so there would be no doubt when reading the code without an IDE at hand. From my experience this can be discussed a lot, so for now I just prefer consistency. :-)

bundles/org.openhab.binding.icloud/src/test/resources/simplelogger.properties
jlaur: In that case I would say this file doesn't belong in the repository, so I would kindly ask for removal. I don't know exactly the purpose, but you might have a look at #12824 if you want to control the log output from tests. It's even possible to check that specific messages are logged.

bundles/org.openhab.binding.icloud/src/main/java/org/openhab/binding/icloud/internal/ICloudApiResponseException.java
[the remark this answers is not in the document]
maihacke: I'll changed it to checked exception.

bundles/org.openhab.binding.icloud/src/main/java/org/openhab/binding/icloud/internal/ICloudService.java
```diff
@@ -0,0 +1,296 @@
...
+     *
+     * @param forceRefresh Force a new authentication
+     * @return {@code true} if authentication was successful
+     * @throws IOException if I/O error occurred
+     * @throws InterruptedException if request was interrupted
+     */
+    public boolean authenticate(boolean forceRefresh) throws IOException, InterruptedException {
+
+        boolean loginSuccessful = false;
+        // pyicloud 286
```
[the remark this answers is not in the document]
maihacke: Removed

bundles/org.openhab.binding.icloud/src/main/java/org/openhab/binding/icloud/internal/handler/ICloudDeviceHandler.java
maihacke: Did it get it right, your suggestion is to remove this if clause, because the potential situation is handled by the framework?

lsiepel:
Just to verify: If i understand it right from the community forums, this binding is unusable in it currrent state without this PR?

bundles/org.openhab.binding.icloud/src/main/java/org/openhab/binding/icloud/internal/handler/ICloudDeviceHandler.java
maihacke: Did change this according to my understanding, maybe double check if this was your intended change.

maihacke:
> Just to verify: If i understand it right from the community forums, this binding is unusable in it currrent state without this PR?

Yes, correct.

maihacke:
> > Thanks for your detailed review. I did most of the request changes. Some I left for discussion, so please check.
> 
> Is it possible you missed the comments hidden by GitHub? I see only a few fixes in the latest three commits, and the comments still being unresolved without comments?

Yes, I overlooked it, since github collapse a lot of them. Sorry for that.

bundles/org.openhab.binding.icloud/src/main/java/org/openhab/binding/icloud/internal/utilities/CustomCookieStore.java
maihacke: This produces a lot of error because of unmatched null-typed from supertype. Tried a few variants, but couldn't get it compile. Do you have a hint or is it okay to have this derived class without this annotation?

maihacke:
@jlaur 
I added the requests @NonNullByDefault annotations, in many cases I do not understand why the compiler still complains.
E.g.

```
    public boolean authenticateWithToken() throws IOException, InterruptedException {

        Map<String, Object> requestBody = new HashMap<>();
        if (session.getAccountCountry() != null) {
            requestBody.put("accountCountryCode", session.getAccountCountry());
        }
```

Produces
`[ERROR] /Users/simon/Projects/devon-ide/workspaces/openhab/openhab-addons/bundles/org.openhab.binding.icloud/src/main/java/org/openhab/binding/icloud/internal/ICloudService.java:[141,51] Null type mismatch (type annotations): required 'java.lang.@NonNull Object' but this expression has type 'java.lang.@Nullable String'`

lsiepel:
> @jlaur I added the requests @NonNullByDefault annotations, in many cases I do not understand why the compiler still complains. E.g.
> 
> ```
>     public boolean authenticateWithToken() throws IOException, InterruptedException {
> 
>         Map<String, Object> requestBody = new HashMap<>();
>         if (session.getAccountCountry() != null) {
>             requestBody.put("accountCountryCode", session.getAccountCountry());
>         }
> ```
> 
> Produces `[ERROR] /Users/simon/Projects/devon-ide/workspaces/openhab/openhab-addons/bundles/org.openhab.binding.icloud/src/main/java/org/openhab/binding/icloud/internal/ICloudService.java:[141,51] Null type mismatch (type annotations): required 'java.lang.@NonNull Object' but this expression has type 'java.lang.@Nullable String'`

Because ‘ session.getAccountCountry()’ may change to null between the two calls as it is accessed out of scope.
If you create a local var, check it and use it the problem is gone.

String localCountry = session.getAccountCountry();
If (localCountry != null) {
use localCountry and not session. getAccountCountry()
}

maihacke:
> > @jlaur I added the requests @NonNullByDefault annotations, in many cases I do not understand why the compiler still complains. E.g.
> > ```
> >     public boolean authenticateWithToken() throws IOException, InterruptedException {
> > 
> >         Map<String, Object> requestBody = new HashMap<>();
> >         if (session.getAccountCountry() != null) {
> >             requestBody.put("accountCountryCode", session.getAccountCountry());
> >         }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Produces `[ERROR] /Users/simon/Projects/devon-ide/workspaces/openhab/openhab-addons/bundles/org.openhab.binding.icloud/src/main/java/org/openhab/binding/icloud/internal/ICloudService.java:[141,51] Null type mismatch (type annotations): required 'java.lang.@NonNull Object' but this expression has type 'java.lang.@Nullable String'`
> 
> Because `session.getAccountCountry()` may change to null between the two calls as it is accessed out of scope. If you create a local var, check it and use it the problem is gone.
> 
> ```
> String localAccountCountry = session.getAccountCountry();
> if (localAccountCountry != null) {
>     requestBody.put("accountCountryCode", localAccountCountry);
> }
> ```

Ok, that helped. Thanks. Here is another one:

```
        String storedData = stateStorage.get(SESSION_DATA_KEY);
        if (storedData != null) {
            data = gson.fromJson(storedData, ICloudSessionData.class);
        }
```

```
[ERROR] /Users/simon/Projects/devon-ide/workspaces/openhab/openhab-addons/bundles/org.openhab.binding.icloud/src/main/java/org/openhab/binding/icloud/internal/ICloudSession.java:[77,20] Null type mismatch (type annotations): required 'org.openhab.binding.icloud.internal.ICloudSession.@NonNull ICloudSessionData' but this expression has type 'org.openhab.binding.icloud.internal.ICloudSession.@Nullable ICloudSessionData'
```

lsiepel:
> ICloudSessionData

If you read the warning carefully, you might notice `ICloudSessionData` is nullable. and you access a property if it. So thi sis in essence the same as the previos warning. Create local var, check it, use it.

maihacke:
Sorry, what to do in such a case with the null check? I could in theory store getTrustToken in a local var, but thats not what I want, because than I have to duplicate the logic from hasToken everywhere...:

        if (session.hasToken()) {
            requestBody.put("trustToken", session.getTrustToken());
        } else {
            requestBody.put("trustToken", "");
        }

lsiepel:
> Sorry, what to do in such a case with the null check? I could in theory store getTrustToken in a local var, but thats not what I want, because than I have to duplicate the logic from hasToken everywhere...:
> 
> ```
>     if (session.hasToken()) {
>         requestBody.put("trustToken", session.getTrustToken());
>     } else {
>         requestBody.put("trustToken", "");
>     }
> ```

A matter of design. Don't know about session, but can you make getTrustToken NonNullable, maybe return empty string if there is a problem? If so you might adapt the hasToken() to check for empty string.  There are so many solutions, it's just a matter of design on how to fix this.

maihacke:
> > Sorry, what to do in such a case with the null check? I could in theory store getTrustToken in a local var, but thats not what I want, because than I have to duplicate the logic from hasToken everywhere...:
> > ```
> >     if (session.hasToken()) {
> >         requestBody.put("trustToken", session.getTrustToken());
> >     } else {
> >         requestBody.put("trustToken", "");
> >     }
> > ```
> 
> A matter of design. Don't know about session, but can you make getTrustToken NonNullable, maybe return empty string if there is a problem? If so you might adapt the hasToken() to check for empty string. There are so many solutions, it's just a matter of design on how to fix this.

Yeah sure, I liked my design ;-) But that is not compatible with the null checks... I will rework it.

maihacke:
> > ICloudSessionData
> 
> If you read the warning carefully, you might notice `ICloudSessionData` is nullable. and you access a property if it. So this is in essence the same as the previous warning. Create local var, check it, use it.

The problem is, that it is not ICloudSessionData is not nullable. According the interface contract of gson it won't return null.

lsiepel:
I'm working on some fixes to get you going. Temporary set this to WIP. Try to get some more SAT issues fixed.

maihacke:
@lsiepel Thanks for your changes. Btw is it possible to have the exact same null checks in eclipse. Currently it is total confusing because eclipse warns me for this code:

```
            ICloudSessionData localSessionData = gson.fromJson(storedData, ICloudSessionData.class);
            if (localSessionData != null) {
                data = localSessionData;
            }
```

"localSessionData cannot be null at this point"...
Normally I'm used to try fix those warning inside the IDE and do not have to rely on maven to do checks with an contradicting outcome...

bundles/org.openhab.binding.icloud/src/main/java/org/openhab/binding/icloud/internal/handler/ICloudDeviceHandler.java
```diff
@@ -91,35 +96,33 @@ public void deviceInformationUpdate(List<ICloudDeviceInformation> deviceInformat
...
         if (bridge != null) {
-            bridge.refreshData();
+            ((ICloudAccountBridgeHandler) bridge.getHandler()).registerListener(this);
+        }
+
+        if (bridge.getStatus() == ThingStatus.ONLINE) {
+            ((ICloudAccountBridgeHandler) bridge.getHandler()).refreshData();
+            updateStatus(ThingStatus.ONLINE);
+        } else {
+            updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.BRIDGE_UNINITIALIZED);
```
jlaur, line 114: Something here is still not correct. If bridge is null, you can exit immediately:

```java
Bridge bridge = getBridge();
if (bridge == null) {
    updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.CONFIGURATION_ERROR, "Bridge is not configured");
    return;
}
```

After this point, you can set status to **UNKNOWN** if you cannot quickly determine if online or not.

If bridge is offline, **BRIDGE_OFFLINE** is the correct detail.

lsiepel: > Something here is still not correct. If bridge is null, you can exit immediately:
> 
> ```java
> Bridge bridge = getBridge();
> if (bridge == null) {
>     updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.CONFIGURATION_ERROR, "Bridge is not configured");
>     return;
> }
> ```
> 
> After this point, you can set status to **UNKNOWN** if you cannot quickly determine if online or not.
> 
> If bridge is offline, **BRIDGE_OFFLINE** is the correct detail.

Think this improved with the commit i just did. I am calling it a day for now. I might proceed tomorrow. Depends on @maihacke what he wants, as i commit to his repository/PR. 

There are several other warnings to be fixed and i just drop in, so not sure

lsiepel:
> @lsiepel Thanks for your changes. Btw is it possible to have the exact same null checks in eclipse. Currently it is total confusing because eclipse warns me for this code:
> 
> ```
>             ICloudSessionData localSessionData = gson.fromJson(storedData, ICloudSessionData.class);
>             if (localSessionData != null) {
>                 data = localSessionData;
>             }
> ```
> 
> "localSessionData cannot be null at this point"... Normally I'm used to try fix those warning inside the IDE and do not have to rely on maven to do checks with an contradicting outcome...

Let me know if this puts you on the right track and/or if you need more help. I'm no pro, but maybe together we can get this done. I'll check in tomorrow.

bundles/org.openhab.binding.icloud/src/main/java/org/openhab/binding/icloud/internal/handler/ICloudDeviceHandler.java
jlaur: Btw, see also #12330.

maihacke:
@jlaur / @lsiepel Thank a lot for the contributions a lot of the warnings are no gone. Would have been better if I would have been aware of the @NonNullByDefault with my initial contribution, so it caused a lot of additionial work.
I lost a bit of overview what still has to be done. There are a few warnings left for missing @NonNullByDefault, for classes which already existed before my contribution here. If necessary I can fix these too.

bundles/org.openhab.binding.icloud/src/main/java/org/openhab/binding/icloud/internal/utilities/CustomCookieStore.java
jlaur: All new classes should be annotated. I'll see if I can find time to have a look in the evening, if not @lsiepel will not have checked already. :)

bundles/org.openhab.binding.icloud/src/main/java/org/openhab/binding/icloud/internal/utilities/Pair.java
```diff
@@ -0,0 +1,67 @@
...
+    /**
+     * Create pair with key and value. Both of type {@link String}.
+     *
+     * @param key Key
+     * @param value Value
+     * @return Pair with given key and value
+     */
+    public static Pair<String, String> of(String key, String value) {
+        Pair<String, String> p = new Pair<>(key, value);
+        return p;
```
jlaur, line 46:
```suggestion
        return new Pair<>(key, value);
```

jlaur (changes requested):
A few additional comments from quick run-through. I will have a closer look in the evening.

maihacke:
I took care of some more of the warnings. What' left is CustomCookieStore and a case I could not cleanly handle in ICloudAccountBridgeHandler:

```
    @SuppressWarnings("null")
    public void findMyDevice(String deviceId) throws IOException, InterruptedException {
        callApiWithRetryAndExceptionHandling(() -> {
            // callApiWithRetryAndExceptionHanlding ensures that iCloudService is not null when the following is
            // called. Cannot use method local iCloudService instance here, because instance may be replaced with a new
            // one during retry.
            iCloudService.getDevices().playSound(deviceId);
            return null;
        });
    }
```
I think during the review one of you suggested the use of a local instance of iCloudService here. But this is not possible. The surrounding code in callApiWithRetryAndExceptionHandling may provide a fresh instance of iCloudService. Which is necessary to handle the authentication in some error cases. Concurrency is also handled with a sync block.

lsiepel:
Left two warnings, I could not get the annotations working with the `CustomCookieStore`.
Later this evening i have more time to look at the open review comments and do some basic testing. @maihacke you want me to look at something sepcific?

bundles/org.openhab.binding.icloud/src/main/java/org/openhab/binding/icloud/internal/handler/ICloudAccountBridgeHandler.java
```diff
@@ -89,18 +107,151 @@ public void handleCommand(ChannelUID channelUID, Command command) {
...
+                }
+            }
+
+            if (authState == AuthState.INITIAL) {
+                success = checkLogin();
+            } else if (authState == AuthState.WAIT_FOR_CODE) {
+                try {
+                    success = handle2FAAuthentication();
+                } catch (IOException | InterruptedException | ICloudApiResponseException ex) {
+                    logger.warn("Error while validating 2-FA code.", ex);
```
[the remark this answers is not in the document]
lsiepel: @maihacke Should the logger.warn line be replace with ?
```
updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.CONFIGURATION_ERROR,
                    "Error while validating 2-FA code.");
```

lsiepel:
Ok, so i think i stumbled upon a bug. As no matter how i annotate `CustomCookieStore` it always throws a build error
I set the class as `@NonNullByDefault`
Then the return types (3) with a list fail with the warning that it is not compatible with the `super` type, I tried:
```
public @Nullable List<@NonNull HttpCookie> getCookies()
public @Nullable List<@Nullable HttpCookie> getCookies()
```
Both fail, I'm not aware of another null state :-/ 

The rest of the issues have been fixed as far as i know. One questions to @maihacke  is still open. I'm about to perform some basic tests now. @jlaur could you do another review pass?

bundles/org.openhab.binding.icloud/src/main/java/org/openhab/binding/icloud/internal/RetryException.java
```diff
@@ -0,0 +1,36 @@
...
+
+/**
+ * This exception is thrown when a retry finally fails.
+ *
+ * @author Simon Spielmann - Initial contribution
+ */
+@NonNullByDefault
+public class RetryException extends RuntimeException {
+
+    private static final long serialVersionUID = 1L;
```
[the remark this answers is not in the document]
jlaur: Marked as resolved without comment or code change?

bundles/org.openhab.binding.icloud/src/main/java/org/openhab/binding/icloud/internal/handler/ICloudDeviceHandler.java
jlaur: This is still not resolved, so reopening. I'll try to add some more explanation.

bundles/org.openhab.binding.icloud/src/main/java/org/openhab/binding/icloud/internal/handler/ICloudDeviceHandler.java
```diff
@@ -91,35 +94,43 @@ public void deviceInformationUpdate(List<ICloudDeviceInformation> deviceInformat
...
-        ICloudAccountBridgeHandler handler = getIcloudAccount();
-        if (handler != null) {
-            refreshData();
-        } else {
-            updateStatus(OFFLINE, BRIDGE_UNINITIALIZED, "Bridge not found");
-        }
-    }
+        Bridge bridge = getBridge();
+        Object bridgeStatus = (bridge == null) ? null : bridge.getStatus();
+        this.logger.debug("initializeThing thing [{}]; bridge status: [{}]", getThing().getUID(), bridgeStatus);
```
jlaur, line 102: This can be removed.

bundles/org.openhab.binding.icloud/src/main/java/org/openhab/binding/icloud/internal/handler/ICloudDeviceHandler.java
```diff
@@ -91,35 +94,43 @@ public void deviceInformationUpdate(List<ICloudDeviceInformation> deviceInformat
...
+                if (bridge.getStatus() == ThingStatus.ONLINE) {
+                    handler.refreshData();
+                    updateStatus(ThingStatus.ONLINE);
+                }
+            } else {
+                updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.BRIDGE_UNINITIALIZED,
+                        "Bridge handler is not configured");
+            }
+        } else {
+            updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.BRIDGE_UNINITIALIZED, "Bridge is not configured");
```
jlaur, line 117:
```suggestion
            updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.CONFIGURATION_ERROR, "Bridge is not configured");
```

lsiepel: Fixed

bundles/org.openhab.binding.icloud/src/main/java/org/openhab/binding/icloud/internal/handler/ICloudDeviceHandler.java
lsiepel: Fixed

bundles/org.openhab.binding.icloud/src/main/java/org/openhab/binding/icloud/internal/handler/ICloudDeviceHandler.java
```diff
@@ -91,35 +94,43 @@ public void deviceInformationUpdate(List<ICloudDeviceInformation> deviceInformat
...
-    private void refreshData() {
-        ICloudAccountBridgeHandler bridge = getIcloudAccount();
         if (bridge != null) {
-            bridge.refreshData();
+            ICloudAccountBridgeHandler handler = (ICloudAccountBridgeHandler) bridge.getHandler();
+            if (handler != null) {
+                handler.registerListener(this);
+                if (bridge.getStatus() == ThingStatus.ONLINE) {
+                    handler.refreshData();
+                    updateStatus(ThingStatus.ONLINE);
```
jlaur, line 107: Perhaps it would be worth considering to simply set status **UNKNOWN** here after registering as listener, and let the bridge handler override [childHandlerInitialized](https://www.openhab.org/javadoc/latest/org/openhab/core/thing/binding/bridgehandler#childHandlerInitialized(org.openhab.core.thing.binding.ThingHandler,org.openhab.core.thing.Thing)). Then bridge would call `deviceInformationUpdate` normally as soon as child handlers are ready, without this need to request the bridge for callback. It would have the benefit of only setting the things online when in fact data was received, as opposed to when data is requested (like now). WDYT?

bundles/org.openhab.binding.icloud/src/main/java/org/openhab/binding/icloud/internal/handler/ICloudAccountBridgeHandler.java
```diff
@@ -86,21 +99,152 @@ public void handleCommand(ChannelUID channelUID, Command command) {
...
+
+        if (authState == AuthState.AUTHENTICATED) {
+            ICloudAccountThingConfiguration config = getConfigAs(ICloudAccountThingConfiguration.class);
+            this.refreshJob = this.scheduler.scheduleWithFixedDelay(this::refreshData, 0, config.refreshTimeInMinutes,
+                    MINUTES);
+        } else {
+            cancelRefresh();
+        }
+        logger.debug("iCloud bridge handler initialized.");
+    }
```
jlaur, line 135: Before leaving this method, bridge status should be set. Are we sure it is?

bundles/org.openhab.binding.icloud/src/main/java/org/openhab/binding/icloud/internal/utilities/CustomCookieStore.java
lsiepel: Sorry, I'm unable to get the annotation to work for this class. Can you help out? Or merge as is?

jlaur: Unfortunately I was not able to sort it out tonight. Let's focus on the other comments first, and this one we can let slip if we won't manage in time.

bundles/org.openhab.binding.icloud/src/main/java/org/openhab/binding/icloud/internal/RetryException.java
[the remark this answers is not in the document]
lsiepel: Not sure if it was me? Didn't see this comment at all, so if i did it was by accident. 
Anyway, i tried about an hour to get vscode to generate one. `serialver` always reports class not found. So maybe someone else can generate this ? I see many bindings with a 1L value, so doubt if this is an actual problem.

jlaur, line 26:
```suggestion
    private static final long serialVersionUID = 4367157567009973413L;
```

bundles/org.openhab.binding.icloud/src/main/java/org/openhab/binding/icloud/internal/handler/ICloudDeviceHandler.java
lsiepel: That would be an improvement. As i only stepped in to help out @maihacke  with the annotations/warnings, i don;t think it is up to me. If so i would need more time to understand the workings, read documentation and test and check. As the binding works and i see other bindings do similar thing i would suggest to merge it as is. During the christmass holidays i'm less time cosntraint and might comcome up with a new PR tryging to fix the remaining issues, or maybe @maihacke has other plans?

bundles/org.openhab.binding.icloud/src/main/java/org/openhab/binding/icloud/internal/handler/ICloudAccountBridgeHandler.java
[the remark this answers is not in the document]
lsiepel: Updated.

bundles/org.openhab.binding.icloud/src/main/java/org/openhab/binding/icloud/internal/handler/ICloudAccountBridgeHandler.java
lsiepel: I'm allmost certain that it is. I tested again and followed all code paths and did not see any path not leading to a status update

bundles/org.openhab.binding.icloud/src/main/java/org/openhab/binding/icloud/internal/RetryException.java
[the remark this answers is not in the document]
lsiepel: Thanks, fixed.

bundles/org.openhab.binding.icloud/src/main/java/org/openhab/binding/icloud/internal/RetryException.java
```diff
@@ -23,7 +23,7 @@
 @NonNullByDefault
 public class RetryException extends RuntimeException {
 
-    private static final long serialVersionUID = 1L;
+    private static final long serialVersionUID = 4367157567009973413L;
```
maihacke, line 26: I prefer 1L here. Putting a serialVersionUID is best practice to allow "manual" management of the UID if the class is serialized and we want to maintain backward compatibility. From my point of view this is more explicit and easier to handle if it is a simple value (like 1) and not the one which would have been generated by the javac.

bundles/org.openhab.binding.icloud/src/main/java/org/openhab/binding/icloud/internal/ICloudSession.java
```diff
@@ -137,7 +137,8 @@ private String request(String method, String url, @Nullable String body,
 
         HttpResponse<?> response = this.client.send(request, BodyHandlers.ofString());
 
-        String responseBodyAsString = response.body() != null ? response.body().toString() : "";
+        Object responseBody = response.body();
+        String responseBodyAsString = responseBody != null ? responseBody.toString() : "";
```
maihacke, line 141: Why is this change needed? Response is a local value, so no other thread could have modified it in between...

bundles/org.openhab.binding.icloud/src/test/java/org/openhab/binding/icloud/TestICloud.java
```diff
@@ -73,7 +74,10 @@ public void testAuth() throws IOException, InterruptedException, ICloudApiRespon
         ICloudService service = new ICloudService(iCloudTestEmail, iCloudTestPassword, stateStorage);
         service.authenticate(false);
         if (service.requires2fa()) {
-            System.out.print("Code: ");
+            PrintStream consoleOutput = System.out;
```
maihacke, line 77: Is it really technically possible that stdout is NULL?
What about System.in (next lines)?

lsiepel: Don’t think it will ever be null. But it is nullable and the compiler warned about it so I added this check.

bundles/org.openhab.binding.icloud/src/test/java/org/openhab/binding/icloud/TestICloud.java
```diff
@@ -74,7 +74,9 @@ public void testAuth() throws IOException, InterruptedException, ICloudApiRespon
         service.authenticate(false);
         if (service.requires2fa()) {
             System.out.print("Code: ");
-            String code = new Scanner(System.in).nextLine();
+            Scanner in = new Scanner(System.in);
+            String code = in.nextLine();
+            in.close();
```
maihacke, line 83: System.in should not be closed. It is managed by the JVM.

bundles/org.openhab.binding.icloud/src/main/java/org/openhab/binding/icloud/internal/ICloudSession.java
lsiepel: Response.body() is nullable.

bundles/org.openhab.binding.icloud/src/test/java/org/openhab/binding/icloud/TestICloud.java
maihacke: OK -1 for readabillity ;-), that the problem with those static code analyzer, they are often a bit silly. But I can live with it.

bundles/org.openhab.binding.icloud/src/test/java/org/openhab/binding/icloud/TestICloud.java
lsiepel: compiler warned about it, did some reading and yeah compilers are sometimes silly. Fixed by surpressing the warning.

bundles/org.openhab.binding.icloud/src/main/java/org/openhab/binding/icloud/internal/RetryException.java
jlaur: Since this is an exception which is unlikely to be persisted anywhere between versions, I agree the value is not important. I'm sorry for the confusion.

lsiepel: @maihacke @jlaur both is fine by me. At the moment it is 1L due to this change being reverted.

lsiepel:
I'll be available if anything is needed, but i prefer to step out of this PR as it is not my PR and the annotations that where blocking are fixed. Offcourse if needed i'm available, but i think the rest is best to be solved by @maihacke 
Thanks for all the efforts!

bundles/org.openhab.binding.icloud/src/main/java/org/openhab/binding/icloud/internal/handler/ICloudDeviceHandler.java
maihacke: I created a WIP issue to collect those future todos: https://github.com/openhab/openhab-addons/issues/13942

maihacke:
I also did some local tests with the latests code base. Everything still seems to work. I created a new RC-4 to have more people test it https://github.com/maihacke/openhab-addons/releases/tag/rc-4.

jlaur:
> Ok, thats fine for me. Thanks for your help. I could resolve remaining issues. So I'm not sure, what is left - @jlaur?

I will double-check comments and latest commits, I think we are quite ready. One last thing I would ask: Can you double-check that bridge handler doesn't perform any blocking network calls during `initialize()`, and that status is set before leaving the method? In both cases, this would be a bug, and it could be easily fixed:
- In case of network calls, this should happen in separate thread, e.g. `scheduler->execute()`
- In case of missing status, `updateStatus(ThingStatus.UNKNOWN);` before starting anything async which could update the status.

For reference:
- https://www.openhab.org/javadoc/latest/org/openhab/core/thing/binding/thinghandler#initialize()
- https://www.openhab.org/docs/concepts/things.html#status-transitions

bundles/org.openhab.binding.icloud/src/main/java/org/openhab/binding/icloud/internal/utilities/ICloudTextTranslator.java
```diff
@@ -37,11 +39,13 @@ public ICloudTextTranslator(Bundle bundle, TranslationProvider i18nProvider, Loc
...
-        return i18nProvider != null ? i18nProvider.getText(bundle, key, getDefaultText(key), locale, arguments) : key;
+        Locale locale = localeProvider.getLocale();
+        String retText = i18nProvider.getText(bundle, key, getDefaultText(key), locale, arguments);
+        return retText != null ? retText : key;
     }
 
     public String getDefaultText(@Nullable String key) {
-        return i18nProvider.getText(bundle, key, key, Locale.ENGLISH);
+        String retText = i18nProvider.getText(bundle, key, key, Locale.ENGLISH);
+        return retText != null ? retText : key != null ? key : "UNKNOWN_TEXT";
```
jlaur, line 49: Any reason for key to be nullable? I believe this logic could be simplified, since it doesn't make sense to call this method without a valid argument.

bundles/org.openhab.binding.icloud/src/main/java/org/openhab/binding/icloud/internal/handler/ICloudAccountBridgeHandler.java
```diff
@@ -86,21 +99,154 @@ public void handleCommand(ChannelUID channelUID, Command command) {
...
+                        checkLogin();
+                    }
+                } catch (ICloudApiResponseException e) {
+                    logger.debug("ICloudApiResponseException with status code {}", e.getStatusCode());
+                    if (e.getStatusCode() == 450) {
+                        checkLogin();
+                    }
+                } catch (IllegalStateException e) {
+                    logger.debug("Need to authenticate first.", e);
+                    updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.CONFIGURATION_PENDING, "Wait for login");
```
jlaur, line 186: This is not a valid combination of status and details, and probably **CONFIGURATION_PENDING** is not correct according to definition - see https://www.openhab.org/docs/concepts/things.html#status-details. Perhaps **NONE**?

maihacke: Ok, configuration_pending would have fit from semantic point of view, but if it's not allowed NONE is the next best fit. I'll change it.

bundles/org.openhab.binding.icloud/src/main/java/org/openhab/binding/icloud/internal/utilities/ICloudTextTranslator.java
maihacke: The initial code wasn't written by be and the change is from @lsiepel . I guess is nullable, because it is also in the called TranslationProvider:

```
@NonNullByDefault
public interface TranslationProvider {

    /**
     * Returns a translation for the specified key in the specified locale (language) by only
     * considering the translations within the specified module.
     * <p>
     * If no translation could be found, the specified default text is returned.<br>
     * If the specified locale is {@code null}, the default locale is used.
     *
     * @param bundle the module to be used for the look-up (could be null)
     * @param key the key to be translated (could be null or empty)
     * @param defaultText the default text to be used (could be null or empty)
     * @param locale the locale (language) to be used (could be null)
     * @return the translated text or the default text (could be null or empty)
     */
    @Nullable
    String getText(@Nullable Bundle bundle, @Nullable String key, @Nullable String defaultText,
            @Nullable Locale locale);
```

maihacke:
> > Ok, thats fine for me. Thanks for your help. I could resolve remaining issues. So I'm not sure, what is left - @jlaur?
> 
> I will double-check comments and latest commits, I think we are quite ready. One last thing I would ask: Can you double-check that bridge handler doesn't perform any blocking network calls during `initialize()`, and that status is set before leaving the method? In both cases, this would be a bug, and it could be easily fixed:
> 
> * In case of network calls, this should happen in separate thread, e.g. `scheduler->execute()`
> * In case of missing status, `updateStatus(ThingStatus.UNKNOWN);` before starting anything async which could update the status.
> 
> For reference:
> 
> * https://www.openhab.org/javadoc/latest/org/openhab/core/thing/binding/thinghandler#initialize()
> * https://www.openhab.org/docs/concepts/things.html#status-transitions

Currently there is a network call. I'll try to change it, hope it doesn't have any side-effects.

jlaur:
> * `updateStatus(ThingStatus.UNKNOWN);` before starting anything async which could update the status.

I believe this part is still missing. Should be called before this line:
```java
this.scheduler.schedule(initialAuthentication, 0, TimeUnit.SECONDS);
```

bundles/org.openhab.binding.icloud/src/main/java/org/openhab/binding/icloud/internal/handler/ICloudAccountBridgeHandler.java
```diff
@@ -86,21 +100,155 @@ public void handleCommand(ChannelUID channelUID, Command command) {
...
+
+        if (authState != AuthState.WAIT_FOR_CODE) {
+            authState = AuthState.INITIAL;
+        }
+
+        Callable<?> initialAuthentication = () -> callApiWithRetryAndExceptionHandling(() -> {
+            logger.debug("Dummy call for initial authentication.");
+            return null;
+        });
+        this.scheduler.schedule(initialAuthentication, 0, TimeUnit.SECONDS);
```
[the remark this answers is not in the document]
jlaur: If I understand it correctly, you need to sequentially ensure that initial authentication is done, then schedule refresh job. Perhaps it would be better to call initialAuthentication **from** the refresh job when not yet authenticated? This would also ensure that it's retried if the network call will fail during an authentication attempt. I didn't check if there is any other retry mechanism in place, but if not, this would improve reliability and at the same time solve the problem of multiple async jobs.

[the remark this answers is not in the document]
jlaur: Additionally, with current approach the future returned by `schedule()` should be stored, so it can be cancelled by `dispose()`.

maihacke:
> > * `updateStatus(ThingStatus.UNKNOWN);` before starting anything async which could update the status.
> 
> I believe this part is still missing. Should be called before this line:
> 
> ```java
> this.scheduler.schedule(initialAuthentication, 0, TimeUnit.SECONDS);
> ```

Ah sorry, hat problems with the solution and tried different approaches. At the end I forgot this line.

bundles/org.openhab.binding.icloud/src/main/java/org/openhab/binding/icloud/internal/handler/ICloudAccountBridgeHandler.java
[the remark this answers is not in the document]
maihacke: The flow here is a bit tricky. The refreshJob is only scheduled, if the authentication was successful:

```
        if (authState == AuthState.AUTHENTICATED) {
```

The problem is that the initial authentication call triggers the request for a 2-FA code. This request must not be automatically retried. Otherwise multiple codes could be requested. Since this is a oneshot job, I do not see see the need to cancel it.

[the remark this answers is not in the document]
maihacke: Yes you are right, this has to be done in the async task.

bundles/org.openhab.binding.icloud/src/main/java/org/openhab/binding/icloud/internal/handler/ICloudDeviceHandler.java
```diff
@@ -105,6 +105,8 @@ public void initialize() {
                 if (bridge.getStatus() == ThingStatus.ONLINE) {
                     handler.refreshData();
                     updateStatus(ThingStatus.ONLINE);
+                } else {
+                    updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.BRIDGE_OFFLINE);
```
jlaur, line 109: Can you check if this is still needed with the latest changes in the bridge handler? Perhaps it should be UNKNOWN here as bridge will probably also be UNKNOWN. If bridge is OFFLINE, I believe child handlers will be automatically OFFLINE with BRIDGE_OFFLINE when not having provided a custom implementation of `bridgeStatusChanged`:

https://github.com/openhab/openhab-core/blob/e420cf16476ed6b08b8fab2626ce6d08beecd995/bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/binding/BaseThingHandler.java#L643-L651

bundles/org.openhab.binding.icloud/src/main/java/org/openhab/binding/icloud/internal/handler/ICloudAccountBridgeHandler.java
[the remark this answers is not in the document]
jlaur: The only question left is what will happen if the authentication will fail because of network issues, for example if internet connectivity is unavailable when the binding is started?

One consequence of the one-shot approach could be that authentication will not be retried at all, so that the binding will stay offline permanently. This requires users to manually stop and start the binding in such cases, so this is a reliability issue.

I haven't checked in detail if this is actually the case, just wanted to mention it. Even if we have this problem, I don't see it as a showstopper, this can be improved further later.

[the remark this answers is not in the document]
maihacke: I think this could happen. I do not like the current approach for 2-FA auth as part of init at all. It males the init complex and error prone. I‘m thinking of implementing a servlet or similar to provide a extra UI for it

bundles/org.openhab.binding.icloud/src/main/java/org/openhab/binding/icloud/internal/handler/ICloudDeviceHandler.java
maihacke: Yes, it is necessary. Otherwise the devices stuck in "INITIALIZING" when the bridge is online during device init. Then they will never init again, even if the bridge goes ONLINE.

jlaur: @maihacke - FYI in case you didn't notice this comment.

maihacke: @jlaur We both commented in the same second ;-)

jlaur: OK, for this last minute bugfix, let's keep it like this then - it can be improved later.

bundles/org.openhab.binding.icloud/src/main/java/org/openhab/binding/icloud/internal/handler/ICloudAccountBridgeHandler.java
```diff
@@ -86,21 +103,160 @@ public void handleCommand(ChannelUID channelUID, Command command) {
...
+                }
+
+                retryCount++;
+                try {
+                    Thread.sleep(200);
+                } catch (InterruptedException e) {
+                    Thread.interrupted();
+                }
+            } while (!success && retryCount < 3);
+            throw new RetryException(lastException);
```
jlaur, line 215: `lastException` is always null?

jlaur (approved):
Thanks for all the effort to fix this binding, and also for the last minutes fixes just before the code freeze.

A few findings from the review were not addressed, so there is some room for improvement in the handling of status transitions and reliability of establishing the initial connection. This is probably status quo when comparing with the previous behavior, so this is something which can be addressed later with more care.

If you need to perform some additional tests, please do so and let me know the results. If not, let's get this merged. 😄

maihacke:
> Thanks for all the effort to fix this binding, and also for the last minutes fixes just before the code freeze.
> 
> A few findings from the review were not addressed, so there is some room for improvement in the handling of status transitions and reliability of establishing the initial connection. This is probably status quo when comparing with the previous behavior, so this is something which can be addressed later with more care.
> 
> If you need to perform some additional tests, please do so and let me know the results. If not, let's get this merged. 😄

Even if there are only minor differences I will create a new RC with the latest commits. So some more people could test it.

maihacke:
@jlaur Having it running without errors for a few days now and the code freeze tomorrow, I suggest somebody merges this. So we might also have time to add the missing i18n strings.

kaikreuzer:
Awesome! Thanks @maihacke and everyone else involved here!

PR #13695, opened 2022-11-11
[jsscripting] Reimplement timer creation method of `ScriptExecution`

florian-h05:
## Description

This PR reimplements the openHAB Core API for creating timers to make this also threadsafe.

As discussed on the forum https://community.openhab.org/t/js-scripting-why-are-we-deprecated-createtimer/140748, the `setTimeout` and `setInterval` methods do not provide the functionality that advanced users need, and to avoid that those users and users coming from Rules DSL leave JS Scripting because of this missing API, we have to provide thread-safe access to it.

Furthermore,  `JSScriptServiceUtil` was introduced to provide easy access from to script services provided by core. This replaces the usage of the `JSScriptServiceUtil` from the Rules DSL parts of core and therefore reduces dependency on a part that might become an addon in the future.

openhab-js is also upgraded to 2.1.1.

## Testing

Use the following file-based script to test:

```javascript
const { time} = require('openhab');

// Note that all timers are scheduled to nearly the same time, so without the synchronization mechanism in place, this would lead to a multi-thread exception
ThreadsafeTimers.createTimer(time.toZDT(), () => {
    console.info("Hello world from createTimer without identifier")
});

ThreadsafeTimers.createTimer('createTimer', time.toZDT(), () => {
    console.info("Hello world from createTimer with identifier")
});
```

### DIFF
--- a/bundles/org.openhab.automation.jsscripting/src/main/java/org/openhab/automation/jsscripting/internal/GraalJSScriptServiceUtil.java
+++ b/bundles/org.openhab.automation.jsscripting/src/main/java/org/openhab/automation/jsscripting/internal/GraalJSScriptServiceUtil.java
@@ -1,0 +1,49 @@
+/**
+ * Copyright (c) 2010-2022 Contributors to the openHAB project
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package org.openhab.automation.jsscripting.internal;
+
+import java.util.Objects;
+
+import org.eclipse.jdt.annotation.NonNullByDefault;
+import org.eclipse.jdt.annotation.Nullable;
+import org.openhab.core.automation.module.script.action.ScriptExecution;
+import org.openhab.core.scheduler.Scheduler;
+import org.osgi.service.component.annotations.Activate;
+import org.osgi.service.component.annotations.Component;
+import org.osgi.service.component.annotations.Reference;
+
+/**
+ * Utility class for providing easy access to script services.
+ *
+ * @author Florian Hotze - Initial contribution
+ */
+@Component(immediate = true)
+@NonNullByDefault
+public class GraalJSScriptServiceUtil {
+    private static @Nullable Scheduler scheduler;
+    private static @Nullable ScriptExecution scriptExecution;
+
+    @Activate
+    public GraalJSScriptServiceUtil(final @Reference Scheduler scheduler, final @Reference ScriptExecution scriptExecution) {
+        GraalJSScriptServiceUtil.scheduler = scheduler;
+        GraalJSScriptServiceUtil.scriptExecution = scriptExecution;
+    }
+
+    public static Scheduler getScheduler() {
+        return Objects.requireNonNull(scheduler);
+    }
+
+    public static ScriptExecution getScriptExecution() {
+        return Objects.requireNonNull(scriptExecution);
+    }
+}
--- a/bundles/org.openhab.automation.jsscripting/src/main/java/org/openhab/automation/jsscripting/internal/GraalJSScriptServiceUtil.java
+++ b/bundles/org.openhab.automation.jsscripting/src/main/java/org/openhab/automation/jsscripting/internal/GraalJSScriptServiceUtil.java
@@ -34,7 +34,8 @@ public class GraalJSScriptServiceUtil {
     private static @Nullable ScriptExecution scriptExecution;
 
     @Activate
-    public GraalJSScriptServiceUtil(final @Reference Scheduler scheduler, final @Reference ScriptExecution scriptExecution) {
+    public GraalJSScriptServiceUtil(final @Reference Scheduler scheduler,
+            final @Reference ScriptExecution scriptExecution) {
         GraalJSScriptServiceUtil.scheduler = scheduler;
         GraalJSScriptServiceUtil.scriptExecution = scriptExecution;
     }
--- a/bundles/org.openhab.automation.jsscripting/src/main/java/org/openhab/automation/jsscripting/internal/JSScriptServiceUtil.java
+++ b/bundles/org.openhab.automation.jsscripting/src/main/java/org/openhab/automation/jsscripting/internal/JSScriptServiceUtil.java
@@ -12,10 +12,7 @@
  */
 package org.openhab.automation.jsscripting.internal;
 
-import java.util.Objects;
-
 import org.eclipse.jdt.annotation.NonNullByDefault;
-import org.eclipse.jdt.annotation.Nullable;
 import org.openhab.core.automation.module.script.action.ScriptExecution;
 import org.openhab.core.scheduler.Scheduler;
 import org.osgi.service.component.annotations.Activate;
@@ -23,27 +20,27 @@ import org.osgi.service.component.annotations.Component;
 import org.osgi.service.component.annotations.Reference;
 
 /**
- * Utility class for providing easy access to script services.
+ * OSGi utility service for providing easy access to script services.
  *
  * @author Florian Hotze - Initial contribution
  */
-@Component(immediate = true)
+@Component(immediate = true, service = JSScriptServiceUtil.class)
 @NonNullByDefault
 public class JSScriptServiceUtil {
-    private static @Nullable Scheduler scheduler;
-    private static @Nullable ScriptExecution scriptExecution;
+    private final Scheduler scheduler;
+    private final ScriptExecution scriptExecution;
 
     @Activate
     public JSScriptServiceUtil(final @Reference Scheduler scheduler, final @Reference ScriptExecution scriptExecution) {
-        JSScriptServiceUtil.scheduler = scheduler;
-        JSScriptServiceUtil.scriptExecution = scriptExecution;
+        this.scheduler = scheduler;
+        this.scriptExecution = scriptExecution;
     }
 
-    public static Scheduler getScheduler() {
-        return Objects.requireNonNull(scheduler);
+    public Scheduler getScheduler() {
+        return scheduler;
     }
 
-    public static ScriptExecution getScriptExecution() {
-        return Objects.requireNonNull(scriptExecution);
+    public ScriptExecution getScriptExecution() {
+        return scriptExecution;
     }
 }
--- a/bundles/org.openhab.automation.jsscripting/src/main/java/org/openhab/automation/jsscripting/internal/JSScriptServiceUtil.java
+++ b/bundles/org.openhab.automation.jsscripting/src/main/java/org/openhab/automation/jsscripting/internal/JSScriptServiceUtil.java
@@ -29,15 +29,15 @@ import org.osgi.service.component.annotations.Reference;
  */
 @Component(immediate = true)
 @NonNullByDefault
-public class GraalJSScriptServiceUtil {
+public class JSScriptServiceUtil {
     private static @Nullable Scheduler scheduler;
     private static @Nullable ScriptExecution scriptExecution;
 
     @Activate
-    public GraalJSScriptServiceUtil(final @Reference Scheduler scheduler,
-            final @Reference ScriptExecution scriptExecution) {
-        GraalJSScriptServiceUtil.scheduler = scheduler;
-        GraalJSScriptServiceUtil.scriptExecution = scriptExecution;
+    public JSScriptServiceUtil(final @Reference Scheduler scheduler,
+                               final @Reference ScriptExecution scriptExecution) {
+        JSScriptServiceUtil.scheduler = scheduler;
+        JSScriptServiceUtil.scriptExecution = scriptExecution;
     }
 
     public static Scheduler getScheduler() {
--- a/bundles/org.openhab.automation.jsscripting/src/main/java/org/openhab/automation/jsscripting/internal/JSScriptServiceUtil.java
+++ b/bundles/org.openhab.automation.jsscripting/src/main/java/org/openhab/automation/jsscripting/internal/JSScriptServiceUtil.java
@@ -34,8 +34,7 @@ public class JSScriptServiceUtil {
     private static @Nullable ScriptExecution scriptExecution;
 
     @Activate
-    public JSScriptServiceUtil(final @Reference Scheduler scheduler,
-                               final @Reference ScriptExecution scriptExecution) {
+    public JSScriptServiceUtil(final @Reference Scheduler scheduler, final @Reference ScriptExecution scriptExecution) {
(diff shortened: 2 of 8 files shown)


florian-h05:
@rkoshak 
PR is now open.

@jlaur @jpg0 Could you please give your feedback on this?

bundles/org.openhab.automation.jsscripting/src/main/java/org/openhab/automation/jsscripting/internal/threading/ThreadsafeTimers.java
```diff
@@ -95,8 +141,11 @@ public long setTimeout(Runnable callback, Long delay) {
      */
     public long setTimeout(Runnable callback, Long delay, Object... args) {
         long id = lastId.incrementAndGet();
-        ScheduledCompletableFuture<Object> future = createFuture(id, ZonedDateTime.now().plusNanos(delay * 1000000),
-                callback);
+        ScheduledCompletableFuture<Object> future = scheduler.schedule(() -> {
+            synchronized (lock) {
+                callback.run();
```
digitaldan, line 146: Should we remove the future from the map here ?  so `idSchedulerMapping.remove(id)` ?  I noticed the only time we remove references to futures is when we completely clear the map at init/reload time?

bundles/org.openhab.automation.jsscripting/src/main/java/org/openhab/automation/jsscripting/internal/threading/ThreadsafeTimers.java
```diff
@@ -55,19 +60,60 @@ public void setIdentifier(String identifier) {
...
+
+    /**
+     * Schedules a block of code (with argument) for later execution
+     *
+     * @param instant the point in time when the code should be executed
+     * @param arg1 the argument to pass to the code block
+     * @param closure the code block to execute
+     * @return a handle to the created timer, so that it can be canceled or rescheduled
+     */
+    public Timer createTimerWithArgument(ZonedDateTime instant, Object arg1, Consumer<Object> closure) {
```
jpg0, line 97: I have two questions about this method. 

1) Presumably your test should actually be:

```
ThreadsafeTimers.createTimerWithArgument(time.toZDT(), arg1, (newArg) => {
    console.info(`Hello world from createTimerWithArgument without identifier and argument: ${newArg}`);
});
```

2) What is the use case here? Why not allow the closure to capture the state? This is effectively what your original test actually did, as the passed parameter was unused. Which is why I'm not sure I understand the use-case.

bundles/org.openhab.automation.jsscripting/src/main/java/org/openhab/automation/jsscripting/internal/threading/ThreadsafeTimers.java
jpg0: It's also cleared on `clearTimeout`, but I agree that it should be cleared here too.

bundles/org.openhab.automation.jsscripting/src/main/java/org/openhab/automation/jsscripting/internal/GraalJSScriptServiceUtil.java
```diff
@@ -0,0 +1,50 @@
...
+import org.osgi.service.component.annotations.Reference;
+
+/**
+ * Utility class for providing easy access to script services.
+ *
+ * @author Florian Hotze - Initial contribution
+ */
+@Component(immediate = true)
+@NonNullByDefault
+public class GraalJSScriptServiceUtil {
```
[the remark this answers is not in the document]
J-N-K: I would suggest to leave out Graal. 3.4 will likely be the last one with integrated Nashorn JS, so my guess would be that someone in a later release will either install Nashorn JS or this one.

bundles/org.openhab.automation.jsscripting/src/main/java/org/openhab/automation/jsscripting/internal/threading/ThreadsafeTimers.java
florian-h05: I agree, I'll change this.

bundles/org.openhab.automation.jsscripting/src/main/java/org/openhab/automation/jsscripting/internal/threading/ThreadsafeTimers.java
florian-h05: > Presumably your test should actually be:

Yes, thanks.

> What is the use case here? Why not allow the closure to capture the state?

TBH, I am not sure what the use case is here. I do not use this in any of my scripts.

jpg0: In that case I would suggest that we remove the `createTimeWithArgument` methods. I _suspect_ that the reason they are here is because:

- The original Java class included this method because it's not so simple to include state when calling from Java, so passing state in (whilst providing the same callback interface) is very convenient
- These were by default just extended to scripts

I don't really see the need in JS because people will pass closures that capture the state implicitly. Having these extra methods just mean extra code to maintain, plus a more confusing API for script writers.

florian-h05: Good idea, I also don't see the need. 

> plus a more confusing API for script writers.

That is a really good argumen.

florian-h05:
@jpg0 I addressed your review, you could please re-review?

digitaldan (approved):
LGTM

bundles/org.openhab.automation.jsscripting/src/main/java/org/openhab/automation/jsscripting/internal/JSScriptServiceUtil.java
```diff
@@ -0,0 +1,49 @@
...
+
+/**
+ * Utility class for providing easy access to script services.
+ *
+ * @author Florian Hotze - Initial contribution
+ */
+@Component(immediate = true)
+@NonNullByDefault
+public class JSScriptServiceUtil {
+    private static @Nullable Scheduler scheduler;
```
jpg0, line 33: This is a slightly odd use of OSGi. You have a component which is activated but then permanently stores (as static) the passed in references. Typically these references would not be static, however it means that consumers of this class would need it injecting too, rather that using static references to it.

I'm not sure it (a) this is a common pattern in openHAB (in which case go ahead with it), or (b) if it will cause any real problems or not. There _may_ be a problem with plugin reloading here, as the component will be reloaded and the fields will then be reset to new values, and any consumers (via the static ref) will see updated values, whereas consumers via injection will see two versions of `JSScriptServiceUtil` where the old one internally gets updated to the values of the new one. I am also not sure if there could be issues with components hanging on to static class references (of `JSScriptServiceUtil`) and hence preventing class collection by the GC.

The alternative to all this would be to (a) make these references non-static, (b) make `ThreadsafeTimers` a component and inject this class as a `@Reference` into it, and (c) fetch an instance of `ThreadsafeTimers` from the bundle context (rather than creating your own).

Personally I think this is a better way to go, however if this is an existing pattern in openHAB then I don't have an issue with it.

jpg0: Oh, and doing without static refs means that you can get rid of the `@Nullable` annotations as the references can never be null.

jpg0 (approved):
Only issue I have is with the use of OSGi. If it's a common pattern in openHAB I'm not concerned about it though.

bundles/org.openhab.automation.jsscripting/src/main/java/org/openhab/automation/jsscripting/internal/JSScriptServiceUtil.java
J-N-K: I agree with @jpg0. This is an ugly workaround that is used in the DSL Script engine because it would (at least) require very heavy refactoring to make the OSGi components available to XText. It should not be used if there is a better way.

florian-h05: Could I also make `JSScriptServiceUtil` non-static, initialize it from somewhere (OpenHABGraalJSScriptEngine) and use the constructor to get `Scheduler` and `ScriptExecution` from OSGi?

florian-h05: @jpg0 @J-N-K See commit https://github.com/openhab/openhab-addons/pull/13695/commits/fedbbe61fbbfec6b90c0e58692f1d40b06d31dbd for how I am retrieving the OSGi components.

jpg0: Your updated method is slightly better, but why not just inject the dependencies? The reason for injection is to break the coupling between classes, to allow things like testing them independently. Putting all the lookup code in the constructor now ties the class to OSGi too, as well as making it more complex. Are you doing that to make the lookup from JS easier? (There is an OSGi lookup ability in JS already.)

florian-h05: > why not just inject the dependencies?

You are referring to the solution you propose above, correctly?

> The alternative to all this would be to (a) make these references non-static, (b) make ThreadsafeTimers a component and inject this class as a @Reference into it, and (c) fetch an instance of ThreadsafeTimers from the bundle context (rather than creating your own).

So my question is, how to get the lock object into ThreadsafeTimers when I fetch it from the bundle context.

J-N-K: You could inject the `ScriptExecution` to the `GraalJSScriptEngineFactory` which is already a component and then pass it in the constructor to the `OpenhabGraalJSScriptEngine`. 

A "better" but more complex but better solution would be to make a factory component that creates a thread-safe component instance for each caller. The factory component could then get the `ScriptExecution` and also the `Scheduler` via injection and create a new lock object each time it creates a new instance.

florian-h05: > You could inject the ScriptExecution to the GraalJSScriptEngineFactory which is already a component and then pass it in the constructor to the OpenhabGraalJSScriptEngine.

Can I also inject the `JSScriptServiceUtil` (as long as it is a `@Component`) to the `GraalJSScriptEngineFactory`?

> A "better" but more complex but better solution would be to make a factory component that creates a thread-safe component instance for each caller. The factory component could then get the ScriptExecution and also the Scheduler via injection and create a new lock object each time it creates a new instance.

I am not sure if I able to implement this. I will try.

J-N-K: If you need assistance, feel free to contact me by private mail.

florian-h05: > If you need assistance, feel free to contact me by private mail.

Thanks for the offer.

Could you please check if my current implementation is okay?

Regarding the "better" solution: I am asking myself if is worth the additional work.
If we need access to more OSGi service, we can just extend `JSScriptServiceUtil` and do not have to change the injection.

J-N-K: Much better. I have made some minor suggestions in a PR to your fork.

jlaur:
@kaikreuzer - it seems that @jpg0 is missing as member of the addons contributors group, could you add him?

bundles/org.openhab.automation.jsscripting/src/main/java/org/openhab/automation/jsscripting/internal/JSScriptServiceUtil.java
florian-h05: @J-N-K See https://github.com/openhab/openhab-addons/pull/13695/commits/77bc18a8da5cd5615837ec561b0f3b5500e7f663 for the changes I spoke about in your PR.

florian-h05:
FYI: DCO is failing because of @J-N-K GitHub username is not his real name, which he used to signoff.

bundles/org.openhab.automation.jsscripting/src/main/java/org/openhab/automation/jsscripting/internal/JSScriptServiceUtil.java
J-N-K: LGTM

florian-h05: Okay, could you then please give an approving review?

florian-h05:
@jlaur This is ready for your review and merging now.

FYI: digitaldan and jpg0 approved at some point, after that I’ve changed the implementation of `JSScriptServiceUtil` (as jpg0 and J-N-K asked me to). J-N-K gave his approval in https://github.com/openhab/openhab-addons/pull/13695#discussion_r1020812558.
After all that reviews, I only upgraded the included openhab-js version and removed some unused code.

bundles/org.openhab.automation.jsscripting/src/main/java/org/openhab/automation/jsscripting/internal/threading/ThreadsafeTimers.java
```diff
@@ -95,8 +109,12 @@ public long setTimeout(Runnable callback, Long delay) {
...
     public long setTimeout(Runnable callback, Long delay, Object... args) {
         long id = lastId.incrementAndGet();
-        ScheduledCompletableFuture<Object> future = createFuture(id, ZonedDateTime.now().plusNanos(delay * 1000000),
-                callback);
+        ScheduledCompletableFuture<Object> future = scheduler.schedule(() -> {
+            synchronized (lock) {
+                callback.run();
+                idSchedulerMapping.remove(id);
+            }
+        }, identifier + ".timeout." + id, ZonedDateTime.now().plusNanos(delay * 1000000).toInstant());
```
jlaur, line 117:
```suggestion
        }, identifier + ".timeout." + id, Instant.now().plusMillis(delay));
```

jlaur (changes requested):
Thanks! Only a small simplification/optimization suggested.

florian-h05:
@jlaur I addressed your review, let's wait for the CI builds and then merge.

jlaur (approved):
LGTM

florian-h05:
@jlaur All checks have passed.

PR #13704, opened 2022-11-12
[tesla] Add steering wheel heater command support for Model 3/Y

theater:
- Added steering wheel heater commands support for Model 3 / Model Y
- Updated a throw clause so it has more context in it.

Implements #11331 
Tested on Model Y

Cheers,
Konstantin


bundles/org.openhab.binding.tesla/src/main/resources/OH-INF/thing/modely.xml
```diff
@@ -119,6 +120,7 @@
 			<channel id="valetmode" typeId="valetmode"/>
 			<channel id="valetpin" typeId="valetpin"/>
 			<channel id="wakeup" typeId="wakeup"/>
+			<channel id="wiperbladeheater" typeId="wiperbladeheater"/>
```
lolodomo, line 123: This is an undocumented change. 
Maybe you can create a separate PR for that?

bundles/org.openhab.binding.tesla/src/main/resources/OH-INF/thing/modely.xml
```diff
@@ -108,6 +108,7 @@
 			<channel id="shiftstate" typeId="shiftstate"/>
 			<channel id="sidemirrorheaters" typeId="sidemirrorheaters"/>
 			<channel id="smartpreconditioning" typeId="smartpreconditioning"/>
+			<channel id="steeringwheelheater" typeId="steeringwheelheater"/>
```
lolodomo, line 111: Please also update the channel type description (like you did in README) and the associated entry in i18n tesla.properties file. 

Please also mention in your PR description that you added that channel for model Y,

lolodomo (changes requested):

bundles/org.openhab.binding.tesla/src/main/resources/OH-INF/thing/modely.xml
lolodomo: Please also fix the channel type to remove the readonly attribute.

lolodomo:
And please sign off properly your contribution (DCO KO).

bundles/org.openhab.binding.tesla/src/main/resources/OH-INF/thing/modely.xml
theater: Where should I document this? Instead of separate PR, rather let's document it wherever applicable.

bundles/org.openhab.binding.tesla/src/main/resources/OH-INF/thing/modely.xml
theater: OK. Will do. Btw according to what I read this feature should be available also for Model 3 and as far as I know it was initially available for Model S/X. Should I transfer these changes to the other models (if they're not yet there)?

theater: Just checked now... modelX and S has the steeringwheelheater channel in the XMLs so question is now only for Model3 and Y (already done).

bundles/org.openhab.binding.tesla/src/main/resources/OH-INF/thing/modely.xml
lolodomo: In the description of your PR. 
But as it is something unrelated to the object of this PR, a separate PR would have been clearly better in particular for the release notes and to inform users. 
If you prefer a unique PR with multiple objects, at least describe all if them.

theater: OK. I got your point. It makese sense to have it as a separate PR so I did it. Please approve the other one first and then we can go over this one. I think I've fixed your comments (mostly).

bundles/org.openhab.binding.tesla/src/main/resources/OH-INF/thing/modely.xml
theater: https://github.com/openhab/openhab-addons/pull/13711

bundles/org.openhab.binding.tesla/src/main/resources/OH-INF/thing/modely.xml
lolodomo: I believe you fully misunderstood my comment. 
My comment was not about steeringwheelheater but about wiperbladeheater you also added. Is there an obvious link between them? 

For simply steeringwheelheater, there is of course no need to create several PRs.

bundles/org.openhab.binding.tesla/src/main/resources/OH-INF/thing/modely.xml
lolodomo: You missed me! Why creating another PR for exactly the same thing? 

I had the feeling you added a second new channel corresponding to a fully different feature without saying it.

lolodomo: > Btw according to what I read this feature should be available also for Model 3 and as far as I know it was initially available for Model S/X. Should I transfer these changes to the other models (if they're not yet there)?

Would make sense.

theater: > You lost me! Why creating another PR for exactly the same thing?
> 
> I had the feeling you added a second new channel corresponding to a fully different feature without saying it.

OK mate. I will remove this channel from the XML. I thought that you want the metadata to be pushed separately in order to appear in the release notes of the candidate with more details. Initially I wanted to implement also the wiper blades heater functionality but I couldn't find the command API call for it. I guess it makes sense to leave it for another PR.

bundles/org.openhab.binding.tesla/src/main/resources/OH-INF/thing/channels.xml
```diff
@@ -607,7 +607,7 @@
 		<item-type>Switch</item-type>
 		<label>Steering Wheel Heater</label>
 		<description>Indicates if the steering wheel heater is switched on</description>
```
lolodomo, line 609: Please keep the same description there, in the i18n file and preferably in documentation (README).
With your change, we have now 3 different descriptions.
I would suggest to keep the one you set in the i18n file, that is:  Turns On/Off the steering wheel heater

lolodomo (changes requested):

lolodomo:
@kgoderis : would you like to review this small enhancement or can I merge as soon as the channel description is consistent in all places ?

lolodomo (approved):
LGTM, thank you

PR #13705, opened 2022-11-12
[ecobee] Add air quality channels for Ecobee Smart Thermostat Premium

robnielsen:
The Ecobee Smart Thermostat Premium returns the following additional runtime parameters: 
```
"runtime": {
  ...
  "actualVOC": 1045, 
  "actualCO2": 738,
  "actualAQAccuracy": 4,
  "actualAQScore": 73,
...
}
```
This PR creates the associated channels for them.

### DIFF
--- a/bundles/org.openhab.binding.ecobee/src/main/resources/OH-INF/thing/thing-types.xml
+++ b/bundles/org.openhab.binding.ecobee/src/main/resources/OH-INF/thing/thing-types.xml
@@ -320,12 +320,12 @@
 		<state readOnly="true" pattern="%.1f %unit%"/>
 	</channel-type>
 	<channel-type id="actualAQAccuracy" advanced="true">
-		<item-type>Number:Dimensionless</item-type>
+		<item-type>Number</item-type>
 		<label>Actual Air Quality Accuracy</label>
 		<state readOnly="true" pattern="%.0f"/>
 	</channel-type>
 	<channel-type id="actualAQScore" advanced="true">
-		<item-type>Number:Dimensionless</item-type>
+		<item-type>Number</item-type>
 		<label>Actual Air Quality Score</label>
 		<state readOnly="true" pattern="%.0f"/>
 	</channel-type>
--- a/bundles/org.openhab.binding.ecobee/src/main/resources/OH-INF/thing/thing-types.xml
+++ b/bundles/org.openhab.binding.ecobee/src/main/resources/OH-INF/thing/thing-types.xml
@@ -322,22 +322,22 @@
 	<channel-type id="actualAQAccuracy" advanced="true">
 		<item-type>Number</item-type>
 		<label>Actual Air Quality Accuracy</label>
-		<state readOnly="true" pattern="%.0f"/>
+		<state readOnly="true" pattern="%d"/>
 	</channel-type>
 	<channel-type id="actualAQScore" advanced="true">
 		<item-type>Number</item-type>
 		<label>Actual Air Quality Score</label>
-		<state readOnly="true" pattern="%.0f"/>
+		<state readOnly="true" pattern="%d"/>
 	</channel-type>
 	<channel-type id="actualCO2" advanced="true">
 		<item-type>Number:Dimensionless</item-type>
 		<label>Actual CO2</label>
-		<state readOnly="true" pattern="%.0f %unit%"/>
+		<state readOnly="true" pattern="%d %unit%"/>
 	</channel-type>
 	<channel-type id="actualVOC" advanced="true">
 		<item-type>Number:Dimensionless</item-type>
 		<label>Actual VOC</label>
-		<state readOnly="true" pattern="%.0f %unit%"/>
+		<state readOnly="true" pattern="%d %unit%"/>
 	</channel-type>
 
 	<channel-group-type id="settings">
--- a/bundles/org.openhab.binding.ecobee/src/main/resources/OH-INF/thing/thing-types.xml
+++ b/bundles/org.openhab.binding.ecobee/src/main/resources/OH-INF/thing/thing-types.xml
@@ -202,6 +202,10 @@
 			<channel id="desiredHeatRangeHigh" typeId="desiredHeatRangeHigh"/>
 			<channel id="desiredCoolRangeLow" typeId="desiredCoolRangeLow"/>
 			<channel id="desiredCoolRangeHigh" typeId="desiredCoolRangeHigh"/>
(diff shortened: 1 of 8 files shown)


bundles/org.openhab.binding.ecobee/src/main/resources/OH-INF/thing/thing-types.xml
```diff
@@ -315,6 +319,26 @@
 		<label>Desired Cool Range High</label>
 		<state readOnly="true" pattern="%.1f %unit%"/>
 	</channel-type>
+	<channel-type id="actualAQAccuracy" advanced="true">
+		<item-type>Number:Dimensionless</item-type>
```
lolodomo, line 323: This channel has no unit set when updating its state, so it should be linked to an item of type "Number", not "Number:Dimensionless".
README should be also updated.

bundles/org.openhab.binding.ecobee/src/main/resources/OH-INF/thing/thing-types.xml
```diff
@@ -315,6 +319,26 @@
...
+		<label>Actual Air Quality Accuracy</label>
+		<state readOnly="true" pattern="%.0f"/>
+	</channel-type>
+	<channel-type id="actualAQScore" advanced="true">
+		<item-type>Number:Dimensionless</item-type>
```
lolodomo, line 328: This channel has no unit set when updating its state, so it should be linked to an item of type "Number", not "Number:Dimensionless".
README should be also updated.

lolodomo (changes requested):

bundles/org.openhab.binding.ecobee/src/main/resources/OH-INF/thing/thing-types.xml
robnielsen: Fixed

bundles/org.openhab.binding.ecobee/src/main/resources/OH-INF/thing/thing-types.xml
robnielsen: Fixed

robnielsen:
@lolodomo, I made the requested changes from `Number:Dimensionless` to `Number`.  This caused the values to now have a .0  in the events.log file:

```
2022-11-13 09:14:37.991 [INFO ] [openhab.event.ItemStateChangedEvent ] - Item 'eAirQualityScore' changed from 173 to 168
2022-11-13 09:17:43.278 [INFO ] [openhab.event.ItemStateChangedEvent ] - Item 'eAirQualityScore' changed from 168 to 164
2022-11-13 09:21:30.913 [INFO ] [openhab.event.ItemStateChangedEvent ] - Item 'eAirQualityScore' changed from NULL to 164
2022-11-13 09:21:30.939 [INFO ] [openhab.event.ItemStateChangedEvent ] - Item 'eAirQualityAccuracy' changed from NULL to 4
2022-11-13 09:21:31.311 [INFO ] [openhab.event.ItemStateChangedEvent ] - Item 'eAirQualityScore' changed from 164 to 164.0
2022-11-13 09:21:31.330 [INFO ] [openhab.event.ItemStateChangedEvent ] - Item 'eAirQualityAccuracy' changed from 4 to 4.0
2022-11-13 09:22:31.164 [INFO ] [openhab.event.ItemStateChangedEvent ] - Item 'eAirQualityScore' changed from 164.0 to 165.0
2022-11-13 09:25:36.310 [INFO ] [openhab.event.ItemStateChangedEvent ] - Item 'eAirQualityScore' changed from 165.0 to 166.0
```

lolodomo:
> This caused the values to now have a .0 in the events.log file:

Probably because you are using EcobeeUtils.undefOrDecimal which works with double rather than integer.

lolodomo:
I have also a doubt about the dimension you choose for co2. I have to find what is done in other bindings handling co2.

lolodomo:
Maybe Number:Density is the appropriate dimension?

robnielsen:
> I have also a doubt about the dimension you choose for co2. I have to find what is done in other bindings handling co2.

> Maybe Number:Density is the appropriate dimension?

Netatmo uses Number:Dimensionless as well for co2.

lolodomo:
> Netatmo uses Number:Dimensionless as well for co2.

Might be a bug in Netatmo README. In fact, the binding retrieves the dimension from the unit. 
In Openeearhermap, Number:Density is used. 
Number:Density is I think the right dimension but I will check in uom.

lolodomo:
I checked and Units.PARTS_PER_MILLION / Units.PARTS_PER_BILLION are Dimensionless. So that is fine.

bundles/org.openhab.binding.ecobee/src/main/resources/OH-INF/thing/thing-types.xml
```diff
@@ -315,6 +319,26 @@
 		<label>Desired Cool Range High</label>
 		<state readOnly="true" pattern="%.1f %unit%"/>
 	</channel-type>
+	<channel-type id="actualAQAccuracy" advanced="true">
+		<item-type>Number</item-type>
+		<label>Actual Air Quality Accuracy</label>
+		<state readOnly="true" pattern="%.0f"/>
```
lolodomo, line 325: You should normally rather use `%d` if the value is an integer. Same for actualAQScore.

robnielsen:
>>Netatmo uses Number:Dimensionless as well for co2.
>
>Might be a bug in Netatmo README. In fact, the binding retrieves the dimension from the unit.
>In Openweathermap, Number:Density is used.
>Number:Density is I think the right dimension but I will check in uom.

According to https://www.openhab.org/docs/concepts/units-of-measurement.html, Dimensionless is correct for ppm

bundles/org.openhab.binding.ecobee/src/main/resources/OH-INF/thing/thing-types.xml
robnielsen: fixed

robnielsen:
>>> This caused the values to now have a .0 in the events.log file:
>>
>>Probably because you are using EcobeeUtils.undefOrDecimal which works with double rather than integer.
>
>Did you try to add a new method EcobeeUtils.undefOrDecimal taking as parameter an Integer ?

Yes, I added a new EcobeeUtils.undefOrLong method that solved it.

lolodomo (approved):
LGTM, thank you

PR #13710, opened 2022-11-13
[evcc] Fix data type issues in DTO

### LINKED ISSUE #13646: [evcc] Failed to get state after evcc update to 0.106
## Expected Behavior
evcc device is correctly connected

## Current Behavior
after updating to evcc 0.106 the binding fails. Trace log shows a failure message and the result of the underlying call to `api/state`

See https://github.com/evcc-io/evcc/issues/5035

florian-h05:
Fixes #13646.

This PR fixes an issue, where evcc changed their API to return a float instead of an int for `vehicleCapacity`, but the DTO expected an int and parsing the JSON response failed.
To avoid such issues in the future, all JSON response fields that might be changed to float in the future now use float instead of int.
Furthermore, the usage of double was replaced by float, as the higher precision of double is not required.

@pauxus I will also publish this bugfix to the addon marketplace, so that you can update the binding ASAP. See https://community.openhab.org/t/evcc-binding-electric-vehicle-charging-control/135209.
@andig FYI: This fixes https://github.com/evcc-io/evcc/issues/5035.

### DIFF
--- a/bundles/org.openhab.binding.evcc/src/main/java/org/openhab/binding/evcc/internal/EvccHandler.java
+++ b/bundles/org.openhab.binding.evcc/src/main/java/org/openhab/binding/evcc/internal/EvccHandler.java
@@ -52,7 +52,7 @@ import org.slf4j.LoggerFactory;
  * The {@link EvccHandler} is responsible for handling commands, which are
  * sent to one of the channels.
  *
- * @author Florian Hotze - Initial contribution
+ * @author Florian Hotze - Initial contribution; Avoid data type issues by using float instead of int
  */
 @NonNullByDefault
 public class EvccHandler extends BaseThingHandler {
@@ -85,7 +85,7 @@ public class EvccHandler extends BaseThingHandler {
                 return;
             }
             String channelIdWithoutGroup = channelUID.getIdWithoutGroup();
-            int loadpoint = Integer.parseInt(groupId.toString().substring(9));
+            int loadpoint = Integer.parseInt(groupId.substring(9));
             EvccAPI evccAPI = this.evccAPI;
             if (evccAPI == null) {
                 return;
@@ -186,7 +186,7 @@ public class EvccHandler extends BaseThingHandler {
      */
     private void refresh() {
         logger.debug("Running refresh job ...");
-        EvccAPI evccAPI = null;
+        EvccAPI evccAPI;
         evccAPI = this.evccAPI;
         if (evccAPI == null) {
             return;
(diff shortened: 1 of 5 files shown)


bundles/org.openhab.binding.evcc/src/main/java/org/openhab/binding/evcc/internal/EvccHandler.java
```diff
@@ -52,7 +52,7 @@
  * The {@link EvccHandler} is responsible for handling commands, which are
  * sent to one of the channels.
  *
- * @author Florian Hotze - Initial contribution
+ * @author Florian Hotze - Initial contribution; Avoid data type issues by using float instead of int
```
jlaur, line 55: I think this change is not needed as you are already mentioned as original contributor. Only reason for change would be another contributor making a substantial contribution: https://www.openhab.org/docs/developer/guidelines.html#c-documentation

@openhab/add-ons-maintainers - please correct me if I'm wrong here.

florian-h05: I am not sure, but at JS Scripting, we also did some smaller changes without adding a new author comment. I'll remove that.

The contribution guidelines also say:

> Data-transfer-objects (DTOs map from JSON/XML to Java classes) do not require JavaDoc.

bundles/org.openhab.binding.evcc/src/main/java/org/openhab/binding/evcc/internal/api/EvccAPI.java
```diff
@@ -32,13 +32,13 @@
 /**
  * The {@link EvccAPI} is responsible for API calls to evcc.
  * 
- * @author Florian Hotze - Initial contribution
+ * @author Florian Hotze - Initial contribution; Avoid data type issues by using float instead of int
```
jlaur, line 35: This one remaining.

jlaur (changes requested):
Thanks! One comment - everything else looks good to me.

bundles/org.openhab.binding.evcc/src/main/java/org/openhab/binding/evcc/internal/api/EvccAPI.java
florian-h05: Oops, I removed it.

jlaur (approved):
LGTM

florian-h05:
@jlaur Checks have also passed now.

PR #1529, opened 2016-12-05
[Onkyo] Fix various issues

marcelrv:
Fixes various issues for Onkyo binding
* Fix lookup of listening modes & audio sources below #10
* Adds several missing listening modes
* Adds several missing sources
* Fix Exceptions statusUpdateReceived in case of N/A
* Fix issue when looking up sources for zone2

Signed-off-by: Marcel Verpaalen <marcel@verpaalen.com>

PR #13714, opened 2022-11-14
[jsscripting] Fix failure on some platforms & JDKs

florian-h05:
## Description

The community reported several cases where JS Scripting was not working due to some issue with the injection of the global script.
This issue seems to only occur on armv7l (e.g. Raspberry Pi 32bit) and OpenJDK 11.0.16.
Investigation showed that the occurrence of the problem depends on the GraalJS version, so we have to downgrade the version. (FYI: The version was upgraded in https://github.com/openhab/openhab-addons/pull/13582.)

See https://community.openhab.org/t/js-scripting-all-scripts-stop-working-when-upgrading-to-3-4-0-m4/140837.

## Testing/Verification

I have verified that JS Scripting is now properly working on my Raspberry Pi 3B:
- `uname -a`:
  ```
  Linux homeberrypi 5.10.103-v7+ #1529 SMP Tue Mar 8 12:21:37 GMT 2022 armv7l GNU/Linux
  ```
- `java --version`:
  ```
  openjdk 11.0.16 2022-07-19
  OpenJDK Runtime Environment (build 11.0.16+8-post-Raspbian-1deb10u1)
  OpenJDK Server VM (build 11.0.16+8-post-Raspbian-1deb10u1, mixed mode)
  ```

It is still working on my dev system (`Linux florian 5.19.16-200.fc36.x86_64 #1 SMP PREEMPT_DYNAMIC Sun Oct 16 22:50:04 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux`) with `openjdk 11.0.17 2022-10-18`.

Anyway, I would prefer that some of the users that reported that problem in the community also verify that it is working now.
From 6 users who reported the problem (including myself), until now 4 have verified that the fix works.

@jlaur Can you also verify the fix? Artifact download link is at the community.

### DIFF
--- a/bundles/org.openhab.automation.jsscripting/src/main/resources/node_modules/@jsscripting-globals.js
+++ b/bundles/org.openhab.automation.jsscripting/src/main/resources/node_modules/@jsscripting-globals.js
@@ -1,189 +1,189 @@
 // ThreadsafeTimers is injected into the JS runtime
 
 (function (global) {
-    'use strict';
+  'use strict';
 
-    // Append the script file name OR rule UID depending on which is available
-    const defaultIdentifier = "org.openhab.automation.script" + (globalThis["javax.script.filename"] ? ".file." + globalThis["javax.script.filename"].replace(/^.*[\\\/]/, '') : globalThis["ruleUID"] ? ".ui." + globalThis["ruleUID"] : "");
-    const System = Java.type('java.lang.System');
-    const formatRegExp = /%[sdj%]/g;
-    // Pass the defaultIdentifier to ThreadsafeTimers to enable naming of scheduled jobs
-    ThreadsafeTimers.setIdentifier(defaultIdentifier);
+  // Append the script file name OR rule UID depending on which is available
+  const defaultIdentifier = 'org.openhab.automation.script' + (globalThis['javax.script.filename'] ? '.file.' + globalThis['javax.script.filename'].replace(/^.*[\\\/]/, '') : globalThis.ruleUID ? '.ui.' + globalThis.ruleUID : '');
+  const System = Java.type('java.lang.System');
+  const formatRegExp = /%[sdj%]/g;
+  // Pass the defaultIdentifier to ThreadsafeTimers to enable naming of scheduled jobs
+  ThreadsafeTimers.setIdentifier(defaultIdentifier);
 
-    function createLogger(name = defaultIdentifier) {
-        return Java.type("org.slf4j.LoggerFactory").getLogger(name);
+  function createLogger (name = defaultIdentifier) {
+    return Java.type('org.slf4j.LoggerFactory').getLogger(name);
+  }
+
+  // User configurable
+  let log = createLogger();
+
+  function stringify (value) {
+    try {
+      if (Java.isJavaObject(value)) {
+        return value.toString();
+      } else {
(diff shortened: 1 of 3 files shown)


florian-h05:
@jpg0 Can you have a look at this PR?

jlaur:
@florian-h05 - I can confirm that jsscripting is working again for me after the downgrade. Thanks for caring care that fast!

jlaur (approved):
Thanks for the fix!

florian-h05:
@jlaur Now 4 of the 6 users that have reported that issue verified that the fix is working for them, so I think we can expect this fix to work properly. Feel free to merge!

jlaur:
@florian-h05 - I cannot review `@jsscripting-globals.js`, so hoping for @jpg0 or @digitaldan to have a look before merging.

florian-h05:
Although it is only some linting (I am not sure why GitHub shows that large changes), I understand your point.

bundles/org.openhab.automation.jsscripting/src/main/resources/node_modules/@jsscripting-globals.js
```diff
@@ -1,189 +1,189 @@
 // ThreadsafeTimers is injected into the JS runtime
 
 (function (global) {
-    'use strict';
```
jlaur, line 4: Is indentation changed from 2 to 4 spaces on purpose?

florian-h05: Not really. I have applied the semistandard JavaScript linter style (that is the same linter we use at the openHAB JavaScript library).
The changes are mostly intendation, double/single quotes as well as let and const instead of var (before linting, var and let/const were used mixed).

florian-h05:
Here are some screenshots from IntelliJ, which display the changes much better than GitHub (GitHub says that I replaced blocks of code):

image
image
image
image

bundles/org.openhab.automation.jsscripting/src/main/resources/node_modules/@jsscripting-globals.js
florian-h05: I did this to have the same codestyle, and I personally prefer the "clean" code style that linters enforce.

jlaur: Not needed, was just curious if done by purpose to achieve some specific code style in JavaScript, or by accident. Thanks for explaining.

jlaur:
> Here are some screenshots from IntelliJ, which display the changes much better than GitHub (GitHub says that I replaced blocks of code):

Okay, I didn't realize changes were that minimal, I guess I went through too quickly. Thanks for sharing, and let's merge!

dilyanpalauzov:
If the Java 11.0.16 problem in GraalVM does not happen with Java 11.0.18, then the version of GraalVM could be lifted.

It is now anyway Java 17 required, so the reason to stick to GraalVM version 22.0.0.2 is likely not relevant any more.

florian-h05:
The problem is not only related to Java version, but more to platform. It seems that GraalVM dropped 32-Bit support in newer versions and I once also read about that.

PR #13726, opened 2022-11-15
[jdbc] Improve error handling safety

jlaur:
This is a continuation of #13719 with two purposes:
- Introduce abstraction by wrapping `YankSQLException` into custom exception. This way Yank can be refactored or replaced without any impact on callers.
- Make custom exception checked to enforce handling - like `SQLException`.

Related to #13718

PR #13728, opened 2022-11-15
[jdbc] Fix possible index corruption when uninitialized

### LINKED ISSUE #13727: [jdbc] Index corruption after failed initialization
If initialization will fail for any reason, first attempt to store an item will result in index corruption by creating a duplicate row.

## Expected Behavior
When uninitialized, any attempt to store an item should rejected. Index should not be changed.

## Current Behavior
After failed initialization, index is untouched:
| ItemId | ItemName |
| -----: | -------- |
|      1 | TestItem |

After first attempt to store this item:
| ItemId | ItemName |
| -----: | -------- |
|      1 | TestItem |
|      2 | TestItem |

No table is created, so only Item0001 exists at this point, but since index is corrupted, it will not work after this.

The reason is that the mapping table (item name -> table name) is not populated, thus lookup will fail, and this triggers the logic to create the missing table for the item.

## Possible Solution
`store()` could check if initialized, and fail gracefully if not.

## Steps to Reproduce
Developer test:
- Make sure query `sqlGetItemTables` will not work, for example by putting "X" just before "SELECT".
- Install.
- Update item state on any existing persisted item.

jlaur:
Depends on #13726
Fixes #13727


bundles/org.openhab.persistence.jdbc/src/main/java/org/openhab/persistence/jdbc/internal/JdbcMapper.java
```diff
@@ -316,20 +317,23 @@ private void populateItemNameToTableNameMap() throws JdbcSQLException {
...
+        }
 
         // Table already exists - return the name
+        String tableName = itemNameToTableNameMap.get(itemName);
         if (!Objects.isNull(tableName)) {
             return tableName;
         }
 
-        logger.debug("JDBC::getTable: no table found for item '{}' in sqlTables", itemName);
+        logger.debug("JDBC::getTable: no table found for item '{}' in itemNameToTableNameMap", itemName);
```
lsiepel, line 332: Minor question; `itemNameToTableNameMap` is not known to anyone except if you have the java code. sqlTables is more meaningfull, as the table is actually not in de db. Is this intended?

lsiepel (approved):
Looks good to me, only minor question about logging message.

bundles/org.openhab.persistence.jdbc/src/main/java/org/openhab/persistence/jdbc/internal/JdbcMapper.java
jlaur: This is a debug log message. `sqlTables` was renamed to `itemNameToTableNameMap` in #13544, but I missed this log message.

lsiepel: Ah thanks, I see now.

lolodomo (approved):
LGTM, thank you

PR #13729, opened 2022-11-16
[networkupstools] Add missing status

mlobstein:
Add state option display string for my APC UPS that is reporting an alarm indicating the battery needs replacing.

image


jlaur (approved):
Thanks!

PR #13730, opened 2022-11-16
[awattar] Fix that allows bridge to be added via UI.

dougculnane:
As per discussion in the forum: https://community.openhab.org/t/awattar-binding-beta-and-discussion/110497/70

the awatar bridge can not be added as a thing with the UI 3.  This fixes (or works around) the problem.

Signed-off-by: Doug Culnane <doug@culnane.net>


bundles/org.openhab.binding.awattar/src/main/resources/OH-INF/thing/bridge-types.xml
```diff
@@ -1,14 +0,0 @@
-<?xml version="1.0" encoding="UTF-8"?>
-<thing:thing-descriptions bindingId="openweathermap"
```
Hilbrand, line 2: The reason it didn't work is that the `bindingId` is wrong in this file

dougculnane: @Hilbrand OK.  That is the real problem then. My fix was more of a work around that worked.  

What is better, my fix or changing the bindingId?

Hilbrand: In general there is no best choice. Either solution is ok. In this case it probably makes more sense to have everything in one file as it's only a small set of things. I would say your fix is ok.

Hilbrand (approved):
LGTM

bundles/org.openhab.binding.awattar/src/main/resources/OH-INF/thing/bridge-types.xml
Wolfgang1966: Sorry for the copy-paste error, was my bad :-(

Wolfgang1966:
LGTM

Wolfgang1966 (approved):

bundles/org.openhab.binding.awattar/src/main/resources/OH-INF/thing/bridge-types.xml
dougculnane: @Wolfgang1966 no problem it gave me an easy fix (work around).

lolodomo (approved):
LGTM, thank you

PR #13734, opened 2022-11-16
[jdbc] Fix date filter to be inclusive

### LINKED ISSUE #9906: [jdbc] JDBC persistence ignores operator set in FilterQuery
Given that `FilterQuery` makes it possible to specify additional operator I am convinced that ignoring it is a bug. There is currently no way to tell JDBC persistence to be inclusive.

The generated query always uses `>` for begin date and `<` for end date. There are cases where these should be `=>` or `<=`.

jlaur:
Fixes #9906


lolodomo:
@cweitkamp : I let you merge as you already reviewed and approved.

PR #13737, opened 2022-11-17
[jdbc] Make itemsManageTable configurable

### LINKED ISSUE #9637: [jdbc.mysql] MySQL migration - Items table is not used for index but creates a new table called items
Migration fro OH2.5.11 to 3.0

As part of the upgrade and persistence migration from old mysql to jdbc, you have to change the config to ensure it keeps your item mappings.

However, even if you use the recommended config
```

tableNamePrefix=Item
tableUseRealItemNames=false
tableIdDigitCount=0
```
it uses an index table called items and not Items. As a result, everything is out of whack as OH reindexes all items requiring persistence. 

I assume this is not by-design?

jlaur:
Items are by default prefixed with `item` and the items management table is named `items`.

When migrating from the MySQL Persistence Service, users are advised to use `tableNamePrefix=Item` in configuration, but it is not mentioned that they need to manually rename the items management table from `Items` to `items`. Failing to do so will result in a new empty `items` table being created alongside the existing `Items` table. This is counter-intuitive.

This pull request introduces a configuration option to set the name of the items management table, so existing `Items` table from a migration can be used as is. This eases the migration and adds more flexibility as any desired name can be configured, instead of the items management table name being hardcoded.

Fixes #9637

### DIFF
--- a/bundles/org.openhab.persistence.jdbc/src/main/resources/OH-INF/config/config.xml
+++ b/bundles/org.openhab.persistence.jdbc/src/main/resources/OH-INF/config/config.xml
@@ -134,6 +134,10 @@
 
 		<!--
 			# T A B L E O P E R A T I O N S
+			# Items Manage Table (optional, default: "items")
+			# for Migration from MYSQL-Bundle set to 'Items'.
+			#itemsManageTable=Items
+
 			# Tablename Prefix String (optional, default: "item")
 			# for Migration from MYSQL-Bundle set to 'Item'.
 			#tableNamePrefix=Item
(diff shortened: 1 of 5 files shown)


bundles/org.openhab.persistence.jdbc/src/main/resources/OH-INF/config/config.xml
```diff
@@ -156,6 +160,11 @@
 			# USE WITH CARE! Deactivate after Renaming is done!
 			#rebuildTableNames=true
 		-->
+		<parameter name="itemsManageTable" type="text">
+			<label>Items Manage Table</label>
+			<description><![CDATA[Items Manage Table <br>(optional, default: "items"). <br>
```
lolodomo, line 165: No need to set the default tag?

jlaur: You might have a point here. Currently this is defaulted in code the same way as for all other existing configuration parameters.

I think this could be improved though, and I know there are issues when removing a parameter from a configuration file - this will not go back to default in some cases.

Reworking this is probably outside the scope of this pull request, but I can have a look at it.

lolodomo: Ok, let's keep that for a future enhancement.

PR #13735, opened 2022-11-17
[vitotronic] Add channel "currentmode" to "circuit" thing

Fejitatete:

PR #13739, opened 2022-11-18
[vitotronic] Add channel "currentmode" to thing "circuit"

Fejitatete:
Signed-off-by: Fejitatete <fejitatete@msn.com>


lolodomo:
A new thing channel with no Java code to manage it?

jlaur:
See also #13735.

Fejitatete:
Sry guys, I don't understand why this need some Java code. Could you explain or take a look in sources plz.
Thx

Hilbrand:
@Fejitatete is right. It doesn't need code. The binding derives the channel from the data it receives from the device. It would be nice to add the channel to the readme. Also the channel seems to represent a mode. I suspect the mode numbers represent textuale description of what each number means. In that case you can add the option to the channel, so the user sees the textual description instead of just a number

bundles/org.openhab.binding.vitotronic/src/main/resources/OH-INF/thing/thing-types.xml
```diff
@@ -528,6 +529,13 @@
 		<state pattern="%d" readOnly="false" max="4" min="0" step="1"/>
 	</channel-type>
 
+	<channel-type id="currentmode">
+		<item-type>Number</item-type>
+		<label>Currentmode</label>
+		<description>Currentmode</description>
```
lolodomo, line 535: Description to be removed if it adds no value to the label.
At least, use correct English, that is current mode (2 words).

bundles/org.openhab.binding.vitotronic/src/main/resources/OH-INF/thing/thing-types.xml
```diff
@@ -528,6 +529,13 @@
...
+		<state pattern="%d" readOnly="true" max="3" min="0" step="1"/>
```
lolodomo, line 536: As it is read only, min/max/step have no real interest, U believe,

lolodomo (changes requested):

lolodomo:
The vitotronic.properties file has to be updated to include the new channel type and the 4 states.

New channel has to be also added in documentation (README.md file).

@Hilbrand : can you take the lead to review/merge this PR as you are apparently more aware than us how it is implemented ?

Fejitatete:
> The vitotronic.properties file has to be updated to include the new channel type and the 4 states.
> 
> New channel has to be also added in documentation (README.md file).
> 
> @Hilbrand : can you take the lead to review/merge this PR as you are apparently more aware than us how it is implemented ?

lolodomo:
No chance to have a proper signing ?

lolodomo (approved):
Thing definition and translation file are not in sync (label/description) for the new channel type.
I will merge as it is and provide a fix in a separate PR.

lolodomo:
Making an exception and merge the PR even if not properly signed off.

Issue #13682, opened 2022-11-08
[lifx] LIFX thing stays in offline state after power break


paulianttila:
My openHAB instance is UPS battery backed-up, so it doesn't go down if there is a "short" power break, but LIFX devices obviusly will. When LIFX devices go down because of power break, biding will recognise that and but things in offline state (COMMUNICATION_ERROR). When power resume, thing status stays in OFFLINE state. 

I enabled trace level logs to the binding and following can be seen from the log.

```
19:06:42.487 [TRACE] [internal.LifxLightOnlineStateUpdater] - 192.168.40.35 : Light is not online, unicasting request
19:06:42.487 [TRACE] [.lifx.internal.util.LifxSelectorUtil] - 192.168.40.35 : Sending packet type 'GetServiceRequest' from '/172.18.0.12:40024' to '/192.168.40.35:56700' for 'D0:73:D5:24:86:58' with sequence '187' and source '2f49573d'
```

My LIFX devices are in different VLAN than openHAB instance, therefore binding use unicast communication (host ip is set to the thing configuration).

I can see that UDP packet is sent to the LIFX device (tcpdump), but device doesn't response it. Should all LIFX devices support GetServiceRequest?

LIFX details:

```
productVersion  131072
productId       27
hostVersion     2.80
host            192.168.40.35
wifiVersion     2.80
vendorId        1
vendorName      LIFX
productName     LIFX A19
```

Because thing state stays in offline, binding doesn't poll the state either

```
19:07:00.078 [TRACE] [nternal.LifxLightCurrentStateUpdater] - 192.168.40.35 : The light is not online, there is no point polling it
```

After I send command to the light, thing status will be changed to online.

```
19:42:12.811 [DEBUG] [.lifx.internal.LifxLightStateChanger] - 192.168.40.35 : Sending SetLightPowerRequest packet

19:42:12.812 [TRACE] [.lifx.internal.util.LifxSelectorUtil] - 192.168.40.35 : Sending packet type 'SetLightPowerRequest' from '/172.18.0.12:40024' to '/192.168.40.35:56700' for 'D0:73:D5:24:86:58' with sequence '74' and source '2f49573d'

19:42:12.836 [TRACE] [ternal.LifxLightCommunicationHandler] - 192.168.40.35 : Packet type 'AcknowledgementResponse' received from '/192.168.40.35:56700' for 'D0:73:D5:24:86:58' with sequence '74' and source '2f49573d'

[...]

PR #13744, opened 2022-11-19
[lifx] Set tagged flag accordingly for unicast and broadcast messages

paulianttila:
Signed-off-by: Pauli Anttila <pauli.anttila@gmail.com>


paulianttila:
Fix #13682

lolodomo:
@wborn : as the maintainer of this binding, you are welcome for review that fix.

wborn:
I will have a look soon. :slightly_smiling_face:

wborn:
Hi @paulianttila, do I understand correctly that the light does initially go ONLINE but due to the short power interruption stops responding to `GetServiceRequest` ? Is it also reproducible by manually toggling power on a light or would that cause the light to restart properly?

paulianttila:
@wborn in fact, light doesn't go to ONLINE state after lifx binding start in normal ciscumstances either (not power break situation). I haven't recognise that so far as it haven't cause any issue for me. The reason is that lifx lights (which I have) doesn't 'never' response to unicast GetServiceRequests. 

Binding send unicast GetServiceRequests also during start in LifxLightCommunicationHandler, but there is no response for that either, so maybe that should be changed as well.

Here are the test result with two different bulb versions which I have; LIFX A19 and LIFX Color 1000. 

Full log with unmodified lifx binding [lifx-org.log](https://github.com/openhab/openhab-addons/files/10146064/lifx-org.log)

Full log with modified lifx binding which sent both GetServiceRequests and GetEchoRequest in LifxLightOnlineStateUpdater (also trace logs 'Channel is ready for reading' and 'No datagram is available' was removed in the test version). [lifx-mod.log](https://github.com/openhab/openhab-addons/files/10146072/lifx-mod.log)

Lifx [documents](https://lan.developer.lifx.com/docs/querying-the-device-for-data#discovery) says that GetService is typically used for broadcast, so maybe intentation was not even use the message for unicast.

wborn:
Thanks for the info @paulianttila. I tried to reproduce the issue in a VM so openHAB is in its own subnet to be sure broadcasts don't work. I was unable to reproduce your issue a couple of days ago like this and all seemed fine just like when I implemented the unicast.

So I thought let's try to reproduce it with a Color 1000 which I also have. However when I tried today all the previously configured lights were offline after starting openHAB in the VM. It seems that when you first add the lights by IP they are online... but if you then restart the LIFX bundle the lights remain offline. :slightly_frowning_face:

I'll have a look at the code why this is. Are they online for you too after first adding them?

paulianttila:
> Are they online for you too after first adding them?

No, thing stays offline state also when add it. But if you bind any channels from the thing, then state immediately changes to online state.

wborn:
>  Did you use files or UI for your tests?

So far I've tested by adding things using the UI.

paulianttila:
@wborn I think I found the root cause. Reason why bulb doesn't response to unicast GetServiceRequest is the fact that `tagged` flag is not correctly set.

Lifx [document](https://lan.developer.lifx.com/docs/packet-contents) states following

> The tagged field is a boolean flag that indicates whether the Frame Address target field is being used to address an individual device or all devices. When you broadcast a message to the network (i.e. broadcasting a [GetService (2)](https://lan.developer.lifx.com/docs/querying-the-device-for-data#getservic--e-packet-2) for discovery) you should set this field to 1 and for all other messages you should set this to 0.

Binding use value 1 also for unicast requests. I changed sendPacket methods to set the tagged flag accordingly and now my bulbs response to  unicast GetServiceRequests.

wborn:
The fix works fine for me too! :+1:

I think it is better to also remove the calls to `setTagged` from the Packets because the value will always be overwritten with these changes. I've created https://github.com/paulianttila/openhab-addons/pull/3 for that.

wborn (approved):
Thanks!

PR #13746, opened 2022-11-19
[Nanoleaf] New Channel: State

austvik:
This is an addition that shows an image of the state of the panels with color.

Also makes the layout slightly prettier. This is less functional than the layout, and more eye-candy.

Idea by Stefan Hoehn from https://github.com/openhab/openhab-addons/pull/13552


austvik:
spaceinvader
squares
theduck
wings
lasvegas

bundles/org.openhab.binding.nanoleaf/src/main/java/org/openhab/binding/nanoleaf/internal/layout/shape/BarycentricTriangleGradient.java
```diff
@@ -0,0 +1,152 @@
...
+ *
+ * Inspired by
+ * https://github.com/hageldave/JPlotter/blob/9c92731f3b29a2cdb14f3dfdeeed6fffde37eee4/jplotter/src/main/java/hageldave/jplotter/util/BarycentricGradientPaint.java,
+ * for how to integrate it into Java AWT but kept so simple that I could understand it. It was however far too big to
+ * use as a dependency.
+ *
+ * @author Jørgen Austvik - Initial contribution
+ */
+@NonNullByDefault
+public class BarycentricTriangleGradient implements Paint {
```
stefan-hoehn, line 46: I wouldn't even have known what a barycentric gradient is ;-)

bundles/org.openhab.binding.nanoleaf/src/main/java/org/openhab/binding/nanoleaf/internal/layout/shape/BarycentricTriangleGradient.java
```diff
@@ -0,0 +1,152 @@
...
+import org.eclipse.jdt.annotation.Nullable;
+import org.openhab.binding.nanoleaf.internal.layout.ImagePoint2D;
+
+/**
+ * Paint for triangles with one color in each corner. Used to make gradients between the colors when
+ * dividing a hexagon into 6 triangles.
+ *
+ * https://codeplea.com/triangular-interpolation is instructive for the math.
+ *
+ * Inspired by
```
stefan-hoehn, line 38: fine to use it - I checked the licence -> it is MIT

bundles/org.openhab.binding.nanoleaf/src/main/java/org/openhab/binding/nanoleaf/internal/layout/shape/Square.java
```diff
@@ -27,7 +28,7 @@
  * @author Jørgen Austvik - Initial contribution
  */
 @NonNullByDefault
-public class Square extends Shape {
+public class Square extends SingleLightShape {
```
stefan-hoehn, line 31: Do you mind adding comment why it is called a "Single"-LightShape instead of Shape?

stefan-hoehn (changes requested):
That review was a huge one - I hope the next ones are going to be smaller ;-)
...but the outcome is worth taking the time.
Mainly it looks good to me: there only minor comments from me. 

@jlaur after @austvik has done the small amendments I recommend you merge it (if you don't have the time to fully review it).

bundles/org.openhab.binding.nanoleaf/README.md
```diff
@@ -137,10 +137,12 @@ The controller bridge has the following channels:
 | colorTemperatureAbs | Number    | Color temperature (in Kelvin, 1200 to 6500) of all light panels                                           | No        |
 | colorMode           | String    | Color mode of the light panels                                                                            | Yes       |
 | effect              | String    | Selected effect of the light panels                                                                       | No        |
+| layout              | Image     | Shows the layout of your panels with IDs.                                                                 | Yes       |
```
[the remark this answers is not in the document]
austvik: The heading for this column is "Read Only"

I believe Yes here is correct. It is just an image of the current layout. The user can't paint on the image and expect the panels to move on the wall. But e.g. for on/off state I can both see what it currently is (read), and change the state of it to e.g. turn it off (write).

bundles/org.openhab.binding.nanoleaf/src/main/java/org/openhab/binding/nanoleaf/internal/layout/shape/BarycentricTriangleGradient.java
austvik: It took some research and made it an interesting challenge :)

bundles/org.openhab.binding.nanoleaf/src/main/java/org/openhab/binding/nanoleaf/internal/layout/shape/BarycentricTriangleGradient.java
austvik: First I tried to use the jar, but it had so many dependencies. I could have copied just the one file, but it was interesting, so I tried to write one myself. I used the MIT code for how to integrate it into AWT (wouldn't have thought about writing a new Paint), and used the article to implement the render method.

austvik: And actually, because the one here doesn't implement all that antialising, it looks slightly better because the other code produced some strange lines towards the center where the lines met.

bundles/org.openhab.binding.nanoleaf/src/main/java/org/openhab/binding/nanoleaf/internal/layout/shape/Square.java
austvik: (Thinking out loud) Background: We have these things we can put on the wall. They have different shapes, hex, triangle, square, line. Most of the old style things have a single light. The Hexagon is either blue or red. They are described by one point, and as long as we have the outline of these elements, we can draw them with one algorithm. So this "SingleLightShape" is the thing that lets me write that algorithm once.

Then we have Elements Hegagons which have a light in each corner - so 6 colors and 6 panel ids in one thing you put on the wall. Images of lines also makes me believe lines can have two colors. 

So I was thinking "Shape" as something you put on the wall, and then the old style triangle/square/hex was those that had only one color. But what if we use the product names instead?

If I rename Shape to Panel, that would be the plastic thing that is placed on the wall.
Then I rename SingleLightShape to just Shape as that is the product name of those things with just one color.

stefan-hoehn: I never questioned that the name makes sense, I rather didn't know what the idea behind the name was. Ok, now I get it. Thanks.

bundles/org.openhab.binding.nanoleaf/README.md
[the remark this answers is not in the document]
stefan-hoehn: Ah, sorry, yes, you are right - my bad. This is the OUTPUT channel - I should have looked at the type. It read "shows" which told "me" it was a switch with which I can choose. Maybe it is better say "provides" instead of "shows"?

bundles/org.openhab.binding.nanoleaf/README.md
```diff
@@ -137,10 +137,12 @@ The controller bridge has the following channels:
...
 | rhythmState         | Switch    | Connection state of the rhythm module                                                                     | Yes       |
 | rhythmActive        | Switch    | Activity state of the rhythm module                                                                       | Yes       |
 | rhythmMode          | Number    | Sound source for the rhythm module. 0=Microphone, 1=Aux cable                                             | No        |
-| swipe               | Trigger   | [Canvas / Shapes Only] Detects Swipes over the panel.LEFT, RIGHT, UP, DOWN events are supported.          | YES        |
+| state               | Image     | Shows the current state of your panels with colors.                                                       | Yes       |
```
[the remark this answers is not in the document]
austvik: Added picture to README

bundles/org.openhab.binding.nanoleaf/src/main/java/org/openhab/binding/nanoleaf/internal/layout/shape/Shape.java
```diff
@@ -21,59 +23,74 @@
 import org.openhab.binding.nanoleaf.internal.layout.PanelState;
 import org.openhab.binding.nanoleaf.internal.layout.Point2D;
 import org.openhab.binding.nanoleaf.internal.layout.ShapeType;
+import org.openhab.core.library.types.HSBType;
 
 /**
- * Shape that can be drawn.
+ * Draws shapes is panel with a single LED.
```
stefan-hoehn, line 29: is -> as

stefan-hoehn:
@austvik only a very small typo, other than that, it looks good to me.

@jlaur IMHO you can merge it if you like.

austvik:
Thanks, Stefan. Reworded the comment.

bundles/org.openhab.binding.nanoleaf/README.md
```diff
@@ -104,7 +104,15 @@ Compare the following output with the right picture at the beginning of the arti
                                     41451                                     
 
 ```
-         
+
+## State
+
+The state channel shows an image of the panels on the wall. You have to configure things for each panel to
```
jlaur, line 110: Please reformat the text so that there's once sentence per line. This will make it easier to diff/comment etc.

jlaur (changes requested):
Excellent, thanks! Only a minor comment about README formatting.

bundles/org.openhab.binding.nanoleaf/README.md
austvik: Thanks! Fixed.

bundles/org.openhab.binding.nanoleaf/README.md
```diff
@@ -104,7 +104,14 @@ Compare the following output with the right picture at the beginning of the arti
                                     41451                                     
 
 ```
-         
+
+## State
+
+The state channel shows an image of the panels on the wall. You have to configure things for each panel to get the correct color.
```
jlaur, line 110:
```suggestion
The state channel shows an image of the panels on the wall.
You have to configure things for each panel to get the correct color. 
```

jlaur (approved):
LGTM

PR #13750, opened 2022-11-20
[pid] Fix PID controller state recovery

lnagel:
This pull requests fixes an issue where PID controller state was not recovered due to invalid isinstance checks on values.

I've simplified the code that obtains the current value to use `item.getStateAs(DecimalType.class)` instead.

There is one remaining issue: how to recover the state after OpenHAB is restarted. This doesn't work reliably when using InfluxDB, since the `PIDControllerTriggerHandler` seems to be initialized before the persistence service performs the `restoreOnStartup` procedure.  I will address this in another pull request hopefully.

Signed-off-by: Lenno Nagel <lenno@nagel.ee>

### DIFF
--- a/bundles/org.openhab.automation.pidcontroller/src/main/java/org/openhab/automation/pidcontroller/internal/handler/PIDControllerTriggerHandler.java
+++ b/bundles/org.openhab.automation.pidcontroller/src/main/java/org/openhab/automation/pidcontroller/internal/handler/PIDControllerTriggerHandler.java
@@ -22,7 +22,6 @@ import java.util.concurrent.TimeUnit;
 
 import org.eclipse.jdt.annotation.NonNullByDefault;
 import org.eclipse.jdt.annotation.Nullable;
-import org.openhab.automation.pidcontroller.internal.PIDException;
 import org.openhab.core.automation.ModuleHandlerCallback;
 import org.openhab.core.automation.Trigger;
 import org.openhab.core.automation.handler.BaseTriggerModuleHandler;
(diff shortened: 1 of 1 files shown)


Hilbrand:
This does change the handling certain value. This removes support for `StringType` and allows `Double.NaN`/`Infinite` values to be passed on. Is that intended?

fwolter:
Just for curiosity, what was the exact problem with the current implementation?

lnagel:
Sorry, I intended this PR to be in draft status until I've got all the cases covered.

> This does change the handling certain value. This removes support for `StringType` and allows `Double.NaN`/`Infinite` values to be passed on. Is that intended?

I'll double-check this, thanks!

> Just for curiosity, what was the exact problem with the current implementation?

Well, my starting point is actually the larger task - to be able to restart openhab without losing controller state. While the rule is able to recover state during edits and pause/unpause from the UI, then it is not able to recover state during OpenHAB's initial startup.

The problem I have with this is that I'm using the PID controller to control underfloor heating with a very long response time and thus a very long integration period. I am not able to reboot openhab without losing temperature stability in my rooms. I want to fix this in the long run.

fwolter:
Ah, I thought you already managed to keep the Item states across reboots with the current implementation.

jlaur:
@lnagel - are you still looking into this? 🙂

lnagel:
> @lnagel - are you still looking into this? 🙂

Well, I have to admit, I couldn't figure this one out. I ended up scrapping this idea and wrote a script instead that runs after OH start up is complete. It stops PID controllers, restores integral components from last hours average and starts controllers again.

Should we pursue further? I'm out of ideas..

jlaur:
> I ended up scrapping this idea and wrote a script instead that runs after OH start up is complete.

I think in that case you could close the pull request. If you find new inspiration and want to continue the work, you can always come back and reopen it. 🙂

Issue #13447, opened 2022-09-26
[resol] New binding in OH3.3 does not work with device SKSR 1/2/3


paicl01:
## Expected Behavior

I have used the previous version of the RESOL binding 
(https://github.com/wborn/openhab-addons/releases/download/resol-20201215/org.openhab.binding.resol-3.0.0-SNAPSHOT.jar)
without issues.

After upgrade OH3.0 --> OH3.3 the RESOL binding  is now  officially included and I changed to the new binding. 
I expected that the new binding would work as the previous one. 

## Current Behavior
The KM1 is recognized by the new binding, and a new thing SKSR 1/2/3 is found in the inbox, which goes online without error. However, after connecting the existing or new  items to the channels of this new thing, they are not updated. 
I dont know how to proceed. 

## Steps to Reproduce (for Bugs)

1.
2.

## Your Environment

* Version used: (e.g., openHAB and add-on versions)
* Environment name and version (e.g. Chrome 76, Java 8, Node.js 12.9, ...):
* Operating System and version (desktop or mobile, Windows 10, Raspbian Buster, ...):

BillGOH:
KM1 working fine here with the binding since openHAB 3.2. It controls a Remacal device.

ramack:
Did you delete all things that were created with earlier versions and recreated them? - I have no idea why that should help, but maybe its worth a try.

I have OH3.3 running smoothly with the resol binding.

And for debugging it would be great if you could enable trace logging for the resol binding by adding
`     <logger name="org.openhab.binding.resol" level="INFO" />`
into log4j2.xml and check your logs for details or paste it here or [there](https://community.openhab.org/t/resol-vbus-channels-not-updated/129591).

paicl01:
I added your proposed line to the log entries, this now looks as follows:
`openhab> Log:get
Logger                                             │ Level

ROOT                                               │ WARN
javax.jmdns                                        │ ERROR
javax.mail                                         │ ERROR
openhab.event                                      │ WARN
openhab.event.AddonEvent                           │ ERROR
openhab.event.InboxUpdatedEvent                    │ ERROR
openhab.event.ItemAddedEvent                       │ ERROR
openhab.event.ItemChannelLinkAddedEvent            │ ERROR
openhab.event.ItemChannelLinkRemovedEvent          │ ERROR
openhab.event.ItemRemovedEvent                     │ ERROR
openhab.event.ItemStateEvent                       │ ERROR
[...]

PR #13753, opened 2022-11-20
[resol] Fix handling of WeekTime fields on vbus

ramack:
Controllers (for example the SKSR 1/2/3) on the VBUS transmitting fields of type WeekTime did not work as decribed in #13447 which is fixed by this PR.

The resol-vbus-java library used by this binding was also improved for this controller in the meantime, that's why this PR integrates an upgrade to the latest version of this lib.

A compiled version of this code change can be found [here](https://github.com/ramack/openhab-addons/blob/51dd67d764f11084cb904aeabf7a02cb52f4791e/bundles/org.openhab.binding.resol/target/org.openhab.binding.resol-3.4.0-SNAPSHOT.jar?raw=true) and @paicl01 tested it successfully.

### DIFF
--- a/bundles/org.openhab.binding.resol/src/main/java/org/openhab/binding/resol/handler/ResolThingHandler.java
+++ b/bundles/org.openhab.binding.resol/src/main/java/org/openhab/binding/resol/handler/ResolThingHandler.java
@@ -48,6 +48,7 @@ import de.resol.vbus.SpecificationFile;
 import de.resol.vbus.SpecificationFile.Enum;
 import de.resol.vbus.SpecificationFile.EnumVariant;
 import de.resol.vbus.SpecificationFile.Language;
+import tech.units.indriya.AbstractUnit;
 
 /**
  * The {@link ResolThingHandler} is responsible for handling commands, which are
@@ -254,7 +255,7 @@ public class ResolThingHandler extends ResolBaseThingHandler {
                                         this.updateState(channelId, q);
                                     } catch (IllegalArgumentException e) {
                                         logger.debug("unit of '{}' unknown in openHAB", str);
-                                        QuantityType<?> q = new QuantityType<>(dd.toString(), Locale.getDefault());
+                                        QuantityType<?> q = new QuantityType<>(dd, AbstractUnit.ONE);
                                         this.updateState(channelId, q);
                                     }
                                 }
--- a/bundles/org.openhab.binding.resol/src/main/java/org/openhab/binding/resol/handler/ResolThingHandler.java
+++ b/bundles/org.openhab.binding.resol/src/main/java/org/openhab/binding/resol/handler/ResolThingHandler.java
@@ -67,7 +67,7 @@ public class ResolThingHandler extends ResolBaseThingHandler {
 
     private static final SimpleDateFormat TIME_FORMAT = new SimpleDateFormat("HH:mm");
 
-    private static final SimpleDateFormat WEEK_FORMAT = new SimpleDateFormat("E, HH:mm");
+    private static final SimpleDateFormat WEEK_FORMAT = new SimpleDateFormat("EEE,HH:mm");
 
     static {
         synchronized (DATE_FORMAT) {
@@ -158,13 +158,13 @@ public class ResolThingHandler extends ResolBaseThingHandler {
 
             Thing thing = getThing();
             switch (pfv.getPacketFieldSpec().getType()) {
-                case WeekTime:
                 case DateTime:
                     acceptedItemType = "DateTime";
                     break;
                 case Number:
                     acceptedItemType = ResolChannelTypeProvider.itemTypeForUnit(pfv.getPacketFieldSpec().getUnit());
                     break;
+                case WeekTime:
                 case Time:
                 default:
                     acceptedItemType = "String";
@@ -249,11 +249,12 @@ public class ResolThingHandler extends ResolBaseThingHandler {
                                     this.updateState(channelId, q);
                                 } else {
                                     try {
-                                        QuantityType<?> q = new QuantityType<>(str);
+                                        QuantityType<?> q = new QuantityType<>(str, Locale
+                                                .getDefault()); /* vbus library returns the value in default locale */
                                         this.updateState(channelId, q);
                                     } catch (IllegalArgumentException e) {
                                         logger.debug("unit of '{}' unknown in openHAB", str);
-                                        QuantityType<?> q = new QuantityType<>(dd.toString());
+                                        QuantityType<?> q = new QuantityType<>(dd.toString(), Locale.getDefault());
                                         this.updateState(channelId, q);
                                     }
                                 }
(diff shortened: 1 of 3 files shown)


lolodomo:
This looks like a breaking change to be mentioned in the release notes.

ramack:
In a sense I agree, but as it breaks only for devices using the WeekTime type before it doesn't matter, as those did not work at all :smile:

lolodomo:
But users will have to change their item type to make it work. So it is better to inform them.

ramack:
yea why not. Where do I need to put it to make it occur in the release notes? The Readme in the binding doesn't seem to be the best place - on https://www.openhab.org/docs/developer/bindings/ I did not find anything and I am simply not contributing often enough to know it...

bundles/org.openhab.binding.resol/src/main/java/org/openhab/binding/resol/handler/ResolThingHandler.java
```diff
@@ -249,11 +249,12 @@ protected void packetReceived(Specification spec, Language lang, Packet packet)
...
                                 } else {
                                     try {
-                                        QuantityType<?> q = new QuantityType<>(str);
+                                        QuantityType<?> q = new QuantityType<>(str, Locale
+                                                .getDefault()); /* vbus library returns the value in default locale */
                                         this.updateState(channelId, q);
                                     } catch (IllegalArgumentException e) {
                                         logger.debug("unit of '{}' unknown in openHAB", str);
-                                        QuantityType<?> q = new QuantityType<>(dd.toString());
+                                        QuantityType<?> q = new QuantityType<>(dd.toString(), Locale.getDefault());
```
lolodomo, line 257: `dd` is a `Double`. I don't think you need to pass the locale there.

You are creating a `QuantityType` without any unit ?? That looks strange to me.

bundles/org.openhab.binding.resol/src/main/java/org/openhab/binding/resol/handler/ResolThingHandler.java
```diff
@@ -287,13 +287,15 @@ protected void packetReceived(Specification spec, Language lang, Packet packet)
...
                             ResolBridgeHandler handler = (ResolBridgeHandler) b.getHandler();
                             String value;
+                            Locale loc;
                             if (handler != null) {
-                                value = pfv.formatTextValue(pfv.getPacketFieldSpec().getUnit(), handler.getLocale());
+                                loc = handler.getLocale();
                             } else {
-                                value = pfv.formatTextValue(pfv.getPacketFieldSpec().getUnit(), Locale.getDefault());
+                                loc = Locale.getDefault();
                             }
```
lolodomo, line 295: Your comment above: "vbus library returns the value in default locale".
So why not just using there like above `Locale.getDefault()` ?

lolodomo (changes requested):

lolodomo:
> yea why not. Where do I need to put it to make it occur in the release notes? The Readme in the binding doesn't seem to be the best place - on https://www.openhab.org/docs/developer/bindings/ I did not find anything and I am simply not contributing often enough to know it...

No problem, I will guide you once the PR is merged. This will be added in a second step, outside the current PR.

bundles/org.openhab.binding.resol/src/main/java/org/openhab/binding/resol/handler/ResolThingHandler.java
lolodomo: Ok, you are passing a string without any unit and a locale. I don't know what will be the unit considered by the core framework in that case.

lolodomo: I would have imagined to rather use in that case the constructor with a Number and a unit.

https://github.com/openhab/openhab-core/blob/4f35d474fc63f487b10042d05f01ea524fe312e8/bundles/org.openhab.core/src/main/java/org/openhab/core/library/types/QuantityType.java#L164

ramack: I would prefer that yes, but this code is only executed if the resol-vbus library returns a unit that is not known by openHAB. So I cannot convert it to a reasonable unit. I actually don't remember whether it commonly happens or whether it is only for a case that could theoretically. In this PR I only added the locale to cover the localized decimal point correctly - which was a change in the lib between 0.7 and 0.10.

ramack: ah no, dd is the Double, so after the toString the QuantityType should not need special handling of the decimal point. You're totally right, adding the locale here is useless and even wrong if the default local uses , as decimal separator.

Would 
`QuantityType<?> q = new QuantityType<>(dd, AbstractUnit.ONE);`
be better?

bundles/org.openhab.binding.resol/src/main/java/org/openhab/binding/resol/handler/ResolThingHandler.java
lolodomo: By the way, you changed nothing there, so let's close my comment.

bundles/org.openhab.binding.resol/src/main/java/org/openhab/binding/resol/handler/ResolThingHandler.java
```diff
@@ -249,11 +250,12 @@ protected void packetReceived(Specification spec, Language lang, Packet packet)
...
                                 } else {
                                     try {
-                                        QuantityType<?> q = new QuantityType<>(str);
+                                        QuantityType<?> q = new QuantityType<>(str, Locale
+                                                .getDefault()); /* vbus library returns the value in default locale */
                                         this.updateState(channelId, q);
                                     } catch (IllegalArgumentException e) {
                                         logger.debug("unit of '{}' unknown in openHAB", str);
-                                        QuantityType<?> q = new QuantityType<>(dd.toString());
+                                        QuantityType<?> q = new QuantityType<>(dd, AbstractUnit.ONE);
```
lolodomo, line 258: Please rather use `Units.ONE`, importing Units from org.openhab.core.library.unit.Units

lolodomo (changes requested):

lolodomo (approved):
LGTM, thank you.

lolodomo:
@ramack : please create a PR to explain the breaking change.
Here is an example of what to do (only one line to add in the same file):
https://github.com/openhab/openhab-distro/pull/1420

ramack:
> @ramack : please create a PR to explain the breaking change. Here is an example of what to do (only one line to add in the same file): [openhab/openhab-distro#1420](https://github.com/openhab/openhab-distro/pull/1420)
like this?
https://github.com/openhab/openhab-distro/pull/1427

Issue #11691, opened 2021-12-02
[telegram] Can't send message with underscore character


zolakk:
I noticed when trying to set up a rule to send telegram messages when my zwave devices drop offline, I was getting the error "Failed to send telegram message: Bad Request: can't parse entities: Can't find end of the entity starting at byte offset 182". I narrowed it down by counting the characters and then attempting to send a message with just an _ in it, which resulted in "Failed to send telegram message: Bad Request: can't parse entities: Can't find end of the entity starting at byte offset 0"

## Expected Behavior
The message should send without issue

## Current Behavior
Attempting to send a message with a _ (underscore) character gives the message  "Failed to send telegram message: Bad Request: can't parse entities: Can't find end of the entity starting at byte offset <character position of the _ character>". Appears to happen with the sendTelegram, sendTelegramQuery, and sendTelegramQuery actions

## Possible Solution
Internally escape the character before sending if it's an invalid character on the telegram API

## Steps to Reproduce (for Bugs)
This python script:
```
telegramAction = actions.get("telegram","<botThingID>")
telegramAction.sendTelegram("_")
telegramAction.sendTelegramQuery("_")
```
will produce these lines in the log:
```
2021-12-02 11:14:26.246 [WARN ] [gram.internal.action.TelegramActions] - Failed to send telegram message: Bad Request: can't parse entities: Can't find end of the entity starting at byte offset 0

2021-12-02 11:14:26.407 [WARN ] [gram.internal.action.TelegramActions] - Failed to send telegram message: Bad Request: can't parse entities: Can't find end of the entity starting at byte offset 0
```

## Context
I first noticed while trying to send a log snippet to my telegram with the text "COMMUNICATION_ERROR" in it but was failing with the above errors. I am able to work around that by replacing the underscore with a space but it would be good to not have to do that.

## Your Environment
* Version used: openHAB 3.2.0.M4
* Operating System and version (desktop or mobile, Windows 10, Raspbian Buster, ...): Docker on raspbian buster

lolodomo:
Closed by #13758

PR #13758, opened 2022-11-21
[Telegram] Can't send message with underscore character

clinique:
Escapes the "_" in the message. Solves issue #11691

### DIFF
--- a/bundles/org.openhab.binding.telegram/src/main/java/org/openhab/binding/telegram/internal/action/TelegramActions.java
+++ b/bundles/org.openhab.binding.telegram/src/main/java/org/openhab/binding/telegram/internal/action/TelegramActions.java
@@ -237,7 +237,8 @@ public class TelegramActions implements ThingActions {
         }
         TelegramHandler localHandler = handler;
         if (localHandler != null) {
-            SendMessage sendMessage = new SendMessage(chatId, message);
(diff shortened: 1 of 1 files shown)


lolodomo (approved):
LGTM

lolodomo:
@clinique : was it tested ?

clinique:
> @clinique : was it tested ?

Yes. I replicated the example presented in the issue.

mtzro2003:
Hi guys!

Implementing this breaks the formatting of the messages using `markdown` as the underscore means italic...
Of course, I could switch to using `HTML` in `parseMode`, but then `markdown` is not fully functional, so maybe it could be entirely removed from thing configuration.
Up until now, I escaped underscore (and other markdown characters) when sending telegrams with items names with a simple 
```
result.replace("_", "\\_").replace("*", "\\*").replace("[", "\\[").replace("`", "\\`")
```

lolodomo:
A new issue would have been better. Without, there is a good risk to "loose" your message.
Pinging @clinique.

RubenKona:
New implementation on 3.4 break original messages with "_". Before (3.3 and earlier, Windows 10) works perfectly with messages with underscores. Now always add extra characters, so, for example, URLs or user names and passwords on messages doesn't works anymore.

PR #13764, opened 2022-11-22
[jsscripting] Fix add-on name in feature/pom

florian-h05:
Working with the UI, I noticed that the JS Scripting addon there is actually called _JSScripting_ (without the space).
This PR should correct this.

### DIFF
--- a/bundles/org.openhab.automation.jsscripting/pom.xml
+++ b/bundles/org.openhab.automation.jsscripting/pom.xml
@@ -12,7 +12,7 @@
 
   <artifactId>org.openhab.automation.jsscripting</artifactId>
 
-  <name>openHAB Add-ons :: Bundles :: Automation :: JSScripting</name>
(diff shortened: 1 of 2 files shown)


lolodomo:
JS Scripting or JavaScript Scripting like defined in README title?

florian-h05:
I’d prefer JS Scripting, because it seems like JavaScript Scripting is too long for the add-on store of MainUI.

lolodomo:
Do you want to adjust README in consequence or not?

florian-h05:
I‘ve just checked using the Chrome dev tools with device simulation, JavaScript Scripting seems to work. I‘ll update this to use JavaScript Scripting as the name.

The thing I don‘t like with JS Scripting, is that is might not be clear for all which language is meant.

lolodomo (approved):

PR #13776, opened 2022-11-25
[knx] Remove obsolete public modifier in tests

Riggs333:
Since JUnit Jupiter test classes and methods do not need to be `public`.

It is recommended to omit the `public` modifier.
See https://junit.org/junit5/docs/current/user-guide/#writing-tests-classes-and-methods

> "Class and method visibility
> Test classes, test methods, and lifecycle methods are not required to be `public`, but they must not be `private`.
> 
> It is generally recommended to omit the `public` modifier for test classes, test methods, and lifecycle methods unless there is a technical reason for doing so (...)"

Signed-off-by: Thomas Traude <t-riggs@gmx.net>


jlaur (approved):
LGTM - thanks.

PR #13795, opened 2022-11-28
[evcc] Extend accepted types for commands & Handle quantity types of commands

florian-h05:
Should fix an issue on the forum, where a user sent a command of DecimalType, but the binding expected a QuantityType.
See https://community.openhab.org/t/evcc-binding-electric-vehicle-charging-control/135209/15.

## Description

- The accepted command types are extended for all channels if possible.
- If a command's type is not accepted, an Exception is now thrown.
- For QuantityType commands the command is converted to the required unit.
- Allows the evcc URL to end with a slash (in the past, this lead to failing commanding API requests).

## Testing

- I haven't been able to test, whether DecimalType commands work now, but they should (the code is really simple).
- But I was able to verify that the unit conversion is working correctly.


bundles/org.openhab.binding.evcc/src/main/java/org/openhab/binding/evcc/internal/EvccHandler.java
```diff
@@ -95,16 +95,27 @@ public void handleCommand(ChannelUID channelUID, Command command) {
                     case CHANNEL_LOADPOINT_MODE:
                         if (command instanceof StringType) {
                             evccAPI.setMode(loadpoint, command.toString());
+                        } else {
+                            throw new EvccException("Command has wrong type, StringType required!");
```
jlaur, line 99: Out of curiosity: Why throw exception here, which is caught locally within the same method, instead of controlling the flow by logging and returning? Throwing and handling exceptions is expensive.

bundles/org.openhab.binding.evcc/src/main/java/org/openhab/binding/evcc/internal/EvccHandler.java
```diff
@@ -129,27 +142,41 @@ public void handleCommand(ChannelUID channelUID, Command command) {
...
+                            evccAPI.setMaxCurrent(loadpoint, ((DecimalType) command).intValue());
+                        } else {
+                            throw new EvccException("Command has wrong type, QuantityType or DecimalType required!");
                         }
                         break;
                     default:
                         return;
                 }
-            } catch (EvccApiException e) {
+            } catch (EvccException e) {
```
jlaur, line 179: Should `refresh()` below be called when command is invalid?

jlaur (changes requested):
Thanks for the fix. Just two minor clarifying questions.

bundles/org.openhab.binding.evcc/src/main/java/org/openhab/binding/evcc/internal/EvccHandler.java
florian-h05: I first thought of logging, but then I thought, that exceptions are handled anyway, so let‘s throw an exception instead.
As you obviously prefer logging, I‘ll change that to logging,

bundles/org.openhab.binding.evcc/src/main/java/org/openhab/binding/evcc/internal/EvccHandler.java
florian-h05: Yes, because otherwise the Item would have a wrong state.
I don't know why the Item is auto updating on command, because I set the autoUpdatePolicy for the read-write channels to veto, but it does.

bundles/org.openhab.binding.evcc/src/main/java/org/openhab/binding/evcc/internal/EvccHandler.java
florian-h05: I've update the PR.

florian-h05:
@jlaur 
I addressed your review, please re-review.

jlaur (approved):
LGTM

PR #13796, opened 2022-11-28
[homekit] make sure to convert step values to Celsius

ccutrer:
otherwise if your step is 1.0 in fahrenheit, then your values will get rounded to 1.0 celsius, and you might not even notice you've lost precision in the Home app.

### DIFF
--- a/bundles/org.openhab.io.homekit/src/main/java/org/openhab/io/homekit/internal/HomekitTaggedItem.java
+++ b/bundles/org.openhab.io.homekit/src/main/java/org/openhab/io/homekit/internal/HomekitTaggedItem.java
@@ -18,6 +18,8 @@ import java.math.BigDecimal;
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
 
+import javax.measure.Unit;
+
 import org.eclipse.jdt.annotation.NonNullByDefault;
 import org.eclipse.jdt.annotation.Nullable;
 import org.openhab.core.items.GroupItem;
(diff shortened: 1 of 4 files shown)


bundles/org.openhab.io.homekit/src/main/java/org/openhab/io/homekit/internal/HomekitTaggedItem.java
```diff
@@ -415,14 +417,34 @@ public double getConfigurationAsDouble(String key, double defaultValue) {
...
+
+        if (relativeConversion) {
+            convertedValue = parsedValue.toUnitRelative(defaultValue.getUnit());
+        } else {
+            convertedValue = parsedValue.toInvertibleUnit(defaultValue.getUnit());
+        }
+        // not convertible? just assume it's in the item's unit
+        if (convertedValue == null) {
+            Unit unit;
+            if (getBaseItem() instanceof NumberItem && (unit = ((NumberItem) getBaseItem()).getUnit()) != null) {
```
yfre, line 437: i have not seen the assignment (=) in IF statement that often in Java. but remember it from old good C/C++ days.
but if it works, why not. probably saves another IF.

yfre (approved):
right, good point. steps require conversion as well. 
thanks for fixing it. also good to see getConfigurationAsQuantity used at different places and saves some work

LGTM

jlaur (approved):
LGTM - thank you!

PR #13801, opened 2022-11-29
[mqtt.espmilighthub] Use system channel types for color temp and brightness

### LINKED ISSUE #10096: [mqtt.espmilighthub] expose color temperature as Number in mireds
Right now color temperature is exposed as a Dimmer from 0-100. However, HomeKit integration requires the color temperature in mireds (from 50 to 400; aka 2500K to 20000K in kelvin). Conveniently, the underlying
MQTT data already exposes it in mireds (from 153 to 370; aka 2700K to 6500K in kelvin). It'd be great if the binding just exposed this directly, so I can hook it up to HomeKit and be done, instead of using rules.

As a wish list item, it'd be nice if there was also a channel that exposed the temperature as a Number in Kelvin (1,000,000 / mireds), since that makes more sense for say Basic UI. I'm currently using a transform service on the items state description, but that means the slider still shows 0-100 when you're adjusting it.

ccutrer:
closes #10096

also add absolute color temp channel. and remove level channel when we already have a color channel (you can link a DimmerItem to a color channel if your light doesn't happen to have color or you want direct control of brightness only for some reason)

Note that this _is_ a breaking change (brightness channel is no longer available on RGB or RGB+CCT bulbs in favor of linking Dimmer items to the color channel). My understanding is that this PR goes as is, then a separate PR is made to the -distro repo calling out the breaking changes? Do I need to handle anything of automatically pruning the extra channel from things that were created before this code existed?


Skinah:
Thank you for the contribution it will be great to have this feature. I think it would be best to wait until AFTER the STABLE build is released in 2-3 weeks before merging any breaking changes so they can be trialed in Milestone builds first. So I will make just a few quick comments now.

Yes a separate PR is made to the -distro repo calling out the breaking changes and you mark it with a label AWAITIN|G OTHER PR and give a brief description that it is waiting on this PR to be merged first.

"Do I need to handle anything of automatically pruning the extra channel from things that were created before this code existed?"
No but this is what you need to mention in the alert for the breaking change. In the file there will be other bindings that alert that you need to delete and re-create the THINGS so copy their text if you wish. For people that do texual config there is nothing that needs to be done, but for GUI users the thing must have a summery of the channels that it held back when the thing was created, so it is necessary to delete and re-add or you can get some very strange behavior that people will blame on the binding having a bug.

As for the breaking changes, do we really need to remove the level channel? What do you think of making it marked as ADVANCED so it is not presented to users and leaving it in place? There is also another way of marking a channel as hidden from memory (not well documented that this is there so I would have to hunt it down and confirm its not my bad memory) so it still works but cannot be seen even if the advanced box is ticked. Two alternative ways forward which I would lean towards the `advanced` method.

I do believe I liked the separate level channel as I also went to remove it once in the past. The reason was based on the color channel only sends a RGB/HSB command and this was causing the code to drop back to using the RGB leds when you wanted to stay on the WHITE leds. Your recent PR that is merged does solve this.

Skinah:
If removing the level channel depending on result of discussion of the above, the readme will need to be updated most likely it will be listed and described in there. 

Is there any other breaking changes? Does the slider change the colour temp the same way/direction, so rules and controls are not affected by the changes?

jlaur:
@Skinah, didn't read the code yet, but from your comments, this might be relevant (or not): #11668

Skinah:
Thanks for posting @jlaur was good to read that for my knowledge. I am hoping that the channel is not removed, as I suspect that it is still needed but I am open to be wrong as I often am. It has been a long time since I looked into this binding in depth and done tests. Also, I took a quick look and it appears to only be for a THING that you can prevent it from being listed in the UI, not sure if it is supported for channels to use `listed="false"`.
https://www.openhab.org/docs/developer/bindings/thing-xml.html#things

ccutrer:
> I do believe I like/d the separate level channel, as I also went to remove it once in the past and then had to put it back in. The reason was based on the color channel only sends a RGB/HSB command

No, it is completely unnecessary to have a dedicated level channel. To begin with, it is perfectly permissible to send a PercentType command to an HSBType item/channel. And in fact, the handler for this binding already treats them identically:

```java
            case CHANNEL_COLOUR:
            case CHANNEL_LEVEL:
                handleLevelColour(command);
                break;
```

It's also perfectly permissible to link a DimmerItem to an HSBType channel (analogous to linking a SwitchItem to a PercentType channel. Or even an HSBType channel). Also note that at least in BasicUI, the "up" and "down" buttons next to the color-picker already send PercentType.HUNDRED and PercentType.ZERO commands. If you want to explicitly treat it as the level, you would either link a DimmerItem to the color channel, or in your sitemap explicitly select a Slider control rather than the default or an explicit Colorpicker. Even if you leave it as a Colorpicker, the slider for brightness when drilling into the color will still send only PercentType commands, not full HSBType commands. Allowing you to "stay on white" if you're just wanting to dim the current white color temperature.

Skinah:
I have read the changes and am happy to have them merged as is, just please make changes to the readme to reflect the changes to the channels, and also we need to put the `Alert` in place for the breaking change/s as already mentioned. Lastly it may be necessary to re-run the mvn command to generate what is needed for doing translations, not sure if it is needed when they are system channels or the removed channels mean it is needed.

https://www.openhab.org/docs/developer/utils/i18n.html#generating-i18n-properties-file

Regarding your last post, I agree and know that the color channel can be used for PercentType and slider controls. I do not know where/why I had it in my head it was needed, but I can see no reason to keep it by reading the code there would be no difference.

lolodomo (dismissed):
LGTM, thank you

lolodomo:
Code LGTM but as @Skinah said, it remains to adjust README and i18n default translations.

ccutrer:
sorry it took me so long to get back to this. it was very low priority for me, and I had lots of other things I wanted to fix first. I updated the readme, and removed the now-unused translations. The breaking-change alert is pending in distro (https://github.com/openhab/openhab-distro/pull/1463). I rebased to resolve a trivial merge conflict in the README; no code changes through any of this.

ccutrer:
I just did a simple rebase with no merge conflicts to resolve CI failing on the copyright year line

bundles/org.openhab.binding.mqtt.espmilighthub/src/main/resources/OH-INF/update/instructions.xml
```diff
@@ -0,0 +1,83 @@
...
+	xmlns:update="https://openhab.org/schemas/update-description/v1.0.0"
+	xsi:schemaLocation="https://openhab.org/schemas/update-description/v1.0.0 https://openhab.org/schemas/update-description-1.0.0.xsd">
+
+	<thing-type uid="mqtt:rgb_cct">
+		<instruction-set targetVersion="1">
+			<remove-channel id="level"/>
+			<update-channel id="colorTemperature">
+				<type>system.color-temperature</type>
+			</update-channel>
+			<update-channel id="colourTemperatureAbs">
```
jlaur, line 12: This is a new channel, right? Should be `<add-channel>`.

Same below.

jlaur (changes requested):
Thanks for the update. I have added a few new comments.

bundles/org.openhab.binding.mqtt.espmilighthub/src/main/resources/OH-INF/update/instructions.xml
```diff
@@ -0,0 +1,83 @@
...
+	</thing-type>
+
+	<thing-type uid="mqtt:rgb">
+		<instruction-set targetVersion="1">
+			<remove-channel id="level"/>
+			<update-channel id="colour">
+				<type>system.color</type>
+			</update-channel>
+		</instruction-set>
+	</instruction-set>
```
jlaur, line 80:
```suggestion
```

bundles/org.openhab.binding.mqtt.espmilighthub/src/main/resources/OH-INF/update/instructions.xml
```diff
@@ -0,0 +1,83 @@
+<?xml version="1.0" encoding="UTF-8" standalone="yes" ?>
+<update:update-descriptions xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+	xmlns:update="https://openhab.org/schemas/update-description/v1.0.0"
+	xsi:schemaLocation="https://openhab.org/schemas/update-description/v1.0.0 https://openhab.org/schemas/update-description-1.0.0.xsd">
+
+	<thing-type uid="mqtt:rgb_cct">
+		<instruction-set targetVersion="1">
+			<remove-channel id="level"/>
+			<update-channel id="colorTemperature">
+				<type>system.color-temperature</type>
```
jlaur, line 10:
```suggestion
				<type>system:color-temperature</type>
```

Please check others. It might be worth verifying the file against the XSD after fixing these. I was checking because I was unsure if `update-channel`/`add-channel` can be freely mixed or had to be in specific order (add/update/remove) and found these issues.

I'm wondering if the upgrade actually worked despite these issues?

jlaur (changes requested):
Sorry, found some more issues.

lolodomo:
@ccutrer : it looks like there are not a lot of things to do to finish this PR. Could you take a look to @jlaur last comments ?

lolodomo:
@ccutrer : maybe it's a shame not to finish this PR ? You have just two comments from @jlaur to consider.

ccutrer:
It's not super important, so I'm okay with it slipping past 4.1. I was planning getting to it during the holiday break.

bundles/org.openhab.binding.mqtt.espmilighthub/src/main/resources/OH-INF/update/instructions.xml
ccutrer: I corrected the XML, the type references, ran the XML through a schema validator, and also recreated my things with the current main, then added with this PR, and confirmed my channels properly changed type on existing things.

ccutrer:
I had a few minutes this morning, so this is ready now.

lolodomo (approved):
LGTM, thank you

Issue #13709, opened 2022-11-13
[systeminfo] memory#availableHeap reports wrong values


UrsusS:
OH 3.4.0.M4
After having used the UI configuration for the systeminfo thing, the reported value for available heap is much to low.
Before with M3 (and textual config), the values have been around 50 - 120 MB, now it is ~ 2.1E-16 MB.

This happens on my both machines where I've upgraded and changed the kind of configuration.

![grafik](https://user-images.githubusercontent.com/37875289/201518214-052c4c20-3d1f-419a-a0f8-88e497f4b5e4.png)

mherwege:
@UrsusS Did you try removing the state description? The value should show in Bytes by default. I tested on a RPi 4 with Java OpenJDK 11.0.16 and it works correctly for me.
What operating system and linux version did you test with?
The code retrieving this was not touched in the most recent PR's to this binding.

UrsusS:
The item was defined as:
`Number:DataAmount       num_Server_Heap_Available                 "Verfügbarer Heap Speicher [%d MB]"         <none>          (gSystemInfo, gRestoreOnStartup, gLog, gDel1M)    {channel="systeminfo:computer:kopf:memory#availableHeap"}`
changing it to 
`Number:DataAmount       num_Server_Heap_Available                 ""         <none>          (gSystemInfo, gRestoreOnStartup, gLog, gDel1M)    {channel="systeminfo:computer:kopf:memory#availableHeap"}`
gives a very high number in Byte.

  version: 3.4.0.M4
  javaVersion: 11.0.16
  javaVendor: Raspbian
  osName: Linux
  osVersion: 5.15.76-v7l+

mherwege:
Does the value in Bytes correspond to the available heap? I know it is a high value. This code has not changed. The only thing I can see is that there have been quite a few changes to units in core.
Did you try %.0f MB as a state description?
@J-N-K Is there a chance that conversion from Bytes to MBytes using a state description broke in one way or the other with M4?

UrsusS:
Hard to say if these values correspond to the available heap. The value isn't constant, it is highly fluctuating.
%.0f MB didn't solve it, I just get 0 MB

mherwege:
[...]

Issue #3183, opened 2018-01-29
[homematic] homegear integration partially broken by #3140


pauxus:
Unfortunately, PR #3140 introduced a nasty bug:

When using homegear with additional modules like philipshue or Sonos (which is useful, because these modules support some features that OH does not), the device types in homegear usually contain spaces (Like : `Sonos PLAY:1`)

This scenario was previously handled by the ListDevicesParser by calling `MiscUtils.validateCharacters`, which replaced the spaces and colons with dashes (Sonos-PLAY-1), emitting a warning in the logs.

This validation (accidentially) got dropped in that PR, leading to an invalid HmDevice, which breaks the device loading loop completely:

```
java.lang.IllegalArgumentException: UID segment 'HG-Philips LCT001' contains invalid characters. Each segment of the UID must match the pattern [A-Za-z0-9_-]*.
        at org.eclipse.smarthome.core.thing.UID.validateSegment(UID.java:95) [116:org.eclipse.smarthome.core.thing:0.10.0.201801231340]
        at org.eclipse.smarthome.core.thing.UID.<init>(UID.java:73) [116:org.eclipse.smarthome.core.thing:0.10.0.201801231340]
        at org.eclipse.smarthome.core.thing.ThingTypeUID.<init>(ThingTypeUID.java:39) [116:org.eclipse.smarthome.core.thing:0.10.0.201801231340]
        at org.openhab.binding.homematic.internal.type.UidUtils.generateThingTypeUID(UidUtils.java:41) [248:org.openhab.binding.homematic:2.3.0.201801282104]
        at org.openhab.binding.homematic.internal.type.HomematicTypeGeneratorImpl.generate(HomematicTypeGeneratorImpl.java:126) [248:org.openhab.binding.homematic:2.3.0.201801282104]
        at org.openhab.binding.homematic.handler.HomematicBridgeHandler.onDeviceLoaded(HomematicBridgeHandler.java:273) [248:org.openhab.binding.homematic:2.3.0.201801282104]
        at org.openhab.binding.homematic.internal.communicator.AbstractHomematicGateway.loadAllDeviceMetadata(AbstractHomematicGateway.java:387) [248:org.openhab.binding.homematic:2.3.0.201801282104]
        at org.openhab.binding.homematic.internal.discovery.HomematicDeviceDiscoveryService$1.run(HomematicDeviceDiscoveryService.java:103) [248:org.openhab.binding.homematic:2.3.0.201801282104]
[...]

PR #3207, opened 2018-02-04
[Kodi] Added dynamic state descriptions for opening pvr stream channels

cweitkamp:
- [x] Added dynamic state descriptions for opening pvr stream channels
- [x] Added `REQUEST_CACHE`
- [x] Added model for Kodi specific entities (e.g. `KodiPVRChannelGroup`, `KodiPVRChannel`, `KodiBaseItem`, etc.)
- [x] Adapted README.md

Also-by: Gregory Moyer <moyerg@syphr.com>
Signed-off-by: Christoph Weitkamp <github@christophweitkamp.de>

PR #13804, opened 2022-11-29
[systeminfo] QuantityTypes and state descriptions cleanup

mherwege:

Changes in core UOM handling make the system uptime channel interpret the minutes as seconds when linking an Number:Time item to the channel. This is because the state description has a wrong UOM. See core issue [#3183](https://github.com/openhab/openhab-core/issues/3183).
This PR does a cleanup of descriptions and default state descriptions to avoid this issue.

Closes  [#3183](https://github.com/openhab/openhab-core/issues/3183).

Another issue has been discovered [#13709](https://github.com/openhab/openhab-addons/issues/13709) with unit conversions in relation to this binding. I don't see an immediate reason for this in the binding. If you have ideas, please suggest and I will include in this PR.


bundles/org.openhab.binding.systeminfo/src/main/resources/OH-INF/thing/channels.xml
```diff
@@ -159,271 +159,271 @@
...
+		<description>The full path.</description>
 		<state readOnly="true" pattern="%s"/>
 		<config-description-ref uri="channel-type:systeminfo:lowpriority_process"/>
 	</channel-type>
 
 	<channel-type id="available">
 		<item-type>Number</item-type>
 		<label>Available</label>
-		<description>Available size in MB</description>
+		<description>Available size.</description>
```
lolodomo, line 178: While your channel is expected to be linked to a Number item without unit, it is better to mention the unit in the description. 

Same remark for all other similar channels. 

Please also adjust the default translations file.

lolodomo (changes requested):

bundles/org.openhab.binding.systeminfo/src/main/resources/OH-INF/thing/channels.xml
mherwege: I will adjust the default translations.

However, what about linking the plain number channel to a Number:DataAmount item. I understood it should then convert from the provided MB, as this comes with the default state description. That’s why I removed the units from the description.
It would probably even be better to convert all of these channels to QuantityType channels. Attaching an plain Number type to them would still give the same result as today.

lolodomo: > However, what about linking the plain number channel to a Number:DataAmount item. I understood it should then convert from the provided MB, as this comes with the default state description. That’s why I removed the units from the description.

No, if you defined a plain number channel, there will be no unit conversion. If "MB" is in the state pattern, it will be just added when formatting the state.
If you want UOM with conversion, you need to define a channel with a number with dimension.

> It would probably even be better to convert all of these channels to QuantityType channels. Attaching an plain Number type to them would still give the same result as today.

Yes, it is of course better to use UoM to get unit conversion feature. I have just a doubt if it would be a breaking change or not for users who defined things with UI.

lolodomo: if you defined a plain number channel, we can expect the binding is building a DecimalType state and so no unit involved.

mherwege:
Test fail because DataAmount unit conversions do not work properly, see [#3207](https://github.com/openhab/openhab-core/issues/3207).
The code is complete from my perspective. If unit conversions for DataAmount are fixed, this can be reviewed and merged.

lolodomo (changes requested):

bundles/org.openhab.binding.systeminfo/src/main/java/org/openhab/binding/systeminfo/internal/model/SysteminfoInterface.java
```diff
@@ -119,101 +126,101 @@ public interface SysteminfoInterface {
...
      */
-    public @Nullable DecimalType getMemoryUsedPercent();
+    public @Nullable PercentType getMemoryUsedPercent();
 
     // Swap memory info
     /**
      * Returns total size of swap memory
      *
-     * @return memory size in MB or 0, if no there is no swap memory
+     * @return memory size or 0, if there is no swap memory
```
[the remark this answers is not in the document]
mherwege: It is 0, so indeed @Nullable can be removed.

itests/org.openhab.binding.systeminfo.tests/src/main/java/org/openhab/binding/systeminfo/test/SysteminfoOSGiTest.java
```diff
@@ -272,7 +280,8 @@ private void initializeThing(Configuration thingConfiguration, String channelID,
         intializeItem(channelUID, DEFAULT_TEST_ITEM_NAME, acceptedItemType);
     }
 
-    private void assertItemState(String acceptedItemType, String itemName, String priority, State expectedState) {
+    private <T extends State> void assertItemState(String acceptedItemType, String itemName, String priority,
```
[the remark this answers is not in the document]
mherwege: Indeed, there is. Leftover from a trial with providing the State type as a parameter to the method. Shouldn't be there.

lolodomo:
This a breaking change as users will have to change their item types.

lolodomo:
> Test fail because DataAmount unit conversions do not work properly, see [#3207](https://github.com/openhab/openhab-core/issues/3207).
> The code is complete from my perspective. If unit conversions for DataAmount are fixed, this can be reviewed and merged.

What are we doing ?
Can you comment the failing integration until the conversion is fixed in core framework ? In that case, I could merge your changes.
Or should we wait for a fix in core framework ?

lolodomo:
Code is good for me, I am just waiting for a successful compilation.

J-N-K:
> > DataAmount unit conversions do not work properly, see [openhab/openhab-core#3207](https://github.com/openhab/openhab-core/issues/3207).
> 
> Maybe (just an idea) the problem could be that we are using a unit defined by the OH core framework (Units.MEBIBYTE), not an official unit ?

This will also fail. The fix is in https://github.com/openhab/openhab-core/pull/3208.

lolodomo:
> This will also fail. The fix is in [openhab/openhab-core#3208](https://github.com/openhab/openhab-core/pull/3208). But it should be checked if the binding really delivers `kB` (1000 B) or `kiB` (1024 B), same for `MB`/`MiB` and `GB`/`GiB` and also the data transfer rates.

Good remark. Hoping @mherwege can retrieve this information.

> This is quite confusing but hard to change now without breaking existing installations.

Of course, users will have to change their item types. But as this PR improve something (ability for the users to choose their displayed units), this is acceptable, no ?
By the way, @mherwege already broke something for this binding in OH3.4 so this is just something additional to clearly mentioned in the release notes.
Is it ok for you @J-N-K ?

J-N-K:
The binding should return the correct value (probably `MiB` and not `MB`). My comment was regarding changing `1 MB` to `1024 kB` in core. Even if I doubt that someone really expects the 1000 instead of 1024 when using these units…

lolodomo:
> The binding should return the correct value (probably `MiB` and not `MB`). My comment was regarding changing `1 MB` to `1024 kB` in core. Even if I doubt that someone really expects the 1000 instead of 1024 when using these units…

Understood. 
Hopefully, this was not too much used until now. In bindings. 
But your remark means @mherwege should also think if the best channel pattern should use MB or MiB. 

For the 2 other bindings, I hope there is no mismatch between MB and MiB.

mherwege:
So far, the binding always returned MiB (so 1024 x 1024 bytes), but the state description was MB. I kept the state description, but the methods return MiB (there is even a method in the binding ([getSizeInMB](https://github.com/openhab/openhab-addons/blob/fe269b127ae83b09fb90eeb9df646d7599c30365/bundles/org.openhab.binding.systeminfo/src/main/java/org/openhab/binding/systeminfo/internal/model/OSHISysteminfo.java#L484) that actually returns MiB, rather than MB. All methods returning data amounts rely in this method.
I opted to keep the default state description, but this will lead to slight changes in value. As an alternative, I could change the default state descriptions to MiB, this will then keep the values constant. I can live with whatever choice.

mherwege:
> This a breaking change as users will have to change their item types.

I think you should still be able to just keep a Number item without dimension linked to the channels with dimension. So I am not sure this makes it a breaking change.
The MB vs MiB issue is however breaking, whatever choice we make.

mherwege:
> The binding should return the correct value (probably `MiB` and not `MB`).

Makes sense, so went for that in the last commit.

lolodomo (approved):
LGTM

lolodomo:
Can I merge or should I wait for the fix in core framework? 

Can you prepare something for the release notes?

mherwege:
@lolodomo Yes, I will prepare something for the release notes. It should work now as long as one does not do unit conversions on DataAmount and sticks with MiB. That is no worse then it was before, and will get better ones the core change is done.
I now also adjusted the README. Should I still push a commit in this PR, or create a separate one for this?

lolodomo:
> I now also adjusted the README. Should I still push a commit in this PR, or create a separate one for this?

Please do it here. I will wait.

mherwege:
Where do I have to push the content for the release notes?

lolodomo (approved):
LGTM, thank you

lolodomo:
> Where do I have to push the content for the release notes?

Your PR in openhab-distro is fine and should be sufficient.
I don't know if Kai will ask for help when he will write the 3.4 release notes.

PR #13805, opened 2022-11-29
[ipcamera] Instar API updates for new 2k+ range

Skinah:
This PR contains 1 bug fix in the ONVIF connecting code that was causing issues with Instar and possibly other brands as well. The Host http header was blank.
The rest of the changes are to provide support for a newer 2k range of cameras that have a different API implemented in them.

Jar can be RIGHT HAND CLICKED and downloaded from:

### DIFF
--- a/bundles/org.openhab.binding.ipcamera/src/main/java/org/openhab/binding/ipcamera/internal/InstarHandler.java
+++ b/bundles/org.openhab.binding.ipcamera/src/main/java/org/openhab/binding/ipcamera/internal/InstarHandler.java
@@ -133,7 +133,17 @@ public class InstarHandler extends ChannelDuplexHandler {
                     }
                     break;
                 default:
-                    ipCameraHandler.logger.debug("Unknown reply from URI:{}", requestUrl);
+                    if (requestUrl.startsWith("/param.cgi?cmd=setasaction&-server=1&enable=1")
+                            && content.contains("response=\"200\";")) {// new
+                        ipCameraHandler.newInstarApi = true;
+                        ipCameraHandler.logger.debug("Alarm server sucessfully setup for a 2k+ Instar camera");
+                    } else if (requestUrl.startsWith("/param.cgi?cmd=setmdalarm&-aname=server2&-switch=on&-interval=1")
+                            && content.startsWith("[Succeed]set ok")) {
+                        ipCameraHandler.newInstarApi = false;
+                        ipCameraHandler.logger.debug("Alarm server sucessfully setup for a 1080p Instar camera");
+                    } else {
+                        ipCameraHandler.logger.debug("Unknown reply from URI:{}", requestUrl);
+                    }
             }
         } finally {
             ReferenceCountUtil.release(msg);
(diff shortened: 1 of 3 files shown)


bundles/org.openhab.binding.ipcamera/src/main/java/org/openhab/binding/ipcamera/internal/InstarHandler.java
```diff
@@ -62,43 +63,61 @@ public void channelRead(@Nullable ChannelHandlerContext ctx, @Nullable Object ms
             ipCameraHandler.logger.trace("HTTP Result back from camera is \t:{}:", content);
             switch (requestUrl) {
                 case "/param.cgi?cmd=getinfrared":
-                    if (content.contains("var infraredstat=\"auto")) {
+                    if (content.contains("var infraredstat=\"auto") || content.contains("infraredstat=\"2")) {
```
lolodomo, line 66: Isn't a final `\"` missing to match value 2?

bundles/org.openhab.binding.ipcamera/src/main/java/org/openhab/binding/ipcamera/internal/InstarHandler.java
```diff
@@ -62,43 +63,61 @@ public void channelRead(@Nullable ChannelHandlerContext ctx, @Nullable Object ms
...
                     if (content.contains("var m1_enable=\"1\"")) {
                         ipCameraHandler.setChannelState(CHANNEL_ENABLE_MOTION_ALARM, OnOffType.ON);
                     } else {
                         ipCameraHandler.setChannelState(CHANNEL_ENABLE_MOTION_ALARM, OnOffType.OFF);
                     }
                     break;
-                case "/cgi-bin/hi3510/param.cgi?cmd=getaudioalarmattr":// Audio Alarm
-                    if (content.contains("var aa_enable=\"1\"")) {
+                case "/param.cgi?cmd=getalarmattr":// Motion Alarm new
+                    if (content.contains("armed=\"1")) {
```
lolodomo, line 95: Isn't a final `\"` missing to match value 1?

bundles/org.openhab.binding.ipcamera/src/main/java/org/openhab/binding/ipcamera/internal/InstarHandler.java
Skinah: Agree safer in case api gets changed. Made change so vale 22, 222 etc do not match

bundles/org.openhab.binding.ipcamera/src/main/java/org/openhab/binding/ipcamera/internal/InstarHandler.java
Skinah: Agree safer in case api gets changed. Made change so vale 11, 111 etc do not match

Skinah:
All review comments have been addressed.

bundles/org.openhab.binding.ipcamera/src/main/java/org/openhab/binding/ipcamera/internal/InstarHandler.java
```diff
@@ -150,9 +150,32 @@ public void channelRead(@Nullable ChannelHandlerContext ctx, @Nullable Object ms
...
+                    if (ipCameraHandler.newInstarApi) {
+                        ipCameraHandler.sendHttpGET("/param.cgi?cmd=getalarmattr");
+                    } else {
+                        ipCameraHandler.sendHttpGET("/cgi-bin/hi3510/param.cgi?cmd=getmdattr");
+                    }
+                    return;
+                case CHANNEL_ENABLE_PIR_ALARM:
+                    ipCameraHandler.sendHttpGET("/param.cgi?cmd=getpirattr");
+                case CHANNEL_AUTO_LED:
+                    ipCameraHandler.sendHttpGET("/param.cgi?cmd=getinfrared");
```
lolodomo, line 174: Is it expected to have no return/break for these 2 cases ?

As you have a return at the exit of the switch, I would suggest to use break rather than return in each case, WDYT ?

lolodomo (changes requested):

bundles/org.openhab.binding.ipcamera/src/main/java/org/openhab/binding/ipcamera/internal/InstarHandler.java
Skinah: Thanks for picking that up, have now fixed.

lolodomo (approved):
LGTM, thank you

PR #13810, opened 2022-11-30
[jrubyscripting] implement dependency tracking

ccutrer:
watchers had to be refactored similiar to jsscripting. it supports watching any directory referenced from RUBYLIB, as well as the gem home. it properly excludes lib and gem home (as well as other gem homes if you have multiple jruby versions installed) from loading as regular scripts.

this is a breaking change if you don't have RUBYLIB explicitly configured, and you are using the old default directory.

it's expected that the detection of what files and gems any given script uses will be self-identified by the script, presumably by the helper library.

JRubyScriptEngineConfiguration was largely refactored as part of this.
 * CONFIGURATION_PARAMETERS was renamed, and is no longer static, since it's modified every time the configuration is changed
 * OptionalConfigurationElement was simplified since default values are always provided now. this also simplified lots of other code that accesses the current settings.


ccutrer:
https://github.com/ccutrer/openhab-jrubyscripting/commit/3ba1054adff6028d161da3e10349e5a8236b48e9 is the commit to enable this functionality in the helper library

jimtng:
This works well in my installation.

bundles/org.openhab.automation.jrubyscripting/src/main/java/org/openhab/automation/jrubyscripting/internal/JRubyScriptEngineConfiguration.java
```diff
@@ -124,24 +126,62 @@ void configureScriptEngine(ScriptEngineFactory factory) {
...
+
+    /**
+     * Makes Gem home directory if it does not exist
+     */
+    private void ensureGemHomeExists(String gemHome) {
+        File gemHomeDirectory = new File(gemHome);
+        if (!gemHomeDirectory.exists()) {
+            logger.debug("gem_home directory does not exist, creating");
+            if (!gemHomeDirectory.mkdirs()) {
+                logger.warn("Error creating gem_home directory");
```
jlaur, line 183: Should the throw an exception to be handled or return false, so caller would be notified and could avoid continuing when this directory doesn't exist, possibly running into problems?

bundles/org.openhab.automation.jrubyscripting/src/main/java/org/openhab/automation/jrubyscripting/internal/watch/BidiSetBag.java
```diff
@@ -0,0 +1,111 @@
...
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.Map;
+import java.util.Set;
+import java.util.concurrent.locks.ReentrantReadWriteLock;
+
+import org.eclipse.jdt.annotation.NonNullByDefault;
+
+// Copy of org.openhab.core.automation.module.script.rulesupport.internal.loader.collection.BidiSetBag
```
jlaur, line 24: For my understanding, why does it need to be copied? Is it modified in any way, or just plain duplication?

jlaur (changes requested):
Thanks for this adaptation/improvement! Just two questions from my side. @J-N-K, perhaps you could have a quick look as you recently did the same with #13756?

bundles/org.openhab.automation.jrubyscripting/src/main/java/org/openhab/automation/jrubyscripting/internal/watch/BidiSetBag.java
J-N-K: It's in an internal package and therefore not exported from the core package.

bundles/org.openhab.automation.jrubyscripting/src/main/java/org/openhab/automation/jrubyscripting/internal/watch/JRubyGemWatchService.java
```diff
@@ -0,0 +1,74 @@
...
+import org.eclipse.jdt.annotation.Nullable;
+import org.openhab.core.service.AbstractWatchService;
+
+/**
+ * Watches a gem home
+ *
+ * @author Cody Cutrer - Initial contribution
+ */
+@NonNullByDefault
+public class JRubyGemWatchService extends AbstractWatchService {
```
J-N-K, line 33: I'm not so happy with these tons of new watch services, but this might improve after we have https://github.com/openhab/openhab-core/pull/3004 merged which simplifies the interface.

bundles/org.openhab.automation.jrubyscripting/src/main/java/org/openhab/automation/jrubyscripting/internal/watch/JRubyDependencyTracker.java
```diff
@@ -0,0 +1,106 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Tracks Ruby dependencies
+ *
+ * @author Cody Cutrer - Initial contribution
+ */
+@NonNullByDefault
+public class JRubyDependencyTracker implements ScriptDependencyTracker {
```
J-N-K, line 35: This could be a service (component) itself which then is injected into the script engine factory (like in js). I didn't check where the individual configurations values come from, but likely they can be derived from the ConfigAdmin.

J-N-K (commented):
Otherwise LGTM

bundles/org.openhab.automation.jrubyscripting/src/main/java/org/openhab/automation/jrubyscripting/internal/watch/BidiSetBag.java
ccutrer: Honestly, I'd prefer if I didn't have to copy _or_ reference it. I had to because the AbstractDependencyTracker isn't flexible enough for my needs - it assumes a single library directory that it must watch. If it were to be improved, then I could not need to reference this class at all.

bundles/org.openhab.automation.jrubyscripting/src/main/java/org/openhab/automation/jrubyscripting/internal/watch/JRubyDependencyTracker.java
ccutrer: I tried _really_ hard to get this to be an OSGi component. I actually started by just copying the JS files. But I couldn't get around the circular dependence - `JRubyScriptEngineFactory` needs to return the dependency tracker. But the dependency tracker needs access to the config. On the one hand, like you say, I could just use `ConfigAdmin` directly to get the current configuration values, but the raw configuration is wrapped by Java both to provide non-constant defaults (well, it is theoretically constant, but dependent on values from the JRuby bundle), as well as providing some post-processing to the values. I definitely didn't want to attempt to duplicate that.

I also considered just having the `JRubyScriptEngineFactory` implement `ScriptDependencyTracker` like you said, but I wanted to keep it separate in hopes of eliminating the bulk of this class by inheriting from `AbstractScriptDependencyTracker` as discussed in the other comment.

Another option I've thought of, but didn't really explore, is having the configuration itself be an independent component that both `JRubyScriptEngineFactory` and `JRubyDependencyTracker` depend on. If you're adamant that you really want this to be an OSGi component, I can explore that.

bundles/org.openhab.automation.jrubyscripting/src/main/java/org/openhab/automation/jrubyscripting/internal/watch/JRubyGemWatchService.java
ccutrer: Oooh. That'll be nice. To slightly assuage your concerns in the meantime, while this is setup to allow an arbitrary number of watch services, 99% of the time it will be the default config, and there will be just 2 - one for GEM_HOME, and a single one for RUBYLIB. And both of those will be directories nested underneath the JRubyScriptFileWatcher, so the underlying WatchQueueReader won't actually be watching any more, but will have more bookkeeping.

bundles/org.openhab.automation.jrubyscripting/src/main/java/org/openhab/automation/jrubyscripting/internal/JRubyScriptEngineConfiguration.java
ccutrer: +1, I can add.

bundles/org.openhab.automation.jrubyscripting/src/main/java/org/openhab/automation/jrubyscripting/internal/watch/JRubyDependencyTracker.java
J-N-K: I would suggest to keep it like it is now and refactor it later when the core watch service is changed. As a result it should also be possible to easily adapt the `AbstractScriptDependencyTracker` to watch for more than one directory. Since that one needs refactoring anyway with the new watch service, I would prefer not to touch it before the release because time is a bit short.

ccutrer:
@jlaur: sounds like we're going to punt on @J-N-K's comments until core has an updated file watcher. I've addressed your comment on the gem home directory creation. Anything else?

jlaur:
> I've addressed your comment on the gem home directory creation. Anything else?

No, let's merge. :-)

jlaur (approved):
LGTM

PR #13812, opened 2022-12-01
Typos a/an

dilyanpalauzov:


fwolter:
Thank you for this PR! Catching also things like "a {@link AsynchronousSocketChannel}" looks like a bit more sophisticated script. I guess you didn't do that by hand...?

dilyanpalauzov:
I found the places with `git grep` and then changed the files by hand.

fwolter (approved):
Thanks, good work!

fwolter:
There's a formatting error due to a line, which became one character longer. Can you run `mvn spotless:apply`?

wborn (approved):

PR #13821, opened 2022-12-02
[homematic] Fix min/max values for rollershutters

maniac103:
For dimmers, the 1.0 max value sent by CCU was already converted to percent values in the item state description. Do the same thing also for roller shutters.

Previously, the item state description looked e.g. like this:
```
{
    "link": "http://server:8080/rest/items/BathroomBlindsLevel",
    "state": "100",
    "stateDescription": {
        "minimum": 0.0,
        "maximum": 1.0,
        "pattern": "%d %%",
        "readOnly": false,
        "options": [

        ]
    },
    "editable": true,
    "type": "Rollershutter",
    "name": "BathroomBlindsLevel",
    "label": "Level",
    "category": "Blinds",
    "tags": [
        "Point"
    ],
    "groupNames": [
    ]
}
```

... which is obviously wrong. Both `ITEM_DIMMER` and `ITEM_ROLLERSHUTTER` use `PercentTypeConverter` when converting item commands to CCU values, which is dividing the value by 100, so in both cases the maximum value in the state description should look the same.

### DIFF
--- a/bundles/org.openhab.binding.homematic/src/main/java/org/openhab/binding/homematic/internal/type/HomematicTypeGeneratorImpl.java
+++ b/bundles/org.openhab.binding.homematic/src/main/java/org/openhab/binding/homematic/internal/type/HomematicTypeGeneratorImpl.java
@@ -31,7 +31,6 @@ import org.openhab.binding.homematic.internal.model.HmChannel;
 import org.openhab.binding.homematic.internal.model.HmDatapoint;
 import org.openhab.binding.homematic.internal.model.HmDevice;
 import org.openhab.binding.homematic.internal.model.HmParamsetType;
-import org.openhab.binding.homematic.internal.type.MetadataUtils.OptionsBuilder;
 import org.openhab.core.config.core.ConfigDescriptionBuilder;
 import org.openhab.core.config.core.ConfigDescriptionParameter;
 import org.openhab.core.config.core.ConfigDescriptionParameterBuilder;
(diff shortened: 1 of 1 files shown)


maniac103:
@MHerbst Mind having a look at this?

MHerbst:
@maniac103 I know from the past of different problems with blinds and dimmers. I could never test with these devices myself, and therefore I can't judge whether any of your changes may have a negative side effect. 
But in general, the changes look OK to me. Would be good if they would be tested with real devices.

maniac103:
I just remembered I also have a dimmer lying around, will do another test with that one tomorrow. No HMIP devices here though...

lolodomo (approved):
LGTM, thank you

lolodomo:
@maniac103 : shall I wait for you test before merging ?

maniac103:
@lolodomo I think it should be fine to merge. Just tested state description output with a roller shutter:

```
{
    "link": "http://127.0.0.1:8080/rest/items/MEQ0391932_Level",
    "state": "0",
    "stateDescription": {
        "minimum": 0,
        "maximum": 100,
        "pattern": "%d %%",
        "readOnly": false,
        "options": [

        ]
    },
    "editable": true,
    "type": "Rollershutter",
    "name": "MEQ0391932_Level",
    "label": "Level",
    "category": "Blinds",
    "tags": [
        "Point"
    ],
    "groupNames": [

    ]
}
```

and a dimmer:
```
{
    "link": "http://127.0.0.1:8080/rest/items/MEQ1118997_Level",
    "state": "0",
    "stateDescription": {
        "minimum": 0,
        "maximum": 100,
        "pattern": "%d %%",
        "readOnly": false,
        "options": [

        ]
    },
    "editable": true,
    "type": "Dimmer",
    "name": "MEQ1118997_Level",
    "label": "Level",
    "category": "",
    "tags": [
        "Point"
    ],
    "groupNames": [

    ]
}
```

and output looks as expected in both cases now.

PR #13823, opened 2022-12-02
[ecowatt] Add channels for Day+2 and Day+3

lolodomo:


lolodomo:
@jlaur : could you please review this very small PR ?

jlaur:
@lolodomo - it looks simple, but something it itching me with the model. Is there no other way to do this, other than having separate items named as relative offsets from today? It seems to me we lack something, like item arrays. I'm not aware of anything existing though, so just raising this as a design concern to brain storm ideas before merging.

jlaur:
Btw, I had a similar issue with this: https://community.openhab.org/t/dishwasher-price-calculation-automation/139207 where I just ended up with a JSON array. Someone raised the ideas of creating a binding for energy calculations. Here we would have a similar problem.

lolodomo:
The API just exposes the data for 4 days (today, tomorrow, D+2 and D+3).

> before merging

Please merge before 3.4 is released. Unfortunately, we really need that for January in France !

lolodomo:
As there is only 4 channels required, I prefer have them all immediately available for the user.
An alternative would have been to use extensible channels with the offset as a parameter but this looks like not the best solution in the current context.

jlaur (approved):
LGTM

lolodomo:
Thank you

Issue #12577, opened 2022-04-04
[jsscripting] Loading script files crashes openHAB


jlaur:
This is a follow-up bug report similar to #11952 which I was hoping would have been fixed by openhab/openhab-core#2681 and #12022, but the issue seems to remain in 3.4M5.

## Expected Behavior
Loading a JS script from a file should not make the system unstable.

## Current Behavior
After reloading a script file multiple times the system becomes slower and slower until it ultimately will become very unstable with timeouts.

## Steps to Reproduce
Create this file somewhere as _test1.sh_:
```
rules.JSRule({
  name: "Reproduce jsscripting issue",
  description: "Reproduce jsscripting issue",
  triggers: [triggers.ItemStateChangeTrigger('Issue12577', 'OFF', 'ON')],
  execute: data => {
    console.log("Test item triggered: 1");
  }
});
```

And this one as _test2.sh_:
```
rules.JSRule({
  name: "Reproduce jsscripting issue",
  description: "Reproduce jsscripting issue",
  triggers: [triggers.ItemStateChangeTrigger('Issue12577', 'OFF', 'ON')],
  execute: data => {
    console.log("Test item triggered: 2");
  }
});
```

And finally, create this script as _reproduce_issue_12577.sh_:

```
#!/bin/bash
echo "Switch Issue12577" > $OPENHAB_CONF/items/issue_12577.items
i=1
while [ $i -lt 50 ]
do
    if [ $((i % 2)) -eq 0 ]
    then
        cp test1.js "$OPENHAB_CONF/automation/js/test.js"
    else
        cp test2.js "$OPENHAB_CONF/automation/js/test.js"
    fi
    sleep 5
    ((i++))
done
```

Set execute permission:
```
chmod +x reproduce_issue_12577.sh
```

Now run this script and wait a few minutes. For me, after running all 50 iterations, the last log items will look something like this:

```
2022-04-04 22:27:53.195 [INFO ] [rulesupport.loader.ScriptFileWatcher] - Loading script '/etc/openhab/automation/js/test.js'
2022-04-04 22:28:48.771 [INFO ] [org.openhab.automation.script.rules ] - Adding rule: Reproduce jsscripting issue
2022-04-04 22:28:52.169 [INFO ] [rulesupport.loader.ScriptFileWatcher] - Loading script '/etc/openhab/automation/js/test.js'
[...]

PR #13824, opened 2022-12-02
[jsscripting] Fix memory leak that crashes openHAB

florian-h05:
Fixes https://github.com/openhab/openhab-addons/issues/12577.

## Description

When a `.js` script file was changed and reloaded by openHAB, a memory leak in the JS Scripting add-on lead to an openHAB crash after some time.
This PR finally fixes the memory leak causing this.

I found the memory leak after reproducing the issue and taking several heap snapshots. Analyzation of the heap snapshots has shown, that a extremely high number of GraalJS related objects without a GC root existed in the heap. 
Looking after `OpenhabGraalJSScriptEngine` in the heap snapshot, I noticed that there were also numerous instances of the engine, which were not cleaned by GC because they were referenced by `com.oracle.truffle.host.HostObject`.

Research on the web lead me to reading https://github.com/oracle/graaljs/issues/121, where other users describe similar problems. In the issue conversation, one of the GraalJS developers said, that one should not recreate the `GraalJSScriptEngine` each time, but instead share the `Context` (reference https://github.com/oracle/graaljs/issues/121#issuecomment-880056648). If that is not possible, because other code (openhab core in our case) requires a `ScriptEngine`, `HostAccess` should not be created each time, but instead shared as a static variable (reference https://github.com/oracle/graaljs/issues/121#issuecomment-690179954).

During working on this issue, I also noticed that the `close` method of the `GraalJSScriptEngine` was never called (at least my breakpoint was never triggered), until I added a call to `close`.

## Testing

Use the reproduce script from @jlaur to verify the fix:

<details>
<summary>Create this file somewhere as `reproduce_issue_12577.sh`</summary>

```shell
#!/bin/bash
echo "Switch Issue12577" > $OPENHAB_CONF/items/issue_12577.items
i=1
while [ $i -lt 150 ]
do
    echo "Iteration ${i} starting..."
    if [ $((i % 2)) -eq 0 ]
    then
        cp test1.js "$OPENHAB_CONF/automation/js/test.js"
    else
        cp test2.js "$OPENHAB_CONF/automation/js/test.js"
    fi
    sleep 5
    ((i++))
done
```

</details>

<details>
<summary>Add these JavaScript files to the same location:</summary>

test1.js:
```javascript

rules.JSRule({
  name: 'Reproduce jsscripting issue 1',
  description: 'Reproduce jsscripting issue',
  triggers: [triggers.ItemStateChangeTrigger('Issue12577', 'OFF', 'ON')],
  execute: data => {
    console.log('Test item triggered: 1');
  }
});
```

test2.js: 
```javascript

rules.JSRule({
  name: 'Reproduce jsscripting issue 2',
  description: 'Reproduce jsscripting issue',
  triggers: [triggers.ItemStateChangeTrigger('Issue12577', 'ON', 'OFF')],
  execute: data => {
    console.log('Test item triggered: 2');
  }
});
```
</details>

Using this verification approach, I have confirmed that openHAB is still running without any problems after 250!! iterations of the reproduce script, my heap usage graph shows that GC works now:

image

### DIFF
--- a/bundles/org.openhab.automation.jsscripting/src/main/java/org/openhab/automation/jsscripting/internal/OpenhabGraalJSScriptEngine.java
+++ b/bundles/org.openhab.automation.jsscripting/src/main/java/org/openhab/automation/jsscripting/internal/OpenhabGraalJSScriptEngine.java
@@ -234,6 +234,7 @@ public class OpenhabGraalJSScriptEngine
     @Override
     public void close() {
         jsRuntimeFeatures.close();
+        delegate.close();
     }
 
     /**
--- a/bundles/org.openhab.automation.jsscripting/src/main/java/org/openhab/automation/jsscripting/internal/OpenhabGraalJSScriptEngine.java
+++ b/bundles/org.openhab.automation.jsscripting/src/main/java/org/openhab/automation/jsscripting/internal/OpenhabGraalJSScriptEngine.java
@@ -68,6 +68,20 @@ public class OpenhabGraalJSScriptEngine
     private static final String REQUIRE_WRAPPER_NAME = "__wraprequire__";
     // final CommonJS search path for our library
     private static final Path NODE_DIR = Paths.get("node_modules");
+    // Custom translate JS Objects - > Java Objects
+    private static final HostAccess HOST_ACCESS = HostAccess.newBuilder(HostAccess.ALL)
+            // Translate JS-Joda ZonedDateTime to java.time.ZonedDateTime
+            .targetTypeMapping(Value.class, ZonedDateTime.class, (v) -> v.hasMember("withFixedOffsetZone"), v -> {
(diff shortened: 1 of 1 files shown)


digitaldan (approved):
Nice Job!  and I think i was the one who authored the HOST_ACCESS logic in the first place. 
LGTM

J-N-K:
Your analysis is the same I made when I looked into that issue, but I didn‘t find a solution for it. Thanks!

jlaur (approved):
Awesome, thank you very much for this fix! I have waited almost a year for someone to figure this out. :-)

florian-h05:
You're welcome :-)

I also had trouble with this and crashed my openHAB several times, but because I'm on not on Pi (anymore) I have more RAM and a larger heap available, so I didn't encounter a crash as fast as on the Pi.

FYI: I'll likely come up with a PR to upgrade the README and the openhab-js library in the next days, would be great if you could merge this as fast as usual

PR #13827, opened 2022-12-03
[unifi] Migrate uptime channels to Number:Time

### LINKED ISSUE #13826: [unifi] Implement UoM for uptime
`uptime` channels are currently of type `Number` and provided in seconds. This could be migrated to `Number:Time` so UoM would be fully supported.

jlaur:
In order to fully support UoM, the uptime channel type is migrated from `Number` to `Number:Time`.

With this change, the unit "s" is automatically appended in logs etc. when not otherwise specified.

Fixes #13826

### DIFF
--- a/bundles/org.openhab.binding.unifi/README.md
+++ b/bundles/org.openhab.binding.unifi/README.md
@@ -161,7 +161,7 @@ The `wirelessClient` information that is retrieved is available as these channel
 | ap         | String               | Access point (AP) the client is connected to                         | Read        |
 | essid      | String               | Network name (ESSID) the client is connected to                      | Read        |
 | rssi       | Number               | Received signal strength indicator (RSSI) of the client              | Read        |
-| uptime     | Number               | Uptime of the client (in seconds)                                    | Read        |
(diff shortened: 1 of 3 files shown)


jlaur:
@Hilbrand, @lolodomo - WDYT? It could potentially break rules as now being `QuantityType`. Would it be acceptable with a notice about this in openhab-distro?

lolodomo (approved):
LGTM, thank you

lolodomo:
The breaking change is because users will have to change item type ?

jlaur:
@lolodomo - that would be one of the scenarios, yes. Now we also have openhab/openhab-core#3153 which might help when existing item is defined as `Number` and channel is now changed to `Number:Time`. I will need to perform some tests to know for sure if/how the change can cause problems.

jlaur:
@lolodomo - I have just tested on a fresh snapshot system and created a `Number` item from UI using the snapshot version of the binding. Then upgraded the binding to new version, and it's fully compatible:
image

So I don't see any compatibility problems or reason to mention in upgrade notes.

jlaur:
Perhaps openhab/openhab-core#3115 also helped ensuring compatibility for above case.

Issue #13577, opened 2022-10-21
[wundergroundupdatereceiver] Invalid description of Channel rain#rainin


MikeTheTux:
The channel `rain#rainin` is using the parameter `rainin` from the WU query.
`rainin` is the current rainfall-rate (e.g. in `mm/h`) and not the "Hourly Rain" (rain during last hour, like for the day, week, month and year channels), as mentioned in the Channel description.

## Possible Solution
* Update channel description towards _rainfall-rate_
* Update UOM from `Number:Length` (e.g. `mm`) to `Number:VolumetricFlowRate` (e.g. `mm/h`)

In addition the readme shall be updated:

### Old:

> The server thus listens at http(s)😕/:/weatherstation/updateweatherstation.php ...

### New:

> The server thus listens at `http(s)://<openHAB-IP>:<openHAB-Port>/weatherstation/updateweatherstation.php` ...

You may also want to provide an example, how the weatherstation needs to be configured. For my station I need to add the `?` behind the `php.`

## Your Environment
* Version used: openHAB 3.3.0

danieldemus:
According to the documentation at wunderground.com: "rainin - [rain inches over the past hour)] -- the accumulated rainfall in the past 60 min."
I can't provide an useful example of how to configure a weatherstation, because they are probably configured very differently. On mine, you just enter the hostname of the server in one field and a port number in another and it works.

MikeTheTux:
> According to the documentation at wunderground.com: "rainin - [rain inches over the past hour)] -- the accumulated rainfall in the past 60 min."

I can follow your argumentation when following the WU API documentation.

Nevertheless I can see, that my weatherstation is reporting a rainfall rate and not an accumulation over the last hour - otherwise the graph would show a ramp during continuous rain:
![grafik](https://user-images.githubusercontent.com/44850211/210166803-d05bc17d-e281-4f1e-9bf4-8944a9c8f1cf.png)

danieldemus:
Apparently this must be an implementation detail of the various weather stations. 
![rain](https://user-images.githubusercontent.com/4918595/210171464-61f66f20-1a70-46df-b16b-34138d5d84a7.png)
While it does say hourly rain rate in the ipobserver gui, the plateaus in the above graphs are decidedly not a current flow rate, but some aggregation for a period.
[...]

PR #13831, opened 2022-12-03
[wundergroundupdatereceiver] Use html entities in normal body otherwise they are parsed as tags

danieldemus:
Fix spelling
Fixes the addendum in #13577 
Signed-off-by: Daniel Demus <daniel-github@demus.dk>


jlaur (approved):
Thanks.

PR #13832, opened 2022-12-03
[plugwiseha] Fix set the scene for a zone

lsiepel:
Fixes community reported bug: https://community.openhab.org/t/plugwise-home-automation-binding-for-openhab/81299/140

The request body was not properly formatted and i added additional trace logging to make future troubleshooting easier.


bundles/org.openhab.binding.plugwiseha/src/main/java/org/openhab/binding/plugwiseha/internal/api/model/PlugwiseHAControllerRequest.java
```diff
@@ -225,6 +227,22 @@ private ContentResponse getContentResponse(int retries) throws PlugwiseHAExcepti
...
+        if (logger.isTraceEnabled()) {
+            String content = "";
+            final ContentProvider provider = request.getContent();
+            if (provider != null) {
+                final Iterator<ByteBuffer> it = provider.iterator();
+                while (it.hasNext()) {
+                    final ByteBuffer next = it.next();
+                    final byte[] bytes = new byte[next.capacity()];
+                    next.get(bytes);
+                    content = new String(bytes, StandardCharsets.UTF_8);
```
lolodomo, line 239: That us strange, you are looping but finally consider only the last item.

lsiepel: Not intentionally. This remote API always responds resulting with 1 item in the iterator. I changed it so that if it holds more it is stored to the output. Thanks, good catch!

lolodomo (approved):
LGTM, thank you

PR #13662, opened 2022-11-05
[jdbc] Add console maintenance commands

jlaur:
This introduces some basic tools for verifying integrity of persisted items and corresponding tables.

### Example
```
openhab> openhab:jdbc
Usage: openhab:jdbc tables list [all] - list tables (all = include valid)
Usage: openhab:jdbc tables clean [<itemName>] [force] - clean inconsistent items (remove from index and drop tables)
openhab> openhab:jdbc tables list
Table     Row count  Item                  Status
--------  ---------  --------------------  ----------------------
Item0281      54217  WiserSocket1_Voltage  Item missing
Item0283          0  NotFound              Item and table missing
Item0284          0  NotFound2             Table missing
blah              1                        Orphan table
openhab> openhab:jdbc tables clean
Cleaning up all inconsistent items...
WiserSocket1_Voltage -> skipped/failed.
FindesIkke -> done.
FindesIkke2 -> done.
openhab> openhab:jdbc tables list
Table     Row count  Item                  Status
--------  ---------  --------------------  ----------------------
Item0281      54217  WiserSocket1_Voltage  Item missing
blah              1                        Orphan table
openhab> openhab:jdbc tables clean WiserSocket1_Voltage force
Cleaning up item WiserSocket1_Voltage... done.
openhab> openhab:jdbc tables list
Table     Row count  Item  Status
--------  ---------  ----  ----------------------
blah              1        Orphan table
```

### Test

JAR for testing: [org.openhab.persistence.jdbc-3.4.0-SNAPSHOT.jar](https://drive.google.com/file/d/1bVcBkz2IiPw2w47DxdxPMWaEeKmPS-v8/view?usp=sharing)

Place the corresponding database driver into the addons directory. For example, download the [MySQL connector](https://repo1.maven.org/maven2/mysql/mysql-connector-java/8.0.30/mysql-connector-java-8.0.30.jar) and rename it to `mysql-connector-java.jar` or the [MariaDB connector](https://repo1.maven.org/maven2/org/mariadb/jdbc/mariadb-java-client/3.0.8/mariadb-java-client-3.0.8.jar) and rename it to `mariadb-java-client.jar`.

Resolves #13661

Issue #13661, opened 2022-11-05
[jdbc] Provide console tools for identifying problems and assisting migration


jlaur:
It would be cool having a way to interact with JDBC persistence to help identifying issues:
- Tables where referenced item no longer exists.
- Mapped items where table doesn't exist.
- Mapped items where neither table nor item exists.
- Orphan tables (not having any mapping).

Perhaps some kind of assisted repair could be added also.

It would also be nice to be able to do a dry-run migration to verify everything before performing a real migration. See also #13544.

lsiepel:
Would it also be possible to add a check if the column type matches the item type ?

jlaur:
> Would it also be possible to add a check if the column type matches the item type ?

I actually have that on my internal TODO. :-) It's a bit more tricky because I need to check column definitions in a way that works for all databases, and I only have a setup for MySQL. Which database do you use?

jlaur:
@lsiepel - I have a draft version now for finding incorrect column types. It needs further work, but I can see already that I have many value columns with unsigned integer types which are expected to be signed. I will need to have a closer look at this and understand how this discrepancy was introduced. I would also like to provide a repair option for safe fixes which cannot truncate data.

PR #13837, opened 2022-12-04
[jdbc] Fix console command 'tables' for SQLite

jlaur:
Follow-up to #13662 so the console command now also supports SQLite.

See https://community.openhab.org/t/hildebrand-glowmarkt-api-binding/139091/20

Related to #13661


lolodomo (approved):
LGTM

PR #13841, opened 2022-12-04
[miele] Migrate time channels from DateTime to Number:Time

### LINKED ISSUE #13840: [miele] Migrate time channels from DateTime to Number:Time
This issue is for discussing and proposing a migration from `DateTime` to `Number:Time` for channels **duration**, **elapsed** and **finish**. This is a continuation from #13316.

This would be a breaking change requiring users to recreate items with the new type. In case anyone would persist such channels (which I doubt), for JDBC it would mean migrating from TIMESTAMP to DOUBLE, which would also be breaking.

I'm not sure if this proposal is controversial, since `DateTimeType` actually supports time-only. However, this implementation doesn't seem "clean" to me, at least not when we have `QuantityType` supporting time as unit. Example:

```xtend
rule "Oven 5 minutes left"
when
    Item Oven_FinishTime changed to "1970-01-01T00:05:00.000+0100"
then
   //
end
```

or (because of time-zone issues, epoch seconds cannot be used directly as there will be an offset from local time-zone):

```javascript
val DateTimeType durationState = (Dishwasher_Duration.state as DateTimeType)
val ZonedDateTime durationZoned = durationState.getZonedDateTime()
val endTime = now.plusSeconds(durationZoned.getHour() * 60 * 60 + durationZoned.getMinute() * 60).toInstant()
```

and in logs:
```
2022-12-03 22:22:22.368 [INFO ] [openhab.event.ItemStateChangedEvent ] - Item 'TumbleDryer_FinishTime' changed from 1970-01-01T00:57:00.000+0100 to 1970-01-01T00:56:00.000+0100
2022-12-03 22:23:22.482 [INFO ] [openhab.event.ItemStateChangedEvent ] - Item 'TumbleDryer_FinishTime' changed from 1970-01-01T00:56:00.000+0100 to 1970-01-01T00:55:00.000+0100
```

Usually users will probably prefer to display these values in the format HH:MM. This is equally possible with `Number:Time` by simply using state pattern `%1$tR`. However, this would simplify rules and make logs more readable:

```xtend
rule "Oven 5 minutes left"
when
    Item Oven_FinishTime changed to "5 min"
then
   //
end
```

And in logs:
```
2022-12-03 22:50:21.737 [INFO ] [openhab.event.ItemStateChangedEvent ] - Item 'TumbleDryer_FinishTime' changed from 29 min to 28 min
```

Since time-only `DateTime` is overlapping in functionality with `Number:Time`, I'm not sure about the reason for having a time-only DateTime at all, and when to actually prefer it over Number:Time?

jlaur:
Migrate channels **duration**, **elapsed** and **finish** from `DateTime` to `Number:Time` since they represent an amount of time, not a specific point in time.

This is a breaking change requiring UI things to be recreated and linked items for these channels to be recreated. Persistence of these channels is unlikely, but would also break. Rules would need adaptation. Therefore I will provide a note in openhab-distro. After recreation of items, the default display format is the same, so UI's should be unaffected. openhab/openhab-distro#1435 prepared.

See also #13316 where related channels were fixed to represent a point in time.

Additional details are available in the linked issue.

Fixes #13840

### DIFF
--- a/bundles/org.openhab.binding.miele/README.md
+++ b/bundles/org.openhab.binding.miele/README.md
@@ -126,9 +126,9 @@ Channels available for each appliance type are listed below.
 | rawPhase            | Number               | Read       | Current phase of the program running on the appliance as raw number |
 | start               | DateTime             | Read       | Programmed start time of the program                                |
 | end                 | DateTime             | Read       | End time of the program (programmed or running)                     |
-| duration            | DateTime             | Read       | Duration of the program running on the appliance                    |
-| elapsed             | DateTime             | Read       | Time elapsed in the program running on the appliance                |
-| finish              | DateTime             | Read       | Time to finish the program running on the appliance                 |
+| duration            | Number:Time          | Read       | Duration of the program running on the appliance                    |
+| elapsed             | Number:Time          | Read       | Time elapsed in the program running on the appliance                |
+| finish              | Number:Time          | Read       | Time to finish the program running on the appliance                 |
 | door                | Contact              | Read       | Current state of the door of the appliance                          |
 | switch              | Switch               | Write      | Switch the appliance on or off                                      |
 | powerConsumption    | Number:Power         | Read       | Power consumption by the currently running program on the appliance |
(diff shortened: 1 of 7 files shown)


kaikreuzer (approved):
lgtm and makes sense to fix even though it is a breaking change.

PR #13842, opened 2022-12-04
[bosesoundtouch] Improve SAT erros and remove dependency

lsiepel:

I tried to fix most SAT errors, mainly null annotations, some are left. I find it hard to get those fixed. As i don't have the devices myself, i'm unable to do testing and don't want it to break just for the sake of SAT. Also removed the dependency on the `org.apache.commons` 
Relative to https://github.com/openhab/openhab-addons/issues/7722 

A jar for testing is here: https://1drv.ms/u/s!AnMcxmvEeupwjpdVz3L1-7qFEd_Prw?e=l3k56x Should only be merged if some testing is done.
I left the obsolete packages `org.xml.sax.helpers.XMLReaderFactory` untouched as that would need more rewriting and without a device it would take too much time.

### DIFF
--- a/bundles/org.openhab.binding.bosesoundtouch/src/main/java/org/openhab/binding/bosesoundtouch/internal/CommandExecutor.java
+++ b/bundles/org.openhab.binding.bosesoundtouch/src/main/java/org/openhab/binding/bosesoundtouch/internal/CommandExecutor.java
@@ -18,6 +18,9 @@ import java.util.Collection;
 import java.util.HashMap;
 import java.util.Map;
 
+import org.eclipse.jdt.annotation.NonNullByDefault;
+import org.eclipse.jdt.annotation.Nullable;
+import org.eclipse.jetty.websocket.api.Session;
 import org.openhab.binding.bosesoundtouch.internal.handler.BoseSoundTouchHandler;
 import org.openhab.core.library.types.DecimalType;
 import org.openhab.core.library.types.NextPreviousType;
@@ -36,14 +39,15 @@ import org.slf4j.LoggerFactory;
  * @author Thomas Traunbauer - Initial contribution
  * @author Kai Kreuzer - code clean up
  */
+@NonNullByDefault
 public class CommandExecutor implements AvailableSources {
     private final Logger logger = LoggerFactory.getLogger(CommandExecutor.class);
 
     private final BoseSoundTouchHandler handler;
 
     private boolean currentMuted;
-    private ContentItem currentContentItem;
-    private OperationModeType currentOperationMode;
+    private @Nullable ContentItem currentContentItem;
+    private @Nullable OperationModeType currentOperationMode;
 
     private Map<String, Boolean> mapOfAvailableFunctions;
 
@@ -53,6 +57,7 @@ public class CommandExecutor implements AvailableSources {
      * @param handler the handler that created this CommandExecutor
      */
     public CommandExecutor(BoseSoundTouchHandler handler) {
+        mapOfAvailableFunctions = new HashMap<>();
         this.handler = handler;
         init();
     }
@@ -66,11 +71,7 @@ public class CommandExecutor implements AvailableSources {
     public void updatePresetContainerFromPlayer(Map<Integer, ContentItem> playerPresets) {
         playerPresets.forEach((k, v) -> {
             try {
-                if (v != null) {
-                    handler.getPresetContainer().put(k, v);
-                } else {
-                    handler.getPresetContainer().remove(k);
-                }
+                handler.getPresetContainer().put(k, v);
             } catch (ContentItemNotPresetableException e) {
                 logger.debug("{}: ContentItem is not presetable", handler.getDeviceName());
             }
@@ -104,7 +105,10 @@ public class CommandExecutor implements AvailableSources {
      */
     public void addCurrentContentItemToPresetContainer(DecimalType command) {
         if (command.intValue() > 6) {
-            addContentItemToPresetContainer(command.intValue(), currentContentItem);
+            ContentItem localContentItem = currentContentItem;
+            if (localContentItem != null) {
+                addContentItemToPresetContainer(command.intValue(), localContentItem);
+            }
         } else {
             logger.warn("{}: Only PresetID >6 is allowed", handler.getDeviceName());
         }
@@ -118,8 +122,11 @@ public class CommandExecutor implements AvailableSources {
     public void getInformations(APIRequest apiRequest) {
         String msg = "<msg><header " + "deviceID=\"" + handler.getMacAddress() + "\"" + " url=\"" + apiRequest
                 + "\" method=\"GET\"><request requestID=\"0\"><info type=\"new\"/></request></header></msg>";
-        handler.getSession().getRemote().sendStringByFuture(msg);
-        logger.debug("{}: sending request: {}", handler.getDeviceName(), msg);
+        Session localSession = handler.getSession();
+        if (localSession != null) {
+            localSession.getRemote().sendStringByFuture(msg);
+            logger.debug("{}: sending request: {}", handler.getDeviceName(), msg);
+        }
     }
 
     /**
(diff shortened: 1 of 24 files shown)


bundles/org.openhab.binding.bosesoundtouch/src/main/java/org/openhab/binding/bosesoundtouch/internal/CommandExecutor.java
```diff
@@ -53,6 +57,7 @@ public class CommandExecutor implements AvailableSources {
      * @param handler the handler that created this CommandExecutor
      */
     public CommandExecutor(BoseSoundTouchHandler handler) {
+        mapOfAvailableFunctions = new HashMap<>();
```
wborn, line 60: It's even better to assign the value where the field is declared and add the `final` modifier to the field.

bundles/org.openhab.binding.bosesoundtouch/src/main/java/org/openhab/binding/bosesoundtouch/internal/PresetContainer.java
```diff
@@ -41,6 +44,7 @@ public class PresetContainer {
      * Creates a new instance of this class
      */
     public PresetContainer(Storage<ContentItem> storage) {
+        mapOfPresets = new HashMap<>();
```
wborn, line 47: It's even better to assign the value where the field is declared and add the `final` modifier to the field. While you're at it you might also wanto to change the field type from the `HashMap` implementation to the `Map` interface.

bundles/org.openhab.binding.bosesoundtouch/src/main/java/org/openhab/binding/bosesoundtouch/internal/XMLResponseHandler.java
```diff
@@ -82,19 +84,28 @@ public XMLResponseHandler(BoseSoundTouchHandler handler,
         this.handler = handler;
         this.commandExecutor = handler.getCommandExecutor();
         this.stateSwitchingMap = stateSwitchingMap;
-        init();
+        states = new Stack<>();
```
wborn, line 87: Can this also be assigned where the field is declared?

bundles/org.openhab.binding.bosesoundtouch/src/main/java/org/openhab/binding/bosesoundtouch/internal/XMLResponseProcessor.java
```diff
@@ -29,10 +30,12 @@
  * @author Christian Niessner - Initial contribution
  * @author Thomas Traunbauer - Initial contribution
  */
+
+@NonNullByDefault
 public class XMLResponseProcessor {
     private BoseSoundTouchHandler handler;
 
-    private Map<XMLHandlerState, Map<String, XMLHandlerState>> stateSwitchingMap;
+    private Map<XMLHandlerState, Map<String, XMLHandlerState>> stateSwitchingMap = new HashMap<>();
```
wborn, line 38: Can it be `final` too?

bundles/org.openhab.binding.bosesoundtouch/src/main/java/org/openhab/binding/bosesoundtouch/internal/handler/BoseSoundTouchHandler.java
```diff
@@ -359,96 +365,113 @@ public PresetContainer getPresetContainer() {
...
     }
 
     @Override
-    public void onWebSocketError(Throwable e) {
-        logger.debug("{}: Error during websocket communication: {}", getDeviceName(), e.getMessage(), e);
-        updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.COMMUNICATION_ERROR, e.getMessage());
-        if (commandExecutor != null) {
-            commandExecutor.postOperationMode(OperationModeType.OFFLINE);
+    public void onWebSocketError(@Nullable Throwable e) {
+        Throwable localThrowable = (e != null) ? e : new Throwable("Null Exception passed to onWebSocketError");
```
wborn, line 377: It's usually better to create exceptions of a specific type:

```suggestion
        Throwable localThrowable = (e != null) ? e : new IllegalStateException("Null Exception passed to onWebSocketError");
```

wborn (commented):
Thanks for the improvements! I have added a few comments below:

wborn:
> Applied review comments and also got rid of the obsolete dependency. Please have another look. I also updated the JAR for testing.

I looked everywhere but could not find any changes? :thinking:

lsiepel:
Sor

> > Applied review comments and also got rid of the obsolete dependency. Please have another look. I also updated the JAR for testing.
> 
> I looked everywhere but could not find any changes? 🤔

Sorry, doing too much at the same time. I really need to stop multitask. I just forgot to push the local commits. All should be fine now.

bundles/org.openhab.binding.bosesoundtouch/src/main/java/org/openhab/binding/bosesoundtouch/internal/ContentItem.java
```diff
@@ -27,30 +28,23 @@
...
     public ContentItem() {
-        source = "";
-        sourceAccount = null;
-        location = null;
-        presetable = false;
-        itemName = null;
-        presetID = 0;
-        containerArt = null;
-        additionalAttributes = new HashMap<>();
     }
```
wborn, line 48: Since it no longer has an implementation, it is equal to the default constructor, so you can probably remove the whole constructor.

bundles/org.openhab.binding.bosesoundtouch/src/main/java/org/openhab/binding/bosesoundtouch/internal/XMLResponseHandler.java
```diff
@@ -131,21 +135,18 @@ public void startElement(String uri, String localName, String qName, Attributes
...
                 break;
             case MsgHeader:
                 if ("request".equals(localName)) {
                     state = XMLHandlerState.Unprocessed; // TODO implement request id / response tracking...
                 } else {
-                    if (logger.isDebugEnabled()) {
-                        logger.warn("{}: Unhandled XML entity during {}: '{}'", handler.getDeviceName(), curState,
-                                localName);
-                    }
+                    logger.warn("{}: Unhandled XML entity during {}: '{}'", handler.getDeviceName(), curState,
```
wborn, line 148: I don't know how often it happens, but it might spam the log now. It's logged on debug above as well.

```suggestion
                    logger.debug("{}: Unhandled XML entity during {}: '{}'", handler.getDeviceName(), curState,
```

bundles/org.openhab.binding.bosesoundtouch/src/main/java/org/openhab/binding/bosesoundtouch/internal/ContentItem.java
lsiepel: Done

bundles/org.openhab.binding.bosesoundtouch/src/main/java/org/openhab/binding/bosesoundtouch/internal/XMLResponseHandler.java
lsiepel: done

grassel82:
Hi everybody,

I would like to thank you for your great work. I started with installing a new openHAB server in version 3.4.0 and I have 2 Soundtouch devices - Soundtouch 10 and Soundtouch 130 which in reality is a Cinemate 130 with a SoundTouch Wireless link adapter. I had issues with state updates and I did a lot of debugging to understand what is happening. I even put my openhab log settings in Debug mode and I saw that the requests were often send twice or not correctly parsed. This has all been fixed since I downloaded your last JAR file from your onedrive account to use it to bind my soundtouch devices.

I still have 2 problems and I do not find any solution to this. As I'm quite new to openHAB, I'm not sure if it is a configuration issue on my side or still problems in the binding.
- PRESETS are not shown and I cannot select any other configured PRESETS on both devices.
- operationMode changes are not working. You cannot switch to TUNEIN from SPOTIFY or any other mode

I could take some traces to send it to you for debugging if this would help?

Kind regards,
Gilles

lsiepel:
I created an issue, let's continue from there

Issue #13679, opened 2022-11-08
[keba] State Item should be Number not String


maihacke:
The example items configuration defines the item for the state as string item:

`String 						KebaState  				"Operating State [%s]"						{channel="keba:kecontact:1:state"}`

This in incorrect end results in the state being shown as a bogus datetime.
Correct would be a number item:

`Number 						KebaState  				"Operating State [%s]"						{channel="keba:kecontact:1:state"}`

lolodomo:
That is true, the "state" channel type is defined as Number. Can you please propose a fix of the README example ? I will then merge your proposal.

maihacke:
Yes, will do in the next days.

maihacke:
Interestingly when trying to "correct" the readme example there was already the right type. Nevertheless in the describing table it was wrong and is now corrected.

PR #13845, opened 2022-12-05
[keba] Correct item type for state item in readme.md

maihacke:
# Minor change in documentation for state item

Fix documentation to correct item type for state item
Should fix https://github.com/openhab/openhab-addons/issues/13679

# Description
Fix documentation to correct item type for state item


lolodomo:
The channel definition still mention Number and not String. 
https://github.com/openhab/openhab-addons/blob/main/bundles/org.openhab.binding.keba/src/main/resources/OH-INF/thing/kecontact.xml#L75

lolodomo:
And the binding is building a DecimalType for this channel:
https://github.com/openhab/openhab-addons/blob/da59cdd255a66275dd7ae11dd294fedca4942d30/bundles/org.openhab.binding.keba/src/main/java/org/openhab/binding/keba/internal/handler/KeContactHandler.java#L315

So Number looks good and String is wrong.

lolodomo:
In fact, you need to make the inversed fix. 
Read the issue.

maihacke:
Damn was totally confused this morning. Now it's (hopefully) the intended change. Sorry.

lolodomo (approved):
LGTM

Issue #13775, opened 2022-11-24
[jdbc] Charts don't show the correct dates/times because persistence doesn't check with the right timezeone


edman007:
Hi,
When I chart my data, it does not show the most recent data, this happened in OpenHab 2.5.x, I recently upgraded to 3.3.0, and it's still happening. Specifically, I live in New York, my timezone is America/New_York (GMT-05:00). I have this correctly set in the region settings.

## Expected Behavior
Charts should show up to the second data

## Current Behavior
Charts omit the last 5 hours of data.
![screenshot](https://user-images.githubusercontent.com/7518958/203830681-455774a9-78c7-42c4-8b81-68382dadda65.png)

## Possible Solution
Queries should be written in UTC, I do not beleive the fix in https://github.com/openhab/openhab-addons/pull/9445 was ever actually correct

## Steps to Reproduce (for Bugs)
1. Set timezone to something other than UTC (I think it needs to be a GMT-XX:XX. America/New_York is a good example
2. Persist some data
3. Chart the data

## Context
This is related to the bugs referenced in  https://github.com/openhab/openhab-addons/pull/9445

I can see that the queries insert with NOW() for time, and that's stored in a native database format. That means dates/times need to be queried and displayed in the connection local. This typically should be UTC for a variety of reasons, regardless of the users local.

However, that's not what's happening, the insert of data looks like this (and it's correct)
`
11:26:29.069 [DEBUG] [g.openhab.persistence.jdbc.dto.ItemVO] - JDBC:ItemVO tableName=solar_system_power_0103; newTableName=null; 
11:26:29.070 [DEBUG] [enhab.persistence.jdbc.db.JdbcBaseDAO] - JDBC::getItemType: Try to use ItemType NUMBERITEM for Item solar_system_power
11:26:29.070 [DEBUG] [enhab.persistence.jdbc.db.JdbcBaseDAO] - JDBC::storeItemValueProvider: item 'solar_system_power' as Type 'NUMBERITEM' in 'solar_system_power_0103' with state '3.926'
11:26:29.070 [DEBUG] [enhab.persistence.jdbc.db.JdbcBaseDAO] - JDBC::storeItemValueProvider: itemState: '3.926'
11:26:29.071 [DEBUG] [g.openhab.persistence.jdbc.dto.ItemVO] - JDBC:ItemVO setValueTypes dbType=DOUBLE; javaType=class java.lang.Double;
[...]

PR #13850, opened 2022-12-05
[jdbc] Fix date filter to be inclusive (Derby)

### LINKED ISSUE #9906: [jdbc] JDBC persistence ignores operator set in FilterQuery

jlaur:
Derby implementation was missed in #13734 and again in #13843

Discovered by @edman007 in relation to another JDBC issue: https://github.com/openhab/openhab-addons/issues/13775#issuecomment-1336655836

Fixes #9906


wborn (approved):
LGTM!

PR #13855, opened 2022-12-05
[jsscripting] Fix "Error removing engine" because context is already closed

florian-h05:
In recent PR #13824, I added closing the context when the engine is closed, but core seems to have problems with that.

It logs: "[ERROR] [ipt.internal.ScriptEngineManagerImpl] - Error removing ScriptEngine java.lang.IllegalStateException: The Context is already closed."

Using the reproduce strategy from #13824, I can confirm that the change of this PR does not bring back the memory leak!

### DIFF
--- a/bundles/org.openhab.automation.jsscripting/src/main/java/org/openhab/automation/jsscripting/internal/OpenhabGraalJSScriptEngine.java
+++ b/bundles/org.openhab.automation.jsscripting/src/main/java/org/openhab/automation/jsscripting/internal/OpenhabGraalJSScriptEngine.java
@@ -237,7 +237,6 @@ public class OpenhabGraalJSScriptEngine
     @Override
     public void close() {
         jsRuntimeFeatures.close();
-        delegate.close();
     }
 
     /**


florian-h05:
@J-N-K Does core close the engine/context then?

jlaur:
> Does core close the engine/context then?

`GraalJSScriptEngine` implements `AutoCloseable`, could this be the reason?

florian-h05:
> GraalJSScriptEngine implements AutoCloseable, could this be the reason?

I‘ve checked core, it calls close if the engine implements AutoCloseable (see https://github.com/openhab/openhab-core/blob/ff75505e703d55a6d62a800561ccee9b39ab7ca1/bundles/org.openhab.core.automation.module.script/src/main/java/org/openhab/core/automation/module/script/internal/ScriptEngineManagerImpl.java#L222).

The error was coming from there, because the context was already closed. So, seems like core would close the context.

> This PR will simply revert https://github.com/openhab/openhab-addons/commit/d5502c0cd9d2f3247a260892bf2c2d59998e3d18 which was not directly related to the memory leak fixed in https://github.com/openhab/openhab-addons/pull/13855, right @florian-h05?

Yes, correct. I had a breakpoint inside the GraalJSScriptEngine `close` method and noticed that my breakpoint was never called, so I added this. It could possibly leak memory when the engine is not properly closed, but I expect core to do that (see above).

I guess it is safe to merge, but I‘d feel better if anyone could verify that the memory leak is not back. Unfortunately, I haven‘t got the time left this evening. So it would either be tomorrow, or you can check @jlaur. Attaching VisualVM and watching the heap usage should be the best option.

jlaur:
> I guess it is safe to merge, but I‘d feel better if anyone could verify that the memory leak is not back. Unfortunately, I haven‘t got the time left this evening. So it would either be tomorrow, or you can check @jlaur. Attaching VisualVM and watching the heap usage should be the best option.

Unfortunately I was not even able to test the memory leak fix because I couldn't figure out how to compile jsscripting with core prior to openhab/openhab-core#3168 being merged. I manually reverted #13756 for the sake of this, but didn't succeed. Also out of time for today. If you would prefer to wait until tomorrow to fully verify, that's fine with me. We can also merge and you could still check tomorrow. I would suspect that the other commits you did in #13855 fixed the problem, while the last commit just introduced this double dispose problem.

jlaur:
See also openhab/openhab-core#2681

florian-h05:
> If you would prefer to wait until tomorrow to fully verify, that's fine with me.

I’ll verify today and ping you after then.
FYI: I usually also have problems „backporting“ the addon, but since I am using SNAPSHOT on my dev system I don’t have to and can just use the latest version.

> See also https://github.com/openhab/openhab-core/pull/2681

Thanks for the link, I’ve already seen this, so core should close the engine.

florian-h05:
@jlaur I've tested this change by using the reproduce script with 150 iterations, heap usage is normal, also looking at the heap dump, I can confirm that the memore leak is not back!!

This can be safely merged now.

jlaur (approved):
Thanks!

PR #13869, opened 2022-12-07
Upgrade commons-net to 3.9.0

wborn:
Related to openhab/openhab-core#3200

Also reworks a few POMs/features so the commons-net OHC feature is used and the dependency is not embedded. 

Replaces: 

* https://github.com/openhab/openhab-addons/pull/13856
* https://github.com/openhab/openhab-addons/pull/13857


goopilot (approved):

jlaur (approved):
Thanks!

PR #13873, opened 2022-12-07
[TR064] Fix SAT warnings

lsiepel:
Nothing fanyc, just fixed some SAT warnings. I left one for the Test, it is partly annotated, so i assume i cannot improve that ;-)


wborn:
> so i assume i cannot improve that ;-)

It was really an easy fix. :slightly_smiling_face:

jlaur (approved):
Thanks.

Issue #12627, opened 2022-04-18
[mielecloud] Integration tests failure


lolodomo:
It is failing apparently sysrematically.
Build on Windows:
```
[INFO] --- bnd-testing-maven-plugin:6.2.0:testing (default) @ org.openhab.binding.mielecloud.tests ---
[INFO] Matching glob *
[INFO] Testing D:\dev\openhab3\git\openhab-addons\itests\org.openhab.binding.mielecloud.tests\itest.bndrun:
org.openhab.binding.mielecloud [org.openhab.binding.mielecloud.internal.config.servlet.CreateBridgeServlet] WARN : Thing creation failed because there was no binding available that supports the thing.
org.openhab.binding.mielecloud [org.openhab.binding.mielecloud.internal.config.servlet.CreateBridgeServlet] WARN : Cannot create bridge: E-mail address is missing.
org.openhab.binding.mielecloud [org.openhab.binding.mielecloud.internal.config.servlet.CreateBridgeServlet] WARN : Cannot create bridge: E-mail address is missing.
org.openhab.binding.mielecloud [org.openhab.binding.mielecloud.internal.config.servlet.CreateBridgeServlet] WARN : Bridge exists but has no handler.
org.openhab.binding.mielecloud [org.openhab.binding.mielecloud.internal.config.servlet.CreateBridgeServlet] WARN : Cannot create bridge: Bridge UID is missing.
org.openhab.binding.mielecloud [org.openhab.binding.mielecloud.internal.webservice.sse.SseStreamParser] WARN : SSE connection failed unexpectedly: org.eclipse.jetty.client.HttpResponseException: HTTP protocol violation: Authentication challenge without WWW-Authenticate header
org.openhab.binding.mielecloud [org.openhab.binding.mielecloud.internal.webservice.sse.SseStreamParser] WARN : SSE connection failed unexpectedly: org.eclipse.jetty.client.HttpResponseException: HTTP protocol violation: Authentication challenge without WWW-Authenticate header
...
org.openhab.binding.mielecloud [org.openhab.binding.mielecloud.internal.webservice.sse.SseStreamParser] WARN : SSE connection failed unexpectedly: org.eclipse.jetty.client.HttpResponseException: HTTP protocol violation: Authentication challenge without WWW-Authenticate header
[...]

PR #13875, opened 2022-12-08
[mielecloud] Integration test fixes: remove things after each integration test

BjoernLange:
When comparing the integration tests of the mielecloud binding to the ones of the nest binding I saw that things are removed after each integration test run. This PR adds this behavior to the mielecloud integration tests, hoping that this is a cause for them failing in the CI build. Also see #12627.


wborn (approved):
Let's see if it helps! 👍

PR #13878, opened 2022-12-08
[homekit] implement basic command completion

ccutrer:
just the sub-commands.

since the show command is already a "search", I didn't feel it's super important to implement on that


yfre (approved):
LGTM
thank you!

lolodomo (approved):
LGTM

PR #13880, opened 2022-12-08
[Nanolaef] Visual State Bugfix

austvik:
Fix the visual state channel name.

Also:
- Better name (from "state" to "visual state") of the (new, not released yet) channel
- Better logs which has helped us debug the problem
- Some more information on when it will work in the README

Discussed on https://community.openhab.org/t/state-and-layout-are-unreliable/141798/19


jlaur (approved):
Thanks! Indeed that is a better channel name.

austvik:
Thanks! All naming credit should go to Stefan!

PR #13883, opened 2022-12-08
[livisismarthome] Fix checkstyle

lsiepel:
Minor checkstyle fixes. Could not find away to use the org.slf4j.Logger in the Test classes and left it as is.

### DIFF
--- a/bundles/org.openhab.binding.livisismarthome/src/main/java/org/openhab/binding/livisismarthome/internal/manager/DeviceStructureManager.java
+++ b/bundles/org.openhab.binding.livisismarthome/src/main/java/org/openhab/binding/livisismarthome/internal/manager/DeviceStructureManager.java
@@ -44,8 +44,8 @@ public class DeviceStructureManager {
     private final Logger logger = LoggerFactory.getLogger(DeviceStructureManager.class);
 
     private final FullDeviceManager deviceManager;
-    private final Map<String, DeviceDTO> deviceMap;
-    private final Map<String, DeviceDTO> capabilityIdToDeviceMap;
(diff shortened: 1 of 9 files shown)


bundles/org.openhab.binding.livisismarthome/src/main/java/org/openhab/binding/livisismarthome/internal/manager/DeviceStructureManager.java
```diff
@@ -44,8 +44,8 @@ public class DeviceStructureManager {
     private final Logger logger = LoggerFactory.getLogger(DeviceStructureManager.class);
 
     private final FullDeviceManager deviceManager;
-    private final Map<String, DeviceDTO> deviceMap;
-    private final Map<String, DeviceDTO> capabilityIdToDeviceMap;
+    private final Map<String, DeviceDTO> deviceMap = Collections.synchronizedMap(new HashMap<>());
```
lolodomo, line 47: A reason to not use a ConcurrentHashMap like at the next line?

lolodomo (commented):
Just one question

bundles/org.openhab.binding.livisismarthome/src/main/java/org/openhab/binding/livisismarthome/internal/manager/DeviceStructureManager.java
lsiepel: I only moved it and did not analyse if it could be altered to a ` ConcurrentHashMap`. Now i did and i believe a lock on the entire object is not needed, a lock on the item would be enough. So changed it.

lolodomo (approved):
LGTM

lolodomo:
Do not compile

Issue #9007, opened 2020-11-12
[evohome] Setpoint Send command 0 doesn't cancel the permanent override


HellStorm666:
since a few weeks I have a problem with the evohome binding.

Before, when I sended setpoint.sendCommand(0) to the evohome the permanent override got cancelled and thus evohome started to follow the schedule again.

However, if I now send this command, it tries to set the temp to 0, that fails and the temperature stays at a given temperature with permanent override.

I don't see any strange errors in the logs.

Also, sending a setpoint.sendCommand(10) does change the info in AopenHab, but doenst change it in the Evohome app, so my guess is that it doens't write to evohome.
If I change settings in evohome, it will display the changed values in Openhab2, so reading the data does work.

HellStorm666:
Any update on this?

Looks like writing any value to the evohome does get changed in OpenHAB, but a few seconds after that it gets overwritten with the value in the Evohome app value.

So, as example, the set-temp in the evohome app (and display) is 17C.
Setting this to 10C in OpenHAB gives a 'set temp to 10C' and 'expected temp to become 10C'. And after a few seconds I get a temp is 17C.

So it looks like the value isn't send to evohome, or evohome ignores the value.

Birchcroft:
I have the same problem on 1 of my heating zones.  On careful inspection I found that I had set the "Set Point Item" to type "number:temperature" for that heating zone, whereas it was set to type "number" for the others.   Changing it to type "number" appears to have resolved the problem, although I need to monitor it for a while to check.

lsiepel:
I might look into this issue and come up with a fix. Are you available for performing some tests, run it for a couple of days?

Birchcroft:
Don't really have a lot of time at the moment, although I might be able to play a bit.

I can confirm that since my last post I believe that the problem relates to the "Type" of the Item associated with the setpoint.

If I use Type = number everything works as expected 

However 

If I use Type = number: temperature then I cannot change the setpoint using openhab.  Every time I try it reverts to the value set by the EvoHome controller.

So I just use a dimensionless number for set point and don't have a problem.

On 7 January 2023 12:44:37 GMT, lsiepel ***@***.***> wrote:
>I might look into this issue and come up with a fix. Are you available for performing some tests, run it for a couple of days?
>
>
>-- 
>Reply to this email directly or view it on GitHub:
>https://github.com/openhab/openhab-addons/issues/9007#issuecomment-1374469150
>You are receiving this because you commented.
>
[...]

PR #13885, opened 2022-12-09
[evohome] Add null annotation and minor refactoring

lsiepel:

- Fixes: https://github.com/openhab/openhab-addons/issues/9007
- Added UoM for both temperature channels
- Added default semantic tags for both temperature channels
- Added null annotations.
- Fixed checkstyle SAT warnings

Left the warnings about unused private fields untouched as these are part of the DTO responses and might be used in the future. 

The jar (3.4.1) is here: https://1drv.ms/u/s!AnMcxmvEeupwjp4MT2ZumTKtulvBWg?e=yrmAmM
(updated after last commit as of 2023-01-07 15:19)

I keep this here as draft while some tests are being performed. As i don't own such a device, we have to wait for others to complete the testing.


bundles/org.openhab.binding.evohome/src/main/java/org/openhab/binding/evohome/internal/api/ApiAccess.java
```diff
@@ -188,18 +192,18 @@ private <TOut> TOut doRequest(HttpMethod method, String url, Map<String, String>
...
-
-            headers.put("Authorization", "Bearer " + authenticationData.getAccessToken());
-            headers.put("applicationId", applicationId);
+    private @Nullable <TOut> TOut doAuthenticatedRequest(HttpMethod method, String url,
+            @Nullable Object requestContainer, @Nullable Class<TOut> outClass) throws TimeoutException {
+        Map<String, String> headers = new HashMap<>();
+        Authentication localAuthenticationData = authenticationData;
+        String localApplicaitonId = applicationId;
+
+        if (localAuthenticationData != null && localApplicaitonId != null) {
```
jlaur, line 201: Behavior is changed, is it intentional? Previously `applicationId` header would be set to null, and other headers set correctly. Now no headers will be set, but request will still be made (without any headers).

bundles/org.openhab.binding.evohome/src/main/java/org/openhab/binding/evohome/internal/api/ApiAccess.java
```diff
@@ -37,17 +39,17 @@
  * @author Jasper van Zuijlen - Initial contribution
  *
  */
+@NonNullByDefault
 public class ApiAccess {
     private static final int REQUEST_TIMEOUT_SECONDS = 5;
     private final Logger logger = LoggerFactory.getLogger(ApiAccess.class);
     private final HttpClient httpClient;
-    private final Gson gson;
+    private final @Nullable Gson gson = new GsonBuilder().create();
```
jlaur, line 47: Previously it didn't have to be nullable because constructor was guaranteed to assign it. What is the advantage of this change?

bundles/org.openhab.binding.evohome/src/main/java/org/openhab/binding/evohome/internal/api/models/v2/dto/request/HeatSetPoint.java
```diff
@@ -43,12 +43,15 @@ public class HeatSetPoint {
         timeUntil = null;
     }
 
+    @SuppressWarnings("unused")
```
jlaur, line 46: Why is this suppression needed?

bundles/org.openhab.binding.evohome/src/main/java/org/openhab/binding/evohome/internal/discovery/EvohomeDiscoveryService.java
```diff
@@ -86,13 +89,21 @@ private void discoverDevices() {
...
+                        for (TemperatureControlSystem tcs : gateway.getTemperatureControlSystems()) {
+                            if (tcs != null) {
+                                addDisplayDiscoveryResult(location, tcs);
+                                for (Zone zone : tcs.getZones()) {
+                                    if (zone != null) {
+                                        addZoneDiscoveryResult(location, zone);
+                                    }
+                                }
+                            }
+                        }
```
jlaur, line 106: Consider reducing nesting here by inverting logic with `return` and `continue`.

bundles/org.openhab.binding.evohome/src/main/java/org/openhab/binding/evohome/internal/handler/EvohomeAccountBridgeHandler.java
```diff
@@ -163,27 +184,32 @@ private boolean checkInstallationInfoHasDuplicateIds(Locations locations) {
         return result;
     }
 
+    @SuppressWarnings("null")
```
jlaur, line 187: I think this is hiding a problem, see below.

bundles/org.openhab.binding.evohome/src/main/java/org/openhab/binding/evohome/internal/handler/EvohomeAccountBridgeHandler.java
```diff
@@ -163,27 +184,32 @@ private boolean checkInstallationInfoHasDuplicateIds(Locations locations) {
...
     private void disposeApiClient() {
-        if (apiClient != null) {
-            apiClient.logout();
+        synchronized (this) {
+            if (this.apiClient != null) {
+                this.apiClient.logout();
```
jlaur, line 191:
```suggestion
            EvohomeApiClient localApiCLient = apiClient;
            if (localApiCLient != null) {
                localApiCLient.logout();
```

bundles/org.openhab.binding.evohome/src/main/java/org/openhab/binding/evohome/internal/handler/EvohomeAccountBridgeHandler.java
```diff
@@ -163,27 +184,32 @@ private boolean checkInstallationInfoHasDuplicateIds(Locations locations) {
...
+                this.apiClient = null;
+            }
         }
-        apiClient = null;
     }
 
+    @SuppressWarnings("null")
```
jlaur, line 197: Same here.

bundles/org.openhab.binding.evohome/src/main/java/org/openhab/binding/evohome/internal/handler/EvohomeAccountBridgeHandler.java
```diff
@@ -163,27 +184,32 @@ private boolean checkInstallationInfoHasDuplicateIds(Locations locations) {
...
     }
 
-    private boolean checkConfig() {
+    private boolean checkConfig(EvohomeAccountConfiguration configuration) {
         String errorMessage = "";
 
-        if (configuration == null) {
-            errorMessage = "Configuration is missing or corrupted";
-        } else if (configuration.username == null || configuration.username.isEmpty()) {
+        if (configuration.username.isEmpty()) {
```
jlaur, line 210: Or even:
```suggestion
        if (configuration.username.isBank()) {
```

bundles/org.openhab.binding.evohome/src/main/java/org/openhab/binding/evohome/internal/handler/EvohomeAccountBridgeHandler.java
```diff
@@ -163,27 +184,32 @@ private boolean checkInstallationInfoHasDuplicateIds(Locations locations) {
...
             errorMessage = "Username not configured";
-        } else if (configuration.password == null || configuration.password.isEmpty()) {
+        } else if (configuration.password.isEmpty()) {
```
jlaur, line 212:
```suggestion
        } else if (configuration.password.isBlank()) {
```

bundles/org.openhab.binding.evohome/src/main/java/org/openhab/binding/evohome/internal/handler/EvohomeAccountBridgeHandler.java
```diff
@@ -236,15 +266,29 @@ private void updateThings() {
...
+                                    for (ZoneStatus zone : tcs.getZones()) {
+                                        String zoneId = zone.getZoneId();
+                                        if (zoneId != null) {
+                                            idToZoneMap.put(zoneId, zone);
+                                            zoneIdToTcsIdMap.put(zoneId, systemId);
+                                        }
+                                    }
+                                }
+                            }
+                        }
```
jlaur, line 291: I would suggest to make this more readable by reducing the nesting.

jlaur (changes requested):
Thanks for the refactoring. I have added some comments and questions (as I didn't pull branch and loaded into my IDE).

bundles/org.openhab.binding.evohome/src/main/java/org/openhab/binding/evohome/internal/api/ApiAccess.java
lsiepel: Think i posible wrong an assumption here. Changed it, thanks.

bundles/org.openhab.binding.evohome/src/main/java/org/openhab/binding/evohome/internal/api/EvohomeApiClient.java
```diff
@@ -139,33 +142,33 @@ private void setHeatingZoneOverride(String zoneId, HeatSetPoint heatSetPoint) th
...
-    private UserAccount requestUserAccount() throws TimeoutException {
+    private @Nullable UserAccount requestUserAccount() throws TimeoutException {
         String url = EvohomeApiConstants.URL_V2_BASE + EvohomeApiConstants.URL_V2_ACCOUNT;
         return apiAccess.doAuthenticatedGet(url, UserAccount.class);
     }
 
     private Locations requestLocations() throws TimeoutException {
-        Locations locations = new Locations();
-        if (useraccount != null) {
+        Locations locations = null;
```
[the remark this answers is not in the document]
lsiepel: This was intentionally. Because `return apiAccess.doAuthenticatedGet(url, Locations.class);` returns nullable Location. A check is needed before returning it. So i also have to declare it before the if statement. If you have further suggestions, let me know.

bundles/org.openhab.binding.evohome/src/main/java/org/openhab/binding/evohome/internal/api/ApiAccess.java
lsiepel: Your right.

bundles/org.openhab.binding.evohome/src/main/java/org/openhab/binding/evohome/internal/api/EvohomeApiClient.java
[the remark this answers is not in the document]
jlaur: Ah, I missed that. It would be better if `doAuthenticatedGet` either returned non-null or threw an exception, but this would probably be outside scope of your PR.

bundles/org.openhab.binding.evohome/src/main/java/org/openhab/binding/evohome/internal/api/models/v2/dto/request/HeatSetPoint.java
lsiepel: As this is a DTO class for sending a request this might be part of the json model to send to the server. I did not want to touch it, but i do want to get rid of the compiler warning. Nothing more. It would be nice if SAT would detect such DTO class and hides the unsued warning when the property has been annotated with SerializedName.

bundles/org.openhab.binding.evohome/src/main/java/org/openhab/binding/evohome/internal/handler/EvohomeAccountBridgeHandler.java
lsiepel: Because the block is `synchronized` it should not be needed to use a local variabel. The compiler still shows a warning, i have seen this pattern in many bindings. I could change it to a local variabel. But why is that preferred over a synchronized block?

bundles/org.openhab.binding.evohome/src/main/java/org/openhab/binding/evohome/internal/handler/EvohomeAccountBridgeHandler.java
lsiepel: https://github.com/openhab/openhab-addons/pull/13885#discussion_r1057214429

bundles/org.openhab.binding.evohome/src/main/java/org/openhab/binding/evohome/internal/handler/EvohomeAccountBridgeHandler.java
lsiepel: https://github.com/openhab/openhab-addons/pull/13885#discussion_r1057214429

bundles/org.openhab.binding.evohome/src/main/java/org/openhab/binding/evohome/internal/handler/EvohomeAccountBridgeHandler.java
jlaur: Hmm, is it really needed to synchronize that block over just using a local variable?

lsiepel: i guess not. changed it. thanks

bundles/org.openhab.binding.evohome/src/main/java/org/openhab/binding/evohome/internal/api/ApiAccess.java
```diff
@@ -89,18 +91,16 @@ public void setApplicationId(String applicationId) {
...
      * @throws TimeoutException Thrown when a request times out
      */
-    public <TOut> TOut doRequest(HttpMethod method, String url, Map<String, String> headers, String requestData,
-            String contentType, Class<TOut> outClass) throws TimeoutException {
-        TOut retVal = null;
+    public @Nullable <TOut> TOut doRequest(HttpMethod method, String url, Map<String, String> headers,
+            @Nullable String requestData, String contentType, @Nullable Class<TOut> outClass) throws TimeoutException {
         logger.debug("Requesting: [{}]", url);
-
+        @Nullable
```
jlaur, line 97: Is this annotation needed for local variable?

lsiepel: When removed this warning appears:
`[WARNING] C:\oh-dev\openhab-addons\bundles\org.openhab.binding.evohome\src\main\java\org\openhab\binding\evohome\internal\api\ApiAccess.java:[98,23] Null type mismatch (type annotations): 'null' is not compatible to the free type variable 'TOut'`

bundles/org.openhab.binding.evohome/src/main/java/org/openhab/binding/evohome/internal/handler/EvohomeAccountBridgeHandler.java
```diff
@@ -236,15 +262,31 @@ private void updateThings() {
...
-        for (LocationStatus location : apiClient.getInstallationStatus()) {
-            for (GatewayStatus gateway : location.getGateways()) {
-                for (TemperatureControlSystemStatus tcs : gateway.getTemperatureControlSystems()) {
-                    idToTcsMap.put(tcs.getSystemId(), tcs);
-                    tcsIdToGatewayMap.put(tcs.getSystemId(), gateway);
-                    for (ZoneStatus zone : tcs.getZones()) {
-                        idToZoneMap.put(zone.getZoneId(), zone);
-                        zoneIdToTcsIdMap.put(zone.getZoneId(), tcs.getSystemId());
+        EvohomeApiClient localApiClient = apiClient;
+        if (localApiClient != null) {
```
jlaur, line 266: Again nesting. :)
```suggestion
        if (localApiClient == null) {
            return;
        }
```

bundles/org.openhab.binding.evohome/src/main/java/org/openhab/binding/evohome/internal/handler/EvohomeAccountBridgeHandler.java
```diff
@@ -236,15 +262,31 @@ private void updateThings() {
...
+            // First, create a lookup table
+            LocationsStatus localLocationsStatus = localApiClient.getInstallationStatus();
+            if (localLocationsStatus != null) {
```
jlaur, line 269:
```suggestion
            if (localLocationsStatus == null) {
                return;
            }
```

jlaur:
@lsiepel - should we wait for someone to test this, or are you confident about merging this? @Nebula83, are you still around?

bundles/org.openhab.binding.evohome/src/main/java/org/openhab/binding/evohome/internal/handler/EvohomeAccountBridgeHandler.java
lsiepel: This was intentionally, so that the second block at line 297 of this method would still be executed. I changed it as you suggested. 
The consequence is that when the apiclient == null, all thinghandlers do not update their things (reset them to no zones etc). That maybe a good, on the other hand it looks a functionall change to me.

jlaur: Sorry, I missed that, I'll have another look.

jlaur: It doesn't seem to be a functional change since all maps would be empty, thus the called methods would only be provided with null arguments, thus wouldn't do anything. Or am I still missing something?

lsiepel:
> @lsiepel - should we wait for someone to test this, or are you confident about merging this? @Nebula83, are you still around?

I think it would be good to have some actual tests as there are no tests to this bundle and null annotation refactoring have been the cause for many bugs :-) I'll also ask on the community forums, but without some enhancements not much users are willing to spend their time on it.

bundles/org.openhab.binding.evohome/src/main/java/org/openhab/binding/evohome/internal/handler/EvohomeAccountBridgeHandler.java
```diff
@@ -236,15 +262,33 @@ private void updateThings() {
...
             for (GatewayStatus gateway : location.getGateways()) {
+                if (gateway == null) {
+                    continue;
+                }
                 for (TemperatureControlSystemStatus tcs : gateway.getTemperatureControlSystems()) {
-                    idToTcsMap.put(tcs.getSystemId(), tcs);
-                    tcsIdToGatewayMap.put(tcs.getSystemId(), gateway);
+                    String systemId = tcs.getSystemId();
+                    if (systemId == null) {
+                        continue;
```
jlaur, line 282: I guess it could still be relevant to iterate `tcs.getZones()` and populate `idToZoneMap`?

bundles/org.openhab.binding.evohome/src/main/java/org/openhab/binding/evohome/internal/handler/EvohomeAccountBridgeHandler.java
lsiepel: Again, you right. I thought empty HashMap would be passed, but i double checked and null values are passed. So i changed it should be good now.

jlaur: I read read handler code but would need to double-check, but actually I found one update which accepts null arguments and use that to mark thing as offline? If I found the correct handler, actually you are correct that it's a functional change, and probably undesired. But it's very convoluted...

bundles/org.openhab.binding.evohome/src/main/java/org/openhab/binding/evohome/internal/handler/EvohomeAccountBridgeHandler.java
lsiepel: I partly reverted the commits that fixed the nesting review comments as otherwise the `zoneHandler.update` and `tcsHandler.update` are not called. As these methods have functionality to set the things offline etc when the passed arguments are null. It would change the functionality of the binding and that is beyond the scope of this PR.

jlaur: The old code populated `idToZoneMap` even if `tcs.getSystemId()` would return null, so there is still a (probably undesired) change in behavior.

lsiepel: Might be true, changed it so that idToZoneMap should always be populated

bundles/org.openhab.binding.evohome/src/main/java/org/openhab/binding/evohome/internal/handler/EvohomeAccountBridgeHandler.java
lsiepel: fixed

lsiepel:
Ready for merge as tests have been performed, all seems fine.

bundles/org.openhab.binding.evohome/src/main/java/org/openhab/binding/evohome/internal/handler/EvohomeAccountBridgeHandler.java
```diff
@@ -202,7 +224,11 @@ private void startRefreshTask() {
 
     private void update() {
         try {
-            apiClient.update();
+            EvohomeApiClient localApiCLient = apiClient;
+            if (localApiCLient != null) {
+                localApiCLient.update();
+                apiClient = localApiCLient;
```
jlaur, line 230: What is the purpose of this line?

lsiepel: None :-)

jlaur (approved):
Thanks!

jlaur:
@lsiepel - you could change the PR title to be about the UoM channel additions, and we could add "enhancement" label to have this mentioned in the release notes.

lsiepel:
> @lsiepel - you could change the PR title to be about the UoM channel additions, and we could add "enhancement" label to have this mentioned in the release notes.

Updated it, feel free to adjust it to release note needs.

PR #13888, opened 2022-12-09
[siemensrds] Binding no longer fails if thermostat does not use optional features

### LINKED ISSUE #13887: [siemensrds] Binding would fail if the thermostat did not use an optional feature
### Issue

A user reported on the forum that if a thermostat does not use an optional feature, then the binding would fail.
Particular cases are..
1. The outside temperature sensor connect is not used.
1. The second relay output is not used. e.g. for domestic hot water

https://community.openhab.org/t/siemens-rds-bindings/127062/6

### Solution

The binding should not fail if the thermostat does not use a particular feature; instead the respective channel should simply display the state 'UNDEF'.

andrewfg:
Fixes #13887

### Issue

As described in #13887 the binding would fail if the thermostat did not use an optional feature.

Furthermore, unfortunately, during the creation and testing of this PR it was discovered that the name of the JUnit test class 'RdsTestData' was such that the respective tests were not actually being run during the GitHub CI Build process. And since the tests were never being run, it had not been noticed that those tests would have been failing anyway, because of a) the test JSON files were encoded using the ANSI character set instead of UTF-8, and b) some of the JUnit 'assertEquals' tests were failing due to changes in the mean-time to the way that QuantityType UoM values were being handled in the binding and/or the OH core classes.

### Solutions

This PR fixes the following things..

1. The binding no longer fails if the thermostat does not use a particular feature. Instead the channel state becomes 'UNDEF'.
2. Test class 'RdsTestData' has been renamed to 'TestRdsData' so that the unit tests are now actually run during CI build.
3. The QuantityType UoM code has been refactored to eliminate a compiler warning and to add some missing UoMs.
4. The charset encoding of the JSON test files has been changed from ANSI to UTF-8

### DIFF
--- a/bundles/org.openhab.binding.siemensrds/src/main/java/org/openhab/binding/siemensrds/internal/RdsDataPoints.java
+++ b/bundles/org.openhab.binding.siemensrds/src/main/java/org/openhab/binding/siemensrds/internal/RdsDataPoints.java
@@ -83,8 +83,7 @@ public class RdsDataPoints {
      * execute an HTTP GET command on the remote cloud server to retrieve the JSON
      * response from the given urlString
      */
-    protected static String httpGenericGetJson(String apiKey, String token, String urlString)
-            throws RdsCloudException, IOException {
+    protected static String httpGenericGetJson(String apiKey, String token, String urlString) throws IOException {
         /*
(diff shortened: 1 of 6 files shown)


lolodomo (approved):
LGTM, thank you

PR #13889, opened 2022-12-09
[neohub] Recover faster if NeoHub produces empty responses

### LINKED ISSUE #13829: [neohub] Hub goes offline due to 'Null or invalid response'
On a system with two NeoHubs, one of the two hubs is sporadically going offline with the warning message 'Null or invalid response'.

From the log, it looks like the request times out with no response after 10 seconds, however a response does finally arrive after 12 seconds. It seems very unlikely that the hub would actually take as much as 12 seconds to send a response (it is usually "all or nothing"), so I suspect that there may be a thread synchronisation issue in the binding (or the OH core, or dependency libraries). The issue has only been observed since OH 3.4.M4 so I am wondering if recent core changes, or dependency library changes, in (say) the WebSocket code may have triggered this issue. Also I am wondering if the issue may be limited to my atypical installation which is running two instances of the NeoHub bridge handler within the binding.

I will investigate and eventually post a fix.

```
2022-12-03 18:08:42.915 [TRACE] [ding.neohub.internal.NeoHubWebSocket] - Sending request: {"message_type":"hm_get_command_queue","message":"{\"token\":\"xxx\",\"COMMANDS\":[{\"COMMAND\":\"{'GET_LIVE_DATA':0}\",\"COMMANDID\":1}]}"}
2022-12-03 18:08:52.930 [DEBUG] [ding.neohub.internal.NeoHubWebSocket] - Null or invalid response.
2022-12-03 18:08:52.931 [WARN ] [inding.neohub.internal.NeoHubHandler] - hub 'neohub:neohub:g24x' device data polling error: failed to create device data response
2022-12-03 18:08:54.723 [TRACE] [ding.neohub.internal.NeoHubWebSocket] - onMessage: msg:{"command_id":1,"device_id":"54:10:EC:A8:07:CA","message_type":"hm_set_command_response","response":"
..
```

andrewfg:
Resolves #13829

### Issue

As mentioned in #13829 the bridge sporadically goes offline (due to 'Null or invalid response'), and it remains offline until the next polling cycle produces a valid response, whereupon it goes online again.

Apparently the reason for this is that there is a bug in the NeoHub, whereby it does indeed sporadically produce empty response packages!!

Note: the NeoHub does produce a response so it **is** technically online, however the response is empty, so the binding has no data to update the channels.

The issue is being discussed on the [HeatMiser Developer portal](https://dev.heatmiser.com/t/neohub-produces-sporadic-empty-responses/485), and it seems the solution (work around) is to simply repeat the command immediately so the NeoHub produces a non empty response.

Furthermore, during testing of this PR I became aware of some inconsistencies between the various methods of the `NeoHubSocket` resp. `NeoHubWebSocket` classes..
- The exception types `NeoHubException` resp. `IOException` were used inconsistently between the two classes.
- The logging messages, formatting, and methodology was different between the two classes.
- The JSON validity checking was inconsistent between the two classes.

### Solution

This PR makes the following changes..
1. If the hub produces an empty response, instead of waiting for the next polling cycle, the request is repeated immediately.
2. Use of exception types `NeoHubException` / `IOException` is now consistent over `NeoHubSocket` / `NeoHubWebSocket`.
3. Logging messages, formatting, and methodology are now consistent over `NeoHubSocket` / `NeoHubWebSocket`.
4. JSON validity checking is now consistent over `NeoHubSocket` / `NeoHubWebSocket`.
5. Made some improvements to JavaDocs.
6. Added another JUnit test.

### DIFF
--- a/bundles/org.openhab.binding.neohub/src/main/java/org/openhab/binding/neohub/internal/NeoHubException.java
+++ b/bundles/org.openhab.binding.neohub/src/main/java/org/openhab/binding/neohub/internal/NeoHubException.java
@@ -15,7 +15,9 @@ package org.openhab.binding.neohub.internal;
 import org.eclipse.jdt.annotation.NonNullByDefault;
 
 /**
- * The {@link NeoHubException} is a custom exception for NeoHub
+ * The {@link NeoHubException} is a custom checked exception for NeoHubs. It is thrown when the NeoHub encounters an
+ * error that is NOT I/O related. i.e. if the binding is able to connect to the NeoHub but the NeoHub returns unexpected
+ * results.
  *
  * @author Andrew Fiddian-Green - Initial contribution
  *
--- a/bundles/org.openhab.binding.neohub/src/main/java/org/openhab/binding/neohub/internal/NeoHubHandler.java
+++ b/bundles/org.openhab.binding.neohub/src/main/java/org/openhab/binding/neohub/internal/NeoHubHandler.java
@@ -148,11 +148,11 @@ public class NeoHubHandler extends BaseBridgeHandler {
         NeoHubSocketBase socket;
         try {
             if (config.useWebSocket) {
-                socket = new NeoHubWebSocket(config);
+                socket = new NeoHubWebSocket(config, thing.getUID().getAsString());
             } else {
-                socket = new NeoHubSocket(config);
+                socket = new NeoHubSocket(config, thing.getUID().getAsString());
             }
-        } catch (NeoHubException e) {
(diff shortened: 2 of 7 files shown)


lolodomo:
It looks to me more an enhancement than a bug resolution?

andrewfg:
^
Changed tag from 'bug' to 'enhancement'

andrewfg:
@lolodomo .. but even if it is 'only' an enhancement it would be nice if it could get into RC1

lolodomo (approved):
LGTM, thank you

PR #13893, opened 2022-12-09
[nanoleaf] More color for less network calls

austvik:
This is a refactoring that moves the "get panel color" out of the panel handler and into a separate class, with callbacks.

This makes us do only one REST call to get colors instead of one per panel that is a thing. It also lets us retrieve colors for all panels -
 also those that doesn't have a thing in OpenHAB,

While testing this out, I found a bug where solid colors set in the app wasn't reflected in neither the controller nor panel channels, and that should also be fixed now.


bundles/org.openhab.binding.nanoleaf/src/main/java/org/openhab/binding/nanoleaf/internal/colors/NanoleafPanelColors.java
```diff
@@ -0,0 +1,148 @@
...
+public class NanoleafPanelColors {
+
+    private final Logger logger = LoggerFactory.getLogger(NanoleafPanelColors.class);
+
+    // holds current color data per panel
+    private final Map<Integer, HSBType> panelColors = new ConcurrentHashMap<>();
+    private final Map<Integer, NanoleafPanelColorChangeListener> panelChangeListeners = new ConcurrentHashMap<>();
+    private final List<NanoleafControllerColorChangeListener> controllerListeners = new CopyOnWriteArrayList<>();
+
+    private boolean updatePanelColorNoController(Integer panelId, HSBType color) {
```
jlaur, line 43: Here and in other method signatures in this class: Could primitive type (`int`) be used?

bundles/org.openhab.binding.nanoleaf/src/main/java/org/openhab/binding/nanoleaf/internal/handler/NanoleafControllerHandler.java
```diff
@@ -919,6 +932,117 @@ private void sendRhythmCommand(Command command) throws NanoleafException {
...
+            updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.COMMUNICATION_ERROR,
+                    "@text/error.nanoleaf.panel.communication");
+            logger.debug("Panel data could not be retrieved: {}", nue.getMessage());
+        }
+    }
+
+    void parsePanelData(NanoleafControllerConfig config, ContentResponse panelData) {
+        // panelData is in format (numPanels, (PanelId, 1, R, G, B, W, TransitionTime) * numPanel)
+        @Nullable
+        Write response = gson.fromJson(panelData.getContentAsString(), Write.class);
```
jlaur, line 1012: `JsonSyntaxException` possible here, seems unhandled?

bundles/org.openhab.binding.nanoleaf/src/main/java/org/openhab/binding/nanoleaf/internal/handler/NanoleafPanelHandler.java
```diff
@@ -166,33 +159,39 @@ private void stopAllJobs() {
...
         updateStatus(panelStatus.getStatus(), panelStatus.getStatusDetail());
+        updateState(CHANNEL_PANEL_COLOR, currentPanelColor);
         logger.debug("Panel {} status changed to {}-{}", this.getThing().getUID(), panelStatus.getStatus(),
                 panelStatus.getStatusDetail());
+
+        Bridge bridge = getBridge();
+        if (bridge != null) {
+            NanoleafControllerHandler handler = (NanoleafControllerHandler) bridge.getHandler();
+            if (handler != null) {
+                handler.getColorInformation().registerChangeListener(getPanelID(), this);
```
jlaur, line 170: To have both null and type safety, you could consider something like this:
```suggestion
            ThingHandler handler = bridge.getHandler();
            if (handler instanceof NanoleafControllerHandler) {
                ((NanoleafControllerHandler) handler).getColorInformation().registerChangeListener(getPanelID(), this);
```

bundles/org.openhab.binding.nanoleaf/src/main/java/org/openhab/binding/nanoleaf/internal/handler/NanoleafPanelHandler.java
```diff
@@ -286,98 +276,33 @@ public void updatePanelGesture(int gesture) {
         }
     }
 
-    public String getPanelID() {
+    public Integer getPanelID() {
```
jlaur, line 279: Would primitive type `int` be sufficient here?

bundles/org.openhab.binding.nanoleaf/src/main/java/org/openhab/binding/nanoleaf/internal/handler/NanoleafPanelHandler.java
```diff
@@ -286,98 +276,33 @@ public void updatePanelGesture(int gesture) {
...
         String panelID = getThing().getConfiguration().get(CONFIG_PANEL_ID).toString();
-        return panelID;
+        return Integer.valueOf(panelID);
```
jlaur, line 281: Would this work as it's defined as `Integer` in configuration definition?
```java
return (Integer) this.getConfig().get(CONFIG_PANEL_ID);
```

jlaur (changes requested):
Thanks, it looks good! I have only a few comments for consideration.

bundles/org.openhab.binding.nanoleaf/src/main/java/org/openhab/binding/nanoleaf/internal/colors/NanoleafPanelColors.java
```diff
@@ -0,0 +1,148 @@
...
+import org.openhab.binding.nanoleaf.internal.handler.NanoleafPanelHandler;
+import org.openhab.core.library.types.HSBType;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Stores information about panels and their colors, while sending notifications to panels and controllers
+ * about updated states.
+ *
+ * @author Jørgen Austvik - Refactoring to own class
```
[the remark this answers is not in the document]
austvik: Thanks. It started out with more code from the PanelHandler and I didnt want to take cretdit for that, but that code was later put it in the ControllerHandler as that was doing the communication with the controller, so now it is only new code as well.

bundles/org.openhab.binding.nanoleaf/src/main/java/org/openhab/binding/nanoleaf/internal/handler/NanoleafPanelHandler.java
austvik: It would. But it is an Integer in the configuration, and it is used to look up in some other maps as an Integer, so I was thinking keeping it as Integer would save us from some unboxing/boxing?

bundles/org.openhab.binding.nanoleaf/src/main/java/org/openhab/binding/nanoleaf/internal/colors/NanoleafPanelColors.java
austvik: It could, but the map on the inside has to have Integer, so this saves some boxing in case the variable on the outside is an Integer.

bundles/org.openhab.binding.nanoleaf/src/main/java/org/openhab/binding/nanoleaf/internal/handler/NanoleafPanelHandler.java
austvik: When propving to myself that the config could only be Integer, I found NanoleafPanelConfig.  For me it looks like it is not used. Could it be that this is some historic artifact and was used before openHAB used Maps for config?

bundles/org.openhab.binding.nanoleaf/src/main/java/org/openhab/binding/nanoleaf/internal/handler/NanoleafControllerHandler.java
austvik: Good catch!

austvik:
Thanks for the thorough review, Jacob! I always learn something from them!

Stefan said he would play with it on Sunday. Hes has more devices, colors and style than me :)

bundles/org.openhab.binding.nanoleaf/src/main/java/org/openhab/binding/nanoleaf/internal/handler/NanoleafControllerHandler.java
```diff
@@ -1009,7 +1025,16 @@ private void updatePanelColors() {
...
         // panelData is in format (numPanels, (PanelId, 1, R, G, B, W, TransitionTime) * numPanel)
         @Nullable
-        Write response = gson.fromJson(panelData.getContentAsString(), Write.class);
+        Write response = null;
+
+        String panelDataContent = panelData.getContentAsString();
+        try {
+            response = gson.fromJson(panelDataContent, Write.class);
+        } catch (JsonSyntaxException jse) {
+            logger.error("Unable to parse panel data information from Nanoleaf", jse);
```
jlaur, line 1034: Unless this will stop the binding from working, log level should be reduced to at least warning.

bundles/org.openhab.binding.nanoleaf/src/main/java/org/openhab/binding/nanoleaf/internal/handler/NanoleafControllerHandler.java
```diff
@@ -919,6 +948,126 @@ private void sendRhythmCommand(Command command) throws NanoleafException {
...
+        // panelData is in format (numPanels, (PanelId, 1, R, G, B, W, TransitionTime) * numPanel)
+        @Nullable
+        Write response = null;
+
+        String panelDataContent = panelData.getContentAsString();
+        try {
+            response = gson.fromJson(panelDataContent, Write.class);
+        } catch (JsonSyntaxException jse) {
+            logger.error("Unable to parse panel data information from Nanoleaf", jse);
+            logger.trace("Panel Data which couldnt be parsed: {}", panelDataContent);
```
jlaur, line 1035:
```suggestion
            logger.trace("Panel Data which couldn't be parsed: {}", panelDataContent);
```

bundles/org.openhab.binding.nanoleaf/src/main/java/org/openhab/binding/nanoleaf/internal/handler/NanoleafPanelHandler.java
jlaur: It looks like you are right, so it could be removed. I don't know the reason for it not being used, it's still common to use a DTO for configuration.

jlaur: Since it does not cause harm, let's merge. It can be refactored later.

jlaur (approved):
LGTM, thanks!

bundles/org.openhab.binding.nanoleaf/src/main/java/org/openhab/binding/nanoleaf/internal/handler/NanoleafControllerHandler.java
```diff
@@ -919,6 +948,126 @@ private void sendRhythmCommand(Command command) throws NanoleafException {
...
+    void parsePanelData(NanoleafControllerConfig config, ContentResponse panelData) {
+        // panelData is in format (numPanels, (PanelId, 1, R, G, B, W, TransitionTime) * numPanel)
+        @Nullable
+        Write response = null;
+
+        String panelDataContent = panelData.getContentAsString();
+        try {
+            response = gson.fromJson(panelDataContent, Write.class);
+        } catch (JsonSyntaxException jse) {
+            logger.error("Unable to parse panel data information from Nanoleaf", jse);
```
jlaur, line 1034:
```suggestion
            logger.warn("Unable to parse panel data information from Nanoleaf", jse);
```

jlaur:
@austvik - I did two small commits from my phone in order to get this merged before today's milestone. DCO failure can be ignored. However, there are merge conflicts that would need to be resolved. I cannot help right now since I'm not near a computer. If you would be able to resolve within the next hour, we can merge.

austvik:
Thanks @jlaur! I resolved the conflict.

stefan-hoehn:
@austvik I wasn't able to review it before due to serious private issues. I appreciate adding this optimisation I asked you to do 🥇 

Thanks, @jlaur for taking care of the review.

Issue #11909, opened 2021-12-31
[tr064] wanBlockIPs, incorrect regular-expression for IP-address validation


syntacrsc:
When specifying ip-address for parameter "wanBlockIPs" for Bridge tr064:fritzbox:SomeName addresses such as 10.0.0.35 won't be accepted due to an invalid regular expression for vaildating the IP addresses.

## Expected Behavior
an IP adresses containing zeros (e.g. 10.0.0.51) should be accepted for wanBlockIPs.

## Current Behavior
Event log shows the following errors when trying to setup the wanBlock items:

"==> /var/log/openhab/openhab.log <==

2021-12-31 10:07:05.939 [WARN ] [hab.binding.tr064.internal.util.Util] - Removing 10.0.0.51 while processing wanBlockByIP, does not match pattern ([1-9]\d{0,2}\.){3}[1-9]\d{0,2}(\s*#.*)*, check config.

2021-12-31 10:07:05.943 [WARN ] [hab.binding.tr064.internal.util.Util] - Removing 10.0.0.36 while processing wanBlockByIP, does not match pattern ([1-9]\d{0,2}\.){3}[1-9]\d{0,2}(\s*#.*)*, check config.

2021-12-31 10:07:05.945 [WARN ] [hab.binding.tr064.internal.util.Util] - Removing 10.0.0.42 while processing wanBlockByIP, does not match pattern ([1-9]\d{0,2}\.){3}[1-9]\d{0,2}(\s*#.*)*, check config."

## Possible Solution
Adjust the regular expression for validating IP addresses for parameter "wanBlockIPs"

## Steps to Reproduce (for Bugs)
Setup a tr06 bridge and specify IP addresses for parameter wanBlockIPs containing zeros (e.g. 10.0.0.35)
example:
Bridge tr064:fritzbox:Fritz7490 "FritzBox 7490 Bridge"	@ "Office"
[
	host="10.0.0.1",
	user="tr064", 
	password="somePassword", 
	refresh=30,
	missedCallDays="7",
  	tamIndices="2",
  	outboundCallDays="7",
	inboundCallDays="7",
	rejectedCallDays="7", 
	callListDays="7",
	wanBlockIPs="10.0.0.35",	
				"10.0.0.42",	
				"10.0.0.51",	
	callDeflectionIndices="5" 

Now restart the openhab service and watch for the error messages in the event log.

## Context
[...]

Issue #11224, opened 2021-09-08
[tr064] Failed to get list from FritzBox


Ghelten:
There is still a bug in the tr064 binding. Issue  #9689 was closed 09.01.2021 but the bug is not corrected.

Some people, include me, get errors like this: 

![grafik](https://user-images.githubusercontent.com/36737211/132517859-23832bd7-dab3-4f87-a2fe-cb228a0ddf8e.png)

My Equipment: Raspbian Buster OH 3.1.0 Release Build FritzBox 7490 7.27

Can someone please check this. Thanks.
Günter

(It's my first bug report, hope i made not to many mistakes)

J-N-K:
Please provide a log at TRACE level (request before that message and result). Otherwise we can't see what went wrong.

sheilbronn:
> Please provide a log at TRACE level (request before that message and result). Otherwise we can't see what went wrong.
Sorry, I'm changing log levels for the first time: Do I only need to set the tr064 binding to log level TRACE?

Ghelten:
Log trace:

2021-09-11 04:54:32.222 [TRACE] [hab.binding.tr064.internal.util.Util] - Storing in cache org.openhab.binding.tr064.internal.dto.additions.Root@1a4b2fd
2021-09-11 04:54:32.225 [TRACE] [hab.binding.tr064.internal.util.Util] - Returning from cache: org.openhab.binding.tr064.internal.dto.additions.Root@1a4b2fd
2021-09-11 04:54:32.233 [TRACE] [ng.tr064.internal.soap.SOAPConnector] - Request: <SOAP-ENV:Envelope xmlns:SOAP-ENV="http://schemas.xmlsoap.org/soap/envelope/" SOAP-ENV:encodingStyle="http://schemas.xmlsoap.org/soap/encoding/"><SOAP-ENV:Header/><SOAP-ENV:Body><u:GetMessageList xmlns:u="urn:dslforum-org:service:X_AVM-DE_TAM:1"><NewIndex>1</NewIndex></u:GetMessageList></SOAP-ENV:Body></SOAP-ENV:Envelope>
2021-09-11 04:54:32.262 [TRACE] [ng.tr064.internal.soap.SOAPConnector] - Received response: <?xml version="1.0"?>
 <s:Envelope xmlns:s="http://schemas.xmlsoap.org/soap/envelope/" s:encodingStyle="http://schemas.xmlsoap.org/soap/encoding/">
<s:Body>
[...]

PR #13894, opened 2022-12-09
[TR064] Fix pattern excluding some valid ip-addresses

lsiepel:
Fixes https://github.com/openhab/openhab-addons/issues/11909 Regex excluding some valid ip-addresses
Fixes https://github.com/openhab/openhab-addons/issues/11224 Extend timeout and logging for some specific requests


lolodomo (approved):
Do not understand your pattern, let's assume you know what you are doing.

PR #13896, opened 2022-12-10
[ipcamera] Add a check to see if FFmpeg is frozen for mjpeg creation

Skinah:
Had a user report that a Unifi protect camera does not get detected as OFFLINE by the binding as the base station/hub/bridge stays online whilst the camera itself does a reboot. FFmpeg hangs if the stream stops midway so this fix will check if the thread is still working when it is needed. Also hard coded URLS in for newer range Instar cameras.

Reported via forum here:
https://community.openhab.org/t/ipcamera-streams-not-automatically-updating-cached/141223/15

Jar can be RIGHT HAND CLICKED and `save link as` to get around the jfrog login from:

### DIFF
--- a/bundles/org.openhab.binding.ipcamera/src/main/java/org/openhab/binding/ipcamera/internal/handler/IpCameraHandler.java
+++ b/bundles/org.openhab.binding.ipcamera/src/main/java/org/openhab/binding/ipcamera/internal/handler/IpCameraHandler.java
@@ -906,9 +906,9 @@ public class IpCameraHandler extends BaseThingHandler {
             case MJPEG:
                 if (ffmpegMjpeg == null) {
                     if (inputOptions.isEmpty()) {
-                        inputOptions = "-hide_banner -loglevel warning";
+                        inputOptions = "-hide_banner";
                     } else {
-                        inputOptions += " -hide_banner -loglevel warning";
+                        inputOptions += " -hide_banner";
                     }
                     ffmpegMjpeg = new Ffmpeg(this, format, cameraConfig.getFfmpegLocation(), inputOptions, rtspUri,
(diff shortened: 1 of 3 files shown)


bundles/org.openhab.binding.ipcamera/src/main/java/org/openhab/binding/ipcamera/internal/handler/IpCameraHandler.java
```diff
@@ -1575,6 +1575,13 @@ void pollCameraRunnable() {
                 setupFfmpegFormat(FFmpegFormat.RTSP_ALARMS);
             }
         }
+        if (ffmpegMjpeg != null) {// check if the thread has frozen due to camera doing a soft reboot
+            localFfmpeg = ffmpegMjpeg;
+            if (localFfmpeg == null || !localFfmpeg.getIsAlive()) {
```
lolodomo, line 1580: `localFfmpeg` cannot be null at this place.

Maybe rather:
```
        localFfmpeg = ffmpegMjpeg;
        if (localFfmpeg != null) {// check if the thread has frozen due to camera doing a soft reboot
            if (!localFfmpeg.getIsAlive()) {
                setupFfmpegFormat(FFmpegFormat.MJPEG);
                logger.debug("MJPEG was not being produced by FFmpeg when it should have been, restarting FFmpeg.");
            }
        }
```

lolodomo (changes requested):

bundles/org.openhab.binding.ipcamera/src/main/java/org/openhab/binding/ipcamera/internal/handler/IpCameraHandler.java
Skinah: fixed and Resolved all comments.

lolodomo (approved):
LGTM, thank you

PR #13897, opened 2022-12-10
[knx] Improve handling of serial gateways

### LINKED ISSUE #13892: [knx] FT12 serial interfaces stop working after configuration change via UI
After a config change via UI, FT12 serial interfaces stop working.
Saving the config triggers a shutdown of the thing (as expected), but restarting does not work. Comm thread is not reconnecting.

IP based KNX gateways are not affected.
Things loaded from file are also not affected.

Workaround is to disable the thing and to re-enable it afterwards (or to restart the KNX binding).

Running on OH 3.4 SNAPSHOT, KNX binding built from current main branch, openHABian on PI4.

holgerfriedrich:
Take over initialization logic from KNX IP gateway for serial gateway. 
Properly re-initialize serial gateway after configuration changes done in UI.
Fixes #13892.

### DIFF
--- a/bundles/org.openhab.binding.knx/src/main/java/org/openhab/binding/knx/internal/handler/SerialBridgeThingHandler.java
+++ b/bundles/org.openhab.binding.knx/src/main/java/org/openhab/binding/knx/internal/handler/SerialBridgeThingHandler.java
@@ -51,16 +51,13 @@ public class SerialBridgeThingHandler extends KNXBridgeBaseThingHandler {
         // create new instance using current configuration settings;
         // when a parameter change is done from UI, dispose() and initialize() are called
         SerialBridgeConfiguration config = getConfigAs(SerialBridgeConfiguration.class);
-        client = new SerialClient(config.getAutoReconnectPeriod(), thing.getUID(),
-                config.getResponseTimeout().intValue(), config.getReadingPause().intValue(),
-                config.getReadRetriesLimit().intValue(), getScheduler(), config.getSerialPort(), config.useCemi(),
-                this);
+        client = new SerialClient(config.getAutoReconnectPeriod(), thing.getUID(), config.getResponseTimeout(),
+                config.getReadingPause(), config.getReadRetriesLimit(), getScheduler(), config.getSerialPort(),
+                config.useCemi(), this);
 
         updateStatus(ThingStatus.UNKNOWN);
         // delay actual initialization, allow for longer runtime of actual initialization
-        initJob = scheduler.submit(() -> {
-            initializeLater();
-        });
+        initJob = scheduler.submit(this::initializeLater);
     }
 
     public void initializeLater() {
--- a/bundles/org.openhab.binding.knx/src/main/java/org/openhab/binding/knx/internal/handler/SerialBridgeThingHandler.java
+++ b/bundles/org.openhab.binding.knx/src/main/java/org/openhab/binding/knx/internal/handler/SerialBridgeThingHandler.java
@@ -71,7 +71,7 @@ public class SerialBridgeThingHandler extends KNXBridgeBaseThingHandler {
     public void dispose() {
         final var tmpInitJob = initJob;
         if (tmpInitJob != null) {
-            while (!tmpInitJob.isDone()) {
+            if (!tmpInitJob.isDone()) {
                 logger.trace("Bridge {}, shutdown during init, trying to cancel", thing.getUID());
                 tmpInitJob.cancel(true);
                 try {
--- a/bundles/org.openhab.binding.knx/src/main/java/org/openhab/binding/knx/internal/handler/SerialBridgeThingHandler.java
+++ b/bundles/org.openhab.binding.knx/src/main/java/org/openhab/binding/knx/internal/handler/SerialBridgeThingHandler.java
@@ -12,12 +12,18 @@
  */
 package org.openhab.binding.knx.internal.handler;
 
+import java.util.concurrent.Future;
+
 import org.eclipse.jdt.annotation.NonNullByDefault;
-import org.openhab.binding.knx.internal.client.AbstractKNXClient;
+import org.eclipse.jdt.annotation.Nullable;
+import org.openhab.binding.knx.internal.client.KNXClient;
+import org.openhab.binding.knx.internal.client.NoOpClient;
 import org.openhab.binding.knx.internal.client.SerialClient;
 import org.openhab.binding.knx.internal.config.SerialBridgeConfiguration;
 import org.openhab.core.thing.Bridge;
 import org.openhab.core.thing.ThingStatus;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /**
  * The {@link IPBridgeThingHandler} is responsible for handling commands, which are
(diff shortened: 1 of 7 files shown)


bundles/org.openhab.binding.knx/src/main/java/org/openhab/binding/knx/internal/handler/SerialBridgeThingHandler.java
```diff
@@ -31,31 +37,69 @@
...
+    }
+
+    @Override
+    public void initialize() {
+        // create new instance using current configuration settings;
+        // when a parameter change is done from UI, dispose() and initialize() are called
         SerialBridgeConfiguration config = getConfigAs(SerialBridgeConfiguration.class);
         client = new SerialClient(config.getAutoReconnectPeriod(), thing.getUID(),
                 config.getResponseTimeout().intValue(), config.getReadingPause().intValue(),
                 config.getReadRetriesLimit().intValue(), getScheduler(), config.getSerialPort(), config.useCemi(),
```
lolodomo, line 56: Could you also fix the class `BridgeConfiguration` to have `readingPause`, `readRetriesLimit` and `responseTimeout` defined with type `int` and default value set respectively to 50, 3 and 10 to be inline with the thing definitions (ip and serial) ?

bundles/org.openhab.binding.knx/src/main/java/org/openhab/binding/knx/internal/handler/SerialBridgeThingHandler.java
```diff
@@ -31,31 +37,69 @@
...
+        final var tmpInitJob = initJob;
+        if (tmpInitJob != null) {
+            while (!tmpInitJob.isDone()) {
+                logger.trace("Bridge {}, shutdown during init, trying to cancel", thing.getUID());
+                tmpInitJob.cancel(true);
+                try {
+                    Thread.sleep(1000);
+                } catch (InterruptedException e) {
+                    logger.trace("Bridge {}, cancellation interrupted", thing.getUID());
+                }
```
lolodomo, line 84: Generally, we cancel the job without waiting.
Do you really need to wait for the cancel to complete ?
By the waay, I am not fully sure that looping on isDone() is correct. To be verified.

bundles/org.openhab.binding.knx/src/main/java/org/openhab/binding/knx/internal/handler/SerialBridgeThingHandler.java
```diff
@@ -31,31 +37,69 @@
...
-    }
 
-    @Override
-    public void initialize() {
         updateStatus(ThingStatus.UNKNOWN);
-        client.initialize();
+        // delay actual initialization, allow for longer runtime of actual initialization
+        initJob = scheduler.submit(() -> {
+            initializeLater();
+        });
```
lolodomo, line 63: You can simplify like that: `initJob = scheduler.submit(this::initializeLater);`

bundles/org.openhab.binding.knx/src/main/java/org/openhab/binding/knx/internal/handler/SerialBridgeThingHandler.java
```diff
@@ -31,31 +37,69 @@
...
         SerialBridgeConfiguration config = getConfigAs(SerialBridgeConfiguration.class);
         client = new SerialClient(config.getAutoReconnectPeriod(), thing.getUID(),
                 config.getResponseTimeout().intValue(), config.getReadingPause().intValue(),
                 config.getReadRetriesLimit().intValue(), getScheduler(), config.getSerialPort(), config.useCemi(),
                 this);
-    }
 
-    @Override
-    public void initialize() {
         updateStatus(ThingStatus.UNKNOWN);
```
lolodomo, line 59: I do not see where the status is finally set to ONLINE but it may be in another class.

lolodomo (changes requested):

bundles/org.openhab.binding.knx/src/main/java/org/openhab/binding/knx/internal/handler/SerialBridgeThingHandler.java
holgerfriedrich: thanks, looks much cleaner now

bundles/org.openhab.binding.knx/src/main/java/org/openhab/binding/knx/internal/handler/SerialBridgeThingHandler.java
holgerfriedrich: yes.... ```AbstractKnxClient``` will do...

bundles/org.openhab.binding.knx/src/main/java/org/openhab/binding/knx/internal/handler/SerialBridgeThingHandler.java
holgerfriedrich: Allright, lets do some cleanup here, though it makes the patch harder to read. But agreed, I see no reason to use ```BigDecimal``` when the xml description uses ```int``` as well.
For the default values: I do not see a reason to set something in the code, 0 is fine.

lolodomo: > For the default values: I do not see a reason to set something in the code, 0 is fine.

The good practice is to align the default in class and the default in XML parameter definition.

holgerfriedrich: Hope this does not break anything. For ```autoReconnectPeriod``` I have to choose, 0 (i.e. "never") from ```serial.xml``` vs 60s from ```ip.xml```.

holgerfriedrich: @lolodomo I think I want to stick to the values were before. To make is sound, I would need to bring in the defaults in ```BridgeConfiguration.java```, ```IPBridgeConfiguration.java```, ```SerialBridgeConfiguration.java``` and ```DeviceConfig.java```. I do not want to break something shortly before the release. Let us do this in a later patch.

lolodomo: OK

bundles/org.openhab.binding.knx/src/main/java/org/openhab/binding/knx/internal/handler/SerialBridgeThingHandler.java
holgerfriedrich: Looping is probably unnecessary, as the platform docs state the following for ```cancel()```:

> "After this method returns, subsequent calls to isDone() will always return true."

But same here, I would play it safe for the release. The code is in use for IP bridge and has already some test mileage.

lolodomo:
Do not forget to sign off your last commit. 

And you missed one change, leading to a compilation error.

bundles/org.openhab.binding.knx/src/main/java/org/openhab/binding/knx/internal/handler/SerialBridgeThingHandler.java
lolodomo: This is new code and in SerialBridgeThingHandler class, so I guess it is not already used for the IP handler.
So why adding something not clean? I do not understand your argument, sorry.

lolodomo: At least do it as expected here, even if it is not done properly elsewhere.

holgerfriedrich: Ok, lets do the following: I change the while to an if. Then it should be more clear and actually do the same in case the doc is correct. It will cancel the init job if it is not yet started, and try to cancel it in case it is already running, and do nothing if it already completed. WDYT?

Just a sidenote: What I actually meant with test mileage is the hours of debugging I spent to get the SECURETUNNEL for IP working properly - without eating up all tunnel connections due to improperly closing. The code for initialization of the thing should be quite similar for IP and serial, but for the serial gateway I did not dare to change anything without having a device at hand for testing. Delayed initialization was already required due to IP secure for IP, but not yet in serial gateways. Data secure - both relevant to serial and IP gateways - is still under development (open PR, needs some more work). So actually the code I picked for the PR here was already available.

lolodomo: Why keeping a sleep of 1s just after cancelling the thread?

holgerfriedrich: To let both sides properly disconnect. Not sure if this is a hard requirement for serial, but it will not hurt. I have only one device at hand for testing.
Just think about when this code will be running: only if the thing is quickly reloaded, typically when someone saves the text config more than once.```initialize()``` schedules the initialization (which typically starts very soon after this), and the initialization is finished within a few sec.

lolodomo: @kaikreuzer : as naintainer of this binding, are you ok with this unusual code pattern in disposr() method.  If you are, you can merge the PR.

kaikreuzer: While not looking beautifully, I think with the if instead of the while, we can accept it. 😉

PR #13895, opened 2022-12-10
[airvisualnode] Add null annotations

lsiepel:
Signed-off-by: Leo Siepel <leosiepel@gmail.com>

added null annotations, some very minor refactoring.
Left one `UnusedPrivateField` warning that allready existed. It is part of json/dto and might be essential in request/response model binding.

Edit: I thik it is ready for a review. Maybe not merge it so short before 3.4.0 as it does not solve bugs (but tries to prevent them) a JAR is available here: https://1drv.ms/u/s!AnMcxmvEeupwjpdU8sz_NrUHNz7Diw?e=DN2WPo

PR #13903, opened 2022-12-10
[bigassfan] Null annotations

lsiepel:

Changes:
added null annotations, 
minor refactoring
fixed all SAT warnings

Edit: 4.0.0 JAR is available here: https://1drv.ms/u/s!AnMcxmvEeupwjp4Kbj6wM5O_dnnDew?e=z2G4qO


bundles/org.openhab.binding.bigassfan/src/main/java/org/openhab/binding/bigassfan/internal/discovery/BigAssFanDiscoveryService.java
```diff
@@ -116,20 +116,27 @@ protected void stopBackgroundDiscovery() {
...
 
     private void startListenerJob() {
-        if (listenerJob == null) {
-            terminate = false;
-            logger.debug("Starting discovery listener job in {} seconds", BACKGROUND_DISCOVERY_DELAY);
-            listenerJob = scheduledExecutorService.schedule(listenerRunnable, BACKGROUND_DISCOVERY_DELAY,
-                    TimeUnit.SECONDS);
+        logger.debug("Starting discovery listener job in {} seconds", BACKGROUND_DISCOVERY_DELAY);
+
+        synchronized (this) {
```
wborn, line 121: It would be better to synchronize on some private object instead of `this` because you never know what other logic is also synchronizing on the object instance.

If you do ever want to ever synchronize a whole method on `this` it's easier to add the `synchronized` modifier between `private void` so you do not end up with all the unncessary indentation and 2 more lines of code. :wink:

lsiepel: It was a code suggestion from fwolter or cpmeister a year ago i used for PlugwiseHA binding. So i might have to alter that too. i'll have a look at how other bindings solve this. Thank you

bundles/org.openhab.binding.bigassfan/src/main/java/org/openhab/binding/bigassfan/internal/BigAssFanHandlerFactory.java
```diff
@@ -53,7 +53,10 @@ public boolean supportsThingType(ThingTypeUID thingTypeUID) {
     protected @Nullable ThingHandler createHandler(Thing thing) {
         ThingTypeUID thingTypeUID = thing.getThingTypeUID();
         if (SUPPORTED_THING_TYPES_UIDS.contains(thingTypeUID)) {
-            return new BigAssFanHandler(thing, networkAddressService.getPrimaryIpv4HostAddress());
+            String address = networkAddressService.getPrimaryIpv4HostAddress();
+            if (address != null) {
+                return new BigAssFanHandler(thing, address);
+            }
```
jlaur, line 59: I'm not sure if good or bad, but there is a functional change here. Previously, if returned address was null, handler would still be created. This would then create a new `ConnectionManager` with address being null. Then [InetAddress.getByName](https://docs.oracle.com/javase/8/docs/api/java/net/InetAddress.html#getByName-java.lang.String-) would be called with null argument which would return a representation of the loopback interface. Just something to consider. At least there should be a warning.

You might also want to check other implementations - for example squeezebox and chromecast.

bundles/org.openhab.binding.bigassfan/src/main/java/org/openhab/binding/bigassfan/internal/discovery/BigAssFanDiscoveryService.java
```diff
@@ -115,21 +115,25 @@ protected void stopBackgroundDiscovery() {
         cancelListenerJob();
     }
 
-    private void startListenerJob() {
-        if (listenerJob == null) {
-            terminate = false;
-            logger.debug("Starting discovery listener job in {} seconds", BACKGROUND_DISCOVERY_DELAY);
-            listenerJob = scheduledExecutorService.schedule(listenerRunnable, BACKGROUND_DISCOVERY_DELAY,
+    private synchronized void startListenerJob() {
```
jlaur, line 118: Does it really need to be synchronized? I would propose to follow the common pattern of using a local variable.

bundles/org.openhab.binding.bigassfan/src/main/java/org/openhab/binding/bigassfan/internal/discovery/BigAssFanDiscoveryService.java
```diff
@@ -115,21 +115,25 @@ protected void stopBackgroundDiscovery() {
         cancelListenerJob();
     }
 
-    private void startListenerJob() {
-        if (listenerJob == null) {
-            terminate = false;
```
jlaur, line 120: Removed on purpose?

bundles/org.openhab.binding.bigassfan/src/main/java/org/openhab/binding/bigassfan/internal/discovery/BigAssFanDiscoveryService.java
```diff
@@ -115,21 +115,25 @@ protected void stopBackgroundDiscovery() {
...
         }
     }
 
-    private void cancelListenerJob() {
-        if (listenerJob != null) {
-            logger.debug("Canceling discovery listener job");
-            listenerJob.cancel(true);
-            terminate = true;
-            listenerJob = null;
+    @SuppressWarnings("null")
```
jlaur, line 127: This is potentially hiding bugs, can you use a local variable instead?

bundles/org.openhab.binding.bigassfan/src/main/java/org/openhab/binding/bigassfan/internal/discovery/BigAssFanDiscoveryService.java
```diff
@@ -115,21 +115,25 @@ protected void stopBackgroundDiscovery() {
...
+    private synchronized void cancelListenerJob() {
+        logger.debug("Canceling discovery listener job");
+
+        if (this.listenerJob != null) {
+            if (!this.listenerJob.isCancelled()) {
```
jlaur, line 132: This check is not needed.

bundles/org.openhab.binding.bigassfan/src/main/java/org/openhab/binding/bigassfan/internal/discovery/BigAssFanDiscoveryService.java
```diff
@@ -115,21 +115,25 @@ protected void stopBackgroundDiscovery() {
...
+                this.listenerJob.cancel(true);
+                terminate = true;
+                listenerJob = null;
```
jlaur, line 135:
```suggestion
                this.listenerJob = null;
```

bundles/org.openhab.binding.bigassfan/src/main/java/org/openhab/binding/bigassfan/internal/discovery/BigAssFanDiscoveryService.java
```diff
@@ -167,11 +173,11 @@ private synchronized void listen() {
                 break;
             }
         }
-        discoveryListener.shutdown();
+        localDiscoveryListener.shutdown();
         logger.debug("DiscoveryListener job is exiting");
     }
 
-    private void processMessage(BigAssFanDevice device) {
+    private void processMessage(@Nullable BigAssFanDevice device) {
```
jlaur, line 180: Perhaps it would be better to not call this method with a null argument, since the method won't do anything useful then?

bundles/org.openhab.binding.bigassfan/src/main/java/org/openhab/binding/bigassfan/internal/discovery/BigAssFanDiscoveryService.java
```diff
@@ -242,22 +248,28 @@ private synchronized void deviceDiscovered(BigAssFanDevice device) {
...
-        if (pollJob != null) {
+    @SuppressWarnings("null")
+    private synchronized void cancelPollJob() {
+        if (this.pollJob != null) {
             logger.debug("Canceling poll job");
-            pollJob.cancel(true);
+            this.pollJob.cancel(true);
             pollJob = null;
         }
     }
```
jlaur, line 275: Please see comments above.

bundles/org.openhab.binding.bigassfan/src/main/java/org/openhab/binding/bigassfan/internal/discovery/BigAssFanDiscoveryService.java
```diff
@@ -242,22 +248,28 @@ private synchronized void deviceDiscovered(BigAssFanDevice device) {
                 .withRepresentationProperty(THING_PROPERTY_MAC).withLabel(device.getLabel()).build());
     }
 
-    private void schedulePollJob() {
-        logger.debug("Scheduling discovery poll job to run every {} seconds starting in {} sec", POLL_FREQ, POLL_DELAY);
-        cancelPollJob();
```
jlaur, line 247: Is the cancellation of any existing running instance intentional?

bundles/org.openhab.binding.bigassfan/src/main/java/org/openhab/binding/bigassfan/internal/discovery/BigAssFanDiscoveryService.java
```diff
@@ -242,22 +248,28 @@ private synchronized void deviceDiscovered(BigAssFanDevice device) {
...
-        pollJob = scheduler.scheduleWithFixedDelay(() -> {
-            try {
-                discoveryListener.pollForDevices();
-            } catch (RuntimeException e) {
-                logger.warn("Poll job got unexpected exception: {}", e.getMessage(), e);
-            }
-        }, POLL_DELAY, POLL_FREQ, TimeUnit.SECONDS);
+    private synchronized void schedulePollJob() {
+        if (this.pollJob == null) {
```
jlaur, line 252: Probably you want local variable here, see other comments.

bundles/org.openhab.binding.bigassfan/src/main/java/org/openhab/binding/bigassfan/internal/discovery/DiscoveryListener.java
```diff
@@ -54,52 +60,68 @@ public DiscoveryListener() throws IOException, SocketException {
...
-        dSocket.receive(rcvPacket);
+        DatagramPacket localPacket = rcvPacket;
+        DatagramSocket localDatagramSocket = dSocket;
+
+        if (localPacket != null) {
+            localPacket.setLength(rcvBuffer.length);
+        }
+
+        if (localDatagramSocket != null && localPacket != null) {
+            localDatagramSocket.receive(rcvPacket);
```
jlaur, line 87: I assume you meant:
```suggestion
            localDatagramSocket.receive(localPacket);
```
?

bundles/org.openhab.binding.bigassfan/src/main/java/org/openhab/binding/bigassfan/internal/discovery/DiscoveryListener.java
```diff
@@ -54,52 +60,68 @@ public DiscoveryListener() throws IOException, SocketException {
...
-        device.setIpAddress(rcvPacket.getAddress().getHostAddress());
-        String message = (new String(rcvBuffer, 0, rcvPacket.getLength()));
-        device.setDiscoveryMessage(message);
-        logger.debug("RECEIVED packet of length {} from {}: {}", message.length(), device.getIpAddress(), message);
+            // Process the received packet
+            device.reset();
+
+            String address = localPacket.getAddress().getHostAddress();
+            if (address != null) {
+                device.setIpAddress(address);
```
jlaur, line 94: This is a functional change as IP address will no longer be cleared (previously null, now empty string). Intentional?

bundles/org.openhab.binding.bigassfan/src/main/java/org/openhab/binding/bigassfan/internal/discovery/DiscoveryListener.java
```diff
@@ -54,52 +60,68 @@ public DiscoveryListener() throws IOException, SocketException {
...
             return;
         }
 
         logger.debug("Sending poll request for fans: {}", POLL_MESSAGE);
         try {
-            dSocket.send(bcastPacket);
-        } catch (IOException ioe) {
-            logger.warn("IOException sending poll request for fans: {}", ioe.getMessage(), ioe);
+            localDatagramSocket.send(bcastPacket);
+        } catch (IllegalArgumentException | SecurityException | IOException ioe) {
```
jlaur, line 115: Since no longer only catching `IOException`, probably name could be more generic:
```suggestion
        } catch (IllegalArgumentException | SecurityException | IOException e) {
```

bundles/org.openhab.binding.bigassfan/src/main/java/org/openhab/binding/bigassfan/internal/handler/BigAssFanHandler.java
```diff
@@ -324,8 +330,10 @@ private void adjustMaxSpeed(PercentType command, String channelId, String comman
     private void adjustMinSpeed(PercentType command, String channelId, String commandFragment) {
         int newMax = command.intValue();
         int currentMin = PercentType.HUNDRED.intValue();
-        if (fanStateMap.get(channelId) != null) {
-            currentMin = ((PercentType) fanStateMap.get(channelId)).intValue();
+        @Nullable
```
jlaur, line 333: This is not needed here.

bundles/org.openhab.binding.bigassfan/src/main/java/org/openhab/binding/bigassfan/internal/handler/BigAssFanHandler.java
```diff
@@ -449,8 +457,10 @@ private void handleLightLevelMax(Command command) {
     private void adjustMaxLevel(PercentType command) {
         int newMin = command.intValue();
         int currentMax = PercentType.ZERO.intValue();
-        if (fanStateMap.get(CHANNEL_LIGHT_LEVEL_MAX) != null) {
-            currentMax = ((PercentType) fanStateMap.get(CHANNEL_LIGHT_LEVEL_MAX)).intValue();
+        @Nullable
```
jlaur, line 460: This is not needed here.

bundles/org.openhab.binding.bigassfan/src/main/java/org/openhab/binding/bigassfan/internal/handler/BigAssFanHandler.java
```diff
@@ -461,8 +471,10 @@ private void adjustMaxLevel(PercentType command) {
     private void adjustMinLevel(PercentType command) {
         int newMax = command.intValue();
         int currentMin = PercentType.HUNDRED.intValue();
-        if (fanStateMap.get(CHANNEL_LIGHT_LEVEL_MIN) != null) {
-            currentMin = ((PercentType) fanStateMap.get(CHANNEL_LIGHT_LEVEL_MIN)).intValue();
+        @Nullable
```
jlaur, line 474: This is not needed here.

bundles/org.openhab.binding.bigassfan/src/main/java/org/openhab/binding/bigassfan/internal/handler/BigAssFanHandler.java
```diff
@@ -589,13 +596,15 @@ public void startFanListener() {
             }
         }
 
-        public void stopFanListener() {
+        @SuppressWarnings("null")
```
jlaur, line 599: See previous comments.

bundles/org.openhab.binding.bigassfan/src/main/java/org/openhab/binding/bigassfan/internal/handler/BigAssFanHandler.java
```diff
@@ -660,7 +669,7 @@ private String readMessage() {
             return message;
         }
 
-        private void processMessage(String incomingMessage) {
+        private void processMessage(@Nullable String incomingMessage) {
```
jlaur, line 672: Consider not calling method with null argument.

bundles/org.openhab.binding.bigassfan/src/main/java/org/openhab/binding/bigassfan/internal/handler/BigAssFanHandler.java
```diff
@@ -1045,26 +1052,28 @@ protected synchronized void connect() {
...
             // Open socket
             try {
-                fanSocket = new Socket();
-                fanSocket.bind(new InetSocketAddress(ifAddress, 0));
-                fanSocket.connect(new InetSocketAddress(ipAddress, BAF_PORT), SOCKET_CONNECT_TIMEOUT);
-            } catch (IOException e) {
-                logger.debug("IOException connecting to  {} at {}: {}", thing.getUID(), ipAddress, e.getMessage());
+                localFanSocket.bind(new InetSocketAddress(ifAddress, 0));
+                localFanSocket.connect(new InetSocketAddress(ipAddress, BAF_PORT), SOCKET_CONNECT_TIMEOUT);
+                fanSocket = localFanSocket;
```
jlaur, line 1060: In case of exceptions thrown, do you want `fanSocket` to be assigned new `Socket()` or not? Previously it was assigned immediately, also in case of exceptions.

bundles/org.openhab.binding.bigassfan/src/main/java/org/openhab/binding/bigassfan/internal/handler/BigAssFanHandler.java
```diff
@@ -1081,17 +1090,20 @@ protected synchronized void disconnect() {
             logger.debug("Disconnecting from {} at {}", thing.getUID(), ipAddress);
 
             try {
-                if (fanWriter != null) {
-                    fanWriter.close();
+                DataOutputStream localFanWriter = fanWriter;
+                if (localFanWriter != null) {
+                    localFanWriter.close();
```
jlaur, line 1095: Should it still be assigned after this? Otherwise we could risk close() being called again.

bundles/org.openhab.binding.bigassfan/src/main/java/org/openhab/binding/bigassfan/internal/handler/BigAssFanHandler.java
```diff
@@ -1141,19 +1158,24 @@ private boolean isConnected() {
          * Periodically validate the command connection to the device by executing a getversion command.
          */
         private void scheduleConnectionMonitorJob() {
-            if (connectionMonitorJob == null) {
-                logger.debug("Starting connection monitor job in {} seconds for {} at {}", CONNECTION_MONITOR_DELAY,
-                        thing.getUID(), ipAddress);
-                connectionMonitorJob = scheduler.scheduleWithFixedDelay(connectionMonitorRunnable,
-                        CONNECTION_MONITOR_DELAY, CONNECTION_MONITOR_FREQ, TimeUnit.SECONDS);
+            synchronized (this) {
```
jlaur, line 1161: See previous comments.

bundles/org.openhab.binding.bigassfan/src/main/java/org/openhab/binding/bigassfan/internal/handler/BigAssFanHandler.java
```diff
@@ -1141,19 +1158,24 @@ private boolean isConnected() {
...
+                if (connectionMonitorJob == null) {
+                    logger.debug("Starting connection monitor job in {} seconds for {} at {}", CONNECTION_MONITOR_DELAY,
+                            thing.getUID(), ipAddress);
+                    connectionMonitorJob = scheduler.scheduleWithFixedDelay(connectionMonitorRunnable,
+                            CONNECTION_MONITOR_DELAY, CONNECTION_MONITOR_FREQ, TimeUnit.SECONDS);
+                }
             }
         }
 
+        @SuppressWarnings("null")
```
jlaur, line 1171: See previous comments.

bundles/org.openhab.binding.bigassfan/src/main/java/org/openhab/binding/bigassfan/internal/utils/BigAssFanConverter.java
```diff
@@ -41,6 +43,7 @@ public class BigAssFanConverter {
      * Dimmer item will produce PercentType value, which is 0-100
      * Convert that value to what the fan expects, which is 0-7
      */
+    @SuppressWarnings("null")
```
jlaur, line 46: This can be removed.

bundles/org.openhab.binding.bigassfan/src/main/java/org/openhab/binding/bigassfan/internal/utils/BigAssFanConverter.java
```diff
@@ -57,6 +60,7 @@ public static PercentType speedToPercent(String speed) {
      * Dimmer item will produce PercentType value, which is 0-100
      * Convert that value to what the light expects, which is 0-16
      */
+    @SuppressWarnings("null")
```
jlaur, line 63: This can be removed.

bundles/org.openhab.binding.bigassfan/src/main/java/org/openhab/binding/bigassfan/internal/utils/BigAssFanConverter.java
```diff
@@ -73,6 +77,7 @@ public static PercentType levelToPercent(String level) {
      * Dimmer item will produce PercentType value, which is 0-100
      * Convert that value to what the light expects, which is 2200-5000
      */
+    @SuppressWarnings("null")
```
jlaur, line 80: This can be removed.

jlaur (changes requested):
Thanks for the refactoring. I have provided some feedback.

bundles/org.openhab.binding.bigassfan/src/main/java/org/openhab/binding/bigassfan/internal/BigAssFanHandlerFactory.java
lsiepel: I think it is needed here. As checking for null (even on a local var) and if null scheduling a task might cause multiple task to be scheduled. For the other way around (safely accessing a property) i agree.
When looking for `synchronized void start` i found many bindings scheduling a task like this. There are also more bindings that have the `synchronised` block on this within the method, so there might be many more bindings.

bundles/org.openhab.binding.bigassfan/src/main/java/org/openhab/binding/bigassfan/internal/discovery/BigAssFanDiscoveryService.java
lsiepel: I think it is needed here. As checking for null (even on a local var) and if null scheduling a task might cause multiple task to be scheduled. For the other way around (safely accessing a property) i agree.
When looking for `synchronized void start` i found many bindings scheduling a task like this. There are also more bindings that have the `synchronised` block on this within the method, so there might be many more bindings.

bundles/org.openhab.binding.bigassfan/src/main/java/org/openhab/binding/bigassfan/internal/discovery/BigAssFanDiscoveryService.java
lsiepel: no, missed it

bundles/org.openhab.binding.bigassfan/src/main/java/org/openhab/binding/bigassfan/internal/discovery/BigAssFanDiscoveryService.java
lsiepel: While i'm fine with adding `this`, what is the point? I have seen some PR's last week where a contributor was asked to remove `this` prefix so is i see some individuall preferences here.

bundles/org.openhab.binding.bigassfan/src/main/java/org/openhab/binding/bigassfan/internal/discovery/BigAssFanDiscoveryService.java
lsiepel: https://github.com/openhab/openhab-addons/pull/13903#discussion_r1057289352

bundles/org.openhab.binding.bigassfan/src/main/java/org/openhab/binding/bigassfan/internal/discovery/BigAssFanDiscoveryService.java
lsiepel: I removed the active cancellation in the revised method. But now that you mention it, there might be more reasons to cancel it first and then start it. Looking from a distance this method was re-scheduling the task and not any scheduling it. So i changed the method to the old behaviour. 

On a side note, i don't think any user will ever notice this.

bundles/org.openhab.binding.bigassfan/src/main/java/org/openhab/binding/bigassfan/internal/discovery/DiscoveryListener.java
lsiepel: good catch! thanks

bundles/org.openhab.binding.bigassfan/src/main/java/org/openhab/binding/bigassfan/internal/discovery/DiscoveryListener.java
lsiepel: Not intentional, too much focused on the null case...

bundles/org.openhab.binding.bigassfan/src/main/java/org/openhab/binding/bigassfan/internal/handler/BigAssFanHandler.java
lsiepel: copy/paste

bundles/org.openhab.binding.bigassfan/src/main/java/org/openhab/binding/bigassfan/internal/handler/BigAssFanHandler.java
lsiepel: I rather leave it as is. I would need to move the nullness check from here to the caller. The same, but different.

bundles/org.openhab.binding.bigassfan/src/main/java/org/openhab/binding/bigassfan/internal/discovery/BigAssFanDiscoveryService.java
jlaur: Sorry, that was not my intention of the comment, which was about nullifying the instance variable, not the local variable. But I mixed it up and was too fast (some prefer `this.job` and `job` and others `job` and `localJob`). So please disregard that comment.

On a side note, I once had an instance variable named `host`. Naming that `localHost` was to me confusing, so I avoided that. :)

bundles/org.openhab.binding.bigassfan/src/main/java/org/openhab/binding/bigassfan/internal/handler/BigAssFanHandler.java
jlaur: The only difference is that it would be cleaner not calling a method without needed arguments in order to do nothing, since this could hide bugs. It could then even throw an [IllegalArgumentException](https://docs.oracle.com/javase/7/docs/api/java/lang/IllegalArgumentException.html) to make it fail fast and prevent introduction of bugs. However, this is out of scope here, just wanted to mention this difference.

bundles/org.openhab.binding.bigassfan/src/main/java/org/openhab/binding/bigassfan/internal/handler/BigAssFanHandler.java
lsiepel: Good one!

bundles/org.openhab.binding.bigassfan/src/main/java/org/openhab/binding/bigassfan/internal/handler/BigAssFanHandler.java
lsiepel: From the docs it is not very clear what happens when you call close to an allready closed :
_When not already closed, the close method of FilterOutputStream calls its flush method, and then calls the close method of its underlying output stream._

Similar for the `Socket.Close()` method 9 lines lower. That is also not clear. To be sure i assigned null

bundles/org.openhab.binding.bigassfan/src/main/java/org/openhab/binding/bigassfan/internal/handler/BigAssFanHandler.java
lsiepel: https://github.com/openhab/openhab-addons/pull/13903#discussion_r1057289352

bundles/org.openhab.binding.bigassfan/src/main/java/org/openhab/binding/bigassfan/internal/BigAssFanHandlerFactory.java
lsiepel: Changed it and made it similar to many other bindings. When `getPrimaryIpv4HostAddress` returns null a `warn` message is logged and the `createHandler` method is returned immediate.

jlaur: Can you mention some examples, so I could check? It's different from squeezebox and chromecast which I mentioned.

lsiepel: You're right. The bindings i looked at all have this like chomecast. They wrapped the call to `getPrimaryIpv4HostAddress` in a `callbackurl` method and pass that to the handler constructor. In this case there is no wrapper, i'll come up with a fix. Thanks.

bundles/org.openhab.binding.bigassfan/src/main/java/org/openhab/binding/bigassfan/internal/handler/BigAssFanHandler.java
jlaur: Not yet resolved, so reopening.

bundles/org.openhab.binding.bigassfan/src/main/java/org/openhab/binding/bigassfan/internal/handler/BigAssFanHandler.java
jlaur: Redundant assignment now as already assigned in line 1053.

bundles/org.openhab.binding.bigassfan/src/main/java/org/openhab/binding/bigassfan/internal/handler/BigAssFanHandler.java
```diff
@@ -1141,18 +1158,21 @@ private boolean isConnected() {
          * Periodically validate the command connection to the device by executing a getversion command.
          */
         private void scheduleConnectionMonitorJob() {
-            if (connectionMonitorJob == null) {
-                logger.debug("Starting connection monitor job in {} seconds for {} at {}", CONNECTION_MONITOR_DELAY,
-                        thing.getUID(), ipAddress);
-                connectionMonitorJob = scheduler.scheduleWithFixedDelay(connectionMonitorRunnable,
-                        CONNECTION_MONITOR_DELAY, CONNECTION_MONITOR_FREQ, TimeUnit.SECONDS);
+            synchronized (this) {
```
jlaur, line 1161: Since the entire method body is within the `synchronized` block, you could add the keyword to the method signature instead. This would also ensure exact lock on this method, avoiding blocking and potentially deadlocking if other methods would also synchronize on `this`.

bundles/org.openhab.binding.bigassfan/src/main/java/org/openhab/binding/bigassfan/internal/BigAssFanHandlerFactory.java
lsiepel: Looked at it again, i changed it back to the original functionality. Both solutions have pros/cons. just lets get this forward

bundles/org.openhab.binding.bigassfan/src/main/java/org/openhab/binding/bigassfan/internal/handler/BigAssFanHandler.java
lsiepel: Not intentionally, sorry. just fixed it.

lsiepel:
All comments have been adressed, thanks.

jlaur:
> All comments have been adressed, thanks.

So just one final confirmation - since not labelled "additional testing preferred", I assume you have given it a good spin yourself 🙂?

@mhilbush, you are code owner, would you like to have a look before merging?

lsiepel:
No i don't have this device, it is just from technicall code perspective. As it isn't that super complex, i'm confident we dont have any regressions like the bosesoundtouch. Yeah, i learned from that ;-)

Also with 4.0 release far away any regressions can be fixed on time before a release, and besides snapshots being unstable at the moment i would not count on anyone testing at the moment.  I leave it to you or @mhilbush

jlaur:
> No i don't have this device, it is just from technicall code perspective. As it isn't that super complex, i'm confident we dont have any regressions like the bosesoundtouch. Yeah, i learned from that ;-)

We usually assume that the code has been tested, so thanks for the honesty here. My general advise would be in line with @fwolter's here: https://github.com/openhab/openhab-addons/pull/14023#discussion_r1057301382.

Although I agree that the risk is low, there is a real non-zero risk of breaking something which worked fine (and was tested by many users) without null annotations. See for example https://github.com/openhab/openhab-addons/pull/13895#discussion_r1057265140. With ~200 lines changed, we both could have missed something. I recently also did this in #14010 with less confidence, even though just two lines changed, but in this case we had more to win than we had to lose.

Summing up, my general advise would be to prioritize bindings which you can either test yourself, or with backup from someone who can test. If untested, please clearly state this in the description and also use the label "Additional testing preferred". This can give a hint to reviewers to take more care, and could also trigger some coordination to possibly have someone help testing it.

Yet, here we are. I don't see any problems which were not already mentioned, but would like input from @mhilbush. If the JAR is up-to-date, perhaps we can look for a tester in previous PR's/issues for this binding or in the community?

Sorry for being overly cautious, but better safe than sorry. 😄 I don't think we need to rush it.

lsiepel:
> > No i don't have this device, it is just from technicall code perspective. As it isn't that super complex, i'm confident we dont have any regressions like the bosesoundtouch. Yeah, i learned from that ;-)
> 
> We usually assume that the code has been tested, so thanks for the honesty here. My general advise would be in line with @fwolter's here: [#14023 (comment)](https://github.com/openhab/openhab-addons/pull/14023#discussion_r1057301382).
> 
> Although I agree that the risk is low, there is a real non-zero risk of breaking something which worked fine (and was tested by many users) without null annotations. See for example [#13895 (comment)](https://github.com/openhab/openhab-addons/pull/13895#discussion_r1057265140). With ~200 lines changed, we both could have missed something. I recently also did this in #14010 with less confidence, even though just two lines changed, but in this case we had more to win than we had to lose.
> 
> Summing up, my general advise would be to prioritize bindings which you can either test yourself, or with backup from someone who can test. If untested, please clearly state this in the description and also use the label "Additional testing preferred". This can give a hint to reviewers to take more care, and could also trigger some coordination to possibly have someone help testing it.
> 
> Yet, here we are. I don't see any problems which were not already mentioned, but would like input from @mhilbush. If the JAR is up-to-date, perhaps we can look for a tester in previous PR's/issues for this binding or in the community?
> 
> Sorry for being overly cautious, but better safe than sorry. 😄 I don't think we need to rush it.

I understand your position and do take your and fwolter's advice serieus.. Couple of weeks ago i started to fix SAT issues to get to know more about openHAB's ecosystem and be of any value. And as you might have seen, after some of those PR's i stopped and switched to bugs/enhancements. While this and some PR's where allready there, i hope they are of added value and get merged and if not so be it.

mhilbush:
Yes, I'd like to look it over. And I'd like to actually run it for a while 
in my environment where I have 3 of these fans.

I'm sorry I haven't been available to review this. Lots going on in 2022 
that took much of my time. That will change in 2023.

mhilbush:
So if you could hold off on merging I would appreciate it.

mhilbush:
@jlaur Thank you for the time you invested in the review so far.

mhilbush:
Built and installed the binding, and the handlers won't start.
```
2023-01-04 13:18:55.410 [DEBUG] [inding.bigassfan.internal.handler.BigAssFanHandler] - BigAssFanHandler for bigassfan:fan:----- is initializing
2023-01-04 13:18:55.410 [DEBUG] [inding.bigassfan.internal.handler.BigAssFanHandler] - BigAssFanHandler config for bigassfan:fan:----- is BigAssFanConfig{label=Master Bedroom Fan, ipAddress=---, macAddress=---}
2023-01-04 13:18:55.410 [DEBUG] [inding.bigassfan.internal.handler.BigAssFanHandler] - BigAssFanHandler config of bigassfan:fan:--- is invalid. Check configuration
```

bundles/org.openhab.binding.bigassfan/src/main/java/org/openhab/binding/bigassfan/internal/BigAssFanConfig.java
```diff
@@ -58,16 +61,7 @@ public void setMacAddress(String macAddress) {
...
-            return false;
-        }
-        if (ipAddress == null || ipAddress.isBlank()) {
-            return false;
-        }
-        if (macAddress == null || macAddress.isBlank()) {
-            return false;
-        }
-        return true;
+        return label.isBlank() || ipAddress.isBlank() || macAddress.isBlank();
```
mhilbush, line 64: The logic here is wrong.

lsiepel: Fixed

lsiepel: I have added a 4.0.0 jar to the start post.

mhilbush:
Thanks. Handlers have started successfully.

I will continue to run for a while in my environment.

bundles/org.openhab.binding.bigassfan/src/main/java/org/openhab/binding/bigassfan/internal/BigAssFanConfig.java
```diff
@@ -58,16 +61,7 @@ public void setMacAddress(String macAddress) {
...
-            return false;
-        }
-        if (ipAddress == null || ipAddress.isBlank()) {
-            return false;
-        }
-        if (macAddress == null || macAddress.isBlank()) {
-            return false;
-        }
-        return true;
+        return !(label.isBlank() || ipAddress.isBlank() || macAddress.isBlank());
```
jlaur, line 64: Just for consideration - might be simpler to read (from left to right without two passes):
```suggestion
        return !label.isBlank() && !ipAddress.isBlank() && !macAddress.isBlank());
```

bundles/org.openhab.binding.bigassfan/src/main/java/org/openhab/binding/bigassfan/internal/BigAssFanHandlerFactory.java
```diff
@@ -37,6 +39,7 @@
 @Component(service = ThingHandlerFactory.class, configurationPid = "binding.bigassfan")
 public class BigAssFanHandlerFactory extends BaseThingHandlerFactory {
 
+    private final Logger logger = LoggerFactory.getLogger(BigAssFanHandlerFactory.class);
```
mhilbush, line 42: Why did you add a logger when the logger is not used?

bundles/org.openhab.binding.bigassfan/src/main/java/org/openhab/binding/bigassfan/internal/BigAssFanHandlerFactory.java
```diff
@@ -53,7 +56,8 @@ public boolean supportsThingType(ThingTypeUID thingTypeUID) {
     protected @Nullable ThingHandler createHandler(Thing thing) {
         ThingTypeUID thingTypeUID = thing.getThingTypeUID();
         if (SUPPORTED_THING_TYPES_UIDS.contains(thingTypeUID)) {
-            return new BigAssFanHandler(thing, networkAddressService.getPrimaryIpv4HostAddress());
+            String ipAddress = networkAddressService.getPrimaryIpv4HostAddress();
+            return new BigAssFanHandler(thing, ipAddress);
```
mhilbush, line 60: What was the point in breaking this into 2 lines?

bundles/org.openhab.binding.bigassfan/src/main/java/org/openhab/binding/bigassfan/internal/discovery/BigAssFanDiscoveryService.java
```diff
@@ -115,21 +115,23 @@ protected void stopBackgroundDiscovery() {
         cancelListenerJob();
     }
 
-    private void startListenerJob() {
-        if (listenerJob == null) {
+    private synchronized void startListenerJob() {
+        logger.debug("Starting discovery listener job in {} seconds", BACKGROUND_DISCOVERY_DELAY);
```
mhilbush, line 119: Why log this here when the discovery listener won't be started if listenerJob is null?

bundles/org.openhab.binding.bigassfan/src/main/java/org/openhab/binding/bigassfan/internal/discovery/BigAssFanDiscoveryService.java
```diff
@@ -115,21 +115,23 @@ protected void stopBackgroundDiscovery() {
...
+            this.listenerJob = scheduledExecutorService.schedule(listenerRunnable, BACKGROUND_DISCOVERY_DELAY,
                     TimeUnit.SECONDS);
         }
     }
 
     private void cancelListenerJob() {
-        if (listenerJob != null) {
-            logger.debug("Canceling discovery listener job");
-            listenerJob.cancel(true);
+        logger.debug("Canceling discovery listener job");
```
mhilbush, line 129: Why log this here when the discovery listener won't be canceled if listenerJob is not null?

bundles/org.openhab.binding.bigassfan/src/main/java/org/openhab/binding/bigassfan/internal/handler/BigAssFanHandler.java
```diff
@@ -96,18 +99,19 @@ public BigAssFanHandler(Thing thing, String ipv4Address) {
     public void initialize() {
         logger.debug("BigAssFanHandler for {} is initializing", thing.getUID());
 
-        config = getConfig().as(BigAssFanConfig.class);
-        logger.debug("BigAssFanHandler config for {} is {}", thing.getUID(), config);
+        BigAssFanConfig configuration = getConfig().as(BigAssFanConfig.class);
+        logger.debug("BigAssFanHandler config for {} is {}", thing.getUID(), configuration);
 
-        if (!config.isValid()) {
+        if (!configuration.isValid()) {
```
mhilbush, line 105: What was the point of changing the name to configuration? Was there something wrong with config?

mhilbush (commented):
I only have a few comments. Might have some more when I do a second pass.

bundles/org.openhab.binding.bigassfan/src/main/java/org/openhab/binding/bigassfan/internal/handler/BigAssFanHandler.java
lsiepel: First config was removed as it was class scope and not needed at class level. Then i introduced a local scoped var and gave it this name. Nothing intentionally, nothing wrong with config :-),  just something that happens in the workflow. But if you want config back, i'll make a commit ;-/

bundles/org.openhab.binding.bigassfan/src/main/java/org/openhab/binding/bigassfan/internal/discovery/BigAssFanDiscoveryService.java
lsiepel: Due to the introduction of a local var this line ended up on the wrong side. Fixed.

bundles/org.openhab.binding.bigassfan/src/main/java/org/openhab/binding/bigassfan/internal/discovery/BigAssFanDiscoveryService.java
lsiepel: fixed

bundles/org.openhab.binding.bigassfan/src/main/java/org/openhab/binding/bigassfan/internal/handler/BigAssFanHandler.java
mhilbush: No, just wanted to understand why it was changed.

bundles/org.openhab.binding.bigassfan/src/main/java/org/openhab/binding/bigassfan/internal/BigAssFanHandlerFactory.java
lsiepel: Same cause as the logger. Don't think it is wrong, but you're right, old was better.

bundles/org.openhab.binding.bigassfan/src/main/java/org/openhab/binding/bigassfan/internal/BigAssFanHandlerFactory.java
lsiepel: Used it for a moment, but removed it and did not remove the decleration. Fixed

mhilbush (approved):
This has been running well for me. And I don't have any other changes or comments. Therefore I'm good with merging.

LGTM

jlaur (approved):
LGTM

PR #13904, opened 2022-12-10
[jrubyscripting] Reverses order of class imports and ScriptEngine puts calls

boc-tothefuture:
This appears to fix an incompatibility introduced after M5, more details available [here](https://github.com/boc-tothefuture/openhab-jruby/pull/659).  To be honest, I am not sure why the order of these calls matter or what changed between M5 and now to make the order matter? 

Because the resolution is so unclear (at least to me) @ccutrer and @jimtng can you verify this doesn't break anything on your end? 

The other changes are from running spotless:apply

### DIFF
--- a/bundles/org.openhab.automation.jrubyscripting/src/main/java/org/openhab/automation/jrubyscripting/internal/JRubyScriptEngineFactory.java
+++ b/bundles/org.openhab.automation.jrubyscripting/src/main/java/org/openhab/automation/jrubyscripting/internal/JRubyScriptEngineFactory.java
@@ -61,7 +61,8 @@ public class JRubyScriptEngineFactory extends AbstractScriptEngineFactory {
 
     private JRubyDependencyTracker jrubyDependencyTracker;
 
-    // Adds $ in front of a set of variables so that Ruby recognizes them as global variables
+    // Adds $ in front of a set of variables so that Ruby recognizes them as global
+    // variables
     private static Map.Entry<String, Object> mapGlobalPresets(Map.Entry<String, Object> entry) {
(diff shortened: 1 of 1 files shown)


boc-tothefuture:
FYI - I did sign off on the commit, it appears to be a matching problem.

ccutrer:
LGTM. Everything loads fine in my env with this commit. And I confirmed that a bare script without any requires can now run ```ruby
logger = org.slf4j.LoggerFactory.get_logger("org.openhab.automation.rule.test.rb")

logger.warn(ON.inspect)
```

whereas before it could not

ccutrer:
I bet https://github.com/openhab/openhab-addons/commit/a68ff57c170007b07eb47ee15681e68f16befcbd broke it since M5. I _have_ noted that `$se.import_preset` doesn't work - it runs properly, and returns a map of the new constants/globals that should have been applied, but they're not actually applied. I just assumed it meant that `JRubyEngine.put` won't work being called from within a script, but it might be related to this problem.

ccutrer (approved):

ccutrer:
@boc-tothefuture: @kai just got me properly added as a co-code-owner with you (for jrubyscripting, and other addons that I'm a major contributor to) so I can give you the proper approval now. I'm still learning what this means. I was able to override the DCO, since you clearly did sign-off on the commit. But I can't merge it for you, which I'm not really surprised at (my intention at being labeled code owner was simply to have GitHub auto-assign me as a reviewer as appropriate), and I wouldn't have done anyway without asking what the proper etiquette is first. So we still need to wait for someone like @jlaur or @lolodomo to merge.

wborn (approved):
Thanks!

PR #13917, opened 2022-12-11
[Nanoleaf] Fix wrong image and website build (again)

Confectrician:
next.openhab.org website isn't building because of a wrong includeded image in nanoleaf binding based on the build errors.
A quick merge would be apprecieated.

Also we should check this a bit deeper for current nanoleaf contributions.
This is the second time i have to fix a build caused from the nanonleaf readme.
(Picture uploaded as `png` but referenced as `jpg`.)

I will try to have an eye on this too but would appreciate, if @openhab/add-ons-maintainers would also do so.


kaikreuzer (approved):
Thanks!

stefan-hoehn:
@austvik FYI please also be careful when adding pictures to the nanoleaf binding to make sure that the type is consistent with the ending.

austvik:
Thanks for fixing it, @stefan-hoehn!

stefan-hoehn:
Kudos go to @Confectrician, though <3

PR #13918, opened 2022-12-11
[knx] Minor documentation update

### LINKED ISSUE #13602: Docstring improvement
Hi,

I am the maintainer of knxd. I'd like to suggest a small but significant documentation change.

When configuring OpenHAB's knx gateway, the physical address field says "The Physical Address (Individual Address) in x.y.z notation for identification of this KNX/IP gateway within the KNX bus".

This is misleading; we're getting new users every couple of weeks who thought that the address to use here is knxd's own address, or one of the addresses configured for allocating to clients.

I'd like to suggest that replacing "this KNX/IP gateway" with "this OpenHAB instance" would go a long way towards fixing this.

NB: I assume that setting `localSourceAddr="0.0.0"` tells OpenHAB to use the address it gets from the gteway. On the other hand, if a fixed (non-zero) address is entered it **must not** be within the range knxd reserves for tunnel clients. (This also applies when you tell OpenHAB to use routing, of course) Please add these two items to the documentation.

Thank you.

holgerfriedrich:
Slightly change wording of docs related to localSourceAddress to avoid address conflicts during initial setup.
Add remark about knxd.
Fixes #13602.


holgerfriedrich:
@kaikreuzer  we got a request to slightly change our docs to avoid a pitfall new users might run into when using openHAB with knxd, see #13602.
Do you think we should update our docs as in this PR?

jlaur (approved):
Thanks!

smurfix:
One additional improvement, please:

> Or leave it at `0.0.0` and let openHAB decide which address to use.

Umm, openHAB doesn't decide which address to use. It can't, as it doesn't know which addresses **are** in use.

In router/multicast mode, the address needs to be assigned manually. `0.0.0` is unlikely to work.

In tunnel mode, `0.0.0` means that the address is provided by the gateway. In fact, some gateways will either ignore the address you provide, or ignore your packets if it doesn't match the one they sent in the tunnel establishment reply.

PR #13924, opened 2022-12-12
[jsscripting] Extend synchronization to common ScriptEngine methods

florian-h05:
## Description

This extends the multi-thread synchronization to "eval" and "invokeMethod" and moves synchronization for "invokeFunction" to the InvocationInterceptingScriptEngineWithInvocableAndAutoCloseable class.
The synchronization mechanism changed from using synchronized to using a ReentrantLock together with catch_finally to avoid having deadlocks when an exception is thrown.
Fixes the multi-thread access requested warnings described in the community (https://community.openhab.org/t/openhab-3-4-milestone-discussion/138093/130) and related to https://github.com/openhab/openhab-core/pull/3180.

## Testing

I've tested this (see https://community.openhab.org/t/openhab-3-4-milestone-discussion/138093/150), I've also verified on my production system that this does not cause any issues with rules and timers.

/cc @ccutrer @cweitkamp @J-N-K 

The pull request should be built automatically and available at:
https://openhab.jfrog.io/artifactory/libs-pullrequest-local/org/openhab/addons/bundles/org.openhab.automation.jsscripting/3.4.0-SNAPSHOT/org.openhab.automation.jsscripting-3.4.0-SNAPSHOT.jar

### DIFF
--- a/bundles/org.openhab.automation.jsscripting/src/main/java/org/openhab/automation/jsscripting/internal/GraalJSScriptEngineFactory.java
+++ b/bundles/org.openhab.automation.jsscripting/src/main/java/org/openhab/automation/jsscripting/internal/GraalJSScriptEngineFactory.java
@@ -17,6 +17,7 @@ import java.util.Map;
 
 import javax.script.ScriptEngine;
 
+import org.eclipse.jdt.annotation.NonNullByDefault;
 import org.eclipse.jdt.annotation.Nullable;
 import org.openhab.automation.jsscripting.internal.fs.watch.JSDependencyTracker;
 import org.openhab.core.automation.module.script.ScriptDependencyTracker;
@@ -37,6 +38,7 @@ import org.osgi.service.component.annotations.Reference;
 @Component(service = ScriptEngineFactory.class, configurationPid = "org.openhab.jsscripting", property = Constants.SERVICE_PID
         + "=org.openhab.jsscripting")
 @ConfigurableService(category = "automation", label = "JS Scripting", description_uri = "automation:jsscripting")
+@NonNullByDefault
 public final class GraalJSScriptEngineFactory implements ScriptEngineFactory {
     private static final String CFG_INJECTION_ENABLED = "injectionEnabled";
     private static final String INJECTION_CODE = "Object.assign(this, require('openhab'));";
@@ -80,7 +82,7 @@ public final class GraalJSScriptEngineFactory implements ScriptEngineFactory {
     }
 
     @Override
-    public ScriptEngine createScriptEngine(String scriptType) {
+    public @Nullable ScriptEngine createScriptEngine(String scriptType) {
         return new DebuggingGraalScriptEngine<>(
                 new OpenhabGraalJSScriptEngine(injectionEnabled ? INJECTION_CODE : null, jsScriptServiceUtil));
     }
--- a/bundles/org.openhab.automation.jsscripting/src/main/java/org/openhab/automation/jsscripting/internal/OpenhabGraalJSScriptEngine.java
+++ b/bundles/org.openhab.automation.jsscripting/src/main/java/org/openhab/automation/jsscripting/internal/OpenhabGraalJSScriptEngine.java
@@ -30,6 +30,8 @@ import java.time.ZonedDateTime;
 import java.util.Collections;
 import java.util.Map;
 import java.util.Set;
+import java.util.concurrent.locks.Lock;
+import java.util.concurrent.locks.ReentrantLock;
 import java.util.function.Consumer;
 import java.util.function.Function;
 
@@ -58,7 +60,8 @@ import com.oracle.truffle.js.scriptengine.GraalJSScriptEngine;
  * @author Jonathan Gilbert - Initial contribution
  * @author Dan Cunningham - Script injections
  * @author Florian Hotze - Create lock object for multi-thread synchronization; Inject the {@link JSRuntimeFeatures}
- *         into the JS context; Fix memory leak caused by HostObject by making HostAccess reference static
+ *         into the JS context; Fix memory leak caused by HostObject by making HostAccess reference static; Switch to
+ *         {@link Lock} for multi-thread synchronization
  */
 public class OpenhabGraalJSScriptEngine
         extends InvocationInterceptingScriptEngineWithInvocableAndAutoCloseable<GraalJSScriptEngine> {
@@ -84,7 +87,7 @@ public class OpenhabGraalJSScriptEngine
             .build();
 
     /** Shared lock object for synchronization of multi-thread access */
-    private final Object lock = new Object();
+    private final Lock lock = new ReentrantLock();
     private final JSRuntimeFeatures jsRuntimeFeatures;
 
     // these fields start as null because they are populated on first use
@@ -174,6 +177,8 @@ public class OpenhabGraalJSScriptEngine
 
     @Override
     protected void beforeInvocation() {
+        lock.lock();
+
         if (initialized) {
             return;
         }
@@ -227,11 +232,8 @@ public class OpenhabGraalJSScriptEngine
     }
 
     @Override
-    public Object invokeFunction(String s, Object... objects) throws ScriptException, NoSuchMethodException {
-        // Synchronize multi-thread access to avoid exceptions when reloading a script file while the script is running
-        synchronized (lock) {
-            return super.invokeFunction(s, objects);
-        }
+    protected void afterInvocation() {
+        lock.unlock();
     }
 
     @Override
--- a/bundles/org.openhab.automation.jsscripting/src/main/java/org/openhab/automation/jsscripting/internal/OpenhabGraalJSScriptEngine.java
+++ b/bundles/org.openhab.automation.jsscripting/src/main/java/org/openhab/automation/jsscripting/internal/OpenhabGraalJSScriptEngine.java
@@ -45,7 +45,7 @@ import org.openhab.automation.jsscripting.internal.fs.DelegatingFileSystem;
 import org.openhab.automation.jsscripting.internal.fs.PrefixedSeekableByteChannel;
 import org.openhab.automation.jsscripting.internal.fs.ReadOnlySeekableByteArrayChannel;
 import org.openhab.automation.jsscripting.internal.fs.watch.JSDependencyTracker;
-import org.openhab.automation.jsscripting.internal.scriptengine.InvocationInterceptingScriptEngineWithInvocableAndAutoCloseable;
+import org.openhab.automation.jsscripting.internal.scriptengine.InvocationInterceptingScriptEngineWithInvocableAndAutoCloseableAndSynchronization;
 import org.openhab.core.automation.module.script.ScriptExtensionAccessor;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -61,7 +61,7 @@ import com.oracle.truffle.js.scriptengine.GraalJSScriptEngine;
  *         into the JS context; Fix memory leak caused by HostObject by making HostAccess reference static
  */
 public class OpenhabGraalJSScriptEngine
-        extends InvocationInterceptingScriptEngineWithInvocableAndAutoCloseable<GraalJSScriptEngine> {
+        extends InvocationInterceptingScriptEngineWithInvocableAndAutoCloseableAndSynchronization<GraalJSScriptEngine> {
 
     private static final Logger LOGGER = LoggerFactory.getLogger(OpenhabGraalJSScriptEngine.class);
     private static final String GLOBAL_REQUIRE = "require(\"@jsscripting-globals\");";
@@ -83,8 +83,6 @@ public class OpenhabGraalJSScriptEngine
                     }, HostAccess.TargetMappingPrecedence.LOW)
             .build();
 
-    /** Shared lock object for synchronization of multi-thread access */
-    private final Object lock = new Object();
     private final JSRuntimeFeatures jsRuntimeFeatures;
 
     // these fields start as null because they are populated on first use
@@ -223,14 +221,6 @@ public class OpenhabGraalJSScriptEngine
             eval(globalScript);
         } catch (ScriptException e) {
             LOGGER.error("Could not inject global script", e);
-        }
-    }
-
-    @Override
-    public Object invokeFunction(String s, Object... objects) throws ScriptException, NoSuchMethodException {
-        // Synchronize multi-thread access to avoid exceptions when reloading a script file while the script is running
-        synchronized (lock) {
-            return super.invokeFunction(s, objects);
         }
     }
 
--- a/bundles/org.openhab.automation.jsscripting/src/main/java/org/openhab/automation/jsscripting/internal/OpenhabGraalJSScriptEngine.java
+++ b/bundles/org.openhab.automation.jsscripting/src/main/java/org/openhab/automation/jsscripting/internal/OpenhabGraalJSScriptEngine.java
@@ -92,7 +92,7 @@ public class OpenhabGraalJSScriptEngine
 
     // these fields start as null because they are populated on first use
     private String engineIdentifier;
-    private Consumer<String> scriptDependencyListener;
+    private @Nullable Consumer<String> scriptDependencyListener;
 
     private boolean initialized = false;
     private final String globalScript;
--- a/bundles/org.openhab.automation.jsscripting/src/main/java/org/openhab/automation/jsscripting/internal/OpenhabGraalJSScriptEngine.java
+++ b/bundles/org.openhab.automation.jsscripting/src/main/java/org/openhab/automation/jsscripting/internal/OpenhabGraalJSScriptEngine.java
@@ -239,9 +239,9 @@ public class OpenhabGraalJSScriptEngine
     }
 
     @Override
-    protected ScriptException afterThrowsInvocation(ScriptException se) {
+    protected Exception afterThrowsInvocation(Exception e) {
         lock.unlock();
-        return se;
+        return e;
     }
 
     @Override
--- a/bundles/org.openhab.automation.jsscripting/src/main/java/org/openhab/automation/jsscripting/internal/OpenhabGraalJSScriptEngine.java
+++ b/bundles/org.openhab.automation.jsscripting/src/main/java/org/openhab/automation/jsscripting/internal/OpenhabGraalJSScriptEngine.java
@@ -237,6 +237,18 @@ public class OpenhabGraalJSScriptEngine
     }
 
     @Override
+    protected Object afterInvocation(Object obj) {
+        afterInvocation();
+        return obj;
+    }
+
+    @Override
+    protected ScriptException afterThrowsInvocation(ScriptException se) {
+        lock.unlock();
+        return se;
+    }
+
+    @Override
     public void close() {
         jsRuntimeFeatures.close();
     }
--- a/bundles/org.openhab.automation.jsscripting/src/main/java/org/openhab/automation/jsscripting/internal/OpenhabGraalJSScriptEngine.java
+++ b/bundles/org.openhab.automation.jsscripting/src/main/java/org/openhab/automation/jsscripting/internal/OpenhabGraalJSScriptEngine.java
@@ -122,8 +122,9 @@ public class OpenhabGraalJSScriptEngine
                             @Override
                             public SeekableByteChannel newByteChannel(Path path, Set<? extends OpenOption> options,
                                     FileAttribute<?>... attrs) throws IOException {
-                                if (scriptDependencyListener != null) {
-                                    scriptDependencyListener.accept(path.toString());
+                                Consumer<String> localScriptDependencyListener = scriptDependencyListener;
+                                if (localScriptDependencyListener != null) {
+                                    localScriptDependencyListener.accept(path.toString());
                                 }
 
                                 if (path.toString().endsWith(".js")) {
--- a/bundles/org.openhab.automation.jsscripting/src/main/java/org/openhab/automation/jsscripting/internal/OpenhabGraalJSScriptEngine.java
+++ b/bundles/org.openhab.automation.jsscripting/src/main/java/org/openhab/automation/jsscripting/internal/OpenhabGraalJSScriptEngine.java
@@ -86,7 +86,7 @@ public class OpenhabGraalJSScriptEngine
                     }, HostAccess.TargetMappingPrecedence.LOW)
             .build();
 
-    /** Shared lock object for synchronization of multi-thread access */
+    /** {@link Lock} synchronization of multi-thread access */
     private final Lock lock = new ReentrantLock();
     private final JSRuntimeFeatures jsRuntimeFeatures;
 
@@ -232,13 +232,8 @@ public class OpenhabGraalJSScriptEngine
     }
 
     @Override
-    protected void afterInvocation() {
-        lock.unlock();
-    }
-
-    @Override
     protected Object afterInvocation(Object obj) {
-        afterInvocation();
+        lock.unlock();
         return obj;
     }
 
--- a/bundles/org.openhab.automation.jsscripting/src/main/java/org/openhab/automation/jsscripting/internal/OpenhabGraalJSScriptEngine.java
+++ b/bundles/org.openhab.automation.jsscripting/src/main/java/org/openhab/automation/jsscripting/internal/OpenhabGraalJSScriptEngine.java
@@ -45,7 +45,7 @@ import org.openhab.automation.jsscripting.internal.fs.DelegatingFileSystem;
 import org.openhab.automation.jsscripting.internal.fs.PrefixedSeekableByteChannel;
 import org.openhab.automation.jsscripting.internal.fs.ReadOnlySeekableByteArrayChannel;
 import org.openhab.automation.jsscripting.internal.fs.watch.JSDependencyTracker;
-import org.openhab.automation.jsscripting.internal.scriptengine.InvocationInterceptingScriptEngineWithInvocableAndAutoCloseableAndSynchronization;
+import org.openhab.automation.jsscripting.internal.scriptengine.InvocationInterceptingScriptEngineWithInvocableAndAutoCloseable;
 import org.openhab.core.automation.module.script.ScriptExtensionAccessor;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -61,7 +61,7 @@ import com.oracle.truffle.js.scriptengine.GraalJSScriptEngine;
  *         into the JS context; Fix memory leak caused by HostObject by making HostAccess reference static
  */
 public class OpenhabGraalJSScriptEngine
-        extends InvocationInterceptingScriptEngineWithInvocableAndAutoCloseableAndSynchronization<GraalJSScriptEngine> {
+        extends InvocationInterceptingScriptEngineWithInvocableAndAutoCloseable<GraalJSScriptEngine> {
 
     private static final Logger LOGGER = LoggerFactory.getLogger(OpenhabGraalJSScriptEngine.class);
     private static final String GLOBAL_REQUIRE = "require(\"@jsscripting-globals\");";
@@ -83,6 +83,8 @@ public class OpenhabGraalJSScriptEngine
                     }, HostAccess.TargetMappingPrecedence.LOW)
             .build();
 
+    /** Shared lock object for synchronization of multi-thread access */
+    private final Object lock = new Object();
     private final JSRuntimeFeatures jsRuntimeFeatures;
 
     // these fields start as null because they are populated on first use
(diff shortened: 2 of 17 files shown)


florian-h05:
@jpg0 Can you please have a look at this?

I know that you might not like the synchronization inside `DelegatingScriptEngineWithInvocableAndAutocloseable`, but for me this seemed to be the best place, because:
- `InvocationInterceptingScriptEngineWithInvocableAndAutoCloseable` was no option to me since it is there for
   > interception of calls, either before Invocation, or upon a ScriptException being thrown.
- `OpenhabGraalJSScriptEngine`: the place where you wanted me to put the synchronization for `invokeFunction`, but adding the overrides with the synchronization logic here makes it unnecessarily “puffed up” IMO.
- `DelegatingScriptEngineWithInvocableAndAutocloseable` is there to _Allow overriding of specific methods._, which is exactly what we need here.

bundles/org.openhab.automation.jsscripting/src/main/java/org/openhab/automation/jsscripting/internal/scriptengine/DelegatingScriptEngineWithInvocableAndAutocloseableAndSynchronization.java
```diff
@@ -26,47 +26,85 @@
...
     @Override
     public @Nullable Object eval(String s, ScriptContext scriptContext) throws ScriptException {
-        return Objects.nonNull(delegate) ? delegate.eval(s, scriptContext) : null;
+        if (Objects.nonNull(delegate)) {
+            synchronized (lock) {
+                return delegate.eval(s, scriptContext);
+            }
+        } else {
+            return null;
+        }
```
J-N-K, line 52: In general using `== null` or `!= null` is preferred over `Objects.isNull` and `Objects.nonNull`. The static methods from `Objects` should only be used when predicates in filter expressions are needed.

That aside, this type of null-check is not thread-safe because the value could be changed by another thread. Use a local copy instead:

```suggestion
        T localDelegate = delegate)
        if (localDelegate != null) {
            synchronized (lock) {
                return localDelegate.eval(s, scriptContext);
            }
        } else {
            return null;
        }
```

Otherwise LGTM, but I'm not an expert on the internals of this script engine either.

jpg0:
You are correct that `DelegatingScriptEngineWithInvocableAndAutocloseable` is there to allow overriding of methods. But it is not there to add specific functionality itself, rather to make it easy for other classes to do so.

The way these delegates work is my providing the 'template' for implementors to add the functionality. Their responsibility is providing 'hooks' for implementors to use. In this case one option would be to subclass the delegate and add your specific functionality to the subclass. Another would be to create an additional delegate type which wraps this delegate - you could even do this with a dynamic proxy if you just wanted to synchronise every method.

The reason that I am against adding it to the delegate itself is that the purpose of the delegate is to create a separation from target class (the script engine) to specific behaviour you want added in other classes. If you start putting the specific behaviour in the delegate itself, then there is no longer any separation anyway, so the purpose of the class existing is lost. 

Also if you look in the openhab-core codebase you will see many `DelegatingXXX` classes. These all behave in exactly the same way, because a delegate is a design pattern that works this way and this makes it easy for developers to immediately understand what the class does.

bundles/org.openhab.automation.jsscripting/src/main/java/org/openhab/automation/jsscripting/internal/scriptengine/DelegatingScriptEngineWithInvocableAndAutocloseableAndSynchronization.java
florian-h05: While thinking about your comment, I am wondering if we need those null checks or if we should annotate `delegate` with `@NonNull`. It is not initialized in this class, but it is in the constructor of `OpenhabGraalJSScriptEngine`. So WDYT?

florian-h05:
@jpg0 So you are proposing to add another delegate class?
What do you think is the “best” approach of the three you described?

jpg0:
Looking at it again, the best way to do this is to subclass the `InvocationInterceptingScriptEngineWithInvocableAndAutoCloseable` - this is already designed to intercept all calls. You would need to add an `afterInvocation` to release the lock though (I never did this as I didn't have a need for it, but it's a common pattern). 

The only problem is that you would need to switch to explicit use of a [ReentrantLock](https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/locks/ReentrantLock.html) rather than use of the `synchronized` keyword. TBH this is probably not a bad idea anyway as we are already going fairly deep on synchronisation here and this will give you more options.

jpg0:
(Oh and if the previous suggestion is too complex for you, I'd also support just subclassing `DelegatingScriptEngineWithInvocableAndAutocloseable` and adding the synchronisation in the subclass.)

florian-h05:
> You would need to add an afterInvocation to release the lock though

Also to `OpenhabGraalJSScriptEngine`?
So lock in `beforeInvocation` and unlock in `afterInvocation` then?

jpg0:
> Also to OpenhabGraalJSScriptEngine?

No, this doesn't need it, unless you want to put the synchronisation into it directly (and then wouldn't need a subclass).

> So lock in beforeInvocation and unlock in afterInvocation then?

Yes. You don't need to call `afterInvocation` after the catch, as long as you also release the lock in `afterThrowsInvocation`. It's more typical this way (so that `afterInvocation` is not called upon exception).

florian-h05:
I think I will put the synchronizatuon logic directly into the `OpenhabGraalJSScriptEngine` since until now synchronization was also provided inside it.

> You don't need to call afterInvocation after the catch, as long as you also release the lock in afterThrowsInvocation. It's more typical this way (so that afterInvocation is not called upon exception).

Don't I need to call `afterInvocation` in the `finally` block to unlock the lock if the try succeeds and if it fails?

florian-h05:
@jpg0 Can you please have a look at my current implementation?

@J-N-K Can you please have a look at my null annotations changes/compiler warnings reducal?

jpg0:
You have put `afterInvocation` in a finally block meaning that it's called even when an exception is thrown. This is non -standard for interception code like this, plus prevents consumers hooking non-exception invocations only. Better is to keep inside the try block so that it's not called upon exception and ensure that the implementor also overrides `afterException` to release the lock in this case.

The locking code does lgtm though and will work.

florian-h05:
Putting `afterInvocation` inside the try block wouldn’t work (if I’m wrong please correct me), since it would need to come after the return statement:

```java
    @Override
    public Object eval(String s, ScriptContext scriptContext) throws ScriptException {
        try {
            beforeInvocation();
             return super.eval(s, scriptContext);
            // afterInvocation wouldn’t work here
         } catch (ScriptException se) {
             throw afterThrowsInvocation(se);
         } finally { // Make sure that Lock is unlocked regardless of an exception is thrown or not to avoid deadlocks
             afterInvocation();
         }
     }
```

The only possible option is to store the result of `eval` in a local var, then call `afterInvocation` and then return the local var. I’m not sure if that would change the behaviour of the script engine or not, I’m no ScriptEngine expert.

jpg0:
I would do it this way:

```
@Override
    public Object eval(String s, ScriptContext scriptContext) throws ScriptException {
        try {
            beforeInvocation();
            return afterInvocation(super.eval(s, scriptContext));
         } catch (ScriptException se) {
             throw afterThrowsInvocation(se);
         }
     }
```

Then you get the added benefit that subclasses can modify or wrap the return value if they wish. (`afterInvocation` would accept and return an `Object`, and may require casting in some cases).

You can also use a local variable if you like; it shouldn't affect the delegate, regardless of whether it's a script engine or anything else.

florian-h05:
@jpg0 Can you please have a look at my commit https://github.com/openhab/openhab-addons/pull/13924/commits/7b7e01fb3f2925a2446cbbf67afd95777a23a72e?

From my side this only needs some testing and then can be merged.

J-N-K (commented):
Just one comment: According to https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/locks/Lock.html the pattern for locks should be

```
lock.lock();
try {
    // do something
} finally {
    lock.unlock();
}
```


while you use

```
try {
    lock.lock();
    // do something
} finally {
    lock.unlock();
}
```

Since we usually use the first, it would be preferable to use that.

jpg0:
> @jpg0 Can you please have a look at my commit [7b7e01f](https://github.com/openhab/openhab-addons/commit/7b7e01fb3f2925a2446cbbf67afd95777a23a72e)?

Looks good, only change would be to remove the left-over parameterless `afterInvocation`.

bundles/org.openhab.automation.jsscripting/src/main/java/org/openhab/automation/jsscripting/internal/OpenhabGraalJSScriptEngine.java
```diff
@@ -84,12 +87,12 @@
             .build();
 
     /** Shared lock object for synchronization of multi-thread access */
```
florian-h05, line 89: Reminder: Update the comment here.

florian-h05:
@J-N-K I‘ve addressed your review comment (not pushed yet).

But I found a deadlock due to a `NoSuchMethodException` being thrown. I already have a fix, but I have to keep `afterInvocation` since it universally allows us to unlock the lock without passing or returning an argument, e.g. in case we need to handle an exception that is no ScriptException. @jpg0 So I’ll keep the parameter-less afterInvocation.

I will try to finish this this evening, would be nice if we could get this merged before the next SNAPSHOT. 

@jlaur Are you available?

jpg0:
> So I’ll keep the parameter-less afterInvocation.

Don't do this, just make sure that the catch is wide enough that everything is caught: `catch Throwable`.

jlaur:
@florian-h05 - I'll be available in the evening, if you need an addon maintainer for merging. However, I don't know how much time I will have for reviewing this and also don't feel comfortable in this area, so I would suggest approval from @jpg0 and/or @J-N-K first. @kaikreuzer, keeping you in the loop also as this seems like an important fix, but probably also with some risk.

bundles/org.openhab.automation.jsscripting/src/main/java/org/openhab/automation/jsscripting/internal/scope/AbstractScriptExtensionProvider.java
```diff
@@ -41,7 +45,6 @@ protected void addType(String name, Function<String, Object> value) {
 
     @Activate
     public void activate(final BundleContext context) {
-        types = new HashMap<>();
         initializeTypes(context);
```
jlaur, line 48: Too lazy/busy to check right now: Will `initializeTypes` clear the HashMap, so it will be okay if it already has previous contents if `activate` would be called again?

bundles/org.openhab.automation.jsscripting/src/main/java/org/openhab/automation/jsscripting/internal/scriptengine/DelegatingScriptEngineWithInvocableAndAutocloseable.java
```diff
@@ -33,94 +32,90 @@
  */
 public abstract class DelegatingScriptEngineWithInvocableAndAutocloseable<T extends ScriptEngine & Invocable & AutoCloseable>
         implements ScriptEngine, Invocable, AutoCloseable {
-    protected T delegate;
+    protected @NonNull T delegate;
```
J-N-K, line 35: Why do you use `@NonNull`? Wouldn't it make sense to annotate the class with `@NonNullByDafault`? Are you running into trouble with inherited annotations? In that case ignore my comment, this can be fixed later.

bundles/org.openhab.automation.jsscripting/src/main/java/org/openhab/automation/jsscripting/internal/OpenhabGraalJSScriptEngine.java
```diff
@@ -227,11 +232,20 @@ protected void beforeInvocation() {
...
-        }
+    protected void afterInvocation() {
+        lock.unlock();
+    }
+
+    @Override
+    protected Object afterInvocation(Object obj) {
+        afterInvocation();
+        return obj;
+    }
```
J-N-K, line 243: Are you sure that the overridden methods are always the same as they are now (i.e. do nothing/return object)? Otherwise it would be good to do something like

```suggestion
    protected void afterInvocation() {
        lock.unlock();
        super.afterInvocation();
    }

    @Override
    protected Object afterInvocation(Object obj) {
        lock.unlock();
        return super.afterInvocation(obj);
    }
```

bundles/org.openhab.automation.jsscripting/src/main/java/org/openhab/automation/jsscripting/internal/scope/AbstractScriptExtensionProvider.java
J-N-K: Difficult to say. There is no class extending this class. I would vote for `types.clear();` before initializing.

bundles/org.openhab.automation.jsscripting/src/main/java/org/openhab/automation/jsscripting/internal/scriptengine/DelegatingScriptEngineWithInvocableAndAutocloseable.java
florian-h05: I tried `@NonNullByDefault` and it causes problems with inherited annotations.

florian-h05: I'll split the null annotation changes to a separate PR to focus on the more important fix here.

J-N-K: All fine with what we have. It's better than before and correct, as far as I can see.

florian-h05: Okay, then I'll keep as it is.

jlaur:
> Just one comment: According to https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/locks/Lock.html the pattern for locks should be

Perhaps because `lock` could throw an unchecked exception, in which case there would be no reason to unlock?

bundles/org.openhab.automation.jsscripting/src/main/java/org/openhab/automation/jsscripting/internal/OpenhabGraalJSScriptEngine.java
florian-h05: Note that the methods that are overriden here are newly introduced together with their overrides. I followed the pattern of `beforeInvocation` when I added `afterInvocation`: have a method at the `InvocationInterceptingXXX` class and override it in the `OpenhabGraalJSScriptEngine`.

florian-h05: The `beforeInvocation` override also doesn't call it's super method, so I'd keep as it is and resolve this comment.

florian-h05:
Commit https://github.com/openhab/openhab-addons/pull/13924/commits/e17f869c4e8b6459cfbf1510e3117a0c4745f2d2 resolves:
- https://github.com/openhab/openhab-addons/pull/13924#pullrequestreview-1215431734
- https://github.com/openhab/openhab-addons/pull/13924#issuecomment-1348523283 
- https://github.com/openhab/openhab-addons/pull/13924#issuecomment-1349254163 

Commit https://github.com/openhab/openhab-addons/pull/13924/commits/87af8e7a366dc09c13e7d770aecba219fc4af7bc fixes the deadlock I was talking about in https://github.com/openhab/openhab-addons/pull/13924#issuecomment-1349098515.

I have tested this on my dev system and also on my production system (with about 70 JS rules using timers and multiple triggers), works fine :+1: 

@J-N-K Can you please review? 
@jpg0 feel free to also have a look, but your comment about the parameter-less `afterInvocation` is addressed (I removed it as asked) and the other changes are minimal.

bundles/org.openhab.automation.jsscripting/src/main/java/org/openhab/automation/jsscripting/internal/scope/AbstractScriptExtensionProvider.java
florian-h05: Done in https://github.com/openhab/openhab-addons/commit/e17f869c4e8b6459cfbf1510e3117a0c4745f2d2.

bundles/org.openhab.automation.jsscripting/src/main/java/org/openhab/automation/jsscripting/internal/scriptengine/InvocationInterceptingScriptEngineWithInvocableAndAutoCloseable.java
```diff
@@ -116,6 +113,10 @@ public Object invokeMethod(Object o, String s, Object... objects) throws ScriptE
             return afterInvocation(super.invokeMethod(o, s, objects));
         } catch (ScriptException se) {
             throw afterThrowsInvocation(se);
+        } catch (NoSuchMethodException e) { // Make sure to unlock on a NoSuchMethodException to avoid deadlocks
```
J-N-K, line 116: As @jpg0 mentioned: if you catch `Throwable` we are sure that the lock is always removed. (Also below) Otherwise LGTM.

florian-h05: And then rethrow it? The method expects a `return` statement, so I either need to return something or throw something. If I want to rethrow it, I need to add `throws Throwable`.

J-N-K:
Either that, or catch `Exception`, I guess we‘ll have other sorts of issues if we run into an `Error`.

bundles/org.openhab.automation.jsscripting/src/main/java/org/openhab/automation/jsscripting/internal/scriptengine/InvocationInterceptingScriptEngineWithInvocableAndAutoCloseable.java
florian-h05: @J-N-K
I've fixed this in https://github.com/openhab/openhab-addons/pull/13924/commits/250601585414c50e92fc33937ebf3656fffb4ec8, but since I cannot throw a general `Exception` from these methods since that would clash with `Invocable.invokeMethod` and `Invoke.invokeFunction`, I handle all exceptions (except NPEs because SAT fails then) that can be thrown from these two methods.

And, I always need to return a value.

J-N-K: LGTM

florian-h05: Can you please then add an approving review so we can ping @jlaur for merging?

I think we don't need to wait for @jpg0 since his concerns were addressed and the other changes are exception handling which is nothing GraalJSScriptEngine specific.

florian-h05:
@jlaur Can you merge?
See https://github.com/openhab/openhab-addons/pull/13924#discussion_r1047819859 and  https://github.com/openhab/openhab-addons/pull/13924#issuecomment-1349706948.

bundles/org.openhab.automation.jsscripting/src/main/java/org/openhab/automation/jsscripting/internal/scriptengine/InvocationInterceptingScriptEngineWithInvocableAndAutoCloseable.java
jpg0: FYI the typical pattern here is to:

- Catch either any `Exception` (or `Throwable` in some cases), and allow consumers to process any of them (not just specific ones as you have done)
- If the resulting exception is not in the throws list, wrap it in an `InvocationTargetException` before rethrowing

It's not a big deal because your implementation should work, but is a bit more complex than it needs to be at the cost of preventing some consumer options (and is a non-standard interception advice pattern).

florian-h05: @jpg0
Throwing a `InvocationTargetException` doesn‘t work because this gives a compilation error:
```
Project org.openhab.automation.jsscripting: Compilation failure: Compilation failure: 
[ERROR] /home/florianh/media/data_m2/git/openhab-addons/bundles/org.openhab.automation.jsscripting/src/main/java/org/openhab/automation/jsscripting/internal/scriptengine/InvocationInterceptingScriptEngineWithInvocableAndAutoCloseable.java:[126,13] Unhandled exception type java.lang.reflect.InvocationTargetException
[ERROR] /home/florianh/media/data_m2/git/openhab-addons/bundles/org.openhab.automation.jsscripting/src/main/java/org/openhab/automation/jsscripting/internal/scriptengine/InvocationInterceptingScriptEngineWithInvocableAndAutoCloseable.java:[144,13] Unhandled exception type java.lang.reflect.InvocationTargetException
``` 

If I add `InvocationTargetException` to `throws`, this clashed with the methods from the `Invocable` interface.

@J-N-K
I can now also rethrow a NPE because I added PMD suppresions for that class. 

Now all exceptions that are thrown by the two methods from the `Invocable` interface are properly rethrown.

jpg0: Oops sorry, it's `UndeclaredThrowableException` that you want.

florian-h05: @jpg0 
This works, thanks. See https://github.com/openhab/openhab-addons/pull/13924/commits/8a6834d96c27c72d99cbdf2011ecf789f5421768. 

If you are happy with the changes, can you please add an approving review or comment so that we can merge this?

jpg0: I think that you mixed up `afterThrowsInvocation` and `afterInvocation` in the catch block? You shouldn't need to cast the return type, as `afterThrowsInvocation` should accept and return `Exception`?

florian-h05: No, I deliberately did it that way.

I use `afterThrowsInvocation` only for exceptions (i.e. `ScriptException`) that the `OpenhabGraalJSScriptEngine` should be informed about. (FYI `afterThrowsInvocation` only accepts `ScriptException` and before this PR was also only called on a `ScriptException`, but not on `NoSuchMethodException` etc.).

For the other exceptions that are not a `ScriptException`, I use `afterInvocation` because they shouldn‘t be handled by the `OpenhabGraalJSScriptEngine` because they are related to missuse of the methods of the `Invocable` interface.

jpg0: I would not advise doing it that way. You cannot enumerate all the exception types which may be thrown by the delegate, only the non `RuntimeException` / declared ones (and even that isn't guaranteed as it's only checked at compile time). Better is to allow the consumer to make the decision on what types of exceptions they care about.

Additionally the current way mixes up whether the consumer is handling an exception or a return value? There is no way to tell if an exception is being thrown or being returned.

I did not notice that the current `afterThrowsInvocation` deals solely with `ScriptException`. I don't believe that is the right way to go.

florian-h05: So what do you recommend to do?
I‘d like to get this finished and merged before the code freeze on Friday and my time is limited.

jpg0: Sorry for the back and forth... This is what I would do:

```
public Object invokeMethod(Object o, String s, Object... objects) throws ScriptException {
        try {
             return afterInvocation(super.invokeMethod(o, s, objects));
         } catch (Exception e) {
             Exception rethrow = afterThrowsInvocation(e);

             if(rethrow instanceof ScriptException) {
                 throw ((ScriptException)rethrow);
             } else if (rethrow instanceof RuntimeException) {
                  throw ((RuntimeException)rethrow);
             } else {
                  throw new UndeclaredThrowableException(rethrow);
             }
        }
}

...


public Exception afterThrowsInvocation(Exception e) { ... }
```

florian-h05: @jpg0 
See https://github.com/openhab/openhab-addons/pull/13924/commits/8a6834d96c27c72d99cbdf2011ecf789f5421768, I hope this is the final commit ;-)

jpg0: Almost :-/

You updated the exception type, but didn't make the switch from `afterInvocation` to `afterThrowsInvocation`.

florian-h05: See https://github.com/openhab/openhab-addons/pull/13924/commits/3140f02a04cd7d4862bf9e27331aad461ee53a1a.
I think I linked the wrong commit.

Maybe just have a look at the full change: https://github.com/openhab/openhab-addons/pull/13924/files.

J-N-K (approved):

bundles/org.openhab.automation.jsscripting/src/main/java/org/openhab/automation/jsscripting/internal/scriptengine/InvocationInterceptingScriptEngineWithInvocableAndAutoCloseable.java
jpg0: Ah yes, I see now. Great! There are two other issues*, but maybe we should look at them another time; I don't want to hold this up any longer and they are not things that are hard to change later.

* the are other methods which just cast the return of `afterThrowsInvocation`to `ScriptException` rather than performing the type checks. These casts may fail. Best is probably to extract the type checks into a separate method which each of these methods use to extract the correct exception type to throw.
* ideally the overriden implementation of `afterThrowsInvocation` should call `super.afterThrowsInvocation(e)` to get the exception to return, rather than using `e` directly, in case the parent class is also doing something with it.

Either way, let's get this merged!

florian-h05:
@jlaur Let‘s get this merged!
See the approval of J-N-K and jpg0‘s comment in https://github.com/openhab/openhab-addons/pull/13924#discussion_r1048777249.

bundles/org.openhab.automation.jsscripting/src/main/java/org/openhab/automation/jsscripting/internal/scriptengine/InvocationInterceptingScriptEngineWithInvocableAndAutoCloseable.java
florian-h05: I‘ve created issue #13950 as a reminder.

bundles/org.openhab.automation.jsscripting/src/main/java/org/openhab/automation/jsscripting/internal/GraalJSScriptEngineFactory.java
```diff
@@ -80,7 +82,7 @@ public void scopeValues(ScriptEngine scriptEngine, Map<String, Object> scopeValu
     }
 
     @Override
-    public ScriptEngine createScriptEngine(String scriptType) {
+    public @Nullable ScriptEngine createScriptEngine(String scriptType) {
         return new DebuggingGraalScriptEngine<>(
                 new OpenhabGraalJSScriptEngine(injectionEnabled ? INJECTION_CODE : null, jsScriptServiceUtil));
     }
```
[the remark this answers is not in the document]
florian-h05: I‘ve created issue #13950 as a reminder.

jlaur (approved):
Thanks!

PR #13925, opened 2022-12-12
[airq] Avoid subscript digits in channel label

### LINKED ISSUE #13909: [airq] naming conflic between CO and CO2 items
When creating items for the channels `CO Concentration` and `CO₂ Concentration`, in both cases a short-name `<thing>_COConcentration` is proposed. This leads to a naming conflict.

The same problem might exist for other CO₂ channels as well.

## Expected Behavior
Item `<thing>_CO2Concentration` for channel `CO₂ Concentration`
Item `<thing>_COConcentration` for channel `CO Concentration`

## Current Behavior
Item `<thing>_COConcentration` for channel `CO₂ Concentration`
Item `<thing>_COConcentration` for channel `CO Concentration`
naming conflict

## Your Environment
openHAB 3.4.0.M4

lolodomo:
Subscript digits in channel label are ignored by UI when the item name is built.

Fix #13909


Hilbrand:
I'm not sure this is the right solution. The problem is actually a UI problem. Not a binding problem, because the UI does throw away the subscript. We do specially use subscript in units, therefore it should be perfectly fine to use them in labels.

The UI problem is on this line: https://github.com/openhab/openhab-webui/blob/5052597b0da86eb18fbcbc0fe58c485c3c73ca23/bundles/org.openhab.ui/web/src/pages/settings/things/link/link-add.vue#L195
And could probably be fixed with this solution: https://codegolf.stackexchange.com/a/238478

lolodomo:
Ok

PR #13927, opened 2022-12-12
[goecharger] Consider correct channel ID (maxCurrentTemp)

### LINKED ISSUE #13891: [goecharger] API V1 can't change amx (temporary max current)
GoECharger binding API V1 can't change amx (temporary max current)

2022-12-09 17:40:00.130 [WARN ] [r.internal.handler.GoEChargerHandler] - Could not update channel maxCurrentTemp with key null and value null

Probably because 

because in thing-types.xml the channel id is defined as 

		<channels>
			<channel id="maxCurrent" typeId="current"/>
			<channel id="maxCurrentTemp" typeId="maxCurrTmp"/>

## Expected Behavior
max Current can be changed by changing the value of item linked to channel maxCurrentTemp

## Current Behavior
the max current is noit changed, only trace
Could not update channel maxCurrentTemp with key null and value null
appears in log file

## Possible Solution
GoEChargerBindingConstants
public static final String MAX_CURRENT_TEMPORARY = "maxCurrentTemporary";

should be 

public static final String MAX_CURRENT_TEMPORARY = "maxCurrentTemp";

try to change the value of item linked to channel maxCurrentTemp

lolodomo:
Fix #13891

### DIFF
--- a/bundles/org.openhab.binding.goecharger/src/main/java/org/openhab/binding/goecharger/internal/GoEChargerBindingConstants.java
+++ b/bundles/org.openhab.binding.goecharger/src/main/java/org/openhab/binding/goecharger/internal/GoEChargerBindingConstants.java
@@ -31,7 +31,7 @@ public class GoEChargerBindingConstants {
 
     // List of all Channel ids
     public static final String MAX_CURRENT = "maxCurrent";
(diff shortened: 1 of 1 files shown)


wborn (approved):
I think this will fix it!

PR #13930, opened 2022-12-12
[miele] Fix dimension for powerConsumption channel

### LINKED ISSUE #13929: [miele] Dimension for channel powerConsumption is wrong
openhab/openhab-core#3143 exposed a problem in the Miele@home binding:
- **waterConsumption:** Type `Number:Volume` and updated with unit "l". Suddenly this was shown with "m3" as unit.
- **powerConsumption:** Type `Number:Power` and updated with unit "kWh". Suddenly this was shown with "W" as unit. 

One thing was the changed default unit - which has since been fixed by openhab/openhab-core#3201. Another thing is how the unit for power consumption could end up as "W", since this is a completely different dimension. And obviously the answer is that it should have been defined as `Number:Energy`. For reference: [List of Units](https://www.openhab.org/docs/concepts/units-of-measurement.html#list-of-units).

## Expected Behavior

Channel powerConsumption is `Number:Energy`.

Ideally channel would also be renamed to energyConsumption, but this would be a (even more) breaking change.

## Current Behavior

Channel powerConsumption is `Number:Power`.

The UoM handling appears to be hiding this bug as the unit "kWh" is accepted for the channel even though it doesn't match the dimension.

https://github.com/openhab/openhab-addons/blob/e027932349dd9d0567fb121f44711be2e00c9e95/bundles/org.openhab.binding.miele/src/main/java/org/openhab/binding/miele/internal/handler/WashingMachineHandler.java#L134-L135

## Possible Solution

Change channel definition:
https://github.com/openhab/openhab-addons/blob/e027932349dd9d0567fb121f44711be2e00c9e95/bundles/org.openhab.binding.miele/src/main/resources/OH-INF/thing/channeltypes.xml#L217-L222

jlaur:
Fixes #13929

Exposed by openhab/openhab-core#3143, thus 3.4 regression as a result.

### DIFF
--- a/bundles/org.openhab.binding.miele/README.md
+++ b/bundles/org.openhab.binding.miele/README.md
@@ -131,7 +131,7 @@ Channels available for each appliance type are listed below.
 | finish              | Number:Time          | Read       | Time to finish the program running on the appliance                 |
 | door                | Contact              | Read       | Current state of the door of the appliance                          |
 | switch              | Switch               | Write      | Switch the appliance on or off                                      |
-| powerConsumption    | Number:Power         | Read       | Power consumption by the currently running program on the appliance |
+| powerConsumption    | Number:Energy        | Read       | Power consumption by the currently running program on the appliance |
 | waterConsumption    | Number:Volume        | Read       | Water consumption by the currently running program on the appliance |
 
 ##### Programs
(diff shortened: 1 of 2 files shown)


lolodomo (approved):
LGTM

lolodomo:
Do you agree that it is a breaking change to mention to users? They will have to change their item types.

jlaur:
@lolodomo - I haven't tested this in UI yet, but can have a closer look tomorrow and create a note if it is in fact breaking. Unfortunately the channel doesn't work anymore because of changes in core, see https://github.com/openhab/openhab-addons/issues/13929#issuecomment-1347316074. So even if it is breaking, the fix is needed.

jlaur:
@lolodomo - I have now verified by installing M6 (core+binding) and creating a thing + linked item for channel. With this version it was possible to update the channel with **kWh** as unit, so openhab/openhab-core#3201 fixed that problem.

After upgrading binding to the build from this PR, channel definition changed from `Number:Power` to `Number:Energy` while my linked item was still `Number:Power` and displaying the correct value and unit (**kWh**). Then after updating the channel with another value/unit, the linked item was still updated correctly (value and unit).

So my conclusion is that this is not a breaking change, but a correctness fix and can potentially also fix problems when restoring persisted values (which are without unit), when item has no state presentation defined, so that default unit for type is used.

Additionally, when new items are created from the channel they will be created correctly now. With latest core and old binding version, default unit would be **W**, so any such item update without explicit unit (**kWh**) would end up wrong. So it would be advisable to update items to `Number:Energy`, but not strictly required.

PR #13933, opened 2022-12-12
[miele] Fix examples after migration of time channels

jlaur:
Update examples in documentation - which was missing in #13841.


lolodomo (approved):
LGTM

PR #13934, opened 2022-12-12
[shieldtv] Initial Contribution - NVIDIA ShieldTV 

morph166955:
This is the initial contribution of the NVIDIA ShieldTV binding.  The intention is to mirror the NVIDIA ShieldTV App (Android version) protocol to enable IP based control over the ShieldTV.  Initial version will be limited to basic X.509 key exchange, auto discovery via mDNS on-screen PIN request process, basic key press functionality, and app control.  Future versions will include anything else we are able to uncover as we explore the (otherwise undocumented) protocol.


morph166955:
Binding tests and works on all of my shields.  
Keypress works no issue.  
Pin registration works no issue.  
App control and status works with caveat that status only seems to report after the shield receives one command (of any kind).

Auto device discovery is not implemented yet.  I may work on that as a follow-up or may commit to here depending on time.

morph166955:
My appologies to anyone who just got spammed.  Apparently updating the readme file via the web is a really bad idea...

ccutrer:
While I think I was pinged by accident because you originally pushed a massive merge that touched a lot of recent history, I'm very interested in this binding. I have 4 SHIELD TVs, that I currently control via a Ruby gem (https://github.com/ccutrer/adb_mqtt_bridge) that uses ADB and exposes them to openHAB via MQTT.

I've just glanced through your readme, and not dove into the code yet, but I have several suggestions so far:
 * don't have a separate channel for reading the currently running app and one for launching a specific app - just have an app channel whose current state is the currently running app, and when you send it a command it sends a launch command.
 * discovery will be a necessity, including the binding doing the work of generating the keys and storing them, rather than relying on the user to use OpenSSL CLI tools, and watching logs.
 * Assuming the thing has been added in the UI (it's a managed thing), is it be possible to store the certificate material in the thing's configuration?
 * Don't use a channel for retrieving the PIN from the user. It takes extra effort to link that channel to an item, and then send a command to an item you wouldn't normally put in any UI. Instead just have a configuration field on the thing that you can see when the thing configuration has changed in your handler, and then complete the handshake.

Now, for my wishlist of features that I would need/want in order to abandon my own ADB based solution. I don't use the app much, so I'm not sure how much of this would even be possible:

 * r/w: the sleep state of the device. Need to know when it turns on or off to take actions on lighting in the room or control other A/V equipment. Bonus points if I can get not just awake/asleep, but also the "dreaming" state.
 * event stream of keys pressed on the remote, not just simulate key presses. Most important would be "the user began pressing volume up/down", "the user released volume up/down", "the user pressed mute", "the user pressed the special App button (Netflix)".
 * the current state of media - stopped, paused, buffering, fast forwarding, rewinding, playing, etc. A lot of this can be captured in a PlayerItem, but some states that are possible are more advanced, and likely need to go to a StringItem.
 * metadata on what's playing. Title, album, season, etc - whatever you can get. Bonus points for current position and total media length.
 * for the current app, it'd be good to have two channels - one for the app's package name, and one for its display name.  The former for automations, the latter to show in UIs.

morph166955:
> While I think I was pinged by accident because you originally pushed a massive merge that touched a lot of recent history, I'm very interested in this binding. I have 4 SHIELD TVs, that I currently control via a Ruby gem (https://github.com/ccutrer/adb_mqtt_bridge) that uses ADB and exposes them to openHAB via MQTT.
> 
Yeah I have no idea how I did what I did just by updating the readme file.  First time for everything.

> I've just glanced through your readme, and not dove into the code yet, but I have several suggestions so far:
> 
> * don't have a separate channel for reading the currently running app and one for launching a specific app - just have an app channel whose current state is the currently running app, and when you send it a command it sends a launch command.
I could probably combine these easily enough.  I'll throw this on the to-do list.

> * discovery will be a necessity, including the binding doing the work of generating the keys and storing them, rather than relying on the user to use OpenSSL CLI tools, and watching logs.
Absolutely agreed.  My goal for this PR was to get basic functionality reviewed and committed in.  I was going to push discovery and keystore management into a second PR as I'm not sure if I'm going to have time in the next few weeks to do the code with the holiday season approaching.  It is not lost on me that the keystore bits are more advanced and need to go.

> * Assuming the thing has been added in the UI (it's a managed thing), is it be possible to store the certificate material in the thing's configuration?
> * Don't use a channel for retrieving the PIN from the user. It takes extra effort to link that channel to an item, and then send a command to an item you wouldn't normally put in any UI. Instead just have a configuration field on the thing that you can see when the thing configuration has changed in your handler, and then complete the handshake.

I attempted to do the PIN via the thing config originally.  There is a timing issue that I ran into as it's only good for 60 seconds so the user has to be quick.  Plus, the user must initiate the PIN request, it's not automatic so this would/could be 2 thing updates which takes more time.  I ended up using the openhab-cli for this and it made it very simple/fast.  Also, if there is any disconnect/reconnect, which was happening when I made a change to the thing, the PIN becomes invalid and you have to start again.  The whole thing is very flaky so I ended up pushing this to a channel for the time being.

> 
> Now, for my wishlist of features that I would need/want in order to abandon my own ADB based solution. I don't use the app much, so I'm not sure how much of this would even be possible:
> 
> * r/w: the sleep state of the device. Need to know when it turns on or off to take actions on lighting in the room or control other A/V equipment. Bonus points if I can get not just awake/asleep, but also the "dreaming" state.
Unfortunately, I've yet to uncover any level information beyond device name, current app, and a list of apps (I've not implemented this, it's a very complex stream I haven't completely deciphered yet).  The android app doesn't get this info either so I have no way to sniff it out.  I'd absolutely love to pull a variety of things from the device, but as of yet I haven't been able to find the appropriate command to provide that info.  

> * event stream of keys pressed on the remote, not just simulate key presses. Most important would be "the user began pressing volume up/down", "the user released volume up/down", "the user pressed mute", "the user pressed the special App button (Netflix)".
I also agree on this and had it on my to-do list.  So to be clear, you would like 3 buttons configured for each?  E.g.  KEY_UP, KEY_UP_PRESS, KEY_UP_RELEASE or something similar?  When doing the testing I found that the platform is VERY fast in terms of how far it goes without a button release.  When testing the fast forward I didn't release for about 2 seconds and it skipped the entire episode.

> * the current state of media - stopped, paused, buffering, fast forwarding, rewinding, playing, etc. A lot of this can be captured in a PlayerItem, but some states that are possible are more advanced, and likely need to go to a StringItem.
> * metadata on what's playing. Title, album, season, etc - whatever you can get. Bonus points for current position and total media length.
See comments above about lack of data from the shield.  Absolutely would love to have this, but at present, nothing that it's sending me is providing this data (as far as I have been able to decipher at least).

> * for the current app, it'd be good to have two channels - one for the app's package name, and one for its display name.  The former for automations, the latter to show in UIs.
I'll look at what I get back from the massive dump of app info, but I don't know if that is actually provided either.  Dicing that up is going to be a significant effort that I haven't undertaken yet.  There is a URL provided that I haven't dug into which may have some additional information about the app.

theater:
Fix your notification list... rebase first on top of master/main and then make the PR so not everyone with commits/PRs in between is involved. I would advise that you close this PR and open a new one once you make the rebase or manually remove all people with PRs in between as notified.

morph166955:
> Fix your notification list... rebase first on top of master/main and then make the PR so not everyone with commits/PRs in between is involved. I would advise that you close this PR and open a new one once you make the rebase or manually remove all people with PRs in between as notified.

From what I can tell, that's all been done already.

morph166955:
Apparently the move from 3.4.0 to 4.0.0 breaks compiling for this.  I've rebased right up to that point.

morph166955:
@ccutrer I've collapsed CURRENTAPP and STARTAPP to just APP as you suggested.  Seems to work just fine.  I've also added _PRESS and _RELEASE to the available buttons so that these can be controlled directly if the user so desires.

morph166955:
The CLI requirement for the PIN process is now removed and handled 100% by the code.  Process is as simple as sending REQUEST to the PINCODE channel and then sending the on-screen PIN now.  Data is stored directly by the binding into the UserData folder.  

To give credit where due, a few certificate functions were adapted from the Bosch binding as it had a similar need for a self-signed certificate.

morph166955:
> > @ccutrer I've collapsed CURRENTAPP and STARTAPP to just APP as you suggested. Seems to work just fine. I've also added _PRESS and _RELEASE to the available buttons so that these can be controlled directly if the user so desires.
> > Community Marketplace updated to match this commit
> 
> Very nice.
> 
> But note that for keydown and keyup events, I'm not talking about wanting to _send_ them (though I suppose that's nice) - I want to be able to _receive_ them. CEC has never been very reliable for me, in a multitude of combinations. So I take the key events _from_ my SHIELD, process them with rules, and then send commands out another binding to my receiver or TV as appropriate. Having keydown and keyup events allows me to implement that if volume down is held down, adjust the volume another step every 0.25s or so.

Let's talk offline about how we can try to uncover that.  I understand what you're looking for, and we can probably/maybe/hopefully dig through the packets that are coming in from the shield to figure that out.  That definitely is something that will take exploration and we probably want to defer to add after we get the basic binding reviewed and committed.

morph166955:
Community marketplace now updated to include certificate commits above.

ccutrer:
> Let's talk offline about how we can try to uncover that.  I understand what you're looking for, and we can probably/maybe/hopefully dig through the packets that are coming in from the shield to figure that out.  That definitely is something that will take exploration and we probably want to defer to add after we get the basic binding reviewed and committed. 

Agreed. That's definitely a beyond-minimum feature. And may not be possible using strictly nVidia's protocol which is exactly what this binding is meant to represent.

mlobstein:
> I attempted to do the PIN via the thing config originally. There is a timing issue that I ran into as it's only good for 60 seconds so the user has to be quick. Plus, the user must initiate the PIN request, it's not automatic so this would/could be 2 thing updates which takes more time.

If you have not looked into it already, the Vizio binding implements a PIN request/response prompt routine using the openHAB console. You might want to just re-use that logic instead of doing it via a channel.

morph166955:
> > I attempted to do the PIN via the thing config originally. There is a timing issue that I ran into as it's only good for 60 seconds so the user has to be quick. Plus, the user must initiate the PIN request, it's not automatic so this would/could be 2 thing updates which takes more time.
> 
> If you have not looked into it already, the Vizio binding implements a PIN request/response prompt routine using the openHAB console. You might want to just re-use that logic instead of doing it via a channel.

I looked at a few bindings that did similar.  The shield is VERY flaky with this process.  With the recent commit that resolves all of the user cli certificate complexity I personally don't have any issue with this being a channel.  I just pushed the new jar to my production setup and it was very easy to just send two commands to the channel and have everything else work in the backend.  I think for the time being, unless this is a significant issue for the reviewers, I'd like to leave it as a channel.

bundles/org.openhab.binding.shieldtv/src/main/java/org/openhab/binding/shieldtv/internal/ShieldTVHandler.java
```diff
@@ -0,0 +1,790 @@
...
+    private ShieldTVPKI shieldtvPKI = new ShieldTVPKI();
+
+    public ShieldTVHandler(Thing thing) {
+        super(thing);
+        shieldtvMessageParser = new ShieldTVMessageParser(this);
+    }
+
+    public void setHostName(String hostName) {
+        this.hostName = hostName;
+        updateState(CHANNEL_DEVICENAME, new StringType(hostName));
```
mlobstein, line 127: Since it is just static metadata, device name should be a Thing Property instead of stored in a channel that will never change.

```suggestion
        thing.setProperty("devicename", hostname);
```

morph166955: I had created this as a channel with the intention of having other channels (eventually) be presented to the user with other info (app for example) as it was uncovered in the protocol.   Unfortunately, decoding the protocol has been slow going.  This also could potentially change if the user makes a change on the device.

mlobstein: Thing Properties are presented to the user on the thing config page. They can be updated dynamically. The hostname is not going to change unless the user re-configures the device  and is not useful in rules so putting it in a channel seems like the wrong approach.

image

morph166955: Sounds good.  I'll make the change.

morph166955:
Updated my build environment to support 4.0.0.  Test build works on my side, pushing this to 4.0.0-SNAPSHOT.

bundles/org.openhab.binding.shieldtv/src/main/java/org/openhab/binding/shieldtv/internal/ShieldTVHandler.java
morph166955: Done and committed.

mlobstein: I think you need to add it to the thing definition xml as well:
image

morph166955: You are correct.  Done and committed.

morph166955:
This is ready for review IMHO.  I'm working on mdns discovery, and if I can get it all figured out prior to review I'll add it, but otherwise I don't have anything on the roadmap to contribute at this time (other than fixes from reviews).

morph166955:
And if someone can tell me what I'm doing wrong on mdns discovery I'd greatly appreciate it. From what I can tell it's not finding the binding as a participant. 

https://github.com/morph166955/openhab-addons/blob/shieldtv-mdns/bundles/org.openhab.binding.shieldtv/src/main/java/org/openhab/binding/shieldtv/internal/discovery/ShieldTVDiscoveryParticipant.java

mlobstein:
> And if someone can tell me what I'm doing wrong on mdns discovery I'd greatly appreciate it. From what I can tell it's not finding the binding as a participant.

Your createResult() method always returns null. It needs to return a DiscoveryResult. Take a look at the vizio binding for an example.

morph166955:
> > And if someone can tell me what I'm doing wrong on mdns discovery I'd greatly appreciate it. From what I can tell it's not finding the binding as a participant.
> 
> Your createResult() method always returns null. It needs to return a DiscoveryResult. Take a look at the vizio binding for an example.

That was a byproduct of my debugging.  I removed a large chunk of the createResult() code to make sure it wasn't causing the issue.  I added it back in just now and retested.  Nada.

I'm 99% sure I'm not even getting close to createResult() firing.  From what I can tell, addMDNSDiscoveryParticipant() isn't firing at all.  It looks like the mdns discovery side of openhab-core isn't seeing my MDNSDiscoveryParticipant.

mlobstein:
> I'm 99% sure I'm not even getting close to createResult() firing. From what I can tell, addMDNSDiscoveryParticipant() isn't firing at all. It looks like the mdns discovery side of openhab-core isn't seeing my MDNSDiscoveryParticipant.

I think you need to add this to feature.xml:
`<feature>openhab-transport-mdns</feature>`

morph166955:
> > I'm 99% sure I'm not even getting close to createResult() firing. From what I can tell, addMDNSDiscoveryParticipant() isn't firing at all. It looks like the mdns discovery side of openhab-core isn't seeing my MDNSDiscoveryParticipant.
> 
> I think you need to add this to feature.xml: `<feature>openhab-transport-mdns</feature>`

Problem was something stuck in my build environment, code was fine.  I'm building cleanly now.  Thanks to @mhilbush for the sanity check!  I'll be sorting through the mdns today and hopefully commit the code later on.  Thanks again!

morph166955:
Discovery tested and committed.  Jar available at https://github.com/morph166955/openhab-addons/releases/download/shieldtv-beta-3/org.openhab.binding.shieldtv-4.0.0-SNAPSHOT-abfcf0e.jar

I have nothing else to submit to this minus any changes from reviews at this time.

morph166955:
@mhilbush noted that the dispose method has gone AWOL from the handler.  Not sure when it got lost but I'll add it back in when I do the commit for whatever the reviewers note. Noting here so I don't forget.

morph166955:
Rebased to include 2023 copyright and re-added dispose() back to the handler.

morph166955:
As this is still pending review, I've pulled the app support upgrades in from my beta branch so they can be included.

bundles/org.openhab.binding.shieldtv/src/main/resources/OH-INF/binding/binding.xml
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<binding:binding id="shieldtv" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+	xmlns:binding="https://openhab.org/schemas/binding/v1.0.0"
+	xsi:schemaLocation="https://openhab.org/schemas/binding/v1.0.0 https://openhab.org/schemas/binding-1.0.0.xsd">
+
+	<name>ShieldTV Binding</name>
+	<description>This is the binding for ShieldTV.</description>
+
+</binding:binding>
```
J-N-K, line 9: Please convert to the new `addon.xml`.

morph166955: Will do. I'll push it with some other things this week.

bundles/org.openhab.binding.shieldtv/src/main/java/org/openhab/binding/shieldtv/internal/ShieldTVHandler.java
```diff
@@ -0,0 +1,832 @@
...
+                    "Error opening SSL connection. Check log.");
+            logger.info("Error opening SSL connection: {}", e.getMessage());
+            disconnect(false);
+            scheduleConnectRetry(reconnectInterval); // Possibly a temporary problem. Try again later.
+            return;
+        }
+
+        updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.NONE, STATUS_INITIALIZING);
+
+        Thread readerThread = new Thread(this::readerThreadJob, "ShiledTV reader");
```
clinique, line 325: Typo on "`Shiled`"

bundles/org.openhab.binding.shieldtv/src/main/java/org/openhab/binding/shieldtv/internal/ShieldTVHandler.java
```diff
@@ -0,0 +1,832 @@
...
+        this.senderThread = senderThread;
+
+        logger.debug("Starting keepalive job with interval {}", heartbeatInterval);
+        keepAliveJob = scheduler.scheduleWithFixedDelay(this::sendKeepAlive, heartbeatInterval, heartbeatInterval,
+                TimeUnit.SECONDS);
+
+        String login = ShieldTVRequest.encodeMessage(ShieldTVRequest.loginRequest());
+        sendCommand(new ShieldTVCommand(login));
+
+        // updateStatus(ThingStatus.ONLINE);
```
clinique, line 342: Leftover comment.

bundles/org.openhab.binding.shieldtv/src/main/resources/OH-INF/i18n/shieldtv.properties
```diff
@@ -0,0 +1,3 @@
+# FIXME: please add all English translations to this file so the texts can be translated using Crowdin
+# FIXME: to generate the content of this file run: mvn i18n:generate-default-translations
```
clinique, line 2: Don't forget this one.

clinique (changes requested):
Minor observations

morph166955:
I appreciate all of the comments/review/feedback.  In an effort to have this binding support a larger part of the community I've decided to re-spin this as an AndroidTV binding and not just a ShieldTV binding.  Part of this will be the incorporation of additional protocols (like the GoogleTV v2 protocol).  As such, I'm making some significant changes to the layout, naming conventions, etc. which will make this very complicated to try to merge into this PR.  I'll be closing this PR and opening a new one in short order for an AndroidTV binding so that it's merged cleaner.  I will make sure to make the changes from the reviews here as well, they all seem relevant to the code still.  Thank you!

morph166955:
Just for note, all review comments have been fixed and merged into androidtv

PR #13946, opened 2022-12-13
[jrubyscripting] Allow choosing among multiple major versions of jruby

ccutrer:
since you may want to use the latest. conversely, don't default to latest, since you may be using gems not compatible with ruby 3.x, since it's such a big jump

### DIFF
--- a/bundles/org.openhab.automation.jrubyscripting/pom.xml
+++ b/bundles/org.openhab.automation.jrubyscripting/pom.xml
@@ -19,13 +19,51 @@
     <jruby.version>9.3.9.0</jruby.version>
   </properties>
 
-  <dependencies>
-    <dependency>
-      <groupId>org.jruby</groupId>
-      <artifactId>jruby-complete</artifactId>
-      <version>${jruby.version}</version>
-      <scope>compile</scope>
-    </dependency>
-  </dependencies>
-
+  <build>
+    <resources>
+      <resource>
+        <directory>${project.build.directory}/jrubies</directory>
+        <includes>
+          <include>*.jar</include>
+        </includes>
+      </resource>
+      <resource>
+        <directory>src/main/resources</directory>
+      </resource>
+    </resources>
+    <plugins>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-dependency-plugin</artifactId>
+        <version>3.4.0</version>
+        <executions>
+          <execution>
+            <id>copy</id>
+            <phase>validate</phase>
+            <goals>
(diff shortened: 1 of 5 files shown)


ccutrer:
this is working, but seems to be _much_ slower at loading ... basically everything ruby related. I suspect it's because there's now an extra level of unzipping with jars-within-jars. we could probably get around that my extracting ruby-complete to a temp directory before loading it up.

ccutrer:
The only thing I confirmed is that the standard compile dependency in the pom only supports a single version.

That said, is there anything in this that feels particularly fragile? I'm super familiar with the class loader stuff because I had to fix some JRuby bugs around classloaders and their interactions with OSGi for my openHAB/JRuby helper library test framework. This is exactly how OSGi manages bundles internally - by having a bajillion classloaders hanging around.

As for the small amount of reflection to get JRuby's version number - same thing - this is how JRuby works internally interfacing with Java. It's only more "fragile" in the same way that you don't have a compile time check that the objects are there. Similar to having or not having @NonNull annotations. And it still has to catch runtime exceptions in case the objects _aren't_ there (which I've chosen to simply swallow and ignore the substitution in the gem home path if somehow these assumptions break in a future version of JRuby. Of course, the list of supported versions is still static in this PR so far, so it's a fairly safe assumption).

That said, dynamic OSGi imports sound like an interesting avenue. And maybe could even get us to the point of not having to package all the jruby versions inside the addon's jar, _and_ allow the addon to pull in new jruby versions as soon as they're released without being statically mentioned. Unfortunately afaik jruby-complete doesn't have the necessary metadata to be a direct OSGi bundle. And the last time I tried to wrap a vanilla jar as an OSGi bundle (for using Jinja in the mqtt.homeassistant binding without directly depending on the jinja transformation service), I crashed and burned hard, and haven't gotten back to it. That'd definitely be awesome if you wanted to investigate this avenue.

ccutrer:
👍 

I'm definitely open to more ideas. This was just a quick proof-of-concept, and could definitely be matured (this its status as a draft PR). Feel free to explore it further and even push up additional commits to it. I mainly want to use it at some point to A/B 9.3.8 vs. 9.3.9 to try and determine a root cause for the scopeValues issue you worked around. But that's low priority for me. Don't expect anything on that investigation or on this PR from me anytime soon.

jimtng:
This can probably be closed, @ccutrer?

ccutrer:
Yeah. This is definitely less important since we made the jump to 9.4 anyway. And I don’t see me getting back to it anytime soon. I know where to find it if it comes up again :).

PR #13951, opened 2022-12-14
[nanoleaf] Bugfix: Handle non-integer panel ids

austvik:
Panel ids are sometimes returned as BigInteger

We haven't been able to understand why this happens somewhere and somewhere not, but this is an sledgehammer attempt to fix it quickly to unblock users, and then we will try to understand it later.

Discussions:
https://community.openhab.org/t/java-lang-classcastexception-class-java-math-bigdecimal-cannot-be-cast-to-class-java-lang-integer/142035/16 https://community.openhab.org/t/nanoleaf-binding-oh3-stabilization-update/116300/61

### DIFF
--- a/bundles/org.openhab.binding.nanoleaf/src/main/java/org/openhab/binding/nanoleaf/internal/handler/NanoleafPanelHandler.java
+++ b/bundles/org.openhab.binding.nanoleaf/src/main/java/org/openhab/binding/nanoleaf/internal/handler/NanoleafPanelHandler.java
@@ -286,7 +286,21 @@ public class NanoleafPanelHandler extends BaseThingHandler implements NanoleafPa
     }
 
     public Integer getPanelID() {
-        return (Integer) getThing().getConfiguration().get(CONFIG_PANEL_ID);
+        Object panelId = getThing().getConfiguration().get(CONFIG_PANEL_ID);
+        if (panelId instanceof Integer) {
+            return (Integer) panelId;
+        } else if (panelId instanceof Number) {
+            return ((Number) panelId).intValue();
+        } else {
+            // Fall back to parsing string representation of panel if it is not returning an integer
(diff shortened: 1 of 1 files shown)


austvik:
@jlaur This is a break I have caused in M6, and potentially a backport?

jlaur:
Bug, but removing label again to not have it mentioned in release notes as it was introduced with M6, thus will not make sense when reading all 3.4 changes.

austvik:
Thanks @jlaur! Stefan mentioned backport here: https://community.openhab.org/t/java-lang-classcastexception-class-java-math-bigdecimal-cannot-be-cast-to-class-java-lang-integer/142035/12

I am not sure if it is done, or if it qualifies, or if I should do something more for it potentially.

stefan-hoehn:
thanks @jlaur for merging that so quickly . 🥇 

 I noticed yesterday that all panel ids are loaded as Big Decimal. It is not yet clear why this is the case.  However the fix is at least a band aid for that. 

Will this become part of the current M6 or only with the upcoming final release?

jlaur:
@stefan-hoehn - M6 was released on Sunday, so the fix will be included with the release version of 3.4 since there are no more planned milestones.

jlaur:
@stefan-hoehn - you could also provide them with a JAR, or they could use the snapshot produced after midnight.

jlaur:
@austvik - about that backport, the link refers to a private page?

stefan-hoehn:
Yes, that was a private conversation between me and Jorgen that he linked here.

jlaur:
> M6 was released on Sunday, so the fix will be included with the release version of 3.4 since there are no more planned milestones.

Actually not entirely correct, there will be a release candidate today in the evening, so (hopefully) last version before the final release.

PR #13952, opened 2022-12-14
[jsscripting] Fix broken event object table

wborn:
The table is not properly rendered when reading the documentation in Main UI.


lolodomo (approved):
LGTM

PR #13955, opened 2022-12-15
[rrd4j] Error handling for broken rrd4j files

holgerfriedrich:
Catch exceptions thrown by getDB(..) and print the name of the affected database file. This allows to identify a broken rrd4j file.

Signed-off-by: Holger Friedrich <mail@holger-friedrich.de>

During testing of the M6 milestone I got stacktraces like this in my log:
```
[WARN ] [ore.internal.scheduler.SchedulerImpl] - Scheduled job '<unknown>' failed and stopped
java.lang.IndexOutOfBoundsException: null
        at java.nio.Buffer.checkIndex(Buffer.java:693) ~[?:?]
        at java.nio.DirectByteBuffer.getInt(DirectByteBuffer.java:758) ~[?:?]
        at org.rrd4j.core.ByteBufferBackend.readInt(ByteBufferBackend.java:130) ~[?:?]
        at org.rrd4j.core.RrdPrimitive.readInt(RrdPrimitive.java:38) ~[?:?]
        at org.rrd4j.core.RrdInt.get(RrdInt.java:35) ~[?:?]
        at org.rrd4j.core.Archive.<init>(Archive.java:45) ~[?:?]
        at org.rrd4j.core.RrdDb.<init>(RrdDb.java:646) ~[?:?]
        at org.rrd4j.core.RrdDb.<init>(RrdDb.java:44) ~[?:?]
        at org.rrd4j.core.RrdDb$Builder.build(RrdDb.java:94) ~[?:?]
        at org.rrd4j.core.RrdDbPool.requestRrdDb(RrdDbPool.java:420) ~[?:?]
        at org.rrd4j.core.RrdDb$Builder.build(RrdDb.java:92) ~[?:?]
        at org.openhab.persistence.rrd4j.internal.RRD4jPersistenceService.getDB(RRD4jPersistenceService.java:349) ~[?:?]
        at org.openhab.persistence.rrd4j.internal.RRD4jPersistenceService.store(RRD4jPersistenceService.java:155) ~[?:?]
        at org.openhab.core.persistence.internal.PersistItemsJob.run(PersistItemsJob.java:60) ~[?:?]
        at org.openhab.core.internal.scheduler.CronSchedulerImpl.lambda$0(CronSchedulerImpl.java:62) ~[?:?]
        at org.openhab.core.internal.scheduler.CronSchedulerImpl.lambda$1(CronSchedulerImpl.java:69) ~[?:?]
[...]

PR #13956, opened 2022-12-15
[rrd4j] Upgrade base library from 3.8.1 to 3.8.2

holgerfriedrich:
RRD4J base library was updated.

See changes here:
https://github.com/rrd4j/rrd4j/compare/3.8.1...3.8.2

Typically I would not recommend an update shortly before a planned release, but in this case there might be a reason to do it: The interesting part is the update of log4j from 2.17.0 to 2.19.0.
@openhab/add-ons-maintainers Can you pls check if we have a log4j problem here? Log4j 2.7.0 used by the older rrd4j 3.8.1 is affected by the known log4j issue, see https://logging.apache.org/log4j/2.x/security.html

### DIFF
--- a/bundles/org.openhab.persistence.rrd4j/pom.xml
+++ b/bundles/org.openhab.persistence.rrd4j/pom.xml
@@ -23,7 +23,7 @@
     <dependency>
       <groupId>org.rrd4j</groupId>
       <artifactId>rrd4j</artifactId>
-      <version>3.8.1</version>
+      <version>3.8.2</version>
     </dependency>
   </dependencies>
 </project>


jlaur:
@kaikreuzer - would this require rrd4j to be uploaded to jfrog? If you agree on @holgerfriedrich's assessment, feel free to upload and merge. @holgerfriedrich, how much were you able to test this upgrade?

holgerfriedrich:
@jlaur actually I did not test much, I have built the binding for #13955 including this fix.
I have not seen a problem, at first glance my items were stored and restored properly.

Functionally, there should not be much of change (if we believe in the tags on GitHub)
https://github.com/rrd4j/rrd4j/compare/3.8.1...3.8.2#diff-0064b069888fef98c4f7dc8e0e36546053a710356c2d6e20ed5b5ec43c0bb2f8
But there were several dependency updates in the pom, I cannot judge the impact.

And I do not know how the dependencies to different versions of libraries are resolved in the framework.

wborn:
rrd4j only has log4j as a test dependency, see:

https://github.com/rrd4j/rrd4j/blob/117120fc4338a36e0fb86d739b915996664ebc8a/pom.xml#L91-L108

jlaur (approved):
Thanks!

jlaur:
@kaikreuzer - can you upload [rrd4j-3.8.2.jar](https://repo1.maven.org/maven2/org/rrd4j/rrd4j/3.8.2/rrd4j-3.8.2.jar) if this is needed?

kaikreuzer:
I already uploaded this today due to https://github.com/openhab/openhab-webui/issues/1595.

PR #13964, opened 2022-12-15
[mapdb] Add some documentation

wborn:
No docs showed for this add-on in Main UI because there is not even a README.md for this useful persistence service. :roll_eyes: 
It's also missing from https://www.openhab.org/addons/ which makes it also hard to discover.
These docs are based on the original OH1 documentation which did contain some info on its advantages and limitations.


jlaur (approved):
Thanks, good finding!

Confectrician:
👌

https://next.openhab.org/addons/persistence/mapdb/

wborn:
Very nice! :slightly_smiling_face: :rocket:

PR #13966, opened 2022-12-16
[mapdb] Fix "a" typo and use "-" for bullets in docs

wborn:
Here are a few small documentation tweaks on top of #13964.


jlaur (approved):
Thanks, I missed that in the previous PR.

PR #13967, opened 2022-12-16
[bluetooth] Null annotations and SAT

lsiepel:
This and some more PR's to follow focus on reducing the SAT warnings and add null annotations to the set of bluetooth bindings. While doing so i tried to change as less as possible to the inner workings as it is a complex set.

Hopefully some (not yet) bugs are fixed in advance and no new ones are introduced.

Depends on https://github.com/openhab/openhab-addons/pull/13974


lsiepel:
I do have some questions. This and the related bluetooth bindings have warnings about the bridge id. I don;t understand why this is in place. So maybe someone can point me in the right direction, if this is fixable or not.

```
[WARNING] OH-INF.thing.thing-types.xml:[0]
Missing the supported bridge with id roaming
[WARNING] OH-INF.thing.thing-types.xml:[0]
Missing the supported bridge with id bluegiga
[WARNING] OH-INF.thing.thing-types.xml:[0]
Missing the supported bridge with id bluez
```

Also the BluetoothDiscoveryService.java shows some null pointer related issues i find hard to addres without making the code sloppy and unreadable. Hopefully someone can point me in the right direction.

kaikreuzer (approved):
lgtm, thanks for this clean-up!

PR #13968, opened 2022-12-16
[bluetooth.ruuvitag] SAT fixes

lsiepel:
Minor null annotations, checkstyle and other SAT-warnings.


lsiepel:
Might need some guidance in solving the last warning:
RuuviTagHandler.java:[103,17] Redundant null check: The variable manufacturerData cannot be null at this location

@Nullable is not possible (due to byte primitive), but it is actually nullable.

lsiepel:
too soon

PR #13975, opened 2022-12-16
[bluetooth.daikinmadoka] null annotations

lsiepel:

null annotations
checkstyle
other SAT warnings
Part of multiple PR's for bluetooth bundles. This may be merged independent.

Some warnings (compile and sat) are left as is. If those need to be fixed, i need some guidance, but i think it can be left alone (for now).

### DIFF
--- a/bundles/org.openhab.binding.bluetooth.daikinmadoka/src/main/java/org/openhab/binding/bluetooth/daikinmadoka/handler/DaikinMadokaHandler.java
+++ b/bundles/org.openhab.binding.bluetooth.daikinmadoka/src/main/java/org/openhab/binding/bluetooth/daikinmadoka/handler/DaikinMadokaHandler.java
@@ -83,7 +83,7 @@ public class DaikinMadokaHandler extends ConnectedBluetoothHandler implements Re
 
     private final Logger logger = LoggerFactory.getLogger(DaikinMadokaHandler.class);
 
-    private @NonNullByDefault({}) DaikinMadokaConfiguration config;
+    private DaikinMadokaConfiguration config = new DaikinMadokaConfiguration();
 
     private @Nullable ExecutorService commandExecutor;
 
@@ -823,7 +823,7 @@ public class DaikinMadokaHandler extends ConnectedBluetoothHandler implements Re
         if (indicatorStatus != null) {
             this.madokaSettings.setCleanFilterIndicator(indicatorStatus);
             updateStateIfLinked(DaikinMadokaBindingConstants.CHANNEL_ID_CLEAN_FILTER_INDICATOR,
-                    indicatorStatus ? OnOffType.ON : OnOffType.OFF);
+                OnOffType.from(indicatorStatus));
         }
     }
 
--- a/bundles/org.openhab.binding.bluetooth.daikinmadoka/src/main/java/org/openhab/binding/bluetooth/daikinmadoka/handler/DaikinMadokaHandler.java
+++ b/bundles/org.openhab.binding.bluetooth.daikinmadoka/src/main/java/org/openhab/binding/bluetooth/daikinmadoka/handler/DaikinMadokaHandler.java
@@ -441,7 +441,7 @@ public class DaikinMadokaHandler extends ConnectedBluetoothHandler implements Re
                 }
             }
 
-            if (command.getState() == BRC1HCommand.State.SENT && config != null) {
+            if (command.getState() == BRC1HCommand.State.SENT) {
                 if (!command.awaitStateChange(config.commandTimeout, TimeUnit.MILLISECONDS,
                         BRC1HCommand.State.SUCCEEDED, BRC1HCommand.State.FAILED)) {
                     logger.debug("[{}] Command {} to device {} timed out", super.thing.getUID().getId(), command,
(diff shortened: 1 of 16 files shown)


bundles/org.openhab.binding.bluetooth.daikinmadoka/src/main/java/org/openhab/binding/bluetooth/daikinmadoka/handler/DaikinMadokaHandler.java
```diff
@@ -84,7 +83,7 @@ public class DaikinMadokaHandler extends ConnectedBluetoothHandler implements Re
 
     private final Logger logger = LoggerFactory.getLogger(DaikinMadokaHandler.class);
 
-    private @Nullable DaikinMadokaConfiguration config;
+    private @NonNullByDefault({}) DaikinMadokaConfiguration config;
```
lolodomo, line 86: Can we avoid that ?
Maybe by initializing it with new DaikinMadokaConfiguration() ?

bundles/org.openhab.binding.bluetooth.daikinmadoka/src/main/java/org/openhab/binding/bluetooth/daikinmadoka/handler/DaikinMadokaHandler.java
```diff
@@ -825,7 +823,7 @@ public void receivedResponse(GetCleanFilterIndicatorCommand command) {
         if (indicatorStatus != null) {
             this.madokaSettings.setCleanFilterIndicator(indicatorStatus);
             updateStateIfLinked(DaikinMadokaBindingConstants.CHANNEL_ID_CLEAN_FILTER_INDICATOR,
-                    indicatorStatus == true ? OnOffType.ON : OnOffType.OFF);
+                    indicatorStatus ? OnOffType.ON : OnOffType.OFF);
```
lolodomo, line 826: Suggestion: OnOffType.fromValue(indicatorStatus)

bundles/org.openhab.binding.bluetooth.daikinmadoka/src/main/java/org/openhab/binding/bluetooth/daikinmadoka/internal/model/commands/GetCleanFilterIndicatorCommand.java
```diff
@@ -30,14 +28,13 @@
...
-    private final Logger logger = LoggerFactory.getLogger(GetCleanFilterIndicatorCommand.class);
-
     private @Nullable Boolean cleanFilterIndicator;
 
     @Override
     public void handleResponse(Executor executor, ResponseListener listener, MadokaMessage mm)
             throws MadokaParsingException {
-
+        // In similar class GetVersionCommand.java is a note about intentionally leaving the NPE, might be intentionally
+        // here too
```
lolodomo, line 37: What is the argument to leave a potential NPE ???

lolodomo (changes requested):

bundles/org.openhab.binding.bluetooth.daikinmadoka/src/main/java/org/openhab/binding/bluetooth/daikinmadoka/internal/model/commands/GetCleanFilterIndicatorCommand.java
lsiepel: In the linked file there is this comment from the developer who build the binding:
```
        // In this method, we intentionally do not check for null values in mv45 and mv46. In case of null pointer
        // access, it will be catched by the global exception and be reported as a Parsing Reponse exception.
```

I see this pattern on all `public void handleResponse` methods in this binding. Only at  the other place there are no comments, so i added them to clarify why they are left behind. I could catch them and throw a `MadokaParsingException` but the dev did this on purpose and it was merged so. 

Let me know if it needs to be changed, otherwise merge as is.

lsiepel: I refactored those cases to do proper null check and throw a `MadokaParsingException` if null.

bundles/org.openhab.binding.bluetooth.daikinmadoka/src/main/java/org/openhab/binding/bluetooth/daikinmadoka/internal/model/commands/GetCleanFilterIndicatorCommand.java
```diff
@@ -51,7 +54,12 @@ public void handleResponse(Executor executor, ResponseListener listener, MadokaM
         }
 
         setState(State.SUCCEEDED);
-        executor.execute(() -> listener.receivedResponse(this));
+        try {
+            executor.execute(() -> listener.receivedResponse(this));
+        } catch (Exception e) {
```
lolodomo, line 59: You cannot catch more specific exception(s) ?

lsiepel: This binding has numerous calls to `executor.execute(() ->` some of them had a catch block, some not. I checked if any exception was catch higher up the chain and if not add the catch block to make to code more robust/same structured.

I can not determine the possible exception types and more refactoring goes beyond the scope of this PR, if you ask me. But if you can give me some tips, i'm willing to look into it and check if i can improve it. Otherwise please merge as is.

lolodomo: The tip would be to check what exception can be thrown by `listener.receivedResponse`

But ok, let's do that in another PR if you like.

lolodomo (approved):
LGTM

PR #13979, opened 2022-12-16
[bluetooth.generic] null annotations and SAT warnings

lsiepel:
Signed-off-by: Leo Siepel [leosiepel@gmail.com](mailto:leosiepel@gmail.com)

null annotations
checkstyle.

=> would be good to remove the forbidden units package, but it depends on changes in core. I could not find if they are added yet. 
=> i see several warnings that i'm unable to fix yet (circular dependency with other classes)

Would be cool if i get some advice in how to fix those remaining issues.

Part of multiple PR's for bluetooth bundles. Can be merged independent.


lolodomo (approved):
LGTM

Issue #10838, opened 2021-06-09
[yamahareceiver] Yamaha Receiver RX-V4A: HANDLER_INITIALIZING_ERROR


MiHechtel:
I tried to connect my Yamaha RX-V4A Receiver to openHAB via the Yamaha Binding. I used the autodetection function  (same behavior when the thing is added manually in things file), which is why there should be no configuration errors here. But nevertheless I get the HANDLER_INITIALIZING_ERROR you can see in the picture. 
This issue occured in openHAB 2 and also after the upgrade in openHAB 3. 

If anyone else got the same problem I would suggest that there could be a bug in the Yamaha Binding?

![image](https://user-images.githubusercontent.com/37272678/121381809-5374f680-c946-11eb-8221-5ac859f2623e.png)

## Expected Behavior
The Receiver Thing should successfully connect to my Yamaha RX-V4A

## Current Behavior
When I try to add the auto-detected Thing it goes directly into the HANDLER_INITIALIZING_ERROR state (same behavior when the thing is added manually in things file)

## Steps to Reproduce (for Bugs)
Install Yamaha Binding
Add Auto-Detect Receiver Thing
-> Thing goes into HANDLER_INITIALIZING_ERROR state

## Your Environment
Openhab 3.0.2
Yamaha Binding 3.0.2

florian-h05:
Does the problem still exist?

florian-h05:
I think I can take a look at this in the next days/weeks. 
Can you please provide me some logs as well as your thing configuration? 
Probably trace logs, you can set the log level to TRACE by running `log:set TRACE org.openhab.binding.yamahareceiver` in the openHAB console.

florian-h05:
I've just had a closer look at this, seems like the binding expects some information from the API that is not returned.

I think that Yamaha changed their API on newer models (your RX-V4A is from 2021, right?) and the Yamaha Receiver binding says:

```
This binding connects openHAB with Yamaha Receivers of product line CX-A5000, RX-A860, RX-A30xx, RX-A20xx, RX-A10xx, RX-Vxxx, RX-Z7, DSP-Z7, RX-S600, RX-S601D, HTR-xxxx.
```

Your receiver is not on the compatibility list, so we can not expect the binding to work with it.

Since your issue is from mid 2021, back then you only had the option to use the Yamaha Receiver binding, but meanwhile the [Yamaha MusicCast binding](https://www.openhab.org/addons/bindings/yamahamusiccast/) was released.

[...]

PR #13985, opened 2022-12-17
[yamaha] README: Recommend correct binding for specific model

florian-h05:
This doc improvement tries to avoid future issue like #10838.


lolodomo (approved):
LGTM

PR #13986, opened 2022-12-17
[bondhome] compile warnings and sat

lsiepel:
Fixed some checkstyle and compile warnings
Also replace deprecated gson calls.


ccutrer (approved):

bundles/org.openhab.binding.bondhome/src/main/java/org/openhab/binding/bondhome/internal/discovery/BondDiscoveryService.java
```diff
@@ -94,7 +96,7 @@ protected synchronized void startScan() {
                 for (final String deviceId : deviceList) {
                     BondDevice thisDevice = api.getDevice(deviceId);
                     String deviceName;
-                    if (thisDevice != null && (deviceName = thisDevice.name) != null) {
+                    if ((deviceName = thisDevice.name) != null) {
```
lolodomo, line 99: `thisDevice` can't be null there? 
That means api.getDevice(deviceId) can't return null?

bundles/org.openhab.binding.bondhome/src/main/java/org/openhab/binding/bondhome/internal/handler/BondBridgeHandler.java
```diff
@@ -242,23 +241,39 @@ public void forwardUpdateToThing(BPUPUpdate pushUpdate) {
      * Returns the Id of the bridge associated with the handler
      */
     public String getBridgeId() {
-        String serialNumber = config.serialNumber;
-        return serialNumber == null ? "" : serialNumber;
+        BondBridgeConfiguration configuration = config;
```
lolodomo, line 244: You could also have config initialized with new BondBridgeConfiguration() so that no need to check if it is null everywhere. This is a common pattern in many bindings. 
That is just a suggestion, the code is also good as it is.

bundles/org.openhab.binding.bondhome/src/main/java/org/openhab/binding/bondhome/internal/discovery/BondDiscoveryService.java
lsiepel: Yes. It returns a BondDevice or it throws `BondException`
```
    public BondDevice getDevice(String deviceId) throws BondException {
        String json = request("/v2/devices/" + deviceId);
        logger.trace("BondHome device info : {}", json);
        try {
            return Objects.requireNonNull(gson.fromJson(json, BondDevice.class));
        } catch (JsonParseException e) {
            logger.debug("Could not parse device {}'s JSON '{}'", deviceId, json, e);
            throw new BondException("@text/offline.comm-error.unparseable-response");
        }
    }
```

bundles/org.openhab.binding.bondhome/src/main/java/org/openhab/binding/bondhome/internal/handler/BondBridgeHandler.java
lsiepel: Refactored the config init

bundles/org.openhab.binding.bondhome/src/main/java/org/openhab/binding/bondhome/internal/handler/BondBridgeHandler.java
```diff
@@ -288,9 +284,11 @@ public void setBridgeOffline(ThingStatusDetail detail, String description) {
      * Called by the UDP listener when it gets a proper response.
      */
     public void setBridgeOnline(String bridgeAddress) {
-        BondBridgeConfiguration localConfig = config;
-        if (localConfig.ipAddress == null || !localConfig.ipAddress.equals(bridgeAddress)) {
-            logger.debug("IP address of Bond {} has changed to {}", localConfig.serialNumber, bridgeAddress);
+        if (config.isValid()) {
```
lolodomo, line 287: I guess the expected condition is rather `!config.isValid()` ?

lolodomo (changes requested):

lolodomo (approved):
LGTM, thank you

PR #13276, opened 2022-08-16
Raise minimum JDK version to 17

J-N-K:
Depends on https://github.com/openhab/openhab-core/pull/2994

Signed-off-by: Jan N. Klug <github@klug.nrw>

PR #13913, opened 2022-12-11
[siemensrds] Update to documentation

andrewfg:
This PR contains the following updates..
- A bug fix that resolves #13906
- An update to the read me documentation file

Signed-off-by: Andrew Fiddian-Green <software@whitebear.ch>

Issue #13906, opened 2022-12-10
[siemensrds] Unit test failing


lolodomo:
On Windows 10:
```
[INFO] -------------------------------------------------------
[INFO]  T E S T S
[INFO] -------------------------------------------------------
[INFO] Running org.openhab.binding.siemensrds.test.TestRdsData
[ERROR] Tests run: 6, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 0.7 s <<< FAILURE! - in org.openhab.binding.siemensrds.test.TestRdsData
[ERROR] org.openhab.binding.siemensrds.test.TestRdsData.testAccessToken  Time elapsed: 0.402 s  <<< FAILURE!
org.opentest4j.AssertionFailedError: expected: <true> but was: <false>
        at org.openhab.binding.siemensrds.test.TestRdsData.testAccessToken(TestRdsData.java:208)

[INFO]
[INFO] Results:
[INFO]
[ERROR] Failures:
[ERROR]   TestRdsData.testAccessToken:208 expected: <true> but was: <false>
[INFO]
[ERROR] Tests run: 6, Failures: 1, Errors: 0, Skipped: 0
[INFO]
[INFO] ------------------------------------------------------------------------
[INFO] BUILD FAILURE
[INFO] ------------------------------------------------------------------------
```

lolodomo:
If I replace the current line by `expDate = new SimpleDateFormat("EEE, dd MMM yyyy HH:mm:ss z", Locale.ENGLISH).parse(expires);`, the test is then OK.

andrewfg:
@lolodomo good catch!

The ClimatixIC cloud server speaks English; AFAIK it is an (American) AWS server. So the `isExpired()` method must indeed parse the date using English day / month names ('Sun' vs 'dim.' / 'Dec' vs 'déc.', etc.).  And therefore currently `isExpired()` will, as you say, fail on computers in regions where the language is not English.

Obviously my Windows PC, my Linux OH system, and the GitHub CI build system, have English locales, which explains why the problem was not broadly identified before now.

> not sure if the problem is only the unit test .. or more generally the binding

Luckily it is NOT a problem in the binding -- because the binding code actually never calls `isExpired()` (!!) -- instead, when the current token expires, the cloud server rejects the next API request with a 'not authorised' response, that forces the binding to request a new token. In other words the expiry check is done by the cloud server and not by the binding.

[...]

PR #13989, opened 2022-12-17
[siemensrds] Fix missing locale in accessToken parsing

J-N-K:
The missing locale results in a test error on MacOS Ventura with german locale. The reason is that with a german locale the weekday cannot be parsed. Probably this is also the case in real life but since there is a fallback if date parsing fails.

Discovered while working on #13276 


lolodomo:
https://github.com/openhab/openhab-addons/blob/main/bundles/org.openhab.binding.siemensrds/src/main/java/org/openhab/binding/siemensrds/internal/RdsAccessToken.java#L134

This was already fixed in PR #13913 (issue #13906). 
You are not in line with the main branch?

J-N-K:
Indeed. My last fetch was a week ago.

jlaur:
> https://github.com/openhab/openhab-addons/blob/main/bundles/org.openhab.binding.siemensrds/src/main/java/org/openhab/binding/siemensrds/internal/RdsAccessToken.java#L134
> 
> This was already fixed in PR #13913 (issue #13906). You are not in line with the main branch?

Nice catch. Consider rebasing to also add the exception to the log.

PR #13992, opened 2022-12-17
Show add-on intros in Main UI

wborn:
Using these Markdown tweaks a small intro will show in Main UI instead of emptiness and a "more" button:

Screenshot from 2022-12-17 22-29-39

There's also https://github.com/openhab/openhab-webui/pull/1587 for UI add-ons.


jlaur (approved):
Thanks!

PR #13998, opened 2022-12-18
[nanoleaf] More robust caching of layout

austvik:
This is a bugfix/enhancement to make sure the caching of the layout (to save it from being recalculated) works better:
- Only save previous layout if indeed painted
- Only save layout from the layout update, not the display state
- Recalculate anyway if current state is null

### DIFF
--- a/bundles/org.openhab.binding.nanoleaf/src/main/java/org/openhab/binding/nanoleaf/internal/handler/NanoleafControllerHandler.java
+++ b/bundles/org.openhab.binding.nanoleaf/src/main/java/org/openhab/binding/nanoleaf/internal/handler/NanoleafControllerHandler.java
@@ -121,6 +121,7 @@ public class NanoleafControllerHandler extends BaseBridgeHandler implements Nano
     private final List<NanoleafControllerListener> controllerListeners = new CopyOnWriteArrayList<NanoleafControllerListener>();
     private PanelLayout previousPanelLayout = new PanelLayout();
     private final NanoleafPanelColors panelColors = new NanoleafPanelColors();
+    private boolean updateVisualLayout = true;
     private byte @Nullable [] layoutImage;
 
     private @NonNullByDefault({}) ScheduledFuture<?> pairingJob;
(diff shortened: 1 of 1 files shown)


austvik:
I don̈́'t have a repro, so would be great if @stefan-hoehn checked if it resolved the issue he is seeing.

stefan-hoehn:
@austvik The code looks good to me and it is also working <3

image

@kaikreuzer As discussed earlier can you cherry pick this one to 3.4.1, please?

stefan-hoehn (approved):
LGTM

jlaur (approved):
Thanks!

stefan-hoehn:
@dougculnane @jlaur Thanks for merging.

Did you see my note above?

As discussed earlier [with Kai] can you cherry pick this one to 3.4.1, please?

jlaur:
@stefan-hoehn - can you either direct me to that discussion, or describe the impact of this bug (which I understand is a regression in 3.4)? We are only cherry-picking critical bugfixes into that branch, so just need to fully understand what we are dealing with here. Cc @kaikreuzer

kaikreuzer:
> can you either direct me to that discussion

@jlaur That was a private discussion, but I only stated that it is in general possible to backport fixes to the 3.4.x branch, if this is necessary. We didn't talk about this concrete bug/enhancement.

> describe the impact of this bug

@stefan-hoehn Please comment on this as it helps the maintainers to decide whether it is worth to include in a potential upcoming 3.4.1 patch release.

stefan-hoehn:
Yes, fair enough, I do understand that. 

The name of this PR is actually a bit misleading as the caching is the cause but not the problem. @austvik has worked the past few weeks to bring this visualization to life but unfortunately with the last commit before this one he broke one functionality that prevents the picture from being generated (basically the caching was too "aggressive") which is a pity as it is really such nice feature he contributed which would be said not be possible to be used before 4.0.

jlaur:
@stefan-hoehn - I agree. Not critical in the sense that binding was broken, but new functionality introduced in 3.4 didn't work at all. I have cherry-picked the commit into the hotfix branch. Thank you both for the enhancements and this fix.

PR #14003, opened 2022-12-18
[jdbc] Update org.xerial/sqlite-jdbc to 3.40.0.0

### LINKED ISSUE #14002: [jdbc] SQLite doesn't work on aarch64 architectures (e.g. openhabian 64bit)
## Expected Behavior
The SQLite driver should work after installing it from the addons store.

## Current Behavior
While it is possible to install the addon, it won't create the database and also not write to it.
The logs show an error: `Error opening connection`.

The behavior was already discussed in the forum: https://community.openhab.org/t/how-can-i-change-oh3-3-to-install-a-aarch64-capable-version-of-sqlite-jdbc/139102

## Possible Solution
As discussed in the forum, the solution is to update the `org.xerial/sqlite-jdbc` dependency. 

1. Setup a raspi 4 with openhabian 64bit (default settings)
2. install and configure the jdbc sqlite addon
3. inspect the logs  

## Your Environment
* Version used: Openhab 3.3.0
* Operating System and version: Linux openhabian 5.15.76-v8+ #1597 SMP PREEMPT Fri Nov 4 12:16:41 GMT 2022 aarch64 GNU/Linux

Ka0o0:
## Description

The current version of org.xerial/sqlite-jdbc does not support aarch64 (e.g. openhabian 64bit).
This PR updates this dependency to the latest version and fixes #14002 .

## Testing

I've tested this on my own raspi and it seems to work just fine with this version.

### DIFF
--- a/bundles/org.openhab.persistence.jdbc/pom.xml
+++ b/bundles/org.openhab.persistence.jdbc/pom.xml
@@ -31,7 +31,7 @@
     <mariadb.version>3.0.8</mariadb.version>
     <mysql.version>8.0.30</mysql.version>
     <postgresql.version>42.4.3</postgresql.version>
-    <sqlite.version>3.16.1</sqlite.version>
+    <sqlite.version>3.40.0.0</sqlite.version>
   </properties>
 
   <dependencies>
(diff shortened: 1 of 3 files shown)


lolodomo:
@jlaur : I let you decide if this is to be merged now or not. Don't know what are the risks.

jlaur (approved):
Thanks for the upgrade and with that the 64 bit CPU support! @kaikreuzer, if you can upload to jfrog today, I would suggest to merge this as SQLite cannot be used by openHABian 64 bit currently.

kaikreuzer:
Uploaded to artifactory!

PR #14012, opened 2022-12-19
[nuvo] Auto update source channel for grouped zones

mlobstein:
For zones that are configured into a source group, the changing of the source in one zone causes the amplifier to change the selected source in all other zones of that group. However no updates for the other zones are reported by the serial protocol and thus the other group members will now show incorrect information for the source channel.

To address this, the configuration for each zone is retrieved at startup and a map object to track group membership information is populated. When any source channel changes in a zone, the new code will check if that zone is in a group and if so update the source channel for the other zones in the group.

### DIFF
--- a/bundles/org.openhab.binding.nuvo/src/main/java/org/openhab/binding/nuvo/internal/handler/NuvoHandler.java
+++ b/bundles/org.openhab.binding.nuvo/src/main/java/org/openhab/binding/nuvo/internal/handler/NuvoHandler.java
@@ -134,7 +134,9 @@ public class NuvoHandler extends BaseThingHandler implements NuvoMessageEventLis
     private static final Pattern DISP_PATTERN = Pattern.compile("^DISPLINE(\\d{1}),\"(.*)\"$");
     private static final Pattern DISP_INFO_PATTERN = Pattern
             .compile("^DISPINFO,DUR(\\d{1,6}),POS(\\d{1,6}),STATUS(\\d{1,2})$");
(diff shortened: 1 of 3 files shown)


bundles/org.openhab.binding.nuvo/src/main/java/org/openhab/binding/nuvo/internal/handler/NuvoHandler.java
```diff
@@ -860,7 +880,18 @@ public void onNewMessageEvent(NuvoMessageEvent evt) {
                     updateChannelState(NuvoEnum.valueOf(ZONE + zoneId), CHANNEL_TYPE_LOUDNESS,
                             ONE.equals(matcher.group(4)) ? ON : OFF);
                 } else {
-                    logger.debug("no match on message: {}", updateData);
+                    matcher = ZONE_CFG_PATTERN.matcher(updateData);
+                    // example: ENABLE1,NAME"Great Room",SLAVETO0,GROUP1,SOURCES63,XSRC0,IR1,DND0,LOCKED0,SLAVEEQ0
+                    if (matcher.find()) {
+                        // Future; utilize other info such as zone name, available sources bitmask, etc.
```
lsiepel, line 886: Could you change Future to TODO, it will then show up in SAT and will be on the 'radar' for future developments.

lsiepel (approved):
LGTM

lolodomo (approved):
LGTM

PR #14013, opened 2022-12-19
[jsscriptingnashorn] JavaScript Scripting Nashorn Automation 

wborn:
This add-on allows you to use your older ECMAScript 5.1 code on newer Java versions until the code is migrated to ECMAScript 2021+. It should only be installed for providing backwards compatibility. When writing new code it is preferred to do this using ECMAScript 2021+ for which support is provided by installing the [JavaScript Scripting](https://www.openhab.org/addons/automation/jsscripting/) add-on.

This add-on uses a standalone [Nashorn Engine](https://github.com/openjdk/nashorn) to run ECMAScript 5.1 code. The Nashorn Engine was pre-installed in openHAB 2 and openHAB 3 because it was part of Java. Since Java 15 the Nashorn Engine has been removed from Java.

---

This PR is based on the [JavaScript Scripting (Nashorn) Marketplace add-on](https://community.openhab.org/t/javascript-scripting-nashorn/134770?u=wborn).

---

To do:

* [x] Add more documentation
* [x] Update MIME types
* [x] Add more tests

### DIFF
--- a/bundles/org.openhab.automation.jsscriptingnashorn/src/main/java/org/openhab/automation/jsscriptingnashorn/internal/NashornScriptEngineFactory.java
+++ b/bundles/org.openhab.automation.jsscriptingnashorn/src/main/java/org/openhab/automation/jsscriptingnashorn/internal/NashornScriptEngineFactory.java
@@ -38,23 +38,11 @@ import org.osgi.service.component.annotations.Component;
 @NonNullByDefault
 public class NashornScriptEngineFactory extends AbstractScriptEngineFactory {
 
-    private static final String NASHORN_ARGS = "nashorn.args";
-
     private final org.openjdk.nashorn.api.scripting.NashornScriptEngineFactory factory = new org.openjdk.nashorn.api.scripting.NashornScriptEngineFactory();
 
     private final List<String> scriptTypes = (List<String>) Stream.of(factory.getExtensions(), factory.getMimeTypes())
             .flatMap(List::stream) //
             .collect(Collectors.toUnmodifiableList());
-
-    public NashornScriptEngineFactory() {
-        removeUnsupportedNashornArgs();
-    }
-
-    private void removeUnsupportedNashornArgs() {
-        String property = System.getProperty(NASHORN_ARGS, "");
-        property = property.replaceAll("--no-deprecation-warning", "").trim();
-        System.setProperty(NASHORN_ARGS, property);
-    }
 
     @Override
     public List<String> getScriptTypes() {
--- a/bundles/org.openhab.automation.jsscriptingnashorn/src/main/java/org/openhab/automation/jsscriptingnashorn/internal/NashornScriptEngineFactory.java
+++ b/bundles/org.openhab.automation.jsscriptingnashorn/src/main/java/org/openhab/automation/jsscriptingnashorn/internal/NashornScriptEngineFactory.java
@@ -40,9 +40,17 @@ public class NashornScriptEngineFactory extends AbstractScriptEngineFactory {
 
     private final org.openjdk.nashorn.api.scripting.NashornScriptEngineFactory factory = new org.openjdk.nashorn.api.scripting.NashornScriptEngineFactory();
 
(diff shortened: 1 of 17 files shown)


bundles/org.openhab.automation.jsscriptingnashorn/src/main/java/org/openhab/automation/jsscriptingnashorn/internal/NashornScriptEngineFactory.java
```diff
@@ -0,0 +1,88 @@
...
+    public NashornScriptEngineFactory() {
+        removeUnsupportedNashornArgs();
+    }
+
+    private void removeUnsupportedNashornArgs() {
+        String property = System.getProperty(NASHORN_ARGS, "");
+        property = property.replaceAll("--no-deprecation-warning", "").trim();
+        System.setProperty(NASHORN_ARGS, property);
+    }
+
```
wborn, line 58:
```suggestion
```

This can be removed once the property is removed from [setenv](https://github.com/openhab/openhab-distro/blob/466a3427711fe90b4b6511808b461edcc6ed6b12/distributions/openhab/src/main/resources/bin/setenv#L109) / [setenv.bat](https://github.com/openhab/openhab-distro/blob/466a3427711fe90b4b6511808b461edcc6ed6b12/distributions/openhab/src/main/resources/bin/setenv.bat#L126). The default Nashorn engine will no longer warn about its deprecation because it has been removed. :wink:

J-N-K:
Would it make sense to integrate the "old" JS transformation or shall we point people to use the `SCRIPT`  transformation? Since add-ons are not allowed to depend on add-ons, we can't ship that alone and I removed it in #13276.

florian-h05:
@wborn Thanks for your PR. I‘ve seen that you listed MIME types as todo, I‘d prefer to discuss that on the issue. For the docs, you could have a look how the GraalJS docs were looking before openhab-js was introduced: https://github.com/openhab/openhab-addons/blob/9912e1afc277f50a8c4f6c1c52634ec7716cc87d/bundles/org.openhab.automation.jsscripting/README.md.

@J-N-K I would vote to remove the „old“ JS transformations because for Item transformations you can just switch to the `SCRIPT` transformations, but we are missing both inline transformation scripts for the `SCRIPT` as well as a `SCRIPT` transformation profile for channel links, see https://www.openhab.org/addons/transformations/javascript/#inline-script-example.

J-N-K:
Inline-Scripting can be added, looks simple, and I can open a successor for https://github.com/openhab/openhab-core/pull/2872 which would add exactly the missing profile, but I don't see how that would help if no agreement on the design can be reached.

florian-h05:
> Inline-Scripting can be added, looks simple, 

Great 👍 This is a huge simplification because I then don’t need a file for each and every SCRIPT transformation.

> would add exactly the missing profile, but I don't see how that would help if no agreement on the design can be reached.

I have had a short look at your PR, I think I need to get my own opinion on this. We need to find a solution for this, otherwise we loose functionality.

wborn:
I've updated the docs and adjusted the script types. I expect the migration of existing scripts to be done in some other PR. :wink:

florian-h05:
The MIME type version looks good and the docs as well.
Reviewing this won’t help since I’m no maintainer.

> I expect the migration of existing scripts to be done in some other PR. 😉

FYI see https://github.com/openhab/openhab-webui/pull/1601, which aims to allow the user to easily change the MIME type of a script. So users can install your Nashorn addon and update the MIME type of their scripts before they have the time to migrate them to GraalJS.

jlaur:
> Reviewing this won’t help since I’m no maintainer.

Sure it will. 🙂

wborn:
I think it's complete now. There's even an itest that spins up the engine. :slightly_smiling_face:

florian-h05:
I'll try to review this soon.

florian-h05 (approved):
Code LGTM, thanks!
I've also tested this with the example code from the README, works fine!
I've also had a look at the tests, they all seem reasonable to me :+1: 

@jlaur From my side, this is ready for merging.

jlaur (approved):
Great, many thanks for providing this backwards compatibility.

PR #14016, opened 2022-12-19
[homekit] Improve multiple instance management

ccutrer:


bundles/org.openhab.io.homekit/src/main/java/org/openhab/io/homekit/internal/HomekitCommandExtension.java
```diff
@@ -206,4 +254,12 @@ private void printAccessory(String id, Console console) {
             }
         });
     }
+
+    private Collection<HomekitAccessory> getIntanceAccessories(@Nullable Integer instance) {
+        if (instance != null) {
```
timcharper, line 259: it's a shame we need two different code paths for this. Could we just refer to instance 0 in the case of non-multi-instance setups?

timcharper: OHHHH I get it now. Instance being null means "all accessories on all instances".

Any opinion on using Optional over nullable integer, instead?

bundles/org.openhab.io.homekit/src/main/java/org/openhab/io/homekit/Homekit.java
```diff
@@ -45,17 +45,42 @@ public interface Homekit {
     void allowUnauthenticatedRequests(boolean allow);
 
     /**
-     * returns list of HomeKit accessories registered at bridge.
+     * returns list of HomeKit accessories registered on all bridges.
```
timcharper, line 48:
```suggestion
     * returns list of HomeKit accessories registered on all bridge instances.
```

bundles/org.openhab.io.homekit/src/main/java/org/openhab/io/homekit/internal/HomekitImpl.java
```diff
@@ -295,21 +346,39 @@ public void allowUnauthenticatedRequests(boolean allow) {
...
     public void clearHomekitPairings() {
+        for (int i = 1; i <= authInfos.size(); ++i) {
+            clearHomekitPairings(i);
+        }
+    }
+
+    @Override
+    public void clearHomekitPairings(int instance) {
+        if (instance < 1 || instance > authInfos.size()) {
+            return;
```
timcharper, line 376: should we log a warning here?

bundles/org.openhab.io.homekit/src/main/java/org/openhab/io/homekit/internal/HomekitImpl.java
```diff
@@ -322,6 +391,20 @@ public void pruneDummyAccessories() {
         }
     }
 
+    @Override
+    public void pruneDummyAccessories(int instance) {
+        if (instance < 1 || instance > authInfos.size()) {
+            return;
```
timcharper, line 397: log a warning?

bundles/org.openhab.io.homekit/src/main/java/org/openhab/io/homekit/internal/HomekitImpl.java
```diff
@@ -211,60 +227,75 @@ private HomekitRoot startBridge(HomekitServer homekitServer, HomekitAuthInfoImpl
...
-            try {
-                networkInterface = InetAddress
-                        .getByName(((settings.networkInterface != null) && (!settings.networkInterface.isEmpty()))
-                                ? settings.networkInterface
-                                : networkAddressService.getPrimaryIpv4HostAddress());
-            } catch (UnknownHostException e) {
-                logger.warn("cannot resolve the Pv4 address / hostname {}.",
-                        networkAddressService.getPrimaryIpv4HostAddress());
+    private void startHomekitServer(int instance) throws IOException, InvalidAlgorithmParameterException {
+        logger.trace("starting HomeKit bridge instance {}", instance + 1);
```
timcharper, line 231: It seems like instance is sometimes 0 based, as seen in this function, and in other functions, 1 based (ie pruneDummyAccessories). This seems likely that some poor soul will get it wrong (and by poor soul, I mean myself). I'm wondering if we should normalize as 0-based at the UI level (parsing the command), and be consistent?

timcharper (approved):
Looks good! A few small suggestions, I think. More strongly the mixing of instance being 0-based vs 1-based, but overall looks great :)

bundles/org.openhab.io.homekit/src/main/java/org/openhab/io/homekit/internal/HomekitImpl.java
```diff
@@ -211,60 +227,75 @@ private HomekitRoot startBridge(HomekitServer homekitServer, HomekitAuthInfoImpl
...
+        }
+
+        try {
+            String interfaceName = ((settings.networkInterface != null) && (!settings.networkInterface.isEmpty()))
+                    ? settings.networkInterface
+                    : networkAddressService.getPrimaryIpv4HostAddress();
+            return (networkInterface = Objects.requireNonNull(InetAddress.getByName(interfaceName)));
+        } catch (UnknownHostException e) {
+            logger.warn("cannot resolve the IPv4 address / hostname {}.",
+                    networkAddressService.getPrimaryIpv4HostAddress());
```
yfre, line 292: i know, this line is from older code, but just saw that the parameter for log message is wrong if settings.networkInterface is not empty.
could you please replace networkAddressService.getPrimaryIpv4HostAddress()
with 
interfaceName

bundles/org.openhab.io.homekit/src/main/java/org/openhab/io/homekit/internal/HomekitImpl.java
```diff
@@ -211,60 +227,75 @@ private HomekitRoot startBridge(HomekitServer homekitServer, HomekitAuthInfoImpl
...
-                authInfos.add(authInfo);
+                startHomekitServer(i);
             }
         } else {
             logger.warn("trying to start HomeKit server but it is already initialized");
         }
     }
 
+    private InetAddress ensureNetworkInterface() throws IOException {
+        InetAddress localNetworkInterface = networkInterface;
```
yfre, line 280: very minor thing but in case you will touch the code anyway 
i think we dont need to create localNetworkInterface here but just

if (networkInterface != null) { return networkInterface }

yfre (approved):
LGTM

yfre:
thank you. good refactoring to support improve support for multiple instances

ccutrer (commented):
Whoops, @timcharper, I guess I didn't get my comments published last week.

ccutrer:
@yfre : I can't find when your code comments went on GitHub, I only see them in my email. anyhow, I addressed the interfaceName getting logged. I did _not_ change the temp var for localNetworkInterface - it's required, because networkInterface is @Nullable, and the compiler continues to complain if you return it directly since theoretically another thread could change it to null between the non-null check and returning it.

bundles/org.openhab.io.homekit/src/main/java/org/openhab/io/homekit/internal/HomekitCommandExtension.java
timcharper: Fair enough :)

bundles/org.openhab.io.homekit/src/main/java/org/openhab/io/homekit/internal/HomekitImpl.java
yfre: agree, even for me who used to count many things starting from 0 in the past :) would be a little strange to refer to "instance 0". i would vote to use "1" and keep it consistence in the communication to the user.

ccutrer:
@lolodomo and/or @jlaur : this is good to go from our end, just need a maintainer review. thanks!

jlaur (approved):
Looks good, thanks!

PR #14026, opened 2022-12-20
[nanoleaf] Handle and provide QuantityType for color-temperature-abs channel

ccutrer:
see https://github.com/openhab/openhab-core/pull/3129

It should be noted I don't have nanoleaf, so I have not tested this.

### DIFF
--- a/bundles/org.openhab.binding.nanoleaf/README.md
+++ b/bundles/org.openhab.binding.nanoleaf/README.md
@@ -235,7 +235,7 @@ Dimmer NanoleafBrightness "Brightness [%.0f]" { channel="nanoleaf:controller:MyL
 String NanoleafHue "Hue [%s]"
 String NanoleafSaturation "Saturation [%s]"
 Dimmer NanoleafColorTemp "Color temperature [%.0f]" { channel="nanoleaf:controller:MyLightPanels:colorTemperature" }
-Number NanoleafColorTempAbs "Color temperature [%.000f]" { channel="nanoleaf:controller:MyLightPanels:colorTemperatureAbs" }
+Number:Temperature NanoleafColorTempAbs "Color temperature [%d K]" { channel="nanoleaf:controller:MyLightPanels:colorTemperatureAbs" }
 String NanoleafColorMode "Color mode [%s]" { channel="nanoleaf:controller:MyLightPanels:colorMode" }
 String NanoleafEffect "Effect" { channel="nanoleaf:controller:MyLightPanels:effect" }
 Switch NanoleafRhythmState "Rhythm connected [MAP(nanoleaf.map):%s]" { channel="nanoleaf:controller:MyLightPanels:rhythmState" }
--- a/bundles/org.openhab.binding.nanoleaf/README.md
+++ b/bundles/org.openhab.binding.nanoleaf/README.md
@@ -139,19 +139,19 @@ This discovers all connected panels with their IDs.
 
 The controller bridge has the following channels:
 
-| Channel             | Item Type | Description                                                                                               | Read Only |
-|---------------------|-----------|-----------------------------------------------------------------------------------------------------------|-----------|
-| color               | Color     | Color, power and brightness of all light panels                                                           | No        |
-| colorTemperature    | Dimmer    | Color temperature (in percent) of all light panels                                                        | No        |
-| colorTemperatureAbs | Number    | Color temperature (in Kelvin, 1200 to 6500) of all light panels                                           | No        |
-| colorMode           | String    | Color mode of the light panels                                                                            | Yes       |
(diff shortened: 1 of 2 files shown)


jlaur:
> It should be noted I don't have LifX, so I have not tested this.

Having Nanoleaf would be even better. 😉 @austvik, @stefan-hoehn - would one of you be able to test this?

stefan-hoehn:
Sure, and, I definitely want to test this first, so it doesn't break anything but I need a few days before I'll have the time.

bundles/org.openhab.binding.nanoleaf/README.md
```diff
@@ -139,19 +139,19 @@ This discovers all connected panels with their IDs.
...
-| state               | Image     | Shows the current state of your panels with colors.                                                       | Yes       |
-| swipe               | Trigger   | [Canvas / Shapes Only] Detects Swipes over the panel.LEFT, RIGHT, UP, DOWN events are supported.          | Yes       |
+| Channel             | Item Type          | Description                                                                                               | Read Only |
+|---------------------|--------------------|-----------------------------------------------------------------------------------------------------------|-----------|
+| color               | Color              | Color, power and brightness of all light panels                                                           | No        |
+| colorTemperature    | Dimmer             | Color temperature (in percent) of all light panels                                                        | No        |
+| colorTemperatureAbs | Number:Temperature | Color temperature (in Kelvin, 1200 to 6500) of all light panels                                           | No        |
```
lolodomo, line 146: Please also update the exemple (nonaleaf.items) at the end of README.

bundles/org.openhab.binding.nanoleaf/src/main/java/org/openhab/binding/nanoleaf/internal/handler/NanoleafControllerHandler.java
```diff
@@ -899,15 +901,25 @@ private void sendStateCommand(String channel, Command command) throws NanoleafEx
...
-                    IntegerState state = new Ct();
                     state.setValue(((DecimalType) command).intValue());
-                    stateObject.setState(state);
+                } else if (command instanceof QuantityType) {
+                    QuantityType<?> tempKelvin = ((QuantityType) command).toInvertibleUnit(Units.KELVIN);
+                    if (tempKelvin == null) {
+                        logger.warn("Cannot convert color temperature {} to Kelvin.", command);
+                        return;
+                    }
+                    state.setValue(tempKelvin.toBigDecimal().intValue());
```
lolodomo, line 916: Any difference with `tempKelvin.intValue()` ?

lolodomo (changes requested):

ccutrer:
> How is this related to Lifx, @ccutrer ?

Just that LifX is one of the other bindings that uses the color-temperature-abs system channel, and I'm going through and making all the bindings that use it compatible with QuantityType. I lost track of which PR I was sending when I wrote the original comment. Edited now for clarity.

lolodomo (approved):
LGTM, thank you

lolodomo:
@raepple @stefan-hoehn: would you like to test the fix before I merge ?

stefan-hoehn:
Yes, I do, 
at: @ccutrer

 I just added that channel to one of my nanoleafs without your fix and shows already the Kelvin Unit

image

What I noticed though. when I tested this, is that my bundle actually throws a warning:

```
2022-12-21 20:40:05.143 [INFO ] [openhab.event.ItemStateChangedEvent ] - Item 'AZ_Trio_Canvas_Color_TemperatureAbs' changed from 3965 K to 1733 K

==> /var/log/openhab/openhab.log <==

2022-12-21 20:40:05.967 [WARN ] [al.handler.NanoleafControllerHandler] - Unhandled command type: org.openhab.core.library.types.QuantityType

```

So, I bet that even fixes a bug in the handler. Let me check.

stefan-hoehn:
> would you like to test the fix before I merge ?

It was a bit hard to test because that branch is already on Java 17 as a requirement, so I needed to backport the changes to a 3.4.0.RC1 before I was able to deploy and test it on my side.

In essence, @ccutrer, thanks for finding this because it is actually a bug and prevented the channel to work. I can confirm that it does work now and I am happy to have it merged, @lolodomo

ccutrer:
> In essence, @ccutrer, thanks for finding this because it is actually a bug and prevented the channel to work. I can confirm that it does work now and I am happy to have it merged, @lolodomo

Heh, you're too generous. It's a bug I caused with https://github.com/openhab/openhab-core/pull/3129. Prior to 3.4.0 your channel wouldn't default to a Number:Temperature, and thus wouldn't have the Kelvin, and thus wouldn't have been sent a QuantityType command at all. Just cleaning up my messes.

PR #14027, opened 2022-12-20
Fix build after core changes

J-N-K:
Related to https://github.com/openhab/openhab-core/pull/3241

Changes here are necessary because the test relied on a fixed order of the channels in the returned list. This change makes the test more robust.

As a side-note: I'm a bit astonished to see thing type upgrade code in the handler because my impression was that the policy is not to allow such code.

### DIFF
--- a/bundles/org.openhab.binding.wundergroundupdatereceiver/src/test/java/org/openhab/binding/wundergroundupdatereceiver/internal/WundergroundUpdateReceiverDiscoveryServiceTest.java
+++ b/bundles/org.openhab.binding.wundergroundupdatereceiver/src/test/java/org/openhab/binding/wundergroundupdatereceiver/internal/WundergroundUpdateReceiverDiscoveryServiceTest.java
@@ -107,22 +107,22 @@ class WundergroundUpdateReceiverDiscoveryServiceTest {
         // Then
         assertThat(actual.size(), is(9));
 
-        assertChannel(actual.get(0), METADATA_GROUP, LAST_RECEIVED, LAST_RECEIVED_DATETIME_CHANNELTYPEUID,
-                ChannelKind.STATE, is("DateTime"));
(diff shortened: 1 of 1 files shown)


jlaur:
> As a side-note: I'm a bit astonished to see thing type upgrade code in the handler because my impression was that the policy is not to allow such code.

Out of curiosity, do you mean this?
https://github.com/openhab/openhab-addons/blob/324483d8e97763ee2a24025a37a7f89e66367b27/bundles/org.openhab.binding.wundergroundupdatereceiver/src/main/java/org/openhab/binding/wundergroundupdatereceiver/internal/WundergroundUpdateReceiverHandler.java#L133-L142

What would be the recommended approach?

J-N-K:
A solution for https://github.com/openhab/openhab-core/issues/1924 (which superseded https://github.com/eclipse-archived/smarthome/issues/2555) would be the correct approach. Or the already known "delete and re-add".

jlaur:
> A solution for [openhab/openhab-core#1924](https://github.com/openhab/openhab-core/issues/1924) (which superseded [eclipse-archived/smarthome#2555](https://github.com/eclipse-archived/smarthome/issues/2555)) would be the correct approach. Or the already known "delete and re-add".

I just need to understand - until we have a solution for the referenced issue, what is it specifically we have a policy against? As an example, some time ago I added this:

https://github.com/openhab/openhab-addons/blob/df8e0bb15b7b71293eed55cc84700bd4a3e133dc/bundles/org.openhab.binding.danfossairunit/src/main/java/org/openhab/binding/danfossairunit/internal/handler/DanfossAirUnitHandler.java#L116-L129

and this:

https://github.com/openhab/openhab-addons/blob/df8e0bb15b7b71293eed55cc84700bd4a3e133dc/bundles/org.openhab.binding.hdpowerview/src/main/java/org/openhab/binding/hdpowerview/internal/handler/HDPowerViewHubHandler.java#L489-L528

would these kludgy backwards compatibility work-arounds also violate a policy? I'm removing them now in 4.0, but am still interested in best practices and policies here.

cweitkamp (approved):
LGTM. Thanks.

PR #14028, opened 2022-12-21
[opengarage] Add door transition status support

timcharper:
# Description

Homekit requires a status for the garage door of OPEN, CLOSED, CLOSING, OPENING. In order to report that, we must provide state transition information. State transition information is inferred when the garage door state is changed. For door_transition_time_seconds since the last open/close command was issued, the binding reports the state as either "closing" or "opening".


psmedley (approved):
This LGTM - I'd thought about implementing something like this myself!

bundles/org.openhab.binding.opengarage/src/main/java/org/openhab/binding/opengarage/internal/OpenGarageHandler.java
```diff
@@ -53,33 +53,42 @@ public class OpenGarageHandler extends BaseThingHandler {
...
     private @NonNullByDefault({}) OpenGarageWebTargets webTargets;
-    private @Nullable ScheduledFuture<?> pollFuture;
+    private @Nullable VariableDelayPoller poller;
+    private Instant lastTransition;
+    private String lastTransitionText;
+
+    public @NonNullByDefault({}) String doorOpeningState;
+    public @NonNullByDefault({}) String doorOpenState;
+    public @NonNullByDefault({}) String doorClosedState;
+    public @NonNullByDefault({}) String doorClosingState;
```
lolodomo, line 66: Rather replace these four lines by:
```
private OpenGarageConfiguration config = new OpenGarageConfiguration();
```
and then use config.xxx in your code.
Initialize `config` in initialize as you did.

bundles/org.openhab.binding.opengarage/src/main/java/org/openhab/binding/opengarage/internal/OpenGarageHandler.java
```diff
@@ -53,33 +53,42 @@ public class OpenGarageHandler extends BaseThingHandler {
     private final Logger logger = LoggerFactory.getLogger(OpenGarageHandler.class);
 
     private long refreshInterval;
+    private long transitionTimeSecs;
```
lolodomo, line 56: Same here, you could use the config object.
Applicable to transitionTimeSecs but also to refreshInterval.

bundles/org.openhab.binding.opengarage/src/main/java/org/openhab/binding/opengarage/internal/OpenGarageHandler.java
```diff
@@ -53,33 +53,42 @@ public class OpenGarageHandler extends BaseThingHandler {
...
+        this.lastTransitionText = "";
     }
 
     @Override
     public void handleCommand(ChannelUID channelUID, Command command) {
         try {
             logger.debug("Received command {} for thing '{}' on channel {}", command, thing.getUID().getAsString(),
                     channelUID.getId());
-            boolean invert = isChannelInverted(channelUID.getId());
+            var maybeInvert = getInverter(channelUID.getId());
```
lolodomo, line 79: I would prefer to avoid "var" in that place because with var you need to look at `getInverter` to guess the type.

bundles/org.openhab.binding.opengarage/src/main/java/org/openhab/binding/opengarage/internal/OpenGarageHandler.java
```diff
@@ -91,78 +100,80 @@ public void handleCommand(ChannelUID channelUID, Command command) {
...
-                case 1:
-                    updateState(OpenGarageBindingConstants.CHANNEL_OG_STATUS, invert ? OnOffType.OFF : OnOffType.ON);
-                    updateState(OpenGarageBindingConstants.CHANNEL_OG_STATUS_SWITCH,
-                            invert ? OnOffType.OFF : OnOffType.ON);
-                    updateState(OpenGarageBindingConstants.CHANNEL_OG_STATUS_ROLLERSHUTTER, UpDownType.UP);
-                    updateState(OpenGarageBindingConstants.CHANNEL_OG_STATUS_CONTACT, OpenClosedType.OPEN);
-                    break;
-                default:
-                    logger.warn("Received unknown door value: {}", controllerVariables.door);
+            var maybeInvert = getInverter(OpenGarageBindingConstants.CHANNEL_OG_STATUS_SWITCH);
```
lolodomo, line 152: I would prefer to avoid "var" in that place because with var you need to look at `getInverter` to guess the type.

bundles/org.openhab.binding.opengarage/src/main/java/org/openhab/binding/opengarage/internal/OpenGarageHandler.java
```diff
@@ -91,78 +100,80 @@ public void handleCommand(ChannelUID channelUID, Command command) {
 
     @Override
     public void initialize() {
-        OpenGarageConfiguration config = getConfigAs(OpenGarageConfiguration.class);
+        var config = getConfigAs(OpenGarageConfiguration.class);
```
lolodomo, line 103: I am not sure to understand the added value of this change.

bundles/org.openhab.binding.opengarage/src/main/java/org/openhab/binding/opengarage/internal/api/ControllerVariables.java
```diff
@@ -40,6 +40,11 @@ public class ControllerVariables {
     private ControllerVariables() {
     }
 
+    public static ControllerVariables empty() {
+        ControllerVariables info = new ControllerVariables();
+        return info;
+    }
```
lolodomo, line 46: Is it used ?
I have the feeling it is not.

lolodomo (changes requested):
Review part 1 of 2

bundles/org.openhab.binding.opengarage/src/main/java/org/openhab/binding/opengarage/internal/api/ControllerVariables.java
timcharper: Ah, indeed, it is not :) I'll remove it

timcharper:
Ok, @lolodomo , thank you for the feedback! I've incorporated your suggestions, and have rebased. (should I squash my commits or leave individual commits with "address feedback"? From the history, it looks like squash-merge is used on other PRs.

lolodomo:
Please do not squash.

bundles/org.openhab.binding.opengarage/src/main/java/org/openhab/binding/opengarage/internal/VariableDelayPoller.java
```diff
@@ -0,0 +1,88 @@
...
+    private long defaultPollSeconds;
+    private AtomicLong lastThreadId = new AtomicLong(0);
+    private Future<?> currentFuture;
+
+    /**
+     * @param scheduler - Reference to the Java Scheduler instance
+     * @param task - task to invoke. Task should return a number of seconds indicating when the next task should be
+     *            called
+     * @param defaultPollSeconds - how long to set the next poll should task fail to return a value due to an uncaught
+     *            exception
```
lolodomo, line 43: Please fix the description

bundles/org.openhab.binding.opengarage/src/main/java/org/openhab/binding/opengarage/internal/VariableDelayPoller.java
```diff
@@ -0,0 +1,88 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Variable poller will schedule a task to run periodically, but will allow the
+ * schedule to be modified on-the-fly
+ *
+ * @author Tim Harper - Initial contribution
+ */
+public class VariableDelayPoller {
```
lolodomo, line 30: Please add `@NonNullByDefault` and then add `@Nullable` when required in the rest of the class.

I have the strange feeling you made stuff complex with that class while not necessary, but let's assume you have a good reason I do not understand.

bundles/org.openhab.binding.opengarage/src/main/java/org/openhab/binding/opengarage/internal/VariableDelayPoller.java
```diff
@@ -0,0 +1,88 @@
...
+        lastThreadId.getAndIncrement(); // prevent future tasks from being scheduled
+        this.currentFuture.cancel(abortIfRunning);
+    }
+
+    private void doPoll(long threadId) {
+        var nextPollDuration = this.defaultPollSeconds;
+        try {
+            nextPollDuration = task.get();
+        } catch (Exception e) {
+            logger.error("error occurred while invoking periodic task", e);
```
lolodomo, line 63: Please do not use ERROR level, either DEBUG or WARN. Use WARN only if you are sure it could not flood the logs.

bundles/org.openhab.binding.opengarage/src/main/java/org/openhab/binding/opengarage/internal/OpenGarageHandler.java
```diff
@@ -52,34 +52,38 @@ public class OpenGarageHandler extends BaseThingHandler {
...
 
-    private long refreshInterval;
-
     private @NonNullByDefault({}) OpenGarageWebTargets webTargets;
-    private @Nullable ScheduledFuture<?> pollFuture;
+    private @Nullable VariableDelayPoller poller;
+    private Instant lastTransition;
+    private String lastTransitionText;
+
+    private @NonNullByDefault({}) OpenGarageConfiguration config;
```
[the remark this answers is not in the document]
timcharper: That's doable, curious as to the logic behind that though. A few lines above, it uses the same convention for a variable that is initialized similarly:

https://github.com/openhab/openhab-addons/pull/14028/files/11941dbb223fd13f5d7e6f0eff0f0be2a08df8f6#diff-3d1db1768fb745c0c9a929f89358366f026a755103bed1745f6c119f7fb074e9R55

bundles/org.openhab.binding.opengarage/src/main/java/org/openhab/binding/opengarage/internal/VariableDelayPoller.java
timcharper: Yes, I have a hunch you are right, and felt this code needed justification. I wrote up a long response to justify it, but after writing it all, I think the reason was excessive paranoia about concurrency and potential deadlocks. I think a simple approach will be fine.

justification I wrote:

 It may be a case of being excessively paranoia about multi-threading and not entirely sure what I can trust to not create a deadlock scenario. I can probably be talked out of this.

With the previous implementation, we would simply run the poll on a loop and be done with it. Ezpz. Scheduler handles it.

- 00:01 - Poll
- 00:30 - Door open command
- 00:31 - Poll
- 01:01 - Poll
- 01:31 - Poll


With the door transition state change, we need to additionally update the state in response to a door state command (in order to convey the transition state wanted by HomeKit), and then, again, when the door transition is completed. Let's assume a transition time of 17 seconds. Here's what the schedule would look like without my variable poller:

- 00:01 - Poll
- 00:30 - Door open command
- 00:30 - "Poll" (in response to door command)
- 00:31 - Poll (on schedule)
- 00:47 - Poll (in response to earlier door command, to publish door transition completion)
- 01:01 - Poll
- 01:31 - Poll

Now, if two of these poll methods happen to run concurrently, then it could cause some trouble. Say 00:30 poll invocation starts, and while it's running, the 00:31 poll invocation runs and completes before the 00:30 poll invocation completes. We'd end up with the state of 00:30 overriding the state of 00:31. Probably relatively harmless? I don't know. Then, on shutdown, we have extra future we have to potentially cancel.

If we put synchronized around the pollStatus loop, we risk clogging the thread pool with up to 3 threads because one thread is taking 30 seconds to time out.

I don't know, the more I'm trying to explain this the less justified I feel it really is. I'm going to try and delete it.

timcharper: I removed it, I'm pretty confident it'll be fine just using synchronized and an extra one-of scheduled task.

timcharper:
@lolodomo thank you for the full review! I've taken all of your suggestions to heart and I think the pull request is much better than when it started. Getting rid of the custom poll logic was a good move, I think the downsides of synchronizing on poll are better than the downsides of having custom concurrency code.

timcharper:
@lolodomo gentle nudge on this :)

bundles/org.openhab.binding.opengarage/src/main/java/org/openhab/binding/opengarage/internal/OpenGarageHandler.java
```diff
@@ -91,103 +106,104 @@ public void handleCommand(ChannelUID channelUID, Command command) {
...
+        this.config = getConfigAs(OpenGarageConfiguration.class);
         logger.debug("config.hostname = {}, refresh = {}, port = {}", config.hostname, config.refresh, config.port);
         if (config.hostname == null) {
             updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.CONFIGURATION_ERROR, "Hostname/IP address must be set");
         } else {
-            webTargets = new OpenGarageWebTargets(config.hostname, config.port, config.password);
-            refreshInterval = config.refresh;
-
-            schedulePoll();
+            int requestTimeout = Math.max(OpenGarageWebTargets.DEFAULT_TIMEOUT_MS, config.refresh * 10000);
```
lolodomo, line 114: Why 10000 ?
You want to convert seconds into milliseconds ?

bundles/org.openhab.binding.opengarage/src/main/java/org/openhab/binding/opengarage/internal/OpenGarageHandler.java
```diff
@@ -91,103 +106,104 @@ public void handleCommand(ChannelUID channelUID, Command command) {
...
+            webTargets = new OpenGarageWebTargets(config.hostname, config.port, config.password, requestTimeout);
+            this.pollScheduledFuture = this.scheduler.scheduleWithFixedDelay(this::poll, 1, config.refresh,
```
lolodomo, line 116: Please set the status to UNKNOW before scheduling your task, the core framework expects a status to be set to ONLINE/OFFLINE/UNKNOWN when returning from initialize()

bundles/org.openhab.binding.opengarage/src/main/java/org/openhab/binding/opengarage/internal/OpenGarageHandler.java
```diff
@@ -91,78 +95,73 @@ public void handleCommand(ChannelUID channelUID, Command command) {
...
+    private long poll() {
+        var pollAgainIn = this.config.refresh;
         try {
             logger.debug("Polling for state");
-            pollStatus();
+            pollAgainIn = pollStatus();
         } catch (IOException e) {
-            logger.debug("Could not connect to OpenGarage controller", e);
+            logger.warn("Could not connect to OpenGarage controller", e);
             updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.COMMUNICATION_ERROR, e.getMessage());
```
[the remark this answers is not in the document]
lolodomo: Keep a DEBUG log,

bundles/org.openhab.binding.opengarage/src/main/java/org/openhab/binding/opengarage/internal/OpenGarageHandler.java
```diff
@@ -91,103 +106,104 @@ public void handleCommand(ChannelUID channelUID, Command command) {
...
+            var inTransition = lastTransitionAgoSecs < this.config.doorTransitionTimeSeconds;
+            if (controllerVariables != null) {
+                updateStatus(ThingStatus.ONLINE);
+                updateState(OpenGarageBindingConstants.CHANNEL_OG_DISTANCE,
+                        new QuantityType<>(controllerVariables.dist, MetricPrefix.CENTI(SIUnits.METRE)));
+                Function<Boolean, Boolean> maybeInvert = getInverter(
+                        OpenGarageBindingConstants.CHANNEL_OG_STATUS_SWITCH);
+
+                if ((controllerVariables.door != 0) && (controllerVariables.door != 1)) {
+                    logger.warn("Received unknown door value: {}", controllerVariables.door);
```
lolodomo, line 147: A DEBUG log looks more appropriate.

bundles/org.openhab.binding.opengarage/src/main/java/org/openhab/binding/opengarage/internal/OpenGarageHandler.java
```diff
@@ -91,103 +106,104 @@ public void handleCommand(ChannelUID channelUID, Command command) {
...
+                        break;
+                    case 1:
+                        updateState(OpenGarageBindingConstants.CHANNEL_OG_VEHICLE, new StringType("Vehicle detected"));
+                        break;
+                    case 2:
+                        updateState(OpenGarageBindingConstants.CHANNEL_OG_VEHICLE,
+                                new StringType("Vehicle status unknown"));
+                        break;
+                    default:
+                        logger.warn("Received unknown vehicle value: {}", controllerVariables.vehicle);
```
lolodomo, line 182: A DEBUG log looks more appropriate.

bundles/org.openhab.binding.opengarage/src/main/resources/OH-INF/i18n/opengarage.properties
```diff
@@ -18,6 +18,16 @@ thing-type.config.opengarage.opengarage.port.label = Port
...
+thing-type.config.opengarage.opengarage.doorTransitionTimeSeconds.label = Door Transition Time
+thing-type.config.opengarage.opengarage.doorTransitionTimeSeconds.description = Specifies how long it takes the garage door to fully open / close. Include time it takes for OpenHab to emit beeps. Round up.
+thing-type.config.opengarage.opengarage.doorOpeningState.label = Door Opening State
+thing-type.config.opengarage.opengarage.doorOpeningState.description = Text state to report when garage is opening
+thing-type.config.opengarage.opengarage.doorOpenState.label = Door Open State
+thing-type.config.opengarage.opengarage.doorOpenState.description = Text state to report when garage is open (and not transitioning)
+thing-type.config.opengarage.opengarage.doorClosingState.label = Door Closing State
+thing-type.config.opengarage.opengarage.doorClosingState.description = Text state to report when garage is closing
+thing-type.config.opengarage.opengarage.doorClosedState.label = Door Closed State
+thing-type.config.opengarage.opengarage.doorClosedState.description = Text state to report when garage is closed (and not transitioning)
```
lolodomo, line 30: Please make consistent the descriptions here and in the thing description (and in the documentation). This is not the case.

bundles/org.openhab.binding.opengarage/src/main/resources/OH-INF/i18n/opengarage.properties
```diff
@@ -31,6 +41,8 @@ channel-type.opengarage.opengarage-status-switch.label = Status
 channel-type.opengarage.opengarage-status-switch.description = On/Off Status of the OG unit
 channel-type.opengarage.opengarage-status.label = Status
 channel-type.opengarage.opengarage-status.description = On/Off Status of the OG unit (now deprecated, use status-switch instead)
+channel-type.opengarage.opengarage-status-text.label = Text status
+channel-type.opengarage.opengarage-status-text.description = Describe the current state textually, including the current transition (opening / closing)
```
lolodomo, line 45: Also this one.

lolodomo (changes requested):

lolodomo:
> @lolodomo gentle nudge on this :)

Done. Should be the last run ;)

Take care, apparently you missed some of my comments. You have to click on "Read more...". I marked some of them as resolved when they are now outdated considering your last changes.
Some of my new comments are probably redundant with old remaining comments as I started a full new review.

bundles/org.openhab.binding.opengarage/src/main/java/org/openhab/binding/opengarage/internal/OpenGarageHandler.java
```diff
@@ -52,34 +54,47 @@ public class OpenGarageHandler extends BaseThingHandler {
...
-                        return;
+                    } else {
+                        var doorOpen = command.equals(OnOffType.ON) || command.equals(UpDownType.UP);
+                        changeStatus(maybeInvert.apply(doorOpen) ? OpenGarageCommand.OPEN : OpenGarageCommand.CLOSE);
+                        this.lastTransition = Instant.now();
+                        this.lastTransitionText = doorOpen ? this.config.doorOpeningState
+                                : this.config.doorClosingState;
+
+                        this.poll(); // invoke poll directly to communicate the door transition state
+                        this.pollScheduledFutureTransition.cancel(false);
```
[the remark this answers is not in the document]
timcharper: correct

bundles/org.openhab.binding.opengarage/src/main/java/org/openhab/binding/opengarage/internal/OpenGarageHandler.java
timcharper: yeah, that's wrong

```suggestion
            int requestTimeout = Math.max(OpenGarageWebTargets.DEFAULT_TIMEOUT_MS, config.refresh * 1000);
```

bundles/org.openhab.binding.opengarage/src/main/java/org/openhab/binding/opengarage/internal/OpenGarageHandler.java
timcharper:
```suggestion
                    logger.debug("Received unknown door value: {}", controllerVariables.door);
```

bundles/org.openhab.binding.opengarage/src/main/java/org/openhab/binding/opengarage/internal/OpenGarageHandler.java
timcharper:
```suggestion
                        logger.debug("Received unknown vehicle value: {}", controllerVariables.vehicle);
```

bundles/org.openhab.binding.opengarage/src/main/java/org/openhab/binding/opengarage/internal/OpenGarageHandler.java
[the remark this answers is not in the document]
timcharper:
```suggestion
            logger.debug("Could not connect to OpenGarage controller", e);
            updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.COMMUNICATION_ERROR, e.getMessage());
```

bundles/org.openhab.binding.opengarage/src/main/java/org/openhab/binding/opengarage/internal/OpenGarageHandler.java
```diff
@@ -91,103 +106,104 @@ public void handleCommand(ChannelUID channelUID, Command command) {
...
+                }
+                updateState(OpenGarageBindingConstants.CHANNEL_OG_VEHICLE_STATUS,
+                        new DecimalType(controllerVariables.vehicle));
+            }
         } catch (IOException e) {
-            logger.debug("Could not connect to OpenGarage controller", e);
+            logger.warn("Could not connect to OpenGarage controller", e);
             updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.COMMUNICATION_ERROR, e.getMessage());
         } catch (RuntimeException e) {
             logger.warn("Unexpected error connecting to OpenGarage controller", e);
```
timcharper, line 191:
```suggestion
            logger.debug("Unexpected error connecting to OpenGarage controller", e);
```

jlaur:
@timcharper - can you resolve the conflict and address @lolodomo's comments? I think this PR could then still make it into 4.0.

timcharper:
@lolodomo I've addressed all of your comments, I believe. Can you give this the ol' green check mark?

bundles/org.openhab.binding.opengarage/src/main/java/org/openhab/binding/opengarage/internal/OpenGarageConfiguration.java
```diff
@@ -12,14 +12,24 @@
...
 /**
  * The OpenGarageConfiguration class contains fields mapping thing configuration parameters.
  *
  * @author Paul Smedley - Initial contribution
  */
+@NonNullByDefault
 public class OpenGarageConfiguration {
-    public String hostname;
-    public long port = 80;
+    public @Nullable String hostname;
```
lolodomo, line 25: `hostname` will never be null because you defined the parameter as required.
The best practice is to not defined it as `@Nullable` and set it to an empty string by default. In your thing handler (initialize), just check if its value is not empty.

bundles/org.openhab.binding.opengarage/src/main/java/org/openhab/binding/opengarage/internal/OpenGarageHandler.java
```diff
@@ -91,107 +106,111 @@ public void handleCommand(ChannelUID channelUID, Command command) {
 
     @Override
     public void initialize() {
-        OpenGarageConfiguration config = getConfigAs(OpenGarageConfiguration.class);
+        this.config = getConfigAs(OpenGarageConfiguration.class);
         logger.debug("config.hostname = {}, refresh = {}, port = {}", config.hostname, config.refresh, config.port);
         if (config.hostname == null) {
```
lolodomo, line 111: As previously mentioned, it will never be null. Rather test an empty string.

bundles/org.openhab.binding.opengarage/src/main/java/org/openhab/binding/opengarage/internal/OpenGarageHandler.java
```diff
@@ -91,107 +106,111 @@ public void handleCommand(ChannelUID channelUID, Command command) {
...
+            int requestTimeout = Math.max(OpenGarageWebTargets.DEFAULT_TIMEOUT_MS, config.refresh * 1000);
+            webTargets = new OpenGarageWebTargets(config.hostname, config.port, config.password, requestTimeout);
+            this.pollScheduledFuture = this.scheduler.scheduleWithFixedDelay(this::poll, 1, config.refresh,
+                    TimeUnit.SECONDS);
         }
     }
 
     @Override
     public void dispose() {
         super.dispose();
```
lolodomo, line 124: I would rather call the super method at the end of the method after stopping the scheduled tasks.

bundles/org.openhab.binding.opengarage/src/main/java/org/openhab/binding/opengarage/internal/OpenGarageHandler.java
```diff
@@ -91,107 +106,111 @@ public void handleCommand(ChannelUID channelUID, Command command) {
...
-
     private void changeStatus(OpenGarageCommand status) throws OpenGarageCommunicationException {
         webTargets.setControllerVariables(status);
     }
 
-    private boolean isChannelInverted(String channelUID) {
+    private Function<Boolean, Boolean> getInverter(String channelUID) {
         Channel channel = getThing().getChannel(channelUID);
-        return channel != null && channel.getConfiguration().as(OpenGarageChannelConfiguration.class).invert;
+        var invert = channel != null && channel.getConfiguration().as(OpenGarageChannelConfiguration.class).invert;
```
lolodomo, line 209: I suggest to replace var by boolean.

bundles/org.openhab.binding.opengarage/src/main/java/org/openhab/binding/opengarage/internal/OpenGarageHandler.java
```diff
@@ -52,34 +54,47 @@ public class OpenGarageHandler extends BaseThingHandler {
...
-                        return;
-                    } else if (command.equals(OnOffType.OFF) || command.equals(UpDownType.DOWN)) {
-                        changeStatus(invert ? OpenGarageCommand.OPEN : OpenGarageCommand.CLOSE);
-                        return;
-                    } else if (command.equals(StopMoveType.STOP) || command.equals(StopMoveType.MOVE)) {
+                    if (command.equals(StopMoveType.STOP) || command.equals(StopMoveType.MOVE)) {
                         changeStatus(OpenGarageCommand.CLICK);
-                        return;
+                    } else {
+                        var doorOpen = command.equals(OnOffType.ON) || command.equals(UpDownType.UP);
```
lolodomo, line 88: I suggest to replace var by boolean

lolodomo (changes requested):
Full new review of the last code.

lolodomo:
@timcharper : I am sorry, there are still few things to improve, but not too much.

bundles/org.openhab.binding.opengarage/src/main/resources/OH-INF/thing/thing-types.xml
```diff
@@ -16,6 +16,7 @@
 			<channel id="status-rollershutter" typeId="opengarage-status-rollershutter"/>
 			<channel id="vehicle" typeId="opengarage-vehicle"/>
 			<channel id="vehicle-status" typeId="opengarage-vehicle-status"/>
+			<channel id="status-text" typeId="opengarage-status-text"/>
```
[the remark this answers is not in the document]
timcharper: I don't understand why this is necessary, and see no mention of upgrade.xml anywhere in the repo. Perhaps you mean something else?

timcharper:
I've addressed what I can, but I cannot find any information on this `upgrade.xml` file you've mentioned. I've scanned the documentation, and also couldn't find mention of it.

I see a handful of `update.xml` files in the repo, but I don't know that that is what you're referring to.

lolodomo:
https://www.openhab.org/docs/developer/bindings/thing-xml.html#updating-thing-types

bundles/org.openhab.binding.opengarage/src/main/resources/OH-INF/thing/thing-types.xml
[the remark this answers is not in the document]
lolodomo: https://www.openhab.org/docs/developer/bindings/thing-xml.html#updating-thing-types

bundles/org.openhab.binding.opengarage/src/main/java/org/openhab/binding/opengarage/internal/OpenGarageHandler.java
```diff
@@ -91,107 +106,111 @@ public void handleCommand(ChannelUID channelUID, Command command) {
 
     @Override
     public void initialize() {
-        OpenGarageConfiguration config = getConfigAs(OpenGarageConfiguration.class);
+        this.config = getConfigAs(OpenGarageConfiguration.class);
         logger.debug("config.hostname = {}, refresh = {}, port = {}", config.hostname, config.refresh, config.port);
-        if (config.hostname == null) {
+        if (config.hostname == "") {
```
lolodomo, line 111:
```
if (config.hostname.isEmpty()) {
```

lolodomo (changes requested):

bundles/org.openhab.binding.opengarage/src/main/java/org/openhab/binding/opengarage/internal/OpenGarageHandler.java
lolodomo: Note that is the line that leads to the failing build. By fixing it, the build should work again.

lolodomo:
Remains one thing to fix + upgrade instructions to provide for the new channel and then I can merge.

bundles/org.openhab.binding.opengarage/src/main/resources/OH-INF/update/instructions.xml
```diff
@@ -0,0 +1,18 @@
+<?xml version="1.0" encoding="UTF-8" standalone="yes" ?>
+<update:update-descriptions xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+	xmlns:update="https://openhab.org/schemas/update-description/v1.0.0"
+	xsi:schemaLocation="https://openhab.org/schemas/update-description/v1.0.0 https://openhab.org/schemas/update-description-1.0.0.xsd">
+
+	<thing-type uid="opengarage:opengarage">
+
+		<instruction-set targetVersion="1">
+			<add-channel id="status-text">
+				<type>opengarage:status-text</type>
```
lolodomo, line 10: Thing type is rather: `opengarage:opengarage-status-text`

bundles/org.openhab.binding.opengarage/src/main/resources/OH-INF/update/instructions.xml
```diff
@@ -0,0 +1,18 @@
...
+				<label>Status Text</label>
+				<description>Text status of the current door state</description>
```
lolodomo, line 12: No need to set label/description, those from the channel type will then be used.

bundles/org.openhab.binding.opengarage/src/main/resources/OH-INF/update/instructions.xml
```diff
@@ -0,0 +1,18 @@
+<?xml version="1.0" encoding="UTF-8" standalone="yes" ?>
+<update:update-descriptions xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+	xmlns:update="https://openhab.org/schemas/update-description/v1.0.0"
+	xsi:schemaLocation="https://openhab.org/schemas/update-description/v1.0.0 https://openhab.org/schemas/update-description-1.0.0.xsd">
+
+	<thing-type uid="opengarage:opengarage">
+
+		<instruction-set targetVersion="1">
```
lolodomo, line 8: You forget to add the thingTypeVersion property in the thing type opengarage:opengarage

lolodomo (changes requested):

lolodomo:
To accelerate the process and have this PR released in OH 4.1, I just pushed the fix for upgrade instructions.

lolodomo (approved):

timcharper:
Thank you @lolodomo ! Wow, 1 year later :) That's persistence!

PR #14021, opened 2022-12-20
[dali] handle and provide QuantityType for color-temperature-abs channel

ccutrer:
see https://github.com/openhab/openhab-core/pull/3129

PR #14029, opened 2022-12-21
[dali] tweak color temperature abs QuantityType fix

ccutrer:
adds an example of usage of the channel to the README, and simplifies one method call

follow-up to #14021, applying analogous changes to feedback received on similar PRs for other bindings.


lolodomo (approved):
LGTM

PR #14030, opened 2022-12-21
[danfossairunit] Remove deprecated channel 'manual_fan_speed'

jlaur:
Related to:
- #11147
- #11668

Now finally removing this channel which was deprecated in 3.2.


lsiepel:
Does this need a mention in the breaking change notices ? A warning has been issued before. But because the channel remained available, we might need to show it again?
https://github.com/openhab/openhab-distro/blob/00bc9270035c98a86b7c4bbc782273d4fab04a31/distributions/openhab/src/main/resources/bin/update.lst#L50

Otherwise LGTM

jlaur:
> Does this need a mention in the breaking change notices ? A warning has been issued before. But because the channel remained available, we might need to show it again?

Yes, I plan to do that, just didn't get around to create the PR yet. Thanks for the reminder. :-)

jlaur:
Created openhab/openhab-distro#1449.

lolodomo (approved):
LGTM

lolodomo:
@pravussum : please have a look.

jlaur:
@lolodomo - I think we can take a cross-binding decision about which breaking changes we can accept for 4.0 in order to clean up the code. #14033 has been merged, and this one is analogue to that: In both cases channels were deprecated in 3.2, but still supported (with deprecation warnings in log). And also in both cases they were mentioned in the 3.2 upgrade notes. So it would be inconsistent to merge one, but not the other. If you agree, please merge. Otherwise let's find some common ground. 🙂

pravussum (approved):
LGTM

Issue #13595, opened 2022-10-24
[HomeKit] Use Number item as Value for Heater/Cooler, Thermostat current and target mode


grzegorz914:
Hi,

Heater/Cooler and Thermostat accessory can be used with string item to control current/target mode, it will be very help full to add possibility use number item. Right now I need convert thing channel number to string item and vice versa.

Thanks
Grzegorz

ccutrer:
FYI @grzegorz914: #14036 is now ready for review, and fully closes this issue (it took a while to get dependent semi-related PRs through first).

grzegorz914:
@ccutrer any progress?

ccutrer:
The PR is up and linked from this issue, and there are comments going back and forth.

grzegorz914:
@ccutrer why the PR is pause, any chance we get any progress wit it?

PR #14036, opened 2022-12-22
[homekit] remove global configuration of enum values

ccutrer:
use metadata configuration only. in particular, for GarageDoorOpener, the global settings didn't even actually work, so they should work now.

this is in preparation for working on #13595


ccutrer:
@timcharper: can you test this GarageDoorOpener portion of this?

ccutrer:
this is not currently working. it needs a helper that's in a PR that's not yet merged; I'm just going to back-burner this until its dependency is merged, then I'll update this PR

yfre:
agree, OH 4.0 is a good opportunity to remove legacy things especially if they dont work

bundles/org.openhab.io.homekit/README.md
```diff
@@ -686,263 +678,267 @@ or using UI
...
+All enum values can be customized via item metadata. I.e. `HEAT="heating", COOL="cooling"`, or `HEAT=5, COOL=7` for a Number.
+<a id="customizeable-enum">Some enums can have the list of valid values customized, meaning that if you customize the mapping, any value that is missing will not be presented to the user.</a>
+They are appropriately marked.
+Enums that are linked to Switches or Contacts have an `inverted` param that will reverse the sense of `ON`/`OFF` or `OPEN`/`CLOSED`.
+
+| Accessory Tag        | Mandatory Characteristics   | Optional Characteristics    | Supported openHAB item types            | Description                                                                                                                                                                                                                                                                                                                                                   | Configuration Options                                                 | Valid Enum Values                                                                                           |
```
ccutrer, line 690: The main changes in this table are:
 * adding additional columns for configuration options and enum values, and moving any references to those things in the description column as appropriate, adopting a consistent formatting and ordering for enum values, as well as item types.
 * Sorting the entire table by accessory type name, then required and optional characteristics, and then characteristic name
 * Ensure all relevant item types are listed for enums (and they're sorted), as appropriate (I didn't list Switch or Contact for enums that didn't seem to have a reasonable use case for only using the first two values, even though technically the code would allow it)

timcharper (changes requested):
Garage door support is broken, please hold merging until we resolve

ccutrer:
I re-reviewed the changes here, and confirmed that garage door is the only enum that previously had custom handling that now needs to be inverted.

ccutrer:
@timcharper : can you confirm with the latest commit your garage door opener is good now?

bundles/org.openhab.io.homekit/src/main/java/org/openhab/io/homekit/internal/accessories/AbstractHomekitAccessoryImpl.java
```diff
@@ -267,29 +268,38 @@ protected <T> T getAccessoryConfiguration(HomekitCharacteristicType characterist
...
-        });
+    protected <T extends Enum<T> & CharacteristicEnum> Map<T, String> createMapping(
+            HomekitCharacteristicType characteristicType, Class<T> klazz) {
+        return createMapping(characteristicType, klazz, null, false);
+    }
+
+    @NonNullByDefault
+    protected <T extends Enum<T> & CharacteristicEnum> Map<T, String> createMapping(
+            HomekitCharacteristicType characteristicType, Class<T> klazz, boolean inverted) {
+        return createMapping(characteristicType, klazz, null, true);
```
timcharper, line 280: Did you mean to omit inverted here?

```suggestion
        return createMapping(characteristicType, klazz, null, inverted);
```

timcharper (commented):
Wondering if inverting the right approach? Did the original invert everything? (Shutters, etc. or just switches?)

Perhaps we could have a Map<OpenHabValue, T> that we pass to override default mappings? Or something to evaluate "truthiness" of different OpenHab values?

I'm also wondering if it makes sense to invert an inversion, double negatives could be a source of confusion that perhaps a mapping could avoid?

ccutrer:
I’m not sure I follow about a partial invert. It’s still the case that “invert” (whether being prescribed by the createMapping call in code, or by user config on the metadata) only affects Switch and Contact items. Both before and after this commit. Before this commit it was done in how parameters are passed to `createBooleanReader` (which is still there for the moment, handling a couple of truly boolean attributes, that aren't backed by an enum).

As for double-invert, I don't see any confusion at all. The `inverted` passed to `createMapping` is a constant decided on by the addon for "we disagree with the HAP specification on their enum values for 0 and 1, so swap them", whereas the user-level invert is "my items happen to be opposite of what normally makes sense for linking this particular item to this particular characteristic" (I think this happens most often in practice with linking a Switch item to a security/sensor characteristic  where OPEN/CLOSED would mean "insecure"/"secure", and would normally be ON to mean "something is detected", but your particular sensor thinks about things the other way, or just happens to be translated the other way due to levels of abstraction by the time the actual value makes it to an item state in openHAB). It's a single conditional in the code that I don't think of as "double-invert" but instead as "combine the two types of inverting into a final value of if we're inverting ON/OFF from HAP or not", meaning the actual inversion code doesn't need to worry about how we arrived at _why_ we're inverting or not, it just does a simple inversion if necessary. I can definitely improve the javadocs on createMapping to make it clear the intent of the `inverted` method parameter vs. the isInverted() coming from the HomekitTaggedItem.

Note also that this commit does _not_ affect AbstractHomekitPositionAccessoryImpl at all which uses inverted not in a boolean sense, but in a 0-100 or 100-0 sense for translating dimmers especially, but also numbers, and even possibly rollershutters, to a full range of "open" to "close" where it's completely arbitrary which end is assigned to which (except that openHAB defines Rollershutters having 0 as OPEN, and HomeKit defines 0 as CLOSED, so we also always invert there by default, doing a "double" inversion as you say if the user requests it because their Dimmer item already matches HomeKit semantics, but not openHAB semantics).

bundles/org.openhab.io.homekit/src/main/java/org/openhab/io/homekit/internal/accessories/AbstractHomekitAccessoryImpl.java
ccutrer: good catch

timcharper (approved):

timcharper:
@ccutrer sounds like you've thought about this deeply. I trust your judgement on it.

I had thought that the GarageDoor implementation supported a roller shutter, but I think I'm conflating it with the OpenGarage binding instead.

Just tested it without inverting the garage door switch in the configuration, it works perfectly.

yfre (approved):
it looks for me some time to understand how you done. pretty smart. using HAP enum for default mapping but especially non-string item mapping. cool. saved many lines of code. thank you

PR LGTM

ccutrer:
@lolodomo can I get a review/merge on this please?

jlaur (approved):
Many thanks!

PR #14041, opened 2022-12-22
Fix "Hello, World!" examples in automation documentation

wborn:
The documentation states incorrectly that ["Hello, World!"](https://en.wikipedia.org/wiki/%22Hello,_World!%22_program) is printed when instead "Hello world!" is printed.


jlaur (approved):
Nice catch. :-) Thanks.

PR #14045, opened 2022-12-23
[homekit] support Rollershutter items for HoldPosition

ccutrer:
just send STOP to them

Note that HoldPosition is not exposed by the native Home app, but it is supported by the Elgato Eve app.


bundles/org.openhab.io.homekit/src/main/java/org/openhab/io/homekit/internal/accessories/HomekitCharacteristicFactory.java
```diff
@@ -471,7 +474,17 @@ private static NameCharacteristic createNameCharacteristic(HomekitTaggedItem tag
...
             HomekitAccessoryUpdater updater) {
-        return new HoldPositionCharacteristic(value -> ((SwitchItem) taggedItem.getItem()).send(OnOffType.from(value)));
+        return new HoldPositionCharacteristic(value -> {
+            if (!value) {
+                return;
+            }
+            Item item = taggedItem.getBaseItem();
+            if (item instanceof SwitchItem) {
+                ((SwitchItem) item).send(OnOffType.ON);
+            } else if (item instanceof RollershutterItem) {
```
timcharper, line 484: possible for item to be neither of these? Maybe we should add a final else close and log an error?

yfre:
oh, there are apps that are actually using holdPosition.
good PR. it is complementary to https://github.com/openhab/openhab-addons/pull/13788 that emulates STOP for home app

probably we need to add guidance to documentation when to use which options

ccutrer:
> probably we need to add guidance to documentation when to use which options

I've got a followup PR (that you might not see for a while, since it depends on the secondary services PR) that does exactly that.

bundles/org.openhab.io.homekit/src/main/java/org/openhab/io/homekit/internal/accessories/HomekitCharacteristicFactory.java
ccutrer: good call. will do (but slightly differently, so the warning comes as soon as the accessory is created, not later when a Hold is actually received).

yfre (approved):
LGTM

timcharper (approved):

lolodomo (approved):
LGTM, thank you

PR #14048, opened 2022-12-23
Remove JavaScript Transformation pom.xml

wborn:
Everything except for this pom.xml got removed in openhab/openhab-addons#13276


jlaur (approved):
Thanks.

PR #14050, opened 2022-12-23
[knx] Improve reading of device properties

holgerfriedrich:
Extend the code reading the KNX device properties:
Add reading additional properties which help to identify a KNX device (and help you to identify when you configured the individual address wrong ;-) ).
Display the APDU size, which tells you if KNX secure can work with your line couplers....
Rename properties (typo).
Remove Firmware.java as functionality is provided by Calimero.
Try blocks around optional properties allow to continue reading others after error.

btw: is there a useful way to "sort" the properties? Inserting in the correct order does not help, using LinkedHashMap did not help as well.


bundles/org.openhab.binding.knx/src/main/java/org/openhab/binding/knx/internal/client/DeviceInspector.java
```diff
@@ -20,24 +20,27 @@
 import java.util.Map;
 import java.util.Set;
 
+import org.apache.commons.lang3.StringUtils;
```
jlaur, line 23: Please avoid this import, see #7722.

bundles/org.openhab.binding.knx/src/main/java/org/openhab/binding/knx/internal/KNXBindingConstants.java
```diff
@@ -38,13 +38,16 @@ public class KNXBindingConstants {
...
 
     // Property IDs
-    public static final String FIRMWARE_TYPE = "firmwaretype";
-    public static final String FIRMWARE_VERSION = "firmwareversion";
-    public static final String FIRMWARE_SUBVERSION = "firmwaresubversion";
-    public static final String MANUFACTURER_NAME = "manfacturername";
-    public static final String MANUFACTURER_SERIAL_NO = "manfacturerserialnumber";
-    public static final String MANUFACTURER_HARDWARE_TYPE = "manfacturerhardwaretype";
-    public static final String MANUFACTURER_FIRMWARE_REVISION = "manfacturerfirmwarerevision";
+    public static final String DEVICE_MASK_VERSION = "devicemaskversion";
```
jlaur, line 41: I would suggest to prefix these constants with `PROPERTY_` also also use camelCase for their contents, e.g.:
```suggestion
    public static final String PROPERTY_DEVICE_MASK_VERSION = "deviceMaskVersion";
```

bundles/org.openhab.binding.knx/src/main/java/org/openhab/binding/knx/internal/client/DeviceInspector.java
```diff
@@ -102,15 +105,125 @@ private Map<String, String> readDeviceProperties(IndividualAddress address) {
...
+                }
+
+                Thread.sleep(OPERATION_INTERVAL);
+                byte[] orderInfo = getClient().readDeviceProperties(address, DEVICE_OBJECT, PID.ORDER_INFO, 1, 1, false,
+                        OPERATION_TIMEOUT);
+                if (orderInfo != null) {
+                    final String hexString = toHex(orderInfo, "");
+                    if ((!"ffffffffffffffffffff".equals(hexString)) && (!"00000000000000000000".equals(hexString))) {
+                        String result = new String(orderInfo);
+                        // result = result.trim();
```
jlaur, line 181: Left-over?

jlaur (changes requested):
Thanks. I didn't study the code closely yet, but hereby provide some initial feedback.

holgerfriedrich:
> Thanks. I didn't study the code closely yet, but hereby provide some initial feedback.

Thanks, @jlaur! I have removed ```StringUtils``` as requested.

bundles/org.openhab.binding.knx/src/main/java/org/openhab/binding/knx/internal/client/DeviceInspector.java
```diff
@@ -102,15 +104,128 @@ private Map<String, String> readDeviceProperties(IndividualAddress address) {
...
+                Thread.sleep(OPERATION_INTERVAL);
+                String firmwareRevision = null;
+                try {
+                    byte[] result = getClient().readDeviceProperties(address, DEVICE_OBJECT, PID.FIRMWARE_REVISION, 1,
+                            1, false, OPERATION_TIMEOUT);
+                    if (result != null) {
+                        firmwareRevision = Integer.toString(toUnsigned(result));
+                    } else {
+                        // try fallback to PID_VERSION
+                        result = getClient().readDeviceProperties(address, DEVICE_OBJECT, PID.VERSION, 1, 1, false,
```
lolodomo, line 126: No sleep (OPERATION_INTERVAL) required before new request ?

bundles/org.openhab.binding.knx/src/main/java/org/openhab/binding/knx/internal/client/DeviceInspector.java
```diff
@@ -102,15 +104,128 @@ private Map<String, String> readDeviceProperties(IndividualAddress address) {
...
+                                OPERATION_TIMEOUT);
+                        if (result != null) {
+                            // data format is DPT217.001
+                            int i = toUnsigned(result);
+                            firmwareRevision = Integer.toString((i & 0xF800) >> 11) + "."
+                                    + Integer.toString((i & 0x07C0) >> 6) + "." + Integer.toString((i & 0x003F));
+                        }
+                    }
+                } catch (Exception e) {
```
lolodomo, line 135: No chance to catch more specialized exceptions ?
You could also name the variable `ignore` instead of `e`
Same for others

bundles/org.openhab.binding.knx/src/main/java/org/openhab/binding/knx/internal/client/DeviceInspector.java
```diff
@@ -102,15 +104,128 @@ private Map<String, String> readDeviceProperties(IndividualAddress address) {
...
+                        maxApdu = "14"; // see spec
+                    }
+                }
+
+                Thread.sleep(OPERATION_INTERVAL);
+                byte[] orderInfo = getClient().readDeviceProperties(address, DEVICE_OBJECT, PID.ORDER_INFO, 1, 1, false,
+                        OPERATION_TIMEOUT);
+                if (orderInfo != null) {
+                    final String hexString = toHex(orderInfo, "");
+                    if ((!"ffffffffffffffffffff".equals(hexString)) && (!"00000000000000000000".equals(hexString))) {
```
lolodomo, line 178: You can remove some parenthesis

bundles/org.openhab.binding.knx/src/main/java/org/openhab/binding/knx/internal/client/DeviceInspector.java
```diff
@@ -102,15 +104,128 @@ private Map<String, String> readDeviceProperties(IndividualAddress address) {
...
+                        String result = new String(orderInfo);
+                        result = result.trim();
+                        if (result.isEmpty()) {
+                            result = "0x" + hexString;
+                        } else {
+                            final String printable = result.replaceAll("[^\\x20-\\x7E]", ".");
+                            if (!printable.equals(result)) {
+                                result = printable + " (0x" + toHex(orderInfo, "") + ")";
```
lolodomo, line 186: You can use `hexString` instead of `toHex(orderInfo, "")`

bundles/org.openhab.binding.knx/src/main/java/org/openhab/binding/knx/internal/client/DeviceInspector.java
```diff
@@ -102,15 +104,128 @@ private Map<String, String> readDeviceProperties(IndividualAddress address) {
...
+                    if ((count != null) && (toUnsigned(count) == 30)) {
+                        StringBuffer buf = new StringBuffer(30);
+                        for (int i = 1; i <= 30; i++) {
+                            Thread.sleep(OPERATION_INTERVAL);
+                            // for some reason, reading more than one character per message fails
+                            // reading only one character is inefficient, but works
+                            byte[] data = getClient().readDeviceProperties(address, ROUTER_OBJECT, PID.FRIENDLY_NAME, i,
+                                    1, false, OPERATION_TIMEOUT);
+                            if (toUnsigned(data) != 0) {
+                                if (data != null) {
```
lolodomo, line 208: Looking at `toUnsigned`, it will return 0 if the parameter is null. So this test is useless as already covered by the previous test.

bundles/org.openhab.binding.knx/src/main/java/org/openhab/binding/knx/internal/client/DeviceInspector.java
```diff
@@ -140,16 +258,28 @@ private Map<String, String> readDeviceDescription(IndividualAddress address) {
...
+            try {
+                final DD0 dd = DeviceDescriptor.DD0.from(data);
+
+                ret.put(DEVICE_MASK_VERSION, String.format("%04X", dd.maskVersion()));
+                ret.put(DEVICE_PROFILE, dd.deviceProfile());
+                ret.put(DEVICE_MEDIUM_TYPE, getMediumType(dd.mediumType()));
+                logger.debug("The device with address {} has mask {} ({}, medium {})", address,
+                        ret.get(DEVICE_MASK_VERSION), ret.get(DEVICE_PROFILE), ret.get(DEVICE_MEDIUM_TYPE));
+            } catch (KNXIllegalArgumentException e) {
+                logger.info("Can not parse Device Descriptor 0 of device with address {}: {}", address, e.getMessage());
```
lolodomo, line 270: Log level looks wrong. Either this is really important and it should be at WARN level, or not and in this case it should be simply at DEBUG level.

bundles/org.openhab.binding.knx/src/main/java/org/openhab/binding/knx/internal/client/DeviceInspector.java
```diff
@@ -140,16 +258,28 @@ private Map<String, String> readDeviceDescription(IndividualAddress address) {
...
-            logger.debug("The KNX device with address {} does not expose a Device Descriptor", address);
+            logger.debug("The device with address {} does not expose a Device Descriptor type 0", address);
+        }
+        data = getClient().readDeviceDescription(address, 2, false, OPERATION_TIMEOUT);
+        if (data != null) {
+            try {
+                final DD2 dd = DeviceDescriptor.DD2.from(data);
+                logger.debug("The device with address {} is has DD2 {}", address, dd.toString());
+            } catch (KNXIllegalArgumentException e) {
+                logger.info("Can not parse device descriptor 2 of device with address {}: {}", address, e.getMessage());
```
lolodomo, line 281: Sale here with log level

bundles/org.openhab.binding.knx/src/main/java/org/openhab/binding/knx/internal/client/DeviceInspector.java
```diff
@@ -140,16 +258,28 @@ private Map<String, String> readDeviceDescription(IndividualAddress address) {
...
+                logger.debug("The device with address {} has mask {} ({}, medium {})", address,
+                        ret.get(DEVICE_MASK_VERSION), ret.get(DEVICE_PROFILE), ret.get(DEVICE_MEDIUM_TYPE));
+            } catch (KNXIllegalArgumentException e) {
+                logger.info("Can not parse Device Descriptor 0 of device with address {}: {}", address, e.getMessage());
+            }
         } else {
-            logger.debug("The KNX device with address {} does not expose a Device Descriptor", address);
+            logger.debug("The device with address {} does not expose a Device Descriptor type 0", address);
+        }
+        data = getClient().readDeviceDescription(address, 2, false, OPERATION_TIMEOUT);
```
lolodomo, line 275: No need for a sleep before a new read ?

bundles/org.openhab.binding.knx/src/main/java/org/openhab/binding/knx/internal/client/DeviceInspector.java
```diff
@@ -140,16 +258,28 @@ private Map<String, String> readDeviceDescription(IndividualAddress address) {
...
+        if (data != null) {
+            try {
+                final DD2 dd = DeviceDescriptor.DD2.from(data);
+                logger.debug("The device with address {} is has DD2 {}", address, dd.toString());
```
lolodomo, line 279: All this block is only for logging a data, no property is set ?

lolodomo (changes requested):

bundles/org.openhab.binding.knx/src/main/java/org/openhab/binding/knx/internal/client/DeviceInspector.java
holgerfriedrich: right, but compiler does not get it, resulting in a compile error without the check,,,,

bundles/org.openhab.binding.knx/src/main/java/org/openhab/binding/knx/internal/client/DeviceInspector.java
holgerfriedrich: Actually I have not seen any device exposing a DD2 descriptor yet. The KNX spec says that each device needs to implement either DD0 or DD2. Maybe this log entry helps to diagnose once we get a complaint about devices which do not show data.

bundles/org.openhab.binding.knx/src/main/java/org/openhab/binding/knx/internal/client/DeviceInspector.java
holgerfriedrich: I don't mind about what exception is thrown. These are optional properties, any failure should result in skipping it and continue reading the next one.
But thanks, I added catching and re-throwing ```InterruptedException``` where applicable.

bundles/org.openhab.binding.knx/src/main/java/org/openhab/binding/knx/internal/client/DeviceInspector.java
lolodomo: So I would suggest to add all this block inside a if block based on condition logger.isDebugEnabled().
Like that, it will not be run for nothing when DEBUG logs is not enabled.

bundles/org.openhab.binding.knx/src/main/java/org/openhab/binding/knx/internal/client/DeviceInspector.java
lolodomo: I do not understand all your logic with `InterruptedException`, can you please explain it ?
Normally, this kind of exception should not be caught, or if caught you should at least "retag" the thread as interrupted.
Is it a documented exception for `getClient().readDeviceProperties` ?
And I see that sometimes you consider it (by catching and rethrowing it), sometimes you ignore it (by catching any `Exception`). What is your general logic ?

holgerfriedrich: @lolodomo 
The concept should be as follows:
Thing creation (and status change of bridge to online) triggers several scheduled tasks, including periodic datapoint reads, reading the device info, etc. This scheduled tasks are cancelled once the thing is disconnected, thus ```InterruptedException``` is thrown. All tasks should end immediately (of course this will only happen at certain points of the code, e.g. sleep).

Reading the device properties is some kind of optional task, and singe properties might not be available. This is why exceptions are ignored and the binding tries to continue with the next property. Anyhow, interruption should stop the whole task.

Catching ```InterruptedException``` was actually targeting ```Thread.sleep(OPERATION_INTERVAL);```. ```readDeviceProperties``` does not throw ```InterruptedException```, it silently fails by catching it and returning null.
See DeviceInfoClientImpl.java:70.

Actually this should be corrected. Give me some time.

holgerfriedrich: @lolodomo does it look more reasonable now? I did the change described above and added javadoc.

holgerfriedrich:
@lolodomo may I ask you to check this PR again? I think I have addressed all review comments and from my perspective it is ready to be merged.

holgerfriedrich:
@kaikreuzer may I ask you to check this PR? KNX binding still has several issues but I don't want to open too many PRs in parallel. Thanks.

lolodomo (approved):
LGTM

PR #14015, opened 2022-12-19
[knx] Upgrade Calimero library to release 2.5.1

holgerfriedrich:
Calimero 2.5.1 was recently released and brings a few fixes needed for further development of KNX data secure feature.

PR #14051, opened 2022-12-23
[knx] Replace gnu.io dependency with serial transport

holgerfriedrich:
~This is still work in progress.~
The goal is to remove dependency on  gnu.io.rxtx and to user openHAB SerialTransport.

Two steps are necessary:
* remove our code in the KNX binding (easy, only used to show available com ports)
* remove the transitive dependency imported with the Calimero library.

Calimero introduced a new way to handle pluggable serial transport classes via the java.util.ServiceLoader in release 2.5.1.
This is in principle working and I was able to connect to our serial communication classes,
~*but* Serviceloader fails once we reload the KNX plugin. I could not get this fixed, so right now openHAB needs a restart to reload the KNX plugin! Any suggestions welcome....~

Calimero seems to change the interface again, after 2.5.1 ~this was removed. I need to have a look how to proceed.~

### DIFF
--- a/bundles/org.openhab.binding.knx/pom.xml
+++ b/bundles/org.openhab.binding.knx/pom.xml
@@ -43,18 +43,6 @@
         </exclusion>
       </exclusions>
     </dependency>
-    <dependency>
-      <groupId>com.github.calimero</groupId>
-      <artifactId>calimero-rxtx</artifactId>
-      <version>2.5.1</version>
-      <scope>compile</scope>
-      <exclusions>
-        <exclusion>
-          <groupId>org.slf4j</groupId>
-          <artifactId>slf4j-api</artifactId>
-        </exclusion>
-      </exclusions>
-    </dependency>
   </dependencies>
 
   <build>
--- a/bundles/org.openhab.binding.knx/pom.xml
+++ b/bundles/org.openhab.binding.knx/pom.xml
@@ -20,6 +20,12 @@
 
   <dependencies>
     <dependency>
+      <groupId>biz.aQute.bnd</groupId>
+      <artifactId>biz.aQute.bnd.annotation</artifactId>
+      <version>${bnd.version}</version>
+      <scope>provided</scope>
+    </dependency>
+    <dependency>
       <groupId>com.github.calimero</groupId>
       <artifactId>calimero-core</artifactId>
       <version>2.5.1</version>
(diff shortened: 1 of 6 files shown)


wborn:
> Serviceloader fails once we reload the KNX plugin. I could not get this fixed, so right now openHAB needs a restart to reload the KNX plugin!

IIRC there is a similar issue with [Mary TTS](https://www.openhab.org/addons/voice/marytts/) which also uses a ServiceLoader. It was only an issue during development when updating the bundle but not for normal users whenever they (un)install the add-on.

wborn:
There are some known issues with the ServiceLoader mechanism in OSGi in which [SPI Fly](https://aries.apache.org/documentation/modules/spi-fly.html) can help.
There are also plans to add better support for it in OSGi R9, see https://github.com/osgi/osgi/issues/372.

holgerfriedrich:
Thanks @wborn for giving me some pointers.
I have seen SPI Fly, but for sure am am lacking the deeper knowledge of OSGI.....

Let's see if I can get this implemented. Otherwise, I will try to get this PR to a mergeable state and open a separate ticket for the issue during binding reload. We already have this issue in the code base, as Calimero 2.5.1 is already in use #14015.

holgerfriedrich:
@wborn it seems that adding ```bnd.bnd``` as in https://github.com/openhab/openhab-addons/pull/14051/commits/7da7dc04c4f3fe83d14a8cdb0353de4c33b51330 does the job....
I have seen this being used in jsscripting as well.
Not sure if this is the correct approach, bit it seems to work.

bundles/org.openhab.binding.knx/bnd.bnd
```diff
@@ -0,0 +1,15 @@
+Bundle-SymbolicName: ${project.artifactId}
+DynamicImport-Package: *
```
wborn, line 2: I think this will do more harm than good. 

```suggestion
```

bundles/org.openhab.binding.knx/bnd.bnd
```diff
@@ -0,0 +1,15 @@
...
+      filter:="(osgi.extender=osgi.serviceloader.processor)",
+    osgi.serviceloader:=
+      filter:="(osgi.serviceloader=tuwien.auto.calimero.serial.spi.SerialCom)";
+      cardinality:=multiple
+
+SPI-Provider: *
+SPI-Consumer: *
+
+-fixupmessages "Classes found in the wrong directory"; restrict:=error; is:=warning
+-fixupmessages "Unused Export-Package instructions"
```
wborn, line 15: Are these fixups really used?

wborn (changes requested):

holgerfriedrich:
@wborn I have now changed the serviceloader registration as implemented for marytts in #14064.
In addition knx binding uses the ```@ServiceProvider``` annotation to create the required metadata file automatically.

This changes the global ```pom.xml```, maybe it would be cleaner to merge #14064 before this one as we have the full discussion about serviceloader registrtion over there.

wborn:
I'll have a look if it is possible to reduce the amount of bnd instruction duplication.

wborn (approved):
Thanks!

holgerfriedrich:
@wborn Don't mind. I am happy to see this finally merged, as it was blocking my next PR for this binding.
I appreciate all the review work also here in the addons repo, especially as there is a lot of work done in the core as well. Thanks again.

PR #14055, opened 2022-12-24
[homekit] Add support for TV accessory

ccutrer:
Now possible since we support multiple secondary services. Just need to explicitly declare that InputSource is a linked service to a Television, not just a secondary service.

Note also that since TV and related services have so many mandatary characteristics that are often static, I introduced a new way to declare characteristics - via metadata on the service's item. Honestly, I feel like it's a lot cleaner to have a factory create the mandatory characteristics the same way as the optional characteristics, and then construct the service ourselves instead of basing the service on the specific accessory interface. But this commit is already big enough, I didn't want to go refactoring _all_ of the accessories to do it that way just yet. This is why I have "unused" metadata characteristic factory methods for AirQuality, HeaterCooler, and Thermostat - I started to make those configurable via metadata, then realized they were mandatory characteristics that couldn't be found from metadata via the current infrastructure.


timcharper (approved):

bundles/org.openhab.io.homekit/src/main/java/org/openhab/io/homekit/internal/accessories/HomekitCharacteristicFactory.java
```diff
@@ -894,8 +1019,8 @@ private static ResetFilterIndicationCharacteristic createFilterResetCharacterist
     private static ActiveCharacteristic createActiveCharacteristic(HomekitTaggedItem taggedItem,
             HomekitAccessoryUpdater updater) {
         return new ActiveCharacteristic(
-                () -> getEnumFromItem(taggedItem, ActiveEnum.ACTIVE, ActiveEnum.INACTIVE, ActiveEnum.INACTIVE),
-                (value) -> setValueFromEnum(taggedItem, value, ActiveEnum.ACTIVE, ActiveEnum.INACTIVE),
+                () -> getEnumFromItem(taggedItem, ActiveEnum.INACTIVE, ActiveEnum.ACTIVE, ActiveEnum.INACTIVE),
```
yfre, line 1022: i think "Active" was not used until now in the accessories, so, it is fine to swap default values

yfre (commented):
good idea. it was indeed very cumbersome to create mandatory characteristics that are actually static / one time configurable. i have not seen the case for air quality but for valve and thermostat. 
but i was afraid of the major redesign and did some awkward workaround. the Java-HAP accessories interfaces design was not helpful either.

thank you for doing it.
 
good approach with metadata factory. but why you dont use it in the TelevisionImpl? i assume, more PRs will follow.

in general, code looks good for me, i was just not sure about the hardcoded mapping of "true"

bundles/org.openhab.io.homekit/src/main/java/org/openhab/io/homekit/internal/accessories/HomekitMetadataCharacteristicFactory.java
```diff
@@ -0,0 +1,295 @@
...
+    }
+
+    private static <T extends Enum<T>> Supplier<CompletableFuture<T>> getEnum(Object value, Class<T> klazz) {
+        T enumValue = Enum.valueOf(klazz, value.toString());
+        return () -> CompletableFuture.completedFuture(enumValue);
+    }
+
+    private static <T extends Enum<T>> Supplier<CompletableFuture<T>> getEnum(Object value, Class<T> klazz, T trueValue,
+            T falseValue) {
+        if (value.equals(true) || value.equals("true")) {
```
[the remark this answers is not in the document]
ccutrer: Because depending on where the metadata came from (.items file, UI, JSR223 script setting it) the Object might be a String, or a Boolean. I'm not sure I follow what you mean by `trueValue would be "false"` - `trueValue` will be a value of a CharacteristicEnum (something like `ActiveEnum.ACTIVE`). There's no need for inversion when setting things like this from metadata, so `value` would never be `false` when you're trying to get `ActiveEnum.ACTIVE`. Also, `false.equals(true) || false.equals("true")` will return false, as will `"false".equals(true) || "false".equals("true")`.

ccutrer:
> but why you dont use it in the TelevisionImpl?

I think because all of the characteristics on Television are optional, and it didn't make sense to me to be able to configure any of them via metadata, because the effect would be the same as if the characteristic wasn't present at all. They're only useful if they're writable.

bundles/org.openhab.io.homekit/src/main/java/org/openhab/io/homekit/internal/accessories/HomekitMetadataCharacteristicFactory.java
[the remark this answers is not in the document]
yfre: right. forgot that this is static and does not require mapping. 
thanks for the explanation.

yfre:
> > but why you dont use it in the TelevisionImpl?
> 
> I think because all of the characteristics on Television are optional, and it didn't make sense to me to be able to configure any of them via metadata, because the effect would be the same as if the characteristic wasn't present at all. They're only useful if they're writable.

ok. need to think about the television accessory more, maybe try it out to understand how it works in Home app. 
the PR looks good for me.
thank you for adding this probably most complex accessory

yfre (approved):
LGTM

ccutrer:
@lolodomo and/or @jlaur : this is good to go from our end, just need a maintainer review. thanks!

ccutrer:
force push is a rebase to resolve a trivial merge conflict in the last line of README.md

ccutrer:
@lolodomo  or @jlaur: 🙏 can I get a review on this? I'm getting quite a stack of additional feature and bug fix commits that depend on this that I haven't sent as separate PRs yet.

bundles/org.openhab.io.homekit/src/main/java/org/openhab/io/homekit/internal/HomekitTaggedItem.java
```diff
@@ -221,6 +222,24 @@ public void send(OnOffType command) {
                 getName());
     }
 
+    /**
+     * Send IncreaseDecreaseType command to a DimmerItem (or a Group:Dimmer)
+     *
+     * @param command
```
jlaur, line 228: I would suggest to either omit this or add description for the parameter.

bundles/org.openhab.io.homekit/src/main/java/org/openhab/io/homekit/internal/accessories/HomekitAccessoryFactory.java
```diff
@@ -387,6 +395,28 @@ private static void addOptionalCharacteristics(HomekitTaggedItem taggedItem, Abs
...
+     * @param taggedItem main item
+     * @param accessory accessory
+     */
+    private static void addOptionalMetadataCharacteristics(HomekitTaggedItem taggedItem,
+            AbstractHomekitAccessoryImpl accessory)
+            throws NoSuchMethodException, IllegalAccessException, InvocationTargetException, HomekitException {
+        // Check every metadata key looking for a characteristics we can create
+        var config = taggedItem.getConfiguration();
+        if (config == null)
+            return;
```
jlaur, line 410:
```suggestion
        if (config == null)
        {
            return;
        }
```

bundles/org.openhab.io.homekit/src/main/java/org/openhab/io/homekit/internal/accessories/HomekitCharacteristicFactory.java
```diff
@@ -176,29 +210,120 @@ public class HomekitCharacteristicFactory {
...
     }
 
+    public static <T extends Enum<T>> Map<T, String> createMapping(HomekitTaggedItem item, Class<T> klazz) {
+        EnumMap<T, String> map = new EnumMap(klazz);
+        for (var k : klazz.getEnumConstants()) {
+            map.put(k, k.toString());
+        }
+        var configuration = item.getConfiguration();
+        if (configuration != null)
+            updateMapping(configuration, map);
```
jlaur, line 270:
```suggestion
        if (configuration != null)
        {
            updateMapping(configuration, map);
        }
```

bundles/org.openhab.io.homekit/src/main/java/org/openhab/io/homekit/internal/accessories/HomekitCharacteristicFactory.java
```diff
@@ -176,29 +210,120 @@ public class HomekitCharacteristicFactory {
...
+        logger.trace("Create characteristic {}", item);
+        if (optional.containsKey(type)) {
+            return optional.get(type).apply(item, updater);
+        }
+        return null;
+    }
+
     /**
-     * create optional HomeKit characteristic
+     * create HomeKit characteristic
```
jlaur, line 244:
```suggestion
     * Create HomeKit characteristic
```

bundles/org.openhab.io.homekit/src/main/java/org/openhab/io/homekit/internal/accessories/HomekitCharacteristicFactory.java
```diff
@@ -176,29 +210,120 @@ public class HomekitCharacteristicFactory {
...
+            map.put(k, k.toString());
+        }
+        var configuration = item.getConfiguration();
+        if (configuration != null)
+            updateMapping(configuration, map);
+        return map;
+    }
+
+    /**
+     * update mapping with values from item configuration.
```
jlaur, line 275:
```suggestion
     * Update mapping with values from item configuration.
```

bundles/org.openhab.io.homekit/src/main/java/org/openhab/io/homekit/internal/accessories/HomekitCharacteristicFactory.java
```diff
@@ -176,29 +210,120 @@ public class HomekitCharacteristicFactory {
...
+     * it checks for all keys from the mapping whether there is configuration at item with the same key and if yes,
```
jlaur, line 276:
```suggestion
     * It checks for all keys from the mapping whether there is configuration at item with the same key and if yes,
```

bundles/org.openhab.io.homekit/src/main/java/org/openhab/io/homekit/internal/accessories/HomekitCharacteristicFactory.java
```diff
@@ -176,29 +210,120 @@ public class HomekitCharacteristicFactory {
...
+            }
+        });
+    }
+
+    public static <T> void updateMapping(Map<String, Object> configuration, Map<T, String> map) {
+        updateMapping(configuration, map, null);
+    }
+
+    /**
+     * takes item state as value and retrieves the key for that value from mapping.
```
jlaur, line 301:
```suggestion
     * Takes item state as value and retrieves the key for that value from mapping.
```

bundles/org.openhab.io.homekit/src/main/java/org/openhab/io/homekit/internal/accessories/HomekitCharacteristicFactory.java
```diff
@@ -176,29 +210,120 @@ public class HomekitCharacteristicFactory {
...
+     * e.g. used to map StringItem value to HomeKit Enum
```
jlaur, line 302:
```suggestion
     * E.g. used to map StringItem value to HomeKit Enum
```

bundles/org.openhab.io.homekit/src/main/java/org/openhab/io/homekit/internal/accessories/HomekitMetadataCharacteristicFactory.java
```diff
@@ -0,0 +1,295 @@
...
+import io.github.hapjava.characteristics.impl.television.SleepDiscoveryModeEnum;
+import io.github.hapjava.characteristics.impl.televisionspeaker.VolumeControlTypeCharacteristic;
+import io.github.hapjava.characteristics.impl.televisionspeaker.VolumeControlTypeEnum;
+import io.github.hapjava.characteristics.impl.thermostat.CurrentHeatingCoolingStateCharacteristic;
+import io.github.hapjava.characteristics.impl.thermostat.CurrentHeatingCoolingStateEnum;
+import io.github.hapjava.characteristics.impl.thermostat.TargetHeatingCoolingStateCharacteristic;
+import io.github.hapjava.characteristics.impl.thermostat.TargetHeatingCoolingStateEnum;
+
+/**
+ * Creates a optional characteristics from metadata
```
jlaur, line 66:
```suggestion
 * Creates an optional characteristics from metadata
```

bundles/org.openhab.io.homekit/README.md
```diff
@@ -21,12 +21,12 @@ HomeKit integration supports following accessory types:
...
-- Slat  
+- Slat
 - Valve
-- Faucet / Shower  
+- Faucet / Shower
 - Speaker
-- SmartSpeaker  
-- Microphone  
+- SmartSpeaker
+- Microphone
```
jlaur, line 29: Should Television be added here?

bundles/org.openhab.io.homekit/README.md
```diff
@@ -828,8 +877,36 @@ or using UI
...
+|                      |                             | InputDeviceType              | String                        | Type of the input device. possible values (OTHER, TV, RECORDING, TUNER, PLAYBACK, AUDIO_SYSTEM). Custom mapping can be defined at item level. Can also be configured via metadata, e.g. [InputDeviceType="OTHER"]. |
+|                      |                             | TargetVisibilityState        | Switch                        | The desired visibility state of the input source. ON = shown, OFF = hidden. |
+| TelevisionSpeaker    |                             |                              |                               | An accessory that can be added to a Television in order to control the speaker associated with it. |
+|                      | Mute                        |                              | Switch                        | If the television is muted. ON = muted, OFF = not muted. |
+|                      |                             | Active                       | Switch                        | ?? |
```
jlaur, line 906: No description?

bundles/org.openhab.io.homekit/README.md
```diff
@@ -828,8 +877,36 @@ or using UI
...
+|                      |                             | Volume                       | Dimmer, Number                | Current volume. min/max/step can configured at item level, e.g. minValue=10.5, maxValue=50, step=2] |
```
jlaur, line 907:
```suggestion
|                      |                             | Volume                       | Dimmer, Number                | Current volume. Min/max/step can configured at item level, e.g. minValue=10.5, maxValue=50, step=2] |
```

jlaur (changes requested):
Thank you for these additions. I have added a few minor comments.

bundles/org.openhab.io.homekit/README.md
ccutrer: Nope. No idea what it does. TV Accessory isn't documented at all in the public documentation. A lot of this is based off of reverse engineering the HAP-JS team has done. I couldn't find that this characteristic had any effect.

bundles/org.openhab.io.homekit/README.md
ccutrer: Yes it should.

bundles/org.openhab.io.homekit/README.md
ccutrer: you mind if I leave this as-is for now? it matches the rest of the table. a subsequent pending PR already re-organizes configuration values in this table, and removes this wording

jlaur: OK.

bundles/org.openhab.io.homekit/README.md
jlaur: Hmm, it looks strange documenting it with "??". But leaving it out when it actually exists would leave it even more undocumented. Perhaps add the text "<undocumented>" or something like that rather than "??". That would at least clearly state that it's not a "TODO" kind of left-over, but rather something which is not possible to document at the moment.

ccutrer: Will do.

jlaur (approved):
Thank you for another nice feature.

jlaur:
@ccutrer - sorry, I think I was waiting for builds to complete and then forgot to get back and merge after that.

ccutrer:
No problem. I figured that's what had happened, and wasn't concerned. Thanks again!

PR #14060, opened 2022-12-24
[unifi] Fix PoE bug with combination of other data/ports

Hilbrand:
- It seems to throw an exception when updating internal cache. It can happen if you have a switch that has both PoE ports and other PoE ports or data in the port override.
- Fixed logout, should be POST instead of GET.
- Fixed typo in channel-type.config.unifi.poeEnable.mode.option.pasv24 should be without appending v.
- Removed compiler warnings.

### DIFF
--- a/bundles/org.openhab.binding.unifi/src/main/java/org/openhab/binding/unifi/internal/api/dto/UniFiSwitchPorts.java
+++ b/bundles/org.openhab.binding.unifi/src/main/java/org/openhab/binding/unifi/internal/api/dto/UniFiSwitchPorts.java
@@ -1,0 +1,122 @@
+/**
+ * Copyright (c) 2010-2022 Contributors to the openHAB project
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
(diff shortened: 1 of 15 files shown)


bundles/org.openhab.binding.unifi/src/main/java/org/openhab/binding/unifi/internal/api/dto/UniFiSwitchPorts.java
```diff
@@ -0,0 +1,122 @@
...
+    /**
+     * Return port data for the given port or if none exists set a new data object and return it.
+     *
+     * @param portIdx port to get the data for
+     * @return Return port data for the given port or if none exists set a new data object and return it.
+     */
+    public UniFiPortTuple computeIfAbsent(final int portIdx) {
+        final UniFiPortTuple tuple = ports.computeIfAbsent(portIdx, t -> new UniFiPortTuple());
+        if (tuple == null) {
+            // This should never happen because ports can never contain a null value, fand computeIfAbsent should never
```
jlaur, line 67:
```suggestion
            // This should never happen because ports can never contain a null value, and computeIfAbsent should never
```

jlaur (changes requested):
Thanks for the fixes! Is this related to https://community.openhab.org/t/unifi-oh-3-4-not-working-any-more/142260? If this is a regression from 3.4, should it be cherry-picked into the 3.4 hotfix branch after being merged?

jlaur (approved):
Thanks!

jlaur:
@Hilbrand, @lolodomo - I would suggest to include this in the hotfix, even if it was not a regression from 3.4. WDYT?

PR #14061, opened 2022-12-24
Set openHAB to correct version in binding skeleton scripts.

Hilbrand:


wborn (approved):

timcharper (approved):

bundles/create_openhab_binding_skeleton.sh
```diff
@@ -2,7 +2,7 @@
 
 [ $# -lt 3 ] && { echo "Usage: $0 <BindingIdInCamelCase> <Author> <GitHub Username>"; exit 1; }
 
-openHABVersion=3.4.0-SNAPSHOT
+openHABVersion=4.0.0-SNAPSHOT
```
splatch, line 5: Out of curiosity, could it be `cat pom.xml|grep '<version>'|cut …` or similar to read it from pom instead of hardcoding it in shell script?
I don't know how it would look for windows, but I could construct one liner for linux.

wborn: I also don't know much about Windows scripts.

You can also use Maven itself to evaluate it though it is usually a bit slower as greps:

https://maven.apache.org/plugins/maven-help-plugin/evaluate-mojo.html

splatch: Making maven call would work same for windows and linux thus it could be a winner :)

Issue #6571, opened 2019-12-11
[astro] Moon phase WANING_GIBBOUS reported before Full moon


Rossko57:
Astro binding 2.4

Things
`astro:moon:fossmoon   [ geolocation="51.783, -1.902", interval=600 ]`
Item
`String moon_phase "Moon Phase" {channel="astro:moon:fossmoon:phase#name"}`
results
`2019-12-11 12:54:37.662 [vent.ItemStateChangedEvent] - moon_phase changed from NULL to WANING_GIBBOUS`

Full moon is due "tomorrow" i.e. 12/12/19 so correct phase would be Waxing gibbous

Problem still present in 2.5M6
https://community.openhab.org/t/astro-binding-reports-wrong-moon-phase/87391

shutterfreak:
I think the main problem is that the moon phase name seems to be calculated only once (at midnight of the Julian date) whereas on a given calendar day more than one lunar phase can occur (e.g., `WAXING_GIBBOUS` &rarr; `FULL` &rarr; `WANING_GIBBOUS`).

A possible workaround is to check the instant moon age (`astro:moon:home:phase#agePercent`: value between 0 and 1) as it is definitely instantly computed by the astro binding. If the moon age increases, we're in the waxing half of the lunar cycle; if the phase decreases, we're in the waning half.

Creating a workaround rule is rather straightforward.

Rossko57:
I'm on GMT and today is the 11th.  The Full moon is on the 12th.  The whole of the 11th should be Waxing, or Full at stretch, and in no case be Waning.

shutterfreak:
I still believe the problem lies somewhere at either a time zone conversion or at a different midnight in use (astronomers have more than one definition of year or day).

I posted a code snippet with a fairly accurate approximation of the lunar phase owing to Jean Meeus. See my reply in the related openHAB community thread.

Rossko57:
Here in GMT it is still the 11th, Full moon is due in a couple of hours "just" into the 12th at 00:12 - I will report again then.
Meantime Astro is still updating age **each hour** as expected -
`2019-12-11 22:30:35.083 [vent.ItemStateChangedEvent] - moon_age changed from 51.713532126958796 % to 51.73700221403369 %`
Reckon this stuff is just a day out.

shutterfreak:
I think that the moon age might indeed be off by about 1 day.

shutterfreak:
[...]

PR #14067, opened 2022-12-26
[astro] Added moon phase precision

lsiepel:
Prior to this PR the moon age was calculated in a double and rounded to a day before it is send to tthe channel.  I removed the rounding.

It has two effects:
1. The age channel (Duration) represeted in days has double value.
2. The calculation of isWaxing is more precise 

Closes https://github.com/openhab/openhab-addons/issues/6571


bundles/org.openhab.binding.astro/src/main/java/org/openhab/binding/astro/internal/model/MoonPhase.java
```diff
@@ -105,7 +105,7 @@ public QuantityType<Time> getAge() {
     /**
      * Sets the age in days.
      */
-    public void setAge(int age) {
+    public void setAge(Double age) {
```
jlaur, line 108:
```suggestion
    public void setAge(double age) {
```

jlaur:
Just to be clear, my comment is a single comment, not a review. 😄 The problem solved here seems complex and I don't have any knowledge about the astro calculations and workings, so I hope @gerrieg will be able to review.

lolodomo (approved):
Code LGTM

lolodomo:
> The problem solved here seems complex and I don't have any knowledge about the astro calculations and workings

Same for me !
But I understand that now getAge() returns a number of days which is not rounded to a day.
@lsiepel : did you check all calls of this method ?

lsiepel:
> 
Because all the complex calculations it might look complex. But in essence it is really simple. The age was first rounded then reported and used for further calculations. More precission was added by removing the rounding.

Only side effect i can think of is that users might have rules / item labels etc. that don't expect a double for MoonAge. But that is easily fixed (if needed at all) by transforming the value to a full int. Atleast the user now has the choice to do so.

lolodomo:
Ok, so let's merge it.

jlaur:
> Because all the complex calculations it might look complex. But in essence it is really simple. The age was first rounded then reported and used for further calculations. More precission was added by removing the rounding.

Would you consider adding a unit test for that? :-) Only sun calculations are currently covered, but perhaps this would be an opportunity to add moon coverage.

clinique:
Thinking of it a bit late, this can break some sitemaps using the int value to provide dynamic icon.

PR #14068, opened 2022-12-26
[rrd4j] Improve logging of exceptions thrown by getDB

holgerfriedrich:
I have seen a discussion about warning messages from rrd4j in the 3.4 release discussion thread.
https://community.openhab.org/t/openhab-3-4-release-discussion/142257/130

The current log may look like this:
> Failed to open rrd4j database ‘...’ (java.lang.IllegalStateException)

This not really helpful for identifying the root cause, as rrd4j library uses ```llegalStateException``` at many different places, but with specific error messages.

So we should make use of the method ```getMessage()```. ```toString()``` will actually print class name and message (if available).
This will it make easier to understand issues being reported.
(should IMHO also be cherry picked for potential 3.4.x patch release)

### DIFF
--- a/bundles/org.openhab.persistence.rrd4j/src/main/java/org/openhab/persistence/rrd4j/internal/RRD4jPersistenceService.java
+++ b/bundles/org.openhab.persistence.rrd4j/src/main/java/org/openhab/persistence/rrd4j/internal/RRD4jPersistenceService.java
@@ -156,7 +156,7 @@ public class RRD4jPersistenceService implements QueryablePersistenceService {
         try {
             db = getDB(name);
         } catch (Exception e) {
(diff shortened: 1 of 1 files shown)


lolodomo (approved):
LGTM, thank you

lolodomo:
> (should IMHO also be cherry picked for potential 3.4.x patch release)

@jlaur, can I let you do that ?

jlaur:
> @jlaur, can I let you do that ?

Done! For reference:
```shell
git checkout 3.4.x
git pull
git cherry-pick e7938ae357687d37d22c1e3820842f3d5f010793
git push
```

(for openhab-addons repository, not your own fork)

PR #14070, opened 2022-12-27
[tesla] Fix discovery of Tesla vehicles from account

kaikreuzer:
So far, vehicles that are asleep (which is the normal situation) are not discovered and thus not put into the Inbox, what is fairly unexpected for the user.
This change makes sure that every vehicle of the account is added to the Inbox. It defers the identification of the model to a later stage as this is only required for creating the Thing with all its channels.


bundles/org.openhab.binding.tesla/src/main/java/org/openhab/binding/tesla/internal/discovery/TeslaVehicleDiscoveryService.java
```diff
@@ -81,31 +81,15 @@ public void deactivate() {
 
     @Override
     public void vehicleFound(Vehicle vehicle, VehicleConfig vehicleConfig) {
```
lolodomo, line 83: Please add `@Nullable` to the parameter `vehicleConfig`.

bundles/org.openhab.binding.tesla/src/main/resources/OH-INF/i18n/tesla.properties
```diff
@@ -206,7 +206,7 @@ channel-type.tesla.minavailabletemp.label = Minimum Temperature
 channel-type.tesla.minavailabletemp.description = Indicates the minimal inside temperature of the vehicle
 channel-type.tesla.mobileenabled.label = Mobile Enabled
 channel-type.tesla.mobileenabled.description = Indicates whether the vehicle can be remotely controlled
-channel-type.tesla.notenoughpower.label = Not Enought Power
+channel-type.tesla.notenoughpower.label = Not Enough Power
```
lolodomo, line 209: Please also update the typo in the XML file.

And as you added a new thing type, you need to add all the new information for this thing type in this file (the easiest way is to run the i18n tool).

lolodomo (changes requested):
I was not aware of the existence of the ThingTypeMigrationService service.

bundles/org.openhab.binding.tesla/src/main/java/org/openhab/binding/tesla/internal/discovery/TeslaVehicleDiscoveryService.java
kaikreuzer: The whole class/binding does not yet have null annotations on it, I wouldn't want to do it as part of a bug fix.

bundles/org.openhab.binding.tesla/src/main/resources/OH-INF/i18n/tesla.properties
kaikreuzer: I always forget this. But I do not seem to be the only one - regenerating the files shows a hell of a lot new entries that are not all related to my changes. 😉

kaikreuzer:
@lolodomo Thanks for the quick review - I have addressed your comments.

lolodomo (approved):
LGTM, thank you

Issue #14069, opened 2022-12-26
[shelly] Wrong profile after restart of thing / rereading configuration on Shelly TRV


goligo:
## Expected Behavior
When OpenHAB is restarted, or a thing is restarted, its configuration should not be changed and should match the actual configuration on the device.

## Current Behavior
When the current selectedProfile is 0 (schedule disabled) and the thing is restarted, the selectedProfile is changed to 1. However only in OpenHAB, the selected profile on the device itself stays 0 (disabled). 

The reason seems to be the way, how the profile state is persisted in the shelly settings:
"schedule":false,"schedule_profile":1

So when the profile is set to 0, the shelly TRV does not set "schedule_profile" to 0, but instead "schedule" to false.

## Possible Solution
I guess the binding needs to take care of this behaviour, and do it the other way round - if "schedule" equals false, then selectedProfile should be set to 0, regardless of the value of "schedule_profile"

## Steps to Reproduce (for Bugs)
1. Have an item linked to the selectedProfile channel of a Shelly TRV
2. Change selectedProfile to 0 (DIsabled)
3. Disable and enable thing to reinitialize configuration
4. See that selectedProfile has wrongly been set to 1 

## Context
UI does show wrong data, which does not match the actual setting on the device.

## Your Environment
OpenHAB 3.4 on a Raspberry 3 with OpenHabian

goligo:
Hmmm... the code looks like it does exactly what I was proposing as a solution. So why does it still go wrong?

https://github.com/openhab/openhab-addons/blob/3.4.x/bundles/org.openhab.binding.shelly/src/main/java/org/openhab/binding/shelly/internal/handler/ShellyComponents.java#L441-L445

goligo:
Ah, I think I found the root cause, in the latest changes to profile handling

https://github.com/openhab/openhab-addons/commit/5b038786d586e47880ffca4343bdd35d9c94101f#diff-719dc6767a12baa72226bcfcb8539394229258660c3b55a6a84b7bc9d2ec09a7

Before it was
```
            if (id <= 0) {
                return "DISABLED";
            }
```
now it is
```
        if (id <= 0 && settings.thermostats != null) {
            id = settings.thermostats.get(0).profile;
        }
```

Which is wrong, if id is 0, it is disabled, it should NOT return the last set profile id, which can be found in the settings object, but just 0.

@markus7017 I guess. you want to have a look at it?

goligo:
[...]

PR #14075, opened 2022-12-27
[shelly] TRV: Add selectedProfile state option "Disabled" and made it selectable

goligo:
The selectedProfile did not have 0:Disabled in its state options and even if adding it manually, it could not be selected. Now the behaviour of the selectedProfile channel matches the profile selection on the device itself, as well as the documentation of the shelly binding.

Fixes https://github.com/openhab/openhab-addons/issues/14069

### DIFF
--- a/bundles/org.openhab.binding.shelly/src/main/java/org/openhab/binding/shelly/internal/api/ShellyDeviceProfile.java
+++ b/bundles/org.openhab.binding.shelly/src/main/java/org/openhab/binding/shelly/internal/api/ShellyDeviceProfile.java
@@ -351,8 +351,8 @@ public class ShellyDeviceProfile {
 
     public String getValueProfile(int valveId, int profileId) {
         int id = profileId;
-        if (id <= 0 && settings.thermostats != null) {
-            id = settings.thermostats.get(0).profile;
+        if (id <= 0 || id > 5) {
+            id = 0;
         }
         return "" + id;
     }
--- a/bundles/org.openhab.binding.shelly/src/main/java/org/openhab/binding/shelly/internal/handler/ShellyBaseHandler.java
+++ b/bundles/org.openhab.binding.shelly/src/main/java/org/openhab/binding/shelly/internal/handler/ShellyBaseHandler.java
@@ -602,6 +602,8 @@ public abstract class ShellyBaseHandler extends BaseThingHandler
             String channelId = mkChannelId(CHANNEL_GROUP_CONTROL, CHANNEL_CONTROL_PROFILE);
             logger.debug("{}: Adding TRV profile names to channel description: {}", thingName, profileNames);
             channelDefinitions.clearStateOptions(channelId);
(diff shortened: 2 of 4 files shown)


goligo:
After having a second look, I think the method getValueProfile can be removed completely. I kept it because it was checking for <= 0, so I thought it might be called with invalid input data, but the only two classes using this method are Shelly1CoiotVersion2 and ShellyComponents, both of which get their value directly from the Shelly TRV, so the value will always be valid.

lolodomo:
@markus7017: can you please review ?

goligo:
@markus7017 Hey, any chance to have look at this?

markus7017:
I don't agree, the problem is that the device not ready has profile 0, but schedule 0 / 1
When the thing is initialized the binding reads the current value from the device. The device doesn't know value 0, this was just a supplication for the implementation  I expect schedule = 0 when profile is disabled. Please check, in this case we have
- schedule = ON,  profile = selected profile
- schedule = OFF, profile = profile value from the device, but not active
please verify

join us here:https://community.openhab.org/t/shelly-binding/56862/3128

goligo:
You don't agree to what? The setting in the device UI, as well as the REST API of the device works like that, you can set selectedProfile to 0 to disable the automatic schedule. This is also worked in the previous release of the Shelly binding, but was broken with your changes as described in https://github.com/openhab/openhab-addons/issues/14069#issuecomment-1365513543 .

This PR just fixes the behaviour to work as expected, as it did before and as documented.

goligo:
Just had a second look - the MQTT and REST API have schedule and schedule_profile as separate parameters and schedule_profile does not accept 0 as a value. However the device UI does show "Disabled" as it was a profile, and from a user point of view it also makes more sense to show "Disabled" as the current schedule, instead of the name of a schedule which is currently not active.

Shelly_Disabled

Also from a compatibility point of view, for users who are already using the Shelly Binding it does make sense to keep it that way. If you think about the UI - you need an additional widget to be able to disable the schedule, instead of having the possibility to disable on the same select widget. I think there is a good reason the Shelly UI designers also implemented it this way.

goligo:
There is another interesting aspect regarding schedule and selectedProfile - if you see these two separated and independent, changing the selectedProfile should not activate the schedule. However this coupling is exactly what you are doing here

https://github.com/openhab/openhab-addons/blob/5b038786d586e47880ffca4343bdd35d9c94101f/bundles/org.openhab.binding.shelly/src/main/java/org/openhab/binding/shelly/internal/api1/Shelly1CoIoTVersion2.java#L108-L113

I think this is right - there should be state option "disabled" and they should be coupled this way. When you set schedule to false, selectedProfile should be switched to "disabled". When you chose a selectedProfile different than "disabled", then schedule should set to true.

goligo:
I don't see any interest in discussing or fixing this issue by the maintainer, so I am closing the issue.

PR #13970, opened 2022-12-16
[renault] Refactor/improve logging and error handleing. New channels …

dougculnane:
As a result of discussion in the forum starting here:
https://community.openhab.org/t/betatest-renault-ze-services-binding/72226/212

The following has been improved in the Renault binding.

- refactor the error logging for clarity and consistency.
- disable features that do not work on certain cars and reduce repeated logging of errors.
- add 3 new channels.
- improve documentation of channels.
- update sitemap

## New Channels

| Channel ID             | Type               | Description                                     | Read Only |
|------------------------|--------------------|-------------------------------------------------|-----------|
| batterystatusupdated   | DateTime           | Timestamp of the last battery status update     | Yes       |
| lockstatus             | String             | Lock status of the car                          | Yes       |
| updatenow              | Switch             | Switch to force an immediate update of channels | No        |

I and user NikM have tested this and it seams stable.  The lock status does not work on our cars but it should work if this documentation is correct. https://renault-api.readthedocs.io/en/latest/endpoints.html#lock-status

PR #14076, opened 2022-12-27
[renault] Add new channels 'batterystatusupdated' and 'locked'

dougculnane:
I resubmit this Pull Request after the commit mess here:
https://github.com/openhab/openhab-addons/pull/13970

As a result of discussion in the forum starting here:
https://community.openhab.org/t/betatest-renault-ze-services-binding/72226/212

The following has been improved in the Renault binding.

    refactor the error logging for clarity and consistency.
    disable features that do not work on certain cars and reduce repeated logging of errors.
    add 2 new channels.
    improve documentation of channels.
    update sitemap
    incorporate feedback from first review.


jlaur:
@dougculnane - will you respond to my comments in the original PR or copy them here?

bundles/org.openhab.binding.renault/README.md
```diff
@@ -19,22 +19,23 @@ No discovery
...
+|-------------------|----------------------------------------------------------------------------|----------------------------------|
+| myRenaultUsername | MyRenault Username.                                                        |                                  |
+| myRenaultPassword | MyRenault Password.                                                        |                                  |
+| locale            | MyRenault Location (language_country).                                     |                                  |
+| vin               | Vehicle Identification Number.                                             |                                  |
+| refreshInterval   | Interval the car is polled in minutes.                                     |                               10 |
+| updateDelay       | How long to wait for commands to reach car and update to server in seconds.|                               30 |
+| kamereonApiKey    | Kamereon API Key.                                                          | VAX7XYKGfa92yMvXculCkEFyfZbuM7Ss |
```
dougculnane, line 30: I added the default values as you suggested because i think this is then clear.

bundles/org.openhab.binding.renault/README.md
```diff
@@ -69,24 +71,24 @@ renaultcar.sitemap:
 sitemap renaultcar label="Renault Car" {
```
dougculnane, line 71: I removed the binary image of the site map as requested but kept the text one.  This gives uses an easy UI but I think this could be improved with the new openhab UI 3 in future.  I kept the icons because the category definitions in the channel XML caused me runtime errors on start.

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/api/Car.java
```diff
@@ -40,19 +42,22 @@ public class Car {
...
     private @Nullable Double batteryLevel;
     private @Nullable Double batteryAvailableEnergy;
+    private @Nullable ZonedDateTime batteryStatusUpdated;
     private @Nullable Integer chargingRemainingTime;
     private @Nullable Boolean hvacstatus;
     private @Nullable Double odometer;
     private @Nullable Double estimatedRange;
     private @Nullable String imageURL;
-    private @Nullable String locationUpdated;
+    private @Nullable ZonedDateTime locationUpdated;
```
dougculnane, line 59: I changed string timestamps to ZonedDateTime as suggested and it works with the supplied format.

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/handler/RenaultHandler.java
```diff
@@ -196,7 +202,16 @@ public void handleCommand(ChannelUID channelUID, Command command) {
...
+                break;
             default:
+                if (command instanceof RefreshType) {
+                    ScheduledFuture<?> job = pollingJob;
+                    if (job != null) {
+                        job.cancel(true);
+                    }
+                    pollingJob = scheduler.scheduleWithFixedDelay(this::getStatus, 0, config.refreshInterval,
+                            TimeUnit.MINUTES);
+                }
```
dougculnane, line 214: Thanks for this suggestion.  It works (I think) and I have removed the update now switch channel.  I want my user to test this and maybe we will add some documentation if it is OK from a code perspective.  I have added null pointer handling as per the warning.

bundles/org.openhab.binding.renault/src/main/resources/OH-INF/i18n/renault.properties
```diff
@@ -55,6 +55,9 @@ thing-type.config.renault.car.vin.description = Vehicle Identification Number
 # channel types
 
 channel-type.renault.batteryavailableenergy.label = Battery Energy Available
+channel-type.renault.batterystatusupdated.label = Battery Status Updated
+channel-type.renault.batterystatusupdated.description = Timestamp of the last battery status update
+channel-type.renault.batterystatusupdated.state.pattern = %1$tH:%1$tM %1$td.%1$tm.%1$tY
```
dougculnane, line 60: I added the English and reverted the French.  I think this is how it is documented to be done.

bundles/org.openhab.binding.renault/src/main/resources/OH-INF/thing/thing-types.xml
```diff
@@ -96,6 +98,12 @@
 		<label>Battery Energy Available</label>
 		<state pattern="%.1f %unit%" readOnly="true"></state>
```
dougculnane, line 99: I reverted this back to unit as the 3.4.0-SNAPSHOT uses kWh as a default and does not convert to Joules.

dougculnane (commented):
@jlaur  I tried to incorporate you feedback as best I could.  I got a better understanding of the rebase flow now, and how to rebase and sign any accepted suggestions from the web UI.

Please check if this is OK and let me do some more testing.  Hopefully we have a good base (or rebaseable) pull request for these improvements now.

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/api/Car.java
```diff
@@ -105,6 +116,9 @@ public void setBatteryStatus(JsonObject responseJson) {
                 if (attributes.get("chargingRemainingTime") != null) {
                     chargingRemainingTime = attributes.get("chargingRemainingTime").getAsInt();
                 }
+                if (attributes.get("timestamp") != null) {
+                    batteryStatusUpdated = ZonedDateTime.parse(attributes.get("timestamp").getAsString());
```
jlaur, line 120: This can throw [DateTimeParseException](https://docs.oracle.com/javase/8/docs/api/java/time/format/DateTimeParseException.html) if the format is not as expected.

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/handler/RenaultHandler.java
```diff
@@ -196,7 +202,16 @@ public void handleCommand(ChannelUID channelUID, Command command) {
                         return;
                     }
                 }
+                break;
             default:
+                if (command instanceof RefreshType) {
```
jlaur, line 207: You might want to do this before checking the channel, since the job will probably refresh values for all channels? With the current implementation, it seems that a request to refresh one of the channels above will not work?

jlaur (changes requested):
Thanks. Only two minor comments left.

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/api/Car.java
dougculnane: I have pushed a try / catch for this.

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/handler/RenaultHandler.java
dougculnane: Good point.  Yes a REFRESH will poll the car and update all channels.  This is because the poll follows the phyton logic and goes through all services.  (If I have time to debug the API request limit counters I might be able to improve this with a later enhancement)

I removed 2 breaks for channels that REFRESH by contacting the car.  The other channel "CHANNEL_HVAC_TARGET_TEMPERATURE" has its own REFRESH logic and does not contact the car so it has a break.  I think this is better and correct.

dougculnane:
@jlaur Thanks for your help.  I will build the latest version and test it myself and put it in the forum for testing.  I will give feedback in a few days if we find any issues.

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/handler/RenaultHandler.java
jlaur: It doesn't look correct. You will now update both **CHANNEL_HVAC_STATUS** and **CHANNEL_CHARGING_MODE** with a value intended for **CHANNEL_HVAC_STATUS** when receiving a command which is not `RefreshType`. In the beginning of the method you might want to do something like this instead:

```java
if (command instanceof RefreshType) {
    switch (channelUID.getId()) {
        case RenaultBindingConstants.CHANNEL_HVAC_STATUS:
        case RenaultBindingConstants.CHANNEL_CHARGING_MODE:
            // Reschedule job for supported channels
    }
    return;
}
```

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/api/Car.java
```diff
@@ -105,6 +117,13 @@ public void setBatteryStatus(JsonObject responseJson) {
                 if (attributes.get("chargingRemainingTime") != null) {
                     chargingRemainingTime = attributes.get("chargingRemainingTime").getAsInt();
                 }
+                if (attributes.get("timestamp") != null) {
+                    try {
+                        batteryStatusUpdated = ZonedDateTime.parse(attributes.get("timestamp").getAsString());
+                    } catch (DateTimeParseException e) {
+                        logger.warn("Error updating battery status updated timestamp. {}", e.getMessage());
```
jlaur, line 124: Log level should be reduced to **debug**. Otherwise users will see this for every poll cycle and all refresh commands in case the service would start to return incompatible values.

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/api/Car.java
```diff
@@ -153,7 +172,24 @@ public void setLocation(JsonObject responseJson) {
                     gpsLongitude = attributes.get("gpsLongitude").getAsDouble();
                 }
                 if (attributes.get("lastUpdateTime") != null) {
-                    locationUpdated = attributes.get("lastUpdateTime").getAsString();
+                    try {
+                        locationUpdated = ZonedDateTime.parse(attributes.get("lastUpdateTime").getAsString());
+                    } catch (DateTimeParseException e) {
+                        logger.warn("Error updating location updated timestamp. {}", e.getMessage());
```
jlaur, line 178: Log level should be reduced to **debug**. Otherwise users will see this for every poll cycle and all refresh commands in case the service would start to return incompatible values.

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/api/Car.java
jlaur: Additionally, should `batteryStatusUpdated` be set to `null` in this case?

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/api/Car.java
jlaur: Additionally, should `locationUpdated` be set to `null` in this case?

bundles/org.openhab.binding.renault/README.md
jlaur: Which error did you get? Are you sure you put the category at the right position (order in XML has to be correct according to XSD)?

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/handler/RenaultHandler.java
dougculnane: ??? Sorry I am confused... What I think I want is that an action will be performed if the channel id and command specify it.  CHANNEL_HVAC_TARGET_TEMPERATURE and REFRESH have their own logic block and a break. All other channels trigger a refresh of all channels if the command is a REFRESH?

At the moment a poll updates all channels.

jlaur: Context: https://github.com/openhab/openhab-addons/pull/13970#discussion_r1052653841

So my understanding is that you would like the possibility to refresh on demand, for example from a rule. This is what `RefreshType` is for. However, this should not be mixed with other commands. You will either send a command to a channel in order to update its state, like:

```java
Renault_ChargingMode.SendCommand("ALWAYS_CHARGING");
```

or you will send a refresh command in order to request a state update from the thing:

```java
sendCommand(Renault_ChargingMode, "REFRESH")
```

What is a little confusing here is that updating CHANNEL_HVAC_STATUS will also reschedule the job. I'm not sure what the reason for that is, but maybe this is some kind of work-around to reflect the new state.

dougculnane: @jlaur  OK now I have a per channel handler with a break, and a default for the other channels.  This is nice and simple and I think will avoid the issue you describe and future confusion.

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/api/Car.java
dougculnane: Done

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/api/Car.java
dougculnane: Done

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/handler/RenaultHandler.java
```diff
@@ -110,7 +112,7 @@ public void initialize() {
             return;
         }
         updateStatus(ThingStatus.UNKNOWN);
-
+        updateState(CHANNEL_LOCK_STATUS, new StringType(car.getLockStatus().name()));
```
jlaur, line 115: Since `car` is not yet initialized, I'm not sure it makes sense to update the channel at this point.

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/handler/RenaultHandler.java
```diff
@@ -146,32 +149,40 @@ public void handleCommand(ChannelUID channelUID, Command command) {
...
-                        && !car.isDisableHvac()) {
-                    final MyRenaultHttpSession httpSession = new MyRenaultHttpSession(this.config, httpClient);
-                    try {
-                        updateState(CHANNEL_HVAC_STATUS, new StringType(Car.HVAC_STATUS_PENDING));
-                        car.resetHVACStatus();
-                        httpSession.initSesssion(car);
-                        httpSession.actionHvacOn(car.getHvacTargetTemperature());
-                        if (pollingJob != null) {
-                            pollingJob.cancel(true);
+                if (!car.isDisableHvac()) {
```
jlaur, line 152: Can the HVAC enabled/disabled state change after polling new data? In this case `RefreshType` should be checked unconditionally.

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/handler/RenaultHandler.java
```diff
@@ -146,32 +149,40 @@ public void handleCommand(ChannelUID channelUID, Command command) {
...
+                    if (command instanceof RefreshType) {
+                        reschedulePollingJob();
+                    } else if (command instanceof StringType && command.toString().equals(Car.HVAC_STATUS_ON)
+                            && !car.isDisableHvac()) {
```
jlaur, line 156: This will always evaluate to `true` as already checked in line 152.

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/handler/RenaultHandler.java
dougculnane: No. Features are switched off (if the service is forbidden etc..) and then is skipped until the thing is deactivated and reactivated.

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/api/Car.java
dougculnane: I set these to null now.  The channel will not get updated if the values are null.  So the last update in the Item value will be the last successful update.

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/api/Car.java
dougculnane: I set these to null now.  The channel will not get updated if the values are null.  So the last update in the Item value will be the last successful update.

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/handler/RenaultHandler.java
dougculnane: This sets the lock status to the default of "unknowen". If this lock status service works then the car will update this.  This is where I set default channel values for initialised sessions...  Maybe there is a better location?

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/handler/RenaultHandler.java
dougculnane: The CHANNEL_HVAC_STATUS reschedules the polling with a delay of a few seconds (config.updateDelay).  This gives the Renault servers time to contact the car and send the command to it.  They then update their HVAC status for the car.  Then we can query it and change our status from "PENDING" to "ON/OFF"

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/handler/RenaultHandler.java
jlaur: I think it would better to not update the channel at all at this point.

Imagine having a persisted item linked to this channel. So the first time the binding is started, item will be updated to **UNKNOWN**, then shortly after to **LOCKED** or **UNLOCKED**. After next openHAB restart, the same will happen. And you will have persisted values UNKNOWN -> LOCKED -> UNKOWN -> LOCKED.

It should only be unknown for a few milliseconds until handler is fully initialized and you have fetched the values. But actually it was also unknown prior to that, when the binding was not yet started. So what would be the advantage of updating to **UNKNOWN** for such short period of time?

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/handler/RenaultHandler.java
```diff
@@ -146,32 +149,39 @@ public void handleCommand(ChannelUID channelUID, Command command) {
...
-                    final MyRenaultHttpSession httpSession = new MyRenaultHttpSession(this.config, httpClient);
-                    try {
-                        updateState(CHANNEL_HVAC_STATUS, new StringType(Car.HVAC_STATUS_PENDING));
-                        car.resetHVACStatus();
-                        httpSession.initSesssion(car);
-                        httpSession.actionHvacOn(car.getHvacTargetTemperature());
-                        if (pollingJob != null) {
-                            pollingJob.cancel(true);
+                if (!car.isDisableHvac()) {
+                    if (command instanceof RefreshType) {
```
jlaur, line 153: Can you check lines 132-134 also? I know you didn't change them, but since you are implementing refresh, you probably want it to work for **CHANNEL_HVAC_TARGET_TEMPERATURE** also, and the implementation is wrong here.

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/handler/RenaultHandler.java
```diff
@@ -320,10 +340,38 @@ private void updateBattery(MyRenaultHttpSession httpSession) {
...
+            } catch (RenaultNotImplementedException e) {
+                logger.warn("Disable lock status update.");
+                car.setDisableLockStatus(true);
+            } catch (RenaultForbiddenException | RenaultUpdateException e) {
+                logger.warn("Error updating lock status.", e);
             }
         }
     }
+
+    private void reschedulePollingJob() {
```
jlaur, line 370: Perhaps `initialize()` should also use this new method?

bundles/org.openhab.binding.renault/README.md
dougculnane: Ah OK.  That might have been the issue.  I will try again.

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/handler/RenaultHandler.java
dougculnane: Good point. I had the problem that if the lock service got disabled the lock status was null not unknown.  I will try to handle this better.

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/handler/RenaultHandler.java
dougculnane: The CHANNEL_HVAC_TARGET_TEMPERATURE is a bit strange.  It only sets this value in the binding "car" object where is is persisted.  The value is not sent to the real car.  This value is only used on the HVAC ON preconditioning action. So there should be not polling job change when this is refreshed or set.

jlaur: Oh. Now I checked it more thoroughly and you are right. I'm also right though, implementation is wrong or a least very odd. But that's outside the scope of your PR to fix:
- If the value cannot be requested from the car, there is no reason for implementing refresh in `handleCommand`. There is nothing to refresh.
- If HVAC is already ON, setting a different target temperature should call `actionHvacOn` to update the value in the car.

I would suggest to leave this for now (in context of this PR), but it seems there is room for improvement here.

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/handler/RenaultHandler.java
```diff
@@ -320,10 +336,39 @@ private void updateBattery(MyRenaultHttpSession httpSession) {
...
         }
     }
+
+    private void updateLockStatus(MyRenaultHttpSession httpSession) {
+        if (!car.isDisableLockStatus()) {
+            try {
+                httpSession.getLockStatus(car);
+                updateState(CHANNEL_LOCK_STATUS, new StringType(car.getLockStatus().name()));
+            } catch (RenaultNotImplementedException e) {
+                updateState(CHANNEL_LOCK_STATUS, new StringType(LockStatus.UNKNOWN.name()));
```
jlaur, line 358:
```suggestion
                updateState(CHANNEL_LOCK_STATUS, UnDefType.UNDEF);
```

jlaur, line 358:
```suggestion
                updateState(CHANNEL_LOCK_STATUS, UnDefType.UNDEF);
```

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/handler/RenaultHandler.java
```diff
@@ -320,10 +336,39 @@ private void updateBattery(MyRenaultHttpSession httpSession) {
...
+                logger.warn("Disable lock status update.");
```
jlaur, line 359: Under which circumstances will this happen? Perhaps there should be a reason mentioned?

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/handler/RenaultHandler.java
```diff
@@ -320,10 +336,39 @@ private void updateBattery(MyRenaultHttpSession httpSession) {
...
+                car.setDisableLockStatus(true);
+            } catch (RenaultForbiddenException | RenaultUpdateException e) {
+                logger.warn("Error updating lock status.", e);
```
jlaur, line 362: Under which circumstances will this happen? Will it be temporary failures (e.g. network issues) or could it be permanently rejection of such requests? In this case the warnings would keep reappearing in logs?

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/handler/RenaultHandler.java
dougculnane: As a result of testing the target temperature has to be 19, 20 or 21... or HVAC ON it does not consistently work.  So it is not work having as a channel.  However maybe one day it will be more useful.  I am not sure what will happen if you try to change the target temperature once the car is doing a pre-heat.  I would need to test this.

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/handler/RenaultHandler.java
dougculnane: I will not commit your suggestion with the Web UI due to the signing issues I had to rebase .... But I will adopt it.  The  LockStatus.UNKNOWN is maybe not useful then?

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/handler/RenaultHandler.java
dougculnane: Not sure.  It does not work on my car or the testers cars.  However it is documented to work here: https://renault-api.readthedocs.io/en/latest/endpoints.html#lock-status  so it might work for some cars.  If it does not work then the service call is disabled to avoid log SPAM and repeated calls to it.  Next time the thing it initialized all services are enabled again so it might start working one day if the service become available for the car/account.

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/handler/RenaultHandler.java
dougculnane: Once the service is disabled due to a HTTP 404. 501, 502 the lock status service is disabled after the first failing call so that it does not keep reappearing in logs.

bundles/org.openhab.binding.renault/README.md
dougculnane: @jlaur I have added the categories after description and this is working.  If the Items are re-created they adopt the category Time and icon in the site map.  I have remove the redundant icon definition from the site map as suggested.

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/handler/RenaultHandler.java
dougculnane: I have pushed this change.

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/handler/RenaultHandler.java
jlaur: So perhaps "Disabling unsupported lock status update" or something like that?

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/handler/RenaultHandler.java
jlaur: I assume these three HTTP codes will cause `RenaultNotImplementedException `then? So are you saying that `RenaultForbiddenException` and `RenaultUpdateException` cannot happen unless `RenaultNotImplementedException` already was thrown prior to that, or that these exceptions will only be temporary?

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/handler/RenaultHandler.java
jlaur: Sure, also, if you had done that, you wouldn't have noticed compiler error because of missing import. :) I guess if would make sense so set UNDEF also if case of UNKNOWN in line 356. Something like:
```java
LockStatus lockStatus = car.getLockStatus();
updateState(CHANNEL_LOCK_STATUS, localStatus == UNKNOWN ? UnDefType.UNDEF : new StringType(lockStatus.name()));
```

dougculnane: @jlaur On reflection keeping the LockStatus.UNKNOWN is helpful if there is a value delivered by the API that I did not expect.  It will give me a clue that I have to debug and add this value mapping.

However I am happy to map UNKNOWN to UNDEF if you think this is better.

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/handler/RenaultHandler.java
dougculnane: I am saying that RenaultForbiddenException and RenaultUpdateException cannot happen **IF** RenaultNotImplementedException was already was thrown prior to that.  

RenaultUpdateException etc can be temporary but the API is a bit unintuitive and it is hard to test...

However the binding is working well on my computer at the moment.

jlaur: Okay, let's see how this will work out in real life. I was concerned that these two exceptions will **not** trigger  `setDisableLockStatus`, so there could be a risk of repeating warnings in log.

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/handler/RenaultHandler.java
jlaur: So perhaps it would then be better to add a **debug** log with this information, but preferably it should also contain the actual value received from the API, unless there is already a **trace** log providing the full response payload?

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/handler/RenaultHandler.java
jlaur: Let's leave this for this PR, you could then improve this further in a new PR if you would like to.

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/handler/RenaultHandler.java
dougculnane: I do a DEBUG log that has the API  request / responses.  Should these be TRACE logs?

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/api/MyRenaultHttpSession.java
```diff
@@ -87,9 +88,11 @@ private void login() throws RenaultException, InterruptedException, ExecutionExc
         fields.add("ApiKey", this.constants.getGigyaApiKey());
         fields.add("loginID", config.myRenaultUsername);
         fields.add("password", config.myRenaultPassword);
-        logger.debug("URL: {}/accounts.login", this.constants.getGigyaRootUrl());
-        ContentResponse response = httpClient.FORM(this.constants.getGigyaRootUrl() + "/accounts.login", fields);
+        final String url = this.constants.getGigyaRootUrl() + "/accounts.login";
+        ContentResponse response = httpClient.FORM(url, fields);
         if (HttpStatus.OK_200 == response.getStatus()) {
+            logger.debug("GigyaApi Request: {} Response: [{}] {}\n{}", url, response.getStatus(), response.getReason(),
```
jlaur, line 94: Dump of full response payload should be **trace** level:
```suggestion
            logger.trace("GigyaApi Request: {} Response: [{}] {}\n{}", url, response.getStatus(), response.getReason(),
```

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/handler/RenaultHandler.java
jlaur: Yes, added separate comment for that. So I think **UNDEF** would be correct. If a user would report this, trace logs can be enabled, and you should have everything needed for investigating this. Out of curiosity: Can lock status be anything other than locked, unlocked and unknown? Perhaps a read-only switch would be even better then? Some cars even support locking/unlocking through the API.

dougculnane: I have changed to API calls to trace.

The API returns either locked or unlocked as a string.  I need an unknown state which is now covered by UNDEF.  So it is technically possible to change the locked status channel to a locked Switch boolean type.  

I do not know if the API will return anything other that the 2 states.  I also do not know why they did not choose a boolean type.  I used a string for flexibility if the API changes.  Also each door status has a opened status... and so a string of "unlocked and 2 doors open" could be added later if it is an information string.  However not sure this is useful in a home automatization context.

I am not sure how users will use the locked status.  I guess a home security rule could check the location and locked status of the car as part of a openHAB home security system rule set.  

That is the design background.  What do you think is better a flexible status string or a Boolean locked?

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/api/MyRenaultHttpSession.java
dougculnane: I have changed this.

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/handler/RenaultHandler.java
jlaur: With the current knowledge, a switch type channel seems more appropriate for mapping this into an openHAB context. If you would later find something like "unlocked and 2 doors open", that would result in UNDEF anyway, so changes would need to be made in the binding. In this case it might even make sense to have both a `locked` channel as well as `doorsOpen` or whatever can be derived from such API strings.

jlaur:
> I have been testing and the binding looks good to me.

Thanks for the update. I would just need to ask you to update the I18N file once again after the last commit.

bundles/org.openhab.binding.renault/src/main/resources/OH-INF/thing/thing-types.xml
```diff
@@ -189,9 +199,16 @@
...
-		<label>Location Update</label>
+		<label>Location Updated</label>
 		<description>Timestamp of the last location update</description>
+		<category>Time</category>
 		<state pattern="%1$tH:%1$tM %1$td.%1$tm.%1$tY" readOnly="true"></state>
 	</channel-type>
+	<channel-type id="locked">
+		<item-type>Switch</item-type>
+		<label>Locked</label>
+		<description>Locked status of the car</description>
```
jlaur, line 210: How about this? 🙂 
```suggestion
		<description>Locked status of the car</description>
		<category>Lock</category>
```

jlaur (approved):
Thanks for the enhancements!

dougculnane:
> Thanks for the enhancements!

Thanks for the review.

PR #12757, opened 2022-05-16
Updated readme, added how to generate i18n en properties file/kar files/use -pl option

Hilbrand:
- Also put command in blocks to make them stand out more in the readme.

PR #14077, opened 2022-12-27
Update i18n command version number

holgerfriedrich:
update i18n example command line to ```3.4.0```

For me it makes more sense to use the stable version ```3.4.0```, even if we are on snapshot.
```4.0.0-SNAPSHOT``` would work as well, but always fetch a lot of artifacts.

### DIFF
--- a/README.md
+++ b/README.md
@@ -125,7 +125,7 @@ In some cases the command does not work, and requires the full plug-in name.
 In that case use:
 
 ```shell
-mvn org.openhab.core.tools:i18n-maven-plugin:3.3.0:generate-default-translations
+mvn org.openhab.core.tools:i18n-maven-plugin:3.4.0:generate-default-translations
 ```
 
 


kaikreuzer (approved):
Makes sense, thanks!

jlaur:
@holgerfriedrich - actually the version number is not needed anymore. See https://www.openhab.org/docs/developer/utils/i18n.html#generating-i18n-properties-file.

holgerfriedrich:
@jlaur reasonable to keep the version number out in the docs - and the README as well....

I am not sure which special cases the README references here, so I am not sure if it would be safe to remove that part.
grafik

Maybe ```mvn org.openhab.core.tools:i18n-maven-plugin:generate-default-translations``` works as well for the fallback case.

Hilbrand:
@holgerfriedrich the special case was added after a comment from @lolodomo https://github.com/openhab/openhab-addons/pull/12757#discussion_r874098500

jlaur:
@holgerfriedrich, @Hilbrand, @lolodomo - sorry, I missed that this was a special case. Perhaps we could just mention something like "x.y" and add some description to replace with current version number, so we don't have to update this line over and over again?

Issue #12767, opened 2022-05-20
[astro] LocalSun_SunPhaseName giving wrong results


Elgin1:
## Expected Behavior
LocalSun_SunPhaseName should correspond to the time of day and certainly not be UNDEF.

## Current Behavior
Today I noticed that LocalSun_SunPhaseName was NIGHT when it should have been DAYLIGHT

My event logs of the past few days show some unexpected behaviours (comment in ()):

`2022-05-17 23:16:00.007 [INFO ] [openhab.event.ItemStateChangedEvent ] - Item 'LocalSun_SunPhaseName' changed from NAUTIC_DUSK to ASTRO_DUSK
**2022-05-18 00:00:30.696 [INFO ] [openhab.event.ItemStateChangedEvent ] - Item 'LocalSun_SunPhaseName' changed from ASTRO_DUSK to UNDEF
2022-05-18 00:59:37.217 [INFO ] [openhab.event.ItemStateChangedEvent ] - Item 'LocalSun_SunPhaseName' changed from UNDEF to NIGHT**
2022-05-18 02:17:00.006 [INFO ] [openhab.event.ItemStateChangedEvent ] - Item 'LocalSun_SunPhaseName' changed from NIGHT to ASTRO_DAWN

2022-05-18 23:19:00.006 [INFO ] [openhab.event.ItemStateChangedEvent ] - Item 'LocalSun_SunPhaseName' changed from NAUTIC_DUSK to ASTRO_DUSK
**2022-05-19 00:00:30.701 [INFO ] [openhab.event.ItemStateChangedEvent ] - Item 'LocalSun_SunPhaseName' changed from ASTRO_DUSK to UNDEF
2022-05-19 01:09:37.216 [INFO ] [openhab.event.ItemStateChangedEvent ] - Item 'LocalSun_SunPhaseName' changed from UNDEF to NIGHT**
2022-05-19 02:06:00.005 [INFO ] [openhab.event.ItemStateChangedEvent ] - Item 'LocalSun_SunPhaseName' changed from NIGHT to ASTRO_DAWN

2022-05-19 23:21:00.004 [INFO ] [openhab.event.ItemStateChangedEvent ] - Item 'LocalSun_SunPhaseName' changed from NAUTIC_DUSK to ASTRO_DUSK
[...]

Issue #12746, opened 2022-05-16
[astro] `EVENING_NIGHT` and `MORNING_NIGHT` are `undef`


ThaDaVos:
## Expected Behavior

Them to not be `undef` so they can be relied on in Time of Day situations, like I used to use them for in combination with motion sensors to trigger lights

## Current Behavior

Currently they show as `undef` which breaks my `Time of Day` system, after looking through all other states, there's none I can really use to determine the start of the evening correctly, see below for the `MORNING_NIGHT` and `EVENING_NIGHT`:
![image](https://user-images.githubusercontent.com/5251028/168542708-c86d091c-4f6a-42f7-8167-a05e93eeec04.png)

## Possible Solution

Can this be fixed? Or perhaps a workaround and reason given to why they are `undef`?

## Steps to Reproduce (for Bugs)

1. Sorry don't know, they are just `undef` randomly on 16th of May 2022 (the day I noticed they were)

## Context

I was using them for the Marketplace Time of Day rule to keep an up to date status of the Time of Day, currently used for motion sensors and lights

## Your Environment

* Version used: 3.3.0.M3
* Environment name and version (e.g. Chrome 76, Java 8, Node.js 12.9, ...): Openhabian
* Operating System and version (desktop or mobile, Windows 10, Raspbian Buster, ...): Openhabian

lsiepel:
Currently investigating similar issues in https://github.com/openhab/openhab-addons/pull/14078
Can you give me the following details?
long/lat/altitude/timezone of openhab/server   (if you don't feel comfortable with giving those, please DM me details.

ThaDaVos:
@lsiepel - I created a paste for you with the data - the password is your username (as written in the @...) - know the paste is burned after reading:
https://pastebin.com/BpP0L8yN

ThaDaVos:
Also @lsiepel I checked and they are no longer `UNDEF` at the moment

lsiepel:
Thanks, will try to reproduce with the details you provided.

lsiepel:
Undef is not something i would expect. But  i have seen strange behaviour when troubleshooting this and other tickets all around  12-22 may. It has to do with some inner lists of sun phases that did not get any sorting. Aftwarewards i have made several tess with the parameters you gave and i have not found any issue again.

Issue #7642, opened 2020-05-14
[astro] Sun phase name incorrect - NIGHT when it should be DAYLIGHT


Dopeyr:
## Expected Behavior

The "Sun Phase Name" to always be correct.

## Current Behavior

The "Sun Phase Name" can be incorrect on certain days, at certain times.

On 13th May 2020 I found that at 05:17 the phase name switched to DAYLIGHT, then a few seconds later to NIGHT. 

* A openhab restart at 09:23 still showed the phase as NIGHT. 
* A openhab restart at 12:12 showed the phase correctly as DAYLIGHT.

This is the events.log filtered by phase name:

```
2020-05-12 13:07:09.285 [vent.ItemStateChangedEvent] - astro_sun_9fe96e5d_phase_name changed from DAYLIGHT to NOON
2020-05-12 13:12:09.284 [vent.ItemStateChangedEvent] - astro_sun_9fe96e5d_phase_name changed from NOON to DAYLIGHT
2020-05-12 20:56:00.010 [vent.ItemStateChangedEvent] - astro_sun_9fe96e5d_phase_name changed from DAYLIGHT to SUN_SET
2020-05-12 21:00:00.008 [vent.ItemStateChangedEvent] - astro_sun_9fe96e5d_phase_name changed from SUN_SET to CIVIL_DUSK
2020-05-12 21:44:00.006 [vent.ItemStateChangedEvent] - astro_sun_9fe96e5d_phase_name changed from CIVIL_DUSK to NAUTIC_DUSK
2020-05-12 22:45:00.009 [vent.ItemStateChangedEvent] - astro_sun_9fe96e5d_phase_name changed from NAUTIC_DUSK to ASTRO_DUSK
2020-05-13 00:00:30.396 [vent.ItemStateChangedEvent] - astro_sun_9fe96e5d_phase_name changed from ASTRO_DUSK to ASTRO_DAWN
2020-05-13 03:27:00.007 [vent.ItemStateChangedEvent] - astro_sun_9fe96e5d_phase_name changed from ASTRO_DAWN to NAUTIC_DAWN
2020-05-13 03:27:09.284 [vent.ItemStateChangedEvent] - astro_sun_9fe96e5d_phase_name changed from NAUTIC_DAWN to NIGHT
2020-05-13 04:29:00.006 [vent.ItemStateChangedEvent] - astro_sun_9fe96e5d_phase_name changed from NIGHT to CIVIL_DAWN
[...]

PR #14078, opened 2022-12-27
[astro] Fix returning wrong sun phase name

lsiepel:
While troubleshooting [issue 7642](https://github.com/openhab/openhab-addons/issues/7642) first a minor bug was found in the calculation of sun phases.

In `Sun.getSunInfo` method for each phase (sunrise, sunset etc.) a range (start/end datetime) where that phase is valid for is calculated. At the end of all the calculations the **current datetime** instead of the **method parameter** was checked against these ranges to determine the current phase. 
This behavior is for almost all use-cases perfectly fine as the binding usually only shows the current sun phase. But when performing unit tests, this limits you to the current date and prevents you from checking historical phases etc. So, i changed it to the method parameter.

Also added three tests related to the issue. This revealed that the Map holding the ranges for all subphases is not sorted. giving strange results when null values get involved.

So, two fixes:
1. Fixed sorting of sun phases /ranges
2. Using SunPhase Calendar parameter instead of `Calendar.Instance` (now)

Closes https://github.com/openhab/openhab-addons/issues/7642
Closes https://github.com/openhab/openhab-addons/issues/12746
Closes https://github.com/openhab/openhab-addons/issues/12767


clinique (approved):

bundles/org.openhab.binding.astro/src/main/java/org/openhab/binding/astro/internal/model/MoonPhase.java
```diff
@@ -85,14 +85,14 @@ public void setThirdQuarter(Calendar thirdQuarter) {
...
-        return _new;
+        return newCalendar;
     }
 
     /**
      * Sets the date of the new moon.
      */
     public void setNew(Calendar _new) {
-        this._new = _new;
+        this.newCalendar = _new;
```
lolodomo, line 95: You could also rename the parameter.

lolodomo (changes requested):

lolodomo (approved):
LGTM, thank you

lolodomo:
Is it something we should cherry-pick in 3.4.x ? 
@jlaur

jlaur:
@lsiepel - good job hunting and nailing this long-time bug!

@lolodomo - I'm not sure. It's an equation of risk vs. impact. It seems this bug happens rarely (next time May?), but how sure are we that this fix will not cause any regressions? Perhaps we should leave it for 4.0 so we would have a full release cycle for detecting any regressions. I didn't look closely into the details, but please correct me if you believe the risk is negligible.

lolodomo:
> It seems this bug happens rarely (next time May?), but how sure are we that this fix will not cause any regressions? Perhaps we should leave it for 4.0 so we would have a full release cycle for detecting any regressions.

Ok

lsiepel:
> @lsiepel - good job hunting and nailing this long-time bug!
> 
> @lolodomo - I'm not sure. It's an equation of risk vs. impact. It seems this bug happens rarely (next time May?), but how sure are we that this fix will not cause any regressions? Perhaps we should leave it for 4.0 so we would have a full release cycle for detecting any regressions. I didn't look closely into the details, but please correct me if you believe the risk is negligible.

Agree, hope the milestones will get some feedback.

PR #9385, opened 2020-12-15
[atlona] Add support for the AT-PRO3HD66M

mlobstein:
Added support for the older (non-UHD) model AT-PRO3HD66M matrix. This device required a slightly different login protocol and selective bypassing of some of the status inquires. This was necessary because sending large amounts of commands to the device that it did not understand would cause it to lock up.

@tmrobert8

PR #14080, opened 2022-12-28
[atlona] Add support for AT-PRO3HD44M

mlobstein:
follow up to #9385
Add support for the older model AT-PRO3HD44M 4x4 matrix that uses the same protocol as the AT-PRO3HD66M matrix. Also fix binding name to follow standard naming conventions.


lsiepel:
Would it be needed to re-create the thing because the channel type changed ? to make use of the system provided ones? If so this is a breaking  change.

mlobstein:
> Would it be needed to re-create the thing because the channel type changed ? to make use of the system provided ones? If so this is a breaking change.
That is a good point. I don't think it warrants the trouble so I reverted those changes.

lsiepel:
As 4.0.0 is a release that is gonna break more stuff, this is the right moment to make such changes. Using the system channel type is an improvement.

lolodomo (approved):
LGTM, thank you

PR #14086, opened 2022-12-28
[miele] Add categories and semantic tags

jlaur:
Add some categories and semantic tags where missing and obvious.

### DIFF
--- a/bundles/org.openhab.binding.miele/src/main/resources/OH-INF/thing/channeltypes.xml
+++ b/bundles/org.openhab.binding.miele/src/main/resources/OH-INF/thing/channeltypes.xml
@@ -104,6 +104,7 @@
 		<item-type>Switch</item-type>
 		<label>Switch</label>
 		<description>Switch the appliance on or off</description>
+		<category>Switch</category>
 	</channel-type>
 
 	<channel-type id="stop">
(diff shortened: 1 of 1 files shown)


bundles/org.openhab.binding.miele/src/main/resources/OH-INF/thing/channeltypes.xml
```diff
@@ -125,6 +126,10 @@
 		<label>Temperature</label>
 		<description>Temperature reported by the appliance</description>
 		<category>Temperature</category>
+		<tags>
+			<tag>Measurement</tag>
+			<tag>Temperature</tag>
+		</tags>
```
lolodomo, line 132: This one is apparently used for the oven thing.
I am not sure that it is really a measurement.

jlaur: You are right. I don't know what they are as they are not available for my oven. My oven has a meat probe, but it's dark magic and only used for automatic programs. I can't read out the temperature on the oven display or through the API. I guess I was hoping those **temp1** and **temp2** channels could be something like this for supported ovens, but I really don't know. I have removed the tags.

lolodomo (approved):
LGTM

PR #14095, opened 2022-12-28
[miele] Rename handler to fix case

jlaur:
Rename handler to fix case and make consistent with corresponding channel selector.

### DIFF
--- a/bundles/org.openhab.binding.miele/src/main/java/org/openhab/binding/miele/internal/MieleHandlerFactory.java
+++ b/bundles/org.openhab.binding.miele/src/main/java/org/openhab/binding/miele/internal/MieleHandlerFactory.java
@@ -26,7 +26,7 @@ import org.eclipse.jdt.annotation.Nullable;
 import org.eclipse.jetty.client.HttpClient;
 import org.openhab.binding.miele.internal.discovery.MieleApplianceDiscoveryService;
 import org.openhab.binding.miele.internal.handler.CoffeeMachineHandler;
-import org.openhab.binding.miele.internal.handler.DishWasherHandler;
+import org.openhab.binding.miele.internal.handler.DishwasherHandler;
 import org.openhab.binding.miele.internal.handler.FridgeFreezerHandler;
 import org.openhab.binding.miele.internal.handler.FridgeHandler;
 import org.openhab.binding.miele.internal.handler.HobHandler;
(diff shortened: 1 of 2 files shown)


lolodomo:
We have WashingMachineHandler and not WashingmachioneHandler.
While would you like to make an exception for DishWasherHandler ?

jlaur:
> We have WashingMachineHandler and not WashingmachioneHandler.
> While would you like to make an exception for DishWasherHandler ?

It's an old inconsistency that has been bugging me - `DishwasherChannelSelector` is correct, but `DishWasherHandler` not. It's not an exception:
- Washing Machine -> WashingMachine
- Tumble Dryer -> TumbleDryer
- Dishwasher -> Dishwasher

Dishwasher is one word, it's not "Dish Washer".

lolodomo:
> Dishwasher is one word, it's not "Dish Washer".

Didn't know that.

lolodomo (approved):
LGTM

jlaur:
> > Dishwasher is one word, it's not "Dish Washer".
> 
> Didn't know that.

I cheated - I had to look it up to be sure. :-)

mhilbush:
@jlaur I tried to update my local addons repo today. When running the merge (git merge origin/main), I get the following.

```
Updating 0357049f9f..31fd91129e
error: The following untracked working tree files would be overwritten by merge:
	bundles/org.openhab.binding.miele/src/main/java/org/openhab/binding/miele/internal/handler/DishwasherHandler.java
Please move or remove them before you merge.
Aborting
```

I don't think I should "move or remove" that file because that's the new name of the handler. I'm not sure how to correctly resolve. Any suggestions?

J-N-K:
That‘s a known issue on windows because it is case-insensitive for file names. It‘s probably easiest to hard-reset.

jlaur:
@mhilbush - are you on Windows? In general this is a mess and also the reason why I hesitated this long before fixing this. I'm running Windows on my development machine, so file system is case insensitive. Therefore I did a git `mv DishWasherHandler.java DishwasherHandler.java` to get this correctly picked up by Git. You could try renaming your local file to Dish**w**asherHandler.java before performing the merge in order to avoid the filename conflict.

jlaur:
@mhilbush - you could also check this option: https://www.git-scm.com/docs/git-config/2.14.6#Documentation/git-config.txt-coreignoreCase

J-N-K:
I use SmartGit, that seems to handle that just fine.

lolodomo:
You just have to set ignore case to true in git config. 
And then unset it when your command succeeded.

mhilbush:
SmartGit looks pretty nice - I applied for a non-commercial license.

When I opened my `openhab-addons` repo in SmartGit, it complained about `core.ignoreCase` not being set properly. So I set it to true and that also resolved the issue I was having running git from the command line.

Thanks for the help guys!

PR #14093, opened 2022-12-28
[jsscripting] Upgrade to ECMAScript 2022 & Dependency improvements

florian-h05:
## Description

- Remove `org.ow2.asm` from dependency, as it is not required by GraalJS since 20.3.0
- Remove `org.graalvm.js`/js-launcher dependency, as we don't need the JavaScript command line interpreter
- ~Disable Nashorn compatiblity mode as it's additional functionality is not used by openhab-js and very unlikely used by any user.~ Don't disable Nashorn compat mode as the library relies on some Nashorn accessors: https://github.com/oracle/graaljs/blob/master/docs/user/NashornMigrationGuide.md#accessors.
- **Enable new ECMAScript-2022 features like private properties for classes**

**Note: I also wanted to upgrade to GraalJS 22.3.0, but it is failing with the same error message as on OpenJDK 11 on armv7l, so we cannot upgrade :cry:** 

## Testing

I have tested this change on these system/JDK combinations and it works on both of them:

x86_64: `Linux fedora 6.0.15-200.fc36.x86_64 #1 SMP PREEMPT_DYNAMIC Wed Dec 21 18:46:09 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux` with 
```
openjdk version "17.0.5" 2022-10-18
OpenJDK Runtime Environment GraalVM CE 22.3.0 (build 17.0.5+8-jvmci-22.3-b08)
OpenJDK 64-Bit Server VM GraalVM CE 22.3.0 (build 17.0.5+8-jvmci-22.3-b08, mixed mode, sharing)
```
and
```
openjdk version "17.0.5" 2022-10-18
OpenJDK Runtime Environment (Red_Hat-17.0.5.0.8-3.fc36) (build 17.0.5+8)
OpenJDK 64-Bit Server VM (Red_Hat-17.0.5.0.8-3.fc36) (build 17.0.5+8, mixed mode, sharing)
```

Raspberry Pi 3B+: `Linux homeberrypi 5.10.103-v7+ #1529 SMP Tue Mar 8 12:21:37 GMT 2022 armv7l GNU/Linux` with
```
openjdk 17.0.5 2022-10-18
OpenJDK Runtime Environment Zulu17.38+21-CA (build 17.0.5+8)
OpenJDK Server VM Zulu17.38+21-CA (build 17.0.5+8, mixed mode)
```

PR #14096, opened 2022-12-28
[jsscripting] Use `application/javascript` as MIME type

florian-h05:
With openHAB on the horizon and the NashornJS engine available as an addon with MIME type `application/javascript;version=ECMAScript-5.1`, GraalJS should take over the `application/javascript` MIME type.


florian-h05:
@jlaur I've added a unit test that calls the `GraalJSScriptEngineFactory`, but the test is failing before it even starts and I don't understand why. Can you please help me?

J-N-K:
Please make sure that `application/javascript;version=ECMAScript-2021` still works. Otherwise user would have to re-create all UI based rules.

florian-h05:
@J-N-K 

I was not sure if we want to keep that backward compatiblity (FYI: #14093 upgrades to ECMA2022), I had it in but removed it then. I'll just revert the removal. FYI, I have a UI PR that will add a menu to change the script type: https://github.com/openhab/openhab-webui/pull/1601.

florian-h05:
@J-N-K Can you help me with the failing unit test? It's failing before it even starts ...

J-N-K:
I guess it's the same as [here](https://stackoverflow.com/questions/74208293/maven-error-failed-to-run-the-ecj-compiler-unrecognized-option-patch-modul), but I have no idea why it failes in `automation.jsscripting` but not in `binding.dmx`.

florian-h05:
The test is nearly the same as for [JS Scripting (Nashorn)](https://github.com/openhab/openhab-addons/blob/main/bundles/org.openhab.automation.jsscriptingnashorn/src/test/java/org/openhab/automation/jsscriptingnashorn/NashornScriptEngineFactoryTest.java) and there it works. As no one has an idea why it fails, I’ll remove it.

J-N-K:
It‘s probably related to something on the classpath.

wborn (approved):
Thanks!

PR #14099, opened 2022-12-29
[mqtt] Fix shutdownRetain default value

jimtng:
The documentation and java code set the default for `shutdownRetain` to true, but the thing-types.xml sets it to false. I came across a bug when I didn't specify `shutdownRetain` in my mqtt bridge config, and it appears to have been set to false, when I expected it to default to true. This resulted in an unexpected behaviour.

The workaround is to explicitly set `shutdownRetain` to true in the thing config, until this PR is merged.


lolodomo (approved):
LGTM, thank you

PR #14103, opened 2022-12-29
[bondhome] Fix fatal Null Pointer errors

mlobstein:
I just got a bond bridge today. I tried to use the binding and ran into several fatal errors that stopped the binding from functioning:

1. When I changed the IP of the bridge the binding would fail connecting to the old address. An error in the translation file caused an exception when calling setBridgeOffline(),
2. In my bridge, I set up a 'Light' aka 'LT' item and this is apparently not supported by BondDeviceType. This caused the discovery service to fail when the BondDevice with a null 'type' was encountered.
3. The list of actions in one or more BondDevice objects contained null entries that would cause an NPE in deleteExtraChannels()

### DIFF
--- a/bundles/org.openhab.binding.bondhome/src/main/java/org/openhab/binding/bondhome/internal/handler/BondDeviceHandler.java
+++ b/bundles/org.openhab.binding.bondhome/src/main/java/org/openhab/binding/bondhome/internal/handler/BondDeviceHandler.java
@@ -554,12 +554,10 @@ public class BondDeviceHandler extends BaseThingHandler {
         Set<String> availableChannelIds = new HashSet<>();
 
         for (BondDeviceAction action : currentActions) {
-            if (action != null) {
-                String actionType = action.getChannelTypeId();
-                if (actionType != null) {
-                    availableChannelIds.add(actionType);
-                    logger.trace(" Action: {}, Relevant Channel Type Id: {}", action.getActionId(), actionType);
-                }
+            String actionType = action.getChannelTypeId();
+            if (actionType != null) {
+                availableChannelIds.add(actionType);
+                logger.trace(" Action: {}, Relevant Channel Type Id: {}", action.getActionId(), actionType);
             }
         }
         // Remove power channels if we have a dimmer channel for them;
@@ -627,7 +625,7 @@ public class BondDeviceHandler extends BaseThingHandler {
             BondDeviceProperties devProperties = this.deviceProperties;
             if (devProperties != null) {
                 double maxSpeed = devProperties.maxSpeed;
-                value = (int) (((double) updateState.speed / maxSpeed) * 100);
+                value = (int) ((updateState.speed / maxSpeed) * 100);
                 logger.trace("Raw fan speed: {}, Percent: {}", updateState.speed, value);
             } else if (updateState.speed != 0 && this.getThing().getThingTypeUID().equals(THING_TYPE_BOND_FAN)) {
                 logger.info("Unable to convert fan speed to a percent for {}!", this.getThing().getLabel());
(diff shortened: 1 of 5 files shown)


lsiepel (commented):
LGTM

ccutrer (approved):

lsiepel (approved):

bundles/org.openhab.binding.bondhome/src/main/java/org/openhab/binding/bondhome/internal/handler/BondDeviceHandler.java
```diff
@@ -554,10 +554,12 @@ private synchronized void deleteExtraChannels(List<BondDeviceAction> currentActi
         Set<String> availableChannelIds = new HashSet<>();
 
         for (BondDeviceAction action : currentActions) {
-            String actionType = action.getChannelTypeId();
-            if (actionType != null) {
-                availableChannelIds.add(actionType);
-                logger.trace(" Action: {}, Relevant Channel Type Id: {}", action.getActionId(), actionType);
+            if (action != null) {
```
lolodomo, line 557: Can `action` really be null ?

mlobstein: There were null items in the currentActions list using the 3.4.0 release code:
image

lolodomo: in case it is not a bug (I mean it is really expected to have some null entries), I would to replace in BondDevice  class:
```
    public List<BondDeviceAction> actions = Arrays.asList(BondDeviceAction.TURN_ON);
```
by
```
    public List<@Nullable BondDeviceAction> actions = Arrays.asList(BondDeviceAction.TURN_ON);
```

ccutrer: I'd vote for making the list NonNullable, and fix whatever populates it to prune out actions that are not supported

mlobstein: It is currently non-nullable and is set by the gson/json parser. I guess one could do some post processing in BondHttpApi.getDevice() after the gson parsing and before the BondDevice is returned... Something like device.actions.removeIf(null);

ccutrer: Oh eww. So presumably gson is bypassing the NonNull restriction because there isn't an enum value for whatever your device's are advertising. Force removing nulls still sounds best to me, immediately after parsing. But I hope you have another PR coming to add support for your device's actions :)

mlobstein: Fixed. The unsupported actions are "TiltClose" and "TiltOpen". Once I get time to dig into the device api, I will try to add the missing features.

lolodomo (approved):
LGTM, thank you

jlaur:
@ccutrer, @lolodomo, @mlobstein  - should we include this in the upcoming 3.4.1 hotfix? In terms of risk, there is probably not a lot of users yet since it's a new binding. But it seems this bug could be quite annoying for impacted new users, so seems critical?

lolodomo:
Take care, there was another PR decent merged. If this one is cherry picked in 3.4.x, the other should be done first.

lolodomo:
I am talking about PR #13986 .

Issue #8955, opened 2020-11-04
[chromecast] Things go offline after a while


andreasscherbaum:
## Current Behavior

I have monitoring for my bindings and most of the Things in place. All my ChromeCasts are monitored, and every hour I get a summary about all the ChromeCasts which are marked offline. Over time (sometimes couple days, sometimes only hours) I can see that more and more ChromeCast devices are marked offline, and never come back online.

When I restart just the binding the devices all come online again.

Enabling logging for the binding reveals not much, but a bit of "unknown" messages:
```
log:set DEBUG org.openhab.binding.chromecast
```

```
2020-11-04 17:33:55.364 [DEBUG] [overy.ChromecastDiscoveryParticipant] - Chromecast Type: null
2020-11-04 17:34:57.105 [DEBUG] [overy.ChromecastDiscoveryParticipant] - Chromecast Type: null
2020-11-04 17:34:57.108 [DEBUG] [overy.ChromecastDiscoveryParticipant] - Chromecast Type: Chromecast Ultra
2020-11-04 17:34:57.111 [DEBUG] [overy.ChromecastDiscoveryParticipant] - Chromecast Found: 192.168.0.67 8009
2020-11-04 17:34:57.113 [DEBUG] [overy.ChromecastDiscoveryParticipant] - Chromecast Type: Chromecast Ultra
2020-11-04 17:37:51.947 [DEBUG] [ast.internal.ChromecastEventReceiver] - Received an 'UNKNOWN' event (class=class com.fasterxml.jackson.databind.JsonNode)
2020-11-04 17:37:51.977 [DEBUG] [ast.internal.ChromecastEventReceiver] - Received an 'UNKNOWN' event (class=class com.fasterxml.jackson.databind.JsonNode)
2020-11-04 17:37:52.049 [DEBUG] [ast.internal.ChromecastEventReceiver] - Received an 'UNKNOWN' event (class=class com.fasterxml.jackson.databind.JsonNode)
2020-11-04 17:37:52.049 [DEBUG] [ast.internal.ChromecastEventReceiver] - Received an 'UNKNOWN' event (class=class com.fasterxml.jackson.databind.JsonNode)
```
Not sure which message belongs to which ChromeCast though, we have a couple of them connected here.

## Expected Behavior

The Thing should not go offline, or should come back online again. The physical device is available in the network, and can be used for streaming. It's just openHAB not getting that info.

## Steps to Reproduce (for Bugs)

[...]

PR #14105, opened 2022-12-29
[chromecast] Fix constant disconnections

jimtng:
When ipv6 is configured on the network, Chromecast devices will randomly return ipv4 and/or multiple ipv6 addresses during mdns discovery. As a result of this, the discovered ip address keeps changing which would trigger a thingUpdated call which will then call dispose/initialize, causing the thing to go offline/online.

I'm experiencing this issue and it's also reported in https://community.openhab.org/t/openhab-3-4-release-discussion/142257/98

### DIFF
--- a/bundles/org.openhab.binding.chromecast/src/main/java/org/openhab/binding/chromecast/internal/discovery/ChromecastDiscoveryParticipant.java
+++ b/bundles/org.openhab.binding.chromecast/src/main/java/org/openhab/binding/chromecast/internal/discovery/ChromecastDiscoveryParticipant.java
@@ -14,6 +14,7 @@ package org.openhab.binding.chromecast.internal.discovery;
 
 import static org.openhab.binding.chromecast.internal.ChromecastBindingConstants.*;
 
+import java.net.Inet4Address;
 import java.util.Dictionary;
 import java.util.Map;
 import java.util.Set;
@@ -89,7 +90,11 @@ public class ChromecastDiscoveryParticipant implements MDNSDiscoveryParticipant 
         if (isAutoDiscoveryEnabled) {
             ThingUID uid = getThingUID(service);
             if (uid != null) {
-                String host = service.getHostAddresses()[0];
+                Inet4Address[] addresses = service.getInet4Addresses();
+                if (addresses.length == 0) {
+                    return null;
+                }
+                String host = addresses[0].getHostAddress();
                 int port = service.getPort();
                 logger.debug("Chromecast Found: {} {}", host, port);
                 String id = service.getPropertyString(PROPERTY_DEVICE_ID);


lolodomo (approved):
LGTM, thank you

lolodomo:
@jlaur : please cherry-pick this PR in 3.4.x.

lolodomo:
This would be a good idea to check in other bindings if a similar fix has to be applied. 
getHostAddresses is found 17 times.

jimtng:
> This would be a good idea to check in other bindings if a similar fix has to be applied. getHostAddresses is found 17 times.

* This could just be Google nest's idiosyncrasy. 
* In case it happens to others, perhaps in core (in MDNSDiscoveryService), filter out ServiceInfo that contains zero ipv4addresses - although this seems like a bad solution.

stefan-hoehn:
First of all, @jimtng , great that you found the route cause and I can confirm it has fixed that problem 🥇 

If I understand the fix right, it makes sure it is only using the IPv4 address (I am not complaining, I am more than happy that it fixes the problem ;-) ) 

Even though I guess the vast majority of people are still using only IPv4s,  we might think about a different solution as well where also v6 is supported. In particular if more bindings would be affected.

jimtng:
> If I understand the fix right, it makes sure it is only using the IPv4 address (I am not complaining, I am more than happy that it fixes the problem ;-) )

Correct. 

The behaviour that I observed is the mdns discovery would sometimes provided _only_ one ipv6 in the getHostAddresses() return value, and at other times it provided both the ipv4 and several, but not all of its ipv6 addresses.

Of the ipv6 addresses that it provided, some of them are unreachable in my system. So I concluded that the ipv6 addresses that we received cannot be relied upon without doing some extra verifications such as making sure they are in the same network, and/or that they are actually reachable. I prefer not having to go to that length just to validate the addresses that were provided in the discovery. 

The only time this would break if someone had completely disabled ipv4 from their network completely. Highly unlikely in the next 10 years at least. 

Perhaps someone who's more familiar with this topic could offer a better solution?

I suspect that there may be a bug in jmdns library that openhab is using for its mdns discovery, but on the other hand, I'm not certain that it's not the google chromecast devices that's acting funny with their mdns discovery e.g. due to a recent update on google's end. Further investigations can be done, but for now I'm happy that this solved the immediate problem.

lsiepel:
Fixes: https://github.com/openhab/openhab-addons/issues/8955

lolodomo:
What we could do for any other binding is to select in priority an IP v4 host address if available, or the first host address if not available. In other words, to give priority to IP v4 over IP v6 when both are provided.

lolodomo:
Or maybe rather sort (in OH core framework) the result of getHostAddresses?

jimtng:
`getHostAddresses` already [returns the ipv4 addresses first](https://github.com/jmdns/jmdns/blob/2c608730527becb19028b2949643425b4c22930d/src/main/java/javax/jmdns/impl/ServiceInfoImpl.java#L412-L423).

The problem is that quite often, `getHostAddresses()` would _only_ return one element, which is a single ipv6 address, and sometimes only several ipv6 addresses, without any ipv4 addresses. It's the reason why I replaced it with `getInet4Addresses()` and when it doesn't return any ipv4 addresses, discard the discovery result.

lolodomo:
So the problem could be really specific to Chromecast and we have probably nothing to do for other bindings. 
By the way, I will check with few bindings I am using how it behaves: hue, free box and remote openHAB.

PR #14128, opened 2022-12-31
[tesla] Synchronize access to account to avoid concurrency issues

kaikreuzer:
When an account contains multiple vehicles, the handler tries to query them concurrently - which results in an exception since the `vehiclesTarget` does not support concurrent executions.


lolodomo (approved):
LGTM, thank you

Issue #11424, opened 2021-10-23
[astro] should respect configured earliest or latest channel time


beulenbilly:
## Your Environment
* Version used: OH 3.1

## Description
Due to a lack of light and temperature sensors I use the astro channels to control the rolloshutter especially in summer. To be sure that is really dark outside and get cooler air inside I use the astronomic dawn. But in summer the range end is null, if you use:
- lat: 53.5214675
- lon: 10.043392
- altitude: 10
- date: 21.06.2021, 8pm

In this case no event is triggered because of org.openhab.binding.astro.internal.job.Job line 122 even if the latest-property is configured for the channel.

## Expected Behavior
If there is a latest time configured for the channel I would expect that the time will be applied.

## Possible Solution
https://github.com/beulenbilly/openhab-addons/commit/2a91a6c312954a887a01ca713996a9ddae3f47ff

I am not really happy about adding the test-dependency as well as mocking the static method calls...

lsiepel:
@beulenbilly i think it might get complex. If both earliest start and end are configured, and the ‘force’ configuration option is set. What value should replace the null? Earliest start or earliest end?

lsiepel:
Think i have a fix, could you look at the PR?

beulenbilly:
Hey @lsiepel the implementation looks really good.
I am not sure if everything works with the time calculation for start and end but your test will proof it.

Thanks.

PR #14132, opened 2023-01-01
[astro] Add option to force event to occur

lsiepel:

Trigger channels have an earliest start / latest end option to 'manualy' control/override event (e.g. sunrise, sunset ) starts/ ends to occur. There are cases that the event is not happening at all (e.g. astro dawn during summer in Sweden). This PR adds an option to the trigger channel config to 'force' the event to take place even when it is actually not happening. For the start/ end it uses the earliest /latest config setting.
this setting defaults to false, so it should not affect existing set-ups.

Also add tests to assure behaviour is consistent with and without the ChannelConfig 'forceEvents' set to true.

Resolves: https://github.com/openhab/openhab-addons/issues/11424


bundles/org.openhab.binding.astro/README.md
```diff
@@ -130,6 +130,12 @@ OR
 
 sunset is 22:10 but `latest` is set to 20:00 so the event/datetime value is moved 20:00.
 
+**Force event:** For each trigger channel and `start`, `end` datetime value, you can force the `earliest`, `latest` time of the day, when the event is actually not taking place (e.g. astronomic dawn during summer in Sweden)
+e.g `sun#astroDawn earliest=8:00, latest=20:00 forceEvent=true`
+
+astronomic dawn start is null but `earliest` is set to 06:00 so the event/datetime value is set to 06:00.
```
lolodomo, line 136: Can you consistent between the example and the explanation, that is either 6 or 8.

bundles/org.openhab.binding.astro/src/main/resources/OH-INF/config/config.xml
```diff
@@ -53,6 +53,11 @@
 			<description>The latest time of the day for the event or the datetime value (hh:mm).</description>
 			<context>time</context>
 		</parameter>
+		<parameter name="forceEvent" type="boolean" required="true">
```
lolodomo, line 56: No need to mark the parameter as required as you provide a default value.

lolodomo (changes requested):

bundles/org.openhab.binding.astro/README.md
lsiepel: :-/

bundles/org.openhab.binding.astro/src/main/resources/OH-INF/config/config.xml
lsiepel: ah, thanks thought it would was necessary to also list it as required.

lolodomo (approved):
LGTM, thank you

PR #14135, opened 2023-01-01
[jsscripting] Cache openhab-js injection to improve performance

florian-h05:
For previous discussion see #14113.

## Description

This PR extends the code caching to the version of the openHAB JavaScript library (openhab-js) that is included in the JS Scripting addon.
A new configuration option is introduced to allow the user to disable usage of the internal openhab-js library and instead load it from the file system.

~**Note: I am currently not super happy with the description and the label of that new configuration option, I would like to get some feedback on them.**~

Code caching of the openhab-js injection brings great performance improvements as the evaluation of the pretty large openhab-js codebase takes some time. 

Note: it is not possible to cache the `Object.assign(this, require("openhab");` therefore I had to take the way of creating an injection script at openhab-js.

I've performed a few test runs on my dev system (that is quite powerful, AMD Ryzen 7) and here are the results:

- openhab-js injection without caching: about 1 second
- openhab-js injection with code caching: between 100 and 200 milliseconds

On my Raspberry Pi 3 (Debian 10 Buster), caching makes a even larger difference:

- openhab-js injection without caching: about 13 seconds
- openhab-js injection with code caching: about 2 seconds

Note that evaluation of the cached version takes a bit longer than usual after the first start of the addon -- the code needs to be evaluated once before it is cached.

## To Do

- [x] Regenerate default translations
- [x] Maybe improve the label and description of the new configuration option

Ping @digitaldan @jlaur

### DIFF
--- a/bundles/org.openhab.automation.jsscripting/src/main/java/org/openhab/automation/jsscripting/internal/OpenhabGraalJSScriptEngine.java
+++ b/bundles/org.openhab.automation.jsscripting/src/main/java/org/openhab/automation/jsscripting/internal/OpenhabGraalJSScriptEngine.java
@@ -125,17 +125,17 @@ public class OpenhabGraalJSScriptEngine
 
     private boolean initialized = false;
     private final boolean injectionEnabled;
-    private final boolean internalLibraryEnabled;
+    private final boolean useIncludedLibrary;
 
     /**
      * Creates an implementation of ScriptEngine (& Invocable), wrapping the contained engine, that tracks the script
      * lifecycle and provides hooks for scripts to do so too.
      */
-    public OpenhabGraalJSScriptEngine(boolean injectionEnabled, boolean internalLibraryEnabled,
+    public OpenhabGraalJSScriptEngine(boolean injectionEnabled, boolean useIncludedLibrary,
             JSScriptServiceUtil jsScriptServiceUtil) {
         super(null); // delegate depends on fields not yet initialised, so we cannot set it immediately
         this.injectionEnabled = injectionEnabled;
-        this.internalLibraryEnabled = internalLibraryEnabled;
+        this.useIncludedLibrary = useIncludedLibrary;
         this.jsRuntimeFeatures = jsScriptServiceUtil.getJSRuntimeFeatures(lock);
 
         LOGGER.debug("Initializing GraalJS script engine...");
@@ -264,7 +264,7 @@ public class OpenhabGraalJSScriptEngine
             LOGGER.debug("Evaluating cached global script...");
             delegate.getPolyglotContext().eval(GLOBAL_SOURCE);
             if (this.injectionEnabled) {
-                if (this.internalLibraryEnabled) {
+                if (this.useIncludedLibrary) {
                     LOGGER.debug("Evaluating cached openhab-js injection...");
                     delegate.getPolyglotContext().eval(OPENHAB_JS_SOURCE);
                 } else {
--- a/bundles/org.openhab.automation.jsscripting/src/main/java/org/openhab/automation/jsscripting/internal/OpenhabGraalJSScriptEngine.java
+++ b/bundles/org.openhab.automation.jsscripting/src/main/java/org/openhab/automation/jsscripting/internal/OpenhabGraalJSScriptEngine.java
@@ -64,7 +64,7 @@ import com.oracle.truffle.js.scriptengine.GraalJSScriptEngine;
  * @author Dan Cunningham - Script injections
  * @author Florian Hotze - Create lock object for multi-thread synchronization; Inject the {@link JSRuntimeFeatures}
  *         into the JS context; Fix memory leak caused by HostObject by making HostAccess reference static; Switch to
- *         {@link Lock} for multi-thread synchronization
+ *         {@link Lock} for multi-thread synchronization; globals & openhab-js injection code caching
  */
 public class OpenhabGraalJSScriptEngine
         extends InvocationInterceptingScriptEngineWithInvocableAndAutoCloseable<GraalJSScriptEngine> {
--- a/bundles/org.openhab.automation.jsscripting/src/main/java/org/openhab/automation/jsscripting/internal/OpenhabGraalJSScriptEngine.java
+++ b/bundles/org.openhab.automation.jsscripting/src/main/java/org/openhab/automation/jsscripting/internal/OpenhabGraalJSScriptEngine.java
@@ -229,13 +229,14 @@ public class OpenhabGraalJSScriptEngine
         ScriptExtensionModuleProvider scriptExtensionModuleProvider = new ScriptExtensionModuleProvider(
                 scriptExtensionAccessor, lock);
 
+        // Wrap the "require" function to also allow loading modules from the ScriptExtensionModuleProvider
         Function<Function<Object[], Object>, Function<String, Object>> wrapRequireFn = originalRequireFn -> moduleName -> scriptExtensionModuleProvider
                 .locatorFor(delegate.getPolyglotContext(), engineIdentifier).locateModule(moduleName)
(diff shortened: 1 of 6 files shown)


bundles/org.openhab.automation.jsscripting/src/main/resources/OH-INF/config/config.xml
```diff
@@ -7,14 +7,27 @@
...
+			If disabled, the openHAB JavaScript library can be imported manually using "<i>require('openhab')</i>"
 			]]></description>
 			<options>
 				<option value="true">Use Built-in Variables</option>
 				<option value="false">Do Not Use Built-in Variables</option>
 			</options>
 			<default>true</default>
 		</parameter>
+		<parameter name="internalLibraryEnabled" type="boolean" required="true">
+			<label>Use Internal Library</label>
```
digitaldan, line 21:
```suggestion
			<label>Use Included openHAB Javascript Library</label>
```

bundles/org.openhab.automation.jsscripting/src/main/resources/OH-INF/config/config.xml
```diff
@@ -7,14 +7,27 @@
...
+			<description><![CDATA[
+			Use the addon-included and cached openHAB JavaScript library for optimal performance.<br>
+			Disable this option to allow loading the library from the disk instead, but be aware that this increases script loading times especially on less powerful systems.
```
digitaldan, line 24:
```suggestion
			Use the included openHAB JavaScript library for optimal performance.<br>
			Disable this option to allow loading the library from the local user configuration directory "automation/js/node_modules".  Using a user provided version of the library may increase script loading times, especially on less powerful systems.
```

bundles/org.openhab.automation.jsscripting/src/main/resources/OH-INF/config/config.xml
```diff
@@ -7,14 +7,27 @@
...
+			]]></description>
+			<options>
+				<option value="true">Use Internal Library</option>
+				<option value="false">Do Not Use Internal Library</option>
```
digitaldan, line 28:
```suggestion
				<option value="true">Use Included Library</option>
				<option value="false">Do Not Use Included Library</option>
```

digitaldan (changes requested):
Great work!  I have a few comments regarding configuration language as you requested, nothing major.  Changes in code looks good to me, was planning on getting my 4.0 environment up today and see if i can test this.

florian-h05:
> Great work! I have a few comments regarding configuration language as you requested, nothing major. 

Thanks :-)

> Changes in code looks good to me, was planning on getting my 4.0 environment up today and see if i can test this.

This PR depends on https://github.com/openhab/openhab-js/pull/218.
FYI: For testing, I generated the webpacked openhab-js with the openhab-js PR and copied it to the `target/js/dist/` folder of the addon. I also locally removed the module bundling from the pom to lower build time. Just use Maven without the `clean` command then.

jlaur:
@florian-h05 - great work again. I'm wondering if it would be possible to also cache the library from `node_modules` and watch the directory in order to invalidate the cached version upon changes? Since latest version is even provided through openHABian, this is quite convenient and I could imagine many "ordinary" users preferring to use this. I already switched and will probably never look back. :-)

florian-h05:
> great work again

Thanks :-)

> I'm wondering if it would be possible to also cache the library from node_modules and watch the directory in order to invalidate the cached version upon changes?

I was trying this, but unfortunantely code caching for loaded code (via `require`) doesn't work (doesn't improve performance). This is because every engine has it's own CommonJS module loader. I was thinking if it would be possible to share the module loader because it is also caching in some ways, but I did not find a docs to do this and looking at the GraalJS code, it seems that the loader is bound to the script.

> Since latest version is even provided through openHABian, this is quite convenient and I could imagine many "ordinary" users preferring to use this.

This is mainly thought for openhab-js updates in case you need a yet not addon included fix or enhancement.
Note that this change makes the addon use the included library unless you explicitely disable this in the addon settings.

digitaldan (approved):
LGTM

bundles/org.openhab.automation.jsscripting/src/main/java/org/openhab/automation/jsscripting/internal/OpenhabGraalJSScriptEngine.java
```diff
@@ -111,15 +124,18 @@
...
-    private final String injectionCode;
+    private final boolean injectionEnabled;
+    private final boolean internalLibraryEnabled;
 
     /**
      * Creates an implementation of ScriptEngine (& Invocable), wrapping the contained engine, that tracks the script
      * lifecycle and provides hooks for scripts to do so too.
      */
-    public OpenhabGraalJSScriptEngine(@Nullable String injectionCode, JSScriptServiceUtil jsScriptServiceUtil) {
+    public OpenhabGraalJSScriptEngine(boolean injectionEnabled, boolean internalLibraryEnabled,
```
jlaur, line 134: See other comment.
```suggestion
    public OpenhabGraalJSScriptEngine(boolean injectionEnabled, boolean includedLibraryEnabled,
```

bundles/org.openhab.automation.jsscripting/src/main/java/org/openhab/automation/jsscripting/internal/OpenhabGraalJSScriptEngine.java
```diff
@@ -111,15 +124,18 @@
     private @Nullable Consumer<String> scriptDependencyListener;
 
     private boolean initialized = false;
-    private final String injectionCode;
+    private final boolean injectionEnabled;
+    private final boolean internalLibraryEnabled;
```
jlaur, line 128: See other comment.
```suggestion
    private final boolean includedLibraryEnabled;
```

jlaur (changes requested):
Thanks. Only a minor comment about consistent naming.

bundles/org.openhab.automation.jsscripting/src/main/resources/OH-INF/i18n/jsscripting.properties
```diff
@@ -1,7 +1,16 @@
...
+automation.config.jsscripting.useIncludedLibrary.option.true = Use Included Library
+automation.config.jsscripting.useIncludedLibrary.option.false = Do Not Use Included Library
+
+automation.config.jsscripting.internalLibraryEnabled.label = Use Included openHAB JavaScript Library
+automation.config.jsscripting.internalLibraryEnabled.description = Use the included openHAB JavaScript library for optimal performance.<br> Disable this option to allow loading the library from the local user configuration directory "automation/js/node_modules". Using a user provided version of the library may increase script loading times, especially on less powerful systems.
+automation.config.jsscripting.internalLibraryEnabled.option.true = Use Included Library
+automation.config.jsscripting.internalLibraryEnabled.option.false = Do Not Use Included Library
```
jlaur, line 13:
```suggestion
```

jlaur (approved):
Thank you!

digitaldan:
@florian-h05 does `require("openhab")` no longer work when dealing with other libraries in the node_modules?  I have a helper script which i use by first  requiring our lib like:

```js
const ns = {};

// assign everything in @oh, so items, actions, triggers, etc....
Object.assign(ns, require("openhab"));
.....
module.exports = ns;
```

i then add some proxy magic for my own helper functions and export the whole thing.  This is throwing errors that it can't find `openhab`.   Did changing the name (@jsscripting-globals) prevent using our lib by the `openhab` name ?

florian-h05:
Oh yes, I see the problem. 
The problem is that there is no `openhab` file in the addon anymore. If you install openhab-js to the disk, it will work again.
I will provide a fix for this.

PR #14138, opened 2023-01-02
[bsblan] Minor SAT fixes

lsiepel:

Minor refactoring:
- add null annotation
- checkstyle

### DIFF
--- a/bundles/org.openhab.binding.bsblan/src/main/java/org/openhab/binding/bsblan/internal/api/BsbLanApiCaller.java
+++ b/bundles/org.openhab.binding.bsblan/src/main/java/org/openhab/binding/bsblan/internal/api/BsbLanApiCaller.java
@@ -17,10 +17,10 @@ import static org.openhab.binding.bsblan.internal.BsbLanBindingConstants.API_TIM
 import java.io.ByteArrayInputStream;
 import java.io.IOException;
 import java.io.InputStream;
-import java.nio.charset.Charset;
 import java.util.Set;
 import java.util.stream.Collectors;
 
+import org.eclipse.californium.elements.util.StandardCharsets;
 import org.eclipse.jdt.annotation.NonNullByDefault;
 import org.eclipse.jdt.annotation.Nullable;
 import org.openhab.binding.bsblan.internal.api.dto.BsbLanApiContentDTO;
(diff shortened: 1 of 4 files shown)


lsiepel:
@jlaur i guess the build failed is not related to this PR? 
Seems something wrong with:  osgi.wiring.package=org.eclipse.californium.elements.util)(version>=2.7.0)(!(version>=3.0.0)))"]

bundles/org.openhab.binding.bsblan/src/main/java/org/openhab/binding/bsblan/internal/api/BsbLanApiCaller.java
```diff
@@ -17,10 +17,10 @@
 import java.io.ByteArrayInputStream;
 import java.io.IOException;
 import java.io.InputStream;
-import java.nio.charset.Charset;
 import java.util.Set;
 import java.util.stream.Collectors;
 
+import org.eclipse.californium.elements.util.StandardCharsets;
```
lolodomo, line 23: Why this strange import? 
You can get UTV 8 standard charger from common Java class.

lolodomo (changes requested):

bundles/org.openhab.binding.bsblan/src/main/java/org/openhab/binding/bsblan/internal/api/BsbLanApiCaller.java
lsiepel: That is where it should have come from.

lolodomo (approved):
LGTM

lolodomo:
Waiting for a GO to merge.

lsiepel:
> Waiting for a GO to merge.

I also wanted to remove the dependency on org.apache.commons. but i have not yet seen an follow-up to the question in that issue about core/comon string util classes.

lsiepel:
> Waiting for a GO to merge.

Thunderbirds GO :-)

jlaur:
I believe this will conflict with #14202.

lsiepel:
> I believe this will conflict with #14202.

I think it is the other way around. ;-) 
Anyway, there is no real conflict as both PR's make the exact same change to the same two lines. Just merge one and the other might have to adapt. Not sure how git handles such case, let's see.

jlaur:
> I think it is the other way around. ;-)

To be clear, I meant they both conflict with each other. 🙂

> Anyway, there is no real conflict as both PR's make the exact same change to the same two lines. Just merge one and the other might have to adapt. Not sure how git handles such case, let's see.

I didn't compare in that detail, but you are right - if the changes are exactly the same, there should be no conflict. We'll see.

jlaur (approved):
Thanks!

PR #14141, opened 2023-01-02
[ism8] Fix SAT warnings

holgerfriedrich:
Reduce number of warnings, mainly using ```"constant".equals(var)``` instead of ```var.equals("constant")```. This is considered to be better, as "constant".equals(null) is null-safe and always returns false. (Though it should not make a difference, we have the annotation ```NonNullByDefault``` in place for the parameters).

Adding ```final``` makes the tool recognize the byte Arrays as constants (though it is not really a constant, but constant ref to a modifiable buffer).


lolodomo (approved):
LGTM

PR #14154, opened 2023-01-04
Update license headers to 2023

jlaur:
First commit updates these two places:
https://github.com/openhab/openhab-addons/blob/31fd91129eb4382f2c44f1b06b1559123ece1110/pom.xml#L434

https://github.com/openhab/openhab-addons/blob/31fd91129eb4382f2c44f1b06b1559123ece1110/tools/static-code-analysis/checkstyle/ruleset.properties#L2

Second commit is search and replace for "Copyright (c) 2010-2022 Contributors to the openHAB project" since `mvn license:format` does not work correctly.

Previous: #11973 (9893 files)

PR #14156, opened 2023-01-04
[paradoxalarm] Fix namespaces for tests

jlaur:
The used namespaces prevented the license headers from being updated with the command `mvn license:format`.

Related to #14154


jlaur:
Marking as draft to avoid collision with #14154. No longer coupled since `mvn license:format` cannot be used.

lolodomo (approved):
LGTM

Issue #9591, opened 2020-12-30
[chromecast] Things go offline when sending stop command while "no application is running"


andiwirs:
## Expected Behavior
Chromecast Binding should ignore the stop command when no application is running

## Current Behavior
I have my google home minis setup with the Chromecast Binding. When sending a stop command when no application is running the follwing happens:

```
2020-12-30 10:10:40.831 [ome.event.ItemCommandEvent] - Item 'Sw_Sp_Stop_Lr' received command ON
2020-12-30 10:10:40.913 [nt.ItemStatePredictedEvent] - Sw_Sp_Stop_Lr predicted to become ON
2020-12-30 10:10:40.979 [hingStatusInfoChangedEvent] - 'chromecast:audio:GHM_LR' changed from ONLINE to OFFLINE (COMMUNICATION_ERROR): No application is running in ChromeCast
2020-12-30 10:10:42.851 [hingStatusInfoChangedEvent] - 'chromecast:audio:GHM_LR' changed from OFFLINE (COMMUNICATION_ERROR): No application is running in ChromeCast to ONLINE
```

## Context
I have made a group of all Chromecast STOP switches so that i can stop them all for example when i leave the house.
Of course not all Chromecast devices are running at all times, leading to these errors. Writing an extra rule that checks if something is running before sending the STOP is not elegant. IMO the binding should handle that.

## Your Environment
openHAB 2.5.11, on an up-to-date Raspbian, running on a NanoPi M4.

lsiepel:
As this issue is two years old, it might allready been fixed as there are several PR's merged since. Could you verify if this is still happening with 3.4.0? Last week there was also some PR merged to the (yet) unreleased 3.4.1 branch that solved some connection issues.

If this issue is still active, could you provide a debug log from the 3.4.0 version of the binding?

andiwirs:
I am on OH 3.4 now and the problem persists:

`2023-01-02 10:21:57.412 [INFO ] [openhab.event.ItemCommandEvent      ] - Item 'Sw_Sp_Stop_Ki' received command ON
2023-01-02 10:21:57.414 [INFO ] [penhab.event.ItemStatePredictedEvent] - Item 'Sw_Sp_Stop_Ki' predicted to become ON
[...]

PR #14158, opened 2023-01-05
[chromecast] Fix thing go offline after stop command

lsiepel:

When the app running on the ChromeCast was stopped by switching the ON/OFF stop channel, the binding tried to stop the app without checking if it actually runs. Because the app wasn;t running, it gives an error and the thing goes offline.
This PR adds a check, and also restructured the class a bit.

Also when a command was givfen to play media by URL, but the URL was invalid or could not be loaded by the ChromeCast, the thing went offline for ~10 seconds. That mechanism is also change so that the Thing only goes offline when there is an actuall IOException, in other case it logs a warning about unplayable url.

3.4.1 Test JAR is here: https://1drv.ms/u/s!AnMcxmvEeupwjp4ILqWfPxAy7qJh4g?e=43fQde

Fixes: https://github.com/openhab/openhab-addons/issues/9591 (tested and confirmed)

### DIFF
--- a/bundles/org.openhab.binding.chromecast/src/main/java/org/openhab/binding/chromecast/internal/ChromecastCommander.java
+++ b/bundles/org.openhab.binding.chromecast/src/main/java/org/openhab/binding/chromecast/internal/ChromecastCommander.java
@@ -238,7 +238,7 @@ public class ChromecastCommander {
             }
             statusUpdater.updateStatus(ThingStatus.ONLINE);
         } catch (final IOException e) {
-            logger.info("Failed starting app: {}. Message: {}", appId, e.getMessage());
+            logger.warn("Failed starting app: {}. Message: {}", appId, e.getMessage());
         }
     }
 
--- a/bundles/org.openhab.binding.chromecast/src/main/java/org/openhab/binding/chromecast/internal/ChromecastCommander.java
+++ b/bundles/org.openhab.binding.chromecast/src/main/java/org/openhab/binding/chromecast/internal/ChromecastCommander.java
@@ -70,7 +70,7 @@ public class ChromecastCommander {
                 handleControl(command);
                 break;
             case CHANNEL_STOP:
-                handleStop(command);
+                handleCloseApp(command);
                 break;
             case CHANNEL_VOLUME:
                 handleVolume(command);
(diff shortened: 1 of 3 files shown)


bundles/org.openhab.binding.chromecast/src/main/java/org/openhab/binding/chromecast/internal/ChromecastEventReceiver.java
```diff
@@ -56,6 +56,8 @@ public void connectionEventReceived(final @NonNullByDefault({}) ChromeCastConnec
 
     @Override
     public void spontaneousEventReceived(final @NonNullByDefault({}) ChromeCastSpontaneousEvent event) {
+        logger.trace("Received an {} event (class={})", event.getType(), event.getData());
```
lolodomo, line 59: Just a note: I see below a similar log (case for unknow) which uses `event.getType().getDataClass()` and not `event.getData()` for class

bundles/org.openhab.binding.chromecast/src/main/java/org/openhab/binding/chromecast/internal/ChromecastCommander.java
```diff
@@ -229,44 +222,69 @@ private void handleMute(final Command command) {
...
                 }
             } else {
-                logger.warn("Missing media player app - cannot process media.");
+                logger.warn("Failed starting app, app probably not installed. Appid: {}", appId);
             }
             statusUpdater.updateStatus(ThingStatus.ONLINE);
         } catch (final IOException e) {
-            logger.debug("Failed playing media: {}", e.getMessage());
-            statusUpdater.updateStatus(ThingStatus.OFFLINE, COMMUNICATION_ERROR, e.getMessage());
+            logger.info("Failed starting app: {}. Message: {}", appId, e.getMessage());
```
lolodomo, line 241: debug or warn, info is generally not recommended.

bundles/org.openhab.binding.chromecast/src/main/java/org/openhab/binding/chromecast/internal/ChromecastCommander.java
```diff
@@ -229,44 +222,69 @@ private void handleMute(final Command command) {
...
-                    } else {
-                        chromeCast.load(title, null, url, mimeType);
-                    }
+                    logger.debug("Application launched: {}", appId);
                 }
             } else {
-                logger.warn("Missing media player app - cannot process media.");
+                logger.warn("Failed starting app, app probably not installed. Appid: {}", appId);
             }
             statusUpdater.updateStatus(ThingStatus.ONLINE);
```
lolodomo, line 239: Should it be kept there or only in playMedia ?

lsiepel: I extracted the starting of the app out of the playMedia method. This way it would be possible to use the same routine to start different apps. So i prefer to leave it here.

bundles/org.openhab.binding.chromecast/src/main/java/org/openhab/binding/chromecast/internal/ChromecastCommander.java
lsiepel: Should be warn indeed.

bundles/org.openhab.binding.chromecast/src/main/java/org/openhab/binding/chromecast/internal/ChromecastEventReceiver.java
lsiepel: Yes, this is intentionally.

lolodomo (approved):
LGTM

PR #14159, opened 2023-01-05
[chromecast] Add tags and category

lsiepel:

- Added default tags and category (including instructions.xml)

During development, i splitted some changes to [another PR](https://github.com/openhab/openhab-addons/pull/15357)


lolodomo:
Please read #12262

lsiepel:
> Please read #12262

Have adjusted the tags as much as possible to the guidelines, let me know is something is not yet compliant.

lolodomo:
> Have adjusted the tags as much as possible to the guidelines, let me know if something is not yet compliant.

Yes please again the rule number 2: No alone "status" tag is expected on read only channels.
I am not sure even your other added tags are compliant with the "obvious" condition but I am going to see more in details your channels.

lsiepel:
> > Have adjusted the tags as much as possible to the guidelines, let me know if something is not yet compliant.
> 
> Yes please again the rule number 2: No alone "status" tag is expected on read only channels. I am not sure even your other added tags are compliant with the "obvious" condition but I am going to see more in details your channels.

Really sorry about that. Not all my changes where committed. Obvious to one, might not so obvious to the other. It should be an improvement, so if it is not i'll remove the tags.

lsiepel:
Gentle ping @lolodomo do you see some tag issues here left to fix? I'm in the understanind everything is fixed.

lsiepel:
I think it is ready for a review. I tried to add the upgrade instructions, but have some questions as this is a new feature and the documentation might need some iterations.
- Why is the channel type mandatory? As this is not changed. Isn't the channel-id enough?
- The XML definition seems to lack a category element. Is this auto updated  from the thing-types.xml?

jlaur:
> * Why is the channel type mandatory? As this is not changed. Isn't the channel-id enough?
> * The XML definition seems to lack a category element. Is this auto updated  from the thing-types.xml?

I think it has something to do with the data actually stored in the JSON file - you could have a look at openhab/openhab-core#3660

> * [edit] The thing-types.xml accepts DateTime, but the upgrade instructions xml does not. What am i missing @jlaur ?

It needs to be channel type, not item type.

lsiepel:
> > * Why is the channel type mandatory? As this is not changed. Isn't the channel-id enough?
> > * The XML definition seems to lack a category element. Is this auto updated  from the thing-types.xml?
> 
> I think it has something to do with the data actually stored in the JSON file - you could have a look at [openhab/openhab-core#3660](https://github.com/openhab/openhab-core/issues/3660)
> 
> > * [edit] The thing-types.xml accepts DateTime, but the upgrade instructions xml does not. What am i missing @jlaur ?
> 
> It needs to be channel type, not item type.

Thanks for pointing me in the right direction. Could you review?

bundles/org.openhab.binding.chromecast/src/main/resources/OH-INF/upgrade/instructions.xml
```diff
@@ -0,0 +1,125 @@
...
+			</update-channel>
+			<update-channel id="duration">
+				<type>chromecast:duration</type>
+				<tags>
+					<tag>Point</tag>
+					<tag>Duration</tag>
+				</tags>
+			</update-channel>
+			<update-channel id="broadcastDate">
+				<type>Number:broadcastDate</type>
```
jlaur, line 23:
```suggestion
				<type>chromecast:broadcastDate</type>
```

bundles/org.openhab.binding.chromecast/src/main/resources/OH-INF/upgrade/instructions.xml
```diff
@@ -0,0 +1,125 @@
...
```
jlaur, line 103:
```suggestion
				<type>chromecast:broadcastDate</type>
```

jlaur (changes requested):
Thanks!

Review of the update instructions: please see my comments. Also you need to add the `thingTypeVersion` property, otherwise the migration will not be triggered. Please see https://github.com/openhab/openhab-addons/pull/15294#discussion_r1274150623 and test the migration.

I would prefer @lolodomo to have a look at the tags as I never fully got into the rules for those.

You could consider splitting the PR into two:
- **Bug:** Bugfix.
- **Enhancement:** Adding tags/categories.

lsiepel:
Splitted this PR, ready for a review.

jlaur:
> Review of the update instructions: please see my comments. Also you need to add the `thingTypeVersion` property, otherwise the migration will not be triggered. Please see [#15294 (comment)](https://github.com/openhab/openhab-addons/pull/15294#discussion_r1274150623) and test the migration.

Please revisit this.

lsiepel:
> > Review of the update instructions: (...) Also you need to add the `thingTypeVersion` property, otherwise the migration will not be triggered. Please see [#15294 (comment)](https://github.com/openhab/openhab-addons/pull/15294#discussion_r1274150623) and test the migration.
> 
> Please revisit this.

Added the thingtypeversion.

lsiepel:
@lolodomo can you look at the tags? Would be appreciated.

lsiepel:
Gentle ping @openhab/add-ons-maintainers

kaikreuzer:
@lolodomo and @jlaur: I guess you would be the preferred reviewers here.

bundles/org.openhab.binding.chromecast/src/main/resources/OH-INF/thing/thing-types.xml
```diff
@@ -200,13 +214,23 @@
 		<item-type>Number:Time</item-type>
 		<label>Current Time</label>
 		<description>Current time of currently playing media</description>
+		<category>Time</category>
+		<tags>
+			<tag>Status</tag>
+			<tag>Duration</tag>
```
jlaur, line 220: This seems to violate rule 3 in #12262, i.e. no property tag expected with the Status point tag.

bundles/org.openhab.binding.chromecast/src/main/resources/OH-INF/thing/thing-types.xml
```diff
@@ -200,13 +214,23 @@
...
 		<state readOnly="true" pattern="%.0f %unit%"/>
 	</channel-type>
 
 	<channel-type id="duration" advanced="true">
 		<item-type>Number:Time</item-type>
 		<label>Duration</label>
 		<description>Length of currently playing media</description>
+		<category>Time</category>
+		<tags>
+			<tag>Point</tag>
```
jlaur, line 231: I'm not sure if "raw" point tag makes sense in combination with a property tag like below. I guess it might viloate rule 1 by not being absolutely obvious. I would need @lolodomo's advise here.

bundles/org.openhab.binding.chromecast/src/main/resources/OH-INF/thing/thing-types.xml
```diff
@@ -245,6 +269,10 @@
 		<label>Broadcast Date</label>
 		<description>The broadcast date of the currently playing media</description>
 		<category>Calendar</category>
+		<tags>
+			<tag>Point</tag>
```
jlaur, line 273: Same here.

bundles/org.openhab.binding.chromecast/src/main/resources/OH-INF/thing/thing-types.xml
```diff
@@ -260,6 +288,10 @@
 		<label>Creation Date</label>
 		<description>The creation date of the currently playing media</description>
 		<category>Calendar</category>
+		<tags>
+			<tag>Point</tag>
```
jlaur, line 292: Same here.

bundles/org.openhab.binding.chromecast/src/main/resources/OH-INF/thing/thing-types.xml
```diff
@@ -303,6 +335,10 @@
 		<label>Release Date</label>
 		<description>The release date of the currently playing media</description>
 		<category>Calendar</category>
+		<tags>
+			<tag>Point</tag>
```
jlaur, line 339: Same here.

jlaur (changes requested):
I'm not an expert on tag conventions and I'm not able to find any documentation that enable me to verify if these tag combinations make sense. I also don't know what impact adding these tags might have. So I would appreciate if @lolodomo could have a look.

I was previously involved in reviewing the update instructions. Currently they look fine to me, but will of course need an update if changes are required for the tags - and a new test run.

bundles/org.openhab.binding.chromecast/src/main/resources/OH-INF/thing/thing-types.xml
lsiepel: For the example temperature i can see it is not expected. But a song/video has a fixed length and this shows the progress (its status), you don't actively measure the progress. So i thought this was a good fit.

jlaur: That might be the case. I can't assess this myself since I can't find proper documentation and my knowledge about how these tags are used by the semantic model in MainUI is almost non-existant. @lolodomo, perhaps you can help here? 

@lsiepel - maybe let me ask to get this going, when adding these tags, which effect do you see in MainUI, rules or elsewhere, i.e. what is gained from this change?

jlaur: > Are you asking why there are semantic tags at all?

No, I'm asking if you can explain or show a screenshot of the difference that can be seen when having the `Duration` tag compared to when not having it.

PR #14160, opened 2023-01-05
[plugwiseha] Added default tags

lsiepel:

- Added default semantic tags.


bundles/org.openhab.binding.plugwiseha/src/main/resources/OH-INF/thing/channels.xml
```diff
@@ -185,14 +213,22 @@
...
+			<tag>Temperature</tag>
+		</tags>
 		<state readOnly="true" pattern="%.1f %unit%"/>
 	</channel-type>
 
 	<channel-type id="maxBoilerTemperature">
 		<item-type>Number:Temperature</item-type>
 		<label>Max Boiler Temperature</label>
-		<description>Gets the maximum temperature ofthis boiler</description>
+		<description>Gets the maximum temperature of this boiler</description>
```
jlaur, line 226: 👍 Please also update in I18N properties file.

lsiepel: Thanks

jlaur (approved):
Thanks! @lolodomo - I'll let you merge. 🙂

bundles/org.openhab.binding.plugwiseha/src/main/resources/OH-INF/thing/channels.xml
```diff
@@ -129,6 +141,10 @@
 		<label>Intended Boiler Temperature</label>
 		<description>Gets the intended temperature of this boiler</description>
 		<category>heating</category>
+		<tags>
+			<tag>Setpoint</tag>
+			<tag>Temperature</tag>
+		</tags>
 		<state readOnly="true" pattern="%.1f %unit%"/>
```
lolodomo, line 148: While the channel is read only, I am not sure it is appropriate to set a "control" tag as in fact the user cannot control it ?

lsiepel: It is a setpoint that can be controlled from the device, not from openHAB. i.m.o thi sis fine, but as stated in the rules it has to be clear to everyone and reading this post i suppose it is not :-) , so i'll remove it.

I'll remove it for now, but i really hope that the minds behind the semantic tags will come up with some more guidance/rules. As at this moment it leads to some frustration (atleast to me).
I've seen others struggle too.

jlaur:
@lolodomo - gentle ping. :)

lolodomo:
My comment regarding setpoint and read only channel was done on one channel but it concerns several channels. So I suggest to not add setpoint tag to any read only channel in order to be consistent.

bundles/org.openhab.binding.plugwiseha/src/main/resources/OH-INF/thing/channels.xml
lolodomo: You have to do the same thing for other channels, I made my comment on this channel but you added the same tags to other channels which are also read only.

lsiepel: I’ll look at the other channels. It was not clear to me that control can never be set to a read-only channel. To me channel is a control related channel, maybe not writable from oh. 
It is said before, and I know you tried to come up with base rules in the other thread, but there are still so many uncertainties. I really hope core/ac devs come up with more guidance. Currently it is hard for binding developers to support this with a good out of the box experience, it can be seen as hardly any binding had pre-set tags.

lolodomo: > It was not clear to me that control can never be set to a read-only channel.

This is not fully clear for me too.
Look at my comment, it was initially a question.

lsiepel: So how can we be sure? Who to ask? @kaikreuzer  ?

lolodomo: Remains 2 to suppress if we want to be consistent.
You do it one by one :)

lolodomo: Yes, I hope @kaikreuzer could have a general advice for "control" tag for read-only channels.

lsiepel: Removed the one remaining read-only + setpoint tag. i.m.h.o. it could be merged. I added a post to https://github.com/openhab/openhab-addons/issues/12262 i think it would be better to continue the tag discussion there. Thanks for your reply @lolodomo

lolodomo (approved):
LGTM

PR #14168, opened 2023-01-05
[ecobee] Make more channels advanced

### LINKED ISSUE #8029: [ecobee] Reduce ecobee:thermostat thing-type default channels
The `ecobee:thermostat` thing type has too many default channels at the moment, which makes it really heavy and messy for users who use the UI setup. 
I think that only the thermostat-related channels can be kept, while everything else should be switched to an advanced channel.

Also, would it be possible to switch all settings-related channels to the thing configuration? 

## Your Environment

* Version used: (e.g., openHAB and add-on versions)
* Operating System and version (desktop or mobile, Windows 10, Raspbian Buster, ...):

mhilbush:

Resolves #8029 

This binding has a lot of channels. Even though many are advanced, the number of visible channels still can be overwhelming, especially to new users. This PR adds more advanced channels to reduce the clutter of channels that are likely to not be used by most users.

### DIFF
--- a/bundles/org.openhab.binding.ecobee/src/main/resources/OH-INF/thing/thing-types.xml
+++ b/bundles/org.openhab.binding.ecobee/src/main/resources/OH-INF/thing/thing-types.xml
@@ -299,22 +299,22 @@
 		<label>Desired Fan Mode</label>
 		<state readOnly="true" pattern="%s"/>
 	</channel-type>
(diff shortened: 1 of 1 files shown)


lolodomo (approved):
LGTM

mhilbush:
> You can already expect some users asking "where are my channels ?" ;)

Yep. I've already anticipated that. LOL

PR #14169, opened 2023-01-06
[homekit] properly expose services from dummy accessories

ccutrer:
lol I'm surprised they've worked as well as they have in the past. basically only the AccessoryInformationService was getting exposed before, accidentally. now we really do publish the full set of original characteristics, just with dummy values.


yfre (approved):
this is probably your smallest PT ever :)
LGTM

jlaur (approved):
Thanks!

PR #14170, opened 2023-01-06
[ecobee] Fix issue with UTC and local dates

mhilbush:
The Ecobee API returns all date/time fields formatted as "yyyy-MM-dd HH:mm:ss". Most of those date/times are in UTC time, however a couple are in local time. This PR correctly handles both cases by:

- introducing a Gson DateDeserializer that handles the dates when they're in UTC time
- introduces a new method that handles the dates when they're in local time

Reference: https://community.openhab.org/t/ecobee-binding-v2/97733/225

### DIFF
--- a/bundles/org.openhab.binding.ecobee/src/main/java/org/openhab/binding/ecobee/internal/dto/thermostat/DateDeserializer.java
+++ b/bundles/org.openhab.binding.ecobee/src/main/java/org/openhab/binding/ecobee/internal/dto/thermostat/DateDeserializer.java
@@ -1,5 +1,5 @@
 /**
- * Copyright (c) 2010-2022 Contributors to the openHAB project
+ * Copyright (c) 2010-2023 Contributors to the openHAB project
  *
  * See the NOTICE file(s) distributed with this work for additional
  * information.
--- a/bundles/org.openhab.binding.ecobee/src/main/java/org/openhab/binding/ecobee/internal/dto/thermostat/DateDeserializer.java
+++ b/bundles/org.openhab.binding.ecobee/src/main/java/org/openhab/binding/ecobee/internal/dto/thermostat/DateDeserializer.java
@@ -1,0 +1,50 @@
+/**
+ * Copyright (c) 2010-2022 Contributors to the openHAB project
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package org.openhab.binding.ecobee.internal.dto.thermostat;
+
+import java.lang.reflect.Type;
+import java.text.ParseException;
+import java.text.SimpleDateFormat;
+import java.util.Date;
+import java.util.TimeZone;
+
+import org.eclipse.jdt.annotation.NonNullByDefault;
+import org.eclipse.jdt.annotation.Nullable;
+
+import com.google.gson.JsonDeserializationContext;
+import com.google.gson.JsonDeserializer;
+import com.google.gson.JsonElement;
+import com.google.gson.JsonParseException;
+
+/**
+ * The {@link DateDeserializer} is responsible for handling the UTC dates returned from
+ * the Ecobee API service.
+ *
+ * @author Mark Hilbush - Initial contribution
+ */
+@NonNullByDefault
+public class DateDeserializer implements JsonDeserializer<Date> {
+
+    @Override
+    public @Nullable Date deserialize(JsonElement element, Type arg1, JsonDeserializationContext arg2)
+            throws JsonParseException {
+        String date = element.getAsString();
+        SimpleDateFormat formatter = new SimpleDateFormat("yyyy-MM-dd HH:mm:ss");
+        formatter.setTimeZone(TimeZone.getTimeZone("UTC"));
+        try {
+            return formatter.parse(date);
+        } catch (ParseException e) {
+            return null;
+        }
+    }
+}
--- a/bundles/org.openhab.binding.ecobee/src/main/java/org/openhab/binding/ecobee/internal/handler/EcobeeUtils.java
+++ b/bundles/org.openhab.binding.ecobee/src/main/java/org/openhab/binding/ecobee/internal/handler/EcobeeUtils.java
@@ -12,7 +12,9 @@
  */
 package org.openhab.binding.ecobee.internal.handler;
 
+import java.time.DateTimeException;
 import java.time.ZonedDateTime;
+import java.time.format.DateTimeFormatter;
 import java.util.Date;
 
 import javax.measure.Unit;
@@ -85,6 +87,20 @@ public final class EcobeeUtils {
 
     public static State undefOrPoint(@Nullable String value) {
         return value == null ? UnDefType.UNDEF : new PointType(value);
+    }
+
+    public static State undefOrLocalDate(@Nullable String date, TimeZoneProvider timeZoneProvider) {
+        if (date == null) {
+            return UnDefType.UNDEF;
+        }
+        try {
+            DateTimeFormatter formatter = DateTimeFormatter.ofPattern("yyyy-MM-dd HH:mm:ss")
+                    .withZone(timeZoneProvider.getTimeZone());
+            ZonedDateTime zoned = ZonedDateTime.parse(date, formatter);
+            return new DateTimeType(zoned);
+        } catch (DateTimeException e) {
+            return UnDefType.UNDEF;
+        }
     }
 
     public static State undefOrDate(@Nullable Date date, TimeZoneProvider timeZoneProvider) {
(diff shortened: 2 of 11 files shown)


bundles/org.openhab.binding.ecobee/src/main/java/org/openhab/binding/ecobee/internal/dto/thermostat/DateDeserializer.java
```diff
@@ -0,0 +1,50 @@
...
+        String date = element.getAsString();
+        SimpleDateFormat formatter = new SimpleDateFormat("yyyy-MM-dd HH:mm:ss");
+        formatter.setTimeZone(TimeZone.getTimeZone("UTC"));
+        try {
+            return formatter.parse(date);
+        } catch (ParseException e) {
+            return null;
+        }
+    }
+}
```
jlaur, line 50: Would you consider the deserializing directly to the intended target type, e.g. `ZonedDateType` instead? See for example:
https://github.com/openhab/openhab-addons/blob/d5dd3c9ea3a2ca609707d7e9b3b51bebf9a48432/bundles/org.openhab.binding.netatmo/src/main/java/org/openhab/binding/netatmo/internal/deserialization/NADeserializer.java#L45-L57

Also, can you explain how this deserializer is used, since the DTO has changed from `Date` to `String`? I'm a little confused by this deserializer and the new version of `undefOrLocalDate` taking a String.

mhilbush: > Would you consider the deserializing directly to the intended target type, e.g. ZonedDateType instead?

I would, but I'm not sure how to do it in this case.

The only changes in the DTO are for two date/time elements representing local time of the form "yyyy-MM-dd HH:mm:ss"

I left the Date elements intact for the remaining date/time elements representing UTC time also of the form "yyyy-MM-dd HH:mm:ss"

The problem I was having was that Ecobee uses the same date/time format (i.e. "yyyy-MM-dd HH:mm:ss") to represent local time and UTC time. Bad design on their part, since the only way to know if a date field is UTC or local time is to look at the Ecobee API documentation. 😢 

So I had to find a way to differentiate how the dates were handled. My solution was to use the Date deserializer for the dates that are in UTC time (the majority), which would force it to be UTC time by explicitly setting the timezone to UTC in the deserializer. In the case of the dates in local time (of which there are only 2), I chose to make them Strings, and handle them explicitly using the method `undefOrLocalDate`.

Hopefully this makes sense. If there's a better way to do it, let's discuss.

jlaur: Thanks for the explanation. Interesting, and not quite straight-forward, it seems. Without going more into the code right now, all I can say is that it would be nice if you could skip conversion to `Date` as intermediate type, since this is an old Java 8 type. Instead you could use `ZonedDateTime` when you need the time-zone or `Instant` when you don't. However, with your explanation, this seems to require some additional refactoring which is probably outside of the scope of this PR. So I will leave this comment and wait for your response, and I'd be happy to merge as is.

mhilbush: Thanks.

It's not urgent to merge this, and I don't think there are that many places where Date is used. So, I'll take a look at changing it, so long as I can still accomplish what I need to (which I see no reason why I can't) and that it doesn't require too many additional changes.

jlaur: I forgot to mention `LocalDateTime`, which may also be of use. For example, DTO's could return either `LocalDateTime` or `Instant`, depending on source format. Applying time-zone from `TimeZoneProvider` could then be handled later and quite in the same fashion.

mhilbush: Does GSON know how to deal with a DTO element defined as an `Instant`? From what I can see, I might need a custom deserializer if I want to use `Instant` in the DTOs for my UTC dates.

jlaur: No, that would require a custom deserializer.

mhilbush: Yep, Instant and LocalDateTime both require custom deserializers. If you would please have a look at my changes. In the Instant deserializer, I had to first convert to LocalDateTime because I was getting exceptions when trying to parse straight to an Instant. I think that was because the format didn't conform what the parser required. This approach required a few more changes, but I also think it's cleaner.

bundles/org.openhab.binding.ecobee/src/main/java/org/openhab/binding/ecobee/internal/handler/EcobeeUtils.java
```diff
@@ -87,9 +89,12 @@ public static State undefOrPoint(@Nullable String value) {
         return value == null ? UnDefType.UNDEF : new PointType(value);
     }
 
-    public static State undefOrDate(@Nullable Date date, TimeZoneProvider timeZoneProvider) {
-        return date == null ? UnDefType.UNDEF
-                : new DateTimeType(ZonedDateTime.ofInstant(date.toInstant(), timeZoneProvider.getTimeZone()));
+    public static State undefOrDate(@Nullable Instant instant) {
+        return instant == null ? UnDefType.UNDEF : new DateTimeType(ZonedDateTime.ofInstant(instant, ZoneId.of("UTC")));
```
jlaur, line 93: This overload should also use `TimeZoneProvider`, i.e.:
```suggestion
    public static State undefOrDate(@Nullable Instant instant, TimeZoneProvider timeZoneProvider) {
        return instant == null ? UnDefType.UNDEF : new DateTimeType(instant.atZone(timeZoneProvider.getTimeZone()));
```

This is an unfortunate side-effect of the way `DateTimeType` currently works - see openhab/openhab-core#2898. Without providing the time-zone, raw formatting will be in UTC.

mhilbush: Now I'm confused. Don't I want the time zone to be UTC, because that's what it is in the json returned from Ecobee. That way when it gets displayed in openHAB it will be shown in the user's local time.

mhilbush: Here is `lastModified` in the console.
```
E1_Info_LastModified (Type=DateTimeItem, State=2023-01-08T12:25:28.000+0000, Label=Last Modified, Category=time, Groups=[gMain_Info])
```
I made a change to the thermostat that caused this item to change at 7:25 AM US EST (UTC-5). So I believe it is correct.

jlaur: I'm sorry for the confusion, and indeed it is kind of a mess until we get that `DateTimeType` issue sorted out. No matter if you provide it in UTC or local time-zone, it will be the same exact moment in time, and with the proper state description, it will be shown as correct local datetime in both cases.

The difference occurs when you omit the state description. In this case the raw `ZonedDateTime` will be shown, and this will of course be different. Here it's preferable to use the configured time-zone, not UTC, as the user will expect to see local time. When the `ZonedDateTime` is constructed with the local time-zone, the raw formatting will also show the time in local time-zone.

So in other words, I'm simply requesting a conversion from UTC to local time-zone for the provided `ZonedDateTime`. In essence it will then be exactly the same as the result of the `undefOrDate` `LocalDateTime` overload. Of course you could argue that all DTO's could have used `Instant` then, but your code looks very clean to me, and you also have the advantage of dealing with the UTC vs. local time logic in your handler rather than deep inside some DTO or deserializer.

mhilbush: Yes, I see now. I was looking at the date in the UI formatted with a state description, which was correct. I see now that if I looked at the raw date without the state description it would be in UTC time.

Thanks for that explanation.

mhilbush: Ok, I think I might finally have it. 🤞

jlaur: It looks good!

bundles/org.openhab.binding.ecobee/src/main/java/org/openhab/binding/ecobee/internal/dto/thermostat/DateDeserializer.java
jlaur: I agree, it's nice and clean - well done!

jlaur (approved):
LGTM - thanks for the fix and cleanup!

mhilbush:
Thanks. I appreciate your help getting this over the finish line!

PR #14178, opened 2023-01-07
Update MapDB and use mmap on 64-bit platforms

Flole998:
This updates MapDB to the latest version. Doing so will break compatibility with the previous version and it will not be possible to load previous databases. A new major version number of openHAB seems like a good idea to do this.

Using mmap will increase performance on supported platforms. We might suffer from the file descriptor leak from http://bugs.java.com/view_bug.do?bug_id=4724038 but as we don't open/close the file often it shouldn't be an issue.

### DIFF
--- a/bundles/org.openhab.persistence.mapdb/pom.xml
+++ b/bundles/org.openhab.persistence.mapdb/pom.xml
@@ -15,14 +15,66 @@
   <name>openHAB Add-ons :: Bundles :: Persistence Service :: MapDB</name>
 
   <properties>
-    <bnd.importpackage>org.openhab.core.library.types</bnd.importpackage>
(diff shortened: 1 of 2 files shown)


Flole998:
The failing build is most likely due to the updated dependency not beeing found yet.

Flole998:
@wborn Maybe you know what's wrong here and can make it build sucessfully.

wborn:
You need to find a way to provide all the Kotlin dependencies. I never got code depending on Kotlin working with OSGi.

Flole998:
Thanks! It's not telling me what's missing though, right?

wborn:
Yes you have to figure out what Maven dependencies to embed into the bundle. From my experience there will be many to be able to use Kotlin so it will significantly increase the bundle size.

Flole998:
It seems like I got (most of) the dependency chaos cleaned up now. However, the tests are failing for some reason. I believe something is not quite right there yet.

I'm not really sure why it doesn't pick up the dependencies of the mapdb library and I need to manually add those.

lsiepel:
@Flole998 Are you able to proceed with this?

Flole998:
@lsiepel No, not really. Feel free to close it if you don't want to work on it either.

lsiepel:
Maybe this can be picked up again in the future. Closing per [author comment. ](https://github.com/openhab/openhab-addons/pull/14178#issuecomment-1937110626)

PR #14191, opened 2023-01-09
[chromecast] Fix for NPE when playerState is null

### LINKED ISSUE #14190: [chromecast] Fix for NPE when playerState is null
My NVIDIA ShieldTV returns a `playerState` of null in the MEDIA_STATUS message, which results in the following NPE.
```
2023-01-09 10:07:04.305 [DEBUG] [ast.internal.ChromecastStatusUpdater] - MEDIA_STATUS MediaStatus{activeTrackIds: <null>, mediaSessionId: 7, playbackRate: 0, playerState: null, currentItemId: 1, currentTime: 0.000000, customData: <null>, loadingItemId: null, items: <null>, preloadedItemId: null, supportedMediaCommands: 0, volume: null, media: null, repeatMode: REPEAT_OFF, idleReason: null}
2023-01-09 10:07:04.305 [WARN ] [mmon.WrappedScheduledExecutorService] - Scheduled runnable ended with an exception:
java.lang.NullPointerException: Cannot invoke "su.litvak.chromecast.api.v2.MediaStatus$PlayerState.ordinal()" because "mediaStatus.playerState" is null
	at org.openhab.binding.chromecast.internal.ChromecastStatusUpdater.updateMediaStatus(ChromecastStatusUpdater.java:160) ~[?:?]
	at org.openhab.binding.chromecast.internal.ChromecastCommander.handleRefresh(ChromecastCommander.java:115) ~[?:?]
	at org.openhab.binding.chromecast.internal.handler.ChromecastHandler$Coordinator.refresh(ChromecastHandler.java:324) ~[?:?]
	at java.util.concurrent.FutureTask.runAndReset(FutureTask.java:305) ~[?:?]
	at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:305) ~[?:?]
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1136) [?:?]
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:635) [?:?]
	at java.lang.Thread.run(Thread.java:833) [?:?]
```
I will submit a fix.

mhilbush:
Fixes #14190 

### DIFF
--- a/bundles/org.openhab.binding.chromecast/src/main/java/org/openhab/binding/chromecast/internal/ChromecastStatusUpdater.java
+++ b/bundles/org.openhab.binding.chromecast/src/main/java/org/openhab/binding/chromecast/internal/ChromecastStatusUpdater.java
@@ -157,30 +157,32 @@ public class ChromecastStatusUpdater {
             return;
         }
 
-        switch (mediaStatus.playerState) {
-            case IDLE:
-                break;
-            case PAUSED:
-                callback.updateState(CHANNEL_CONTROL, PlayPauseType.PAUSE);
-                callback.updateState(CHANNEL_STOP, OnOffType.OFF);
-                break;
-            case BUFFERING:
-            case LOADING:
-            case PLAYING:
-                callback.updateState(CHANNEL_CONTROL, PlayPauseType.PLAY);
-                callback.updateState(CHANNEL_STOP, OnOffType.OFF);
-                break;
-            default:
-                logger.debug("Unknown media status: {}", mediaStatus.playerState);
-                break;
+        if (mediaStatus.playerState != null) {
+            switch (mediaStatus.playerState) {
+                case IDLE:
+                    break;
+                case PAUSED:
+                    callback.updateState(CHANNEL_CONTROL, PlayPauseType.PAUSE);
+                    callback.updateState(CHANNEL_STOP, OnOffType.OFF);
+                    break;
+                case BUFFERING:
+                case LOADING:
+                case PLAYING:
+                    callback.updateState(CHANNEL_CONTROL, PlayPauseType.PLAY);
+                    callback.updateState(CHANNEL_STOP, OnOffType.OFF);
+                    break;
+                default:
+                    logger.debug("Unknown media status: {}", mediaStatus.playerState);
+                    break;
+            }
         }
 
         callback.updateState(CHANNEL_CURRENT_TIME, new QuantityType<>(mediaStatus.currentTime, Units.SECOND));
 
         // If we're playing, paused or buffering but don't have any MEDIA information don't null everything out.
         Media media = mediaStatus.media;
-        if (media == null && (mediaStatus.playerState == PLAYING || mediaStatus.playerState == PAUSED
-                || mediaStatus.playerState == BUFFERING)) {
+        if (media == null && (mediaStatus.playerState == null || mediaStatus.playerState == PLAYING
+                || mediaStatus.playerState == PAUSED || mediaStatus.playerState == BUFFERING)) {
             return;
         }
 


bundles/org.openhab.binding.chromecast/src/main/java/org/openhab/binding/chromecast/internal/ChromecastStatusUpdater.java
```diff
@@ -157,30 +157,32 @@ public void updateMediaStatus(final @Nullable MediaStatus mediaStatus) {
...
-            case BUFFERING:
-            case LOADING:
-            case PLAYING:
-                callback.updateState(CHANNEL_CONTROL, PlayPauseType.PLAY);
-                callback.updateState(CHANNEL_STOP, OnOffType.OFF);
-                break;
-            default:
-                logger.debug("Unknown media status: {}", mediaStatus.playerState);
-                break;
+        if (mediaStatus.playerState != null) {
```
jlaur, line 160: Do we want to log if it's null, similar to line 175 which logs only if new entries are added to:
https://github.com/vitalidze/chromecast-java-api-v2/blob/c1358030b1b6afa11dc2a83a0049783e7b5bef7e/src/main/java/su/litvak/chromecast/api/v2/MediaStatus.java#L41

Of course only if relevant, I don't know if this is a normal situation, or caused by some error/weird state?

jlaur (approved):
Thank you. Just one question about logging.

bundles/org.openhab.binding.chromecast/src/main/java/org/openhab/binding/chromecast/internal/ChromecastStatusUpdater.java
mhilbush: I really don't know if it's a normal situation or not. The Google docs define `playerState` as non-null, so perhaps it should be considered a bug in the NVIDIA implementation?

https://developers.google.com/cast/docs/reference/web_receiver/cast.framework.messages.MediaStatus#playerState

I was hesitant to log it, as I already see a bunch of these constantly in my log file, and I didn't want to add to the noise. I haven't yet tracked down which of my chromecast devices are causing that message. But I don't think it's the NVIDIA device.
```
2023-01-10 08:54:58.285 [DEBUG] [inding.chromecast.internal.ChromecastStatusUpdater] - MEDIA_STATUS null
```

jlaur: Let's treat it as a normal happening situation then, and refrain from logging it.

jlaur:
This should be considered for a 3.4.x backport.

jlaur:
@openhab/add-ons-maintainers - do you agree to include this in next 3.4 hotfix?

lolodomo:
Just check that this is not embedding code from other PRs. IIRC there were several recent PRs merged for this binding.

scatterthought:
I upgraded from 3.3 to 3.4.2 this past weekend, and I've lost the ability to stop casting by sending ON commands to the STOP channel. I manually installed the 3.4.1 binding and immediately regained STOP control, so I think it may be an unintended consequence of the playerState fix.

Looking at the commit, I think there's an extra } on Line 177. Could that be the issue?

lsiepel:
> I upgraded from 3.3 to 3.4.2 this past weekend, and I've lost the ability to stop casting by sending ON commands to the STOP channel. I manually installed the 3.4.1 binding and immediately regained STOP control, so I think it may be an unintended consequence of the playerState fix.
> 
> Looking at the commit, I think there's an extra } on Line 177. Could that be the issue?

Please open an issue so we can teoubleshoot. It would also help if you can share a debug log, device, version etc.

scatterthought:
@lsiepel I've opened an issue, and @mhilbush pointed out that it's more likely due to #14158, which didn't show up in the 3.4.2 release notes.

PR #14192, opened 2023-01-09
[miio] Fix logging in test folder

marcelrv:
Brings back the log messages for o.a. ReadmeHelper that disappeared due to https://github.com/openhab/openhab-core/pull/2982

### DIFF
--- a/bundles/org.openhab.binding.miio/src/main/resources/logback-test.xml
+++ b/bundles/org.openhab.binding.miio/src/main/resources/logback-test.xml
@@ -1,0 +1,75 @@
+<configuration>
+
+	<appender name="STDOUT" class="ch.qos.logback.core.ConsoleAppender">
+		<encoder>
+			<pattern>
+				%d{HH:mm:ss.SSS} [%-5.5p] [%-36.36c] - %m%n
+			</pattern>
+		</encoder>
+	</appender>
+
+	<root level="ERROR">
+		<appender-ref ref="STDOUT"/>
+	</root>
+
+	<!-- openHAB specific logger configuration -->
+
+	<Logger level="${env:logger.openhab.level:-INFO}" name="org.openhab"/>
+
+	<!-- Uncomment this logger and update the name to enable debug logging for a binding -->
+	<!-- <Logger level="DEBUG" name="org.openhab.binding.YOURBINDINGNAME"/> -->
+
+	<Logger level="DEBUG" name="org.openhab.binding.miio"/>
+
+	<!-- You can also override the variables of the loggers below to enable logging without -->
+	<!-- modifying this file. E.g. set 'logger.misc1.name' to 'org.openhab.binding.YOURBINDINGNAME' -->
+	<!-- in the launch configuration of app.bndrun . Similarly you can also override the -->
+	<!--'logger.misc1.level' this way or change the log level of the root and openhab loggers above. -->
+	<Logger level="${env:logger.misc1.level:-DEBUG}" name="${env:logger.misc1.name:-override.me}"/>
+	<Logger level="${env:logger.misc2.level:-DEBUG}" name="${env:logger.misc2.name:-override.me}"/>
+	<Logger level="${env:logger.misc3.level:-DEBUG}" name="${env:logger.misc3.name:-override.me}"/>
+
+	<Logger level="ERROR" name="openhab.event.ItemStateEvent"/>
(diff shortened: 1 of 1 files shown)


lolodomo:
@J-N-K : as you changed the code in the core framework, could you please tell us if these changes look good to you ?

J-N-K:
I don't fully understand what log is really required here. If you only want to see log messages from the binding, the logger configuration should set default logging to error and only the binding logger to INFO or DEBUG. But keep in mind that the logging is reduced during build because it increases the size of the CI build output.

marcelrv:
@J-N-K that is how it worked before. But that stopped since the mentioned PR.
Since then it works like that for the normal binding, but not for the test folder.

or maybe you can advice how to set the logging for a binding for the files from the 'src/test' folder.

background info:
To generate the readme for the miio binding there is
https://github.com/openhab/openhab-addons/blob/main/bundles/org.openhab.binding.miio/src/test/java/org/openhab/binding/miio/internal/ReadmeHelper.java
Since the change, the output is no longer visible even if the logging for the binding is set to DEBUG. Only warn messages are visible.

marcelrv:
@J-N-K this issue is still there.

PR https://github.com/openhab/openhab-core/pull/2982  suppresses the info level, only showing the warn levels with the instroduction of [logback-test.xml](https://github.com/openhab/openhab-core/pull/2982/files#diff-c7125b103880d4834def96dca5c03c3b1f7111bcb6238bfec801ee9756b0a985)
.

to reproduce the issue in eclipse run (right click : run as java)

https://github.com/openhab/openhab-addons/blob/main/bundles/org.openhab.binding.miio/src/test/java/org/openhab/binding/miio/internal/ReadmeHelper.java

since the PR the output is typically empty, as for all the messages with level info nothing is shown. 
With the introduction of this file I can see the messages.
I think this is specifically due to the fact it is in the src/test folder instead of the regular src folders.

the desired situation is that it would follow whatever there is set for the binding in the normal logging setting.

lsiepel:
Gentle ping @J-N-K and @marcelrv

marcelrv:
@J-N-K would appreciate if you could feedback/comment if my clarification was clear and/or what is the concern

bundles/org.openhab.binding.miio/src/main/resources/logback-test.xml
```diff
@@ -0,0 +1,75 @@
...
+	<!-- Filters warnings for events that could not be delivered to a disconnected client. -->
+	<Logger level="ERROR" name="org.apache.cxf.jaxrs.sse.SseEventSinkImpl"/>
+
+	<!-- Filters known issues of javax.mail, see -->
+	<!-- https://github.com/openhab/openhab-addons/issues/5530 -->
+	<Logger level="ERROR" name="javax.mail"/>
+
+	<!-- Filters disconnection warnings of the ChromeCast Java API, see -->
+	<!-- https://github.com/openhab/openhab-addons/issues/3770 -->
+	<Logger level="ERROR" name="su.litvak.chromecast.api.v2.Channel"/>
```
lsiepel, line 73: This (and maybe more) seems beyond the scope of this miio binding.

lsiepel:
@openhab/add-ons-maintainers should we close or merge this PR. I’m not sure what to do here.

jlaur:
@marcelrv - can you set the log level explicitly like here?

https://github.com/openhab/openhab-addons/blob/778f9f61fc84a075cce485fcec38a12010464f98/bundles/org.openhab.binding.bluetooth.grundfosalpha/src/test/java/org/openhab/binding/bluetooth/grundfosalpha/internal/protocol/ResponseMessageTest.java#L27-L41

marcelrv:
> @marcelrv - can you set the log level explicitly like here?
> 
> https://github.com/openhab/openhab-addons/blob/778f9f61fc84a075cce485fcec38a12010464f98/bundles/org.openhab.binding.bluetooth.grundfosalpha/src/test/java/org/openhab/binding/bluetooth/grundfosalpha/internal/protocol/ResponseMessageTest.java#L27-L41

will give that a try and feedback

lsiepel:
@marcelrv you got any results?

lsiepel:
Sorry, but there does not seem to be any progress on this PR. It cannot be merged in its current state. Feel free to re-open if the essential feedback is gathered or it made progress in any other way.

PR #14193, opened 2023-01-09
[miio] Update device names list

marcelrv:
Update the list with latest device names


marcelrv:
Is there any concern about this PR that is blocking the merging.
There is no code that is impacted, this file is only used by the developer support tools

fwolter (approved):

Issue #14195, opened 2023-01-09
[boschshc] Discovery of Bridge and Things


GerdZanker:
Discover the Bosch SmartHomeController Bridge and its Things viw the openhab discovery API.

With mDNS possible IP addresses can be retrieved. 
The SHC returns public information for the bridge without pairing. 
After pairing all the "things" can be auto discovered, so users don't have to dig for the device IDs inside the openHAB logs.

GerdZanker:
Initial Discovery code is now available in related draft pull Request.
Attached you find a Jar to tests with latest stable openhab version - feedback is very welcome.
[org.openhab.binding.boschshc-DISCOVERY-3.4.2.jar.zip](https://github.com/openhab/openhab-addons/files/10796742/org.openhab.binding.boschshc-DISCOVERY-3.4.2.jar.zip)

PR #14197, opened 2023-01-09
[boschshc] Bridge and Device Discovery

GerdZanker:
#14195 [boschshc] Bridge discovery is implemented via mDNS, local IP addresses are checked. If a GET returns the public SHC information,
then this shcIpAddress is reported as a discovered bridge.

Devices are always discovered after successful pairing, but a manual scan is also possible.

### DIFF
--- a/bundles/org.openhab.binding.boschshc/README.md
+++ b/bundles/org.openhab.binding.boschshc/README.md
@@ -206,12 +206,15 @@ The smoke detector warns you in case of fire.
 
 ## Limitations
 
-- Discovery of Things
-- Discovery of Bridge
+No major limitation known.
+Check list of [openhab issues with "boshshc"](https://github.com/openhab/openhab-addons/issues?q=is%3Aissue+boschshc+) 
 
 ## Discovery
 
-Configuration via configuration files or UI (see below).
+Bridge discovery is supported via mDNS.
+Things discovery is started after successful pairing.
+
+Configuration via configuration files or UI supported too (see below).
 
 ## Bridge Configuration
 
@@ -239,19 +242,10 @@ Alternatively, the log can be viewed using the OpenHab Log Viewer (frontail) via
 Example:
 
 ```bash
-2020-08-11 12:42:49.490 [INFO ] [chshc.internal.BoschSHCBridgeHandler] - Found device: name=Heizung id=hdm:HomeMaticIP:3014F711A000XXXXXXXXXXXX
-2020-08-11 12:42:49.495 [INFO ] [chshc.internal.BoschSHCBridgeHandler] - Found device: name=-RoomClimateControl- id=roomClimateControl_hz_1
-2020-08-11 12:42:49.497 [INFO ] [chshc.internal.BoschSHCBridgeHandler] - Found device: name=-VentilationService- id=ventilationService
-2020-08-11 12:42:49.498 [INFO ] [chshc.internal.BoschSHCBridgeHandler] - Found device: name=Großes Fenster id=hdm:HomeMaticIP:3014F711A000XXXXXXXXXXXX
-2020-08-11 12:42:49.501 [INFO ] [chshc.internal.BoschSHCBridgeHandler] - Found device: name=-IntrusionDetectionSystem- id=intrusionDetectionSystem
-2020-08-11 12:42:49.502 [INFO ] [chshc.internal.BoschSHCBridgeHandler] - Found device: name=Rollladen id=hdm:HomeMaticIP:3014F711A000XXXXXXXXXXXX
-2020-08-11 12:42:49.502 [INFO ] [chshc.internal.BoschSHCBridgeHandler] - Found device: name=Heizung id=hdm:HomeMaticIP:3014F711A000XXXXXXXXXXXX
-2020-08-11 12:42:49.503 [INFO ] [chshc.internal.BoschSHCBridgeHandler] - Found device: name=Heizung Haus id=hdm:ICom:819410185:HC1
-2020-08-11 12:42:49.503 [INFO ] [chshc.internal.BoschSHCBridgeHandler] - Found device: name=-RoomClimateControl- id=roomClimateControl_hz_6
-2020-08-11 12:42:49.504 [INFO ] [chshc.internal.BoschSHCBridgeHandler] - Found device: name=PhilipsHueBridgeManager id=hdm:PhilipsHueBridge:PhilipsHueBridgeManager
-2020-08-11 12:42:49.505 [INFO ] [chshc.internal.BoschSHCBridgeHandler] - Found device: name=Rollladen id=hdm:HomeMaticIP:3014F711A000XXXXXXXXXXXX
-2020-08-11 12:42:49.506 [INFO ] [chshc.internal.BoschSHCBridgeHandler] - Found device: name=Rollladen id=hdm:HomeMaticIP:3014F711A000XXXXXXXXXXXX
-2020-08-11 12:42:49.507 [INFO ] [chshc.internal.BoschSHCBridgeHandler] - Found device: name=Central Heating id=hdm:ICom:819410185
+2023-03-20 20:30:48.026 [INFO ] [g.discovery.internal.PersistentInbox] - Added new thing 'boschshc:security-camera-eyes:yourBridgeName:hdm_Cameras_XXXXXXXX-XXXX-XXXX-XXXX-XXXXXXXXXXXX' to inbox.
+2023-03-20 20:30:48.026 [INFO ] [g.discovery.internal.PersistentInbox] - Added new thing 'boschshc:smoke-detector:yourBridgeName:hdm_HomeMaticIP_XXXXXXXXXXXXXXXXXXXXXXXX' to inbox.
+2023-03-20 20:30:48.027 [INFO ] [g.discovery.internal.PersistentInbox] - Added new thing 'boschshc:twinguard:yourBridgeName:hdm_ZigBee_XXXXXXXXXXXXXXXX' to inbox.
+2023-03-20 20:30:48.028 [INFO ] [g.discovery.internal.PersistentInbox] - Added new thing 'boschshc:smart-bulb:yourBridgeName:hdm_PhilipsHueBridge_HueLight_XXXXXXXXXXXXXXXX-XX_XXXXXXXXXXXX' to inbox.
 ```
 
 ## Thing Configuration
--- a/bundles/org.openhab.binding.boschshc/src/main/java/org/openhab/binding/boschshc/internal/devices/bridge/BridgeHandler.java
+++ b/bundles/org.openhab.binding.boschshc/src/main/java/org/openhab/binding/boschshc/internal/devices/bridge/BridgeHandler.java
@@ -16,6 +16,11 @@ import static org.eclipse.jetty.http.HttpMethod.*;
 
 import java.lang.reflect.Type;
 import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.List;
+import java.util.Objects;
+import java.util.Optional;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.TimeUnit;
@@ -33,6 +38,7 @@ import org.openhab.binding.boschshc.internal.devices.bridge.dto.Device;
 import org.openhab.binding.boschshc.internal.devices.bridge.dto.DeviceServiceData;
 import org.openhab.binding.boschshc.internal.devices.bridge.dto.LongPollResult;
 import org.openhab.binding.boschshc.internal.devices.bridge.dto.Room;
+import org.openhab.binding.boschshc.internal.discovery.ThingDiscoveryService;
 import org.openhab.binding.boschshc.internal.exceptions.BoschSHCException;
 import org.openhab.binding.boschshc.internal.exceptions.LongPollingFailedException;
 import org.openhab.binding.boschshc.internal.exceptions.PairingFailedException;
@@ -45,6 +51,7 @@ import org.openhab.core.thing.ThingStatus;
 import org.openhab.core.thing.ThingStatusDetail;
 import org.openhab.core.thing.binding.BaseBridgeHandler;
 import org.openhab.core.thing.binding.ThingHandler;
+import org.openhab.core.thing.binding.ThingHandlerService;
 import org.openhab.core.types.Command;
 import org.osgi.framework.Bundle;
 import org.osgi.framework.FrameworkUtil;
@@ -62,6 +69,7 @@ import com.google.gson.reflect.TypeToken;
  * @author Gerd Zanker - added HttpClient with pairing support
  * @author Christian Oeing - refactorings of e.g. server registration
  * @author David Pace - Added support for custom endpoints and HTTP POST requests
+ * @author Gerd Zanker - added thing discovery
  */
 @NonNullByDefault
 public class BridgeHandler extends BaseBridgeHandler {
@@ -88,10 +96,22 @@ public class BridgeHandler extends BaseBridgeHandler {
 
     private @Nullable ScheduledFuture<?> scheduledPairing;
 
+    /**
+     * SHC thing/device discovery service instance.
+     * Registered and unregistered if service is actived/deactived.
+     * Used to scan for things after bridge is paired with SHC.
+     */
+    private @Nullable ThingDiscoveryService thingDiscoveryService;
+
     public BridgeHandler(Bridge bridge) {
         super(bridge);
 
         this.longPolling = new LongPolling(this.scheduler, this::handleLongPollResult, this::handleLongPollFailure);
+    }
+
+    @Override
+    public Collection<Class<? extends ThingHandlerService>> getServices() {
+        return Collections.singleton(ThingDiscoveryService.class);
     }
 
     @Override
@@ -225,17 +245,19 @@ public class BridgeHandler extends BaseBridgeHandler {
                 return;
             }
 
-            // SHC is online and access is possible
-            // print rooms and devices
-            boolean thingReachable = true;
-            thingReachable &= this.getRooms();
-            thingReachable &= this.getDevices();
-            if (!thingReachable) {
+            // SHC is online and access should possible
+            if (!checkBridgeAccess()) {
                 this.updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.OFFLINE.COMMUNICATION_ERROR,
                         "@text/offline.not-reachable");
                 // restart initial access
                 scheduleInitialAccess(httpClient);
                 return;
+            }
+
+            // do thing discovery after pairing
+            final ThingDiscoveryService discovery = thingDiscoveryService;
+            if (discovery != null) {
+                discovery.doScan();
             }
 
             // start long polling loop
@@ -253,53 +275,129 @@ public class BridgeHandler extends BaseBridgeHandler {
     }
 
     /**
-     * Get a list of connected devices from the Smart-Home Controller
-     *
-     * @throws InterruptedException in case bridge is stopped
+     * Check the bridge access by sending an HTTP request.
+     * Does not throw any exception in case the request fails.
      */
-    private boolean getDevices() throws InterruptedException {
+    public boolean checkBridgeAccess() throws InterruptedException {
         @Nullable
         BoschHttpClient httpClient = this.httpClient;
+
         if (httpClient == null) {
             return false;
         }
 
         try {
-            logger.debug("Sending http request to Bosch to request devices: {}", httpClient);
+            logger.debug("Sending http request to BoschSHC to check access: {}", httpClient);
+            String url = httpClient.getBoschSmartHomeUrl("devices");
+            ContentResponse contentResponse = httpClient.createRequest(url, GET).send();
+
+            // check HTTP status code
+            if (!HttpStatus.getCode(contentResponse.getStatus()).isSuccess()) {
+                logger.debug("Access check failed with status code: {}", contentResponse.getStatus());
+                return false;
+            }
+
+            // Access OK
+            return true;
+        } catch (TimeoutException | ExecutionException e) {
+            logger.warn("Access check failed because of {}!", e.getMessage());
+            return false;
+        }
+    }
+
+    /**
+     * Get a list of connected devices from the Smart-Home Controller
+     *
+     * @throws InterruptedException in case bridge is stopped
+     */
+    public List<Device> getDevices() throws InterruptedException {
+        @Nullable
+        BoschHttpClient httpClient = this.httpClient;
+        if (httpClient == null) {
+            return Collections.emptyList();
+        }
+
+        try {
+            logger.trace("Sending http request to Bosch to request devices: {}", httpClient);
             String url = httpClient.getBoschSmartHomeUrl("devices");
             ContentResponse contentResponse = httpClient.createRequest(url, GET).send();
 
             // check HTTP status code
             if (!HttpStatus.getCode(contentResponse.getStatus()).isSuccess()) {
                 logger.debug("Request devices failed with status code: {}", contentResponse.getStatus());
-                return false;
+                return Collections.emptyList();
             }
 
             String content = contentResponse.getContentAsString();
-            logger.debug("Request devices completed with success: {} - status code: {}", content,
+            logger.trace("Request devices completed with success: {} - status code: {}", content,
                     contentResponse.getStatus());
 
             Type collectionType = new TypeToken<ArrayList<Device>>() {
             }.getType();
-            ArrayList<Device> devices = gson.fromJson(content, collectionType);
-
-            if (devices != null) {
-                for (Device d : devices) {
-                    // Write found devices into openhab.log until we have implemented auto discovery
-                    logger.info("Found device: name={} id={}", d.name, d.id);
-                    if (d.deviceServiceIds != null) {
-                        for (String s : d.deviceServiceIds) {
-                            logger.info(".... service: {}", s);
-                        }
-                    }
-                }
-            }
+            @Nullable
+            List<Device> nullableDevices = gson.fromJson(content, collectionType);
+            return Optional.ofNullable(nullableDevices).orElse(Collections.emptyList());
         } catch (TimeoutException | ExecutionException e) {
-            logger.warn("Request devices failed because of {}!", e.getMessage());
-            return false;
+            logger.debug("Request devices failed because of {}!", e.getMessage());
+            return Collections.emptyList();
+        }
+    }
+
+    /**
+     * Get a list of rooms from the Smart-Home controller
+     *
+     * @throws InterruptedException in case bridge is stopped
+     */
+    public List<Room> getRooms() throws InterruptedException {
+        List<Room> emptyRooms = new ArrayList<>();
+        @Nullable
+        BoschHttpClient httpClient = this.httpClient;
+        if (httpClient != null) {
+            try {
+                logger.trace("Sending http request to Bosch to request rooms");
+                String url = httpClient.getBoschSmartHomeUrl("rooms");
+                ContentResponse contentResponse = httpClient.createRequest(url, GET).send();
+
+                // check HTTP status code
+                if (!HttpStatus.getCode(contentResponse.getStatus()).isSuccess()) {
+                    logger.debug("Request rooms failed with status code: {}", contentResponse.getStatus());
+                    return emptyRooms;
+                }
+
+                String content = contentResponse.getContentAsString();
+                logger.trace("Request rooms completed with success: {} - status code: {}", content,
+                        contentResponse.getStatus());
+
+                Type collectionType = new TypeToken<ArrayList<Room>>() {
+                }.getType();
+
+                ArrayList<Room> rooms = gson.fromJson(content, collectionType);
+                return Objects.requireNonNullElse(rooms, emptyRooms);
+            } catch (TimeoutException | ExecutionException e) {
+                logger.debug("Request rooms failed because of {}!", e.getMessage());
+                return emptyRooms;
+            }
+        } else {
+            return emptyRooms;
+        }
+    }
+
+    public boolean registerDiscoveryListener(ThingDiscoveryService listener) {
+        if (thingDiscoveryService == null) {
+            thingDiscoveryService = listener;
+            return true;
         }
 
-        return true;
+        return false;
+    }
+
+    public boolean unregisterDiscoveryListener() {
+        if (thingDiscoveryService != null) {
+            thingDiscoveryService = null;
+            return true;
+        }
+
+        return false;
     }
 
     /**
@@ -418,51 +516,6 @@ public class BridgeHandler extends BaseBridgeHandler {
         this.updateStatus(ThingStatus.UNKNOWN, ThingStatusDetail.UNKNOWN.NONE,
                 "@text/offline.long-polling-failed.trying-to-reconnect");
         scheduleInitialAccess(httpClient);
-    }
-
-    /**
-     * Get a list of rooms from the Smart-Home controller
-     *
-     * @throws InterruptedException in case bridge is stopped
-     */
-    private boolean getRooms() throws InterruptedException {
-        @Nullable
-        BoschHttpClient httpClient = this.httpClient;
-        if (httpClient != null) {
-            try {
-                logger.debug("Sending http request to Bosch to request rooms");
-                String url = httpClient.getBoschSmartHomeUrl("rooms");
-                ContentResponse contentResponse = httpClient.createRequest(url, GET).send();
-
-                // check HTTP status code
-                if (!HttpStatus.getCode(contentResponse.getStatus()).isSuccess()) {
-                    logger.debug("Request rooms failed with status code: {}", contentResponse.getStatus());
-                    return false;
-                }
-
-                String content = contentResponse.getContentAsString();
-                logger.debug("Request rooms completed with success: {} - status code: {}", content,
-                        contentResponse.getStatus());
-
-                Type collectionType = new TypeToken<ArrayList<Room>>() {
-                }.getType();
-
-                ArrayList<Room> rooms = gson.fromJson(content, collectionType);
-
-                if (rooms != null) {
-                    for (Room r : rooms) {
-                        logger.info("Found room: {}", r.name);
-                    }
-                }
-
-                return true;
-            } catch (TimeoutException | ExecutionException e) {
-                logger.warn("Request rooms failed because of {}!", e.getMessage());
-                return false;
-            }
-        } else {
-            return false;
-        }
     }
 
     public Device getDeviceInfo(String deviceId)
--- a/bundles/org.openhab.binding.boschshc/src/main/java/org/openhab/binding/boschshc/internal/devices/bridge/dto/PublicInformation.java
+++ b/bundles/org.openhab.binding.boschshc/src/main/java/org/openhab/binding/boschshc/internal/devices/bridge/dto/PublicInformation.java
@@ -1,0 +1,45 @@
+/**
+ * Copyright (c) 2010-2023 Contributors to the openHAB project
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package org.openhab.binding.boschshc.internal.devices.bridge.dto;
+
+import java.util.List;
+
+/**
+ * Public Information of the controller.
+ * <p>
+ *
+ * Currently, only the ipAddress is used for discovery. More fields can be added on demand.
+ * <p>
+ * Json example:
+ *
+ * <pre>
+ * {
+ * "apiVersions":["1.2","2.1"],
+ * ...
+ * "shcIpAddress":"192.168.1.2",
+ * ...
+ * }
+ * </pre>
+ *
+ * @author Gerd Zanker - Initial contribution
+ */
+public class PublicInformation {
+    public PublicInformation() {
+        this.shcIpAddress = "";
+        this.shcGeneration = "";
+    }
+
+    public List<String> apiVersions;
+    public String shcIpAddress;
+    public String shcGeneration;
+}
--- a/bundles/org.openhab.binding.boschshc/src/main/java/org/openhab/binding/boschshc/internal/discovery/BridgeDiscoveryParticipant.java
+++ b/bundles/org.openhab.binding.boschshc/src/main/java/org/openhab/binding/boschshc/internal/discovery/BridgeDiscoveryParticipant.java
@@ -1,0 +1,161 @@
+/**
+ * Copyright (c) 2010-2023 Contributors to the openHAB project
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package org.openhab.binding.boschshc.internal.discovery;
+
+import static org.openhab.binding.boschshc.internal.devices.BoschSHCBindingConstants.BINDING_ID;
+
+import java.time.Duration;
+import java.util.Set;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.TimeoutException;
+
+import javax.jmdns.ServiceInfo;
+
+import org.eclipse.jdt.annotation.NonNullByDefault;
+import org.eclipse.jdt.annotation.Nullable;
+import org.eclipse.jetty.client.HttpClient;
+import org.eclipse.jetty.client.api.ContentResponse;
+import org.eclipse.jetty.http.HttpMethod;
+import org.eclipse.jetty.http.HttpStatus;
+import org.eclipse.jetty.util.ssl.SslContextFactory;
+import org.openhab.binding.boschshc.internal.devices.BoschSHCBindingConstants;
+import org.openhab.binding.boschshc.internal.devices.bridge.BoschHttpClient;
+import org.openhab.binding.boschshc.internal.devices.bridge.dto.PublicInformation;
+import org.openhab.core.config.discovery.DiscoveryResult;
+import org.openhab.core.config.discovery.DiscoveryResultBuilder;
+import org.openhab.core.config.discovery.mdns.MDNSDiscoveryParticipant;
+import org.openhab.core.io.net.http.HttpClientFactory;
+import org.openhab.core.thing.ThingTypeUID;
+import org.openhab.core.thing.ThingUID;
+import org.osgi.service.component.annotations.Activate;
+import org.osgi.service.component.annotations.Component;
+import org.osgi.service.component.annotations.Reference;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.google.gson.Gson;
+
+/**
+ * The {@link BridgeDiscoveryParticipant} is responsible discovering the
+ * Bosch Smart Home Controller as a Bridge with the mDNS services.
+ *
+ * @author Gerd Zanker - Initial contribution
+ */
+@NonNullByDefault
+@Component(configurationPid = "discovery.boschsmarthomebridge")
+public class BridgeDiscoveryParticipant implements MDNSDiscoveryParticipant {
+    private static final long TTL_SECONDS = Duration.ofHours(1).toSeconds();
+    public static final Set<ThingTypeUID> SUPPORTED_THING_TYPES_UIDS = Set.of(BoschSHCBindingConstants.THING_TYPE_SHC);
+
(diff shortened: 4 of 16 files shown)


GerdZanker:
Hello @david-pace, can you already take a look at the code and how I now write the test?
There are two TODOs regarding the list of supported things pending which I will update within the next days.

david-pace:
I'll look at it as soon as I find some time, probably at the weekend.

bundles/org.openhab.binding.boschshc/README.md
```diff
@@ -206,12 +206,15 @@ The smoke detector warns you in case of fire.
...
-- Discovery of Things
-- Discovery of Bridge
+No major limitation known.
+Check list of [openhab issues with "boshshc"](https://github.com/openhab/openhab-addons/issues?q=is%3Aissue+boschshc+) 
 
 ## Discovery
 
-Configuration via configuration files or UI (see below).
+Bridge discovery is supported via mDNS.
+Things discovery is started after successfull pairing.
```
david-pace, line 215: typo: successful

bundles/org.openhab.binding.boschshc/src/main/java/org/openhab/binding/boschshc/internal/discovery/BridgeDiscoveryParticipant.java
```diff
@@ -54,6 +54,9 @@ public class BridgeDiscoveryParticipant implements MDNSDiscoveryParticipant {
     private final HttpClient httpClient;
     private final Gson gson = new Gson();
 
+    /// SHC Bridge Information, read via public REST API if bridge is detected. Else String are empty
```
david-pace, line 57: suggestion: Otherwise strings are empty

bundles/org.openhab.binding.boschshc/src/main/java/org/openhab/binding/boschshc/internal/discovery/ThingDiscoveryService.java
```diff
@@ -83,11 +83,19 @@ public ThingDiscoveryService() {
...
         logger.trace("activate");
+        final BridgeHandler handler = shcBridgeHandler;
+        if (handler != null) {
+            handler.registerDiscoveryListener(this);
+        }
     }
 
     @Override
     public void deactivate() {
         logger.trace("DEactivate");
```
david-pace, line 94: typo (case): deactivate

bundles/org.openhab.binding.boschshc/src/main/java/org/openhab/binding/boschshc/internal/discovery/ThingDiscoveryService.java
```diff
@@ -0,0 +1,210 @@
...
+    // BoschSHCBindingConstants.THING_TYPE_WINDOW_CONTACT
+    // TODO add other available SHC things
+    );
+
+    // @formatter:off
+    protected static final Map<String, String> DEVICEMODEL_TO_THINGTYPE_MAP = Map.ofEntries(
+            new AbstractMap.SimpleEntry<>("BBL", BoschSHCBindingConstants.THING_TYPE_SHUTTER_CONTROL.getId()), // Rollladenschalter
+            new AbstractMap.SimpleEntry<>("TWINGUARD", BoschSHCBindingConstants.THING_TYPE_TWINGUARD.getId()), // Twinguard
+            new AbstractMap.SimpleEntry<>("PSM", BoschSHCBindingConstants.THING_TYPE_INWALL_SWITCH.getId()), // Steckdosen Zwischenstecker
+            new AbstractMap.SimpleEntry<>("PLUG_COMPACT", BoschSHCBindingConstants.THING_TYPE_INWALL_SWITCH.getId()) // Steckdosen Zwischenstecker kompakt
```
david-pace, line 62: we now have a dedicated thing type for the plugs (not in-wall switch)

bundles/org.openhab.binding.boschshc/src/test/java/org/openhab/binding/boschshc/internal/discovery/ThingDiscoveryServiceTest.java
```diff
@@ -0,0 +1,246 @@
...
+
+    @Nullable
+    private ThingDiscoveryServiceTest fixture;
+
+    @Nullable
+    private BridgeHandler bridgeHandler;
+
+    @BeforeEach
+    void beforeEach() {
+        fixture = new ThingDiscoveryServiceTest();
```
david-pace, line 61: instead of instantiating the test class here you should instantiate an instance of the class under test (ThingDiscoveryService). Then you can also remove "extends ThingDiscoveryService"

bundles/org.openhab.binding.boschshc/src/test/java/org/openhab/binding/boschshc/internal/discovery/ThingDiscoveryServiceTest.java
```diff
@@ -0,0 +1,246 @@
...
+        bridgeHandler = mock(BridgeHandler.class);
+    }
+
+    private void setMockBridgeAsThingHandler() {
+        assert fixture != null;
```
david-pace, line 66: Those asserts are normally not needed. If beforeEach() is not called for some reason, we will get NullPointers anyway

bundles/org.openhab.binding.boschshc/src/test/java/org/openhab/binding/boschshc/internal/discovery/ThingDiscoveryServiceTest.java
```diff
@@ -0,0 +1,246 @@
...
+    @Nullable
+    private BridgeHandler bridgeHandler;
+
+    @BeforeEach
+    void beforeEach() {
+        fixture = new ThingDiscoveryServiceTest();
+        bridgeHandler = mock(BridgeHandler.class);
+    }
+
+    private void setMockBridgeAsThingHandler() {
```
david-pace, line 65: you could consider calling this in beforeEach() always, instead of calling it in many tests (but both is actually fine)

bundles/org.openhab.binding.boschshc/src/main/java/org/openhab/binding/boschshc/internal/discovery/ThingDiscoveryService.java
```diff
@@ -0,0 +1,210 @@
...
+            // BoschSHCBindingConstants.THING_TYPE_THERMOSTAT,
+            BoschSHCBindingConstants.THING_TYPE_TWINGUARD
+    // BoschSHCBindingConstants.THING_TYPE_WINDOW_CONTACT
+    // TODO add other available SHC things
+    );
+
+    // @formatter:off
+    protected static final Map<String, String> DEVICEMODEL_TO_THINGTYPE_MAP = Map.ofEntries(
+            new AbstractMap.SimpleEntry<>("BBL", BoschSHCBindingConstants.THING_TYPE_SHUTTER_CONTROL.getId()), // Rollladenschalter
+            new AbstractMap.SimpleEntry<>("TWINGUARD", BoschSHCBindingConstants.THING_TYPE_TWINGUARD.getId()), // Twinguard
```
david-pace, line 60: Could we re-use existing constants instead of introducing a new one like "TWINGUARD"?

david-pace (commented):
added some comments about things I saw at first glance

GerdZanker:
> added some comments about things I saw at first glance

Thanks a lot. I will improve your findings together with the updated list/map of devices.
I'm glad to see no big topics regarding my new mockito tests.

bundles/org.openhab.binding.boschshc/src/test/java/org/openhab/binding/boschshc/internal/discovery/ThingDiscoveryServiceTest.java
GerdZanker: Ok, I will give it a try. I wasn't aware that the functions under test can still be called by an "unrelated" class.

bundles/org.openhab.binding.boschshc/src/test/java/org/openhab/binding/boschshc/internal/discovery/ThingDiscoveryServiceTest.java
GerdZanker: My IDE was complaining about this

bundles/org.openhab.binding.boschshc/src/test/java/org/openhab/binding/boschshc/internal/discovery/ThingDiscoveryServiceTest.java
GerdZanker: Hi @david-pace, can you please help here with your Java/mockito experience?

The ThingDiscoveryService calls the ``thingDiscovered`` function of its base class for each discovered SHC thing.
I want to intersept the calls during the test and evaluate the argument.

With a derived class, like the current ThingDiscoveryServiceTest, I'm able to do this. But with the ThingDiscoveryService 
 class under test I fail, because thingDiscovered is a protected function and mockito's spy can't support in this case.

Is it better to create a new ThingDiscoveryService derived class to use this "spy" class in the test code as the fixture instance? Do you know other possibilities?

bundles/org.openhab.binding.boschshc/src/main/java/org/openhab/binding/boschshc/internal/discovery/ThingDiscoveryService.java
GerdZanker: I think no, because the existing strings like the ThinkTypeID in the BoschSHCBindingConstants are not matching to the required device model names from the SHC API.

bundles/org.openhab.binding.boschshc/src/test/java/org/openhab/binding/boschshc/internal/discovery/ThingDiscoveryServiceTest.java
david-pace: I am not aware of a way to achieve this with the standard Mockito toolset. It might be possible with PowerMockito, but this is a bit too risky/unstable in my experience. I think it is also not provided in the openHAB development setup.

I found another solution though: `thingDiscovered` will look for discovery listeners and will forward the result to them. So you can simply register a discovery listener in your `@BeforeEach` method. This listener can either be a custom implementation or also a mock. If it is a mock you can conveniently use Mockito's `verify()` methods to check how often it was called. And you can use argument captors to intercept arguments you are interested in. The `@BeforeEach` method then looks like this:

```
@BeforeEach
void beforeEach() {
    fixture = new ThingDiscoveryService();
    ...        
    DiscoveryListener discoveryListener = mock(DiscoveryListener.class);
    fixture.addDiscoveryListener(discoveryListener);
}
```

Look at `SmartBulbHandlerTest` to see an example for the combination of `verify()` and argument captors.

bundles/org.openhab.binding.boschshc/src/main/java/org/openhab/binding/boschshc/internal/discovery/ThingDiscoveryService.java
```diff
@@ -59,31 +66,22 @@ public class ThingDiscoveryService extends AbstractDiscoveryService implements T
             new AbstractMap.SimpleEntry<>("TWINGUARD", BoschSHCBindingConstants.THING_TYPE_TWINGUARD),
             new AbstractMap.SimpleEntry<>("PSM", BoschSHCBindingConstants.THING_TYPE_INWALL_SWITCH), // Steckdosen Zwischenstecker
             new AbstractMap.SimpleEntry<>("PLUG_COMPACT", BoschSHCBindingConstants.THING_TYPE_SMART_PLUG_COMPACT),
-            new AbstractMap.SimpleEntry<>("CAMERA_EYES", BoschSHCBindingConstants.THING_TYPE_CAMERA_360),
-            new AbstractMap.SimpleEntry<>("BWTH", BoschSHCBindingConstants.THING_TYPE_WALL_THERMOSTAT), // floorheating room control
+            new AbstractMap.SimpleEntry<>("CAMERA_350", BoschSHCBindingConstants.THING_TYPE_CAMERA_360),
```
david-pace, line 69: typo: `CAMERA_350` -> `CAMERA_360`

bundles/org.openhab.binding.boschshc/src/main/java/org/openhab/binding/boschshc/internal/discovery/ThingDiscoveryService.java
```diff
@@ -59,31 +66,22 @@ public class ThingDiscoveryService extends AbstractDiscoveryService implements T
             new AbstractMap.SimpleEntry<>("TWINGUARD", BoschSHCBindingConstants.THING_TYPE_TWINGUARD),
             new AbstractMap.SimpleEntry<>("PSM", BoschSHCBindingConstants.THING_TYPE_INWALL_SWITCH), // Steckdosen Zwischenstecker
```
david-pace, line 67: what does `PSM` mean?

bundles/org.openhab.binding.boschshc/src/main/java/org/openhab/binding/boschshc/internal/discovery/ThingDiscoveryService.java
```diff
@@ -59,31 +66,22 @@ public class ThingDiscoveryService extends AbstractDiscoveryService implements T
...
             new AbstractMap.SimpleEntry<>("PLUG_COMPACT", BoschSHCBindingConstants.THING_TYPE_SMART_PLUG_COMPACT),
-            new AbstractMap.SimpleEntry<>("CAMERA_EYES", BoschSHCBindingConstants.THING_TYPE_CAMERA_360),
-            new AbstractMap.SimpleEntry<>("BWTH", BoschSHCBindingConstants.THING_TYPE_WALL_THERMOSTAT), // floorheating room control
+            new AbstractMap.SimpleEntry<>("CAMERA_350", BoschSHCBindingConstants.THING_TYPE_CAMERA_360),
+            new AbstractMap.SimpleEntry<>("CAMERA_EYES", BoschSHCBindingConstants.THING_TYPE_CAMERA_EYES),
+            new AbstractMap.SimpleEntry<>("BWTH", BoschSHCBindingConstants.THING_TYPE_THERMOSTAT), // wall thermostat
```
david-pace, line 71: Why are some keys abbreviated and others not? I would align it that we have readable names everywhere if possible.

bundles/org.openhab.binding.boschshc/src/main/java/org/openhab/binding/boschshc/internal/discovery/ThingDiscoveryService.java
```diff
@@ -59,31 +66,22 @@ public class ThingDiscoveryService extends AbstractDiscoveryService implements T
...
-
-//            new AbstractMap.SimpleEntry<>("???", BoschSHCBindingConstants.THING_TYPE_WINDOW_CONTACT),
-//            new AbstractMap.SimpleEntry<>("???", BoschSHCBindingConstants.THING_TYPE_THERMOSTAT),
-//            new AbstractMap.SimpleEntry<>("???", BoschSHCBindingConstants.THING_TYPE_CAMERA_EYES),
-//            new AbstractMap.SimpleEntry<>("???", BoschSHCBindingConstants.THING_TYPE_SMART_BULB),
-
-// TODO: map all deviceModel names from the openhab logs to BoschSHC Thing Types
-//
-
+            new AbstractMap.SimpleEntry<>("HUE_LIGHT", BoschSHCBindingConstants.THING_TYPE_SMART_BULB),
```
david-pace, line 77: This is for Ledvance Smart Bulbs that connect via Zigbee directly to the bridge. If I'm not mistaken it has nothing do do with the Hue system.

bundles/org.openhab.binding.boschshc/src/test/java/org/openhab/binding/boschshc/internal/discovery/ThingDiscoveryServiceTest.java
```diff
@@ -133,73 +122,51 @@ public void testAddDevices() {
...
-        assertNotNull(fixture.lastDiscoveryResult);
-        assertThat(fixture.lastDiscoveryResult.getBindingId(), is(BoschSHCBindingConstants.BINDING_ID));
-        assertThat(fixture.lastDiscoveryResult.getThingTypeUID(), is(BoschSHCBindingConstants.THING_TYPE_TWINGUARD));
-        assertThat(fixture.lastDiscoveryResult.getThingUID().getId(), is("testDevice_ID"));
-        assertThat(fixture.lastDiscoveryResult.getBridgeUID().getId(), is("testSHC"));
-        assertThat(fixture.lastDiscoveryResult.getLabel(), is("Test Name"));
-        assertThat(fixture.lastDiscoveryResult.getProperties().get("Location").toString(), is("TestRoom"));
+        verify(discoveryListener).thingDiscovered(discoveryServiceCaptor.capture(), discoveryResultCaptor.capture());
+
+        assertThat(discoveryServiceCaptor.getValue().getClass(), is(ThingDiscoveryService.class));
```
david-pace, line 158: you could extract `discoveryServiceCaptor.getValue()` to a local variable `discoveryService`

david-pace (changes requested):
I found a typo in the key for the 360 camera and I also have some general questions about the keys.

bundles/org.openhab.binding.boschshc/src/main/java/org/openhab/binding/boschshc/internal/discovery/ThingDiscoveryService.java
GerdZanker: Thanks! Good catch!

bundles/org.openhab.binding.boschshc/src/main/java/org/openhab/binding/boschshc/internal/discovery/ThingDiscoveryService.java
GerdZanker: The whole list is reverse engineered, based on own data and on a good guess.

The [SHC API docu](https://local.apidocs.bosch-smarthome.com/#/Devices/get_devices) contains nothing. But in other SHC github projects like [boschshcpy](https://github.com/tschamm/boschshcpy/blob/master/boschshcpy/models_impl.py) you can find similar lists.

The ``PSM`` is the normal Smart Plug. I guess most the stuff is related to the HomeMatic IP system.

bundles/org.openhab.binding.boschshc/src/main/java/org/openhab/binding/boschshc/internal/discovery/ThingDiscoveryService.java
GerdZanker: The key is the content of the DeviceModel field inside the [Device SHC REST Schema](https://local.apidocs.bosch-smarthome.com/#/Devices/get_devices)

I not defined the key and we can't change it.

bundles/org.openhab.binding.boschshc/src/main/java/org/openhab/binding/boschshc/internal/discovery/ThingDiscoveryService.java
GerdZanker: Hmm, this works for me:
The Hue light bulb is connected to a Hue bridge. The Hue bridge is added to the SHC bridge.
I can control a "Smart BULB" light in openhab in the same way as a light via the Bosch SHC App.

GerdZanker:
> I found a typo in the key for the 360 camera and I also have some general questions about the keys.

Fixed the typo, improved the tests code and tried to clarify the questions.
Of cause the available information about your questions is weak and I guess the mapping must be improved later one. But it's the best I know currently.

Soon I will squash and cleanup the commits and provide a bundle Jar for testing with last stable version 3.4.2

GerdZanker:
Test failing, because of updated code. Branch will be rebased and fixed soon.

bundles/org.openhab.binding.boschshc/src/main/java/org/openhab/binding/boschshc/internal/discovery/ThingDiscoveryService.java
david-pace: Got it, these constants are predefined from Bosch and we cannot change them :+1:

bundles/org.openhab.binding.boschshc/src/main/java/org/openhab/binding/boschshc/internal/devices/bridge/BridgeHandler.java
```diff
@@ -225,19 +240,21 @@ private void scheduleInitialAccess(BoschHttpClient httpClient) {
...
+            if (!checkBridgeAccess()) {
                 this.updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.OFFLINE.COMMUNICATION_ERROR,
                         "@text/offline.not-reachable");
                 // restart initial access
                 scheduleInitialAccess(httpClient);
                 return;
             }
 
+            // do thing discovery after pairing
+            final ThingDiscoveryService discovery = thingDiscoveryService;
```
david-pace, line 253: Why is the discovery service accessed via a local variable and not directly via the field `thingDiscoveryService`?

bundles/org.openhab.binding.boschshc/src/main/java/org/openhab/binding/boschshc/internal/devices/bridge/BridgeHandler.java
```diff
@@ -252,54 +269,132 @@ private void scheduleInitialAccess(BoschHttpClient httpClient) {
...
+            // check HTTP status code
+            if (!HttpStatus.getCode(contentResponse.getStatus()).isSuccess()) {
+                logger.debug("Access check failed with status code: {}", contentResponse.getStatus());
+                return false;
+            }
+
+            // Access OK
+            return true;
+        } catch (TimeoutException | ExecutionException | InterruptedException e) {
+            logger.warn("Access check failed because of {}!", e.getMessage());
```
david-pace, line 298: Shouldn't we add the exception as an argument to the logger call so we can see the stack trace in the log?

bundles/org.openhab.binding.boschshc/src/main/java/org/openhab/binding/boschshc/internal/devices/bridge/BridgeHandler.java
```diff
@@ -252,54 +269,132 @@ private void scheduleInitialAccess(BoschHttpClient httpClient) {
...
+
+            // check HTTP status code
+            if (!HttpStatus.getCode(contentResponse.getStatus()).isSuccess()) {
+                logger.debug("Access check failed with status code: {}", contentResponse.getStatus());
+                return false;
+            }
+
+            // Access OK
+            return true;
+        } catch (TimeoutException | ExecutionException | InterruptedException e) {
```
david-pace, line 297: InterruptedExceptions need special handling because of the interrupt flag of the thread

bundles/org.openhab.binding.boschshc/src/main/java/org/openhab/binding/boschshc/internal/devices/bridge/BridgeHandler.java
```diff
@@ -252,54 +269,132 @@ private void scheduleInitialAccess(BoschHttpClient httpClient) {
...
+    }
+
     /**
      * Get a list of connected devices from the Smart-Home Controller
      *
      * @throws InterruptedException in case bridge is stopped
      */
-    private boolean getDevices() throws InterruptedException {
+    public List<Device> getDevices() throws InterruptedException {
+        List<Device> emptyDevices = new ArrayList<Device>();
```
david-pace, line 309: Use the diamond operator: `new ArrayList<>()`. You can also think about removing this list and use ? `Collections.emptyList()` in the early returns.

bundles/org.openhab.binding.boschshc/src/main/java/org/openhab/binding/boschshc/internal/discovery/BridgeDiscoveryParticipant.java
```diff
@@ -0,0 +1,152 @@
...
+            return null;
+        }
+
+        logger.trace("Discovered Bosch Smart Home Controller at {}", bridgeInformation.shcIpAddress);
+
+        return DiscoveryResultBuilder.create(uid)
+                .withLabel("Bosch Smart Home Controller (" + bridgeInformation.shcIpAddress + ")")
+                .withProperty("ipAddress", bridgeInformation.shcIpAddress)
+                .withProperty("shcGeneration", bridgeInformation.shcGeneration)
+                .withProperty("apiVersions", bridgeInformation.apiVersions).withTTL(60 * 60 * 24).build();
```
david-pace, line 100: The expression `60 * 60 * 24` could be extracted to a constant. Also SonarLint complains because one of the operands should be a long (see https://rules.sonarsource.com/java/RSPEC-2184)

bundles/org.openhab.binding.boschshc/src/main/java/org/openhab/binding/boschshc/internal/discovery/BridgeDiscoveryParticipant.java
```diff
@@ -0,0 +1,152 @@
...
+                return new PublicInformation();
+            }
+            // get content from response
+            String content = contentResponse.getContentAsString();
+            logger.trace("Discovered SHC - public info {}", content);
+            PublicInformation bridgeInfo = gson.fromJson(content, PublicInformation.class);
+            if (bridgeInfo.shcIpAddress != null) {
+                return bridgeInfo;
+            }
+        } catch (InterruptedException | TimeoutException | ExecutionException e) {
```
david-pace, line 145: `InterruptedException` should be handled separately because of the thread interrupt flag

bundles/org.openhab.binding.boschshc/src/main/java/org/openhab/binding/boschshc/internal/discovery/ThingDiscoveryService.java
```diff
@@ -0,0 +1,225 @@
...
+            BoschSHCBindingConstants.THING_TYPE_CAMERA_360, BoschSHCBindingConstants.THING_TYPE_CAMERA_EYES,
+            BoschSHCBindingConstants.THING_TYPE_INTRUSION_DETECTION_SYSTEM,
+            BoschSHCBindingConstants.THING_TYPE_SMART_PLUG_COMPACT, BoschSHCBindingConstants.THING_TYPE_SMART_BULB,
+            BoschSHCBindingConstants.THING_TYPE_SMOKE_DETECTOR);
+
+    // @formatter:off
+    protected static final Map<String, ThingTypeUID> DEVICEMODEL_TO_THINGTYPE_MAP = Map.ofEntries(
+            new AbstractMap.SimpleEntry<>("BBL", BoschSHCBindingConstants.THING_TYPE_SHUTTER_CONTROL),
+            new AbstractMap.SimpleEntry<>("TWINGUARD", BoschSHCBindingConstants.THING_TYPE_TWINGUARD),
+            new AbstractMap.SimpleEntry<>("PSM", BoschSHCBindingConstants.THING_TYPE_INWALL_SWITCH), // Steckdosen Zwischenstecker
```
david-pace, line 67: Maybe we should not add German comments, better use something like `// also applies for smart plugs`

bundles/org.openhab.binding.boschshc/src/main/java/org/openhab/binding/boschshc/internal/discovery/ThingDiscoveryService.java
```diff
@@ -0,0 +1,225 @@
...
+                if (d.deviceServiceIds != null) {
+                    for (String s : d.deviceServiceIds) {
+                        logger.info(".... service: {}", s);
+                    }
+                }
+            }
+
+            addDevices(devices, rooms);
+        } catch (InterruptedException e) {
+            logger.trace("scan was interrupted");
```
david-pace, line 169: The interrput flag should be restored using `Thread.currentThread().interrupt();`, see https://rules.sonarsource.com/java/RSPEC-2142

bundles/org.openhab.binding.boschshc/src/main/resources/OH-INF/i18n/boschshc.properties
```diff
@@ -135,6 +135,6 @@ offline.not-reachable = The Bosch Smart Home Controller is not reachable.
 offline.conf-error-ssl = The SSL connection to the Bosch Smart Home Controller is not possible.
 offline.long-polling-failed.http-client-null = Long polling failed and could not be restarted because http client is null.
 offline.long-polling-failed.trying-to-reconnect = Long polling failed, will try to reconnect.
-offline.interrupted = Conneting to Bosch Smart Home Controller was interrupted.
+offline.interrupted = Connecting to Bosch Smart Home Controller was interrupted.
```
david-pace, line 138: Connection (?)

bundles/org.openhab.binding.boschshc/src/main/java/org/openhab/binding/boschshc/internal/discovery/BridgeDiscoveryParticipant.java
```diff
@@ -0,0 +1,152 @@
...
+            // store all information for later access
+            bridgeInformation = getPublicInformationFromPossibleBridgeAddress(address);
+        }
+
+        return bridgeInformation;
+    }
+
+    protected PublicInformation getPublicInformationFromPossibleBridgeAddress(String ipAddress) {
+
+        String url = String.format("https://%s:8446/smarthome/public/information", ipAddress);
```
david-pace, line 128: We already have a method in `BoschHttpClient` that returns this. If possible we should re-use it.

david-pace (changes requested):
See comments, plus some checkstyle warnings have been introduced that should be fixed if possible. See file `target/code-analysis/report.html`. On the `main` branch I currently see 3 warnings related to catching NullPointerExceptions, on your branch I see 19 additional minor things that could be mitigated easily (like unnecessary empty lines etc.)

bundles/org.openhab.binding.boschshc/src/main/java/org/openhab/binding/boschshc/internal/devices/bridge/BridgeHandler.java
GerdZanker: I followed my example/reference code in ``HueBridgeHandler``.
I assume the local final variable avoids multi-threading problems, because the field can be unset/set at any time.

bundles/org.openhab.binding.boschshc/src/main/java/org/openhab/binding/boschshc/internal/devices/bridge/BridgeHandler.java
GerdZanker: I followed the [code guideline/logging](https://www.openhab.org/docs/developer/guidelines.html#f-logging) and a catched exception at this place not a bug in the binding.
We can even discuss if the warning level should be changed to ``info``, because this line is a left over from the original ``getRooms()`` code.

bundles/org.openhab.binding.boschshc/src/main/java/org/openhab/binding/boschshc/internal/devices/bridge/BridgeHandler.java
GerdZanker: - [ ]  Oh yes, the function needs to throw it
Will be changed soon

GerdZanker:
> See comments, 

All comments should be improved, please check how I handled the InterruptException.

> plus some checkstyle warnings have been introduced that should be fixed if possible. See file `target/code-analysis/report.html`. On the `main` branch I currently see 3 warnings related to catching NullPointerExceptions, on your branch I see 19 additional minor things that could be mitigated easily (like unnecessary empty lines etc.)

I did my best to get rid of all findings

bundles/org.openhab.binding.boschshc/src/main/java/org/openhab/binding/boschshc/internal/devices/bridge/dto/PublicInformation.java
```diff
@@ -0,0 +1,41 @@
...
+import java.util.List;
+
+/**
+ * Public Information of the controller.
+ *
+ * Currently, only the ipAddress is used for discovery. More fields can be added on demand.
+ *
+ * Json example:
+ * {
+ * "apiVersions":["1.2","2.1"],
```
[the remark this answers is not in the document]
david-pace: It looks very good now in general, but there is now an unmatched `>` in line 20.

bundles/org.openhab.binding.boschshc/src/main/java/org/openhab/binding/boschshc/internal/discovery/BridgeDiscoveryParticipant.java
david-pace: There is now a `catch` clause for all remaining Exceptions, but according to SonarLint we need an explicit catch clause for `InterruptedException` that contains `Thread.currentThread().interrupt();`, see https://rules.sonarsource.com/java/RSPEC-2142

If you don't have it yet, I can recommend the SonarLint plugin. You can simply install it via the Eclipse Marketplace.

david-pace (changes requested):
Only a few minor things remaining 👍  I also tested your change with real hardware and everything works nicely 😎

One idea for an enhancement would be the labels for the following items. It would look nicer if they did not have the surrounding `-` characters and proper readable names:

grafik

bundles/org.openhab.binding.boschshc/src/main/java/org/openhab/binding/boschshc/internal/discovery/BridgeDiscoveryParticipant.java
```diff
@@ -0,0 +1,153 @@
...
+    private PublicInformation bridgeInformation = new PublicInformation();
+
+    public BridgeDiscoveryParticipant() {
+        // create http client upfront to later request public information from SHC
+        SslContextFactory sslContextFactory = new SslContextFactory.Client.Client(true); // Accept all certificates
+        sslContextFactory.setTrustAll(true);
+        sslContextFactory.setValidateCerts(false);
+        sslContextFactory.setValidatePeerCerts(false);
+        sslContextFactory.setEndpointIdentificationAlgorithm(null);
+        httpClient = new HttpClient(sslContextFactory);
```
jlaur, line 69: This is now possible and will ensure better thread names in the thread pool:
```suggestion
        httpClient = httpClientFactory.createHttpClient(BINDING_ID, sslContextFactory);
```

(of course requiring `httpClientFactory` to be injected)

GerdZanker:
> Only a few minor things remaining 👍 I also tested your change with real hardware and everything works nicely 😎

Thanks  for your testfeedback


> One idea for an enhancement would be the labels for the following items. It would look nicer if they did not have the surrounding `-` characters and proper readable names:

It took some time to find a solution for nice names, but now this enhancement is part of this PR.

bundles/org.openhab.binding.boschshc/src/main/java/org/openhab/binding/boschshc/internal/discovery/BridgeDiscoveryParticipant.java
GerdZanker: Please review my solution.

bundles/org.openhab.binding.boschshc/src/main/java/org/openhab/binding/boschshc/internal/discovery/ThingDiscoveryService.java
```diff
@@ -0,0 +1,244 @@
...
+                thingUID, thingTypeUID, device.id, device.deviceModel);
+    }
+
+    private String getNiceName(String name, String roomName) {
+        // remove "-" from "-IntrusionDetectionSystem-"
+        String niceName = name.replace("-", "");
+        // convert "IntrusionDetectionSystem" into "Intrusion Detection System"
+        String[] allParts = StringUtils.splitByCharacterTypeCamelCase(niceName);
+        String[] parts = Arrays.stream(allParts).filter(p -> !p.trim().isEmpty()).collect(Collectors.toList())
+                .toArray(new String[0]);
```
david-pace, line 225: Array conversion is not needed. You could use `String.join()` providing a list of strings. But there is an even better way: use `Collectors.joining(" ")` instead of `Collectors.toList()` and you're done directly 👍

david-pace (approved):
Awesome, this looks great! Thanks for improving the labels, I see that this was not trivial ;) I tested again with the actual hardware and it looks good 👍 

I just have a minor suggestion which could save one line of code, but this is optional. Good work 😎

bundles/org.openhab.binding.boschshc/src/main/java/org/openhab/binding/boschshc/internal/discovery/ThingDiscoveryService.java
```diff
@@ -0,0 +1,244 @@
...
+
+        // Write found devices into openhab.log to support manual configuration
+        for (Device d : devices) {
+            logger.info("Found device: name={} id={}", d.name, d.id);
+            if (d.deviceServiceIds != null) {
+                for (String s : d.deviceServiceIds) {
+                    logger.info(".... service: {}", s);
+                }
+            }
+        }
```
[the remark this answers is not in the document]
GerdZanker: You are right, the ``logger.info`` is now not the right way anymore and the manual configuration can be done differently.
I will change the log level to ``debug``.

Your hint with the console command sound great. I will keep it in mind for a new, different pull request.

bundles/org.openhab.binding.boschshc/README.md
```diff
@@ -239,19 +242,19 @@ Alternatively, the log can be viewed using the OpenHab Log Viewer (frontail) via
...
-2020-08-11 12:42:49.502 [INFO ] [chshc.internal.BoschSHCBridgeHandler] - Found device: name=Heizung id=hdm:HomeMaticIP:3014F711A000XXXXXXXXXXXX
-2020-08-11 12:42:49.503 [INFO ] [chshc.internal.BoschSHCBridgeHandler] - Found device: name=Heizung Haus id=hdm:ICom:819410185:HC1
-2020-08-11 12:42:49.503 [INFO ] [chshc.internal.BoschSHCBridgeHandler] - Found device: name=-RoomClimateControl- id=roomClimateControl_hz_6
-2020-08-11 12:42:49.504 [INFO ] [chshc.internal.BoschSHCBridgeHandler] - Found device: name=PhilipsHueBridgeManager id=hdm:PhilipsHueBridge:PhilipsHueBridgeManager
-2020-08-11 12:42:49.505 [INFO ] [chshc.internal.BoschSHCBridgeHandler] - Found device: name=Rollladen id=hdm:HomeMaticIP:3014F711A000XXXXXXXXXXXX
-2020-08-11 12:42:49.506 [INFO ] [chshc.internal.BoschSHCBridgeHandler] - Found device: name=Rollladen id=hdm:HomeMaticIP:3014F711A000XXXXXXXXXXXX
-2020-08-11 12:42:49.507 [INFO ] [chshc.internal.BoschSHCBridgeHandler] - Found device: name=Central Heating id=hdm:ICom:819410185
+2023-01-23 20:28:54.725 [INFO ] [rnal.discovery.ThingDiscoveryService] - Found device: name=RoomLightControl id=roomLightControl_hz_XXXXXXXX
```
GerdZanker, line 245: - [ ]  This part needs an update, because the log statements are now removed

GerdZanker: Readme is updated. The new added things are listed in the log.

GerdZanker:
Out of topic, but @jlaur, do you know how I can provide a nice binding icon for boschshc? I wasn't able to find a documentation or used the wrong words.

jlaur:
> Out of topic, but @jlaur, do you know how I can provide a nice binding icon for boschshc? I wasn't able to find a documentation or used the wrong words.

Yes, see here: https://www.openhab.org/docs/developer/bindings/#add-your-binding-s-logo-to-the-openhab-website

bundles/org.openhab.binding.boschshc/README.md
```diff
@@ -239,19 +242,10 @@ Alternatively, the log can be viewed using the OpenHab Log Viewer (frontail) via
...
-2020-08-11 12:42:49.505 [INFO ] [chshc.internal.BoschSHCBridgeHandler] - Found device: name=Rollladen id=hdm:HomeMaticIP:3014F711A000XXXXXXXXXXXX
-2020-08-11 12:42:49.506 [INFO ] [chshc.internal.BoschSHCBridgeHandler] - Found device: name=Rollladen id=hdm:HomeMaticIP:3014F711A000XXXXXXXXXXXX
-2020-08-11 12:42:49.507 [INFO ] [chshc.internal.BoschSHCBridgeHandler] - Found device: name=Central Heating id=hdm:ICom:819410185
+2023-03-20 20:30:48.026 [INFO ] [g.discovery.internal.PersistentInbox] - Added new thing 'boschshc:security-camera-eyes:yourBrideName:hdm_Cameras_XXXXXXXX-XXXX-XXXX-XXXX-XXXXXXXXXXXX' to inbox.
+2023-03-20 20:30:48.026 [INFO ] [g.discovery.internal.PersistentInbox] - Added new thing 'boschshc:smoke-detector:yourBrideName:hdm_HomeMaticIP_XXXXXXXXXXXXXXXXXXXXXXXX' to inbox.
+2023-03-20 20:30:48.027 [INFO ] [g.discovery.internal.PersistentInbox] - Added new thing 'boschshc:twinguard:yourBrideName:hdm_ZigBee_XXXXXXXXXXXXXXXX' to inbox.
+2023-03-20 20:30:48.028 [INFO ] [g.discovery.internal.PersistentInbox] - Added new thing 'boschshc:smart-bulb:yourBrideName:hdm_PhilipsHueBridge_HueLight_XXXXXXXXXXXXXXXX-XX_XXXXXXXXXXXX' to inbox.
```
jlaur, line 248: yourBrideName intended? 👰

GerdZanker: Arg - last minute changes 🤦‍♂️

bundles/org.openhab.binding.boschshc/src/main/java/org/openhab/binding/boschshc/internal/discovery/ThingDiscoveryService.java
```diff
@@ -0,0 +1,243 @@
...
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package org.openhab.binding.boschshc.internal.discovery;
+
+import java.util.*;
+import java.util.stream.Collectors;
+
+import org.apache.commons.lang3.StringUtils;
```
jlaur, line 18: Sorry for such late finding, was just following up on https://github.com/openhab/openhab-addons/pull/14197#discussion_r1141044386 and noticed this. We are actually trying to get rid of all usages of this class. See #7722 and a list PR's using the filter "is:open is:pr in:title org.apache.common".

jlaur (changes requested):
Last minute finding, sorry.

bundles/org.openhab.binding.boschshc/src/main/java/org/openhab/binding/boschshc/internal/discovery/ThingDiscoveryService.java
GerdZanker: No problem, but I'm not able provide a last minute fix today, because it took me some hours to find a solution.
I will replace the used ``StringUtils.splitByCharacterTypeCamelCase`` function during the next days

jlaur: Sure, let's not rush it, it would be counter-productive to introduce more usages while trying to get rid of those we already have. 🙂 I don't have any further comments, so we should be able to merge shortly after refactoring this.

david-pace: You could use https://www.baeldung.com/java-camel-case-title-case-to-words and/or https://stackoverflow.com/questions/7593969/regex-to-split-camelcase-or-titlecase-advanced as a starting point.

jlaur: And also of course: https://github.com/apache/commons-lang/blob/2cf6486197b268bac818d05e38173ab57ce16315/src/main/java/org/apache/commons/lang3/StringUtils.java#L7458-L7460

Perhaps only a subset of that functionality is needed.

GerdZanker: I took the org.apache.commons.lang code as an example to implement own code to add spaces for camel cases.

jlaur (approved):
LGTM

bundles/org.openhab.binding.boschshc/src/main/java/org/openhab/binding/boschshc/internal/discovery/ThingDiscoveryService.java
lsiepel: Sorry, way to late as it is allready merged. Just want to mention an implementation was available, including tests, here:  https://github.com/openhab/openhab-addons/pull/14434/files

PR #14213, opened 2023-01-13
[homematic] Fix updating enum config values

maniac103:
When changing an enum value in the configuration, we used the wrong data type: while the value in the OH config is a string (the 'option value' - see HomematicThingHandler::getValueForConfiguration), internally we use an integer (the 'option index'), so we have to do the option value -> option index conversion when applying the new value. This especially was a problem for HM-MOD-EM-8 devices, which check the CHANNEL_FUNCTION enum value as part of their initialization routine. When disabling/enabling them after changing the CHANNEL_FUNCTION enum value, they went offline, because their initialization failed due to a NumberFormatException (via
HomematicThingHandler::doInitializeInBackground -> HmChannel::checkForChannelFunctionChange ->
HmChannel::getCurrentFunction)


jlaur (approved):
Thanks for the fix!

Issue #14207, opened 2023-01-12
[gree] binding does not allow target temperatures below 16°C


thallaa:
I have a Nordic model Gree AC/heatpump that I'm using to heat my garage. It has a temperature range of 46°F to 86°F or 8°C to 30°C. As far as I can tell, the minimum has been hard coded to 16°C in GreeBindingConstants.java:

```
public static final int TEMP_MIN_C = 16;
public static final int TEMP_MIN_F = 61;
```

I am normally keeping the garage temperature at around 14°C but the binding does not allow this. Also, when trying to set the temperature below 16°C, the error reported in openhab.log is not very clear:

`2023-01-12 13:00:01.541 [INFO ] [ng.gree.internal.handler.GreeHandler] - e890de98ab: Unable to load message for key message.command.invarg
`

Maybe the debug log would have pointed this out but I did not try that before I found the issue.

I did not find this reported before (this is not the same as the 8°C mode if I understand that one correctly).

lsiepel:
Do you have some official documentation stating the min/ max values?

maybe it would be better to remove the range check in the binding and leave it to the user to set this at item level. But this may lead to situations that a temperature is set outside of the device capabilities. It would be interesting to see how the device responds to that.

thallaa:
I have checked the temperature range from the remote controller unit and from the mobile application. The best document I was able to find online was this brochure:
https://www.greenspirit.ee/download4u/gree/gree_eng/GREE-RAC-catalogue-2019-www.pdf
Page 17 mentions the range 8°C to 30°C.
The exact model I have is Gree Amber Nordic 35.

If this change get released, I am volunteering to try out how the device reacts.

lsiepel:
3.4.1 version with the min temp value reduced from 16 to 5 celsius.
https://1drv.ms/u/s!AnMcxmvEeupwjqUKiNB08dZcB8xvlQ?e=kIimEH 

Not a final fix, just to see what is happening if you try to set it below the treshold of 8. Could you enable debug logging en let me know what happens?

thallaa:
I installed the snapshot and tested what happens if I set it below 8°C and it works just fine - the temperature stops at 8°C and simply seems to ignore values below that. The display on the unit itself shows 8° as does the iPhone app.

I do get some log entries but they do not seem to be related to this as I get the same entries for any temperature:

```
[...]

PR #14217, opened 2023-01-14
[gree] Allow wider temperature range

lsiepel:
- Fix checkstyle
 - Allow wider range of temperatures

Previously the binding limits the setpoint temperature from 16-30. As devices have different operating range, it would be better to have a wider allowed range and adjust it with sitemap minValue/ maxValue according to your specific device.

Tested and confirmed. Devices gracefully ignore temperatures outside of the operating range. So regressions or device mallfunction is not to be expected.

Fixes: https://github.com/openhab/openhab-addons/issues/14207

### DIFF
--- a/bundles/org.openhab.binding.gree/src/main/java/org/openhab/binding/gree/internal/GreeBindingConstants.java
+++ b/bundles/org.openhab.binding.gree/src/main/java/org/openhab/binding/gree/internal/GreeBindingConstants.java
@@ -143,9 +143,9 @@ public class GreeBindingConstants {
     // Temperatur types and min/max ranges
     public static final int TEMP_UNIT_CELSIUS = 0;
     public static final int TEMP_UNIT_FAHRENHEIT = 1;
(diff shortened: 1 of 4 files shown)


lolodomo (approved):
LGTM, thank you

jlaur:
@lsiepel - does it work correctly when having this definition where minimum is still 16?
https://github.com/openhab/openhab-addons/blob/b9bec522ee0dc0b95d38bdeca40469448d1b82c5/bundles/org.openhab.binding.gree/src/main/resources/OH-INF/thing/thing-types.xml#L65-L70

lolodomo:
I probably merged too fast, sorry !

lsiepel:
The user that tested was able to send a 7 celsius command, guess he used textual config as the quoted code would only affect Graphic config right? I'll make a PR fast :-)

lsiepel:
> I probably merged too fast, sorry !

Sorry, i also missed it.

PR #14219, opened 2023-01-14
[gree] Fix thing-type min state

lsiepel:

Fix comment: https://github.com/openhab/openhab-addons/pull/14217#issuecomment-1382716371


jlaur (approved):
Thanks!

lsiepel:
I think the bug label should be removed as it is not needed to show up in the release notes. This one is tied closely (and a forgotten change) to https://github.com/openhab/openhab-addons/pull/14217

jlaur:
@lsiepel - it will be excluded from the release notes when "regression" is also present. 🙂

lsiepel:
> notes whe

ah, didn;'t knwo that, thanks!

jlaur:
> > notes whe
> 
> ah, didn't knwo that, thanks!

It's quite new. It makes it easier to manage what will show up in the release notes since labels can be applied immediately without remembering to go back and maintain them later. Also we can correctly categorize bugs that should not be mentioned.

PR #14221, opened 2023-01-14
[venstar] Add semantic tags to channels

digitaldan:

### DIFF
--- a/bundles/org.openhab.binding.venstarthermostat/src/main/resources/OH-INF/thing/thing-types.xml
+++ b/bundles/org.openhab.binding.venstarthermostat/src/main/resources/OH-INF/thing/thing-types.xml
@@ -114,6 +114,9 @@
 		<item-type>String</item-type>
 		<label>System Mode</label>
 		<description>Current System Operating Mode</description>
+		<tags>
+			<tag>Control</tag>
+		</tags>
 		<state readOnly="false">
 			<options>
 				<option value="off">Off</option>
@@ -284,6 +287,9 @@
 		<item-type>String</item-type>
 		<label>System State</label>
 		<description>Current System Operating State</description>
+		<tags>
+			<tag>Status</tag>
+		</tags>
 		<state readOnly="true">
 			<options>
 				<option value="idle">Idle</option>
@@ -307,6 +313,10 @@
 		<label>Heating Setpoint</label>
 		<description>Heating Setpoint</description>
 		<category>Temperature</category>
+		<tags>
+			<tag>Setpoint</tag>
+			<tag>Temperature</tag>
+		</tags>
 		<state readOnly="false" pattern="%.1f %unit%" min="40" max="80" step="1.0"/>
 	</channel-type>
 
@@ -315,6 +325,10 @@
 		<label>Cooling Setpoint</label>
 		<description>Cooling Setpoint</description>
 		<category>Temperature</category>
+		<tags>
+			<tag>Setpoint</tag>
+			<tag>Temperature</tag>
+		</tags>
 		<state readOnly="false" pattern="%.1f %unit%" min="60" max="95" step="1.0"/>
 	</channel-type>
 
@@ -323,6 +337,10 @@
 		<label>Temperature</label>
 		<description>Temperature</description>
 		<category>Temperature</category>
+		<tags>
+			<tag>Measurement</tag>
+			<tag>Temperature</tag>
+		</tags>
 		<state readOnly="true" pattern="%.1f %unit%">
 		</state>
 	</channel-type>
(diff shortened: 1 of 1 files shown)


hww3:
LGTM!

hww3 (approved):
Looks good!

bundles/org.openhab.binding.venstarthermostat/src/main/resources/OH-INF/thing/thing-types.xml
```diff
@@ -114,6 +114,9 @@
 		<item-type>String</item-type>
 		<label>System Mode</label>
 		<description>Current System Operating Mode</description>
+		<tags>
+			<tag>Control</tag>
```
lsiepel, line 118: Might conflict with rule 2: https://github.com/openhab/openhab-addons/issues/12262

digitaldan:
> Might conflict with rule 2: https://github.com/openhab/openhab-addons/issues/12262

Hmm, i would agree that it does as well.  But i am also confused then what "control" is meant with semantic tags?  It would seem reasonable to me that a channel the handles the "control" of a thing, would be tagged as such?  I am very much not clear then how such tags should be used and what benefit there is / is not with such a tag? @lolodomo maybe you can explain ?

lsiepel:
> o me that a channel the handles the "control" of a thing, would be tagged as such? I am very much not clear then how such tags should be used and what benefit there is /

I think the tag alone is not specific enough and it also needs a tag _what_ it controls, like a receiver, hvac etc. But i must agree that i find the semenatic tags a bit of a vague at some point. I really hope someone can create a set of rules and add it to the   xml definition. It starts with the rules tough ;-)

digitaldan:
@lolodomo any feedback?  Thanks!

bundles/org.openhab.binding.venstarthermostat/src/main/resources/OH-INF/thing/thing-types.xml
```diff
@@ -284,6 +287,9 @@
 		<item-type>String</item-type>
 		<label>System State</label>
 		<description>Current System Operating State</description>
+		<tags>
+			<tag>Status</tag>
+		</tags>
```
lolodomo, line 292: This one is unexpected.

lolodomo (changes requested):

bundles/org.openhab.binding.venstarthermostat/src/main/resources/OH-INF/thing/thing-types.xml
lolodomo: Yes, this one is unexpected.

lolodomo:
Only the tags you set for temperature, setpoint and humidity look good to me, considering the discussions about semantic tags we have in the past (in particular with @kaikreuzer) leading to few rules (#12262).

digitaldan:
Thanks @lolodomo , i have read that list, but i'm still not clear about this:

>It is not allowed to set only "Control" or "Switch" tag on a writable channel type. Also not expected to set "Status" tag on all read-only channel types.

I guess i don't understand "why" its not allowed?  And what would be allowed for a "Control" tag?

lolodomo:
> > It is not allowed to set only "Control" or "Switch" tag on a writable channel type. Also not expected to set "Status" tag on all read-only channel types.
> 
> I guess i don't understand "why" its not allowed? And what would be allowed for a "Control" tag?

You have to mention what is controlled, for example "Control" + "Light".

digitaldan:
So in the case of HVAC equipment, the channel which controls the primary operational mode (HEAT,COOL,AUTO,OFF) would have what additional tag?

lsiepel:
> So in the case of HVAC equipment, the channel which controls the primary operational mode (HEAT,COOL,AUTO,OFF) would have what additional tag?

Control + HVAC tags  I suppose

More equipment here:  https://github.com/openhab/openhab-core/blob/main/bundles/org.openhab.core.semantics/model/SemanticTags.csv

lolodomo:
A tag for Point + a tag for Property.
I am not sure what property is appropriate in that case. Maybe "Energy" ?
"Control" + "Energy" ?

lsiepel:
> hich controls the primary op

i would suggest to use: Control + Temperature

lsiepel:
@digitaldan do you need more information or can you proceed with this PR?

lsiepel:
I close this PR due to inactivity. Personally i think that pre-set tags in general can be added to many more bindings, so i hope that at some point this will be picked up again.

However, source is not gone. When you (or someone else) is able to proceed with this PR it can be re-opened, or feel free to create a new PR. I would be happy to review and merge it.

Issue #14222, opened 2023-01-14
Integrate Energi Data Service (Danish)


jlaur:
In order to make data from [Energi Data Service](https://www.energidataservice.dk/) easily accessible in openHAB, I would like to create an add-on implementing this [API](https://www.energidataservice.dk/guides/api-guides).

The first milestone will have focus on [Elspot Prices](https://www.energidataservice.dk/tso-electricity/Elspotprices) and add a channel being updated with the current electricity price (per kilowatt hour). This was immediately achieved when starting the work on this add-on today:

### Data Filters

Filters were based on https://github.com/MTrab/energidataservice/blob/master/custom_components/energidataservice/tariffs/energidataservice/chargeowners.py

They have all been optimized and tested, and issues based on findings were created in this repo.

## Plan

This is the public TODO (in headline form) in order to reach "MVP" and publish a pull request:
- [X] Expose future prices as advanced channel with JSON array (since I'm not aware of any neat way of doing this currently).
- [X] Implement smarter retry policy on failures (e.g. exponential backoff and jitter).
- [X] Add support for tariffs (dataset **DatahubPricelist**).
- [x] Provide pre-made filters for known grid companies.
- [x] Add channel configuration for overriding pre-made filters.
- [x] Add grid companies to option list.
- [x] Add channels for Energinet (TSO) tariffs.
- [x] Add action for calculating price for specified consumption in a period. See https://community.openhab.org/t/dishwasher-price-calculation-automation/139207.
- [x] Load/keep retrospective data, e.g. 12 hours back in time. This would make it possible to write rules giving feedback on price when an appliance has ended a program.
- [x] Documentation cleanup.
- [x] Code cleanup (and a few more unit tests).

# Status
Draft code is here (work in progress):
https://github.com/jlaur/openhab-addons/tree/14222-energidataservice/bundles/org.openhab.binding.energidataservice

## Features
- **Optimized service calls:** Spot prices are retrieved once per day and tariffs only upon expiry (usually months/years).
- **Error handing:** Policies for retry strategies are in place to make sure data will be retrieved as soon as possible after failed calls. This includes exponential back-off with jitter.
- Channel configuration for adding VAT to prices based on openHAB regional settings.
- Grid company is selectable in configuration.
- Pre-configured filters are included for all known grid companies.
- Filters can be overridden by channel configuration.
- Spot prices and all tariffs are supported.
- Properties shows number of remaining calls as well as timestamp of last call.
[...]

PR #14236, opened 2023-01-16
[Tibber] Added "today" API channel

Markinus:
This commit exends the supported Tibber API with channel "today". This channel is very similar to "tomorrow" and gets the hourly prices of the current day  as JSON string.


jlaur:
See also https://github.com/openhab/openhab-addons/pull/13416#discussion_r974668701. It's worth considering now if historical prices are relevant - what are the use-cases?
- Today's prices are partly historical (i.e. some of them might be irrelevant).
- Tomorrow's prices doesn't represent all future prices (there's a gap from now and until midnight).

If the historical prices for today are of interest, why not yesterday's prices also?

If the use-cases are mostly for calculating costs in the future in order to schedule lowest price for running dishwasher etc., if might be worth providing just one channel containing current hour + all available hours into the future (i.e. until midnight tomorrow when available).

I'm not saying anything is wrong with this PR, just wanted to brainstorm a little. I'm also interested in the use-cases. See also #14181 and #14222.

Markinus:
Hi,

I can understand your comment but I already think we should separate this PR and your concern. There are two issues that we need to clarify.

First to the "normal" operators and use cases.

- Normally an operation takes more than an hour. That means that if I want to turn something on then I don't need the cheapest starting point but the cheapest period. This can start the day before.

- Examples for smaller devices are the washing machine, dryer or dishwasher where this deviation may not be so serious.

- With larger devices, however, it looks different. These include, for example:

a. An EV or PHEV. Here, a process sometimes takes hours and that with very high power. The right time slot is very crucial. In addition, it must often be the most favorable slot at night, which may well start on today's day and continue into the next day.

b. A PV storage. In winter, the PV storage is mostly unused. During this time it can be charged with cheap electricity at night and the house can consume it from the storage during the day. Also here it can be that this slot also starts on today's day and goes over into the next day.

c. If the storage is too small and not enough for the whole day. Then I also need the cheapest slot on the day to then switch off the storage and draw electricity from the grid. This can still be today depending on how much additional PV comes over the day.

- So the user needs time slots. I think that would be something that could really help.

a. the cheapest 2/3/4/5 hour time slot, total, day and night.

b. the average price for the respective period

c. a switch that switches it so that the user can use it automatically.

- The data are therefore not only historical. I am only interested in the future data of the current day but these are very relevant in many cases.

- This would all be very useful but also associated with some effort.

 

Now to the "professional" users.

- Currently the binding has no possibility to access today's future data. This is really missing.

- This "today" interface is exactly like the "tomorrow" so that the same tools can be used.

- Someone who can script can solve all the upper use cases
- The solution was very simple and fast

Suggestion:

Let's take a two-step approach. First integrate a solution with this PR at all. Secondly, we can discuss if such time slots as described above would not be something useful.

 

Greetings

jlaur:
@Markinus - thanks for describing the use-cases, this is valuable input. I do think it confirms that the simplest solution (for the user) would be one channel containing both today and tomorrow. This way you don't have to look up the data in two different channels when performing calculations across midnight.

IMHO it shouldn't require much additional work to combine the two payloads, if you agree this would simplify matters for the users. The problem with the two-step approach (which is actually a three-step approach because first step was taken in #13416) is backwards compatibility concerns. It's easier to create new channels that it is to remove existing ones, since this is a breaking change.

WDYT?

One additional clarification:
> Normally an operation takes more than an hour. That means that if I want to turn something on then I don't need the cheapest starting point but the cheapest period. This can start the day before.

I'm not sure I fully understand this. Unless the device is already running and you want to calculate the total price until it completes, prices before current hour shouldn't be needed/relevant? If already running since yesterday and you want to calculate the total cost, you already lack data from yesterday. So having historic prices today seems to only cover a corner-case?

Markinus:
In my opinion, backwards compatibility is not really a problem. Since the channel for "tomorrow" already exists, it can't really be removed. It is then no problem if the one for "today" would also be added. Both are the same and can stay in the binding for professional users.

For normal users, in my opinion, even an array with data from now until tomorrow night is not really easier. The size of the array changes every hour and you have to search for the time again and again.  With "today" or "tomorrow" you know exactly when these areas start.

It helps a user much more if he would get switches for the most favorable time ranges at night and during the day. Then he can choose the right one and switch his devices directly with it (for 2,3,4,5 hours time ranges).

To your question an example: When I come home in the evening and plug in the PV then the charging process should not start directly. The software knows that I leave in the morning and the car needs for example:: 40 kWh. The software then searches from just now and at night the cheapest four hours where the charging can take place. This can be e.g.: from today 23h to tomorrow at 3 o'clock. So I need the data from today and tomorrow. The data from now to the past are irrelevant but they do not hurt.

If the user had direct switch items for the time ranges then he could, depending on how long the time is for the device, directly use them.

But as I said before, this PR does not hurt at all.

Markinus:
@jlaur an another idea. We could replace "tomorrow_prices" with "prices" and send the whole "prices" JSONObject. So every one should have what he need. Maybe we should extend the prices with all informations too. So there wouldn't be a new channel bute the people would have to change their code.

<html><body>
<h3 id="priceinfo" tabindex="-1">PriceInfo</h3>


Field | Argument | Type | Description
-- | -- | -- | --
current | Price | The energy price right now
today | [Price]! | The hourly prices of the current day
tomorrow | [Price]! | The hourly prices of the upcoming day


</body>
</html>PriceInfo
Field 	Argument 	Type 	Description
current 	[Price](https://developer.tibber.com/docs/reference#price) 	

The energy price right now
today 	[[Price](https://developer.tibber.com/docs/reference#price)]! 	

The hourly prices of the current day
tomorrow 	[[Price](https://developer.tibber.com/docs/reference#price)]! 	

The hourly prices of the upcoming day

<h3 id="price" tabindex="-1">Price</h3>
Field | Argument | Type | Description
-- | -- | -- | --
total | Float | The total price (energy + taxes)
energy | Float | Nord Pool spot price
tax | Float | The tax part of the price (guarantee of origin certificate, energy tax (Sweden only) and VAT)
startsAt | String | The start time of the price
currency | String! | The price currency
level | PriceLevel | The price level compared to recent price values


</body>
</html>Price
Field 	Argument 	Type 	Description
total 	[Float](https://developer.tibber.com/docs/reference#float) 	

The total price (energy + taxes)
energy 	[Float](https://developer.tibber.com/docs/reference#float) 	

Nord Pool spot price
tax 	[Float](https://developer.tibber.com/docs/reference#float) 	

The tax part of the price (guarantee of origin certificate, energy tax (Sweden only) and VAT)
startsAt 	[String](https://developer.tibber.com/docs/reference#string) 	

The start time of the price
currency 	[String](https://developer.tibber.com/docs/reference#string)! 	

The price currency
level 	[PriceLevel](https://developer.tibber.com/docs/reference#pricelevel) 	

The price level compared to recent price values

jlaur:
> another idea. We could replace "tomorrow_prices" with "priceInfo" and send the whole "PriceInfo" JSONObject.

In that case I would rather go with the new channel you have implemented in this PR. The problem I see here is that no abstraction is provided for the user, so they will still need to merge two different sources to get relevant future prices.

Just to be very verbose:

**tomorrow_prices** (example reduced):

```json
[
  {
    "startsAt": "2023-01-19T00:00:00.000+02:00",
    "total": 3.8472
  },
  {
    "startsAt": "2023-01-19T01:00:00.000+02:00",
    "total": 3.0748
  }
]
```

**today_prices**

```json
[
  {
    "startsAt": "2023-01-18T22:00:00.000+02:00",
    "total": 4.1014
  },
  {
    "startsAt": "2023-01-18T23:00:00.000+02:00",
    "total": 4.0265
  }
]
```

My proposal:

**future_prices** (when 'now' is 1:30 before midnight, thus everything from **tomorrow_prices**, but filtered prices from before start of current hour from **today_prices**)

```json
[
  {
    "startsAt": "2023-01-18T22:00:00.000+02:00",
    "total": 4.1014
  },
  {
    "startsAt": "2023-01-18T23:00:00.000+02:00",
    "total": 4.0265
  },
  {
    "startsAt": "2023-01-19T00:00:00.000+02:00",
    "total": 3.8472
  },
  {
    "startsAt": "2023-01-19T01:00:00.000+02:00",
    "total": 3.0748
  }
]
```

This would be simpler to use in rules as IMHO this separation of today and tomorrow is artificial IMHO and just reflects how Tibber for some reason has decided to expose the data. Other services behave differently, see for example https://api.energidataservice.dk/dataset/Elspotprices?start=utcnow&filter={%22PriceArea%22:%22DK1%22}&columns=HourUTC,SpotPriceDKK

Markinus:
I understand your point of view, but there is another way to look at it. If you have two separate arrays/maps/lists for both days instead of one for everything, then access is much easier. If you want to know the price for 5 o'clock then you access directly the index 5 of the arrays. If you want to know the price between 4-6 o'clock then you just take a sublist(4-6). For the night you can easily build a sublist from both, directly over the desired times. 
If everything is in one then you always have to search because the length of the arrays/maps/lists is different for each hour (because the old hours of the day are omitted). It is of course possible to deal with this but is it really easier?

Translated with www.DeepL.com/Translator (free version)

jlaur:
> I understand your point of view, but there is another way to look at it. If you have two separate arrays/maps/lists for both days instead of one for everything, then access is much easier. If you want to know the price for 5 o'clock then you access directly the index 5 of the arrays. If you want to know the price between 4-6 o'clock then you just take a sublist(4-6). For the night you can easily build a sublist from both, directly over the desired times. If everything is in one then you always have to search because the length of the arrays/maps/lists is different for each hour (because the old hours of the day are omitted). It is of course possible to deal with this but is it really easier?

Thanks for adding this perspective, I didn't think of the use-case being able to directly reference by index, so probably this is useful in combination with a transformation service?

@kjoglum, @seime - everything okay with you?

bundles/org.openhab.binding.tibber/src/main/java/org/openhab/binding/tibber/internal/handler/TibberHandler.java
```diff
@@ -187,6 +187,10 @@ public void getURLInput(String url) throws IOException {
                             .getAsJsonObject("home").getAsJsonObject("currentSubscription").getAsJsonObject("priceInfo")
                             .getAsJsonArray("tomorrow");
                     updateState(TOMORROW_PRICES, new StringType(tomorrow.toString()));
+                    JsonArray today = rootJsonObject.getAsJsonObject("data").getAsJsonObject("viewer")
+                            .getAsJsonObject("home").getAsJsonObject("currentSubscription").getAsJsonObject("priceInfo")
+                            .getAsJsonArray("today");
+                    updateState(TODAY_PRICES, new StringType(today.toString()));
```
jlaur, line 193: Consider checking if channel is linked to avoid building this array otherwise.

kjoglum:
> @kjoglum, @seime - everything okay with you?

Nothing to add from my side, looks ok.

bundles/org.openhab.binding.tibber/src/main/java/org/openhab/binding/tibber/internal/handler/TibberHandler.java
Markinus: This is just a copy like the other channels are updated. I think it doesn't make sense to add this check for this channel only. This kind of optimisation should be done in the feature for all channels in a new PR.

Markinus:
@jlaur Yes, tranformation service. I have a JS Script and it's used for both today and tomorrow.
```
function(file, id, name) {
	if ( file == null )
        return 0;
	var jsonObj = JSON.parse(file);
	var value;
	if(name == "startsAt") {
		value = jsonObj[parseInt(id)].startsAt;
	} else if (name == "total") {
		value = jsonObj[parseInt(id)].total;
	} else if (name == "number") {
		value = jsonObj.length;
	} else {
		value = "not supportet in JS"
	}

	return value;

})(input, hourId, valueName)
```
In Rule:
```
...
var List<ZonedDateTime> tomorrowTimeStamps = new ArrayList()
var List<Float> tomorrowPrices = new ArrayList()
var List<ZonedDateTime> todayTimeStamps = new ArrayList()
var List<Float> todayPrices = new ArrayList()
...
var Integer nbrData = Integer.parseInt(transform("JS", "TibberParser.js?hourId=9999&valueName=number", tomorrowPricesData))
for(var Integer i=0; i<nbrData; i++) {
	var String time = transform("JS", "TibberParser.js?hourId="+i.toString()+"&valueName=startsAt", tomorrowPricesData)
	var timeZDT = ZonedDateTime.parse(time)
	tomorrowTimeStamps.add(timeZDT.withZoneSameInstant(ZoneId.systemDefault()))
	tomorrowPrices.add(Float.parseFloat(transform("JS", "TibberParser.js?hourId="+i.toString()+"&valueName=total", tomorrowPricesData)))
	time = transform("JS", "TibberParser.js?hourId="+i.toString()+"&valueName=startsAt", todayPricesData)
	timeZDT = ZonedDateTime.parse(time)
	todayTimeStamps.add(timeZDT.withZoneSameInstant(ZoneId.systemDefault()))
	todayPrices.add(Float.parseFloat(transform("JS", "TibberParser.js?hourId="+i.toString()+"&valueName=total", todayPricesData)))
}
...

```

bundles/org.openhab.binding.tibber/src/main/resources/OH-INF/thing/thing-types.xml
```diff
@@ -122,4 +123,10 @@
 		<description>JSON array of tuples startsAt,total, e.g. {["startsAt": "2022-09-10T00:00:00+02:00", "total": 5.332},
 			{"startsAt": ...}]. See binding documantation for full example.</description>
 	</channel-type>
+	<channel-type id="today_prices" advanced="true">
+		<item-type>String</item-type>
+		<label>Prices for today as a JSON array</label>
+		<description>JSON array of tuples startsAt,total, e.g. {["startsAt": "2022-09-10T00:00:00+02:00", "total": 5.332},
+			{"startsAt": ...}]. See binding documantation for full example.</description>
```
jlaur, line 130:
```suggestion
			{"startsAt": ...}]. See binding documentation for full example.</description>
```

bundles/org.openhab.binding.tibber/src/main/resources/OH-INF/i18n/tibber.properties
```diff
@@ -40,4 +40,6 @@ channel-type.tibber.timestamp.description = Timestamp for measurement/change
 channel-type.tibber.voltage.label = Voltage
 channel-type.tibber.voltage.description = Voltage on given Phase
 channel-type.tibber.tomorrow_prices.label = Prices for tomorrow as a JSON array
-channel-type.tibber.tomorrow_prices.description = JSON array of tuples startsAt,total, e.g. {["startsAt": "2022-09-10T00:00:00+02:00", "total": 5.332}, {"startsAt": ...}]}. See binding documantation for full example.
+channel-type.tibber.tomorrow_prices.description = JSON array of tuples startsAt,total, e.g. {["startsAt": "2022-09-10T00:00:00+02:00", "total": 5.332}, {"startsAt": ...}]}. See binding documentation for full example.
+channel-type.tibber.today_prices.label = Prices for today as a JSON array
+channel-type.tibber.today_prices.description = JSON array of tuples statsAt,total, e.g. {["startsAt": "2022-09-10T00:00:00+02:00", "total": 5.332}, {"startsAt": ...}]}. See binding documentation for full example.
```
jlaur, line 45:
```suggestion
channel-type.tibber.today_prices.description = JSON array of tuples startsAt,total, e.g. {["startsAt": "2022-09-10T00:00:00+02:00", "total": 5.332}, {"startsAt": ...}]}. See binding documentation for full example.
```

bundles/org.openhab.binding.tibber/src/main/resources/OH-INF/thing/thing-types.xml
jlaur: It's back. I see you fixed it also for tomorrow in i18n properties - please also fix it accordingly in this file.

bundles/org.openhab.binding.tibber/src/main/resources/OH-INF/i18n/tibber.properties
Markinus: Yes, I saw this and it's already fixed.

jlaur (approved):
LGTM

PR #14238, opened 2023-01-17
[rrd4j] Avoid IAE thrown if e.g. invalid start/end time given

cweitkamp:
- Avoid IAE thrown in constructor of `FetchRequest` called by `createFetchRequest` method (e.g. invalid start/end time given)
- Enable requesting future time ranges including boundary values via REST API

See https://github.com/rrd4j/rrd4j/blob/85183713741c648b88926ed3085c4dd87fd9016f/src/main/java/org/rrd4j/core/FetchRequest.java

### DIFF
--- a/bundles/org.openhab.persistence.rrd4j/src/main/java/org/openhab/persistence/rrd4j/internal/RRD4jPersistenceService.java
+++ b/bundles/org.openhab.persistence.rrd4j/src/main/java/org/openhab/persistence/rrd4j/internal/RRD4jPersistenceService.java
@@ -330,7 +330,7 @@ public class RRD4jPersistenceService implements QueryablePersistenceService {
                 ts += step;
             }
             return items;
-        } catch (IOException e) {
+        } catch (IllegalArgumentException | IOException e) {
             logger.warn("Could not query rrd4j database for item '{}': {}", itemName, e.getMessage());
             return List.of();
         } finally {
--- a/bundles/org.openhab.persistence.rrd4j/src/main/java/org/openhab/persistence/rrd4j/internal/RRD4jPersistenceService.java
+++ b/bundles/org.openhab.persistence.rrd4j/src/main/java/org/openhab/persistence/rrd4j/internal/RRD4jPersistenceService.java
@@ -280,7 +280,10 @@ public class RRD4jPersistenceService implements QueryablePersistenceService {
         }
 
         long start = 0L;
-        long end = filter.getEndDate() == null ? System.currentTimeMillis() / 1000
+        // set end to {@link Instant#MAX} instead of current timestamp to enable requesting future time ranges including
+        // boundary values via REST API
+        // see discussion in https://github.com/openhab/openhab-addons/pull/14238
+        long end = filter.getEndDate() == null ? Instant.MAX.getEpochSecond()
                 : filter.getEndDate().toInstant().getEpochSecond();
 
         try {
@@ -308,11 +311,19 @@ public class RRD4jPersistenceService implements QueryablePersistenceService {
                         start = end;
                     }
                 } else {
-                    throw new UnsupportedOperationException("rrd4j does not allow querys without a begin date, "
-                            + "unless order is descending and a single value is requested");
+                    throw new UnsupportedOperationException(
+                            "rrd4j does not allow querys without a begin date, unless order is descending and a single value is requested");
                 }
             } else {
                 start = filter.getBeginDate().toInstant().getEpochSecond();
+            }
+
+            // do not call method {@link RrdDb#createFetchRequest(ConsolFun, long, long, long)} if start > end to avoid
+            // an IAE to be thrown
+            if (start > end) {
+                logger.warn("Could not query rrd4j database for item '{}': start ({}) > end ({})", itemName, start,
+                        end);
+                return List.of();
             }
 
             FetchRequest request = db.createFetchRequest(getConsolidationFunction(db), start, end, 1);
@@ -330,7 +341,7 @@ public class RRD4jPersistenceService implements QueryablePersistenceService {
                 ts += step;
             }
             return items;
-        } catch (IllegalArgumentException | IOException e) {
+        } catch (IOException e) {
             logger.warn("Could not query rrd4j database for item '{}': {}", itemName, e.getMessage());
             return List.of();
         } finally {


bundles/org.openhab.persistence.rrd4j/src/main/java/org/openhab/persistence/rrd4j/internal/RRD4jPersistenceService.java
```diff
@@ -330,7 +330,7 @@ public Iterable<HistoricItem> query(FilterCriteria filter) {
                 ts += step;
             }
             return items;
-        } catch (IOException e) {
+        } catch (IllegalArgumentException | IOException e) {
```
jlaur, line 333: Wouldn't it be better to not call with invalid parameters? It seems we are simply passing an interval that doesn't make sense here, because `end` could then become smaller than `start`?
```java
long end = filter.getEndDate() == null ? System.currentTimeMillis() / 1000
                : filter.getEndDate().toInstant().getEpochSecond();
```

jlaur (changes requested):
Thanks for the fix. I have left one comment with a suggestion. Unless I'm mistaken this would fix the root cause rather than the symptom.

bundles/org.openhab.persistence.rrd4j/src/main/java/org/openhab/persistence/rrd4j/internal/RRD4jPersistenceService.java
cweitkamp: Yes, it would. Do you have a suggestion what we can do? Compare `start` and `end` against each other and switch values if `end` is less than `start`? Or skip querying the database?

jlaur: The equivalent of the current fix would be around line 317 to add:
```java
if (start > end) {
    logger.warn("Could not query rrd4j database for item '{}': start > end", itemName);
    return List.of();
}
```

i.e. avoid calling with invalid parameters. That would be the minimal fix.

Now, the question is which logic is actually desired for missing end date in the filter? For JDBC, missing end date is not interpreted as current date - the condition is simply left out of the query:
https://github.com/openhab/openhab-addons/blob/80094b5e9d293dfb73c81377325da13172d6baaf/bundles/org.openhab.persistence.jdbc/src/main/java/org/openhab/persistence/jdbc/internal/db/JdbcBaseDAO.java#L543-L556

I don't know what would be correct for this service.

cweitkamp: Me neither. That's the reason why I just added the catch for IAE.

jlaur: Hmm, the intended behavior seems undocumented: https://www.openhab.org/javadoc/latest/org/openhab/core/persistence/filtercriteria

But this also means that it's not documented that end date being null should be interpreted as current date. Since persistence currently only represents historical values (i.e. in the past), a filter with a future start date should not produce any results. So the end result should be the same no matter if we treat null as some distant future date and call, or if we return fast without any results. And I'm not even sure if it makes sense to log a warning.

Do you know the use-case - can you reproduce it, or did any user report this? It would be interesting knowing what lead to this.

cweitkamp: The exception has been found while testing https://github.com/openhab/openhab-webui/pull/1634. The REST API for querying historic item states provides a `boundary` parameter to return an additional value before and after the given time range. If one tries to fetch data for the current month from RRD4J persistence the IAE occures. Persistence Resource in OHC creates a `FilterCriteria` without an end date to get the value after the given time range.

https://github.com/openhab/openhab-core/blob/8f9dafd1d68a92248cd39414145175952d61cd46/bundles/org.openhab.core.io.rest.core/src/main/java/org/openhab/core/io/rest/core/internal/persistence/PersistenceResource.java#L346-L351

I have the slight feeling that the IAE in RRD4J can be forced by calling `PersistenceExtensions.minimumBetween()` or `PersistenceExtensions.maximumBetween()` methods from rules with parameters `begin > end`. OHC does not check that.

jlaur: It seems there is room for improvements in core also (at least documentation) - thanks for the link. I might pick this up when having a bit of time. Reading that code seems to confirm that `end == null` means "no end date" (see difference between these methods):
https://github.com/openhab/openhab-core/blob/dd756b8e2505c6cf9c7555ec7cb73965fd3a5328/bundles/org.openhab.core.persistence/src/main/java/org/openhab/core/persistence/extensions/PersistenceExtensions.java#L420-L451

So for this PR, perhaps the solution would be to treat no end date as `Instant.MAX.getEpochSecond()` rather than `System.currentTimeMillis().` Something like (line 283-284):

```java
long end = (filter.getEndDate() == null ? Instant.MAX : filter.getEndDate().toInstant()).getEpochSecond();
```

bundles/org.openhab.persistence.rrd4j/src/main/java/org/openhab/persistence/rrd4j/internal/RRD4jPersistenceService.java
```diff
@@ -308,13 +311,21 @@ public Iterable<HistoricItem> query(FilterCriteria filter) {
...
                 start = filter.getBeginDate().toInstant().getEpochSecond();
             }
 
+            // do not call method {@link RrdDb#createFetchRequest(ConsolFun, long, long, long)} if start > end to avoid
+            // an IAE to be thrown
+            if (start > end) {
+                logger.warn("Could not query rrd4j database for item '{}': start ({}) > end ({})", itemName, start,
+                        end);
+                return List.of();
+            }
```
jlaur, line 327: We could consider throwing `IllegalArgumentException` by ourselves here, since this would now be a bug on caller side, where previously it was an internal bug within this method. But let's leave something for future improvement also. 🙂

jlaur (approved):
Thanks for the fix and patience.

bundles/org.openhab.persistence.rrd4j/src/main/java/org/openhab/persistence/rrd4j/internal/RRD4jPersistenceService.java
cweitkamp: I incorporated the remarks and no longer catch the IAE thrown.

My feeling regarding IAE reproducible via `PersistenceExtension` methods have been confirmed in
https://github.com/openhab/openhab-webui/pull/1634#issuecomment-1397744206.

crnjan: @cweitkamp @jlaur - I was just testing this change and found [this](https://github.com/openhab/openhab-webui/pull/1649#issuecomment-1407621529) - while digging a bit, it seems internally `fetchData` uses something like
```
        int ptsCount = (int) ((fetchEnd - fetchStart) / arcStep + 1);
        long[] timestamps = new long[ptsCount];
```
and since now `fetchEnd` is set to MAX we get OOM. I tried (locally) changing back to `System.currentTimeMillis() / 1000` and OOM is gone. I can create a PR with the fix later today, please let me know what you think, thank you!

jlaur: I was actually afraid something like that could happen because of this very high value of MAX and not knowing the implementation on "the other side", but assumed this would be detected during testing. The logic seems correct IMHO, but probably implementation needs to be adapted. If you could propose a fix, that would be awesome.

PR #14240, opened 2023-01-17
[mqtt.homeassistant] Mark disabled by default components as advanced channels

ccutrer:
openHAB doesn't have the concept of "enabled by default", since _everything_ is technically "disabled" until you link it to an item. It seems devices use disabled by default on less common entities, such as zigbee2mqtt marking a binary_sensor as a "backup" method to an update entity for when an update is available, or a sensor as a "backup" method to a select entity. So in that case, just hiding these channels unless the user clicks "Show Advanced" seems to map best.


lolodomo (approved):
Technically, code looks good to me.

lolodomo:
@antroids : your review is welcome.

PR #14245, opened 2023-01-18
[evcc] Adjust to breaking API change & Add new channel

### LINKED ISSUE #14231: [evcc] problems with EVCC binding after upgrade to 3.4.1
Hello,

after the upgrade to 3.4.1 the evcc binding only works to a very limited extent. Control of charging settings from openHAB is no longer possible.
I have installed openHAB and evcc on the same rasperry using openHABian. The binding has worked very perfect since October 2022. but now I only see some read-only power-data from evcc.

## Expected Behavior
if I change items in openHAB, the changes should be applied synchronously to evcc and vice versa

## Current Behavior
it is not possible to change the items in openHAB. The charging settings are also shown wrong in openHAB.

## Possible Solution
maybe a problem with write-permission?

The charging settings is can not be changed. They show wrong values and can not be changed

```
Setpoint item=evcc_xx_loadpoint0_targetSoC minValue=5 maxValue=100 step=5
Setpoint item=evcc_xx_loadpoint0_minCurrent minValue=6 maxValue=96 step=2
Setpoint item=evcc_xx_loadpoint0_maxCurrent minValue=6 maxValue=96 step=2
Setpoint item=evcc_xx_loadpoint0_minSoC minValue=0 maxValue=100 step=5
Setpoint item=evcc_xx_loadpoint0_phases minValue=1 maxValue=3 step=2
```

## Context
I can use evcc only with evcc gui. My openHAB sitemap and openHAB rules do not work any longer.

## Your Environment

openHAB 3.4.1 - Release Build
Release = Raspbian GNU/Linux 11 (bullseye)
Kernel = Linux 5.15.84-v7+
Platform = Raspberry Pi 3 Model B Plus Rev 1.3
evcc v0.111.1

florian-h05:
Fixes #14231.

evcc changed the API with version 0.111.0, see https://github.com/evcc-io/evcc/discussions/5522.
This API change breaks the binding:

> - all indexes become 1-based (instead of 0-based for REST and 1-based for MQTT)
> - SoC becomes Soc

I have also added a new channel for the home battery's capacity which is a new value available over the API.

### DIFF
--- a/bundles/org.openhab.binding.evcc/src/main/java/org/openhab/binding/evcc/internal/EvccBindingConstants.java
+++ b/bundles/org.openhab.binding.evcc/src/main/java/org/openhab/binding/evcc/internal/EvccBindingConstants.java
@@ -31,7 +31,8 @@ public class EvccBindingConstants {
     public static final ThingTypeUID THING_TYPE_DEVICE = new ThingTypeUID(BINDING_ID, "device");
 
     // List of all Channel Type UIDs
-    public static final ChannelTypeUID CHANNEL_TYPE_UID_BATTERY_CAPACITY = new ChannelTypeUID(BINDING_ID, "batteryCapacity");
+    public static final ChannelTypeUID CHANNEL_TYPE_UID_BATTERY_CAPACITY = new ChannelTypeUID(BINDING_ID,
+            "batteryCapacity");
     public static final ChannelTypeUID CHANNEL_TYPE_UID_BATTERY_POWER = new ChannelTypeUID(BINDING_ID, "batteryPower");
     public static final ChannelTypeUID CHANNEL_TYPE_UID_BATTERY_SOC = new ChannelTypeUID(BINDING_ID, "batterySoC");
     public static final ChannelTypeUID CHANNEL_TYPE_UID_BATTERY_PRIORITY_SOC = new ChannelTypeUID(BINDING_ID,
--- a/bundles/org.openhab.binding.evcc/src/main/java/org/openhab/binding/evcc/internal/EvccBindingConstants.java
+++ b/bundles/org.openhab.binding.evcc/src/main/java/org/openhab/binding/evcc/internal/EvccBindingConstants.java
@@ -31,6 +31,7 @@ public class EvccBindingConstants {
     public static final ThingTypeUID THING_TYPE_DEVICE = new ThingTypeUID(BINDING_ID, "device");
 
     // List of all Channel Type UIDs
+    public static final ChannelTypeUID CHANNEL_TYPE_UID_BATTERY_CAPACITY = new ChannelTypeUID(BINDING_ID, "batteryCapacity");
     public static final ChannelTypeUID CHANNEL_TYPE_UID_BATTERY_POWER = new ChannelTypeUID(BINDING_ID, "batteryPower");
     public static final ChannelTypeUID CHANNEL_TYPE_UID_BATTERY_SOC = new ChannelTypeUID(BINDING_ID, "batterySoC");
     public static final ChannelTypeUID CHANNEL_TYPE_UID_BATTERY_PRIORITY_SOC = new ChannelTypeUID(BINDING_ID,
@@ -88,6 +89,7 @@ public class EvccBindingConstants {
             "vehicleTitle");
 
     // List of all Channel ids
+    public static final String CHANNEL_BATTERY_CAPACITY = "batteryCapacity";
     public static final String CHANNEL_BATTERY_POWER = "batteryPower";
     public static final String CHANNEL_BATTERY_SOC = "batterySoC";
     public static final String CHANNEL_BATTERY_PRIORITY_SOC = "batteryPrioritySoC";
--- a/bundles/org.openhab.binding.evcc/src/main/java/org/openhab/binding/evcc/internal/EvccBindingConstants.java
+++ b/bundles/org.openhab.binding.evcc/src/main/java/org/openhab/binding/evcc/internal/EvccBindingConstants.java
@@ -27,68 +27,6 @@ public class EvccBindingConstants {
 
     private static final String BINDING_ID = "evcc";
 
-    // List of all Thing Type UIDs
-    public static final ThingTypeUID THING_TYPE_DEVICE = new ThingTypeUID(BINDING_ID, "device");
-
-    // List of all Channel Type UIDs
-    public static final ChannelTypeUID CHANNEL_TYPE_UID_BATTERY_CAPACITY = new ChannelTypeUID(BINDING_ID,
-            "batteryCapacity");
-    public static final ChannelTypeUID CHANNEL_TYPE_UID_BATTERY_POWER = new ChannelTypeUID(BINDING_ID, "batteryPower");
-    public static final ChannelTypeUID CHANNEL_TYPE_UID_BATTERY_SOC = new ChannelTypeUID(BINDING_ID, "batterySoC");
-    public static final ChannelTypeUID CHANNEL_TYPE_UID_BATTERY_PRIORITY_SOC = new ChannelTypeUID(BINDING_ID,
-            "batteryPrioritySoC");
-    public static final ChannelTypeUID CHANNEL_TYPE_UID_GRID_POWER = new ChannelTypeUID(BINDING_ID, "gridPower");
-    public static final ChannelTypeUID CHANNEL_TYPE_UID_HOME_POWER = new ChannelTypeUID(BINDING_ID, "homePower");
-    public static final ChannelTypeUID CHANNEL_TYPE_UID_PV_POWER = new ChannelTypeUID(BINDING_ID, "pvPower");
-    public static final ChannelTypeUID CHANNEL_TYPE_UID_LOADPOINT_ACTIVE_PHASES = new ChannelTypeUID(BINDING_ID,
-            "activePhases");
-    public static final ChannelTypeUID CHANNEL_TYPE_UID_LOADPOINT_CHARGE_CURRENT = new ChannelTypeUID(BINDING_ID,
-            "chargeCurrent");
-    public static final ChannelTypeUID CHANNEL_TYPE_UID_LOADPOINT_CHARGE_DURATION = new ChannelTypeUID(BINDING_ID,
-            "chargeDuration");
-    public static final ChannelTypeUID CHANNEL_TYPE_UID_LOADPOINT_CHARGE_POWER = new ChannelTypeUID(BINDING_ID,
-            "chargePower");
-    public static final ChannelTypeUID CHANNEL_TYPE_UID_LOADPOINT_CHARGE_REMAINING_DURATION = new ChannelTypeUID(
-            BINDING_ID, "chargeRemainingDuration");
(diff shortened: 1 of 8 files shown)


bundles/org.openhab.binding.evcc/src/main/java/org/openhab/binding/evcc/internal/EvccHandler.java
```diff
@@ -85,7 +85,7 @@ public void handleCommand(ChannelUID channelUID, Command command) {
                 return;
             }
             String channelIdWithoutGroup = channelUID.getIdWithoutGroup();
-            int loadpoint = Integer.parseInt(groupId.substring(9));
+            int loadpoint = Integer.parseInt(groupId.substring(9)) + 1;
```
jlaur, line 88: Is this fix backwards compatible, or do we need to treat this according to actual version?

jlaur: Is it always possible to upgrade evcc version, or can you imagine any users being stock with older versions? Perhaps it could be asked in the forum, or initially we could assume version 0.111.1 as first version needing this change?

mj6517: I'm relatively sure that version 0.111.1 is the first version needs the change. 
I'm afraid, but I think additional changes are necessary, my environment is still not working correct, problems with Battery SoC, Target SoC and Minimum SoC.

mstormi:
> In the last, evcc numbered its loadpoints 0, 1, ...
>  Now it is 1, 2, ...

Seems this is only for writing. Reading via `/api/state` still exposes loadpoints[0].

Are you sure this is an intentional API change or rather an EVCC bug ? Did you ask over at their forum ?

florian-h05:
Reading though https://github.com/evcc-io/evcc/discussions/5522, it seems that this change was intentional. It explains as well why the SoCs aren’t working anymore.

bundles/org.openhab.binding.evcc/src/main/java/org/openhab/binding/evcc/internal/EvccHandler.java
florian-h05: Looking at what changed in the last time at evcc, it is `0.111.0` that brings API breaking changes. `0.111.1` is just a small patch release.

> Is it always possible to upgrade evcc version, or can you imagine any users being stock with older versions?

I can’t imagine anyone being stock on older version, because I don’t see a reason why one should. evcc requires only very little resources and I can’t imagine a technical reason why one shouldn’t be able to upgrade.

florian-h05: > I'm afraid, but I think additional changes are necessary, my environment is still not working correct, problems with Battery SoC, Target SoC and Minimum SoC.

Should work again now (at least in my tests it works). I'll try to provide an artifact tomorrow.

mj6517: thank you very much, I will test it right away

florian-h05:
https://github.com/evcc-io/evcc/pull/5649 probably brings more breaking changes.

bundles/org.openhab.binding.evcc/src/main/java/org/openhab/binding/evcc/internal/EvccHandler.java
mj6517: > Should work again now (at least in my tests it works). I'll try to provide an artifact tomorrow.

I assume you are working on that bigger API change?

florian-h05: I’ve actually finished work when I wrote that comment, but haven’t had the time (yet) to provide an artifact.

florian-h05: @mj6517  I created an artifact for 3.4.x, you can download it from WeTransfer: https://we.tl/t-NP2vaXWFdm.

mj6517: thank's a lot. I've implemented and tested it. Everything works like before. prefect!

florian-h05: Thanks for testing!

florian-h05:
@jlaur Since this has been verified to fix the issue, this is ready for merging from my side!

bundles/org.openhab.binding.evcc/src/main/java/org/openhab/binding/evcc/internal/EvccHandler.java
jlaur: I had a brief look at the code, and it seems you don't have the version information available (currently), so I'll resolve this comment and we'll rely on the documentation asking users to upgrade firmware.

bundles/org.openhab.binding.evcc/src/main/java/org/openhab/binding/evcc/internal/EvccBindingConstants.java
```diff
@@ -31,6 +31,8 @@ public class EvccBindingConstants {
     public static final ThingTypeUID THING_TYPE_DEVICE = new ThingTypeUID(BINDING_ID, "device");
 
     // List of all Channel Type UIDs
+    public static final ChannelTypeUID CHANNEL_TYPE_UID_BATTERY_CAPACITY = new ChannelTypeUID(BINDING_ID,
+            "batteryCapacity");
```
jlaur, line 35: Just curious, why not use the already defined constant here?
```suggestion
            CHANNEL_BATTERY_CAPACITY);
```

bundles/org.openhab.binding.evcc/src/main/java/org/openhab/binding/evcc/internal/EvccHandler.java
```diff
@@ -257,6 +258,7 @@ public void dispose() {
     private void createChannelsGeneral() {
         final String channelGroup = "general";
         if (batteryConfigured) {
+            createChannel(CHANNEL_BATTERY_CAPACITY, channelGroup, CHANNEL_TYPE_UID_BATTERY_CAPACITY, "Number:Energy");
```
jlaur, line 261: Just a comment for future consideration: Instead of dynamically creating channels, they could be statically declared in XML and then removed when not needed. See for example #13275. The reason for doing so would be simplification.

bundles/org.openhab.binding.evcc/src/main/resources/OH-INF/i18n/evcc.properties
```diff
@@ -104,8 +106,3 @@ channel-type.evcc.vehicleSoC.label = Vehicle SoC
 channel-type.evcc.vehicleSoC.description = Current State of Charge of EV
 channel-type.evcc.vehicleTitle.label = Vehicle title
 channel-type.evcc.vehicleTitle.description = Name of EV
-
-# channel types
-
-offline.configuration-error.no-host = No host configured
```
jlaur, line 110: Why removed? Seems still referenced here:
https://github.com/openhab/openhab-addons/blob/fe9f0583b5d8dcc88d40c9517f63f327b61261b8/bundles/org.openhab.binding.evcc/src/main/java/org/openhab/binding/evcc/internal/EvccHandler.java#L197-L198

bundles/org.openhab.binding.evcc/src/main/resources/OH-INF/i18n/evcc.properties
```diff
@@ -104,8 +106,3 @@ channel-type.evcc.vehicleSoC.label = Vehicle SoC
...
-offline.communication-error.request-failed = Request failed
```
jlaur, line 111: And here:
https://github.com/openhab/openhab-addons/blob/fe9f0583b5d8dcc88d40c9517f63f327b61261b8/bundles/org.openhab.binding.evcc/src/main/java/org/openhab/binding/evcc/internal/EvccHandler.java#L228-L229

bundles/org.openhab.binding.evcc/src/main/java/org/openhab/binding/evcc/internal/EvccBindingConstants.java
florian-h05: Good idea!

florian-h05: Done, see https://github.com/openhab/openhab-addons/pull/14245/commits/c9811852f2db1f6b45e94a315311628460170ec4.

bundles/org.openhab.binding.evcc/src/main/resources/OH-INF/i18n/evcc.properties
florian-h05: I regenerated the file using `mvn i18n:generate-default-translations` and this section was removed then, I obviously didn't notice. I've reverted this, see https://github.com/openhab/openhab-addons/pull/14245/commits/8142ecd45648689ad88bfde47a2687ec83c584b9.

bundles/org.openhab.binding.evcc/src/main/resources/OH-INF/i18n/evcc.properties
florian-h05: See above.

bundles/org.openhab.binding.evcc/src/main/resources/OH-INF/i18n/evcc.properties
jlaur: I wonder why it did that - does it happen if you run it again? If yes, can you try to fix the header (line 108) to:

```
# thing status descriptions
```

Maybe the tool is using the header, which is currently wrong, so that could explain it?

florian-h05: Hmm, that's weird. If I run again, it doesn't change anything. I guess I probably did something else with this file...
Anyway, I'll correct that header.

jlaur (approved):
LGTM

florian-h05:
@jlaur Do you think this is something to include in a 3.4.2 Patch release (if there’s one)?

jlaur:
@florian-h05 - ideally it should then have been a pure bugfix PR without introduction of a new channel. But since I understand that the binding is quite broken for all users having upgraded to evcc 0.111, I agree it's a candiate. @openhab/add-ons-maintainers, okay to cherry-pick this one?

jlaur:
@lolodomo - would you agree to cherry-pick this?

jlaur:
@kaikreuzer - if you agree, please cherry-pick this commit, f808edf725bf52942245ba2033b529a5e0b1ba3b, and 988fde1abadcadc13b82bb618912f033a86aa7f2 as well (from #14381) into the 3.4.x branch before starting a build. Thanks!

kaikreuzer:
Sure @jlaur, I have cherry-picked those two fixes to `3.4.x`.

Issue #14246, opened 2023-01-18
Binding support for Tesla T-Smart Smart Immersion Heater / Thermostat


jamesmelville:
I am working on a binding to support the Tesla T-Smart Smart Immersion Heater / Thermostat device https://www.tsmart.co.uk/.

The device allows monitoring and control over the temperature and heating of water in a cylinder.

The build is derived from the API docs shared here https://community.home-assistant.io/t/telsa-t-smart-immersion-heater/345853 and some inspiration from https://github.com/britkat1980/tsmart.

I've already written a good proportion of the binding code with the following features

- Monitor Device
    - Temperature
    - Setpoint
    - Mode
    - Relay State
    - Power
- Set Setpoint / Mode
- Discovery

I'm just in the process of tidying the code with a view to committing it and issuing a PR soon (and probably need to work out how to get my environment and the PR to java 17/ OH 4.0)- wondering if anyone would help with testing / feedback etc.? Please get in contact.

PR #14247, opened 2023-01-18
[tsmart] Binding support for Tesla T-Smart Smart Immersion Heater / Thermostat

jamesmelville:
# Title

Binding to support the Tesla T-Smart Smart Immersion Heater / Thermostat device https://www.tsmart.co.uk/. (No relation to Tesla vehicles).
The device allows monitoring and control over the temperature and heating of water in a cylinder.

# Description

Introduce a new binding to support the Tesla T-Smart device.

Mentioned in an issue and on the community
https://github.com/openhab/openhab-addons/issues/14246
https://community.openhab.org/t/tesla-t-smart-smart-immersion-heater-thermostat-binding/143521


bundles/org.openhab.binding.tsmart/README.md
```diff
@@ -0,0 +1,53 @@
+# Tesla T-Smart Binding
+
+This binding is designed to connect to the Tesla T-Smart Smart Immersion Heater / Thermostat device https://www.tsmart.co.uk/.
+
+The device allows monitoring and control over the temperature and heating of water in a water cylinder.
+
+The build is derived from the API docs shared here https://community.home-assistant.io/t/telsa-t-smart-immersion-heater/345853 and some inspiration from https://github.com/britkat1980/tsmart.
```
lsiepel, line 7: You can remove this from the user documentation, if needed you can add it to the start post of the PR or create a developer/usefull links topic at the end of the docs.

bundles/org.openhab.binding.tsmart/README.md
```diff
@@ -0,0 +1,53 @@
...
+| Channel         | Type               | Read/Write | Description                                |
+|-----------------|--------------------|------------|--------------------------------------------|
+| `power`           | Switch             | RW         | Power for the thermostat                   |
+| `mode`            | String             | RW         | Operating mode of the thermostat           |
+| `setpoint`        | Number:Temperature | RW         | Setpoint of the hot water temperature      |
+| `temperature`     | Number:Temperature | R          | Measured temperature of the water cylinder |
+| `temperatureHigh` | Number:Temperature | R          | Temperature of the high thermostat         |
+| `temperatureLow`  | Number:Temperature | R          | Temperature of the low thermostat          |
+| `relay`           | Switch             | R          | Relay state of the immersion heater        |
+| `smartState`           | String             | R          | State of the thermostat's Smart Mode     |
```
lsiepel, line 35: Minor space issue not aligned pipes

bundles/org.openhab.binding.tsmart/README.md
```diff
@@ -0,0 +1,53 @@
...
+|-----------------|---------|---------------------------------------|---------|----------|----------|
+| hostname        | text    | Hostname or IP address of the device  | N/A     | yes      | no       |
+| refreshInterval | integer | Interval the device is polled in sec. | 60      | no       | yes      |
+
+## Channels
+
+| Channel         | Type               | Read/Write | Description                                |
+|-----------------|--------------------|------------|--------------------------------------------|
+| `power`           | Switch             | RW         | Power for the thermostat                   |
+| `mode`            | String             | RW         | Operating mode of the thermostat           |
```
lsiepel, line 29: Either the docs is wrong or the thing-types.xml as the docs say it is RW and the xml has readonly=true

bundles/org.openhab.binding.tsmart/src/main/resources/OH-INF/addon/addon.xml
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<addon:addon xmlns:addon="https://openhab.org/schemas/addon/v1.0.0"
+	xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" id="tsmart"
+	xsi:schemaLocation="https://openhab.org/schemas/addon/v1.0.0 https://openhab.org/schemas/addon-1.0.0.xsd">
+
+	<type>binding</type>
+	<name>T-Smart Binding</name>
+	<description>This is the binding for Tesla T-Smart Immersion Heater Thermostats.</description>
```
lsiepel, line 8:
```suggestion
    <connection>local</connection>
```

bundles/org.openhab.binding.tsmart/src/main/resources/OH-INF/i18n/tsmart.properties
```diff
@@ -0,0 +1,3 @@
+# FIXME: please add all English translations to this file so the texts can be translated using Crowdin
```
lsiepel, line 1: Please run mvn i18n command

bundles/org.openhab.binding.tsmart/src/main/java/org/openhab/binding/tsmart/internal/TSmartDiscoveryService.java
```diff
@@ -0,0 +1,94 @@
...
+    }
+
+    /**
+     * Start Discovery of T-Smart devices
+     */
+    @Override
+    protected void startScan() {
+        TSmartUDPListener.startDiscovery(this);
+
+        try {
```
lsiepel, line 57:
```suggestion
            logger.debug("Started discovery by sending UDP packet to broadcast address");
```

lsiepel (changes requested):
Thanks for the contribution. Overall code looks good, there are some comments left that needs to be fixed. I leave the rest to addons-maintainers.

bundles/org.openhab.binding.tsmart/pom.xml
```diff
@@ -0,0 +1,17 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns="http://maven.apache.org/POM/4.0.0"
+  xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd">
+
+  <modelVersion>4.0.0</modelVersion>
+
+  <parent>
+    <groupId>org.openhab.addons.bundles</groupId>
+    <artifactId>org.openhab.addons.reactor.bundles</artifactId>
+    <version>4.0.0-SNAPSHOT</version>
```
wborn, line 10:
```suggestion
    <version>4.1.0-SNAPSHOT</version>
```

bundles/org.openhab.binding.tsmart/src/main/resources/OH-INF/thing/thing-types.xml
```diff
@@ -0,0 +1,102 @@
...
+	xmlns:thing="https://openhab.org/schemas/thing-description/v1.0.0"
+	xsi:schemaLocation="https://openhab.org/schemas/thing-description/v1.0.0 https://openhab.org/schemas/thing-description-1.0.0.xsd">
+
+	<thing-type id="tsmart">
+
+		<label>Tesla T-Smart</label>
+		<description>Tesla T-Smart Immersion Heater Thermostat</description>
+
+		<channels>
+			<channel id="power" typeId="power"/>
```
jlaur, line 13: System channel type can be used:
```suggestion
			<channel id="power" typeId="system.power"/>
```

bundles/org.openhab.binding.tsmart/src/main/resources/OH-INF/thing/thing-types.xml
```diff
@@ -0,0 +1,102 @@
...
+				<default>60</default>
+			</parameter>
+		</config-description>
+	</thing-type>
+
+	<channel-type id="power">
+		<item-type>Switch</item-type>
+		<label>Power</label>
+	</channel-type>
+
```
jlaur, line 50: See comment above.
```suggestion
```

bundles/org.openhab.binding.tsmart/src/main/resources/OH-INF/thing/thing-types.xml
```diff
@@ -0,0 +1,102 @@
...
+		<label>Tesla T-Smart</label>
+		<description>Tesla T-Smart Immersion Heater Thermostat</description>
+
+		<channels>
+			<channel id="power" typeId="power"/>
+			<channel id="temperature" typeId="temperature">
+				<label>Temperature</label>
+				<description>Measured temperature across the sensors</description>
+			</channel>
+			<channel id="temperatureHigh" typeId="temperature">
```
jlaur, line 18: Please adapt all channel ids and channel type ids according to [naming convention]:(https://www.openhab.org/docs/developer/guidelines.html#naming-convention)
```suggestion
			<channel id="temperature-high" typeId="temperature">
```

bundles/org.openhab.binding.tsmart/src/main/resources/OH-INF/thing/thing-types.xml
```diff
@@ -0,0 +1,102 @@
...
+		<description>Temperature of water in cylinder sensed by thermostat</description>
+		<category>Temperature</category>
+		<state readOnly="true" pattern="%.0f %unit%"/>
+	</channel-type>
+
+	<channel-type id="setpoint">
+		<item-type>Number:Temperature</item-type>
+		<label>Setpoint</label>
+		<description>Setpoint currently set on device</description>
+		<category>Heating</category>
```
jlaur, line 68: You could add semantic tags also:
```suggestion
		<category>Heating</category>
		<tags>
			<tag>Setpoint</tag>
			<tag>Temperature</tag>
		</tags>
```

bundles/org.openhab.binding.tsmart/src/main/java/org/openhab/binding/tsmart/internal/TSmartHandler.java
```diff
@@ -0,0 +1,238 @@
...
+        } else {
+            switch (channelUID.getId()) {
+                case CHANNEL_POWER:
+                    if (command instanceof OnOffType) {
+                        setPower(OnOffType.ON.equals(command));
+                    }
+                    break;
+                case CHANNEL_SETPOINT:
+                    if (command instanceof QuantityType) {
+                        setSetpoint((QuantityType<?>) command);
```
jlaur, line 79:
```suggestion
                    if (command instanceof QuantityType<?> quantityCommand) {
                        setSetpoint(quantityCommand);
```

bundles/org.openhab.binding.tsmart/src/main/java/org/openhab/binding/tsmart/internal/TSmartHandler.java
```diff
@@ -0,0 +1,238 @@
...
+                    }
+                    break;
+                case CHANNEL_MODE:
+                    if (command instanceof StringType) {
+                        setMode((StringType) command);
```
jlaur, line 84:
```suggestion
                    if (command instanceof StringType stringCommand) {
                        setMode(stringCommand);
```

bundles/org.openhab.binding.tsmart/src/main/java/org/openhab/binding/tsmart/internal/TSmartHandler.java
```diff
@@ -0,0 +1,238 @@
...
+
+    /**
+     * Initialise ThingHandler - start listening for UDP responses,
+     * and sending status requests on the configured frequency.
+     */
+    @Override
+    public void initialize() {
+        config = getConfigAs(TSmartConfiguration.class);
+        try {
+            InetAddress destination = InetAddress.getByName(config.hostname);
```
jlaur, line 99: Please move this to the scheduled job. `initialize()` is expected to return fast, so it should not make network calls.

bundles/org.openhab.binding.tsmart/src/main/java/org/openhab/binding/tsmart/internal/TSmartHandler.java
```diff
@@ -0,0 +1,238 @@
...
+     * Close down ThingHandler - cease sending UDP requests
+     * and listening for UDP responses for this device.
+     */
+    @Override
+    public void dispose() {
+        InetAddress destination = destAddr;
+        if (destination != null) {
+            TSmartUDPListener.removeHandler(destination);
+        }
+        if (statusRefreshJob != null) {
```
jlaur, line 124: I would expect null annotation warnings from this:
```suggestion
        ScheduledFuture<?> statusRefreshJob = this.statusRefreshJob;
        if (statusRefreshJob != null) {
```

bundles/org.openhab.binding.tsmart/src/main/java/org/openhab/binding/tsmart/internal/TSmartHandler.java
```diff
@@ -0,0 +1,238 @@
...
+            statusRefreshJob.cancel(true);
```
jlaur, line 125:
```suggestion
            statusRefreshJob.cancel(true);
            this.statusRefreshJob = null;
```

bundles/org.openhab.binding.tsmart/src/main/java/org/openhab/binding/tsmart/internal/TSmartHandler.java
```diff
@@ -0,0 +1,238 @@
...
+        updateState(CHANNEL_TEMPERATURE_HIGH, high);
+        updateState(CHANNEL_TEMPERATURE_LOW, low);
+        updateState(CHANNEL_TEMPERATURE, avg);
+        updateState(CHANNEL_RELAY, OnOffType.from(buffer[9] == 0x1));
+        updateState(CHANNEL_MODE, getMode(buffer[6]));
+        updateState(CHANNEL_SMART_STATE, getSmartState(buffer[10]));
+    }
+
+    private void requestStatusRefresh() {
+
```
jlaur, line 155:
```suggestion
```

bundles/org.openhab.binding.tsmart/src/main/java/org/openhab/binding/tsmart/internal/TSmartHandler.java
```diff
@@ -0,0 +1,238 @@
...
+        sendUDPPacket(new byte[] { (byte) 0xF1, (byte) 0x00, (byte) 0x00 });
+
+        unRespondedRequests++;
+        if (unRespondedRequests > 5) {
+            updateStatus(ThingStatus.OFFLINE);
+        }
+    }
+
+    private QuantityType<Temperature> getTemperature(byte firstByte, byte secondByte) {
+
```
jlaur, line 165:
```suggestion
```

bundles/org.openhab.binding.tsmart/src/main/java/org/openhab/binding/tsmart/internal/TSmartDiscoveryService.java
```diff
@@ -0,0 +1,95 @@
...
+ * @author James Melville - Initial contribution
+ */
+@NonNullByDefault
+@Component(service = DiscoveryService.class, configurationPid = "discovery.tsmart")
+public class TSmartDiscoveryService extends AbstractDiscoveryService {
+
+    private final Logger logger = LoggerFactory.getLogger(TSmartDiscoveryService.class);
+
+    public TSmartDiscoveryService() {
+        super(Collections.unmodifiableSet(new HashSet<>(Arrays.asList(THING_TYPE_T_SMART))), 60, false);
```
jlaur, line 47:
```suggestion
        super(Set.of(THING_TYPE_T_SMART), 60, false);
```

bundles/org.openhab.binding.tsmart/src/main/java/org/openhab/binding/tsmart/internal/TSmartDiscoveryService.java
```diff
@@ -0,0 +1,95 @@
...
+    @Override
+    protected void startScan() {
+        TSmartUDPListener.startDiscovery(this);
+
+        try {
+            logger.debug("Started discovery by sending UDP packet to broadcast address");
+            new TSmartUDPUtils().sendUDPPacket(InetAddress.getByName("255.255.255.255"),
+                    new byte[] { (byte) 0x01, (byte) 0x00, (byte) 0x00 });
+        } catch (UnknownHostException e) {
+
```
jlaur, line 62: Consider to log as debug and include `e.getMessage()`.

jlaur (changes requested):
Thank you for your contribution! And sorry for the long delay. I have provided some feedback from my first review round.

bundles/org.openhab.binding.tsmart/README.md
```diff
@@ -0,0 +1,51 @@
...
+
+| Name            | Type    | Description                           | Default | Required | Advanced |
+|-----------------|---------|---------------------------------------|---------|----------|----------|
+| hostname        | text    | Hostname or IP address of the device  | N/A     | yes      | no       |
+| refreshInterval | integer | Interval the device is polled in sec. | 60      | no       | yes      |
+
+## Channels
+
+| Channel         | Type               | Read/Write | Description                                |
+|-----------------|--------------------|------------|--------------------------------------------|
```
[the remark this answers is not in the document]
jamesmelville: I removed the backticks around the channel names instead, that seemed common across recent readmes

bundles/org.openhab.binding.tsmart/src/main/java/org/openhab/binding/tsmart/internal/TSmartHandler.java
```diff
@@ -0,0 +1,240 @@
...
+    private void startup() {
+        try {
+            InetAddress destination = InetAddress.getByName(config.hostname);
+
+            destAddr = destination;
+
+            TSmartUDPListener.addHandler(destination, this);
+
+            updateStatus(ThingStatus.UNKNOWN);
```
jlaur, line 110: This must be moved back to `initialize()` because otherwise it most likely won't be set before leaving that method.

bundles/org.openhab.binding.tsmart/src/main/java/org/openhab/binding/tsmart/internal/TSmartHandler.java
```diff
@@ -0,0 +1,240 @@
...
+
+            statusRefreshJob = scheduler.scheduleWithFixedDelay(this::requestStatusRefresh, 0, config.refreshInterval,
+                    TimeUnit.SECONDS);
+        } catch (UnknownHostException e) {
+            updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.COMMUNICATION_ERROR, e.getMessage());
```
jlaur, line 115: In this case the binding won't be able to recover automatically. A real-life scenario could be after a power outage if openHAB is faster to start than the router, for example. See also #11891 which is almost exactly the same scenario. What I did here was to simply include the initialization into the scheduled job.

jlaur (changes requested):
While checking comment resolutions, I found two new things.

jamesmelville:
> While checking comment resolutions, I found two new things.

Thanks Jacob, sorry I should have marked this PR as draft as I didn't finish working on updates to your comments. I think I've resolved everything now.

bundles/org.openhab.binding.tsmart/src/main/java/org/openhab/binding/tsmart/internal/TSmartHandler.java
```diff
@@ -0,0 +1,262 @@
...
+
+    /**
+     * Initialise ThingHandler - start listening for UDP responses,
+     * and sending status requests on the configured frequency.
+     */
+    @Override
+    public void initialize() {
+        config = getConfigAs(TSmartConfiguration.class);
+        TSmartConfiguration thisConfig = config;
+        if (thisConfig == null || thisConfig.hostname.isBlank() || thisConfig.refreshInterval <= 0) {
```
jlaur, line 100: With a slight change you can avoid config being null:
```suggestion
        TSmartConfiguration config = getConfigAs(TSmartConfiguration.class);
        this.config = config;
        if (thisConfig.hostname.isBlank() || thisConfig.refreshInterval <= 0) {
```

bundles/org.openhab.binding.tsmart/src/main/java/org/openhab/binding/tsmart/internal/TSmartHandler.java
```diff
@@ -0,0 +1,262 @@
...
+            updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.CONFIGURATION_ERROR);
```
jlaur, line 101: Perhaps you could check the two parameters individually and provide a thing status description text instructing the user which one is invalid?

See for example:
https://github.com/openhab/openhab-addons/blob/118afa31fd1de31c1e77bd7242c75438080fbff6/bundles/org.openhab.binding.energidataservice/src/main/java/org/openhab/binding/energidataservice/internal/handler/EnergiDataServiceHandler.java#L120-L145

and:
https://github.com/openhab/openhab-addons/blob/118afa31fd1de31c1e77bd7242c75438080fbff6/bundles/org.openhab.binding.energidataservice/src/main/resources/OH-INF/i18n/energidataservice.properties#L93-L97

bundles/org.openhab.binding.tsmart/src/main/java/org/openhab/binding/tsmart/internal/TSmartUDPListener.java
```diff
@@ -0,0 +1,143 @@
...
+     * Maintain a mapping between a device address and the openHAB handler
+     * in order to route any received UDP packets to the correct handler.
+     * This method adds the address and ThingHandler to the map.
+     *
+     * @param addr Network address of device
+     * @param handler Instance of ThingHander
+     */
+    public static void addHandler(InetAddress addr, TSmartHandler handler) {
+        if (handlerAddressMap.containsKey(addr)) {
+            // throw error
```
jlaur, line 63: That would probably make sense to fail fast on duplicate subscriptions.

bundles/org.openhab.binding.tsmart/src/main/java/org/openhab/binding/tsmart/internal/TSmartUDPListener.java
```diff
@@ -0,0 +1,143 @@
...
+
+    /**
+     * Maintain a mapping between a device address and the openHAB handler
+     * in order to route any received UDP packets to the correct handler.
+     * This method removes the mapping.
+     *
+     * @param addr Network address of device
+     */
+    public static void removeHandler(InetAddress addr) {
+        if (handlerAddressMap.containsKey(addr)) {
```
jlaur, line 78: This check is not needed. `remove` will return null if nothing was removed.

bundles/org.openhab.binding.tsmart/src/main/java/org/openhab/binding/tsmart/internal/TSmartUDPListener.java
```diff
@@ -0,0 +1,143 @@
...
+                }
+            }
+            socket.close();
+            listener = null;
+            discoService = null;
+        } catch (SocketTimeoutException ex) {
+            logger.debug("Timeout error: {}", ex.getMessage());
+        } catch (IOException ex) {
+            logger.debug("Client error: {}", ex.getMessage());
+        }
```
jlaur, line 137: Should this be propagated back to the handlers, so they could take appropriate actions? For example, a thing handler might want to set status to offline with communication error.

bundles/org.openhab.binding.tsmart/src/main/java/org/openhab/binding/tsmart/internal/TSmartDiscoveryService.java
```diff
@@ -0,0 +1,93 @@
...
+     * @param buffer Packet bytes
+     */
+    public void handleDiscoveryResponse(InetAddress addr, byte[] buffer) {
+        String name = new String(Arrays.copyOfRange(buffer, 9, 20));
+        String id = String.format("%02X%02X%02X%02X", buffer[8], buffer[7], buffer[6], buffer[5]);
+
+        Map<String, Object> properties = new HashMap<>(3);
+        properties.put("hostname", addr.getHostAddress());
+        properties.put("id", id);
+        properties.put("name", name);
```
jlaur, line 88: Consider adding constants for these keys.

I recognize `hostname` as a configuration parameter. Are `id` and `name` properties? Will they also be populated when creating the thing manually, or only through discovery?

bundles/org.openhab.binding.tsmart/src/main/java/org/openhab/binding/tsmart/internal/TSmartHandler.java
```diff
@@ -0,0 +1,262 @@
...
+        } else {
+            updateStatus(ThingStatus.UNKNOWN);
+            statusRefreshJob = scheduler.scheduleWithFixedDelay(this::requestStatusRefresh, 0,
+                    thisConfig.refreshInterval, TimeUnit.SECONDS);
+        }
+    }
+
+    private synchronized boolean initializeProperties() {
+        if (propertiesInitializedSuccessfully) {
```
jlaur, line 110: The method name and instance variable name seems a bit misleading. As I'm reading the method it is responsible for subscribing to UDP events, not initializing properties?

jlaur (changes requested):
Now fully reviewed. Here are my findings from this last look.

bundles/org.openhab.binding.tsmart/src/main/java/org/openhab/binding/tsmart/internal/TSmartDiscoveryService.java
jamesmelville: Sorry, I wrote this code a long time ago and my mind is not fresh! Now I look into it discovery isn't working for me, so will debug my environment and get it working before any resubmission.

- Yes, hostname is the configuration parameter.
- Name is user entered in the device's app, so I figured it should be the default label for a new device.
- Id is fixed and derived from the MAC address, hence why I used it as the representation property. But I assume that wouldn't work properly because I'm not setting the property in normal usage.

I think the best fix for this is setting the name and id properties in runtime, outside of the discovery process.

bundles/org.openhab.binding.tsmart/src/main/java/org/openhab/binding/tsmart/internal/TSmartHandler.java
jamesmelville: Agree, the code is resolving the hostname (if a hostname and not plain IP, and creating a UDP listener / adding an address to listen to). Will resolve

bundles/org.openhab.binding.tsmart/src/main/java/org/openhab/binding/tsmart/internal/TSmartUDPListener.java
jamesmelville: Agree - these should both just mark all Things as offline because the UDP listener has failed. Will implement

bundles/org.openhab.binding.tsmart/src/main/java/org/openhab/binding/tsmart/internal/TSmartDiscoveryService.java
jlaur: Wrt label, it's already provided below with `.withLabel(name)`, so as a property I'm not sure if relevant.

Mostly it's relevant to provide properties to feel the configuration parameters where this is possible. For the rest, you are right that it should be populated by the handler itself so that it will also work for manual configuration and even file-based configuration.

bundles/org.openhab.binding.tsmart/src/main/resources/OH-INF/thing/thing-types.xml
```diff
@@ -0,0 +1,105 @@
...
+	<channel-type id="temperature-setpoint">
+		<item-type>Number:Temperature</item-type>
+		<label>Setpoint</label>
+		<description>Setpoint currently set on device</description>
+		<category>Heating</category>
+		<tags>
+			<tag>Setpoint</tag>
+			<tag>Temperature</tag>
+		</tags>
+		<state min="15" max="70" step="5" readOnly="false" pattern="%.0f %unit%"/>
```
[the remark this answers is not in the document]
jamesmelville: No, it'll be celsius only. Is your suggestion that I should I set the pattern to celsius, and this will enforce the limit correctly for an item set to another unit? I've tried to follow all the recent enhancement to UoM but have to admit I hadn't thought about this scenario!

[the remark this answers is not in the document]
jlaur: No, my only point is that I don't think `min`, `max` and `step` will take unit into consideration, but tbh, I'm not really sure. You could probably easily test that by changing the item's unit to °F and trying to go below/above defined boundaries?

If in fact UoM is not considered, you would either have to provide them dynamically (which is somewhat more complex) or set the boundaries to the lowest/highest allowed value in any of those units. So max would be 158 because 70 °C is 158 °F. Probably it can be assumed that no one would configure this to be in Kelvin. 😉

lolodomo:
@jlaur : does it remain something to change ?

lolodomo:
Build is failing in TSmartHandler => thisConfig

jlaur:
> does it remain something to change ?

It seems this one slipped. I just need to check that my last comments are addressed, but unfortunately don't have time today.

jlaur:
> > does it remain something to change ?
> 
> It seems this one slipped. I just need to check that my last comments are addressed, but unfortunately don't have time today.

Actually it's still waiting for comments to be resolved and build to be fixed. FYI @jamesmelville

bundles/org.openhab.binding.tsmart/src/main/resources/OH-INF/thing/thing-types.xml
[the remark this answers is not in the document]
jamesmelville: I'm not sure where to go with this, and don't have too much time to invest. This is what I know:
1. This is a water heater, there's a bigger problem that openhab being confused if we're looking at 15°F or 158°C :)
2. min/max/step in MainUI reflect the configured unit as you thought
3. I think the device is only sold in the UK so 90%+ of users will work in °C
4. Device resets to 15°C if I try to go below it (e.g. 35°F).
5. Device goes to 70°C if I try to go to negative °C, (e.g. 15°F) - suspect this is conversion to bytes, needs looking into.
6. Device can be set to 10°C, just not via the setpoint, only via travel mode.
7. Device doesn't need steps of 5°C, appears to respect a step of 0.1°C. (I did this as mirroring the companion app)
8. I looked at some other thermostats (Nest, Tado, Wiser). All configure a setpoint as %unit%, only one sets a step of 0.1, none have min/max.

I think I should do the following (~mirrors other bindings):
1. Code guards to prevent setting min / max temps outside 15-70°C and log error.
2. Remove min, max, step properties from Setpoint

Alternative is to set the pattern="%.0f °C" and explore how the REST API, MainUI and sitemaps behave if my item has unit of °F?

bundles/org.openhab.binding.tsmart/pom.xml
```diff
@@ -0,0 +1,17 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns="http://maven.apache.org/POM/4.0.0"
+  xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd">
+
+  <modelVersion>4.0.0</modelVersion>
+
+  <parent>
+    <groupId>org.openhab.addons.bundles</groupId>
+    <artifactId>org.openhab.addons.reactor.bundles</artifactId>
+    <version>4.1.0-SNAPSHOT</version>
```
wborn, line 10: 1
```suggestion
    <version>4.2.0-SNAPSHOT</version>
```

bundles/org.openhab.binding.tsmart/pom.xml
```diff
@@ -0,0 +1,17 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns="http://maven.apache.org/POM/4.0.0"
+  xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd">
+
+  <modelVersion>4.0.0</modelVersion>
+
+  <parent>
+    <groupId>org.openhab.addons.bundles</groupId>
+    <artifactId>org.openhab.addons.reactor.bundles</artifactId>
+    <version>4.2.0-SNAPSHOT</version>
```
wborn, line 10:
```suggestion
    <version>4.3.0-SNAPSHOT</version>
```

lsiepel:
Gentle ping @jamesmelville i know this is marked as draft, but would also like to know if you are able to proceed with this PR and if you need any help to get this to the finish line.

jamesmelville:
Thanks for the ping @lsiepel. I think this is pretty close, the binding has been running without issue on my system for a long time now.
- I've not pushed this mainly because there's nobody pushing me to release this - no response at https://community.openhab.org/t/tesla-t-smart-smart-immersion-heater-thermostat-binding/143521 - my thinking was it;s probably best maintainers look at what's being asked for, but no harm in getting it finished if you have the bandwidth!
- Also I hadn't pushed it because discovery seemed broken, but I think it's working fine and suspect the problem there was me using codespaces.
- I'd appreciate any guidance you can share in this thread https://github.com/openhab/openhab-addons/pull/14247#discussion_r1406821570. I think they're all good questions but not held against other bindings, and I'm not sure what to do as it doesn't seem like these points are handled with UoM in core?
- Otherwise it's probably just things I can handle, update to build latest version, fix against codestyle changes etc.

lsiepel:
> * I'd appreciate any guidance you can share in this thread [[tsmart] Binding support for Tesla T-Smart Smart Immersion Heater / Thermostat #14247 (comment)](https://github.com/openhab/openhab-addons/pull/14247#discussion_r1406821570). I think they're all good questions but not held against other bindings, and I'm not sure what to do as it doesn't seem like these points are handled with UoM in core?

UI and UoM are in this case 'not connected'. Meaning the min/max/step is not unit aware. This is a limitation and a common approach is as follows:

1. The UI is there to guide the user; setting the min to 15 and max to 157 already help, agree it could be better, but that is just how it is now.
2. In the `handleCommand` method you have more control over the value being send to the device. Here you can check if the set value is within the min/max range while respecting the unit. I guess all input is converted to celsius before sending it to the server, check this celcius value to be within the range and log a warn message if it is outside the range. It is up to you to decide if you still wnat to send it to the server, drop it, or adjust the value to be within the boundaries.

Hopefully this gives you some options, if you need more, please let me know and we will see how to fix this.

lsiepel:
@jamesmelville hope you can bring this to the finish line as this contribution is a good addition and too much time has been spend to throw it away. Let me know if you need anything

lsiepel:
After some time without any updates I close this PR. It is not lost, you will still be able to work on it at your own branch. Feel free to re-open once you can move this PR forward. An alternative could  be  to add this binding to the marketplace.

PR #14248, opened 2023-01-19
[nuvo] Publish ALLOFF event for use in source rules

mlobstein:
Publish the ALLOFF event to all button channels for use in source rules. ALLOFF is sent by the amplifier 5 minutes after all zones are switched off.

Also follow-up to #14012. Add the missing feature that will update the other group zone's source channel when a source channel receives a command.


lolodomo (approved):
LGTM, thank you

PR #14249, opened 2023-01-19
[unifi] Various stability improvements

Hilbrand:
- Fixed bug, causing nullpointer, in devices not reporting voltage/ampere values on PoE ports.
- Added some null checks to avoid having null pointer exceptions.
- Handled timeout exception that is in ExecutionException.
- Improved handling case where server returns data, but it's not from the api, and therefor can't be parsed.
- Improved adding additional text to the error messages by adding the reported exception message too.


bundles/org.openhab.binding.unifi/src/main/resources/OH-INF/i18n/unifi.properties
```diff
@@ -132,10 +132,11 @@ channel-type.config.unifi.poeEnable.mode.option.passthrough = Passthrough
...
 
-error.bridge.offline.communication_error = Error communicating with the UniFi controller.
-error.bridge.offline.invalid_credentials = Invalid username and/or password - please double-check your configuration.
-error.bridge.offline.invalid_hostname = Invalid hostname - please double-check your configuration.
-error.bridge.offline.ssl_error = Error establishing an SSL connection with the UniFi controller.
+error.bridge.offline.communication_error = Error communicating with the UniFi controller: {0}
+error.bridge.offline.invalid_credentials = Invalid username and/or password - please double-check your configuration: {0}
+error.bridge.offline.invalid_hostname = Invalid hostname - please double-check your configuration: {0}
+error.bridge.offline.ssl_error = Error establishing an SSL connection with the UniFi controller: {0}
+error.controller.parse_error = Could not parse json from the UniFi Controller. Is the bridge configured with the correct host and/or port?
```
jlaur, line 139:
```suggestion
error.controller.parse_error = Could not parse JSON from the UniFi Controller. Is the bridge configured with the correct host and/or port?
```

bundles/org.openhab.binding.unifi/src/main/java/org/openhab/binding/unifi/internal/handler/UniFiControllerThingHandler.java
```diff
@@ -138,14 +138,14 @@ private void start(final UniFiController uc) {
             uc.start();
             startRefresh = true;
         } catch (final UniFiCommunicationException e) {
-            updateStatus(OFFLINE, COMMUNICATION_ERROR, STATUS_DESCRIPTION_COMMUNICATION_ERROR);
+            updateStatus(OFFLINE, COMMUNICATION_ERROR, STATUS_DESCRIPTION_COMMUNICATION_ERROR + e.getMessage());
```
jlaur, line 141: Will this work without surrounding `[]`? Like here:

https://github.com/openhab/openhab-addons/blob/80094b5e9d293dfb73c81377325da13172d6baaf/bundles/org.openhab.binding.danfossairunit/src/main/java/org/openhab/binding/danfossairunit/internal/handler/DanfossAirUnitHandler.java#L107-L108

https://github.com/openhab/openhab-addons/blob/80094b5e9d293dfb73c81377325da13172d6baaf/bundles/org.openhab.binding.danfossairunit/src/main/resources/OH-INF/i18n/danfossairunit.properties#L84

Just asking, since I'm not aware of this syntax separating I18N text reference and parameter value with just a space, and don't have time to test it right now.

jlaur (changes requested):
Thanks for the fixes! Just one clarification needed, and a very minor text correction suggested.

bundles/org.openhab.binding.unifi/src/main/java/org/openhab/binding/unifi/internal/handler/UniFiControllerThingHandler.java
Hilbrand: Yes it works because core uses handles this in a lenient way. See https://github.com/openhab/openhab-core/blob/dd756b8e2505c6cf9c7555ec7cb73965fd3a5328/bundles/org.openhab.core.thing/src/main/java/org/openhab/core/thing/i18n/ThingStatusInfoI18nLocalizationService.java#L174

I had looked in the code, but could not directly see what the intended usage was. However I've now changed it to use the string array notation as that seems to be what is the official syntax it should be.

jlaur: Thanks for the reference and explanation.

bundles/org.openhab.binding.unifi/src/main/java/org/openhab/binding/unifi/internal/api/UniFiControllerRequest.java
```diff
@@ -128,10 +131,20 @@ public void setQueryParameter(final String key, final Object value) {
...
+                final JsonObject jsonObject = JsonParser.parseString(json).getAsJsonObject();
 
-            if (jsonObject.has(PROPERTY_DATA) && jsonObject.get(PROPERTY_DATA).isJsonArray()) {
-                result = (T) gson.fromJson(jsonObject.getAsJsonArray(PROPERTY_DATA), resultType);
+                if (jsonObject.has(PROPERTY_DATA) && jsonObject.get(PROPERTY_DATA).isJsonArray()) {
+                    result = (T) gson.fromJson(jsonObject.getAsJsonArray(PROPERTY_DATA), resultType);
+                }
+            } catch (final JsonParseException e) {
+                logger.debug(
+                        "Could not parse content retrieved from the server. Is the configuration poiting to the right server/port?, {}",
```
jlaur, line 142:
```suggestion
                        "Could not parse content retrieved from the server. Is the configuration pointing to the right server/port?, {}",
```

PR #14163, opened 2023-01-05
[wemo] Fix stability issues

jlaur:
Fixes #14162.

Tested on my system for a few weeks now and didn't see any blocked threads since. Also my system in general is now stable and responsive after running for more than a week without restarts.

JARs for testing:
- [org.openhab.binding.wemo-3.4.2-SNAPSHOT.jar](https://drive.google.com/file/d/1q_RcE8JnqDQeJI6c5RkGcDZuP8Uyo_Qc/view?usp=sharing)
- [org.openhab.binding.wemo-4.0.0-SNAPSHOT.jar](https://drive.google.com/file/d/1ASIeKW1wopgw0Ac9DQiRwyY1yi-cUgEG/view?usp=sharing)

Issue #12121, opened 2022-01-26
[openhabcloud] Tuning of reconnection parameters


ssalonen:
As [discussed in community forums](https://community.openhab.org/t/disconnected-from-the-openhab-cloud-service/132335/28?u=ssalonen) with @digitaldan 

> ... consider tuning the reconnection parameters such that it would be more graceful to the cloud?

> That would be great, the initial websocket connection to our cloud service is an expensive operation, so when we have tens of thousands of OH’s all trying to connect at once, we almost DDOS the system.

To my knowledge, currently the cloud connector re-connects based on two different logic
1. automatically as dictated by IO.socket
2. "manually" by the CloudClient, when socket is disconnected and we receive Socket.EVENT_ERROR

Both of the places uses exponential backoff and random jitter to distribute the reconnects.

Parameters are as follows

IO.socket reconnects (case 1): 
- defaults, that is: 

https://github.com/socketio/socket.io-client-java/blob/89ef9d09cee799ffb85e0252bcaf1993f9d49af9/src/main/java/io/socket/client/Manager.java#L146-L149

```java
this.reconnectionDelay(opts.reconnectionDelay != 0 ? opts.reconnectionDelay : 1000); // "MIN"
this.reconnectionDelayMax(opts.reconnectionDelayMax != 0 ? opts.reconnectionDelayMax : 5000); // "MAX" 
this.randomizationFactor(opts.randomizationFactor != 0.0 ? opts.randomizationFactor : 0.5); // "JITTER"
```

Cloud client reconnects, manually set (case 2):

```java
reconnectBackoff.setMin(1000);
reconnectBackoff.setMax(30_000);
reconnectBackoff.setJitter(0.5);
```

The delay is calculated as follows

https://github.com/socketio/socket.io-client-java/blob/89ef9d09cee799ffb85e0252bcaf1993f9d49af9/src/main/java/io/socket/backo/Backoff.java#L16-L27

```java
    public long duration() {
// [this.ms = "MIN"]
        BigInteger ms = BigInteger.valueOf(this.ms)
                .multiply(BigInteger.valueOf(this.factor).pow(this.attempts++));
        if (jitter != 0.0) {
            double rand = Math.random();
            BigInteger deviation = BigDecimal.valueOf(rand)
                    .multiply(BigDecimal.valueOf(jitter))
                    .multiply(new BigDecimal(ms)).toBigInteger();
[...]

PR #14251, opened 2023-01-21
[openhabcloud] Reconnection Fixes

digitaldan:

### DIFF
--- a/bundles/org.openhab.io.openhabcloud/src/main/java/org/openhab/io/openhabcloud/internal/CloudClient.java
+++ b/bundles/org.openhab.io.openhabcloud/src/main/java/org/openhab/io/openhabcloud/internal/CloudClient.java
@@ -25,6 +25,7 @@ import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.TimeUnit;
 
 import org.eclipse.jdt.annotation.Nullable;
@@ -139,10 +140,12 @@ public class CloudClient {
 
     /*
      * Delay reconnect scheduler pool
-     * 
+     *
      */
     protected final ScheduledExecutorService scheduler = ThreadPoolManager
             .getScheduledPool(ThreadPoolManager.THREAD_POOL_NAME_COMMON);
+
+    private ScheduledFuture<?> reconnectFuture;
 
     /**
      * Constructor of CloudClient
@@ -325,7 +328,8 @@ public class CloudClient {
                             logger.warn("Error connecting to the openHAB Cloud instance. Reconnecting.");
                         }
                         socket.close();
-                        scheduler.schedule(new Runnable() {
+                        stopReconnectFuture();
+                        reconnectFuture = scheduler.schedule(new Runnable() {
                             @Override
                             public void run() {
                                 socket.connect();
@@ -671,6 +675,7 @@ public class CloudClient {
      */
     public void shutdown() {
         logger.info("Shutting down openHAB Cloud service connection");
+        stopReconnectFuture();
         socket.disconnect();
     }
 
@@ -684,6 +689,14 @@ public class CloudClient {
 
     public void setListener(CloudClientListener listener) {
         this.listener = listener;
+    }
+
+    private void stopReconnectFuture() {
+        ScheduledFuture<?> reconnectFuture = this.reconnectFuture;
+        if (reconnectFuture != null) {
+            reconnectFuture.cancel(true);
+            this.reconnectFuture = null;
+        }
     }
 
     private JSONObject getJSONHeaders(HttpFields httpFields) {
--- a/bundles/org.openhab.io.openhabcloud/src/main/java/org/openhab/io/openhabcloud/internal/CloudClient.java
+++ b/bundles/org.openhab.io.openhabcloud/src/main/java/org/openhab/io/openhabcloud/internal/CloudClient.java
@@ -293,13 +293,17 @@ public class CloudClient {
                 .on(Socket.EVENT_RECONNECT_FAILED,
                         args -> logger.debug("Socket.IO re-connect attempts failed. Stopping reconnection."))//
                 .on(Socket.EVENT_DISCONNECT, args -> {
-                    if (args.length > 0) {
-                        logger.warn("Socket.IO disconnected: {}", args[0]);
-                    } else {
-                        logger.warn("Socket.IO disconnected");
-                    }
+                    String message = args.length > 0 ? args[0].toString() : "";
+                    logger.warn("Socket.IO disconnected: {}", message);
                     isConnected = false;
                     onDisconnect();
+                    // https://github.com/socketio/socket.io-client/commit/afb952d854e1d8728ce07b7c3a9f0dee2a61ef4e
+                    if ("io server disconnect".equals(message)) {
+                        socket.close();
+                        long delay = reconnectBackoff.duration();
+                        logger.warn("Reconnecting after {} ms.", delay);
+                        scheduleReconnect(delay);
+                    }
                 })//
                 .on(Socket.EVENT_ERROR, args -> {
                     if (CloudClient.this.socket.connected()) {
@@ -345,12 +349,7 @@ public class CloudClient {
                             logger.warn("Error connecting to the openHAB Cloud instance. Reconnecting.");
                         }
                         socket.close();
-                        reconnectFuture.getAndSet(Optional.of(scheduler.schedule(new Runnable() {
-                            @Override
-                            public void run() {
-                                socket.connect();
-                            }
-                        }, delay, TimeUnit.MILLISECONDS))).ifPresent(future -> future.cancel(true));
+                        scheduleReconnect(delay);
                     }
                 })//
 
@@ -697,6 +696,15 @@ public class CloudClient {
 
     public void setListener(CloudClientListener listener) {
         this.listener = listener;
+    }
+
+    private void scheduleReconnect(long delay) {
+        reconnectFuture.getAndSet(Optional.of(scheduler.schedule(new Runnable() {
+            @Override
+            public void run() {
+                socket.connect();
+            }
+        }, delay, TimeUnit.MILLISECONDS))).ifPresent(future -> future.cancel(true));
     }
 
     private JSONObject getJSONHeaders(HttpFields httpFields) {
--- a/bundles/org.openhab.io.openhabcloud/src/main/java/org/openhab/io/openhabcloud/internal/CloudClient.java
+++ b/bundles/org.openhab.io.openhabcloud/src/main/java/org/openhab/io/openhabcloud/internal/CloudClient.java
@@ -186,14 +186,14 @@ public class CloudClient {
         try {
             Options options = new Options();
             options.transports = new String[] { WebSocket.NAME };
-            options.reconnection = true; // default value true
+            options.reconnection = true;
             options.reconnectionAttempts = Integer.MAX_VALUE;
-            options.reconnectionDelay = RECONNECT_MIN; // default value 1_000
-            options.reconnectionDelayMax = RECONNECT_MAX; // default value 5_000
-            options.randomizationFactor = RECONNECT_JITTER; // default value 0.5
-            options.timeout = READ_TIMEOUT; // default value 20_000
+            options.reconnectionDelay = RECONNECT_MIN;
+            options.reconnectionDelayMax = RECONNECT_MAX;
+            options.randomizationFactor = RECONNECT_JITTER;
+            options.timeout = READ_TIMEOUT;
             Builder okHttpBuilder = new Builder();
-            okHttpBuilder.readTimeout(READ_TIMEOUT, TimeUnit.MILLISECONDS); // default 10_000
+            okHttpBuilder.readTimeout(READ_TIMEOUT, TimeUnit.MILLISECONDS);
             if (logger.isTraceEnabled()) {
                 // When trace level logging is enabled, we activate further logging of HTTP calls
                 // of the Socket.IO library
--- a/bundles/org.openhab.io.openhabcloud/src/main/java/org/openhab/io/openhabcloud/internal/CloudClient.java
+++ b/bundles/org.openhab.io.openhabcloud/src/main/java/org/openhab/io/openhabcloud/internal/CloudClient.java
@@ -72,6 +72,15 @@ import okhttp3.logging.HttpLoggingInterceptor.Level;
  * @author Kai Kreuzer - migrated code to new Jetty client and ESH APIs
  */
 public class CloudClient {
+
+    private static final long RECONNECT_MIN = 1_000;
+
+    private static final long RECONNECT_MAX = 60_000;
+
+    private static final double RECONNECT_JITTER = 0.5;
+
+    private static final long READ_TIMEOUT = 60_0000;
+
     /*
      * Logger for this class
      */
@@ -164,9 +173,9 @@ public class CloudClient {
         this.remoteAccessEnabled = remoteAccessEnabled;
         this.exposedItems = exposedItems;
         this.jettyClient = httpClient;
-        reconnectBackoff.setMin(1000);
-        reconnectBackoff.setMax(30_000);
-        reconnectBackoff.setJitter(0.5);
+        reconnectBackoff.setMin(RECONNECT_MIN);
+        reconnectBackoff.setMax(RECONNECT_MAX);
+        reconnectBackoff.setJitter(RECONNECT_JITTER);
     }
 
     /**
@@ -176,20 +185,25 @@ public class CloudClient {
     public void connect() {
         try {
             Options options = new Options();
-            // we always use websockets, this prevents unnecessary polling requests
             options.transports = new String[] { WebSocket.NAME };
-
+            options.reconnection = true; // default value true
+            options.reconnectionAttempts = Integer.MAX_VALUE;
+            options.reconnectionDelay = RECONNECT_MIN; // default value 1_000
+            options.reconnectionDelayMax = RECONNECT_MAX; // default value 5_000
+            options.randomizationFactor = RECONNECT_JITTER; // default value 0.5
+            options.timeout = READ_TIMEOUT; // default value 20_000
+            Builder okHttpBuilder = new Builder();
+            okHttpBuilder.readTimeout(READ_TIMEOUT, TimeUnit.MILLISECONDS); // default 10_000
             if (logger.isTraceEnabled()) {
                 // When trace level logging is enabled, we activate further logging of HTTP calls
                 // of the Socket.IO library
-                Builder okHttpBuilder = new Builder();
                 HttpLoggingInterceptor loggingInterceptor = new HttpLoggingInterceptor();
                 loggingInterceptor.setLevel(Level.BASIC);
                 okHttpBuilder.addInterceptor(loggingInterceptor);
                 okHttpBuilder.addNetworkInterceptor(loggingInterceptor);
-                options.callFactory = okHttpBuilder.build();
-                options.webSocketFactory = okHttpBuilder.build();
             }
+            options.callFactory = okHttpBuilder.build();
+            options.webSocketFactory = okHttpBuilder.build();
             socket = IO.socket(baseURL, options);
             URL parsed = new URL(baseURL);
             protocol = parsed.getProtocol();
--- a/bundles/org.openhab.io.openhabcloud/src/main/java/org/openhab/io/openhabcloud/internal/CloudClient.java
+++ b/bundles/org.openhab.io.openhabcloud/src/main/java/org/openhab/io/openhabcloud/internal/CloudClient.java
@@ -73,11 +73,11 @@ import okhttp3.logging.HttpLoggingInterceptor.Level;
  */
 public class CloudClient {
 
-    private static final long RECONNECT_MIN = 5_000;
+    private static final long RECONNECT_MIN = 2_000;
 
     private static final long RECONNECT_MAX = 60_000;
 
-    private static final double RECONNECT_JITTER = 0.5;
+    private static final double RECONNECT_JITTER = 0.75;
 
     private static final long READ_TIMEOUT = 60_0000;
 
--- a/bundles/org.openhab.io.openhabcloud/src/main/java/org/openhab/io/openhabcloud/internal/CloudClient.java
+++ b/bundles/org.openhab.io.openhabcloud/src/main/java/org/openhab/io/openhabcloud/internal/CloudClient.java
@@ -73,7 +73,7 @@ import okhttp3.logging.HttpLoggingInterceptor.Level;
  */
 public class CloudClient {
 
-    private static final long RECONNECT_MIN = 1_000;
+    private static final long RECONNECT_MIN = 5_000;
 
     private static final long RECONNECT_MAX = 60_000;
 
--- a/bundles/org.openhab.io.openhabcloud/src/main/java/org/openhab/io/openhabcloud/internal/CloudClient.java
+++ b/bundles/org.openhab.io.openhabcloud/src/main/java/org/openhab/io/openhabcloud/internal/CloudClient.java
@@ -22,11 +22,13 @@ import java.nio.charset.StandardCharsets;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicReference;
 
 import org.eclipse.jdt.annotation.Nullable;
 import org.eclipse.jetty.client.HttpClient;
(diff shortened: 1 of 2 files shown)


digitaldan:
@ssalonen there seems to be a pretty bad issue with our cloud binding.  Would appreciate some eyes on this one.  Unfortunetly I'm leaving town tomorrow, but will try and look at this when i can until i'm back mid next week.

When debugging connection issues on our cloud service,  i noticed our nginx load balancers are being slammed with the following logs hundreds of time per minute (or more, its very noisy).

```
1.2.3.4 - - [21/Jan/2023:04:53:02 +0000] "GET /socket.io/?EIO=3&transport=polling HTTP/1.1" 200 99 "-" "okhttp/3.8.1"
1.2.3.4 - - [21/Jan/2023:04:53:02 +0000] "GET /socket.io/?EIO=3&transport=polling&sid=9nGwDNdsO20YmYRsKHNf HTTP/1.1" 200 21 "-" "okhttp/3.8.1"
1.2.3.4 - - [21/Jan/2023:04:53:02 +0000] "POST /socket.io/?EIO=3&transport=polling&sid=9nGwDNdsO20YmYRsKHNf HTTP/1.1" 200 2 "-" "okhttp/3.8.1"
1.2.3.4 - - [21/Jan/2023:04:53:03 +0000] "GET /socket.io/?EIO=3&transport=websocket&sid=9nGwDNdsO20YmYRsKHNf HTTP/1.1" 400 18 "-" "okhttp/3.8.1"
1.2.3.4 - - [21/Jan/2023:04:53:32 +0000] "GET /socket.io/?EIO=3&transport=polling HTTP/1.1" 200 99 "-" "okhttp/3.8.1"
1.2.3.4 - - [21/Jan/2023:04:53:32 +0000] "GET /socket.io/?EIO=3&transport=polling&sid=QEz-JkWRBJNHY-D3KHPe HTTP/1.1" 200 21 "-" "okhttp/3.8.1"
1.2.3.4 - - [21/Jan/2023:04:53:32 +0000] "POST /socket.io/?EIO=3&transport=polling&sid=QEz-JkWRBJNHY-D3KHPe HTTP/1.1" 200 2 "-" "okhttp/3.8.1"
1.2.3.4 - - [21/Jan/2023:04:53:33 +0000] "GET /socket.io/?EIO=3&transport=websocket&sid=QEz-JkWRBJNHY-D3KHPe HTTP/1.1" 400 18 "-" "okhttp/3.8.1"
```

If you look you can see that the there are a few polling calls then a websocket attempt, which fails with a 400 error.  At first i thought we had an issue with something server side preventing websockets and forcing polling, so started going down that route.  

But on my test system i noticed that even when I stopped the cloud binding on my local machine through the karaf console, the binding was continuing to log connection events !!!

```
2023-01-20 21:04:42.922 [DEBUG] [io.openhabcloud.internal.CloudClient] - Socket.IO ping
2023-01-20 21:04:42.996 [DEBUG] [io.openhabcloud.internal.CloudClient] - Socket.IO pong: 74 m
```

What i have realized is that the binding is somehow creating orphaned socket.io connections ,  so a thread leak.  I don't think these connections don't actually fully connect, the session gets rejected with the `400` error, but nevertheless they are slamming our servers and making debugging very hard.

As a test i stopped the binding (using the console) on my home production system which has been running for a few days, and can see the same behavior, my log continues to show these ping and pong messages,  so i assume this is extremely widespread.

As a start i went through the code looking for how something like this could occur, the first thing i noticed was the reconnect code schedules a future, but does not keep a reference to it for clean up purposes, this seems like a good place to start, as i could see this connecting a socket after the binding was stopped or restarted (like when saving configs).  I'm going to let my test system run for a bit, but so far i am not seeing the same behavior after this fix, but need to do more testing.

digitaldan:
FYI, the `400` errors and the leak here may not be related, i thought they were, but they may be openhabs that have installed the binding but have not actually signed up for an account (so likely a lot).  The leak problem still exists, and i'm not sure if it could cause other issues, but may not be responsible for the noise in the logs.

ssalonen:
I will have a look with this PR.

In https://github.com/openhab/openhab-addons/pull/11153 the reconnect behaviour changed such that the addon retries connect, also with "errors during connection": https://github.com/openhab/openhab-addons/blob/dff0bda3c1883c0c9cc0c4f9eea677c40319d083/bundles/org.openhab.io.openhabcloud/src/main/java/org/openhab/io/openhabcloud/internal/CloudClient.java#L266

It was later made to use scheduler: https://github.com/openhab/openhab-addons/pull/13421

re. Http 400 and retry

Actually invalid uuid was one of the methods I tested https://github.com/openhab/openhab-addons/pull/11153#issuecomment-905567791

bundles/org.openhab.io.openhabcloud/src/main/java/org/openhab/io/openhabcloud/internal/CloudClient.java
```diff
@@ -25,6 +25,7 @@
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.TimeUnit;
```
ssalonen, line 29:
```suggestion
import java.util.concurrent.ScheduledFuture;
import java.util.concurrent.TimeUnit;
import java.util.concurrent.atomic.AtomicReference;
```

bundles/org.openhab.io.openhabcloud/src/main/java/org/openhab/io/openhabcloud/internal/CloudClient.java
```diff
@@ -25,6 +25,7 @@
 import java.util.Set;
```
ssalonen, line 25:
```suggestion
import java.util.Set;
import java.util.Optional;
```

bundles/org.openhab.io.openhabcloud/src/main/java/org/openhab/io/openhabcloud/internal/CloudClient.java
```diff
@@ -139,11 +140,13 @@ public class CloudClient {
...
     /*
      * Delay reconnect scheduler pool
-     * 
+     *
      */
     protected final ScheduledExecutorService scheduler = ThreadPoolManager
             .getScheduledPool(ThreadPoolManager.THREAD_POOL_NAME_COMMON);
 
+    private ScheduledFuture<?> reconnectFuture;
+
```
ssalonen, line 149:
```suggestion
    private final AtomicReference<Optional<ScheduledFuture<?>>> reconnectFuture = new AtomicReference<>(Optional.empty());

```

bundles/org.openhab.io.openhabcloud/src/main/java/org/openhab/io/openhabcloud/internal/CloudClient.java
```diff
@@ -325,7 +328,8 @@ public void call(Object... args) {
                             logger.warn("Error connecting to the openHAB Cloud instance. Reconnecting.");
                         }
                         socket.close();
-                        scheduler.schedule(new Runnable() {
+                        stopReconnectFuture();
```
ssalonen, line 331: Whole stop/reschedule could be:
```java
reconnectFuture.getAndSet(Optional.of(scheduler.schedule(() -> socket.connect()))).ifPresent(prevFuture -> prevFuture.cancel(true));
```

bundles/org.openhab.io.openhabcloud/src/main/java/org/openhab/io/openhabcloud/internal/CloudClient.java
```diff
@@ -671,6 +675,7 @@ public boolean isConnected() {
      */
     public void shutdown() {
         logger.info("Shutting down openHAB Cloud service connection");
+        stopReconnectFuture();
```
ssalonen, line 678:
```suggestion
        reconnectFuture.getAndSet(Optional.empty()).ifPresent(future -> future.cancel(true));
```

bundles/org.openhab.io.openhabcloud/src/main/java/org/openhab/io/openhabcloud/internal/CloudClient.java
```diff
@@ -686,6 +691,14 @@ public void setListener(CloudClientListener listener) {
...
     }
 
+    private void stopReconnectFuture() {
+        ScheduledFuture<?> reconnectFuture = this.reconnectFuture;
+        if (reconnectFuture != null) {
+            reconnectFuture.cancel(true);
+            this.reconnectFuture = null;
+        }
+    }
+
```
ssalonen, line 701:
```suggestion
```

ssalonen (changes requested):
I suggest to follow the thread-safe pattern established in many bindings, e.g https://github.com/openhab/openhab-addons/blob/main/bundles/org.openhab.binding.webthing/src/main/java/org/openhab/binding/webthing/internal/WebThingHandler.java

digitaldan:
>In https://github.com/openhab/openhab-addons/pull/11153 the reconnect behaviour changed such that the addon retries connect, also with "errors during connection":

Reading that PR again, i noticed you mentioned:
> One way to confirm changed behaviour is to introduce invalid UUID, which is rejected by the server ("not authorized" error). With the new code, client would automatically retry. Old implementation stopped alltogether without retriying.

So with the new code (well newish at this point) , we are now trying indefinitely where before the binding gave up?  I wondering if this is the source of so many rejections (the logs are full of them).  I have noticed this in the node logs as well. 

>Probably tuning of time between reconnects would be nice?

If the server specifically rejects us, worst case we should not be trying very frequently to reconnect, ideally we would not try and reconnect at all until the user takes some action, but the binding is not set up for that right now.

Long term its probably a good idea to revisit the whole UUID / Secret thing, make this part of the binding configuration and maybe have the server generate the secret and have the user enter it in the binding config in the Main UI, which at that point we try a connection, but not retry if the server specifically rejects it.  (this is all of the top of my head, so probably need to think about it more).   

>I suggest to follow the thread-safe pattern established in many bindings, 

Will do, and thanks, i was also concerned about the multi threaded aspect of this quick fix, but wanted to get something up for discussion.

digitaldan:
Also whatever we do here , i would want to make this part of the 3.4.2 fix release.  Let me know what you think about tuning specific server rejections, would love to reduce that load if possible.

ssalonen:
Yes, agreed that we should not retry if server explicitly rejects. That makes sense.

Perhaps there is a way to differentiate this from other errors...

And indeed: before the PR, any error during connect phase was "fatal", no new attempts were made.

digitaldan:
>Perhaps there is a way to differentiate this from other errors...

yeah, i don't see that specifically in the cloud code, which is unfortunate.  We could add this, so send back a 401 or similar if we can't match credentials.  I'm not sure how that get passed down to the java socket.io client, would have to do some testing.

ssalonen:
in the worst case, we should be able to look for some string "unauthorized" in the error ... whatever gets logged down

digitaldan:
Hi @ssalonen , i'm back spending some time on this.  My latest push incorporates you suggestions, i also added an option to have the library only use websocket connections which removes the annoying polling requests to create the session and instead does it all on a single request (and that we never actually use polling ever,  afaik).

**so this part  is a little tl;dr** , but i would appreciate you taking a look since you know this code better then most.

I have another question for you about the socket.io automatic reconnect logic.  During an event when my system experienced this problem, my logs show that actually the custom reconnect logic was never actually used.  Socket.io was reconnecting internally , which lead to the split brain.  I think i may understand why, but would like your opinion since you also worked on this.

Before i post my logs, i'll mention a couple important things.

It seems that the socket.io client has reconnects enabled by default, so if the connection dies, it tries to connect again behind the scenes.  I believe you put in some nice logging here, so it is much easier to debug the connection lifecycle.  

The other piece here is the okhttp client, by default it has a read timeout of 10 seconds before throwing an error. This will be important below.

So in the logs below you will see that we 

1. have a ping timeout 
2. socket.io disconnects
3. socket.io tries to reconnect (attempt 1), 
4. after 10 seconds we get this "EngineIOException xhr poll error"  
5. socket.io immediately tries again (attempt 2), 
6. socket.io then successfully connects, 
7. but then disconnects 3 seconds later with this same poll error 
8. and starts over again with new attempt 1, which now succeeds.

The problem i believe is that on the first disconnect socket.io tries reconnecting, but then gives up after 10 seconds.  Unfortunately the server may have been very busy and is still processing that request, but the client is now immediately connecting again, so our service is processing 2 connections, and we get a race condition which can lead to the first reconnection overwriting the DB on the second reconnection (which started after, but also completed before the first one)

```
2023-01-25 22:36:04.851 [DEBUG] [io.openhabcloud.internal.CloudClient] - Socket.IO CLOSE: ping timeout
2023-01-25 22:36:04.853 [WARN ] [io.openhabcloud.internal.CloudClient] - Socket.IO disconnected: ping timeout
2023-01-25 22:36:04.854 [INFO ] [io.openhabcloud.internal.CloudClient] - Disconnected from the openHAB Cloud service (UUID = 1b...8a, base URL = http://localhost:8080)
2023-01-25 22:36:05.517 [DEBUG] [io.openhabcloud.internal.CloudClient] - Socket.IO re-connecting (attempt 1)
2023-01-25 22:36:16.249 [DEBUG] [io.openhabcloud.internal.CloudClient] - Error connecting to the openHAB Cloud instance: EngineIOException xhr poll error. Should reconnect automatically.
2023-01-25 22:36:16.251 [DEBUG] [io.openhabcloud.internal.CloudClient] - Socket.IO re-connect attempt error: SocketIOException Connection error
2023-01-25 22:36:18.971 [DEBUG] [io.openhabcloud.internal.CloudClient] - Socket.IO re-connecting (attempt 2)
2023-01-25 22:36:19.612 [DEBUG] [io.openhabcloud.internal.CloudClient] - Socket.IO OPEN
2023-01-25 22:36:19.613 [DEBUG] [io.openhabcloud.internal.CloudClient] - Socket.IO re-connected successfully (attempt 2)
2023-01-25 22:36:23.370 [DEBUG] [io.openhabcloud.internal.CloudClient] - Socket.IO connected
2023-01-25 22:36:23.370 [INFO ] [io.openhabcloud.internal.CloudClient] - Connected to the openHAB Cloud service (UUID = 1b...8a, base URL = http://localhost:8080)
2023-01-25 22:36:26.079 [WARN ] [io.openhabcloud.internal.CloudClient] - Error during communication: EngineIOException xhr poll error
2023-01-25 22:36:26.081 [DEBUG] [io.openhabcloud.internal.CloudClient] - Socket.IO CLOSE: transport error
2023-01-25 22:36:26.082 [WARN ] [io.openhabcloud.internal.CloudClient] - Socket.IO disconnected: transport error
2023-01-25 22:36:26.083 [INFO ] [io.openhabcloud.internal.CloudClient] - Disconnected from the openHAB Cloud service (UUID = 1b...8a, base URL = http://localhost:8080)
2023-01-25 22:36:26.767 [DEBUG] [io.openhabcloud.internal.CloudClient] - Socket.IO re-connecting (attempt 1)
2023-01-25 22:36:28.515 [DEBUG] [io.openhabcloud.internal.CloudClient] - Socket.IO OPEN
2023-01-25 22:36:28.516 [DEBUG] [io.openhabcloud.internal.CloudClient] - Socket.IO re-connected successfully (attempt 1)
2023-01-25 22:36:38.364 [DEBUG] [io.openhabcloud.internal.CloudClient] - Socket.IO connected
2023-01-25 22:36:38.365 [INFO ] [io.openhabcloud.internal.CloudClient] - Connected to the openHAB Cloud service (UUID = 1b...8a, base URL = http://localhost:8080)
```

I have a couple of ideas here.  First is that i think we need to make the okhttp client have a larger read timeout, so something like:
```java
okHttpBuilder.readTimeout(1, TimeUnit.MINUTES);
```

Second is that there are a number of options we can set on the socket.io client around reconnects, for example i would maybe suggest a much slower connect cycle with longer timeouts
```java
    .setReconnection(true)
    .setReconnectionAttempts(Integer.MAX_VALUE) //default value Integer.MAX_VALUE
    .setReconnectionDelay(5_000) //default value 1_000
    .setReconnectionDelayMax(60_000) // default value 5_000
    .setRandomizationFactor(0.5) // default value 0.5
    .setTimeout(60_000) // default value 20_000
```

The other thing i am thinking about (but need to sleep on), is using redis as a network lock, so when a new connection comes in, it creates an entry in redis with some TTL on it.  Once the connection is complete that lock is removed, or if something dies, it expires after a reasonable amount of time.  If a second connection comes in, it will see the lock and fail. 

wdyt?

digitaldan:
and i have a redis solution i am testing on my system

https://github.com/openhab/openhab-cloud/compare/main...digitaldan:openhab-cloud:redis-connection-lock?expand=1

digitaldan:
redis PR
https://github.com/openhab/openhab-cloud/pull/413

ssalonen:
I will answer the quick points first

> a much slower connect cycle with longer timeouts

agreed. Those are fairly big numbers(1), but I guess still reasonable: delay of at most 60s between reconnects.

> socket.io automatic reconnect logic. During an event when my system experienced this problem, my logs show that actually the custom reconnect logic was never actually used. 

It will recover certain connection failures which would not be otherwise retried (I can try to find the details from the old PR). Or that is the attempt at least. The intent was to see if this was the culprit, before all the debugging and logging done by the community.

Indeed many failures are connected by socket.io internal logic, and no separate reconnection logic is applied on binding side.

Actually, I have no records of any user reports where this would have "fixed" a connection. I might even go further and say we remove this logic alltogether. Let me still check the old PR, hopefully later today


> we get a race condition which can lead to the first reconnection overwriting the DB on the second reconnection (which started after, but also completed before the first one)

this is a great finding and... plausible

I will check the redis pr, initially this sounds very robust solution and exactly what we need!

(1) see earlier experiments https://github.com/openhab/openhab-addons/issues/12121#issuecomment-1377539222

bundles/org.openhab.io.openhabcloud/src/main/java/org/openhab/io/openhabcloud/internal/CloudClient.java
```diff
@@ -108,11 +121,6 @@ public class CloudClient {
      */
     private boolean isConnected;
 
-    /*
-     * This variable holds version of local openHAB
-     */
-    private String openHABVersion;
```
ssalonen, line 114: Was the openHABVersion always empty and thus cleaned up in this PR?

bundles/org.openhab.io.openhabcloud/src/main/java/org/openhab/io/openhabcloud/internal/CloudClient.java
```diff
@@ -161,9 +173,9 @@ public CloudClient(HttpClient httpClient, String uuid, String secret, String bas
         this.remoteAccessEnabled = remoteAccessEnabled;
         this.exposedItems = exposedItems;
         this.jettyClient = httpClient;
-        reconnectBackoff.setMin(1000);
-        reconnectBackoff.setMax(30_000);
-        reconnectBackoff.setJitter(0.5);
+        reconnectBackoff.setMin(RECONNECT_MIN);
```
ssalonen, line 176: Perhaps it would make sense to have larger initial delay, to be less aggressive to server

ssalonen: [I suggested](https://github.com/openhab/openhab-addons/issues/12121#issuecomment-1377539222) longer delays:  `RECONNECT_MIN=2000`, `jitter=0.75`. Capping to 60 seconds makes sense.

image

```
attempt min/max
0 [500.0, 3500.0],
1 [1000.0, 7000.0],
2 [2000.0, 14000.0],
3 [4000.0, 28000.0],
4 [8000.0, 56000.0],
5 [16000.0, 60000]
```

Compare this to this PR (`RECONNECT_MIN=1000`, `JITTER=0.5`)

```
attempt min/max
0 [500.0, 1500.0],
1 [1000.0, 3000.0],
2 [2000.0, 6000.0],
3 [4000.0, 12000.0],
4 [8000.0, 24000.0],
5 [16000.0, 48000.0]]
```
image

digitaldan: Agreed

bundles/org.openhab.io.openhabcloud/src/main/java/org/openhab/io/openhabcloud/internal/CloudClient.java
digitaldan: I think a long time ago it was needed, but at some point the code was changed to use `OpenHAB.getVersion()` so this variable and getter/setter was not used, so dead code.

digitaldan:
> It will recover certain connection failures which would not be otherwise retried (I can try to find the details from the old PR).

From my testing, that code gets used when the _server_ specifically calls `socket.disconnect()` (which can happen for a number of server related issues) or i think if the initial attempt to connect does not succeed. (you point this out as well in #12121)  Its basically doing the same thing as the internal socket.io code does once its connected.  Its probably still useful, so i would say we leave it in and just tune it.

>Was redis part of the stack already?

Yes, its used as the backing session store for express as well as doing mongo query caching using Mongoose.  Beyond using  it as a network connection lock, i think there is a good opportunity to use it for other purposes, like offline tracking, proxy route look ups, etc.... 

I'm right now looking to upgrade our Redis and Mongo servers to the latest versions, there are a few improvements  that we could use to make the lock solution better with less calls.   
 
>(1) see earlier experiments https://github.com/openhab/openhab-addons/issues/12121#issuecomment-1377539222

Yikes, if i had gone back and read your post again, it probably would have saved me much time as you clearly pointed out the reconnection settings in socket.io !  Apologies for not responding back to that post ! 

Thanks for taking a look!

ssalonen:
> From my testing, that code gets used when the server specifically calls socket.disconnect()

Ok, sounds promising. I must say I have hard time reading the socket/engine client code and understand when that is triggered. I think I have only seen it with first connect errors (wrong uuid example)..

digitaldan:
>  I think I have only seen it with first connect errors (wrong uuid example)..

Actually you may be right, i think this gets called when the server calls `socket.disconnect()`

```java
                .on(Socket.EVENT_DISCONNECT, args -> {
                    if (args.length > 0) {
                        logger.warn("Socket.IO disconnected: {}", args[0]);
                    } else {
                        logger.warn("Socket.IO disconnected");
                    }
                    isConnected = false;
                    onDisconnect();
                })//
```
This means we are no longer connected, but no retry logic is attempted.  Its likely this is occasionally the cause of the binding needed to be restarted after server issues ?

ssalonen:
Hmm, you mean that server calling socket.disconnect would in some cases result in error on client side?

bundles/org.openhab.io.openhabcloud/src/main/java/org/openhab/io/openhabcloud/internal/CloudClient.java
digitaldan: Yeah , that seems more reasonable.  Let give it a go.

digitaldan:
> you mean that server calling socket.disconnect would in some cases result in error on client side?

If the _server_ calls `socket.disconnect()` then that code path gets executed, which does not try to reconnect.  We call disconnect on the server if we have a general runtime error, like if we are not able to talk to mongo .  I'll play with that some more, but  maybe in another PR as i have already made this one a little broad 😱

digitaldan:
Ok, i'm good with this PR, i think this should also be part of the next 3.4.x  release as well.

bundles/org.openhab.io.openhabcloud/src/main/java/org/openhab/io/openhabcloud/internal/CloudClient.java
```diff
@@ -273,13 +293,17 @@ public void call(Object... args) {
...
-                        logger.warn("Socket.IO disconnected: {}", args[0]);
-                    } else {
-                        logger.warn("Socket.IO disconnected");
-                    }
+                    String message = args.length > 0 ? args[0].toString() : "";
+                    logger.warn("Socket.IO disconnected: {}", message);
                     isConnected = false;
                     onDisconnect();
+                    // https://github.com/socketio/socket.io-client/commit/afb952d854e1d8728ce07b7c3a9f0dee2a61ef4e
+                    if ("io server disconnect".equals(message)) {
```
digitaldan, line 301: @ssalonen last change, after more testing i can confirm when the server calls disconnect, our client will not reconnect.  This is actually a design flaw (in my opinion) of the socket.io client library.  You can see in the code comment that this is actually what the official docs suggest doing, checking for this string.  Hopefully this solves yet another reconnect issue !   I have done quite a bit of testing on this and feel good about this change (along with the thread safe scheduling of course)

ssalonen: What a find! Really good

ssalonen (approved):
Looks good. If you like, add a comment referencing socket io java client code

lolodomo (approved):
LGTM, thank you

lolodomo:
> i think this should also be part of the next 3.4.x release as well.

@jlaur : could you please cherry-pick that PR in 3.4.x branch ?

jlaur:
> could you please cherry-pick that PR in 3.4.x branch ?

Done. Can you have a look at these, which I would also consider for 3.4.x:
- #14281
- #14163

I just need a second pair of eyes to make the decision. Especially the last one since it's my own PR.

lolodomo:
I already answered for the wemo PR ;)

Will have a look for the JDBC one.

lolodomo:
Agree with you for the JDBC PR.

digitaldan:
Thanks!

Issue #11517, opened 2021-11-05
[deconz] Issue with Danfoss Ally thermostat


SecurityAddicted:
## Expected Behavior
Hello,
I have a set of Danfoss Ally radiator thermostat configured with deCONZ.
The valve position value reported by the REST API ranges from 0 to 100, however the deCONZ binding shows it from 0 to around 39%, but I think it should just report the value as is as it maps correctly as percentage.

## Current Behavior
The deCONZ API shows a percentage value of 100%:
![image](https://user-images.githubusercontent.com/73715/140440884-06606c6d-4e46-4b27-a96e-2d0b831f8194.png)
but the binding reports it as 39%:
![image](https://user-images.githubusercontent.com/73715/140441018-c906fec2-a3bb-4d9f-beb3-45fdf82d6417.png)
Item declaration in OH2.5:
![image](https://user-images.githubusercontent.com/73715/140521583-fa11145e-ebb1-4235-9859-10ced342ef61.png)

## Possible Solution
I think the problem resides on the following line: 
https://github.com/openhab/openhab-addons/blob/05caddab8593f54eb648fd175d549980797924c2/bundles/org.openhab.binding.deconz/src/main/java/org/openhab/binding/deconz/internal/handler/SensorThermostatThingHandler.java#L156

## Context
Trying to activate the boiler based on valve opening (not the best thing to do tho, the device exposes a "Heat Supply Request" attribute which is not available on this binding, I'll open a feature request later)

## Your Environment
Tested in both OpenHAB 2.5 and 3.1.

SecurityAddicted:
Anyone on this please? Thanks.

J-N-K:
Difficult. According to the test API-response (which I believe is a real-world example), the Eurotronic thermostats report 0-255, so the code would be ok. I'll try to check that with the Deconz people.

danieldemus:
The zigbee spec is wonderfully ambiguous, but says the value is reported as an 8-bit value "in percent" [ZigBee Cluster Library Specification](https://zigbeealliance.org/wp-content/uploads/2019/12/07-5123-06-zigbee-cluster-library-specification.pdf) page 373, section 6.3.2.2.1.8 :-0), so I can understand the danfoss interpretation. The question is how other thermostat makers have interpreted this.
[...]

PR #14259, opened 2023-01-22
[deconz] [Proposal] Normalize valve value to zigbee specification

danieldemus:
A suggestion to work around #11517.

Deconz just passes the PIHeatingDemand without any normalization to the relevant range of valid values from the zigbee 
specification. The thermostat that was used as a template for the relevant code is unfortunately not standards compliant, so any standards compliant thermostats show incorrect valve values.

While it would be preferable to have this fixed in deconz itself, until that happens a workaround would be useful.
This PR is a suggestion to treat #11517. 

Signed-off-by: Daniel Demus <daniel-github@demus.dk>


jlaur:
@danieldemus, @J-N-K - is this PR still relevant? The mentioned issue #11517 is fixed by #14622.

J-N-K:
It can be fixed on deconz-side by adding a proper DDF, so it's a non-issue nowadays.

PR #14235, opened 2023-01-16
Fix i18n for bindings

J-N-K:
Regression of #13289 - see https://github.com/openhab/openhab-addons/pull/13289#issuecomment-1384511682

Signed-off-by: Jan N. Klug <github@klug.nrw>

PR #13289, opened 2022-08-18
Adapt to core changes (addon.xml)

J-N-K:
See https://github.com/openhab/openhab-core/pull/3050

This adapts to the changes in core and adds metadata for all add-ons

PR #14261, opened 2023-01-22
Fix some addon.xml issues

jlaur:
Regression of #13289, #14235


lolodomo (approved):
LGTM

bundles/org.openhab.binding.velux/src/main/resources/OH-INF/thing/binding.xml
```diff
@@ -0,0 +1,22 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
J-N-K, line 1: This should not be named `binding.xml`.

jlaur: I agree it's a bit strange/awkward having **binding** as thing-type, but is the filename actually a problem? What would you suggest?

J-N-K: The most common name seems to be `thing-types.xml`.

lolodomo: thing-types.xml is the default file name (bonding skeleton) and what is commonly used by a lot of bondings.

jlaur: We could prefix the filenames with "thing-type-"? This binding has a separate file per thing type.

lolodomo: Ok sorry, I understood. The thing type id is "binding"! What a strange idea! It should have been noticed at review time. 
Maybe use as filename what should have been used as thing type id?

lolodomo: Looking quickly at the code, my feeling is that this thing type should have never existed. It was created apparently just to provide information about the binding to the user. 

You could call the file binding-information.xml for example.

andrewfg: Yeah. I think Guether created it in order to help him debug issues raised on the community forum. Specifically it puts the OH build version into the log file so one can easily see the context just from the log itself.

jlaur: > You could call the file binding-information.xml for example.

That would be inconsistent as the other files are named 1:1 as their thing type id. I have prefixed all filenames with "thing-type-", let me know what you think.

lolodomo (approved):
LGTM

Issue #14256, opened 2023-01-21
[Netatmo] Weekly (Monthly) Min-Temperature not less then 0°C ?


LukaNoah:
I'm running OH 3.4.1 with the new netatmo-binding on a nuc with ubuntu 22.04 and java 11.

I have a problem with the generated minimum values for the outside temperature. The Min will not be less than 0°C. The daily min comes from the corresponding channel. Weeks and Months-Min are generated via the new ability to add channels.
Configured as a minimum temperature channel with a week or month period.
You can clearly see in the diagram that it was colder than 0 °C last night.

Thanks
Heiko
![image](https://user-images.githubusercontent.com/16850444/213876595-3b61f6dd-3339-44e2-a239-0963f9d1f663.png)

lolodomo:
This is also wrong currently for my feeling temperatures which should be -2 but is 0.

lolodomo:
Maybe INSIDE_TEMPERATURE is wrongly used? 
https://github.com/openhab/openhab-addons/blob/main/bundles/org.openhab.binding.netatmo/src/main/java/org/openhab/binding/netatmo/internal/api/data/NetatmoConstants.java#L81

lolodomo:
I hope the problem is not #13015...

lolodomo:
The feeling temperature is probably the heat index channel (I have to check). In this case, it is computed by the binding using the temperature and the humidity. We may simply use different formula than Netatmo.

LukaNoah:
The feels-like-outside-temperature is then green graph in my chart ;)
Also booth outside-temperatures are "broken".

clinique:
> I hope the problem is not #13015...

Could be. I'll have a look at it, it was my first thought also.

lolodomo:
> The feeling temperature is probably the heat index channel (I have to check => Edit: confirmed). In this case, it is computed by the binding using the temperature and the humidity. We may simply use different formula than Netatmo.

Just to conclude on that subject: the binding computes the heat index while Netatmo computes the feeling temperature which also considers the wind speed. So yes, this is two different data and that is normal they are not identical.

clinique:
> I hope the problem is not #13015...

You are correct. Problem lays here in `MeasureCapability.java`: 
```
                            MeasureClass.AS_SET.stream().filter(mc -> mc.apiDescriptor.equals(descriptor)).findFirst()
                                    .ifPresent(mc -> {
                                        State state = result instanceof ZonedDateTime
[...]

Issue #13015, opened 2022-06-26
[netatmo] Channel type netatmo:temp-measurement is build twice


lolodomo:
@clinique : I believe there is a small problem, you are building the channel type "netatmo:temp-measurement" twice, one for MeasureClass.INSIDE_TEMPERATURE and one for MeasureClass.OUTSIDE_TEMPERATURE. The difference between each of them is the min/max in state description.

As a consequence, the channel provider (NetatmoChannelTypeProvider) will provide only one of both for this UID, the first it finds.

I think this is necessary to have both with a different UID.

lolodomo:
To minimize the breaking changes that such a fix will imply, I suggest to change only MeasureClass.OUTSIDE_TEMPERATURE, meaning only the outside module will be potentially impacted.

clinique:
You're correct. Strictly said I just have used the enum name instead of the apiDescriptor to build the channel type ID. It would be the cleaner but it would break Inside Temp, Outside Temp, Rain Quantity and Humidity (preserving Pressure, CO2, Noise). Not the best choice now that 3.3 is out.

The other option would be to switch Inside Temp and Outside Temp, so the first that comes has the wider measurement range.

clinique:
Thinking of it twice, it is built twice but stored only once as it is in a Map. As Outside Temp comes in second, it is the one that is kept. Not very nice, but should work fine and prevent any bug.

lolodomo:
Does it mean @clinique we change nothing and close this issue ?

clinique:
I'm not sure it is worth the pain to change anything here.

PR #14265, opened 2023-01-23
[Netatmo] Preserving outside temperature minimal ranges for dynamic channels

clinique:
Bringing a solution to issue #14256.
This solution is not the best but takes in account observations made in issue #13015.

### DIFF
--- a/bundles/org.openhab.binding.netatmo/src/main/java/org/openhab/binding/netatmo/internal/handler/capability/MeasureCapability.java
+++ b/bundles/org.openhab.binding.netatmo/src/main/java/org/openhab/binding/netatmo/internal/handler/capability/MeasureCapability.java
@@ -69,26 +69,24 @@ public class MeasureCapability extends RestCapability<WeatherApi> {
         handler.getActiveChannels().filter(channel -> !channel.getConfiguration().getProperties().isEmpty())
                 .forEach(channel -> {
                     ChannelTypeUID channelTypeUID = channel.getChannelTypeUID();
-                    if (channelTypeUID != null) {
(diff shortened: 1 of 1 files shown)


bundles/org.openhab.binding.netatmo/src/main/java/org/openhab/binding/netatmo/internal/handler/capability/MeasureCapability.java
```diff
@@ -69,26 +69,24 @@ private void updateMeasures(WeatherApi api, String deviceId, @Nullable String mo
...
+                    }
+
+                    MeasureConfiguration measureDef = channel.getConfiguration().as(MeasureConfiguration.class);
+                    String descriptor = channelTypeUID.getId().split("-")[0];
+                    try {
+                        Object result = measureDef.limit.isBlank()
+                                ? api.getMeasures(deviceId, moduleId, measureDef.period, descriptor)
+                                : api.getMeasures(deviceId, moduleId, measureDef.period, descriptor, measureDef.limit);
+                        MeasureClass.AS_SET.stream().filter(mc -> mc.apiDescriptor.equals(descriptor))
+                                .reduce((first, second) -> second)
```
clinique, line 83: This is the important line

lolodomo (approved):
LGTM even if it looks more like a workaround than a clean solution.

clinique:
> LGTM even if it looks more like a workaround than a clean solution.

You're correct, this is a workaround.

lolodomo:
Temperature is negative now and I can confirm the fix is working.

PR #14270, opened 2023-01-24
(MyBMW] command "horn-blow" not "horn-low"

igitti:
The command is "horn-blow" not "horn-low".
Tested with my car: "horn-blow" works, "horn-low" does not.


lolodomo:
Looking very quickly at the code, I think there is another mismatch for another command: this is not "flash-lights" but "light-flash". Can you please fix that in README, it is only 4 lines above your current change.

bundles/org.openhab.binding.mybmw/README.md
```diff
@@ -356,7 +356,7 @@ The channel _command_ provides options
 - _vehicle-finder_
 - _door-lock_
 - _door-unlock_
-- _horn-low_
+- _horn-blow_
```
lolodomo, line 359: Looking very quickly at the code, I think there is another mismatch for another command: this is not "flash-lights" but "light-flash". Can you please fix that in README, it is only 4 lines above your current change.

lolodomo (changes requested):

bundles/org.openhab.binding.mybmw/README.md
igitti: Right, it must be "light-flash" not "flash-lights".
I fixed the README.md

lolodomo (approved):
LGTM, thank you

PR #14278, opened 2023-01-26
[jdbc] Perform upsert on conflict like the default for MySQL (PostgreSQL >=9.5)

henfiber:
This patch adds "upsert" behavior for PostgreSql >= 9.5, overwriting the previously stored value instead of raising an exception, when an existing timestamp is provided in a PUT request to the persistence endpoint.

This is the default behavior in the parent JdbcBaseDAO (and thus MySQL since it does not override it) and is equivalent to MySQL: `ON DUPLICATE KEY UPDATE VALUE`

see: https://www.postgresql.org/docs/9.5/sql-insert.html

This is only applied if the PostgreSQL version is >= 9.5, therefore it should not change the behavior for installations with legacy versions and should be backwards compatible with those versions. 

Since `ModifiablePersistenceService` was only recently implemented, I would not consider it a breaking change either, rather than implementing consistent behavior between two major backends.

### DIFF
--- a/bundles/org.openhab.persistence.jdbc/src/main/java/org/openhab/persistence/jdbc/internal/db/JdbcPostgresqlDAO.java
+++ b/bundles/org.openhab.persistence.jdbc/src/main/java/org/openhab/persistence/jdbc/internal/db/JdbcPostgresqlDAO.java
@@ -26,6 +26,7 @@ import org.openhab.core.types.State;
 import org.openhab.persistence.jdbc.internal.dto.ItemVO;
 import org.openhab.persistence.jdbc.internal.dto.ItemsVO;
 import org.openhab.persistence.jdbc.internal.exceptions.JdbcSQLException;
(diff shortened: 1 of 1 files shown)


henfiber:
> Please retarget the PR towards **main**.

Hello @jlaur, thanks for the quick response. 

I tried using the retargeting Github option (Edit > change target branch) but it failed with "This request took too long, please contact us".  Probably because the repo is quite large.

Is there any better way to do it? Or should I create a new PR with a commit on the main branch?

jlaur:
@henfiber - unfortunately I haven't tried this, so can't guide you. Perhaps the easiest solution would be to close the PR and create a new one with a branch relative to main (like you more or less said, check out main and cherry-pick commit).

henfiber:
Fixed the  signed-off-by issue, but will close this PR so I can target main. Thanks

PR #13765, opened 2022-11-22
[jdbc] Add console command for checking/repairing schema integrity

jlaur:
This pull request provides tools for detecting and repairing database schema inconsistencies which may have accumulated over time, when many things changed since tables were created. For example:
- Migration from MySQL persistence where `DATETIME` was used for the time column, whereas this is now `TIMESTAMP(3)`.
- Bugfixes extending or otherwise changing column types, for example #10542 and #9394.
- Manual corrections which are inconsistent with the JDBC types.

### Example

```
openhab> openhab:jdbc schema check
Table        Item         Issue
-----------  -----------  ----------------------------------------------------------------
Anyone_Home  Anyone_Home  Column name 'time' expected, but is 'Time'
Anyone_Home  Anyone_Home  Column type 'TIMESTAMP(3)' expected, but is 'DATETIME'
Anyone_Home  Anyone_Home  Column name 'value' expected, but is 'Value'
Anyone_Home  Anyone_Home  Column type 'VARCHAR(6)' expected, but is 'CHAR(3)'
openhab> openhab:jdbc schema fix
Fixed table 'Anyone_Home' for item 'Anyone_Home'
```

### Test

JAR for testing: [org.openhab.persistence.jdbc-3.4.0-SNAPSHOT.jar](https://drive.google.com/file/d/1aEq2eyimz0V4JfRqBtq98dQ1MKkLNeZy/view?usp=sharing)

Place the corresponding database driver into the addons directory. For example, download the [MySQL connector](https://repo1.maven.org/maven2/mysql/mysql-connector-java/8.0.30/mysql-connector-java-8.0.30.jar) and rename it to `mysql-connector-java.jar` or the [MariaDB client](https://repo1.maven.org/maven2/org/mariadb/jdbc/mariadb-java-client/3.0.8/mariadb-java-client-3.0.8.jar) and rename it to `mariadb-java-client.jar`.

Resolves #13712

PR #14281, opened 2023-01-26
[jdbc] Perform upsert on conflict like the default for MySQL (PostgreSQL >=9.5)

henfiber:
This patch adds "upsert" behavior for PostgreSql >= 9.5, overwriting the previously stored value instead of raising an exception, when an existing timestamp is provided in a PUT request to the persistence endpoint.

This is the default behavior in the parent JdbcBaseDAO (and thus MySQL since it does not override it) and is equivalent to MySQL: `ON DUPLICATE KEY UPDATE VALUE`

see: https://www.postgresql.org/docs/9.5/sql-insert.html

This is only applied if the PostgreSQL version is >= 9.5, therefore it should not change the behavior for installations with legacy versions and should be backwards compatible with those versions.

Since `ModifiablePersistenceService` was only recently implemented, I would not consider it a breaking change either, rather than implementing consistent behavior between two major backends.

### DIFF
--- a/bundles/org.openhab.persistence.jdbc/src/main/java/org/openhab/persistence/jdbc/internal/db/JdbcPostgresqlDAO.java
+++ b/bundles/org.openhab.persistence.jdbc/src/main/java/org/openhab/persistence/jdbc/internal/db/JdbcPostgresqlDAO.java
@@ -26,6 +26,7 @@ import org.openhab.core.types.State;
 import org.openhab.persistence.jdbc.internal.dto.ItemVO;
 import org.openhab.persistence.jdbc.internal.dto.ItemsVO;
 import org.openhab.persistence.jdbc.internal.exceptions.JdbcSQLException;
+import org.openhab.persistence.jdbc.internal.utils.DbMetaData;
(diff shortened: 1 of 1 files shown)


jlaur (approved):
Thanks for the fix!

henfiber:
Thanks for reviewing and merging. 

I will probably work more on the jdbc peristence addon, there are some inconsistencies in the handling of timestamps and time zones (I see also related open issues).

jlaur:
> I will probably work more on the jdbc peristence addon, there are some inconsistencies in the handling of timestamps and time zones (I see also related open issues).

That would be a very welcome addition. I recently did some work on JDBC persistence, but have been taking a break from it since the 3.4 release. 🙂 Was looking at #13775 most recently, you might want to check out the discussions there. AFAIR we didn't quite reach a conclusion.

henfiber:
I am not very experienced with MySQL so I am not sure I can help with this complex issue you have with the session and driver timestamps. 

I added a fix to the PostgreSQL service for now and added a PR here: https://github.com/openhab/openhab-addons/pull/14294 which also adds PG support to the console functionality you introduced at https://github.com/openhab/openhab-addons/pull/13765

PR #12949, opened 2022-06-16
[boschindego] Add localization support for textual states

jlaur:
Add localization support for textual states:
- **Non-redundancy:** State options for **statecode** uses the same texts as **textualstate**.
- **Normalization:** Only unique texts are provided, so duplicate states reference the same text.

Fixes #12941

PR #14282, opened 2023-01-26
[androidtv] AndroidTV Binding initial contribution

### LINKED ISSUE #14315: New binding for AndroidTV
As per guidelines, this is the enhancement associated with #14282 for the AndroidTV binding.

morph166955:
This is the initial contribution of the AndroidTV binding.  AndroidTV has a unique place in the market where it exists in several different vendors in the market.  This creates the basic binding framework and adds the Nvidia ShieldTV protocol (which was reverse engineered from the Nvidia ShieldTV Android App) and version 2 of the GoogleTV protocol (which was reverse engineered from the GoogleTV Android App).  Discovery is provided by mDNS.

All devices use the GoogleTV protocol as the base for command/control/information.  The Nvidia ShieldTV adds the ShieldTV protocol on top for additional metadata and application control.

This version provides connectivity, authentication via the on-screen PIN based process, PKI support for certificates, keypress support for all identified keys, and app control/awareness/information.

This is a follow-on to the original shieldtv PR (#13934) which was closed in favor of a more encompassing androidtv binding to better support the community.

Fixes #14315


morph166955:
A test jar is available at https://github.com/morph166955/openhab-addons/releases/download/androidtv-beta/org.openhab.binding.androidtv-4.0.0-SNAPSHOT-92d3ae3.jar

morph166955:
Created: 

https://github.com/openhab/openhab-docs/pull/2001

and

https://github.com/openhab/openhab-addons/issues/14315

As per guidelines

morph166955:
This is ready for final review.  The next changes to be made will be to add googletv which I won't be doing in the initial PR (but is started for those interested).

morph166955:
I'm not sure why DCO is broken on that commit. Everything looks clean to me.

morph166955:
@openhab/add-ons-maintainers Respectfully requesting a review on this binding. I'd really like to get this into the 4.0.0 release. Thank you in advance for your time!

morph166955:
As the anthem binding was built off of the base of this binding, I've pulled in the relevant changes in from the review that @jlaur did on the anthem binding prior to it being merged https://github.com/openhab/openhab-addons/pull/14311 as of the last commit in hopes that this reduces the needed changes on this binding.  I also resolved all sat and build warnings.

In respect to the review comments about the connection managers on the anthem binding, it would be too complex to remove the connection managers from this binding based on how it's structured across different thing types.

bundles/org.openhab.binding.androidtv/pom.xml
```diff
@@ -0,0 +1,39 @@
...
+
+  <name>openHAB Add-ons :: Bundles :: AndroidTV Binding</name>
+
+  <properties>
+    <ohc.version>3.2.0</ohc.version>
+    <karaf.version>4.3.7</karaf.version>
+    <maven.compiler.source>11</maven.compiler.source>
+    <maven.compiler.target>11</maven.compiler.target>
+  </properties>
+
```
wborn, line 23:
```suggestion
```

bundles/org.openhab.binding.androidtv/src/main/resources/OH-INF/binding/binding.xml
wborn, line 1: This file is no longer used in OH 4.

bundles/org.openhab.binding.androidtv/README.md
```diff
@@ -0,0 +1,478 @@
...
+This binding supports two thing types:
+
+- **googletv** - An AndroidTV running Google Video
+- **shieldtv** - An Nvidia ShieldTV
+
+## Discovery
+
+Both GoogleTVs and ShieldTVs should be added automatically to the inbox through the mDNS discovery process.  
+
+In the case of the ShieldTV, OpenHAB will likely create an inbox entry for both a GoogleTV and a ShieldTV device.
```
wborn, line 18:
```suggestion
In the case of the ShieldTV, openHAB will likely create an inbox entry for both a GoogleTV and a ShieldTV device.
```

bundles/org.openhab.binding.androidtv/README.md
```diff
@@ -0,0 +1,478 @@
...
+Only the ShieldTV device should be configured, the GoogleTV can be ignored.
+There is no benefit to configuring two things for a ShieldTV device.  This could cause undesired effects.
```
wborn, line 20:
```suggestion
There is no benefit to configuring two things for a ShieldTV device.
This could cause undesired effects.
```

bundles/org.openhab.binding.androidtv/README.md
```diff
@@ -0,0 +1,478 @@
...
+If you are on a ShieldTV you must run that process a second time to authenticate the GoogleTV protocol stack.
+
+This completes the PIN process.  Upon reconnection (either from reconfiguration or a restart of OpenHAB), you should now see a message of "Login Successful" in openhab.log
+
+## Full Example
+
+```java
+Thing androidtv:shieldtv:livingroom [ ipAddress="192.168.1.2" ]
+Thing androidtv:googletv:theater [ ipAddress="192.168.1.3" ]
+
```
wborn, line 142:
```suggestion
```

bundles/org.openhab.binding.androidtv/README.md
```diff
@@ -0,0 +1,478 @@
...
+You may also send it a command of the app package name (e.g. com.google.android.youtube.tv) to start/change-to that app.
+
+KEYCODE values are listed at the bottom of this README.
+NOTE: Not all KEYCODES work on all devices.  Keycodes above 255 have not been tested.
+
+## Pin Code Process
+
+For the AndroidTV to be successfully accessed an on-screen PIN authentication is required on the first connection.  
+
+To begin the PIN process, send the text "REQUEST" to the pincode channel while watching your AndroidTV.  A 6 digit PIN should be displayed on the screen.
```
wborn, line 129:
```suggestion
To begin the PIN process, send the text "REQUEST" to the pincode channel while watching your AndroidTV.
A 6 digit PIN should be displayed on the screen.
```

bundles/org.openhab.binding.androidtv/README.md
```diff
@@ -0,0 +1,478 @@
...
+
+To complete the PIN process, send the PIN displayed to the pincode channel.  The display should return back to where it was originally.
```
wborn, line 131:
```suggestion
To complete the PIN process, send the PIN displayed to the pincode channel.
The display should return back to where it was originally.
```

bundles/org.openhab.binding.androidtv/README.md
```diff
@@ -0,0 +1,478 @@
...
+
+If you are on a ShieldTV you must run that process a second time to authenticate the GoogleTV protocol stack.
+
+This completes the PIN process.  Upon reconnection (either from reconfiguration or a restart of OpenHAB), you should now see a message of "Login Successful" in openhab.log
```
wborn, line 135:
```suggestion
This completes the PIN process.
Upon reconnection (either from reconfiguration or a restart of openHAB), you should now see a message of "Login Successful" in openhab.log
```

bundles/org.openhab.binding.androidtv/src/main/java/org/openhab/binding/androidtv/internal/AndroidTVBindingConstants.java
```diff
@@ -0,0 +1,63 @@
...
+public class AndroidTVBindingConstants {
+
+    private static final String BINDING_ID = "androidtv";
+
+    // List of all Thing Type UIDs
+    public static final ThingTypeUID THING_TYPE_GOOGLETV = new ThingTypeUID(BINDING_ID, "googletv");
+    public static final ThingTypeUID THING_TYPE_SHIELDTV = new ThingTypeUID(BINDING_ID, "shieldtv");
+
+    public static final Set<ThingTypeUID> SUPPORTED_THING_TYPES = Collections
+            .unmodifiableSet(Stream.of(THING_TYPE_GOOGLETV, THING_TYPE_SHIELDTV).collect(Collectors.toSet()));
```
wborn, line 39:
```suggestion
    public static final Set<ThingTypeUID> SUPPORTED_THING_TYPES = Set.of(THING_TYPE_GOOGLETV, THING_TYPE_SHIELDTV);
```

bundles/org.openhab.binding.androidtv/src/main/java/org/openhab/binding/androidtv/internal/discovery/GoogleTVDiscoveryParticipant.java
```diff
@@ -0,0 +1,100 @@
...
+    }
+
+    @Override
+    public String getServiceType() {
+        return GOOGLETV_MDNS_SERVICE_TYPE;
+    }
+
+    @Nullable
+    @Override
+    public DiscoveryResult createResult(@Nullable ServiceInfo service) {
```
wborn, line 59:
```suggestion
    @Override
    public @Nullable DiscoveryResult createResult(@Nullable ServiceInfo service) {
```

bundles/org.openhab.binding.androidtv/src/main/java/org/openhab/binding/androidtv/internal/discovery/GoogleTVDiscoveryParticipant.java
```diff
@@ -0,0 +1,100 @@
...
+
+            return DiscoveryResultBuilder.create(uid).withProperties(properties).withLabel(label).build();
+        } else {
+            return null;
+        }
+    }
+
+    @Nullable
+    @Override
+    public ThingUID getThingUID(@Nullable ServiceInfo service) {
```
wborn, line 93:
```suggestion
    @Override
    public @Nullable ThingUID getThingUID(@Nullable ServiceInfo service) {
```

bundles/org.openhab.binding.androidtv/src/main/java/org/openhab/binding/androidtv/internal/discovery/ShieldTVDiscoveryParticipant.java
```diff
@@ -0,0 +1,102 @@
...
+    }
+
+    @Override
+    public String getServiceType() {
+        return SHIELDTV_MDNS_SERVICE_TYPE;
+    }
+
+    @Nullable
+    @Override
+    public DiscoveryResult createResult(@Nullable ServiceInfo service) {
```
wborn, line 59:
```suggestion
    @Override
    public @Nullable DiscoveryResult createResult(@Nullable ServiceInfo service) {
```

bundles/org.openhab.binding.androidtv/src/main/java/org/openhab/binding/androidtv/internal/discovery/ShieldTVDiscoveryParticipant.java
```diff
@@ -0,0 +1,102 @@
...
+        if ((service == null) || !service.hasData()) {
```
wborn, line 60:
```suggestion
        if (service == null || !service.hasData()) {
```

bundles/org.openhab.binding.androidtv/src/main/java/org/openhab/binding/androidtv/internal/discovery/ShieldTVDiscoveryParticipant.java
```diff
@@ -0,0 +1,102 @@
...
+
+            return DiscoveryResultBuilder.create(uid).withProperties(properties).withLabel(label).build();
+        } else {
+            return null;
+        }
+    }
+
+    @Nullable
+    @Override
+    public ThingUID getThingUID(@Nullable ServiceInfo service) {
```
wborn, line 95:
```suggestion
    @Override
    public @Nullable ThingUID getThingUID(@Nullable ServiceInfo service) {
```

bundles/org.openhab.binding.androidtv/src/main/java/org/openhab/binding/androidtv/internal/discovery/ShieldTVDiscoveryParticipant.java
```diff
@@ -0,0 +1,102 @@
...
+        if ((service == null) || !service.hasData()) {
```
wborn, line 96:
```suggestion
        if (service == null || !service.hasData()) {
```

bundles/org.openhab.binding.androidtv/src/main/resources/OH-INF/thing/thing-types.xml
```diff
@@ -0,0 +1,188 @@
...
+
+		<properties>
+			<property name="Device Name">unknown</property>
+			<property name="Device ID">unknown</property>
+			<property name="Architectures">unknown</property>
+			<property name="Manufacturer">unknown</property>
+			<property name="Model">unknown</property>
+			<property name="Android Version">unknown</property>
+			<property name="Remote Server">unknown</property>
+			<property name="Remote Server Version">unknown</property>
```
wborn, line 35: Please use camel-case for the property names, see:

https://next.openhab.org/docs/developer/guidelines.html#naming-convention

bundles/org.openhab.binding.androidtv/src/main/resources/OH-INF/thing/thing-types.xml
```diff
@@ -0,0 +1,188 @@
...
+			<channel id="volume" typeId="system.volume"/>
+			<channel id="mute" typeId="system.mute"/>
+		</channels>
+
+		<properties>
+			<property name="Manufacturer">unknown</property>
+			<property name="Model">unknown</property>
+			<property name="Android Version">unknown</property>
+			<property name="Remote Server">unknown</property>
+			<property name="Remote Server Version">unknown</property>
```
wborn, line 96: Please use camel-case for the property names, see:

https://next.openhab.org/docs/developer/guidelines.html#naming-convention

wborn (changes requested):
Thank your for adding the binding. :+1: The code looks already quite good!
I've added a few comments below:

morph166955:
@wborn I'm just pushed a commit to resolve the majority of the comments.  I am getting build errors on removing the @Nullable from the discovery bits.

```
[ERROR] /home/openhab/androidtv/openhab-addons/bundles/org.openhab.binding.androidtv/src/main/java/org/openhab/binding/androidtv/internal/discovery/ShieldTVDiscoveryParticipant.java:[58,12] The default '@org.eclipse.jdt.annotation.NonNull' conflicts with the inherited '@org.eclipse.jdt.annotation.Nullable' annotation in the overridden method from org.openhab.core.config.discovery.mdns.MDNSDiscoveryParticipant
[ERROR] /home/openhab/androidtv/openhab-addons/bundles/org.openhab.binding.androidtv/src/main/java/org/openhab/binding/androidtv/internal/discovery/ShieldTVDiscoveryParticipant.java:[93,12] The default '@org.eclipse.jdt.annotation.NonNull' conflicts with the inherited '@org.eclipse.jdt.annotation.Nullable' annotation in the overridden method from org.openhab.core.config.discovery.mdns.MDNSDiscoveryParticipant
[ERROR] /home/openhab/androidtv/openhab-addons/bundles/org.openhab.binding.androidtv/src/main/java/org/openhab/binding/androidtv/internal/discovery/GoogleTVDiscoveryParticipant.java:[58,12] The default '@org.eclipse.jdt.annotation.NonNull' conflicts with the inherited '@org.eclipse.jdt.annotation.Nullable' annotation in the overridden method from org.openhab.core.config.discovery.mdns.MDNSDiscoveryParticipant
[ERROR] /home/openhab/androidtv/openhab-addons/bundles/org.openhab.binding.androidtv/src/main/java/org/openhab/binding/androidtv/internal/discovery/GoogleTVDiscoveryParticipant.java:[91,12] The default '@org.eclipse.jdt.annotation.NonNull' conflicts with the inherited '@org.eclipse.jdt.annotation.Nullable' annotation in the overridden method from org.openhab.core.config.discovery.mdns.MDNSDiscoveryParticipant

```

Otherwise everything else looks fine.

morph166955:
I did not notice before that you moved the Nullable, I thought you had removed it. I'll push a commit tonight to move them. That should account for the remainder of the requested changes.

bundles/org.openhab.binding.androidtv/src/main/java/org/openhab/binding/androidtv/internal/AndroidTVHandler.java
```diff
@@ -0,0 +1,261 @@
...
+        if (failed) {
+            updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.NONE, statusMessage);
+        } else {
+            updateStatus(ThingStatus.ONLINE);
+        }
+    }
+
+    @Override
+    public void initialize() {
+        updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.NONE, "Protocols Starting");
```
jlaur, line 143: Consider adding an I18N string here, see https://www.openhab.org/docs/developer/utils/i18n.html#using-custom-keys.

bundles/org.openhab.binding.androidtv/src/main/java/org/openhab/binding/androidtv/internal/AndroidTVHandler.java
```diff
@@ -0,0 +1,261 @@
...
+
+        GoogleTVConfiguration googletvConfig = getConfigAs(GoogleTVConfiguration.class);
+        String ipAddress = googletvConfig.ipAddress;
+
+        if (ipAddress == null || ipAddress.isEmpty()) {
+            updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.CONFIGURATION_ERROR, "googletv address not specified");
```
jlaur, line 149: Consider I18N.

bundles/org.openhab.binding.androidtv/src/main/java/org/openhab/binding/androidtv/internal/AndroidTVHandler.java
```diff
@@ -0,0 +1,261 @@
...
+
+        googletvConnectionManager = new GoogleTVConnectionManager(this, googletvConfig);
+
+        if (THING_TYPE_SHIELDTV.equals(thingTypeUID)) {
+            ShieldTVConfiguration shieldtvConfig = getConfigAs(ShieldTVConfiguration.class);
+            ipAddress = shieldtvConfig.ipAddress;
+
+            if (ipAddress == null || ipAddress.isEmpty()) {
+                updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.CONFIGURATION_ERROR,
+                        "shieldtv address not specified");
```
jlaur, line 161: Consider I18N.

bundles/org.openhab.binding.androidtv/src/main/java/org/openhab/binding/androidtv/internal/protocol/googletv/GoogleTVConfiguration.java
```diff
@@ -0,0 +1,36 @@
...
+
+/**
+ * The {@link GoogleTVConfiguration} class contains fields mapping thing configuration parameters.
+ *
+ * @author Ben Rosenblum - Initial contribution
+ */
+@NonNullByDefault
+public class GoogleTVConfiguration {
+
+    public @Nullable String ipAddress;
```
jlaur, line 26: This would simplify the validation as null wouldn't have to be considered:
```suggestion
    public String ipAddress = "";
```

bundles/org.openhab.binding.androidtv/src/main/java/org/openhab/binding/androidtv/internal/AndroidTVHandler.java
```diff
@@ -0,0 +1,261 @@
...
+    }
+
+    @Override
+    public void initialize() {
+        updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.NONE, "Protocols Starting");
+
+        GoogleTVConfiguration googletvConfig = getConfigAs(GoogleTVConfiguration.class);
+        String ipAddress = googletvConfig.ipAddress;
+
+        if (ipAddress == null || ipAddress.isEmpty()) {
```
jlaur, line 148: See comment in `GoogleTVConfiguration`:
```suggestion
        if (ipAddress.isBlank()) {
```

bundles/org.openhab.binding.androidtv/src/main/resources/OH-INF/i18n/androidtv.properties
```diff
@@ -0,0 +1,73 @@
+# binding
+
+binding.androidtv.name = AndroidTV Binding
+binding.androidtv.description = This is the binding for AndroidTV.
+
```
jlaur, line 5: This can be removed.

bundles/org.openhab.binding.androidtv/src/main/resources/OH-INF/i18n/androidtv.properties
```diff
@@ -0,0 +1,73 @@
...
+channel-type.androidtv.keycode.description = Send keycode
+channel-type.androidtv.keypress.label = Key Press
+channel-type.androidtv.keypress.description = Send key press
+channel-type.androidtv.pincode.label = Pin Code
+channel-type.androidtv.pincode.description = Send Pin Code
+
+# add-on
+
+addon.androidtv.name = AndroidTV Binding
+addon.androidtv.description = This is the addon for AndroidTV.
```
jlaur, line 66: Usually this would be found at the top of the file. Consider recreating with `mvn i18n:generate-default-translations`.

bundles/org.openhab.binding.androidtv/src/main/resources/OH-INF/i18n/androidtv.properties
```diff
@@ -0,0 +1,73 @@
...
+
+# channel types
+
+channel-type.androidtv.raw.label = RAW Command
+channel-type.androidtv.raw.description = RAW command to send (for debugging)
+channel-type.androidtv.rawmsg.label = RAW Message
+channel-type.androidtv.rawmsg.description = RAW message to handle (for debugging)
```
jlaur, line 73: Are these still relevant, or why are they at the bottom and not with the others?

bundles/org.openhab.binding.androidtv/src/main/resources/OH-INF/thing/thing-types.xml
```diff
@@ -0,0 +1,188 @@
...
+				<context>password</context>
+				<label>Keystore Password</label>
+				<description>Password for the keystore file</description>
+			</parameter>
+			<parameter name="reconnect" type="integer">
+				<label>Reconnect Delay</label>
+				<description>Delay between reconnection attempts</description>
+			</parameter>
+			<parameter name="heartbeat" type="integer">
+				<label>Hearbeat Frequency</label>
```
jlaur, line 123:
```suggestion
				<label>Heartbeat Frequency</label>
```

bundles/org.openhab.binding.androidtv/src/main/java/org/openhab/binding/androidtv/internal/protocol/shieldtv/ShieldTVConfiguration.java
```diff
@@ -0,0 +1,35 @@
...
+
+/**
+ * The {@link ShieldTVConfiguration} class contains fields mapping thing configuration parameters.
+ *
+ * @author Ben Rosenblum - Initial contribution
+ */
+@NonNullByDefault
+public class ShieldTVConfiguration {
+
+    public @Nullable String ipAddress;
```
jlaur, line 26:
```suggestion
    public String ipAddress = "";
```

bundles/org.openhab.binding.androidtv/src/main/java/org/openhab/binding/androidtv/internal/discovery/ShieldTVDiscoveryParticipant.java
```diff
@@ -0,0 +1,100 @@
...
+
+        final ThingUID uid = getThingUID(service);
+        if (uid != null) {
+            final String id = uid.getId();
+            final String label = service.getName() + " (" + id + ")";
+            final Map<String, Object> properties = new HashMap<>(2);
+
+            properties.put(IPADDRESS, ipAddress);
+
+            return DiscoveryResultBuilder.create(uid).withProperties(properties).withLabel(label).build();
```
jlaur, line 86: Representation property missing, see https://www.openhab.org/docs/developer/bindings/thing-xml.html#representation-property-and-discovery

jlaur (changes requested):
Thanks for contributing this binding! I had my eyes on it, but didn't yet find the time to do a review, so I'm glad @wborn did. I see now that I had a few initial comments pending, so I'm posting them now, hoping that they might be useful.

morph166955:
@jlaur Thank you for the additional review!  I'll try to push a commit to fix the remaining issues tonight or tomorrow.  I'll check on the i18n bits but I've honestly only ever really run the mvn command for i18n so I'm not entirely sure what else I need to do with adding the custom keys. 

I'll put a few comments in the review on the questions specific at a few of the lines. 

Thanks again!

bundles/org.openhab.binding.androidtv/src/main/java/org/openhab/binding/androidtv/internal/AndroidTVBindingConstants.java
```diff
@@ -0,0 +1,62 @@
...
+    public static final String CHANNEL_PINCODE = "pincode";
+    public static final String CHANNEL_APP = "app";
+    public static final String CHANNEL_APPNAME = "appname";
+    public static final String CHANNEL_APPURL = "appurl";
+    public static final String CHANNEL_POWER = "power";
+    public static final String CHANNEL_VOLUME = "volume";
+    public static final String CHANNEL_MUTE = "mute";
+
+    // List of all config properties
+    public static final String IPADDRESS = "ipAddress";
```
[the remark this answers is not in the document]
morph166955: Can you be more specific on what you mean here?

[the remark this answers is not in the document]
jlaur: I would propose `PROPERTY_IP_ADDRESS` to keep existing convention. See for example: https://www.openhab.org/javadoc/latest/org/openhab/core/thing/thing#field-summary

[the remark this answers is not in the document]
jlaur: - assuming this is a property, not a configuration parameter? Otherwise, sorry.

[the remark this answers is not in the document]
morph166955: I did a quick pass on my phone and this may be a mute point. I think I actually stopped using these in the code all together and they may just be able to be deleted.  That said I'd think I'd get an unused alert somewhere if I wasn't. I'll do a grep when I'm at a computer to verify. If not, I'll address it as requested.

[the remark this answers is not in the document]
morph166955: Found it in the two discovery files once.  I updated it.

morph166955:
@jlaur I've pushed a commit to address all but the i18n strings.  In reality there are probably a bunch more of those buried in the connection managers and in other places that should be added too.  It will take longer than I have tonight to find an update all of those, and I may not have time over the next few days.  Would it be OK with you if I submit a follow-on PR to address the i18n strings specifically sometime in the next week or two?

morph166955:
Also to note, CI seems to be failing but not due to anything relating to the binding.  It compiles clean on my side:

```
downloading https://repo.maven.apache.org/maven2/org/apache/maven/apache-maven/3.8.6/apache-maven-3.8.6-bin.tar.gz
Error: getaddrinfo ENOTFOUND repo.maven.apache.org
```

jlaur:
> Would it be OK with you if I submit a follow-on PR to address the i18n strings specifically sometime in the next week or two?

For me okay, but I'm leaving it to @wborn since I did not read through the code, but only provided a few loose comments I had prepared.

But if we are talking a week or two, wouldn't you rather have the chance to finish in context of this PR? The only difference will be a few more weeks of maintaining the JARs for the community published version rather than having it included in the snapshot builds. Also, translators will have all strings available at once rather than having to come back and translate remaining strings later. Just for consideration...

morph166955:
I'm fine with either option. Just didn't want to hold things up from my side. It will definitely be next week before I get the strings in.

morph166955:
@jlaur Help me understand how the i18n works.  The way I have the thing status structured for all but the very first few updates in the handler, is that the handler calls getStatusMessage() from each of the connectionManagers and aggregates that together.  That's done via a scheduler every 250ms by calling...

https://github.com/openhab/openhab-addons/blob/8fe68d2d152e4cc8622acf48834e19d8ffd32b62/bundles/org.openhab.binding.androidtv/src/main/java/org/openhab/binding/androidtv/internal/AndroidTVHandler.java#L107

So, do I put the @text/my_status_message into the string in the connection manager?

jlaur:
> Help me understand how the i18n works. The way I have the thing status structured for all but the very first few updates in the handler, is that the handler calls getStatusMessage() from each of the connectionManagers and aggregates that together into a singular update. That's done via a scheduler every 250ms by calling...

Well, you cannot concatenate I18N strings, but you can use parameter substitution. See for example:

https://github.com/openhab/openhab-addons/blob/fa57c627a99974765213ec81f372fc5462d2beaf/bundles/org.openhab.binding.danfossairunit/src/main/java/org/openhab/binding/danfossairunit/internal/handler/DanfossAirUnitHandler.java#L107-L108

and:

https://github.com/openhab/openhab-addons/blob/fa57c627a99974765213ec81f372fc5462d2beaf/bundles/org.openhab.binding.danfossairunit/src/main/resources/OH-INF/i18n/danfossairunit.properties#L88

bundles/org.openhab.binding.androidtv/README.md
```diff
@@ -0,0 +1,484 @@
...
+
+To complete the PIN process, send the PIN displayed to the pincode channel.
+
+The display should return back to where it was originally.
+
+If you are on a ShieldTV you must run that process a second time to authenticate the GoogleTV protocol stack.
+
+This completes the PIN process.
+
+Upon reconnection (either from reconfiguration or a restart of OpenHAB), you should now see a message of "Login Successful" in openhab.log
```
jlaur, line 142: Just a quick comment for this process: Consider adding a console command for providing the PIN. It seems a bit cumbersome to temporarily create an item in order link it to a channel, just to delete it again afterwards.

I admit it might be a matter of preference. Unfortunately I can't think of other similar processes right now and how they are implemented, except perhaps OAuth2 flow where servlets are often used.

morph166955:
Interesting. I'm not sure how to do that and achieve the same goal with providing information on all of the protocols in the message. It's not horrible for the googletv only devices, but the shieldtv presents a problem. It's entirely possible for a status to be "GoogleTV: ONLINE | ShieldTV: Connecting..."  the handler in that case is pulling a string from each connection manager and concatenating them.

bundles/org.openhab.binding.androidtv/README.md
morph166955: I'm not sure how to do that. I use the item/channel via karaf personally. E.g. "openhab:send androidtv_pin REQUEST".

morph166955:
Ok, so in theory if I did the getStatusMessage, put them into a temporary string, ran them through i18n, and then concatenated a final message that would work?  How does one call i18n to do that string fetching?

bundles/org.openhab.binding.androidtv/README.md
jlaur: If already using the console, it would be more convenient with a custom command for this to avoid the need for a linked item. See for example:

https://github.com/openhab/openhab-addons/blob/main/bundles/org.openhab.binding.boschindego/src/main/java/org/openhab/binding/boschindego/internal/console/BoschIndegoCommandExtension.java

Documented as last step here:

https://github.com/openhab/openhab-addons/blob/main/bundles/org.openhab.binding.boschindego/README.md#authorization

jlaur:
> Ok, so in theory if I did the getStatusMessage, put them into a temporary string, ran them through i18n, and then concatenated a final message that would work? How does one call i18n to do that string fetching?

You would need `TranslationProvider` and `LocaleProvider` injected from the handler factory, then create some logic using those. See for example #12949.

morph166955:
Let me give that a try. Doesn't sound horrible.

bundles/org.openhab.binding.androidtv/README.md
morph166955: I'll give that a try. Standby.

morph166955: Is there any reason I couldn't just use the CommandExtension as a go-between to the handleCommand() already existing in the handler?  There is a good bit of logic there already to determine what to send to where on each of the connection managers.  It would be trivial if I could just send a "fake command" to handleCommand().

morph166955:
On second thought, and after looking more at that PR you posted, I'd really like to push that into a second PR.  That's a semi-significant add that I would want to do more testing on to make sure it isn't going to cause issues across the different platforms that this supports.  I don't mind doing it, but I'd really rather not push this PR off while waiting to add another component to it.

jlaur:
> On second thought, and after looking more at that PR you posted, I'd really like to push that into a second PR. That's a semi-significant add that I would want to do more testing on to make sure it isn't going to cause issues across the different platforms that this supports. I don't mind doing it, but I'd really rather not push this PR off while waiting to add another component to it.

Are you referring to I18N or the PIN pairing process using a console command?

bundles/org.openhab.binding.androidtv/README.md
morph166955: I looked through what you posted.  Similar to the TranslationProvider, no issues adding this but I'd rather not make large changes to the PR at this point.  I like the idea, I have no problem saying I'll implement it in short order, but I'd rather do it as a small PR next.  After going back to look at the logic in the handler, I'd have to pull a few things out in the connection managers to create a pin function that I could call directly instead of having it buried in the handleCommand() logic.  This is going to need additional testing as well to determine that we don't have a regression across the different hardware platforms that this supports.

morph166955:
> Are you referring to I18N or the PIN pairing process using a console command?

I think we crossed streams.  Both actually now that I've had time to look at it.  While they aren't massive changes, adding both would need a good bit of testing to ensure regressions didn't happen.  PIN is VERY sensitive and breaks very easily.  I really don't want to introduce changes at the 11th hour and potentially commit something unstable.  No issues doing it, but I'd prefer to do each as a separate PR after this to reduce risk.

morph166955:
On that note, @wborn - standing by here, I believe I've covered everything requested in your review.

morph166955:
I just pushed two commits to fix the issue below which was posted in the marketplace.

```
2023-05-03 04:28:20.262 [ERROR] [y.mdns.internal.MDNSDiscoveryService] - Participant 'org.openhab.binding.androidtv.internal.discovery.GoogleTVDiscoveryParticipant' threw an exception
java.lang.ArrayIndexOutOfBoundsException: Index 0 out of bounds for length 0
	at org.openhab.binding.androidtv.internal.discovery.GoogleTVDiscoveryParticipant.createResult(GoogleTVDiscoveryParticipant.java:67) ~[?:?]
	at org.openhab.core.config.discovery.mdns.internal.MDNSDiscoveryService.createDiscoveryResult(MDNSDiscoveryService.java:222) ~[?:?]
	at org.openhab.core.config.discovery.mdns.internal.MDNSDiscoveryService.considerService(MDNSDiscoveryService.java:214) ~[?:?]
	at org.openhab.core.config.discovery.mdns.internal.MDNSDiscoveryService.serviceResolved(MDNSDiscoveryService.java:207) ~[?:?]
	at javax.jmdns.impl.ListenerStatus$ServiceListenerStatus.serviceResolved(ListenerStatus.java:106) ~[?:?]
	at javax.jmdns.impl.JmDNSImpl$1.run(JmDNSImpl.java:911) ~[?:?]
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1136) ~[?:?]
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:635) ~[?:?]
	at java.lang.Thread.run(Thread.java:833) ~[?:?]
```

morph166955:
@wborn Following up on this so we can get it merged in prior to the 4.0 release.  I believe all of the changes requested are completed (minus the two from @jlaur which I will do in follow-on PRs as they require additional time/testing).  Please let me know if there is anything else required on my end.  Thanks!

morph166955:
@openhab/add-ons-maintainers Respectfully requesting this PR be merged in. Reviews are complete. Binding use in the marketplace has been very stable for months now. I have several things I'd like to do still with this binding that I'm holding off on to avoid dealing with rebasing. Thank you!

jlaur:
Since I have posted only some initial comments I prepared in March, but haven't reviewed the binding (see https://github.com/openhab/openhab-addons/pull/14282#pullrequestreview-1359106595), I would prefer if @wborn could check if review comments are addressed and conclude the review.

morph166955:
> I'm OK with this, although it would be preferable to do now to avoid update instructions for removing the obsolete channel again.

Just to be clear on my intentions, I'm fine with adding the CLI commands as an alternative however I'm going to leave the channel in place as well as some users do prefer this method. Also, as this has been on the marketplace for 6 months now we would already have to update there anyway.

morph166955:
@wborn checking in again on this. We're getting close to the code getting locked for the 4.0.0 release and I'd really like to get this merged in for that. Thanks!

morph166955:
Google updated the GoogleTV protocol to version 4.39.  Two minor commits have be pushed to update the login string to resolve an issue posted to the forum and to make a few small changes to the checks and logging for the PIN process.

jlaur:
@wborn - what is the status of your review?

morph166955:
@wborn Please let me know if there's anything else for this. Thanks!

bundles/org.openhab.binding.androidtv/src/main/java/org/openhab/binding/androidtv/internal/protocol/googletv/GoogleTVConnectionManager.java
```diff
@@ -0,0 +1,1383 @@
...
+                logger.trace("Opening GoogleTV shim on port {}", config.port);
+                SSLServerSocket sslServerSocket = (SSLServerSocket) this.sslServerSocketFactory
+                        .createServerSocket(config.port);
+                if (this.config.mode.equals(DEFAULT_MODE)) {
+                    sslServerSocket.setNeedClientAuth(true);
+                } else {
+                    sslServerSocket.setWantClientAuth(true);
+                }
+
+                while (true) {
```
wborn, line 643: I think it is reusing a common openHAB thread pool? If so, it would be better to reschedule a new job when exception occur, so the thread is returned to the pool and can be used by other openHAB add-ons.

bundles/org.openhab.binding.androidtv/src/main/java/org/openhab/binding/androidtv/internal/protocol/googletv/GoogleTVConnectionManager.java
```diff
@@ -0,0 +1,1383 @@
...
+                    break; // exit loop and terminate thread
+                } catch (IOException e) {
+                    logger.warn("{} - Communication error, will try to reconnect GoogleTV. Error: {}",
+                            handler.getThingID(), e.getMessage());
+                    setStatus(false, "Communication error, will try to reconnect");
+                    sendQueue.add(command); // Requeue command
+                    this.isLoggedIn = false;
+                    reconnect();
+                    break; // reconnect() will start a new thread; terminate this one
+                }
```
wborn, line 865: As the loop is always exited in these catches, can the catches be moved to the try-catch outside the loop?

bundles/org.openhab.binding.androidtv/src/main/java/org/openhab/binding/androidtv/internal/protocol/googletv/GoogleTVConnectionManager.java
```diff
@@ -0,0 +1,1383 @@
...
+                        writer.write(command.toString());
+                        writer.flush();
+                    }
+                } catch (InterruptedIOException e) {
+                    logger.debug("Shim interrupted while sending.");
+                    break; // exit loop and terminate thread
+                } catch (IOException e) {
+                    logger.warn("Shim communication error. Error: {}", e.getMessage());
+                    break; // reconnect() will start a new thread; terminate this one
+                }
```
wborn, line 897: Can these also be moved to the try-catch outside the loop?

bundles/org.openhab.binding.androidtv/src/main/java/org/openhab/binding/androidtv/internal/protocol/googletv/GoogleTVConnectionManager.java
```diff
@@ -0,0 +1,1383 @@
...
+    }
+
+    public void sendShim(GoogleTVCommand command) {
+        if (!command.toString().equals("")) {
+            shimQueue.add(command);
+        }
+    }
+
+    public void handleCommand(ChannelUID channelUID, Command command) {
+        logger.debug("{} - Command received: {}", handler.getThingID(), channelUID.getId().toString());
```
wborn, line 1130:
```suggestion
        logger.debug("{} - Command received: {}", handler.getThingID(), channelUID.getId());
```

bundles/org.openhab.binding.androidtv/src/main/java/org/openhab/binding/androidtv/internal/protocol/googletv/GoogleTVConnectionManager.java
```diff
@@ -0,0 +1,1383 @@
...
+                        message = "52040818" + suffix;
+                        break;
+                    case "KEY_VOLDOWN":
+                        message = "52040819" + suffix;
+                        break;
+                    case "KEY_MUTE":
+                        message = "5204085b" + suffix;
+                        break;
+                    default:
+                        logger.debug("Unknown Key {}", command.toString());
```
wborn, line 1217:
```suggestion
                        logger.debug("Unknown Key {}", command.toString());
```
```suggestion
                        logger.debug("Unknown Key {}", command);
```

bundles/org.openhab.binding.androidtv/src/main/java/org/openhab/binding/androidtv/internal/protocol/googletv/GoogleTVConnectionManager.java
```diff
@@ -0,0 +1,1383 @@
...
+                    Certificate[] shimClientChain = this.shimClientChain;
+                    Certificate[] shimServerChain = this.shimServerChain;
+                    Certificate[] shimClientLocalChain = this.shimClientLocalChain;
+                    if (config.mode.equals(DEFAULT_MODE)) {
+                        if ((!isLoggedIn) && (command.toString().equals("REQUEST"))
+                                && (childConnectionManager == null)) {
+                            setStatus(false, "User Forced PIN Process");
+                            logger.debug("{} - User Forced PIN Process", handler.getThingID());
+                            disconnect(true);
+                            startChildConnectionManager(this.config.port + 1, PIN_MODE);
```
wborn, line 1269:
```suggestion
                            startChildConnectionManager(config.port + 1, PIN_MODE);
```

bundles/org.openhab.binding.androidtv/src/main/java/org/openhab/binding/androidtv/internal/protocol/googletv/GoogleTVConnectionManager.java
```diff
@@ -0,0 +1,1383 @@
...
+                if ((volMute && command.equals(OnOffType.OFF)) || (!volMute && command.equals(OnOffType.ON))) {
+                    sendCommand(new GoogleTVCommand(GoogleTVRequest.encodeMessage("5204085b1003")));
+                }
+            }
+        } else if (CHANNEL_DEBUG.equals(channelUID.getId())) {
+            if (command instanceof StringType) {
+                if (command.toString().startsWith("RAW", 9)) {
+                    String newCommand = command.toString().substring(13);
+                    String message = GoogleTVRequest.encodeMessage(newCommand);
+                    logger.trace("Raw Message Decodes as: {}", GoogleTVRequest.decodeMessage(message));
```
wborn, line 1327: Decoding is probably non-trivial so it would be better to make this conditional.

bundles/org.openhab.binding.androidtv/src/main/java/org/openhab/binding/androidtv/internal/protocol/googletv/GoogleTVConnectionManager.java
```diff
@@ -0,0 +1,1383 @@
...
+                    String newCommand = command.toString().substring(13);
+                    messageParser.handleMessage(newCommand);
+                }
+            }
+        } else if (CHANNEL_KEYBOARD.equals(channelUID.getId())) {
+            if (command instanceof StringType) {
+                String keyPress = "";
+                for (int i = 0; i < command.toString().length(); i++) {
+                    keyPress = "aa01071a0512031a01"
+                            + GoogleTVRequest.decodeMessage(new String("" + command.toString().charAt(i)));
```
wborn, line 1339:
```suggestion
                            + GoogleTVRequest.decodeMessage(String.valueOf(command.toString().charAt(i)));
```

bundles/org.openhab.binding.androidtv/src/main/java/org/openhab/binding/androidtv/internal/protocol/googletv/GoogleTVMessageParser.java
```diff
@@ -0,0 +1,334 @@
...
+                String audioMode = "";
+
+                int i = 0;
+                int current = 0;
+
+                for (; i < 12; i++) {
+                    preambleSb.append(charArray[i]);
+                }
+
+                st = "" + charArray[i] + charArray[i + 1];
```
wborn, line 219: Are these `"" + ` in all this code really necessary?

```suggestion
                st = charArray[i] + charArray[i + 1];
```

bundles/org.openhab.binding.androidtv/src/main/java/org/openhab/binding/androidtv/internal/protocol/googletv/GoogleTVRequest.java
```diff
@@ -0,0 +1,151 @@
...
+/**
+ * Contains static methods for constructing LEAP messages
+ *
+ * @author Ben Rosenblum - Initial contribution
+ */
+@NonNullByDefault
+public class GoogleTVRequest {
+
+    public static String encodeMessage(String message) {
+        String reply = new String();
```
wborn, line 29: `StringBuilder`s perform better when doing concatentation in for loops.

bundles/org.openhab.binding.androidtv/src/main/java/org/openhab/binding/androidtv/internal/protocol/googletv/GoogleTVRequest.java
```diff
@@ -0,0 +1,151 @@
...
+        for (int i = 0; i < charArray.length; i = i + 2) {
+            String st = "" + charArray[i] + "" + charArray[i + 1];
+            char ch = (char) Integer.parseInt(st, 16);
+            reply = reply + ch;
+        }
+        return reply;
+    }
+
+    public static String decodeMessage(String message) {
+        StringBuffer sb = new StringBuffer();
```
wborn, line 40: There's no need for the thread safety overhead of a `StringBuffer` here:

```suggestion
        StringBuilder sb = new StringBuilder();
```

bundles/org.openhab.binding.androidtv/src/main/java/org/openhab/binding/androidtv/internal/protocol/googletv/GoogleTVRequest.java
```diff
@@ -0,0 +1,151 @@
...
+        // 4a 02 08 01
+
+        // 42 08 08 7f 10 b4 908a a819
+        // 4a 02 08 7f
+
+        // 42 09 08 8001 10 ed b78a a819
+        // 4a 03 08 8001
+
+        char[] charArray = request.toCharArray();
+        StringBuffer sb = new StringBuffer();
```
wborn, line 124:
```suggestion
        StringBuilder sb = new StringBuilder();
```

bundles/org.openhab.binding.androidtv/src/main/java/org/openhab/binding/androidtv/internal/protocol/googletv/GoogleTVRequest.java
```diff
@@ -0,0 +1,151 @@
...
+        do {
+            int sbLen = sb.toString().length();
+            st = "" + charArray[sbLen - 2] + charArray[sbLen - 1];
+            if (!DELIMITER_10.equals(st)) {
+                sb.setLength(sbLen - 2);
+            }
+        } while (!DELIMITER_10.equals(st));
+        sb.setLength(sb.toString().length() - 2);
+
+        StringBuffer sbReply = new StringBuffer();
```
wborn, line 137: There's no need of the thread-safety overehead of a StringBuffer here:

```suggestion
        StringBuilder sbReply = new StringBuilder();
```

bundles/org.openhab.binding.androidtv/src/main/java/org/openhab/binding/androidtv/internal/protocol/googletv/GoogleTVUtils.java
```diff
@@ -0,0 +1,129 @@
...
+                processMag(digest);
+            } catch (NoSuchAlgorithmException e) {
+                LOGGER.debug("NoSuchAlgorithmException Exception", e);
+            }
+        }
+        return digest;
+    }
+
+    public static String byteArrayToString(byte[] array) {
+        StringBuffer sb = new StringBuffer();
```
wborn, line 101: There's no need of the thread-safety overehead of a `StringBuffer` here:

```suggestion
        StringBuilder sb = new StringBuilder();
```

bundles/org.openhab.binding.androidtv/src/main/java/org/openhab/binding/androidtv/internal/protocol/shieldtv/ShieldTVRequest.java
```diff
@@ -0,0 +1,104 @@
...
+        for (int i = 0; i < charArray.length; i = i + 2) {
+            String st = "" + charArray[i] + "" + charArray[i + 1];
+            char ch = (char) Integer.parseInt(st, 16);
+            reply = reply + ch;
+        }
+        return reply;
+    }
+
+    public static String decodeMessage(String message) {
+        StringBuffer sb = new StringBuffer();
```
wborn, line 39: There's no need of the thread-safety overehead of a `StringBuffer` here:

```suggestion
        StringBuilder sb = new StringBuilder();
```

bundles/org.openhab.binding.androidtv/src/main/java/org/openhab/binding/androidtv/internal/protocol/googletv/GoogleTVMessageParser.java
```diff
@@ -0,0 +1,334 @@
...
+                }
+
+                String st = "";
+                int length = 0;
+                StringBuffer preambleSb = new StringBuffer();
+                StringBuffer manufacturerSb = new StringBuffer();
+                StringBuffer modelSb = new StringBuffer();
+                StringBuffer androidVersionSb = new StringBuffer();
+                StringBuffer remoteServerSb = new StringBuffer();
+                StringBuffer remoteServerVersionSb = new StringBuffer();
```
wborn, line 97: I think there's no need of the thread-safety overehead of a `StringBuffer` here and `StringBuilder`s can be used instead?

bundles/org.openhab.binding.androidtv/src/main/java/org/openhab/binding/androidtv/internal/protocol/googletv/GoogleTVMessageParser.java
```diff
@@ -0,0 +1,334 @@
...
+                // 00 --- 9203 1c 08 03 10 06 1a 0c 42524156494120344b204742 20 02 2800 30 64 38 00 40 00
+                // 01 --- 9203 1c 08 03 10 06 1a 0c 42524156494120344b204742 20 02 2800 30 64 38 01 40 00
+                // 100 -- 9203 1c 08 03 10 06 1a 0c 42524156494120344b204742 20 02 2800 30 64 38 64 40 00
+                // MUTE - 9203 1c 08 03 10 06 1a 0c 42524156494120344b204742 20 02 2800 30 64 38 00 40 01
+
+                String st = "";
+                int length = 0;
+
+                StringBuffer preambleSb = new StringBuffer();
+                StringBuffer modelSb = new StringBuffer();
```
wborn, line 206:
```suggestion
                StringBuilder preambleSb = new StringBuilder();
                StringBuilder modelSb = new StringBuilder();
```

bundles/org.openhab.binding.androidtv/src/main/java/org/openhab/binding/androidtv/internal/protocol/googletv/GoogleTVMessageParser.java
```diff
@@ -0,0 +1,334 @@
...
+                // -----------------LEN-com.google.android.youtube.tv
+                // a201 21 0a 1f 62 1d 636f6d2e676f6f676c652e616e64726f69642e74766c61756e63686572
+                // -----------------LEN-com.google.android.tvlauncher
+                // a201 14 0a 12 62 10 636f6d2e736f6e792e6474762e747678
+                // -----------------LEN-com.sony.dtv.tvx
+                // a201 15 0a 13 62 11 636f6d2e6e6574666c69782e6e696e6a61
+                // -----------------LEN-com.netflix.ninja
+
+                StringBuffer preambleSb = new StringBuffer();
+                StringBuffer appNameSb = new StringBuffer();
```
wborn, line 302:
```suggestion
                StringBuilder preambleSb = new StringBuilder();
                StringBuilder appNameSb = new StringBuilder();
```

bundles/org.openhab.binding.androidtv/src/main/java/org/openhab/binding/androidtv/internal/protocol/googletv/GoogleTVRequest.java
```diff
@@ -0,0 +1,151 @@
...
+        } else {
+            // 080210c801c202 22 0a 20 0e066c3d1c3a6686edb6b2648ff25fcb3f0bf9cc81deeee9fad1a26073645e17
+            // 080210c801c202 22 0a 20 530bb7c7ba06069997285aff6e0106adfb19ab23c18a7422f5f643b35a6467b3
+            // -------------------------SHA HASH OF PIN
+
+            int length = pin.length() / 2;
+            String len1 = GoogleTVRequest.fixMessage(Integer.toHexString(length + 2));
+            String len2 = GoogleTVRequest.fixMessage(Integer.toHexString(length));
+            String reply = "080210c801c202" + len1 + "0a" + len2 + pin;
+            return reply;
```
wborn, line 66:
```suggestion
            return "080210c801c202" + len1 + "0a" + len2 + pin;
```

bundles/org.openhab.binding.androidtv/src/main/java/org/openhab/binding/androidtv/internal/protocol/googletv/GoogleTVUtils.java
```diff
@@ -0,0 +1,129 @@
...
+                processMag(digest);
+                instance.update(r1);
+                instance.update(r2);
+                instance.update(r3);
+                instance.update(r4);
+                instance.update(digest);
+                digest = instance.digest();
+                processMag(digest);
+            } catch (NoSuchAlgorithmException e) {
+                LOGGER.debug("NoSuchAlgorithmException Exception", e);
```
wborn, line 94: If the binding doesn't work due to the missing algorithm it would be better to `LOGGER.warn(..)` users about this.

bundles/org.openhab.binding.androidtv/src/main/java/org/openhab/binding/androidtv/internal/protocol/shieldtv/ShieldTVConnectionManager.java
```diff
@@ -0,0 +1,1218 @@
...
+        initialize();
+    }
+
+    public void setHostName(String hostName) {
+        this.hostName = hostName;
+        handler.setThingProperty("deviceName", hostName);
+    }
+
+    public String getHostName() {
+        return this.hostName;
```
wborn, line 162:
```suggestion
        return hostName;
```

bundles/org.openhab.binding.androidtv/src/main/java/org/openhab/binding/androidtv/internal/protocol/shieldtv/ShieldTVConnectionManager.java
```diff
@@ -0,0 +1,1218 @@
...
+        return handler.getThingID();
+    }
+
+    public void setDeviceID(String deviceId) {
+        this.deviceId = deviceId;
+        handler.setThingProperty("deviceID", deviceId);
+    }
+
+    public String getDeviceID() {
+        return this.deviceId;
```
wborn, line 175:
```suggestion
        return deviceId;
```

bundles/org.openhab.binding.androidtv/src/main/java/org/openhab/binding/androidtv/internal/protocol/shieldtv/ShieldTVConnectionManager.java
```diff
@@ -0,0 +1,1218 @@
...
+    }
+
+    public void setArch(String arch) {
+        this.arch = arch;
+        handler.setThingProperty("architectures", arch);
+    }
+
+    public String getArch() {
+        return this.arch;
```
wborn, line 184:
```suggestion
        return arch;
```

bundles/org.openhab.binding.androidtv/src/main/java/org/openhab/binding/androidtv/internal/protocol/shieldtv/ShieldTVConnectionManager.java
```diff
@@ -0,0 +1,1218 @@
...
+                appURL = appURLDB.get(currentApp);
+                handler.updateChannelState(CHANNEL_APPURL, new StringType(appURL));
+            } else {
+                handler.updateChannelState(CHANNEL_APPURL, new StringType(""));
+            }
+        }
+    }
+
+    public String getStatusMessage() {
+        return this.statusMessage;
```
wborn, line 213:
```suggestion
        return statusMessage;
```

bundles/org.openhab.binding.androidtv/src/main/java/org/openhab/binding/androidtv/internal/protocol/shieldtv/ShieldTVConnectionManager.java
```diff
@@ -0,0 +1,1218 @@
...
+    private void setStatus(boolean isLoggedIn, String statusMessage) {
+        if ((this.isLoggedIn != isLoggedIn) || (!this.statusMessage.equals(statusMessage))) {
+            this.isLoggedIn = isLoggedIn;
+            this.statusMessage = statusMessage;
+            handler.checkThingStatus();
+        }
+    }
+
+    public String getCurrentApp() {
+        return this.currentApp;
```
wborn, line 233:
```suggestion
        return currentApp;
```

bundles/org.openhab.binding.androidtv/src/main/java/org/openhab/binding/androidtv/internal/protocol/shieldtv/ShieldTVConnectionManager.java
```diff
@@ -0,0 +1,1218 @@
...
+    private TrustManager[] defineNoOpTrustManager() {
+        return new TrustManager[] { new X509TrustManager() {
+            @Override
+            public void checkClientTrusted(final X509Certificate @Nullable [] chain, final @Nullable String authType) {
+                logger.debug("Assuming client certificate is valid");
+                if (chain != null && logger.isTraceEnabled()) {
+                    for (int cert = 0; cert < chain.length; cert++) {
+                        logger.trace("Subject DN: {}", chain[cert].getSubjectX500Principal());
+                        logger.trace("Issuer DN: {}", chain[cert].getIssuerX500Principal());
+                        logger.trace("Serial number {}:", chain[cert].getSerialNumber());
```
wborn, line 322:
```suggestion
                        logger.trace("Serial number: {}", chain[cert].getSerialNumber());
```

bundles/org.openhab.binding.androidtv/src/main/java/org/openhab/binding/androidtv/internal/protocol/shieldtv/ShieldTVConnectionManager.java
```diff
@@ -0,0 +1,1218 @@
...
+
+            @Override
+            public void checkServerTrusted(final X509Certificate @Nullable [] chain, final @Nullable String authType) {
+                logger.debug("Assuming server certificate is valid");
+                if (chain != null && logger.isTraceEnabled()) {
+                    for (int cert = 0; cert < chain.length; cert++) {
+                        logger.trace("Subject DN: {}", chain[cert].getSubjectX500Principal());
+                        logger.trace("Issuer DN: {}", chain[cert].getIssuerX500Principal());
+                        logger.trace("Serial number: {}", chain[cert].getSerialNumber());
+                    }
```
wborn, line 335: Can it be extracted to a method and reused instead of being duplicated?

bundles/org.openhab.binding.androidtv/src/main/java/org/openhab/binding/androidtv/internal/protocol/shieldtv/ShieldTVConnectionManager.java
```diff
@@ -0,0 +1,1218 @@
...
+
+/**
+ * The {@link ShieldTVConnectionManager} is responsible for handling connections via the shieldtv protocol
+ *
+ * Significant portions reused from Lutron binding with permission from Bob A.
+ *
+ * @author Ben Rosenblum - Initial contribution
+ */
+@NonNullByDefault
+public class ShieldTVConnectionManager {
```
wborn, line 77: This class seems to duplicate logic from the `GoogleTVConnectionManager`. Can't you create an `AbstractConnectionManager` and then inherit from it in both and use that class for sharing the logic? Many comments I made to the `GoogleTVConnectionManager` also apply to this `ShieldTVConnectionManager`.

wborn (changes requested):
Thanks for addressing the comments. :+1: 
Sorry, I was a bit busy but I've added a few more comments below:

bundles/org.openhab.binding.androidtv/src/main/java/org/openhab/binding/androidtv/internal/protocol/shieldtv/ShieldTVConnectionManager.java
morph166955: I did attempt this early on. Unfortunately while the functions are all similar, and the structure is similar, many have ever so slight but critically different bits for each protocol which made abstracting very hard. For example, on the GTV protocol we have to deal with the message length being read or appended for read/write whereas on STV we have to look for specific delimiters. I can try to see what can be abstracted in the future, but this will take significant effort and I'm not sure we will be successful.

bundles/org.openhab.binding.androidtv/src/main/java/org/openhab/binding/androidtv/internal/protocol/googletv/GoogleTVMessageParser.java
```diff
@@ -0,0 +1,334 @@
...
+
+                    callback.setVolMax(volMax);
+                    callback.setVolCurr(volCurr);
+                    callback.setVolMute(volMute);
+                }
+
+                String preamble = preambleSb.toString();
+                String model = GoogleTVRequest.encodeMessage(modelSb.toString());
+                logger.debug("{} - Device Update: {} \"{}\" {} {} {} {}", callback.getThingID(), preamble, model,
+                        audioMode, volMax, volCurr, volMute);
```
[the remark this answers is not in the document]
morph166955: You're referring to getThingID() correct?

bundles/org.openhab.binding.androidtv/src/main/java/org/openhab/binding/androidtv/internal/protocol/googletv/GoogleTVMessageParser.java
morph166955: Unfortunately yes, I had issues without them. Adding "" + forces the String to build properly when the first variable is by default not a String.

bundles/org.openhab.binding.androidtv/src/main/java/org/openhab/binding/androidtv/internal/protocol/googletv/GoogleTVMessageParser.java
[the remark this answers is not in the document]
morph166955: Would it just be better to call it once at the top and store it for use?  That seems the most efficient since we don't ever change the thingId.

bundles/org.openhab.binding.androidtv/src/main/java/org/openhab/binding/androidtv/internal/protocol/googletv/GoogleTVUtils.java
morph166955: This should NEVER throw.  This was added to java maybe back in the java 7 days.  I'll set this to warn just in case, but more or less we're just catching because it's being thrown by instance.digest();

bundles/org.openhab.binding.androidtv/src/main/java/org/openhab/binding/androidtv/internal/protocol/googletv/GoogleTVMessageParser.java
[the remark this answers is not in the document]
morph166955: I pulled this up to the top since we had a few at warn and info that used getThingID() so we only ever call it once per message.

bundles/org.openhab.binding.androidtv/src/main/java/org/openhab/binding/androidtv/internal/protocol/googletv/GoogleTVConnectionManager.java
morph166955: This is looping not due to an exception but to wait for the next connection.  In theory when we hit line 708 cleanly (meaning the connection was good and nothing failed, we go back to line 644 to wait for the next connection.  Also, shim is a developer item for me to do break and inspect on the app, it is NOT enabled by the end users so this should never impact actual user operations.

bundles/org.openhab.binding.androidtv/src/main/java/org/openhab/binding/androidtv/internal/protocol/googletv/GoogleTVConnectionManager.java
morph166955: We catch inside the loop because we want to reinsert command back into the sendQueue on line 861.  If we pull it outside the loop then command doesn't exist anymore.  I suppose we could declare command above but this was cleaner.

bundles/org.openhab.binding.androidtv/src/main/java/org/openhab/binding/androidtv/internal/protocol/googletv/GoogleTVConnectionManager.java
morph166955: I don't see why we can't, but as this is part of shim it's almost never going to actually.  Also keeping it inside the loop keeps it the same as above which makes it easier for me to debug.

bundles/org.openhab.binding.androidtv/src/main/java/org/openhab/binding/androidtv/internal/protocol/shieldtv/ShieldTVConnectionManager.java
morph166955: To be honest, I've had a ton of issues with the TrustManagers across this whole binding.  They are very finicky and the only thing that actually matters is the getAcceptedIssuers() function.  The two check functions are just for debugging.  Can I add this as a to-do for later as I go through and make the code more efficient?

bundles/org.openhab.binding.androidtv/src/main/java/org/openhab/binding/androidtv/internal/protocol/googletv/GoogleTVConnectionManager.java
```diff
@@ -0,0 +1,1383 @@
...
+            @Override
+            public X509Certificate @Nullable [] getAcceptedIssuers() {
+                X509Certificate[] x509ClientChain = shimX509ClientChain;
+                if (x509ClientChain != null && logger.isTraceEnabled()) {
+                    logger.debug("Returning shimX509ClientChain for getAcceptedIssuers");
+                    for (int cert = 0; cert < x509ClientChain.length; cert++) {
+                        logger.trace("Subject DN: {}", x509ClientChain[cert].getSubjectX500Principal());
+                        logger.trace("Issuer DN: {}", x509ClientChain[cert].getIssuerX500Principal());
+                        logger.trace("Serial number: {}", x509ClientChain[cert].getSerialNumber());
+                    }
```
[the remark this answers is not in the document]
morph166955: See note below.

morph166955:
@wborn I've pushed several commits to resolve the bulk of the comments.  I've left comments on the rest for your consideration.

morph166955:
@wborn Checking in to see if there's anything else on this.

wborn (changes requested):
Thanks for addressing the comments! I've added a few more below. After that I think it's nice and clean. :slightly_smiling_face:

bundles/org.openhab.binding.androidtv/src/main/java/org/openhab/binding/androidtv/internal/protocol/shieldtv/ShieldTVMessageParser.java
```diff
@@ -0,0 +1,448 @@
...
+                    LinkedHashMap<String, String> sortedAppNameDB = new LinkedHashMap<>();
+                    ArrayList<String> valueList = new ArrayList<>();
+                    for (Map.Entry<String, String> entry : appNameDB.entrySet()) {
+                        valueList.add(entry.getValue());
+                    }
+                    Collections.sort(valueList, new Comparator<String>() {
+                        public int compare(String str, String str1) {
+                            return (str).compareTo(str1);
+                        }
+                    });
```
wborn, line 328: It's the default implementation: 

```suggestion
                    Collections.sort(valueList);
```

morph166955:
@wborn I've pushed the fixes as requested.  I'm going to push the current code to the addon marketplace as well.

wborn (approved):
Many thanks! :+1: Let's merge it! :rocket:

morph166955:
Thank you @wborn!  

@jlaur I haven't forgotten you, I'll be working on those follow-on PRs in the next few days in hopes to get them pulled in prior to the final 4.0.0 release.

PR #14283, opened 2023-01-27
[unifi] Added channels name and hostname to client things.

Hilbrand:


bundles/org.openhab.binding.unifi/src/main/java/org/openhab/binding/unifi/internal/api/cache/UniFiClientCache.java
```diff
@@ -48,11 +48,15 @@ public UniFiClientCache() {
             case IP:
                 return client.getIp();
             case HOSTNAME:
-                return client.getHostname();
+                return safeTidy(client.getHostname());
             case ALIAS:
```
jlaur, line 52: Consider if `ALIAS` can/should be renamed to `NAME` also for consistency.

jlaur (changes requested):
Thanks for the channels! I have only one question about the remaining mapping/inconsistency between alias and name. Maybe there is a good reason for keeping it, I'm just a little confused by having both.

jlaur (approved):
LGTM, thanks.

PR #14289, opened 2023-01-29
[hue] Get scenes for groups without 10 minutes delay

### LINKED ISSUE #14137: [hue] "scene" channel on a hue group populated with several minutes delay
Happy new year everyone.

The hue channel type ID "scene" can be used on the bridge or on a hue group. On the bridge it lists all scenes on the bridge with room names in brackets and when used on a hue group it filters the scenes to the scenes in this group. This worked as expected until Milestone M3.

Now it only works at the bridge level. The scene list for a hue group remains empty. Unfortunately this is the most useful feature as you want to select a scene for a room from a filtered list for this room (hue group).

I now downgraded to M3 as this is an essential feature more me. Hopefully there is someone with code knowledge of the hue binding who can support here.

Thanks in advance.

lolodomo:
Fix #14137


jlaur (approved):
Thanks!

Issue #13798, opened 2022-11-28
[tr064] Binding OH3 MAC adresses for LANDevice


FrankH1:
Did install the TR-064 Binding - just with the purpose of detecting the "onlinestatus" of mobiles in the household. Went up to the subdevice (LANDevice). Device and Subdevice (FritzBox) both are online. Need to add I'm doing this via the GUI of OH3. If you start to add several  MAC adresses, separated by "," as requested by description then only for the last adress the system offers channels like MAConline etc. Did discuss this in the german OPENHAB-Forum. The souliution was to skip the commas - so the adresses are filled in the the GUI in separate lines - but without "," or other separators. This works and I'm able to get the status of connected devices.

So either the Addin needs to be changed to work with "," as separator - or the documentation needs update/change.

## Your Environment

* openHAB 3.3.0 Release Build, TR-064 3.3.0
* Raspbian, Raspi 4B

lsiepel:
Can you point me to that discussion ? It looks like it is a documenation problem. I also see a difference in the english and german translation.

I assume you tried this and it failed:
```
Thing subdeviceLan LAN "label LAN"   [ uuid="uuid:xxxxxxxx-xxxx-xxxx-yyyy-xxxxxxxxxxxx",  macOnline="XX:XX:XX:XX:XX:XX", "YY:YY:YY:YY:YY:YY"]
```
                                                          
You then tried this and that succeedded:
```
Thing subdeviceLan LAN "label LAN"   [ uuid="uuid:xxxxxxxx-xxxx-xxxx-yyyy-xxxxxxxxxxxx", macOnline="XX:XX:XX:XX:XX:XX" "YY:YY:YY:YY:YY:YY"]  
```

FrankH1:
Hi, 
I did everything on the gui. In the gui there is a box, where you can put in MAC addresses, and there is an explaining text below, saying adresses need to be separated by commas. ( Can't check as I'm now on holidays far away from cold Germany...)
The original topic where I raised that is in the German forum: https://openhabforum.de/viewtopic.php?t=7578
In the post from sihui is a screenshot, where the corresponding part of the gui is shown. Hope that helps. Cheers

jlaur:
Related PR merged: #13990

PR #13988, opened 2022-12-17
[tr064] Description in LanDevice MAC "comma separated list" wrong

Spiev:
In the opehab binding configuration for the tr064 LANDevice configuration is wrong. It says, that (translate from german in english): 
_comma separated list of MAC-Addresses._ 

But this dosen't work :-D

The setting requires "_one MAC-Address per row_".

I did the changes already to the files where I found the corresponding wording.

Screenshot 2022-12-17 150628
Screenshot 2022-12-17 150535

PR #14291, opened 2023-01-29
[tr064] Update configuration documentation

lsiepel:

- Improves documentation for comma separated list of MAC-Addresses 

Replaces: https://github.com/openhab/openhab-addons/pull/13988/
Fixes: https://github.com/openhab/openhab-addons/issues/13798


bundles/org.openhab.binding.tr064/src/main/resources/OH-INF/i18n/tr064.properties
```diff
@@ -35,29 +35,32 @@ thing-type.config.tr064.fritzbox.rejectedCallDays.label = Rejected Call Days
...
 thing-type.config.tr064.subdeviceLan.refresh.label = Refresh Interval
 thing-type.config.tr064.subdeviceLan.uuid.label = UUID
 thing-type.config.tr064.subdeviceLan.uuid.description = UUID of the sub-device
 
 # thing types config
 
+thing-type.config.tr064.fritzbox.timeout.label = Timeout
+thing-type.config.tr064.fritzbox.timeout.description = Timeout for all requests (SOAP requests, phone book retrieval, call lists, ...).
+thing-type.config.tr064.generic.timeout.label = Timeout
+thing-type.config.tr064.generic.timeout.description = Timeout for all requests (SOAP requests, phone book retrieval, call lists, ...).
```
jlaur, line 60: How did they end here? Perhaps no longer referenced?

jlaur (changes requested):
Thanks! Seems like some I18N cleanup is needed.

bundles/org.openhab.binding.tr064/src/main/resources/OH-INF/i18n/tr064.properties
lsiepel: If they are no longer referenced, i would suspect that the i18n tool would remove them.
I only ran the i18n mvn command and this is the result. As far as i can see there is no real change, the lines are only moved to a different place in the file. So i would suggest to merge it as is. Please advice if you have other suggestions.

jlaur: You can try to delete this section and re-run the tool. If they don't reappear, they are no longer referenced.

lsiepel: Done. Seems there where more issues, i completely remove all contents and re-run the tool. It cleaned more entries.

bundles/org.openhab.binding.tr064/src/main/resources/OH-INF/i18n/tr064.properties
```diff
@@ -35,46 +35,19 @@ thing-type.config.tr064.fritzbox.rejectedCallDays.label = Rejected Call Days
...
-phonebookLookupActionLabel = lookup a phone number
-phonebookLookupActionDescription = Lookup a phone number.
-phonebookLookupActionOutputLabel = Name
-phonebookLookupActionOutputDescription = The name associated with the phone number (or the phone number if no match can be found).
-phonebookLookupActionInputPhoneNumberLabel = Phone Number
-phonebookLookupActionInputPhoneNumberDescription = The phone number to be resolved.
-phonebookLookupActionInputPhoneBookLabel = Phone Book
-phonebookLookupActionInputPhoneBookDescription = The name of the the phone book.
-phonebookLookupActionInputMatchesLabel = Match Count
-phonebookLookupActionInputMatchesDescription = The number of digits matching the incoming value, counted from far right (default is 0 = all matching). Negative numbers skip digits from the left.
```
jlaur, line 80: These are still referenced by `@text/` references here:
https://github.com/openhab/openhab-addons/blob/be6b799ca447b55ce69a068e65af57d3d843d11b/bundles/org.openhab.binding.tr064/src/main/java/org/openhab/binding/tr064/internal/phonebook/PhonebookActions.java#L42-L66

lsiepel: I’ll have a look, probably revert everything as it was. 
This is something the i18n maven command does not pickup?

jlaur: The tool only picks strings from the XML sources. For references like above, the I18N properties file **is** the source.

lsiepel: Checked and fixed the actions. Thanks for the details. Still learning alot here ;-)

lsiepel: I could not find any reference for the `profile-type.transform.xxx` but don't know where to expect those so i might be looking at the wrond place.

jlaur: They have a suspicious duplicated header "# thing types config", so probably they are not relevant anymore. If you want to be 100% sure, you could check git history and find the commit where they moved, and check other changes in that commit.

lsiepel: Seems like they are merged by https://github.com/openhab/openhab-addons/commit/f9695f2d8dd234bdba9b96a5c902b37798b9fded#diff-6114fe980cebbb1ca878651701ea434d443f2ed2071402e103f41683d5679349
and somehow referenced from [bundles/org.openhab.binding.tr064/src/main/resources/OH-INF/config/phonebookProfile.xml](https://github.com/openhab/openhab-addons/commit/f9695f2d8dd234bdba9b96a5c902b37798b9fded#diff-6114fe980cebbb1ca878651701ea434d443f2ed2071402e103f41683d5679349)

The uri in the xml is profile:transform:PHONEBOOK and in the translation is profile.config.transform.PHONEBOOK 

i'm to unfamiliair with this piece of translation / binding specifics, it also goes beyond the goal of this PR. So do you agree that i revert the removal of these lines and you merge it as is?

jlaur (approved):
LGTM

PR #14292, opened 2023-01-29
[rrd4j] fixed oom when requesting data from rrd4j and boundary=true

crnjan:
Using rrd4j (as default provider) and requesting i.e.
```
http://<ip>/rest/persistence/items/gBasement_BoilerRoom_HP_RAD_InOperation_HP?starttime=2022-12-31T23%3A00%3A00.000Z&endtime=2023-01-31T22%3A59%3A59.999Z&boundary=true 
```
will return
```
{
"cause1":"java.lang.OutOfMemoryError: Java heap space",
"servlet":"cxf-servlet",
"cause0":"org.apache.cxf.interceptor.Fault: Java heap space",
"message":"org.apache.cxf.interceptor.Fault: Java heap space",
"url":"/rest/persistence/items/gBasement_BoilerRoom_HP_RAD_InOperation_HP",
"status":"500"
}
```
and requesting the same without the boundary parameter will return datapoints ...

Also, requesting the same range with boundary=true using influxdb also works fine (I usually use influxdb for charts therefore did not see this before):
```
http://<ip>/rest/persistence/items/gBasement_BoilerRoom_HP_RAD_InOperation_HP?starttime=2022-12-31T23%3A00%3A00.000Z&endtime=2023-01-31T22%3A59%3A59.999Z&boundary=true&serviceId=influxdb
```
also returns datapoints for given interval.

Seems to be regression from https://github.com/openhab/openhab-addons/pull/14238, related to https://github.com/openhab/openhab-addons/pull/14238#discussion_r1089949232.

Reverted code change back to current timestamp. Not 100% sure if this is the best way to solve this, but at least with this it's not giving oom anymore.


bundles/org.openhab.persistence.rrd4j/src/main/java/org/openhab/persistence/rrd4j/internal/RRD4jPersistenceService.java
```diff
@@ -283,7 +283,7 @@ public Iterable<HistoricItem> query(FilterCriteria filter) {
         // set end to {@link Instant#MAX} instead of current timestamp to enable requesting future time ranges including
         // boundary values via REST API
         // see discussion in https://github.com/openhab/openhab-addons/pull/14238
-        long end = filter.getEndDate() == null ? Instant.MAX.getEpochSecond()
+        long end = filter.getEndDate() == null ? System.currentTimeMillis() / 1000
```
jlaur, line 286: This reintroduces the faulty logic fixed by #14238 as `end` can now end up being less than `start`, if `start` is in the future. An exception will no longer be thrown, only a warning logged. However, the logic is still not correct.

In the scenario where you are experiencing this, what is the concrete values for filter startDate and the current time? I'm wondering if the delta is only a few milliseconds, of is it's a actual future date.

crnjan: I'm aware, but until a better solution if found, with this change in place at least it works (compared to oom). I'm not at home so cannot dump concrete values, but assuming, based on [code](https://github.com/openhab/openhab-core/blob/8f9dafd1d68a92248cd39414145175952d61cd46/bundles/org.openhab.core.io.rest.core/src/main/java/org/openhab/core/io/rest/core/internal/persistence/PersistenceResource.java#L346-L351) delta might be different, depending on interval, i.e. imagine having per-month chart - based on selected month, `dateTimeEnd` might be (as set in api call):
```
endtime=2022-12-31T22:59:59.999Z
endtime=2022-11-30T22:59:59.999Z
...
```
depending on which month is being displayed/selected in chart view. And end time is used for start time
```
filterAfterEnd.setBeginDate(dateTimeEnd);
```
and no end time is set - resulting in different deltas, right?
```
delta = current timestamp - <2022-12-31T22:59:59.999Z>
```
or
```
delta = current timestamp - <2022-11-30T22:59:59.999Z>
```
...
As said, until a better solution is found, at least it works (renders chart) and probably logs a warning in some cases, but (IMHO) still better compared to not render charts at all due oom ...

bundles/org.openhab.persistence.rrd4j/src/main/java/org/openhab/persistence/rrd4j/internal/RRD4jPersistenceService.java
```diff
@@ -320,8 +317,8 @@ public Iterable<HistoricItem> query(FilterCriteria filter) {
 
             // do not call method {@link RrdDb#createFetchRequest(ConsolFun, long, long, long)} if start > end to avoid
             // an IAE to be thrown
-            if (start > end) {
-                logger.warn("Could not query rrd4j database for item '{}': start ({}) > end ({})", itemName, start,
+            if (start > end && filter.getEndDate() == null) {
```
jlaur, line 320: Logic is now wrong: If `start > end && filter.getEndDate() != null`, we will now call the method and fully reintroduce the bug fixed by #14238. You probably meant something like this:
```java
if (start > end) {
    if (filter.getEndDate() != null)
        logger.debug("Could not query rrd4j database for item '{}': start ({}) > end ({})", itemName, start, end);
    }
    return List.of();
}
```

jlaur (changes requested):
Please have a look at my latest comment. With this fixed, I think it can be merged.

bundles/org.openhab.persistence.rrd4j/src/main/java/org/openhab/persistence/rrd4j/internal/RRD4jPersistenceService.java
crnjan: hm, are you sure? - if end date is not set (which is the case when invoking API with `boundary=true` and we "calculate" end timestamp - only in this case we simply return empty response and don't trigger an exception. Tested locally and it works. If user would invoke a query with explicitly setting end time and `end > start`, "old" logic would kick in. To be honest I'm not sure if we should add (at all) `&& filter.getEndDate() == null` part,  ... and maybe only check `start > end` ... but then explicit queries would also only log `debug` and return empty response. Open for suggestions ...

Also checked a bit the code and it seems datapoints are stored using `long now = System.currentTimeMillis() / 1000;` - https://github.com/openhab/openhab-addons/blob/main/bundles/org.openhab.persistence.rrd4j/src/main/java/org/openhab/persistence/rrd4j/internal/RRD4jPersistenceService.java#L189 so there should be no datapoint "in the future" so having 
```
long end = filter.getEndDate() == null ? System.currentTimeMillis() / 1000
```
should probably be good enough. We might also use `getLastUpdateTime()` but not sure if that would bring much benefit ...

crnjan: Please remember - when rendering chart with i.e.
```
http://<ip>/rest/persistence/items/<item>?starttime=2022-12-31T23:00:00.000Z&endtime=2023-01-31T22:59:59.999Z&boundary=true
```
will query database three times (due `boundary=true`):
1 . for a point before `starttime`, 
2. with explicit start/end- `starttime=2022-12-31T23:00:00.000Z` and `endtime=2023-01-31T22:59:59.999Z`,
3. start time of query = end time (2023-01-31T22:59:59.999Z in above example) and no end time set - which is the offending query before the fix and raised an exception. Therefore `filter.getEndDate() == null` since no end is set in this case

jlaur: If `filter.getEndDate() == null`, `end` is set to `System.currentTimeMillis()` in line 283. This may be before `start`. This scenario is already covered by your code.

If `filter.getEndDate() != null` and `end` is before `start`, this early exit will not happen. Instead `createFetchRequest` will be called below and the exception is thrown.

crnjan: IMHO it should not be triggered (start > end) from UI unless there is a bug. The reason for adding this additional check was (a bit) based on your comment https://github.com/openhab/openhab-addons/pull/14238#discussion_r1082220561 so if user would invoke API and explicitly set `start > end` ... but I guess the exception that would be raised (and I guess it was in the past) is not "the best" ... maybe we should really just get rid of the `&& filter.getEndDate() == null` part?

crnjan: my rationale was - if user (or UI) did not specify end date via REST API call and internal logic results in `start > end` -> return empty list; if user did explicitly set `start > end` -> use existing logic (which triggers exception I guess, but did not try it out yet, will do so later today ...).

jlaur: Thanks for linking that comment of mine. I actually still agree with that (myself), I guess I'm just sleepy today. 🙂 In that case I would take the control then and explicitly throw that exception by ourselves instead of relying on external error handling. So we could validate parameters in the beginning of the method (pseudo code):
```java
ZonedDateTime filterBeginDate = filter.getBeginDate();
ZonedDateTime filterEndDate = filter.getEndDate()
if (filterBeginDate != null && filterEndDate != null && filterBeginDate.isAfter(filterEndDate)) {
    throw new IllegalArgumentException("begin (" + filterBeginDate + ") before filterEndDate (" + filterEndDate + ")");
}
```

Then we wouldn't have to deal with that later in the method.

crnjan: Great suggestion, updated accordingly, seems to work. Thx for your feedback, really appreciated.

crnjan:
just to add - there is a gate before code in this PR - https://github.com/openhab/openhab-core/blob/8f9dafd1d68a92248cd39414145175952d61cd46/bundles/org.openhab.core.io.rest.core/src/main/java/org/openhab/core/io/rest/core/internal/persistence/PersistenceResource.java#L280

So seems, even if one would explicitly set end date before start, it would not enter code in this PR (was not able to get an error when invoking API directly with start > end therefore checked why ...)

jlaur:
> just to add - there is a gate before code in this PR is reached [...]
> So seems, even if one would explicitly set end date before start, it would not enter code in this PR (was not able to get an error when invoking API directly with start > end therefore checked why ...) - but still like the change due consistency.

That's great! Throwing this exception is just a means to enforce the contract, i.e. to reveal bugs elsewhere. It seems we currently do not have such bug, but now we can ensure that we will fail fast in case it would be introduced in the future somewhere. And as a bonus, after parameter validation, usually the actual logic can be kept cleaner.

bundles/org.openhab.persistence.rrd4j/src/main/java/org/openhab/persistence/rrd4j/internal/RRD4jPersistenceService.java
```diff
@@ -280,10 +286,7 @@ public Iterable<HistoricItem> query(FilterCriteria filter) {
         }
 
         long start = 0L;
-        // set end to {@link Instant#MAX} instead of current timestamp to enable requesting future time ranges including
-        // boundary values via REST API
-        // see discussion in https://github.com/openhab/openhab-addons/pull/14238
-        long end = filter.getEndDate() == null ? Instant.MAX.getEpochSecond()
+        long end = filter.getEndDate() == null ? System.currentTimeMillis() / 1000
```
jlaur, line 289: Now that we have a local variable, let's use it:
```suggestion
        long end = filterEndDate == null ? System.currentTimeMillis() / 1000
```

also in next line (and possibly elsewhere) - and similarly for `filter.getBeginDate()`.

bundles/org.openhab.persistence.rrd4j/src/main/java/org/openhab/persistence/rrd4j/internal/RRD4jPersistenceService.java
```diff
@@ -321,8 +324,6 @@ public Iterable<HistoricItem> query(FilterCriteria filter) {
             // do not call method {@link RrdDb#createFetchRequest(ConsolFun, long, long, long)} if start > end to avoid
             // an IAE to be thrown
             if (start > end) {
-                logger.warn("Could not query rrd4j database for item '{}': start ({}) > end ({})", itemName, start,
-                        end);
```
jlaur, line 325: Why is this warning removed?

crnjan: Since this can only happen when "internal" calculations make `start > end` don't think it makes sense to log it ... at least not at warn level?

jlaur: I seem to remember some case where it was only milliseconds in difference, but I'm not sure. I'd suggest to keep it as **debug** to be able to backtrack such cases.

crnjan: done

jlaur (approved):
LGTM - thanks!

PR #14311, opened 2023-02-01
[anthem] Initial contribution of binding for Anthem AV preamp/processors

mhilbush:
This is a binding that can control Anthem AV preamp/processors using the Anthem IP control protocol. It has been tested on model AVM-60.


bundles/org.openhab.binding.anthem/README.md
```diff
@@ -0,0 +1,83 @@
...
+| 1#muteToggle            | Switch  | Toggle the mute on and off  |
+| 1#activeInput           | Number  | The currently active input source  |
+| 1#activeInputShortName  | String  | Short friendly name of the active input  |
+| 1#activeInputLongName   | String  | Long friendly name of the active input |
+| *Zone 2*                |         |   |
+| 2#power                 | Switch  | Power the zone on or off  |
+| 2#volume                | Dimmer  | Increase or decrease the volume level  |
+| 2#volumeDB              | Number  | The actual volume setting  |
+| 2#mute                  | Switch  | Mute the volume  |
+| 2#muteToggle            | Switch  | Toggle the mute on and off  |
```
jlaur, line 51: What is the difference between these two channels?

bundles/org.openhab.binding.anthem/src/main/java/org/openhab/binding/anthem/internal/handler/AnthemConnectionManager.java
```diff
@@ -0,0 +1,433 @@
...
+    private int reconnectIntervalMinutes;
+    private int commandDelayMsec;
+
+    private boolean zone1PreviousPowerState;
+    private boolean zone2PreviousPowerState;
+
+    public AnthemConnectionManager(AnthemHandler handler, AnthemConfiguration configuration) {
+        messageParser = new AnthemCommandParser(this);
+        this.configuration = configuration;
+        this.handler = handler;
```
jlaur, line 89: This class seems very tightly coupled with the handler. Actually it's almost like a partial handler class doing much of the same things you would normally expect a handler to do. It's documented as a class handling the connection to the device, so maybe consider if you could reduce it to do just that, and decouple it from the handler logic.

jlaur (changes requested):
Thanks for providing this new binding. It looks good in general. I have added a few comments and questions from first review iteration.

bom/openhab-addons/pom.xml
```diff
@@ -117,6 +117,11 @@
       <version>${project.version}</version>
     </dependency>
     <dependency>
+      <groupId>org.openhab.addons.bundles</groupId>
+      <artifactId>org.openhab.binding.anthem</artifactId>
+      <version>${project.version}</version>
+    <dependency>
+    </dependency>
```
jlaur, line 124: This will probably fix the build:
```suggestion
    </dependency>
    <dependency>
```

mhilbush: Oh my. Yes I'm sure it will.

jlaur: It failed again, but for another reason. You also need to add your binding to this POM:
https://github.com/openhab/openhab-addons/blob/9a05e9f3b5662ecb9701e95ecb02e07bae37d98c/bundles/pom.xml#L58-L59

mhilbush: Ugh. Sadly it's been so long since I've written a binding I've forgotten all the places that need to be touched. 😢

jlaur: But now you got them all - build succeeded. 🙂

bundles/org.openhab.binding.anthem/README.md
```diff
@@ -0,0 +1,83 @@
...
+
+The binding allows control of Anthem AV processors over an IP connection to the processor.
+
+## Supported Things
+
+The following thing type is supported:
+
+| Thing    | ID       | Discovery | Description |
+|----------|----------|-----------|-------------|
+| Anthem   | anthem   | Manual | Represents a Anthem AV processor |
```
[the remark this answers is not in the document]
mhilbush: Fixed

bundles/org.openhab.binding.anthem/README.md
```diff
@@ -0,0 +1,83 @@
...
+
+Tested models include the AVM-60 11.2-channel preamp/processor.
+
+## Discovery
+
+There is no automatic discovery.
+Things must be added manually using the user interface or a `.things` file.
+
```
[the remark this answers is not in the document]
mhilbush: Removed

bundles/org.openhab.binding.anthem/README.md
mhilbush: The Anthem processors support two styles of muting. There's an ON/OFF mute command, and there's a mute toggle command (sending an ON command will change to the opposite state of whatever it's currently set to). I debated whether to just add the mute ON/OFF channel or add two channels, one for ON/OFF and one for toggling the mute. Do you have an opinion?

bundles/org.openhab.binding.anthem/src/main/java/org/openhab/binding/anthem/internal/AnthemBindingConstants.java
```diff
@@ -0,0 +1,52 @@
...
+ * @author Mark Hilbush - Initial contribution
+ */
+@NonNullByDefault
+public class AnthemBindingConstants {
+    public static final String BINDING_ID = "anthem";
+
+    public static final ThingTypeUID THING_TYPE_ANTHEM = new ThingTypeUID(BINDING_ID, "anthem");
+
+    // List of all Thing Type UIDs
+    public static final Set<ThingTypeUID> SUPPORTED_THING_TYPES_UIDS = Collections.singleton(THING_TYPE_ANTHEM);
```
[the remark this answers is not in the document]
mhilbush: Changed

bundles/org.openhab.binding.anthem/src/main/java/org/openhab/binding/anthem/internal/handler/AnthemConnectionManager.java
mhilbush: Yes, it is. I went down a bit of a rabbit hole after I got started with it. I may just move all the functionality back into the handler class, assuming that's also acceptable to you.

bundles/org.openhab.binding.anthem/src/main/java/org/openhab/binding/anthem/internal/handler/AnthemConnectionManager.java
```diff
@@ -0,0 +1,433 @@
...
+    }
+
+    private void initalize() {
+        reconnectIntervalMinutes = configuration.reconnectIntervalMinutes;
+        commandDelayMsec = configuration.commandDelayMsec;
+        asyncInitializeTask = scheduler.submit(this::connect);
+    }
+
+    private synchronized void connect() {
+        handler.updateThingStatus(ThingStatus.OFFLINE, ThingStatusDetail.NONE, "Initializing");
```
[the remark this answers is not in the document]
mhilbush: Added

bundles/org.openhab.binding.anthem/src/main/java/org/openhab/binding/anthem/internal/handler/AnthemConnectionManager.java
```diff
@@ -0,0 +1,433 @@
...
+        logger.debug("Opening connection to Anthem host {} on port {}", configuration.host, configuration.port);
+        try {
+            Socket socket = new Socket(configuration.host, configuration.port);
+            writer = new BufferedWriter(new OutputStreamWriter(socket.getOutputStream(), StandardCharsets.ISO_8859_1));
+            reader = new BufferedReader(new InputStreamReader(socket.getInputStream(), StandardCharsets.ISO_8859_1));
+            this.socket = socket;
+        } catch (UnknownHostException e) {
+            handler.updateThingStatus(ThingStatus.OFFLINE, ThingStatusDetail.CONFIGURATION_ERROR, "Unknown host");
```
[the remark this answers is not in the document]
mhilbush: Added

bundles/org.openhab.binding.anthem/src/main/java/org/openhab/binding/anthem/internal/handler/AnthemConnectionManager.java
```diff
@@ -0,0 +1,433 @@
...
+            return;
+        } catch (IllegalArgumentException e) {
+            handler.updateThingStatus(ThingStatus.OFFLINE, ThingStatusDetail.CONFIGURATION_ERROR,
+                    "Invalid port number");
```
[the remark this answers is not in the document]
mhilbush: Added

bundles/org.openhab.binding.anthem/src/main/java/org/openhab/binding/anthem/internal/handler/AnthemConnectionManager.java
```diff
@@ -0,0 +1,433 @@
...
+            return;
+        } catch (InterruptedIOException e) {
+            logger.debug("Interrupted while establishing Anthem connection");
+            Thread.currentThread().interrupt();
+            return;
+        } catch (IOException e) {
+            handler.updateThingStatus(ThingStatus.OFFLINE, ThingStatusDetail.COMMUNICATION_ERROR,
+                    "Error opening Anthem connection. Check log.");
```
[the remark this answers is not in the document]
mhilbush: Added

bundles/org.openhab.binding.anthem/src/main/java/org/openhab/binding/anthem/internal/handler/AnthemConnectionManager.java
```diff
@@ -0,0 +1,433 @@
...
+                if (c == COMMAND_TERMINATION_CHAR) {
+                    command = sbReader.toString();
+                    logger.debug("Reader thread sending command to parser: {}", command);
+                    messageParser.parseMessage(command);
+                    sbReader.setLength(0);
+                }
+            }
+        } catch (InterruptedIOException e) {
+            logger.debug("Interrupted while reading");
+            handler.updateThingStatus(ThingStatus.OFFLINE, ThingStatusDetail.COMMUNICATION_ERROR, "Interrupted");
```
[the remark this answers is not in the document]
mhilbush: Added

bundles/org.openhab.binding.anthem/src/main/java/org/openhab/binding/anthem/internal/handler/AnthemConnectionManager.java
```diff
@@ -0,0 +1,433 @@
...
+        } catch (IOException e) {
+            logger.debug("I/O error while reading from socket: {}", e.getMessage());
+            handler.updateThingStatus(ThingStatus.OFFLINE, ThingStatusDetail.COMMUNICATION_ERROR, "I/O Error");
```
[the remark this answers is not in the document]
mhilbush: Added

bundles/org.openhab.binding.anthem/src/main/java/org/openhab/binding/anthem/internal/handler/AnthemConnectionManager.java
```diff
@@ -0,0 +1,433 @@
...
+        } catch (RuntimeException e) {
+            logger.warn("Runtime exception in reader thread", e);
+            handler.updateThingStatus(ThingStatus.OFFLINE, ThingStatusDetail.COMMUNICATION_ERROR, "Runtime exception");
```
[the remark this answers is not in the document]
mhilbush: Added

bundles/org.openhab.binding.anthem/src/main/java/org/openhab/binding/anthem/internal/handler/AnthemConnectionManager.java
```diff
@@ -0,0 +1,433 @@
...
+                    sbReader.setLength(0);
+                }
+            }
+        } catch (InterruptedIOException e) {
+            logger.debug("Interrupted while reading");
+            handler.updateThingStatus(ThingStatus.OFFLINE, ThingStatusDetail.COMMUNICATION_ERROR, "Interrupted");
+        } catch (IOException e) {
+            logger.debug("I/O error while reading from socket: {}", e.getMessage());
+            handler.updateThingStatus(ThingStatus.OFFLINE, ThingStatusDetail.COMMUNICATION_ERROR, "I/O Error");
+        } catch (RuntimeException e) {
```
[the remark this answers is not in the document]
mhilbush: I was worried about an unexpected issue when parsing a command returned from the device, and thought it might be useful to see the stack trace. WDYT?

mhilbush (commented):
I'm addressed almost all of your review feedback with the exception of a couple questions, as well as the refactoring of the AnthemConnectionManager.

bundles/org.openhab.binding.anthem/README.md
jlaur: If I understand this correct, they will both toggle, but one of them is stateless? Will the current mute status be reflected in the state of the "normal" mute channel? If yes, then this would also toggle it? If not, they are actually both stateless?

If the latter is the case, do I get this correct?
- Initial processor status: Muted.
- Initial mute status: OFF
- Initial muteToggle status: OFF
- Set mute to ON. Nothing happens, because already muted.
- Set muteToggle to ON: Now will unmute because it's a toggle.
- Both channels are now ON, but volume is actually unmuted?

I'm very confused. 🙂

bundles/org.openhab.binding.anthem/src/main/java/org/openhab/binding/anthem/internal/handler/AnthemConnectionManager.java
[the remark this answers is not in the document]
jlaur: Wouldn't this be a bug in the command parser class then? It doesn't seem to me than anything external is involved and can throw runtime exceptions here? I just quickly went through that class and you are already catching many exceptions (like array out of bounds) here, so seems quite well guarded. Perhaps consider some unit test coverage for corner cases.

bundles/org.openhab.binding.anthem/src/main/resources/OH-INF/addon/addon.xml
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<addon:addon id="anthem" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+	xmlns:addon="https://openhab.org/schemas/addon/v1.0.0"
+	xsi:schemaLocation="https://openhab.org/schemas/addon/v1.0.0 https://openhab.org/schemas/addon-1.0.0.xsd">
+
+	<type>binding</type>
+	<name>@text/addon.anthem.name</name>
```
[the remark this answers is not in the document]
jlaur: In that case I would encourage you to have a look at:

bundles/org.openhab.binding.anthem/src/main/java/org/openhab/binding/anthem/internal/handler/AnthemConnectionManager.java
jlaur: Of course, only meant to trigger a consideration about this, not a strict requirement. So I would probably either make them loosely coupled (interface/callback) or simply move the logic back into the handler directly as you suggest, as then it would at least be more transparent and life-cycle would be more obvious.

bundles/org.openhab.binding.anthem/src/main/java/org/openhab/binding/anthem/internal/handler/AnthemConnectionManager.java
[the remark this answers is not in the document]
mhilbush: Yes, it definitely would be a bug in the parser if the AV processor sent something that would cause a runtime exception in the parser. I also was pretty confident that the parser was catching most of the expected exceptions. I was just worried about something quite unexpected. I'll remove the RuntimeException in the next iteration.

bundles/org.openhab.binding.anthem/src/main/resources/OH-INF/addon/addon.xml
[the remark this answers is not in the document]
mhilbush: Ok, I'll do that.

But what about the tags I just added to the code for the thing status detail? Will the i18n tool overwrite the properties file that contains those tags?

bundles/org.openhab.binding.anthem/src/main/java/org/openhab/binding/anthem/internal/handler/AnthemConnectionManager.java
mhilbush: It's actually pretty easy to get rid of the `AnthemConnectionManager` completely by moving it all back into the handler. And it will simplify things considerably, so it feels like the right thing to do.

bundles/org.openhab.binding.anthem/README.md
jlaur: Yeah, my example was incomplete because it was created for the case that they both were stateless. If the mute channel is stateful, I see no need for the toggle channel.

bundles/org.openhab.binding.anthem/src/main/resources/OH-INF/addon/addon.xml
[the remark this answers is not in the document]
jlaur: The tool will preserve anything not found in the XML files.

bundles/org.openhab.binding.anthem/README.md
mhilbush: muteToggle channel removed

bundles/org.openhab.binding.anthem/src/main/java/org/openhab/binding/anthem/internal/handler/AnthemConnectionManager.java
mhilbush: AnthemCommunicationManager class refactored into AnthemHandler

bundles/org.openhab.binding.anthem/src/main/java/org/openhab/binding/anthem/internal/handler/AnthemConnectionManager.java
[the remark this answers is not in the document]
mhilbush: RuntimeException catch block removed

bundles/org.openhab.binding.anthem/src/main/resources/OH-INF/addon/addon.xml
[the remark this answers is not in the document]
mhilbush: Generated properties file from xml

bundles/org.openhab.binding.anthem/src/main/resources/OH-INF/thing/thing-types.xml
```diff
@@ -0,0 +1,96 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<thing:thing-descriptions bindingId="anthem"
+	xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+	xmlns:thing="https://openhab.org/schemas/thing-description/v1.0.0"
+	xsi:schemaLocation="https://openhab.org/schemas/thing-description/v1.0.0 https://openhab.org/schemas/thing-description-1.0.0.xsd">
+
+	<thing-type id="anthem">
+		<label>anthem</label>
```
jlaur, line 8:
```suggestion
		<label>Anthem</label>
```

bundles/org.openhab.binding.anthem/src/main/resources/OH-INF/thing/thing-types.xml
```diff
@@ -0,0 +1,96 @@
...
+		</channel-groups>
+
+		<config-description>
+			<parameter name="host" type="text" required="true">
+				<label>Network Address</label>
+				<description>Host name or IP address of the Anthem AV processor</description>
+				<context>network-address</context>
+			</parameter>
+
+			<parameter name="port" type="integer" required="true">
```
jlaur, line 30:
```suggestion
			<parameter name="port" type="integer">
```

bundles/org.openhab.binding.anthem/src/main/resources/OH-INF/thing/thing-types.xml
```diff
@@ -0,0 +1,96 @@
...
+				<label>Network Port</label>
+				<description>Network port number of the Anthem AV processor</description>
+				<default>14999</default>
+				<advanced>true</advanced>
+			</parameter>
+
+			<parameter name="reconnectIntervalMinutes" type="integer" required="true">
```
jlaur, line 37:
```suggestion
			<parameter name="reconnectIntervalMinutes" type="integer">
```

bundles/org.openhab.binding.anthem/src/main/resources/OH-INF/thing/thing-types.xml
```diff
@@ -0,0 +1,96 @@
...
+				<label>Reconnect Interval</label>
+				<description>The time to wait between reconnection attempts (in minutes)</description>
+				<default>2</default>
+				<advanced>true</advanced>
+			</parameter>
+
+			<parameter name="comandDelayMsec" type="integer" required="true">
```
jlaur, line 44:
```suggestion
			<parameter name="comandDelayMsec" type="integer">
```

bundles/org.openhab.binding.anthem/src/main/resources/OH-INF/thing/thing-types.xml
mhilbush: Fixed

bundles/org.openhab.binding.anthem/src/main/resources/OH-INF/thing/thing-types.xml
mhilbush: Fixed

bundles/org.openhab.binding.anthem/src/main/resources/OH-INF/thing/thing-types.xml
mhilbush: Fixed

bundles/org.openhab.binding.anthem/src/main/resources/OH-INF/thing/thing-types.xml
mhilbush: Fixed. Also fixed misspelled parameter name.

bundles/org.openhab.binding.anthem/src/main/resources/OH-INF/thing/thing-types.xml
jlaur: And now the I18N file needs to be regenerated - the price of redundancy. 😉

mhilbush: Ugh. I can almost guarantee I'll forget to do that again in the future. 😭

bundles/org.openhab.binding.anthem/src/main/java/org/openhab/binding/anthem/internal/handler/AnthemConnectionManager.java
jlaur: I will need one more review iteration going through `AnthemCommandParser`.

bundles/org.openhab.binding.anthem/src/main/resources/OH-INF/thing/thing-types.xml
mhilbush: And I had to hand-edit the file after running the tool. Since I changed the name of a config parameter, I had to delete the old properties it left in the file. 🙄

bundles/org.openhab.binding.anthem/src/main/java/org/openhab/binding/anthem/internal/handler/AnthemConnectionManager.java
mhilbush: @jlaur Friendly reminder that you intended to do a final pass through this. No rush, just didn't want it to get lost.

bundles/org.openhab.binding.anthem/src/main/resources/OH-INF/addon/addon.xml
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<addon:addon id="anthem" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+	xmlns:addon="https://openhab.org/schemas/addon/v1.0.0"
+	xsi:schemaLocation="https://openhab.org/schemas/addon/v1.0.0 https://openhab.org/schemas/addon-1.0.0.xsd">
+
+	<type>binding</type>
+	<name>Anthem Binding</name>
+	<description>This is the binding for Anthem AV preamp/processors</description>
```
jlaur, line 8:
```suggestion
	<description>This is the binding for Anthem AV preamp/processors</description>
	<connection>local</connection>
```

bundles/org.openhab.binding.anthem/src/main/java/org/openhab/binding/anthem/internal/handler/AnthemConnectionManager.java
jlaur: Sorry, thanks for the ping. I had another look, and I still believe it would be better to decouple `AthemCommandParser` from the handler. If the `parseMessage` method would return an object containing channel and state, there would be no need for the callbacks anymore, and it would be easy to add unit tests. This is just a suggestion. If you prefer not to refactor this, I won't block the PR because of this, and we can merge as is.

jlaur (changes requested):
Almost there - sorry for the delay.

bundles/org.openhab.binding.anthem/src/main/java/org/openhab/binding/anthem/internal/handler/AnthemConnectionManager.java
mhilbush: I see your point, and I'm not against refactoring this. But I won't be able to test the changes for a least another month or so (the Anthem is in another location and I won't be there until the end of April). I also need to figure out how to handle the cases other than channel/state updates (e.g. property updates, triggering another command to send, etc.). I'll likely need to create several different response object types (one for channel/state updates, one for property updates, etc.). So it's a bit more involved than it was to decouple the sender/reader threads.

I also need to make at least one other enhancement (such as a new channel for sending custom commands), which I also won't be able to test for another month or so.

Therefore, I'd propose that you merge this as is. Then when I make the other change(s), I can also do the decoupling. Let me know if you're ok with that approach.

jlaur: Sounds like a good plan, let's merge it now in good time before next milestone, and you can make additional enhancements in new PR's when you have the opportunity.

bundles/org.openhab.binding.anthem/src/main/java/org/openhab/binding/anthem/internal/handler/AnthemHandler.java
```diff
@@ -0,0 +1,440 @@
...
+        if (!configuration.isValid()) {
+            logger.debug("AnthemHandler: Config of thing '{}' is invalid", thing.getUID().getId());
+            updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.CONFIGURATION_ERROR,
+                    "@text/thing-status-detail-invalidconfig");
+            return;
+        }
+        reconnectIntervalMinutes = configuration.reconnectIntervalMinutes;
+        commandDelayMsec = configuration.commandDelayMsec;
+        asyncInitializeTask = scheduler.submit(this::connect);
+        updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.NONE, "@text/thing-status-detail-connecting");
```
jlaur, line 103: I missed this previously: This should be done before starting the connect task, and with status **UNKNOWN**.

bundles/org.openhab.binding.anthem/src/main/java/org/openhab/binding/anthem/internal/handler/AnthemHandler.java
```diff
@@ -0,0 +1,440 @@
...
+        sendCommand(AnthemCommand.queryActiveInput(zone));
+    }
+
+    private void sendCommand(AnthemCommand command) {
+        logger.debug("Adding command to queue: {}", command);
+        sendQueue.add(command);
+    }
+
+    private synchronized void connect() {
+        updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.NONE, "@text/thing-status-detail-initializing");
```
jlaur, line 255: This should be removed, `initialize()` should set status already before calling this method.

bundles/org.openhab.binding.anthem/src/main/java/org/openhab/binding/anthem/internal/handler/AnthemHandler.java
```diff
@@ -0,0 +1,440 @@
...
+                        localWriter.write(command.toString());
+                        localWriter.flush();
+                    }
+                } catch (InterruptedIOException e) {
+                    logger.debug("Interrupted while sending command");
+                    updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.COMMUNICATION_ERROR,
+                            "@text/thing-status-detail-interrupted");
+                    break;
+                } catch (IOException e) {
+                    logger.warn("Communication error, will try to reconnect. Error: {}", e.getMessage());
```
jlaur, line 392: Status change below is already logged, so this seems unneeded. You could add a third parameter with a description in the `updateStatus()` call. At the very least, this log should be reduced to **debug** if you don't want the exception text shown in the status.

bundles/org.openhab.binding.anthem/src/main/java/org/openhab/binding/anthem/internal/handler/AnthemHandler.java
```diff
@@ -0,0 +1,440 @@
...
+            } catch (IOException e) {
+                logger.debug("Error closing writer: {}", e.getMessage());
+            }
+            this.writer = null;
+        }
+    }
+
+    private synchronized void reconnect() {
+        logger.debug("Attempting to reconnect to the Anthem");
+        updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.COMMUNICATION_ERROR,
```
jlaur, line 368: It seems strange to set "communication error" before attempting a reconnect. If we are not already connected, I'd expect status **offline** already, so no need to set status.

bundles/org.openhab.binding.anthem/src/main/java/org/openhab/binding/anthem/internal/handler/AnthemHandler.java
```diff
@@ -0,0 +1,440 @@
...
+    private void scheduleConnectRetry(long waitMinutes) {
+        logger.debug("Scheduling connection retry in {} minutes", waitMinutes);
+        connectRetryJob = scheduler.schedule(this::connect, waitMinutes, TimeUnit.MINUTES);
+    }
+
+    private synchronized void disconnect() {
+        logger.debug("Disconnecting from Anthem");
+
+        ScheduledFuture<?> localPollingJob = this.pollingJob;
+        if (localPollingJob != null && !localPollingJob.isDone()) {
```
jlaur, line 317: No need to check `isDone`.

bundles/org.openhab.binding.anthem/src/main/java/org/openhab/binding/anthem/internal/handler/AnthemHandler.java
```diff
@@ -0,0 +1,440 @@
...
+                    "@text/thing-status-detail-invalidport");
+            return;
+        } catch (InterruptedIOException e) {
+            logger.debug("Interrupted while establishing Anthem connection");
+            Thread.currentThread().interrupt();
+            return;
+        } catch (IOException e) {
+            updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.COMMUNICATION_ERROR,
+                    "@text/thing-status-detail-openerror");
+            logger.info("Error opening Anthem connection: {}", e.getMessage());
```
jlaur, line 278:
```suggestion
            logger.debug("Error opening Anthem connection: {}", e.getMessage());
```

jlaur (changes requested):
It seems that I didn't previously review the handler. At least I had it marked as "not viewed" and I found some issues with status updates and logging.

bundles/org.openhab.binding.anthem/src/main/java/org/openhab/binding/anthem/internal/handler/AnthemHandler.java
mhilbush: Ok

bundles/org.openhab.binding.anthem/src/main/java/org/openhab/binding/anthem/internal/handler/AnthemHandler.java
mhilbush: Ok

bundles/org.openhab.binding.anthem/src/main/java/org/openhab/binding/anthem/internal/handler/AnthemHandler.java
mhilbush: Reduced to debug

bundles/org.openhab.binding.anthem/src/main/java/org/openhab/binding/anthem/internal/handler/AnthemHandler.java
mhilbush: Ok

bundles/org.openhab.binding.anthem/src/main/java/org/openhab/binding/anthem/internal/handler/AnthemHandler.java
mhilbush: Ok

bundles/org.openhab.binding.anthem/src/main/java/org/openhab/binding/anthem/internal/handler/AnthemHandler.java
mhilbush: Ok

jlaur (approved):
LGTM

morph166955:
@jlaur As the base of this binding was from the androidtv binding which is pending review, I'm assuming it would be prudent to go through the changes here and adapt them into the androidtv PR?

jlaur:
@morph166955 - do you mean that the Android TV binding was based on the Anthem binding? Sure, if we resolved stuff here that you have copied, it would probably be a good idea to go through that and adapt to avoid similar comments in your PR. 🙂

morph166955:
No, @mhilbush pulled from the androidtv binding to build this one.  Either way the changes should apply so I'll make sure to pull them over.

jlaur:
@mhilbush - I forgot the classic reminder about adding a logo. See https://www.openhab.org/docs/developer/bindings/#add-your-binding-s-logo-to-the-openhab-website

mhilbush:
Thanks for the reminder.

Done.

https://github.com/openhab/openhab-docs/pull/2103

PR #14319, opened 2023-02-02
[unifi] Add UoM support for RSSI channel

jlaur:
In order to support UoM, the **rssi** channel type is migrated from `Number` to `Number:Power`.

### DIFF
--- a/bundles/org.openhab.binding.unifi/README.md
+++ b/bundles/org.openhab.binding.unifi/README.md
@@ -176,7 +176,7 @@ The `wirelessClient` information that is retrieved is available as these channel
 | guest      | Switch               | On if this is a guest client                                         | Read        |
 | ap         | String               | Access point (AP) the client is connected to                         | Read        |
 | essid      | String               | Network name (ESSID) the client is connected to                      | Read        |
-| rssi       | Number               | Received signal strength indicator (RSSI) of the client              | Read        |
+| rssi       | Number:Power         | Received signal strength indicator (RSSI) of the client              | Read        |
 | uptime     | Number:Time          | Uptime of the client (in seconds)                                    | Read        |
 | lastSeen   | DateTime             | Date and Time the client was last seen                               | Read        |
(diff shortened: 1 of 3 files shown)


J-N-K:
Does it make sense to do that when we discuss removing these channel types for OH4?

jlaur:
> Does it make sense to do that when we discuss removing these channel types for OH4?

I guess we'll need to go through all bindings when the UoM improvements are done, including this one. It would still need to update the channel with `QuantityType` with appropriate unit to have any kind of UoM support without users having to define the unit manually, right? So I don't see any harm, it just adds another channel for being refactored later.

Hilbrand (approved):
LGTM

lolodomo (approved):
LGTM

lolodomo:
I assume it is a breaking change (Number versus number with dimension) ?

jlaur:
> I assume it is a breaking change (Number versus number with dimension) ?

It is not - unit will be automatically stripped when linked item is a `Number`. I tested this to be sure.

lolodomo:
And having `%unit%` in the default pattern is also not a problem ?

jlaur:
> And having `%unit%` in the default pattern is also not a problem if your item is a simple Number?

No. See openhab/openhab-core#3201.

lolodomo:
Last question for my own understanding, for users having an old item of type Number, will a unit be displayed in (Basic)UI behind the value ? If a unit is displayed, is it the unit used when you built the state (QuantityType) or is it the default unit for Number:Power ?
Sorry for asking, all this is not so clear for me.

jlaur:
> for users having an old item of type Number, will a unit be displayed in (Basic)UI behind the value 

No unit will be displayed. You can also create and link a new `Number` item to a channel with dimension and see the same result.

lolodomo:
> You can also create and link a new `Number` item to a channel with dimension and see the same result.

Good idea, I should take a moment to test how it behaves.

PR #14321, opened 2023-02-03
[nuvo] Fix Essentia G standby mode wake-up

mlobstein:
This PR fixes a long-standing bug when using the Essentia G amplifier. After all zones are switched off, the amplifier will go into a low power mode after 5 minutes. Once in this mode a command from the biding to turn on any zone will not be processed by the amplifier unless it is sent more than once. To fix this problem, the binding will track when the low power mode is entered and then repeat any zone on commands twice to force the amplifier to wake and process the command.


lolodomo (approved):
LGTM, thank you

Issue #14323, opened 2023-02-03
[iCloud] 2FA failing with HTTP 421


maihacke:
According to discussions in the community forum currently there is a problem wit 2FA authentication in some cases:

https://community.openhab.org/t/icloud-binding-3-4-0-2fa-failing/142672
https://community.openhab.org/t/icloud-binding-communication-error/122131/559?u=maihacke

The problem is that even for valid authentication codes the iCloud API returns 421 or 411 HTTP error codes.
The binding stops authentication then. Experiments show that even if 421 or 412 is returned, the authentication code is valid and accept by iCloud.

PR #14324, opened 2023-02-03
[iCloud] Ignore HTTP errors != 400 for 2-FA authentication

maihacke:
This PR should fix https://github.com/openhab/openhab-addons/issues/14323

It changes error handling for iCloud 2-fa authentication to ignore all http error except 400.

### DIFF
--- a/bundles/org.openhab.binding.icloud/src/main/java/org/openhab/binding/icloud/internal/ICloudService.java
+++ b/bundles/org.openhab.binding.icloud/src/main/java/org/openhab/binding/icloud/internal/ICloudService.java
@@ -249,7 +249,7 @@ public class ICloudService {
             this.logger.debug("Exception on code verification. Verification might still be successful.", ex);
             // iCloud API returns different 4xx error codes even if validation is successful
             // currently 400 seems to show that verification "really" failed.
-            if ((ex.getStatusCode() == 400 || ex.getStatusCode() >= 500)) {
+            if (ex.getStatusCode() == 400 || ex.getStatusCode() >= 500) {
                 this.logger.debug("Verification failed with HTTP status {}.", ex.getStatusCode());
                 return false;
             }
--- a/bundles/org.openhab.binding.icloud/src/main/java/org/openhab/binding/icloud/internal/ICloudService.java
+++ b/bundles/org.openhab.binding.icloud/src/main/java/org/openhab/binding/icloud/internal/ICloudService.java
@@ -246,8 +246,12 @@ public class ICloudService {
             this.session.post(AUTH_ENDPOINT + "/verify/trusteddevice/securitycode", JsonUtils.toJson(requestBody),
                     headers);
         } catch (ICloudApiResponseException ex) {
-            logger.debug("Code verification failed.", ex);
-            return false;
+            // iCloud API returns different 4xx error codes even if validation is successful
+            // currently 400 seems to show that verification "really" failed.
+            if (ex.getStatusCode() == 400) {
+                this.logger.debug("Code verification failed.", ex);
+                return false;
+            }
         }
 
         logger.debug("Code verification successful.");
--- a/bundles/org.openhab.binding.icloud/src/main/java/org/openhab/binding/icloud/internal/ICloudService.java
+++ b/bundles/org.openhab.binding.icloud/src/main/java/org/openhab/binding/icloud/internal/ICloudService.java
@@ -246,10 +246,11 @@ public class ICloudService {
             this.session.post(AUTH_ENDPOINT + "/verify/trusteddevice/securitycode", JsonUtils.toJson(requestBody),
                     headers);
         } catch (ICloudApiResponseException ex) {
+            this.logger.debug("Exception on code verification. Verification might still be successful.", ex);
             // iCloud API returns different 4xx error codes even if validation is successful
             // currently 400 seems to show that verification "really" failed.
-            if (ex.getStatusCode() == 400) {
-                this.logger.debug("Code verification failed.", ex);
+            if ((ex.getStatusCode() == 400 || ex.getStatusCode() >= 500)) {
+                this.logger.debug("Verification failed with HTTP status {}.", ex.getStatusCode());
                 return false;
             }
         }


bundles/org.openhab.binding.icloud/src/main/java/org/openhab/binding/icloud/internal/ICloudService.java
```diff
@@ -246,8 +246,12 @@ public boolean validate2faCode(String code) throws IOException, InterruptedExcep
             this.session.post(AUTH_ENDPOINT + "/verify/trusteddevice/securitycode", JsonUtils.toJson(requestBody),
                     headers);
         } catch (ICloudApiResponseException ex) {
-            logger.debug("Code verification failed.", ex);
-            return false;
+            // iCloud API returns different 4xx error codes even if validation is successful
+            // currently 400 seems to show that verification "really" failed.
+            if (ex.getStatusCode() == 400) {
+                this.logger.debug("Code verification failed.", ex);
+                return false;
```
digitaldan, line 253: Not a review, but after quickly reading, i think this is too narrow to only look for a 400 error, as it ignores errors in the 50x range, are their specific 40x errors that are known which would probably be preferable to target , or a small range to check for?

maihacke: Currently known is 412 and 421. I was also in between a positive list and this more radical approach. I decided for the later, because in the end ignoring the errors does not really harm and I feared that we will become aware of other errors which should be ignored one-by-one...

maihacke: How about 

`(ex.getStatusCode() == 400 || ex.getStatusCode() >= 500)`

lolodomo:
Spotless check also failed but it looks unrelated to your code change.

maihacke:
@lolodomo Yes, I also noticed that on my local machine. I didn't had this problem when contributing another PR for the 3.4 release a few weeks ago (https://github.com/openhab/openhab-addons/pull/13691). I think there was something introduced in the 3.4.x-branch later on, which now causes this issue.

lolodomo:
Build is now ok

maihacke:
Any guidance how to proceed? Some users are affected by this problem, fixing is quite easy. But we need to agree on the condition. In my opinion several solutions are valid.

bundles/org.openhab.binding.icloud/src/main/java/org/openhab/binding/icloud/internal/ICloudService.java
lolodomo: @digitaldan :: are you ok with the proposed alternative ?

digitaldan: Yeah, i'm good !

digitaldan: I have not looked at the code, but do we probably want to make sure we have logging in place so if other 4XX codes break this in the future, its easy to see when DEBUG is enable (or this might be hard to track down if its swallowed)

lolodomo: Good idea to also add a DEBUG log showing the status code just before the test.

@maihacke : can you please finalize, with your proposed test and adding a log ?

maihacke: I changed the check as discussed.

bundles/org.openhab.binding.icloud/src/main/java/org/openhab/binding/icloud/internal/ICloudService.java
```diff
@@ -246,8 +246,13 @@ public boolean validate2faCode(String code) throws IOException, InterruptedExcep
             this.session.post(AUTH_ENDPOINT + "/verify/trusteddevice/securitycode", JsonUtils.toJson(requestBody),
                     headers);
         } catch (ICloudApiResponseException ex) {
-            logger.debug("Code verification failed.", ex);
-            return false;
+            this.logger.debug("Exception on code verification. Verification might still be successful.", ex);
```
lolodomo, line 249: The idea was to log `ex.getStatusCode()`, not the full stack trace. So maybe something like
```
logger.debug("Exception on code verification with HTTP status {}. Verification might still be successful.", ex.getStatusCode());
```

lolodomo (changes requested):

bundles/org.openhab.binding.icloud/src/main/java/org/openhab/binding/icloud/internal/ICloudService.java
lolodomo: If in addition to the status code, the stack trace would be helpful, I would suggest to rather use TRACE level:
````
logger.trace("Exception on code verification with HTTP status {}. Verification might still be successful.", ex.getStatusCode(), ex);
```

bundles/org.openhab.binding.icloud/src/main/java/org/openhab/binding/icloud/internal/ICloudService.java
```diff
@@ -246,8 +246,13 @@ public boolean validate2faCode(String code) throws IOException, InterruptedExcep
...
+            // iCloud API returns different 4xx error codes even if validation is successful
+            // currently 400 seems to show that verification "really" failed.
+            if ((ex.getStatusCode() == 400 || ex.getStatusCode() >= 500)) {
```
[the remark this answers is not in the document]
maihacke: ups, happened due to copy&paste. Corrected.

bundles/org.openhab.binding.icloud/src/main/java/org/openhab/binding/icloud/internal/ICloudService.java
maihacke: The full exception is logged in line 249. I avoided logging it twice.

lolodomo: I do not understand your comment. Line 249 is the line I am pointing.

Our request was probably not enough clear. The proposition was to log systematically the status code so that if one day another status code is not caught by your following test, you could easily know what status code is not covered by your test.
Currently you are logging the full stack trace exception, that is even not the data we want to log.
Stack trace is probably even not necessary.

lolodomo:
Your PR is not building but I just discovered you submitted your PR against branch 3.4.x. You have to submit a PR against the main branch.
I close your PR but I can submit a new one against the main branch.

maihacke:
I intentional created the PR against 3.4.x in the expectation of another 3.4.x bugfix release, this PR should be part-of. Is there no such release?

lolodomo:
You should fix the main branch so that your fix will be applied to next coming versions. And then we can decide to "apply" your change to the 3.4.x branch without the need for you to create a specific PR on that branch.

bundles/org.openhab.binding.icloud/src/main/java/org/openhab/binding/icloud/internal/ICloudService.java
maihacke: No worries, but I still don't get why it is a problem to have the stacktrace on DEBUG, but ok it's easy to change that.

maihacke:
Ah ok, could find anything about how this is handled in the dev guide. New PR is here https://github.com/openhab/openhab-addons/pull/14563

Issue #14322, opened 2023-02-03
[fineoffsetweatherstation] no more updates after java.lang.IllegalMonitorStateException warning


TheNinth7:
The binding regularly looses connection to the gateway. The gateway Thing is still shown as ONLINE, but no more updates are received and all the readings (temperature, wind, rain etc.) stay at at their previous value.

## Current Behavior

The following entry is shown in the log, after which no more updates are received:

```
2023-02-02 15:33:28.882 [WARN ] [mmon.WrappedScheduledExecutorService] - Scheduled runnable ended with an exception: 
java.lang.IllegalMonitorStateException: null
	at java.util.concurrent.locks.ReentrantLock$Sync.tryRelease(ReentrantLock.java:149) ~[?:?]
	at java.util.concurrent.locks.AbstractQueuedSynchronizer.release(AbstractQueuedSynchronizer.java:1302) ~[?:?]
	at java.util.concurrent.locks.ReentrantLock.unlock(ReentrantLock.java:439) ~[?:?]
	at org.openhab.binding.fineoffsetweatherstation.internal.service.GatewayQueryService.executeCommand(GatewayQueryService.java:123) ~[?:?]
	at org.openhab.binding.fineoffsetweatherstation.internal.service.FineOffsetGatewayQueryService.executeCommand(FineOffsetGatewayQueryService.java:113) ~[?:?]
	at org.openhab.binding.fineoffsetweatherstation.internal.service.FineOffsetGatewayQueryService.getMeasuredValues(FineOffsetGatewayQueryService.java:93) ~[?:?]
	at org.openhab.binding.fineoffsetweatherstation.internal.handler.FineOffsetGatewayHandler.query(FineOffsetGatewayHandler.java:277) ~[?:?]
	at org.openhab.binding.fineoffsetweatherstation.internal.handler.FineOffsetGatewayHandler.updateLiveData(FineOffsetGatewayHandler.java:159) ~[?:?]
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:515) ~[?:?]
	at java.util.concurrent.FutureTask.runAndReset(FutureTask.java:305) ~[?:?]
	at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:305) ~[?:?]
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128) [?:?]
[...]

PR #14326, opened 2023-02-03
[fineoffsetweatherstation] Fix java.lang.IllegalMonitorStateException

Andy2003:
This fix ensures, that the `ReentrantLock::unlock` method is only called, when the Thread is owning the lock.

resolves: #14322

### DIFF
--- a/bundles/org.openhab.binding.fineoffsetweatherstation/src/main/java/org/openhab/binding/fineoffsetweatherstation/internal/service/GatewayQueryService.java
+++ b/bundles/org.openhab.binding.fineoffsetweatherstation/src/main/java/org/openhab/binding/fineoffsetweatherstation/internal/service/GatewayQueryService.java
@@ -73,14 +73,20 @@ public abstract class GatewayQueryService implements AutoCloseable {
 
     protected byte @Nullable [] executeCommand(String command, byte[] request,
             Function<byte[], Boolean> validateResponse) {
+        try {
+            if (!REQUEST_LOCK.tryLock(30, TimeUnit.SECONDS)) {
+                logger.debug("executeCommand({}): timed out while getting the lock", command);
+                return null;
+            }
+        } catch (InterruptedException e) {
+            logger.debug("executeCommand({}): was interrupted while getting the lock", command);
+            return null;
+        }
+
         byte[] buffer = new byte[2028];
         int bytesRead;
 
         try {
-            if (!REQUEST_LOCK.tryLock(30, TimeUnit.SECONDS)) {
-                logger.trace("executeCommand({}): time out while getting lock", command);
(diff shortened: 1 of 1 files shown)


bundles/org.openhab.binding.fineoffsetweatherstation/src/main/java/org/openhab/binding/fineoffsetweatherstation/internal/service/GatewayQueryService.java
```diff
@@ -73,14 +73,20 @@ public GatewayQueryService(FineOffsetGatewayConfiguration config,
...
-        int bytesRead;
-
         try {
             if (!REQUEST_LOCK.tryLock(30, TimeUnit.SECONDS)) {
-                logger.trace("executeCommand({}): time out while getting lock", command);
+                logger.warn("executeCommand({}): timed out while getting the lock", command);
                 return null;
             }
+        } catch (InterruptedException e) {
+            logger.info("executeCommand({}): was interrupted while getting the lock", command);
```
jlaur, line 82: It seems to me that log level could be reduced to **debug** here. For reference: https://www.openhab.org/docs/developer/guidelines.html#f-logging

bundles/org.openhab.binding.fineoffsetweatherstation/src/main/java/org/openhab/binding/fineoffsetweatherstation/internal/service/GatewayQueryService.java
```diff
@@ -73,14 +73,20 @@ public GatewayQueryService(FineOffsetGatewayConfiguration config,
 
     protected byte @Nullable [] executeCommand(String command, byte[] request,
             Function<byte[], Boolean> validateResponse) {
-        byte[] buffer = new byte[2028];
-        int bytesRead;
-
         try {
             if (!REQUEST_LOCK.tryLock(30, TimeUnit.SECONDS)) {
-                logger.trace("executeCommand({}): time out while getting lock", command);
+                logger.warn("executeCommand({}): timed out while getting the lock", command);
```
jlaur, line 78: Under which circumstances does this happen? Is it really something the user should be warned about?

jlaur (changes requested):
Thanks for the fix! I have only two questions about the log level.

jlaur (approved):
LGTM

jlaur:
@Andy2003, @lolodomo - do you agree we should cherry-pick this into the 3.4.x branch?

Andy2003:
> @Andy2003, @lolodomo - do you agree we should cherry-pick this into the 3.4.x branch?

Definitly

PR #14329, opened 2023-02-03
Fix findbugs instanceof pattern matching

J-N-K:
Findbugs can't properly handle `if (foo instanceof Bar bar)`. The same suppression is done in core.


jlaur:
@J-N-K - thanks! So it fixes this problem, right?

https://github.com/jlaur/openhab-addons/blob/7b8c75b640207302d771d33ac1bdb5289178f63a/bundles/org.openhab.binding.energidataservice/src/main/java/org/openhab/binding/energidataservice/internal/retry/RetryPolicyFactory.java#L43-L46

J-N-K:
Exactly.

J-N-K:
Looks like there are more issues. I'll have a look.

J-N-K:
I really don't understand why these additional suppressions (from core) are needed, but it compiles fine with them and fails with a ton of false-positives if they are missing.

jlaur (approved):
Thanks for taking care of this.

PR #14336, opened 2023-02-05
[neohub] Destroy the web socket client when disposing thing handler

lolodomo:
Signed-off-by: Laurent Garnier <lg.hc@free.fr>

PR #14340, opened 2023-02-05
[yioremote] Stop the web socket client when disposing thing handler

lolodomo:
Avoid a memory leak with web socket client not being stopped.

Also handle properly the case of the thing handler being quickly disposed.

### DIFF
--- a/bundles/org.openhab.binding.yioremote/src/main/java/org/openhab/binding/yioremote/internal/YIOremoteDockHandler.java
+++ b/bundles/org.openhab.binding.yioremote/src/main/java/org/openhab/binding/yioremote/internal/YIOremoteDockHandler.java
@@ -270,7 +270,6 @@ public class YIOremoteDockHandler extends BaseThingHandler {
         } catch (Exception e) {
             logger.debug("Could not stop webSocketClient,  message {}", e.getMessage());
         }
-        webSocketClient.destroy();
     }
 
     @Override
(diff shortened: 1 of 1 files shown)


lolodomo:
Note that this one is clearly more severe.

miloit:
Do you have a Testversion of it @lolodomo . Would be helpful for me ☺️.I am using oh3.4.1.

miloit (approved):
I tested today looks good

@kaikreuzer cherry picking for oh3.4.x would be nice

miloit:
Approved

lolodomo:
@miloit :why did you close my PR?

miloit:
@lolodomo as written via pn it was a mistake..... apologize

lolodomo:
@jlaur : please have a look to this one, it has higher priority than my other PRs related to  using destroy.

jlaur:
> please have a look first to this one, it has higher priority than my other PRs related to using destroy.

My question for this one would be the same: https://github.com/openhab/openhab-addons/pull/14336#issuecomment-1435702918. I think that Jetty manages its own lifecycle so that you shouldn't need to explicitly call `destroy()`, but I'm not 100% sure. Perhaps it would be worth testing? Otherwise I would leave it out of this PR until proven.

lolodomo:
> My question for this one would be the same: [#14336 (comment)](https://github.com/openhab/openhab-addons/pull/14336#issuecomment-1435702918). I think that Jetty manages its own lifecycle so that you shouldn't need to explicitly call `destroy()`, but I'm not 100% sure. Perhaps it would be worth testing? Otherwise I would leave it out of this PR until proven.

The problem here is not the missing destroy but the missing stop.
So I am going tol remove the "destroy" so that you can quickly merge the 2 real problems fixed by this PR.

jlaur:
> The problem here is not the missing destroy but the missing stop.

I understand. 🙂 

> So I am going tol remove the "destroy" so that you can quickly merge the 2 real problems fixed by this PR.

Agreed.

lolodomo:
@jlaur: ready for your review (destroy removed)

jlaur (approved):
Thanks for the fix!

lolodomo:
Thank you.
This is IMHO a candidate for 3.4.x branch.

PR #14342, opened 2023-02-05
[mycroft] Adjust thread name for web socket client

lolodomo:


dalgwen (approved):
👍

lolodomo:
A counter is maybe not the best option because a new thread pool will be created each time the thing handler is initialized.

lolodomo:
@dalgwen : please have a new look.

dalgwen (approved):
OK for me.
But the whole websocket id construction really seems to be some kind of black magic.
I see that you made some merge request on other bindings for the same purpose.
Shouldn't it be a common method somewhere ?
Like in HexUtils or in UIDUtils ?
It is so easy to not using the thread name properly.

lolodomo:
> I see that you made some merge request on other bindings for the same purpose.
> Shouldn't it be a common method somewhere ?
> Like in HexUtils or in UIDUtils ?
> It is so easy to not using the thread name properly.

Could make sense to add a method in the same class as the one that provides createHttpClient and createWebSocketClient.

lolodomo:
Awaiting PR openhab/openhab-core#3359 to be merged to then update this PR

lolodomo:
My initial analysis was wrong, there is no risk of IAE with the current code.

wborn (approved):

PR #14343, opened 2023-02-05
[deconz] Adjust thread name for web socket client

lolodomo:

### DIFF
--- a/bundles/org.openhab.binding.deconz/src/main/java/org/openhab/binding/deconz/internal/handler/DeconzBridgeHandler.java
+++ b/bundles/org.openhab.binding.deconz/src/main/java/org/openhab/binding/deconz/internal/handler/DeconzBridgeHandler.java
@@ -46,6 +46,7 @@ import org.openhab.core.thing.ThingStatusDetail;
 import org.openhab.core.thing.ThingTypeUID;
 import org.openhab.core.thing.binding.BaseBridgeHandler;
 import org.openhab.core.thing.binding.ThingHandlerService;
+import org.openhab.core.thing.util.ThingWebClientUtil;
 import org.openhab.core.types.Command;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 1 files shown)


lolodomo:
A counter is maybe not the best option because a new thread pool will be created each time the thing handler is initialized.

lolodomo:
Whatever the provided name, a new thread pool is created by OH core framework.
Having thread pools with same name is probably allowed.

In case thing UID is too big, we could first try with the thing ID.

PS: I am asking myself if the created thread pool is released when the web socket client is abandoned.

bundles/org.openhab.binding.deconz/src/main/java/org/openhab/binding/deconz/internal/handler/DeconzBridgeHandler.java
```diff
@@ -87,10 +87,11 @@ public DeconzBridgeHandler(Bridge thing, WebSocketFactory webSocketFactory, Asyn
...
         String websocketID = thing.getUID().getAsString().replace(':', '-');
-        if (websocketID.length() < 4) {
-            websocketID = "openHAB-deconz-" + websocketID;
-        } else if (websocketID.length() > 20) {
-            websocketID = websocketID.substring(websocketID.length() - 20);
+        if (websocketID.length() < 4 || websocketID.length() > 20) {
+            websocketID = "deconz-" + thing.getUID().getId();
+        }
+        if (websocketID.length() > 20) {
+            websocketID = "deconz";
```
J-N-K, line 94: That's not a good idea. It should be unique, but repeatable. Maybe hash the full ID and take the last 6 bytes as hex (12 characters + `deconz-` prefix makes a total of 19 characters). I doubt that we'll run into collisions with that and it's nicely reproducible.

lolodomo: It is better if it is unique but not required.
But your idea consisting in hashing the UID is good.

lolodomo: Done

lolodomo: Wait a second...

lolodomo: That is ok now

lolodomo:
Awaiting a new method in core framework.

lolodomo:
Awaiting PR openhab/openhab-core#3359 to be merged to then update this PR

lolodomo:
My initial analysis was wrong, there is no risk of IAE with the current code.

wborn (approved):

Issue #14081, opened 2022-12-28
[boschshc] Control of new thermostats


MiGro1194:
Hello dear OpenHab team,

Bosch has unfortunately updated the thermostats to a new generation after a long time. Reading out the different values still works, but controlling the temperature (target temperature) is no longer possible. The values are displayed correctly in OpenHab as set, but the thermostat remains at the initial temperature. 

Many greetings 
Mitsch

jlaur:
@MiGro1194 - can you share any details about this new thermostat, e.g. model name and log output to support what you are describing? Perhaps @david-pace, @GerdZanker or @coeing would then be willing to have a look.

MiGro1194:
Good morning jlaur,

thank you for your feedback.

The problem is as follows:
- I set the thermostat to 30°C in the Bosch app, for example - but nothing is visible in the log.
- I set the thermostat to 20°C via Openhab - it is visible in the log, but this temperature is not changed on the thermostat.

_2022-12-29 08:30:25.079 [INFO ] [openhab.event.ItemCommandEvent      ] - Item 'KlimasteuerungSpeicher_Zieltemperatur' received command 22
2022-12-29 08:30:25.083 [INFO ] [penhab.event.ItemStatePredictedEvent] - Item 'KlimasteuerungSpeicher_Zieltemperatur' predicted to become 22
2022-12-29 08:30:25.114 [INFO ] [openhab.event.ItemStateChangedEvent ] - Item 'KlimasteuerungSpeicher_Zieltemperatur' changed from 16 °C to 22 °C
2022-12-29 08:30:27.691 [INFO ] [openhab.event.ItemCommandEvent      ] - Item 'KlimasteuerungSpeicher_Zieltemperatur' received command 21
2022-12-29 08:30:27.694 [INFO ] [penhab.event.ItemStatePredictedEvent] - Item 'KlimasteuerungSpeicher_Zieltemperatur' predicted to become 21
2022-12-29 08:30:27.705 [INFO ] [openhab.event.ItemStateChangedEvent ] - Item 'KlimasteuerungSpeicher_Zieltemperatur' changed from 22 °C to 21 °C
2022-12-29 08:31:34.712 [INFO ] [openhab.event.ItemCommandEvent      ] - Item 'KlimasteuerungSpeicher_Zieltemperatur' received command 19
[...]

PR #14352, opened 2023-02-05
[boschshc] allow setting the temperature by disabling read-only state…

vkoop:
… of setpoint-temperature (#14081)

Signed-off-by: Viktor Koop <viktor.koop@gmail.com>


GerdZanker (approved):
Thank you for this Bosch SHC improvement.

jlaur (approved):
Thanks for the fix! This is also consistent with the documentation - already documented as writable.

PR #14362, opened 2023-02-08
[lifx] Add support for newer downlights

wborn:
Allows for discovering the new downlights (PID 121, 122) and their features.

See: https://github.com/LIFX/products/blob/master/products.json


jlaur (approved):
Thanks!

PR #14369, opened 2023-02-09
[jrubyscripting] Update to JRuby 9.3.10.0

jimtng:
This is a maintenance update for openhab 3.4.x branch which contains the latest fixes for jruby 9.3.x.

The main branch for openhab 4 will be upgraded to jruby 9.4.x version which is a major upgrade with "potentially" breaking changes, although so far none is known to conflict with the existing helper libraries.


jlaur:
@kaikreuzer - in continuation of https://github.com/openhab/openhab-addons/pull/14568#issuecomment-1472740945, we also have this PR. This is a little bit more tricky, since the "corresponding" PR for 4.0, #14556, is actually a different upgrade.

kaikreuzer:
Hm, you want to make it more difficult to me with every PR, it seems. 😄 
Ok, I'll then try to collect the 3.x patch release notes by fetching every PR that is marked with "patch" in a given rime range and independent of any milestone being set.

jlaur:
@kaikreuzer - that should work. You could also use your existing filter and add this also: "is:pr is:merged base:3.4.x label:patch". However, it would accomplish the same thing, so I think your approach should be fine.

We should then probably move the "patch" label back to #14568. since it's the actual patch PR. This should be okay with both approaches now.

Issue #14287, opened 2023-01-28
[avmfritz] Fritz DECT 500 lightbulb - not possible to change from color to white mode


outerSpaceToast:
It seems that with the current version of openHAB 3.5 changing color is working now, BUT obviously it is not possible to change back to white mode, after once having set a color option.
Did I miss something?

## Expected Behavior
There is an option to chose white mode (or even better to chose the color temperature) on a Fritz DECT 500 light bulb (created as `HAN_FUN_COLOR_BULB`).
There should be a separate channel for the brightness.

## Current Behavior
As soon as you select a certain color value, the bulb switches to color mode, no more change to white mode possible via openHAB.
Brightness can only be changed, when another manual Dimmer item is linked to the color chaannel.

## Possible Solution
I've tested this changeset from @sommer and @vich-667, which implements a temperature color channel. It is working fine:
https://github.com/vich-667/openhab-addons/pull/1/files
Discussion here: #12427 
What is missing to bring this change into the main development branch?
Can I support somehow?

## Steps to Reproduce (for Bugs)
1. Add bulb as `avmfritz:HAN_FUN_COLOR_BULB`
2. Connect `Color` item to its `color` channel
3. Select a color

## Your Environment
openHAB 3.4.0

Best regards,
Thomas.

vich-667:
Hi @outerSpaceToast!

Thanks four interest on that light blub. I saw that you already mentioned that problem in the old issue! I'm also not happy with the behavior of that Light Blub Thing, but when I had some of my very limited time to play with openhab I had was busy with other things. Yesterday I re setup the dev enviroment, so I hope I can give that old PR a more detailed look and finalize it.

I remember there still was some problem with swiching the light blub totally off. Did it work to you?

I hope my very limited java knowlage is enought to get it working.

KR Vich

outerSpaceToast:
> Yesterday I re setup the dev enviroment, so I hope I can give that old PR a more detailed look and finalize it.

That would be geat :-)

> I remember there still was some problem with swiching the light blub totally off. Did it work to you?

The behaviour is not intuitive, but it works, when you create separate items for the switch, the brightness and the color, which are linked to the color channel.
I think it would be more intuitive, if there were separate channels for that. However, if you know how to do it, then it works well as it is.

> I hope my very limited java knowlage is enought to get it working.

Same situation here, only working with plain C at work.
However, I could merge your changes into the current 3.4-branch locally.

vich-667:
[...]

Issue #12427, opened 2022-03-05
[avmfritz] changing the color of a fritz!dect 500 lightblub isn't working


vich-667:
## Expected Behavior
I have a Fritz!Dect500 LightBlub connected to my Fritzbox 7590, both with the actual firmware. I also own some other stuff for example a Fritz!Dect200 what works well.

In my things inbox a **FRITZ!DECT 500** and a  **HAN_FUN_COLOR_BULB** is shown. The *FRITZ!DECT 500* uses the *ain* shown in the Fritz!Box UI. The *HAN_FUN_COLOR_BULB* uses the same *ain* with an `-1` in addition.

## Current Behavior

- with the **FRITZ!DECT 500** nothing is working
- with the **HAN_FUN_COLOR_BULB** i can change the brightness and switch the light blub on and off, but it is not possible to change hue or saturation. See also community [posts](https://community.openhab.org/t/fritz-dect-500-cannot-control-hue-or-saturation/132076).

## Steps to Reproduce (for Bugs)

1. configure the things from inbox, create items and add them to the semantic model
2. try to change the bightness, hue and saturation.

## Your Environment
* Version used: (e.g., openHAB and add-on versions) openhab 3.2.0-1 and openhab-3.2.0-1
* Environment name and version (e.g. Chrome 76, Java 8, Node.js 12.9, ...): Rasperry Pi3
* Operating System and version (desktop or mobile, Windows 10, Raspbian Buster, ...):  Raspbian GNU/Linux 10 (buster) and OpenJDK 11.0.12
* Fritz!Dect500: 34.10.16.16.015
* Fritzbox 7590: 07.29

sommer:
According to a German [AHA HTTP spec (section 3.5)](https://avm.de/fileadmin/user_upload/Global/Service/Schnittstellen/AHA-HTTP-Interface.pdf) I found online, color bulbs that use color mapping (the FRITZ!DECT 500 seems to have this attribute set) only accept a few select combinations of hue/saturation which can be queried using `getcolordefaults` switchcmd. Any other combination is silently discarded. Crude sample code for OpenHAB 3.2.0 to "snap" selected hue/saturation to a hardcoded table of legal combinations extracted from a sample light bulb might look like this: 
```
diff --git a/bundles/org.openhab.binding.avmfritz/build.sh b/bundles/org.openhab.binding.avmfritz/build.sh
new file mode 100755
index 0000000000..50b7231b2d
[...]

PR #14373, opened 2023-02-09
[avmfritz] Add light blub color temperature support

vich-667:
Added an new channel to HAN_FUN_COLOR_BULB to conrol color temperature. This can be used to set the white color for the FRITZ!DECT 500 or to set it back to white mode if it is in color mode.

- adds a color tempeature channel
- fixes some on off behavior, what was somehow async between OH and the real world Thing

Should fix: https://github.com/openhab/openhab-addons/issues/14287

Thanks to @sommer for posting the base of this code to https://github.com/openhab/openhab-addons/issues/12427


bundles/org.openhab.binding.avmfritz/src/main/java/org/openhab/binding/avmfritz/internal/handler/AVMFritzBaseThingHandler.java
```diff
@@ -218,12 +262,42 @@ private void updateDimmableLight(@Nullable LevelControlModel levelControlModel)
...
 
     private void updateColorLight(@Nullable ColorControlModel colorControlModel,
-            @Nullable LevelControlModel levelControlModel) {
+            @Nullable LevelControlModel levelControlModel, @Nullable SimpleOnOffModel simpleOnOff) {
         if (colorControlModel != null && levelControlModel != null) {
             DecimalType hue = new DecimalType(colorControlModel.hue);
             PercentType saturation = ColorControlModel.toPercent(colorControlModel.saturation);
-            PercentType brightness = new PercentType(levelControlModel.getLevelPercentage());
+            PercentType brightness;
+            if (simpleOnOff.state) {
```
lolodomo, line 270: Take care, simpleOnOff can be null.

bundles/org.openhab.binding.avmfritz/src/main/java/org/openhab/binding/avmfritz/internal/handler/AVMFritzBaseThingHandler.java
```diff
@@ -443,7 +517,42 @@ public void handleCommand(ChannelUID channelUID, Command command) {
...
+                    if (brightness.equals(BigDecimal.ZERO)) {
+                        fritzBox.setSwitch(ain, false);
+                    } else {
+                        fritzBox.setSwitch(ain, true);
+                        fritzBox.setLevelPercentage(ain, brightness);
+                    }
+                }
+                break;
+            case CHANNEL_COLORTEMPERATURE:
+                BigDecimal color_temperature = null;
```
lolodomo, line 529: Please respect convention for variable naming. So rather `colorTemperature`.

bundles/org.openhab.binding.avmfritz/src/main/resources/OH-INF/thing/thing-types.xml
```diff
@@ -353,6 +353,7 @@
 
 		<channels>
 			<channel id="color" typeId="system.color"/>
+			<channel id="color_temperature" typeId="system.color-temperature"/>
```
lolodomo, line 356: Please update the default translation file to add your new channel.

As you are adding a new channel, I assume you should use the new thing upgrade mechanism to trigger the automatic upgrade of things (things created with UI).

lolodomo (changes requested):

lolodomo:
@ccutrer you are our specialist for color temperature ;) Didn't you implemented something that should avoid the hardcoded conversion of values proposed in that PR ?

ccutrer:
I would highly recommend that you add both colorTemperature (as a dimmer), and colorTemperatureAbs (as a Number:Temperature, and system:color-temperature-abs type). You will need to use QuantityType (in Kelvin) when posting updates of your channel's state, and toInvertibleUnit (converting to Kelvin) when accepting commands to make sure automatic unit conversions work (some people will prefer to work in mired/mirek, and some will prefer to work in Kelvin).

There's nothing to help you scale between percent and temperature in Kelvin as in this PR. It looks like your device supports setting a color temperature directly in Kelvin. Does your device not support arbitrary color temperatures, only the few given in your long if/else if structure? If it supports arbitrary, you should definitely just use math and scale the percent value. You may want to do the scaling in mireds instead of Kelvin, since mireds have the nice property that any given interval of say 50 mireds will be the same perceptible change in color, whereas in Kelvin the higher you go in the scale, the larger the interval to achieve the same perceptible change (i.e. 2700K to 3000K is about the same perceptible difference as 5000K to 6000K).

bundles/org.openhab.binding.avmfritz/src/main/java/org/openhab/binding/avmfritz/internal/handler/AVMFritzBaseThingHandler.java
vich-667: done

bundles/org.openhab.binding.avmfritz/src/main/java/org/openhab/binding/avmfritz/internal/handler/AVMFritzBaseThingHandler.java
vich-667: done

bundles/org.openhab.binding.avmfritz/src/main/resources/OH-INF/thing/thing-types.xml
vich-667: Sorry, I'm new here. Do you mand adding the channel like it is done for e.g. the rollershutter here:
https://github.com/openhab/openhab-addons/blob/6e0fe0e0ad3b501977d8c7ea4e7b21fce14749b1/bundles/org.openhab.binding.avmfritz/src/main/resources/OH-INF/i18n/avmfritz.properties#L183

I'm wondering a bit, because the `color` channel is missing in that file.

vich-667:
> I would highly recommend that you add both colorTemperature (as a dimmer), and colorTemperatureAbs (as a Number:Temperature, and system:color-temperature-abs type). You will need to use QuantityType (in Kelvin) when posting updates of your channel's state, and toInvertibleUnit (converting to Kelvin) when accepting commands to make sure automatic unit conversions work (some people will prefer to work in mired/mirek, and some will prefer to work in Kelvin).
> 
> There's nothing to help you scale between percent and temperature in Kelvin as in this PR. It looks like your device supports setting a color temperature directly in Kelvin. Does your device not support arbitrary color temperatures, only the few given in your long if/else if structure? If it supports arbitrary, you should definitely just use math and scale the percent value. You may want to do the scaling in mireds instead of Kelvin, since mireds have the nice property that any given interval of say 50 mireds will be the same perceptible change in color, whereas in Kelvin the higher you go in the scale, the larger the interval to achieve the same perceptible change (i.e. 2700K to 3000K is about the same perceptible difference as 5000K to 6000K).

I will have a look here. I need to test this out if other color temps are working and math can be used. This might take me some days, because currently I'm fighting with some issue that I can not debug.

vich-667:
> I would highly recommend that you add both colorTemperature (as a dimmer), and colorTemperatureAbs (as a Number:Temperature, and system:color-temperature-abs type). You will need to use QuantityType (in Kelvin) when posting updates of your channel's state, and toInvertibleUnit (converting to Kelvin) when accepting commands to make sure automatic unit conversions work (some people will prefer to work in mired/mirek, and some will prefer to work in Kelvin).

I had some look to the hue binding and tried to add that additional channel. At to me it worked fine.


> There's nothing to help you scale between percent and temperature in Kelvin as in this PR. It looks like your device supports setting a color temperature directly in Kelvin. Does your device not support arbitrary color temperatures, only the few given in your long if/else if structure? If it supports arbitrary, you should definitely just use math and scale the percent value. You may want to do the scaling in mireds instead of Kelvin, since mireds have the nice property that any given interval of say 50 mireds will be the same perceptible change in color, whereas in Kelvin the higher you go in the scale, the larger the interval to achieve the same perceptible change (i.e. 2700K to 3000K is about the same perceptible difference as 5000K to 6000K).

The light blub type I have only supports dedicated numbers. Bacause I can configure any value and it jumps to a supported one, I chande to do it via math.
scaling
Because the original steps are in between the scaling in Kelvin and Mired, I used the Mired you reccomended.

ccutrer:
Color temp stuff all looks good to me, except perhaps using some consts for min/max mireds

vich-667:
> Color temp stuff all looks good to me, except perhaps using some consts for min/max mireds

As I put to the inline comment the AHA-HTTP-Inteface docu only gave me that 2700K and 6500K, therefore I didn't introduced any additional magic numbers byside of that ones.

Of course It migt not make sense to calculate the min/max mireds dynamically. If you prefer I also can replace them by const values.

jlaur:
For the new channels added, please add update instructions. For reference, see #14572 and #14587. I believe you can keep target version 1.

vich-667:
@jlaur: I added. can you please check if it's OK?

jlaur:
> I added. can you please check if it's OK?

See https://github.com/openhab/openhab-docs/blob/7bd66d601e0d1421dd18b22e0132ee406127158b/developers/bindings/thing-xml.md#updating-thing-types and https://www.openhab.org/schemas/update-description-1.0.0.xsd. You will need to use `add-channel` for a new channel. Also, you need to set the `thingTypeVersion` property.

vich-667:
@jlaur: Thanks for your hint, I think now I got it.

bundles/org.openhab.binding.avmfritz/src/main/resources/OH-INF/thing/thing-types.xml
vich-667: > As you are adding a new channel, I assume you should use the new thing upgrade mechanism to trigger the automatic upgrade of things (things created with UI).

fixed with the hint from @jlaur

vich-667:
@lolodomo or @cweitkamp: can you please give some feedback here. I think all given comments are fixed.

bundles/org.openhab.binding.avmfritz/src/main/resources/OH-INF/thing/thing-types.xml
```diff
@@ -15,8 +15,14 @@
...
+			<channel id="color_temperature_abs" typeId="system.color-temperature-abs"/>
```
lolodomo, line 19: It would be cool if you could respect the conventions that were recently clarified:
https://next.openhab.org/docs/developer/guidelines.html#naming-convention
What is expected as channel names in your case is `color-temperature `and `color-temperature-abs`.

lolodomo (changes requested):
Just one remark about naming conventions (channel id).

bundles/org.openhab.binding.avmfritz/src/main/resources/OH-INF/thing/thing-types.xml
jlaur: I would recommend sticking to the existing convention in the binding for consistency.

lolodomo (approved):
LGTM, thank upu

PR #14376, opened 2023-02-09
[energidataservice] Initial contribution

jlaur:
This binding integrates the Danish service [Energi Data Service](https://www.energidataservice.dk/), making data available as:

- Channels with current prices.
- An advanced channel with future prices in JSON format.
- Thing action for importing the prices directly into rules.

In this initial version, electricity prices are in focus, specifically the datasets **Elspotprices** and **DatahubPricelist**.

The binding is mainly targeting Denmark because of the tariffs and taxes provided which are specific to the Danish market and in currency DKK. However, spot prices can also be retrieved for other regions, for example Norway and Sweden.

Resolves openhab/openhab-addons#14222

## Prerequisites

- Gson 2.10 is required for deserializing Java records. This dependency currently adds ~300 kB to the size. Hopefully Gson can be upgraded in core before 4.0 - see openhab/openhab-core#3321

## Features

- Optimized service calls: Spot prices are retrieved once per day and tariffs only upon expiry (usually months/years).
- Error handing: Policies for retry strategies are in place to make sure data will be retrieved as soon as possible after failed calls. This includes exponential back-off with jitter.
- Grid company is selectable in configuration.
- Pre-configured filters are included for all known grid companies.
- Filters can be overridden by channel configuration.
- Spot prices and all tariffs are supported.
- Properties shows number of remaining calls as well as timestamp of last/next call.
- Actions for getting prices and performing calculations.

## Note about VAT

See https://github.com/openhab/openhab-addons/issues/14222#issuecomment-1407672100

openhab/openhab-addons#14529 is now merged, and everything related to VAT has been removed from the binding.

## Testing
- 4.0: https://community.openhab.org/t/energi-data-service-binding-4-0-0-0-4-1-0-0/144370 (latest)
- 3.4: https://community.openhab.org/t/energi-data-service-binding-3-4-0-0-4-0-0-0/144381 (backported 27.01.2023, last updated 18.04.2023 23:11)

I have the 3.4.x backport running in my productive system with rules for automation based on the provided thing actions:
https://community.openhab.org/t/dishwasher-price-calculation-automation/139207/4

## Further plans

I've been working on this extensively for a few weeks and all planned MVP features are now implemented. Therefore, I feel the time is right for publishing this pull request. Major changes at this stage are not planned, but can of course be triggered by pull request comments. 😉

A few notes on areas I might revisit on short term:

- Dynamic default value for configuration parameter, if possible: Currently the user has to select price area based on location, either DK1 for West of the Great Belt or DK2 for East of the Great Belt. I would like the value to be defaulted to a calculated value based on the position configured in openHAB's regional settings.
- With dynamic default configuration parameters, currency could also be defaulted based on Locale.
- Channel configuration: Offset for date query parameter filter override.
- Reconsider ways of making the data useful. Check Tibber and aWATTar.

On longer term, gas data could also be added, if there would be an interest in that.

## Future

See openhab/openhab-core#3478 for discussions of more generic character.


masipila:
I did not test this as as I currently have one openHab installation which controls the heating of our house so I did not want to goof with that. However, I read through your code, which seems quite well thought and comprehensive.

By reading the code, I can see that this stores the spot prices as JSON array:
`channel-type.energidataservice.hourlyPrices.description = JSON array with hourly prices from 12 hours ago and onward.
`

Since this is not a proper time series, what is your approach for the basic need of creating a chart of the spot prices? Or is the intention that you will have a Rule that will read the spot prices from this JSON string, parse them and persist them as a real time series so that they can be used like a proper time series?

Cheers,
Markus

jlaur:
> I read through your code, which seems quite well thought and comprehensive.

Thanks!

> By reading the code, I can see that this stores the spot prices as JSON array: `channel-type.energidataservice.hourlyPrices.description = JSON array with hourly prices from 12 hours ago and onward. `
> 
> Since this is not a proper time series, what is your approach for the basic need of creating a chart of the spot prices?

The binding is only a data provider. Charts can be created using other framework components such as persistence. Unfortunately it's currently not possible to provide data with a given timestamp (back and forward in time), so out of the box it will not be possible to create a chart of future prices. You can see historic prices by persisting **CurrentSpotPrice**, for example. 

Persisting of historic and future prices would need to be resolved with openhab/openhab-core#3000 or something similar.

> Or is the intention that you will have a Rule that will read the spot prices from this JSON string, parse them and persist them as a real time series so that they can be plotted as a chart (among other needs for a proper time series)?

The intention was to have some way of exposing future prices, and I didn't want to create tons of channels representing different hours. Initially I used this JSON array in rules to perform calculations. However, this could later be replaced by the action `getPrices` which is much faster. And finally, I created some calculation actions, so in my own current use cases, the calculations can be performed by the binding, and just called by rules. See https://community.openhab.org/t/dishwasher-price-calculation-automation/139207/4

> I mean a chart like this, where the blue area represents the spot prices and the yellow bars indicate the control points (hours) when house heating is allowed to be on.

It looks cool, but this is a problem to be solved outside of the binding IMHO.

masipila:
I agree that fetching the data from a 3rd party and persisting it are two different topics. 

I just left this comment to the PR [#3000](https://github.com/openhab/openhab-core/pull/3000#issuecomment-1225872970). If the conclusion over there will be that bindings must not be able to persist data with future timestamps then I guess all that a binding like this can do is to fetch the data and store it as a serialized JSON array like you are doing here or alternatively have gazillion channels for different timestamps.

And to address the IMHO obvious need to persist that data, there would then have to be for example a Rule which reads the serialized array (or gazillion channels with different timestamps) and then persists the data independently from the Binding that fetched it. 

Did I get this right?

Cheers,
Markus

jlaur:
> I just left this comment to the PR [#3000](https://github.com/openhab/openhab-core/pull/3000#issuecomment-1426761997). If the conclusion over there will be that bindings must not be able to persist data with future timestamps then I guess all that a binding like this can do is to fetch the data and store it as a serialized JSON array like you are doing here or alternatively have gazillion channels for different timestamps.
> 
> And to address the IMHO obvious need to persist that data, there would then have to be for example a Rule which reads the serialized array (or gazillion channels with different timestamps) and then persists the data independently from the Binding that fetched it.
> 
> Did I get this right?

The only thing the binding can do is make the data available. In addition to the two methods you summarized, it can also be made available through actions, which this binding is also offering. When data is made available, users can consume the data in any way they see fit. I don't have any recommendations here. I'm using rules, but the Rest API could also be used to access items, i.e. the serialized JSON.

jlaur:
@openhab/add-ons-maintainers - anyone up for a review, even a partial one? 😉

| Type   | Lines |
| ------ | ----: |
| Code   |  3686 |
| OH-INF |   309 |
| README |   399 |
| Tests  |  5137 |

image

bundles/org.openhab.binding.energidataservice/src/main/java/org/openhab/binding/energidataservice/internal/action/EnergiDataServiceActions.java
```diff
@@ -0,0 +1,381 @@
...
+            logger.warn("{}", e.getMessage());
+            return BigDecimal.ZERO;
+        }
+    }
+
+    @RuleAction(label = "@text/action.calculate-cheapest-period.label", description = "@text/action.calculate-cheapest-period.description")
+    public @ActionOutput(name = "result", type = "java.util.Map<String, Object>") Map<String, Object> calculateCheapestPeriod(
+            @ActionInput(name = "earliestStart", type = "java.time.Instant") Instant earliestStart,
+            @ActionInput(name = "latestEnd", type = "java.time.Instant") Instant latestEnd,
+            @ActionInput(name = "duration", type = "java.time.Duration") Duration duration) {
```
jlaur, line 138: @florian-h05 - perhaps you can advise, (how) is it possible to use these method overloads from JavaScript? I initially tested with DSL rules, but now trying JS:

```javascript
edsActions = actions.get("energidataservice", "energidataservice:service:energidataservice");
var result = edsActions.calculateCheapestPeriod(time.Instant.now(), time.Instant.now().plusSeconds(12*60*60), time.Duration.ofMinutes(90));
```

Result:
```
2023-05-09 22:51:45.308 [ERROR] [enhab.automation.script.file.test.js] - Failed to execute rule Test-EDS-d01ebd58-3ae8-41a8-80cc-769811a64f73: TypeError: invokeMember (calculateCheapestPeriod) on org.openhab.binding.energidataservice.internal.action.EnergiDataServiceActions@7e4decc9 failed due to: no applicable overload found (overloads: [Method[public java.util.Map org.openhab.binding.energidataservice.internal.action.EnergiDataServiceActions.calculateCheapestPeriod(java.time.Instant,java.time.Instant,java.time.Duration,java.util.List,org.openhab.core.library.types.QuantityType)], Method[public java.util.Map org.openhab.binding.energidataservice.internal.action.EnergiDataServiceActions.calculateCheapestPeriod(java.time.Instant,java.time.Instant,java.util.List,java.util.List)], Method[public java.util.Map org.openhab.binding.energidataservice.internal.action.EnergiDataServiceActions.calculateCheapestPeriod(java.time.Instant,java.time.Instant,java.time.Duration,org.openhab.core.library.types.QuantityType)], Method[public java.util.Map org.openhab.binding.energidataservice.internal.action.EnergiDataServiceActions.calculateCheapestPeriod(java.time.Instant,java.time.Instant,java.time.Duration)]], arguments: [DynamicObject<JSOrdinary>@40d066e5 (DefaultLayout), DynamicObject<JSOrdinary>@312634d9 (DefaultLayout), DynamicObject<JSOrdinary>@7cd571ea (DefaultLayout)]): TypeError: invokeMember (calculateCheapestPeriod) on org.openhab.binding.energidataservice.internal.action.EnergiDataServiceActions@7e4decc9 failed due to: no applicable overload found (overloads: [Method[public java.util.Map org.openhab.binding.energidataservice.internal.action.EnergiDataServiceActions.calculateCheapestPeriod(java.time.Instant,java.time.Instant,java.time.Duration,java.util.List,org.openhab.core.library.types.QuantityType)], Method[public java.util.Map org.openhab.binding.energidataservice.internal.action.EnergiDataServiceActions.calculateCheapestPeriod(java.time.Instant,java.time.Instant,java.util.List,java.util.List)], Method[public java.util.Map org.openhab.binding.energidataservice.internal.action.EnergiDataServiceActions.calculateCheapestPeriod(java.time.Instant,java.time.Instant,java.time.Duration,org.openhab.core.library.types.QuantityType)], Method[public java.util.Map org.openhab.binding.energidataservice.internal.action.EnergiDataServiceActions.calculateCheapestPeriod(java.time.Instant,java.time.Instant,java.time.Duration)]], arguments: [DynamicObject<JSOrdinary>@40d066e5 (DefaultLayout), DynamicObject<JSOrdinary>@312634d9 (DefaultLayout), DynamicObject<JSOrdinary>@7cd571ea (DefaultLayout)])
	at <program> (test.js:10)
	at _run (@openhab-globals.js:2)
	at execute (@openhab-globals.js:2)
	at execute (@openhab-globals.js:2)
```

I would like to provide all examples in JavaScript as well in the README, but struggling a bit.

florian-h05: Normally these method overloads work, however it seems that there is an argument type mismatch because of missing "translations" from JS to Java types.
I've just checked the JS Scripting add-on code, we only have "translations" in place for ZonedDateTime and Duration, but not for Instant. Therefore the add-on needs adjustment.

I can have a look into this, should not be too much work or I can give you advice how to do that yourself

jlaur: If you mean add "translation" for `Instant` in the JS Scripting add-on, that would be awesome as it would be a generic fix for all bindings. If you are willing to do this for 4.0, I would highly appreciate that and be grateful. 🙂

I'm in no rush, just wanted to improve the docs, and eventually my plan is to migrate all my rules to JS anyway, so I might as well get started with the rules using the actions of this binding.

florian-h05: I will add this for 4.0, I already have a few code improvements lying around so I’m going to open a PR soon. I’m currently working on the evcc binding, but when I’m finished I will take care of JS Scripting.

I have already migrated most rules to JS (seems reasonable since I am the one doing most work on JS Scripting), IMO it is pretty well developed now and I don’t see any larger changes on the horizon.

florian-h05: @jlaur Do you think that transactions for LocalDate(Time) are also needed anywhere?

florian-h05: @jlaur 
PR is now open, see #14984.
Can you please compile and test it?

jlaur: It works - thanks a lot!

florian-h05: Great!

bundles/org.openhab.binding.energidataservice/src/main/java/org/openhab/binding/energidataservice/internal/action/EnergiDataServiceActions.java
```diff
@@ -0,0 +1,381 @@
...
+        }
+    }
+
+    @RuleAction(label = "@text/action.get-prices.label", description = "@text/action.get-prices.description")
+    public @ActionOutput(name = "prices", type = "java.util.Map<java.time.Instant, java.math.BigDecimal>") Map<Instant, BigDecimal> getPrices() {
+        return getPrices(Arrays.stream(PriceElement.values()).collect(Collectors.toSet()));
+    }
+
+    @RuleAction(label = "@text/action.get-prices.label", description = "@text/action.get-prices.description")
+    public @ActionOutput(name = "prices", type = "java.util.Map<java.time.Instant, java.math.BigDecimal>") Map<Instant, BigDecimal> getPrices(
```
jlaur, line 100: @florian-h05 - sorry to bother you again. I'm progressing with baby steps migrating to JS. Calling this method:

```javascript
edsActions = actions.get("energidataservice", "energidataservice:service:energidataservice");
var priceMap = edsActions.getPrices();
console.log("getPrices" + priceMap.toString());
```

will print:

```
2023-05-17 22:37:58.829 [INFO ] [enhab.automation.script.file.test.js] - getPrices{2023-05-17T14:00:00Z=0.595739002, 2023-05-17T10:00:00Z=0.473758999, 2023-05-18T10:00:00Z=0.915008991, 2023-05-18T06:00:00Z=1.119208973, 2023-05-18T02:00:00Z=1.149668995, 2023-05-17T22:00:00Z=1.195908985, 2023-05-17T18:00:00Z=1.298829029, 2023-05-18T18:00:00Z=1.384179005, 2023-05-18T14:00:00Z=0.979718983, 2023-05-17T15:00:00Z=0.886108998, 2023-05-17T11:00:00Z=0.447169, 2023-05-18T07:00:00Z=1.045108998, 2023-05-18T03:00:00Z=1.179598988, 2023-05-17T23:00:00Z=1.15353899, 2023-05-17T19:00:00Z=1.392889027, 2023-05-18T19:00:00Z=1.319458973, 2023-05-18T15:00:00Z=1.076759022, 2023-05-18T11:00:00Z=0.770899006, 2023-05-17T16:00:00Z=0.931019001, 2023-05-17T12:00:00Z=0.447399, 2023-05-17T08:00:00Z=0.80828899, 2023-05-18T08:00:00Z=0.983449024, 2023-05-18T04:00:00Z=1.194648976, 2023-05-18T00:00:00Z=1.111389027, 2023-05-17T20:00:00Z=1.253629017, 2023-05-18T20:00:00Z=1.263088978, 2023-05-18T16:00:00Z=1.214458973, 2023-05-18T12:00:00Z=0.665819004, 2023-05-17T13:00:00Z=0.461839, 2023-05-17T09:00:00Z=0.717728993, 2023-05-18T09:00:00Z=0.931089009, 2023-05-18T05:00:00Z=1.153018971, 2023-05-18T01:00:00Z=1.109009022, 2023-05-17T21:00:00Z=1.195018971, 2023-05-17T17:00:00Z=0.971458973, 2023-05-18T21:00:00Z=1.18228899, 2023-05-18T17:00:00Z=1.307179005, 2023-05-18T13:00:00Z=0.767628987}
```

But the type seems to be `java.util.HashMap` - which is correct, even if exposed as interface `java.util.Map`. I guess I was somehow expecting a JS `Map`.

Clearly I received the correct data, but I'm not sure how to process it on the JS side. For example, I cannot perform a lookup by key having a js-joda `Instant`. Lookup by string key doesn't work either because of type native Java type.

Do you have any recommendations here?

florian-h05: > But the type seems to be java.util.HashMap - which is correct, even if exposed as interface java.util.Map. I guess I was somehow expecting a JS Map.

That’s the direction that I haven’t a way to do type mappings yet …
openhab-js is internally doing manual Java to JS type mappings, they are available on the `utils` namespace. You could use `utils.javaMapToJsMap` or `utils.javaMapToJsObj`, see https://openhab.github.io/openhab-js/utils.html#.javaMapToJsMap.

For the java.time.Instant - JS-Joda Instant problem, you can either copy the JS type and have the Instants converted to JS-Joda Instant or stringified.

However I’d have to try that myself to give real advice, but this could get a bit complicated.

If you want a Map with the Prices for each hour, I’d rather recommend to create that on the JS side. Use a JS-Joda Instant, get the price for it, put the instant and the price into a js map and so on …

jlaur: Thanks again. I started adding some example JavaScript code in this commit: d512666710debfcb6072d68c3bd8c511f908df1f

`utils.javaMapToJsMap` did the trick, except that keys were still Java `Instant`, so I converted the `Map` to have string keys. I tried also with JS-Joda `Instant` but couldn't make that work. I believe they are immutable, but maybe the `Map` itself doesn't support using such objects as keys. At least now that example works, and it can be improved as I learn more about JavaScript. 🙂

bundles/org.openhab.binding.energidataservice/README.md
```diff
@@ -397,3 +401,24 @@ durationPhases.add(Duration.ofMinutes(104))
 
 var Map<String, Object> result = actions.calculateCheapestPeriod(now.toInstant(), now.plusHours(24).toInstant(), Duration.ofMinutes(236), phases, 0.1 | kWh)
 ```
+
+:::
+
+::: tab JavaScript
+
+```javascript
+edsActions = actions.get("energidataservice", "energidataservice:service:energidataservice");
```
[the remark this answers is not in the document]
jlaur: Yes, good catch. Although it worked without it. 😄

bundles/org.openhab.binding.energidataservice/src/main/java/org/openhab/binding/energidataservice/internal/action/EnergiDataServiceActions.java
florian-h05: The code looks good and simple.


> I tried also with JS-Joda Instant but couldn't make that work. 

What exactly didn't work?

> the Map itself doesn't support using such objects as keys.

It should accept any value as key, see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map.

jlaur: You are right, I tried again, this time with better luck:

```javascript
utils.javaMapToJsMap(edsActions.getPrices()).forEach((value, key) => {
    priceMap.set(time.Instant.parse(key.toString()), value);
});
var hourStart = time.Instant.now().truncatedTo(time.ChronoUnit.HOURS);
console.log("Current total price excl. VAT: " + priceMap.get(hourStart));
```

A bit clumsy (and probably slow) to use `Instant.parse`, don't know if there is a more straight path from `java.time.Instant` to js-joda `Instant`.

florian-h05: > A bit clumsy (and probably slow) to use Instant.parse, don't know if there is a more straight path from java.time.Instant to js-joda Instant.

I think there is a better way, I’ll check that myself and add this to the `utils` so the user doesn’t have to care about the implementation.

florian-h05: I found a much faster way.
Using `Instant.parse` takes 150-200 ms for 50 iterations.
With my approach, 50 iteration only take 7-10 ms.

My benchmark script:

```javascript
utils.javaInstantToJsInstant = function (java) {
  return time.Instant.ofEpochMilli(java.toEpochMilli());
}

var Instant = Java.type('java.time.Instant');
var ins = Instant.now();

console.log('Benchmarking Instant.parse conversion ...')
var start = time.Instant.now().toEpochMilli();

for (var i = 0; i < 50; i++) {
  time.Instant.parse(ins.toString())
}

var end = time.Instant.now().toEpochMilli();
console.log('Benchmark finshed, took ' + (end - start) + ' ms')

console.log('Benchmarking advanced Instant conversion ...')
var start = time.Instant.now().toEpochMilli();

for (var i = 0; i < 50; i++) {
  utils.javaInstantToJsInstant(ins)
}

var end = time.Instant.now().toEpochMilli();
console.log('Benchmark finshed, took ' + (end - start) + ' ms')
```

I will open a PR for the JS Scripting library openhab-js to add this utility function as well as one for ZonedDateTime.

florian-h05: @jlaur PR is open: https://github.com/openhab/openhab-js/pull/267.

bundles/org.openhab.binding.energidataservice/README.md
```diff
@@ -0,0 +1,472 @@
...
+consumptionPhases.push(Quantity("162.162 W"));
+consumptionPhases.push(Quantity("750 W"));
+consumptionPhases.push(Quantity("1500 W"));
+consumptionPhases.push(Quantity("3000 W"));
+consumptionPhases.push(Quantity("1500 W"));
+consumptionPhases.push(Quantity("166.666 W"));
+consumptionPhases.push(Quantity("146.341 W"));
+consumptionPhases.push(Quantity("0 W"));
+
+var result = edsActions.calculateCheapestPeriod(time.Instant.now(), time.Instant.now().plusSeconds(24*60*60), durationPhases, consumptionPhases);
```
jlaur, line 448: @florian-h05 - I was actually surprised to see that this example works, for example with an array of js-joda `Duration`'s being automatically translated into `java.util.List<java.time.Duration>`. It also made we wonder if it could be made in the other direction, so that the `java.util.Map<java.time.Instant, java.math.BigDecimal>` would become a JS map having js-joda `Instant` keys?

jlaur: Sorry, now re-reading this: https://github.com/openhab/openhab-addons/pull/14376#discussion_r1197452833

florian-h05: GraalVM is really great when it comes to mapping JS types to Java types. Most mappings are built-in, however some are manually specified (e.g. Duration). 

I’ve just reread the JavaDoc for GraalVM, and to me it sounds that target type mappings only apply to the JS -> Java direction, see https://www.graalvm.org/truffle/javadoc/org/graalvm/polyglot/HostAccess.Builder.html#targetTypeMapping-java.lang.Class-java.lang.Class-java.util.function.Predicate-java.util.function.Function-.

bundles/org.openhab.binding.energidataservice/src/main/java/org/openhab/binding/energidataservice/internal/action/EnergiDataServiceActions.java
jlaur: @florian-h05 - your optimized utility method works well.

However, I have to back-paddle on this: https://github.com/openhab/openhab-addons/pull/14376#discussion_r1199656891. It does not work. The reason is quite simple: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map and https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Strict_equality. So the key only matches if it's the same object (instance), i.e. `===` is used, not `equals`.

So this is just to conclude that this example should either stay as is with string representation as keys, or alternatively use epoch milliseconds.

florian-h05: > your optimized utility method works well.

Great to hear 👍 

> However, I have to back-paddle on this: https://github.com/openhab/openhab-addons/pull/14376#discussion_r1199656891. It does not work.

Makes sense, unfortunately for objects the lookup is also using `===`, so you need to stick with primitive types like string or number.

jlaur:
@openhab/add-ons-maintainers - no one having time to pick up this PR for review? 🙂 Would be nice as baseline before 4.1 changes for currency, history, etc.

bundles/org.openhab.binding.energidataservice/src/main/java/org/openhab/binding/energidataservice/internal/handler/EnergiDataServiceHandler.java
```diff
@@ -0,0 +1,521 @@
...
+    }
+
+    private void updateHourlyPrices() {
+        if (!isLinked(CHANNEL_HOURLY_PRICES)) {
+            return;
+        }
+        Map<Instant, BigDecimal> spotPriceMap = cacheManager.getSpotPrices();
+        Price[] targetPrices = new Price[spotPriceMap.size()];
+        List<Entry<Instant, BigDecimal>> sourcePrices = spotPriceMap.entrySet().stream()
+                .sorted(Map.Entry.<Instant, BigDecimal> comparingByKey()).toList();
```
fwolter, line 370:
```suggestion
        List<Entry<Instant, BigDecimal>> sourcePrices = spotPriceMap.entrySet().stream()
                .sorted(Map.Entry.comparingByKey()).toList();
```

bundles/org.openhab.binding.energidataservice/src/main/java/org/openhab/binding/energidataservice/internal/handler/EnergiDataServiceHandler.java
```diff
@@ -0,0 +1,521 @@
...
+     * Get cached spot prices or try once to download them if not cached
+     * (usually if no items are linked).
+     *
+     * @return Map of future spot prices
+     */
+    public Map<Instant, BigDecimal> getSpotPrices() {
+        try {
+            downloadSpotPrices();
+        } catch (DataServiceException e) {
+            logger.warn("Error retrieving spot prices");
```
fwolter, line 404: Log the exception's message? Same for below.

bundles/org.openhab.binding.energidataservice/src/main/resources/OH-INF/addon/addon.xml
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<addon:addon id="energidataservice" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+	xmlns:addon="https://openhab.org/schemas/addon/v1.0.0"
+	xsi:schemaLocation="https://openhab.org/schemas/addon/v1.0.0 https://openhab.org/schemas/addon-1.0.0.xsd">
+
+	<type>binding</type>
+	<name>Energi Data Service Binding</name>
+	<description>This is the binding for Energi Data Service.</description>
```
fwolter, line 8: Could you be a bit more verbose, so that somebody who hasn't heard about it gets a clue what this is about?

bundles/org.openhab.binding.energidataservice/src/main/resources/OH-INF/thing/channel-types.xml
```diff
@@ -0,0 +1,32 @@
...
+	xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+	xmlns:thing="https://openhab.org/schemas/thing-description/v1.0.0"
+	xsi:schemaLocation="https://openhab.org/schemas/thing-description/v1.0.0 https://openhab.org/schemas/thing-description-1.0.0.xsd">
+
+	<channel-type id="spot-price">
+		<item-type>Number</item-type>
+		<label>Spot Price</label>
+		<description>Spot price.</description>
+		<category>Price</category>
+		<state readOnly="true" pattern="%.9f"></state>
```
fwolter, line 12: Is 9 positions after the decimal point correct? Same for below (6).

bundles/org.openhab.binding.energidataservice/src/main/resources/OH-INF/thing/thing-service.xml
```diff
@@ -0,0 +1,19 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<thing:thing-descriptions bindingId="energidataservice"
+	xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+	xmlns:thing="https://openhab.org/schemas/thing-description/v1.0.0"
+	xsi:schemaLocation="https://openhab.org/schemas/thing-description/v1.0.0 https://openhab.org/schemas/thing-description-1.0.0.xsd">
+
+	<thing-type id="service">
+
+		<label>Energi Data Service</label>
+		<description>This thing represents the Energi Data Service API.</description>
```
fwolter, line 10:
```suggestion
		<description>This Thing represents the Energi Data Service API.</description>
```

fwolter (changes requested):
This is a reasonable code to test ratio :)

bundles/org.openhab.binding.energidataservice/src/main/resources/OH-INF/addon/addon.xml
jlaur: Good idea, how does it look to you now? The service provides so many different datasets, so it's hard to be very concrete, but at least it should be mentioned it's about energy data.

bundles/org.openhab.binding.energidataservice/src/main/resources/OH-INF/thing/channel-types.xml
jlaur: Yes, it's correct.

bundles/org.openhab.binding.energidataservice/src/main/java/org/openhab/binding/energidataservice/internal/handler/EnergiDataServiceHandler.java
jlaur: Good idea, I must have forgotten this.

jlaur:
@fwolter - thanks for picking up this PR and starting a review. I have fixed your initial comments.

kaikreuzer:
> Would be nice as baseline before 4.1 changes for currency, history, etc.

I was actually not looking at it yet as I assumed you'll depend on https://github.com/openhab/openhab-core/pull/3597.
But I guess it is only the advanced json channel that will get slightly redundant/outdated once we have forecasts in place and I assume you'll want to keep this in place even then anyhow.

jlaur:
> I was actually not looking at it yet as I assumed you'll depend on [openhab/openhab-core#3597](https://github.com/openhab/openhab-core/pull/3597).

It doesn't depend on this, but that will make it possible to store future prices and see graphs with those. However, the binding can already perform calculations based on future prices, or you can make your own calculations based on future prices from the JSON channel or thing actions to retrieve those prices.

> But I guess it is only the advanced json channel that will get slightly redundant/outdated once we have forecasts in place and I assume you'll want to keep this in place even then anyhow.

The JSON channel is already half redundant because of thing actions, but I think it still serves a purpose for some scenarios where you can get future prices without writing rules. However, with openhab/openhab-core#3597 I think it will be fully redundant.

fwolter (approved):
LGTM

[enter your binding's logo reminder here.]

PR #14380, opened 2023-02-10
[dsmr] Declare connection and countries

jlaur:


wborn (approved):

lolodomo:
I will try adding these properties on binding I am maintaining. 
The list of countries is for what exactly? The restricted list of countries in which this binding has value??

wborn:
It's the countries to which the add-on applies, see https://github.com/openhab/openhab-core/issues/2058.
So the big list of add-ons can be filtered so it only shows add-ons that make sense to use in your country. :slightly_smiling_face:

lolodomo (approved):

jlaur:
@Hilbrand - this was based on the documentation. I believe this can/will be used only as a filter, not as a geo-restriction. I don't think it will work in Denmark, though. We do have a P1 port in our Kamstrup meters, which can be read when having the proper encryption keys. See for example https://techblog.vindvejr.dk/?p=523 for the Pow-K+ solution made in Norway and works for both countries at least. There is also a SmartMe module which can be used, but this is a closed cloud-based solution.

PR #14389, opened 2023-02-12
[WIP] Adjust to core-changes (UoM)

J-N-K:
See https://github.com/openhab/openhab-core/pull/3367


Hilbrand (changes requested):
The units for a number of channels of the dsmr binding are not correct. Specifically all volume are not litera. But also not all energy channels have the correct unit. I have no time at the moment to give detailed feedback.

J-N-K:
> The units for a number of channels of the dsmr binding are not correct. Specifically all volume are not liters. But also not all energy channels have the correct unit. I have no time at the moment to give detailed feedback.

The `unit` in a channel definition is used as a replacement for the dimension and will be used only in UI to suggest a unit when creating a new item for a link. It has no technical functionality, so no harm will be done. If the user accepts the suggested item, it behaves like before when no state description was set with a unit that differs from the system default unit.

bundles/org.openhab.binding.roku/src/main/resources/OH-INF/thing/roku.xml
```diff
@@ -175,7 +175,7 @@
 	</channel-type>
 
 	<channel-type id="signalQuality">
-		<item-type>Number:Dimensionless</item-type>
+		<item-type unit="%">Number</item-type>
```
mlobstein, line 178: Shouldn't all the README files be updated to match?
image

PR #14390, opened 2023-02-13
[openwebnet] light switch updates are now triggered

### LINKED ISSUE #14102: [openwebnet] Update trigger does not work on Switch channel
I have been trying to create a Thing listening to general, area or group commands, for example with WHERE=7 for listening to area 7 in the BTicino system. This works but only allows detection of status changes from ON to OFF or vice versa. In the BTicino system an OFF command could be followed by another OFF command, and even the "was updated" trigger is not fired when that happens, therefore making it impossible to process the second event (which should still have an affect on any devices switches ON between the two OFF commands).

Not sure what the expected behavior is, but I thought that the "was updated" trigger should fire also for the second OFF.

Here are screenshots of my setup, first the Thing:

Screenshot - 29_12_2022 , 06_34_23

Then the Item:

Screenshot - 29_12_2022 , 06_32_41

And finally the rule:

Screenshot - 29_12_2022 , 06_35_50

mvalla:
Removed the variable used to check if switch state has changed. 
Now for each light switch update `updateState()` is correctly called on switch channel
Fixes #14102

### DIFF
--- a/bundles/org.openhab.binding.openwebnet/src/main/java/org/openhab/binding/openwebnet/internal/handler/OpenWebNetLightingHandler.java
+++ b/bundles/org.openhab.binding.openwebnet/src/main/java/org/openhab/binding/openwebnet/internal/handler/OpenWebNetLightingHandler.java
@@ -42,7 +42,8 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 /**
- * The {@link OpenWebNetLightingHandler} is responsible for handling commands/messages for a Lighting OpenWebNet device.
+ * The {@link OpenWebNetLightingHandler} is responsible for handling
+ * commands/messages for a Lighting OpenWebNet device.
  * It extends the abstract {@link OpenWebNetThingHandler}.
  *
  * @author Massimo Valla - Initial contribution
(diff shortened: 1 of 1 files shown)


bundles/org.openhab.binding.openwebnet/src/main/java/org/openhab/binding/openwebnet/internal/handler/OpenWebNetLightingHandler.java
```diff
@@ -350,27 +354,17 @@ private void updateOnOffState(Lighting msg) {
...
-                int currentSt = sw[switchId];
-                int newSt = (msg.isOn() ? 1 : 0);
-                if (newSt != currentSt) {
-                    updateState(channelId, (newSt == 1 ? OnOffType.ON : OnOffType.OFF));
-                    sw[switchId] = newSt;
-                    logger.debug("  {} ONOFF CHANGED to {}", ownId, newSt);
-                } else {
-                    logger.debug("  {} ONOFF no change", ownId);
-                }
+                updateState(channelId, (msg.isOn() ? OnOffType.ON : OnOffType.OFF));
```
jlaur, line 367: Could even be simplified to:
```suggestion
                updateState(channelId, OnOffType.from(msg.isOn()));
```

🙂

jlaur (changes requested):
Thanks for the fix! I have added a very minor completely optional suggestion.

bundles/org.openhab.binding.openwebnet/src/main/java/org/openhab/binding/openwebnet/internal/handler/OpenWebNetLightingHandler.java
mvalla: ok!

jlaur (approved):
LGTM

mvalla:
@jlaur I suggest to backport this to 3.4.x, thanks.

jlaur:
> I suggest to backport this to 3.4.x, thanks.

Done.

PR #14399, opened 2023-02-15
Renamed SoC to Soc to comply to release 0.111.0

goebelmeier:
In evcc release 0.111.0 alls occurrences of SoC have been renamed to Soc. This PR will adapt openHAB documentation for this change.


florian-h05 (changes requested):
Whilst I am aware of the fact that evcc renamed `SoC` to `Soc` and adjusted the binding, I decided against renaming the channels because this would be a unneccassary breaking change for openHAB users that doesn't bring any benefit.

The binding is fully compatible to evcc 0.111.x (but currently not to 0.112.x) and there is no need for the openHAB binding to adjust the channel names to this evcc API change.

So I have to at least reject the renaming of the channels. We can discuss if we want to rename the Item names in these examples.

jlaur:
> So I have to at least reject the renaming of the channels. We can discuss if we want to rename the Item names in these examples.

I don't see any reason to do so. The binding should abstract the API, so just because the API renamed a property, we should not need to.

PR #14403, opened 2023-02-16
[ecobee] Remove org.apache.commons

lsiepel:

### DIFF
--- a/bundles/org.openhab.binding.ecobee/src/main/java/org/openhab/binding/ecobee/internal/util/StringUtils.java
+++ b/bundles/org.openhab.binding.ecobee/src/main/java/org/openhab/binding/ecobee/internal/util/StringUtils.java
@@ -23,7 +23,7 @@ import org.eclipse.jdt.annotation.Nullable;
 @NonNullByDefault
 public class StringUtils {
 
-    public static @Nullable String capitalizeWords(@Nullable String input) {
+    public static String capitalizeWords(@Nullable String input) {
         String output = "";
         if (input != null) {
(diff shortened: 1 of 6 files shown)


mhilbush (approved):
Thanks for doing this!

These changes all LGTM.

It's a bit unfortunate that there's no common ExceptionUtils class in core. I would imagine this binding isn't the only one that needs to get the root cause exception.

Same for the utility to capitalize words.

bundles/org.openhab.binding.ecobee/src/main/java/org/openhab/binding/ecobee/internal/util/StringUtils.java
```diff
@@ -0,0 +1,43 @@
...
+
+/**
+ * The {@link StringUtils} class defines static string related methods
+ *
+ * @author Leo Siepel - Initial contribution
+ */
+@NonNullByDefault
+public class StringUtils {
+
+    public static @Nullable String capitalizeWords(@Nullable String input) {
```
lolodomo, line 26: Looks like the method can't return a null value.

lolodomo (changes requested):

bundles/org.openhab.binding.ecobee/src/main/java/org/openhab/binding/ecobee/internal/util/StringUtils.java
lsiepel: Fixed

lolodomo (approved):
LGTM

PR #14408, opened 2023-02-16
[irtrans] Remove org.apache.common

lsiepel:
 - Improved the handlecommand due to allready refactoring 
 - Fixed some checkstyle
 - Fixed some compile warnings.
 

lsiepel:
@kgoderis  are you able to review?

bundles/org.openhab.binding.irtrans/src/main/java/org/openhab/binding/irtrans/internal/handler/EthernetBridgeHandler.java
```diff
@@ -277,9 +277,13 @@ public void handleCommand(ChannelUID channelUID, Command command) {
...
                     if (command instanceof StringType) {
-                        String remoteName = StringUtils.substringBefore(command.toString(), ",");
-                        String irCommandName = StringUtils.substringAfter(command.toString(), ",");
+                        String remoteName = command.toString();
+                        String irCommandName = command.toString();
 
+                        if (command.toString().indexOf(",") > 0) {
+                            remoteName = command.toString().substring(0, command.toString().indexOf(","));
+                            irCommandName = command.toString().substring(command.toString().indexOf(",") + 1);
+                        }
```
jlaur, line 286: I checked the Apache `StringUtils` documentation:
- https://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/StringUtils.html#substringAfter-java.lang.String-java.lang.String-
- https://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/StringUtils.html#substringBefore-java.lang.String-java.lang.String-

And the binding documentation:
- https://www.openhab.org/addons/bindings/irtrans/#channels

If the comma is not found in the command name:
- The old implementation would use full input string for `remoteName` and empty string for `irCommandName`.
- The new implementation will use full input string for both.
- The intention is to ignore the command.

> The IRtrans transceivers store infrared commands in a "remote,command" table, e.g. "telenet,power". Sending the literal text string "telenet,power" to the transceiver will make the transceiver "translate" that into the actual infrared command that will be emitted by the transceiver. A "remote,command" string sent to a Channel that does not match the defined filter will be ignored.

Therefore I think the implementation could be simplified by using https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/lang/String.html#split(java.lang.String,int) with limit 2.

```java
String remoteCommand = command.toString().split(",", 2);
if (remoteCommand.length > 1) {
    String remoteName = remoteCommand[0];
    String irCommandName = remoteCommand[1];
}
// Otherwise ignore command
```

jlaur (changes requested):
Thanks. I analyzed your code and the documentation, and I think this could be further improved.

bundles/org.openhab.binding.irtrans/src/main/java/org/openhab/binding/irtrans/internal/handler/EthernetBridgeHandler.java
jlaur: 👍 I have added a few new comments directly, because the "otherwise ignore command" comment is not fully addressed.

jlaur (changes requested):
Thanks for the update. Just a few comments left.

bundles/org.openhab.binding.irtrans/src/main/java/org/openhab/binding/irtrans/internal/handler/EthernetBridgeHandler.java
```diff
@@ -439,13 +442,13 @@ private String[] getRemoteCommands(String remote, int index) {
         String[] commandList = new String[0];
 
         if (response != null) {
-            String message = stripByteCount(response).split("\0")[0];
-            logger.trace("commands returned {}", message);
-            if (message != null) {
-                if (message.contains("COMMANDLIST")) {
-                    commandList = message.split(",");
+            String[] messages = stripByteCount(response).split("\0");
+            logger.trace("commands returned {}", messages[0]);
```
jlaur, line 446: I suppose this logging will still cause `ArrayIndexOutOfBoundsException` for empty array, since `length` is only checked in next line?

jlaur:
@lsiepel - did you see my comment after the latest additional refactoring commit?

bundles/org.openhab.binding.irtrans/src/main/java/org/openhab/binding/irtrans/internal/handler/EthernetBridgeHandler.java
lsiepel: Double checked and `ArrayIndexOutOfBoundsException` can't happen with java.string.Split so no need to check for length as the array will allways hold 1 value (the original string if the split delemiter was not found)

I updated the fix.

lsiepel: Was partly fixed, should not code anything when to tired.. i corrected it. should be good now.

jlaur (approved):
LGTM

PR #14409, opened 2023-02-16
[keba] Remove org.apache.commons

lsiepel:

Test jar 4.1.0: https://1drv.ms/u/s!AnMcxmvEeupwjq5ZXev_AwpK25Rrvg?e=jQ1Jop


lsiepel:
@kgoderis Could you perform a review/test?

jlaur (changes requested):
This introduces a null annotation warning.

lsiepel:
Fixed the introduced warning and fixed some others.

bundles/org.openhab.binding.keba/src/main/java/org/openhab/binding/keba/internal/handler/KeContactHandler.java
```diff
@@ -529,7 +530,7 @@ public void handleCommand(ChannelUID channelUID, Command command) {
             switch (channelUID.getId()) {
                 case CHANNEL_MAX_PRESET_CURRENT: {
                     if (command instanceof QuantityType<?> quantityCommand) {
-                        QuantityType<?> value = quantityCommand.toUnit("mA");
+                        QuantityType<?> value = Objects.requireNonNull(quantityCommand.toUnit("mA"));
```
jlaur, line 533: This is not safe since `null` is returned in case of error (for example if the unit of the command is not compatible with the target unit here, I would guess): https://www.openhab.org/javadoc/latest/org/openhab/core/library/types/quantitytype#toUnit(java.lang.String)

However, this will cause a `NullPointerException` that would be thrown in next line anyway. So actually the null annotation warning was correct, and now this potential problem is hidden.

Since this is a small issue and it was already not handled appropriately, let's merge anyway.

jlaur (approved):
LGTM

PR #14415, opened 2023-02-17
[neato] Remove org.apache.common

lsiepel:

### DIFF
--- a/bundles/org.openhab.binding.neato/src/main/java/org/openhab/binding/neato/internal/handler/NeatoHandler.java
+++ b/bundles/org.openhab.binding.neato/src/main/java/org/openhab/binding/neato/internal/handler/NeatoHandler.java
@@ -146,7 +146,7 @@ public class NeatoHandler extends BaseThingHandler {
 
         updateState(CHANNEL_STATE, new StringType(neatoState.getRobotState().name()));
 
-        String error = neatoState.getError() != null ? (String) neatoState.getError() : "";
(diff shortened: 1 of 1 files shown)


bundles/org.openhab.binding.neato/src/main/java/org/openhab/binding/neato/internal/handler/NeatoHandler.java
```diff
@@ -146,7 +145,9 @@ private void publishChannels() {
         updateProperty(Thing.PROPERTY_MODEL_ID, neatoState.getMeta().getModelName());
 
         updateState(CHANNEL_STATE, new StringType(neatoState.getRobotState().name()));
-        updateState(CHANNEL_ERROR, new StringType((String) ObjectUtils.defaultIfNull(neatoState.getError(), "")));
+
+        String error = neatoState.getError() != null ? (String) neatoState.getError() : "";
```
lolodomo, line 149: The cast to String is apparently useless as `neatoState.getError()` returns a String

lolodomo (changes requested):

bundles/org.openhab.binding.neato/src/main/java/org/openhab/binding/neato/internal/handler/NeatoHandler.java
lsiepel: True, guess i mixed it up with the original line. anyway, fixed.

lolodomo (approved):
LGTM, thank you

PR #14419, opened 2023-02-17
[onebusaway] Remove org.apache.common

lsiepel:


lolodomo (approved):
LGTM, thank you

PR #14422, opened 2023-02-17
[phc] Remove import org.apache.common

lsiepel:

Provided some tests for the padding, rest is untested, minor refactoring from only from code perspective

### DIFF
--- a/bundles/org.openhab.binding.phc/src/main/java/org/openhab/binding/phc/internal/util/StringUtils.java
+++ b/bundles/org.openhab.binding.phc/src/main/java/org/openhab/binding/phc/internal/util/StringUtils.java
@@ -12,6 +12,7 @@
  */
 package org.openhab.binding.phc.internal.util;
 
+import org.eclipse.jdt.annotation.NonNullByDefault;
 import org.eclipse.jdt.annotation.Nullable;
 
 /**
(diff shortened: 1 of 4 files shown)


bundles/org.openhab.binding.phc/src/main/java/org/openhab/binding/phc/internal/util/StringUtils.java
```diff
@@ -0,0 +1,30 @@
...
+package org.openhab.binding.phc.internal.util;
+
+import org.eclipse.jdt.annotation.Nullable;
+
+/**
+ * The {@link StringUtils} class defines some static string utility methods
+ *
+ * @author Leo Siepel - Initial contribution
+ */
+public class StringUtils {
```
lolodomo, line 22: Can you please add `@NonNullByDefault`

lolodomo (changes requested):

bundles/org.openhab.binding.phc/src/main/java/org/openhab/binding/phc/internal/util/StringUtils.java
lolodomo: Now I see after merging that it was also missing in the new classes you created for the sonos and upnpcontrol bindings.
If you could add them, that would be better.

lsiepel: Yes, done. Will do the others too.

lolodomo (approved):
LGTM, thank you

PR #14165, opened 2023-01-05
Exclude 3rdparty directories from license header generation

jlaur:
Related to #14154, see https://github.com/openhab/openhab-addons/pull/14154#issuecomment-1371319055

PR #14425, opened 2023-02-17
[lametrictime] Move external library to internal code

kaikreuzer:
See https://github.com/openhab/openhab-addons/pull/14165#issuecomment-1432631429

Remarks:
- I moved @syphr42 library code to `src/main/java` (and `src/test/java` for tests).
- I couldn't remove `src/3rdparty/java`, since there are still some external CDDL classes left.
- The CDDL classes do not have openHAB license header, so they won't be an issue for `license:format`.
- I addressed all SAT errors like missing Javadoc headers and author tags.
- ~I didn't address SAT warnings wrt null annotations as a lot of code are pojos that do not really benefit from it and which would have meant significant effort.~


jlaur:
> * I didn't address SAT warnings wrt null annotations as a lot of code are pojos that do not really benefit from it and which would have meant significant effort.

The POJOs could be moved to a folder with the name `dto`, in which case SAT will not complain about missing null annotations.

mhilbush:
> The POJOs could be moved to a folder with the name dto

Yep, for SleepIQ, I just renamed the `model` directory to `dto`.

kaikreuzer:
Thanks, that's a good hint, I didn't know this feature yet!
I've moved the model classes and added null annotations to all other classes now.

jlaur:
> I've moved the model classes and added null annotations to all other classes now.

Awesome, thanks for the effort put into this @kaikreuzer.

jlaur (approved):
LGTM

PR #14426, opened 2023-02-17
[boschshc] Add unit tests

david-pace:
Add additional unit tests for better coverage and stability. This commit boosts the overall src/main/java coverage from 38% to 84%. Especially devices and services are now covered very well (nearly 100%).


david-pace:
Hi @GerdZanker and @jlaur, I wrote lots of unit test for the boschshc binding. Do we need to create a Github issue in openhab-addons for that? I'm not sure whether it's worth mentioning it in the release notes since it's an internal enhancement (not really visibile to users).

jlaur:
> I wrote lots of unit test for the boschshc binding.

👍

> Do we need to create a Github issue in openhab-addons for that?

Not needed.

> I'm not sure whether it's worth mentioning it in the release notes since it's an internal enhancement (not really visibile to users).

We usually only mention PR's which have user impact, so pure technical refactoring (and unit tests) should not be mentioned unless there are other user-related changes in the same PR.

bundles/org.openhab.binding.boschshc/src/main/java/org/openhab/binding/boschshc/internal/devices/bridge/BridgeHandler.java
```diff
@@ -190,8 +193,10 @@ private void scheduleInitialAccess(BoschHttpClient httpClient) {
      * to check if access if possible
      * pairs this Bosch SHC Bridge with the SHC if necessary
      * and starts the first log poll.
+     * <p>
+     * This method is package-protected to enable unit testing.
      */
-    private void initialAccess(BoschHttpClient httpClient) {
+    /* package */ void initialAccess(BoschHttpClient httpClient) {
```
jlaur, line 199: Left-over?

jlaur (changes requested):
Thanks for the tests!

There are some checkstyle warnings left. You could take a look at target/code-analysis/report.html. Please note that **MethodNameCheck** also applies to test methods.

I'll let @GerdZanker and/or @coeing have a look also.

bundles/org.openhab.binding.boschshc/src/main/java/org/openhab/binding/boschshc/internal/devices/bridge/BridgeHandler.java
david-pace: I usually add this comment to indicate that the member is package protected on purpose (and not by accident). I could not find a coding style guide for this aspect in the OpenHAB developer documentation. I can also simply remove `/* package */` if this is preferred.

david-pace:
Thanks for the hint @jlaur :+1: I compared the code analysis report on `main` and my branch. 46 issues were already present on `main`. On my branch there are now 97 issues (mostly related to method naming). Of course I will correct those additional issues on my branch.

However, I would propose changig business logic-related issues in the main code (that already existed on `main` previously and have nothing to do with the unit tests) in a separate pull request to have the purposes of the changes separated cleanly. What do you think?

jlaur:
> Thanks for the hint @jlaur 👍 I compared the code analysis report on `main` and my branch. 46 issues were already present on `main`. On my branch there are now 97 issues (mostly related to method naming). Of course I will correct those additional issues on my branch. I can also correct any issues in existing test code.

I should have used the word "introduced" rather than "left". 🙂 I'm aware that previously some similar issues slipped through.

> However, I would propose changig business logic-related issues in the main code (that already existed on `main` previously and have nothing to do with the unit tests) in a separate pull request to have the purposes of the changes separated cleanly. This will also make reviews easier. What do you think?

I agree. There are also compiler warnings which might be worth checking, and more important than checkstyle warnings. Depending on how much technical debt you want to tackle right now.

bundles/org.openhab.binding.boschshc/src/test/java/org/openhab/binding/boschshc/internal/devices/AbstractBoschSHCHandlerTest.java
```diff
@@ -0,0 +1,113 @@
...
+ * Abstract unit test implementation for all types of handlers.
+ *
+ * @author David Pace - Initial contribution
+ *
+ * @param <T> type of the handler to be tested
+ */
+@ExtendWith(MockitoExtension.class)
+public abstract class AbstractBoschSHCHandlerTest<T extends BoschSHCHandler> {
+
+    private T fixture;
```
david-pace, line 49: When I annotate this class with `@NonNullByDefault` I have the following problem: some members (like in this case `fixture`)  are considered as not always non-null by the compiler. In all other cases i can fix this by annotating the member with `@NonNullByDefault({})`.

The problem is that this only seems to work if the field has a concrete type. With generics (like `T`) in this case this does not work. Do you know a solution for that?

david-pace:
I fixed everything in the test code that I could. One type of issue (2 instances) is left related to `@NonNullByDefault` in combination with generics where I could not find a solution so far (see in-code comment).

bundles/org.openhab.binding.boschshc/src/main/java/org/openhab/binding/boschshc/internal/devices/bridge/BridgeHandler.java
jlaur: If there is no code style rule against this and you prefer it, let's leave it.

bundles/org.openhab.binding.boschshc/src/test/java/org/openhab/binding/boschshc/internal/devices/AbstractBoschSHCHandlerTest.java
jlaur: I can have a look probably tomorrow, I seem to remember having experienced something similar, but don't remember the details.

jlaur: You could try to keep `fixture` non-null and add a constructor:
```java
    public AbstractBoschSHCHandlerTest() {
        fixture = createFixture();
    }
```

david-pace: That works, thanks 😃

I'm not 100% convinced by this solution because now the fixures are created twice (once in the constructor and once in `beforeEach()`), and the ones created in the constructor are in invalid states because the mocks are not initialized yet. But since the fixtures that are created in `beforeEach()` are valid and it works, I would be fine with this solution.

jlaur: Yeah, I'm also not fully satisfied with that hack for the same reason. At least this is test code, so performance should not be much of a concern in this specific case. The same trick is sometimes being used in handlers also because `initialize()` is too late to satisfy the static code analysis.

jlaur (approved):
Thanks!

GerdZanker (commented):
Thank you @david-pace for this huge amount of new tests for the Bosch SHC code.
I learned from several places in your code, how I can write better mockito tests in future pull requests.

PR #14429, opened 2023-02-18
[vizio] Improve handling of TV's self-signed certificate

mlobstein:
Resolve the issue of the binding only attempting to register the self-signed certificate from the TV once during thing initialization. If the TV is off in ECO mode or otherwise not reachable when the thing starts, the registering of the cert will fail and communication cannot be established. With this change the the binding will create its own httpClient with the sslcontextfactory trustAll flag set to true.


lsiepel:
Just out of curiosity, what is the benefit of the trustmanager to accept a self signed certificate, instead of using the client creation (with the sslcintwxtfactoey set to allow all ssl) from https://github.com/openhab/openhab-core/commit/f977a59ebe6d3638e6e39dea2d59d773b0c9cc9e

mlobstein:
> Just out of curiosity, what is the benefit of the trustmanager to accept a self signed certificate, instead of using the client creation (with the sslcontextfactory set to allow all ssl) from [openhab/openhab-core@f977a59](https://github.com/openhab/openhab-core/commit/f977a59ebe6d3638e6e39dea2d59d773b0c9cc9e)

I think it was only way to do it before openhab/openhab-core#3356 was merged 6 days ago. I will look into changing to own http client creation as it would certainly eliminate a lot of complexity here.

bundles/org.openhab.binding.vizio/src/main/java/org/openhab/binding/vizio/internal/handler/VizioHandler.java
```diff
@@ -363,11 +371,14 @@ public void dispose() {
...
-        if (localServiceRegistration != null) {
-            // remove trustmanager service
-            localServiceRegistration.unregister();
-            serviceRegistration = null;
+        try {
+            HttpClient localHttpClient = this.httpClient;
+            if (localHttpClient != null) {
+                localHttpClient.stop();
+            }
+            localHttpClient = null;
```
jlaur, line 379:
```suggestion
            this.httpClient = null;
```

jlaur (changes requested):
Thanks! Only one finding.

bundles/org.openhab.binding.vizio/src/main/java/org/openhab/binding/vizio/internal/handler/VizioHandler.java
jlaur: I see `localHttpClient` is still being nulled, I don't think there is any reason for that? I assumed the local variable was used by mistake rather than the instance variable.

lsiepel (approved):
Besides the open comment from jlaur i have nothing to add.

bundles/org.openhab.binding.vizio/src/main/java/org/openhab/binding/vizio/internal/handler/VizioHandler.java
mlobstein: fixed

jlaur (approved):
LGTM

PR #14430, opened 2023-02-18
[pjlinkdevice] Remove org.apache.common

lsiepel:
- Remove org.apache.commons

While some tests are provided and as far as i can determine all results returned for the subnet methods are identical to the apache lib, some corner cases might happen. I think that is inherent to implementing such method ourselves. 

In totall 5 bindings use this `import org.apache.commons.net.util.SubnetUtils` so i could also prepare a PR for this util class to core repo, and re-use it from these bindings, otherwise, i duplicate this code to the other bindings too. @openhab/core-maintainers  please advise

- dscalarm
- pjlinkdevice
- pclogo
- russound
- zway

Depends on : https://github.com/openhab/openhab-core/pull/4375


lsiepel:
@nils  are you available to review?

nils:
@lsiepel Thanks for the hint, somehow I did not notice your review request. Will check your change sometime next week, latest next weekend...

Are the other bindings using the same bits from org.apache.commons or all different? If they share a lot, I wonder how we can avoid duplicating the code (and if it's worth it, not saying that it's necessarily the case).

lsiepel:
@nils gentle ping, are you able to perform a review?

lsiepel:
@nils another gentle ping

nils:
Before this PR goes into its first anniversary, I finally managed to setup an environment in Docker with some (virtual) devices to find for the discovery.  :fireworks: :champagne: 

image

Looks good to me from a **functionality point of view**, thanks for adding unit tests for the new class! 
I should absolutely do that for the main functionality of this binding as well...

From an **organizational point of view** however, re-implementing this as copies in multiple bindings does not appeal to me very much.
As somebody responsible for this binding, I did not want to worry about this bit of functionality, thus I did use a library. With this change, not only me, but in the worst case also other binding maintainers will have to worry about very special, but reusable functionality. Also over time, the copies might develop each their own set of bugs and glitches.

> In totall 5 bindings use this import org.apache.commons.net.util.SubnetUtils so i could also prepare a PR for this util class to core repo, and re-use it from these bindings, otherwise, i duplicate this code to the other bindings too. @openhab/core-maintainers please advise

I'd appreciate some additional thoughts from some @openhab/core-maintainers too!

lsiepel:
> Before this PR goes into its first anniversary, I finally managed to setup an environment in Docker with some (virtual) devices to find for the discovery. 🎆 🍾
> 
> image
> 
> Looks good to me from a **functionality point of view**, thanks for adding unit tests for the new class! I should absolutely do that for the main functionality of this binding as well...
> 
> From an **organizational point of view** however, re-implementing this as copies in multiple bindings does not appeal to me very much. As somebody responsible for this binding, I did not want to worry about this bit of functionality, thus I did use a library. With this change, not only me, but in the worst case also other binding maintainers will have to worry about very special, but reusable functionality. Also over time, the copies might develop each their own set of bugs and glitches.
> 
> > In totall 5 bindings use this import org.apache.commons.net.util.SubnetUtils so i could also prepare a PR for this util class to core repo, and re-use it from these bindings, otherwise, i duplicate this code to the other bindings too. @openhab/core-maintainers please advise
> 
> I'd appreciate some additional thoughts from some @openhab/core-maintainers too!

Thanks for testing and confirming it works as expected. I’ll raise an issue at the core repo like I did with the string utils class to ultimately create a utility in core making the dependency on org.apache obsolete without creating duplicate code.

jlaur:
> I’ll raise an issue at the core repo like I did with the string utils class to ultimately create a utility in core making the dependency on org.apache obsolete without creating duplicate code.

Are these util methods duplicated/needed by any other bindings? If not, maybe they are not really warranted in core? I didn't check them in details, so I don't know how specific they are, and if Java 17 provides alternatives.

bundles/org.openhab.binding.pjlinkdevice/src/main/java/org/openhab/binding/pjlinkdevice/internal/discovery/AbstractDiscoveryParticipant.java
```diff
@@ -63,7 +60,7 @@ protected ExecutorService getExecutorService() {
     @Override
     protected void startScan() {
         logger.trace("PJLinkProjectorDiscoveryParticipant startScan");
-        Set<InetAddress> addressesToScan = generateAddressesToScan();
+        List<InetAddress> addressesToScan = NetUtil.getFullRangeOfAddressesToScan();
```
lolodomo, line 63: Previous code was ignoring loopback interface and any interface that is not up.
New code does not ?

lsiepel: `NetUtil.getFullRangeOfAddressesToScan()`  calls `NetUtil.getAllInterfaceAddresses()` and that method filters:

https://github.com/openhab/openhab-core/blob/389f6a34343a723442829b7d9841240adc2f2a9c/bundles/org.openhab.core/src/main/java/org/openhab/core/net/NetUtil.java#L344

PR #14431, opened 2023-02-18
Fix build after core-changes

J-N-K:
Thing initialization in core is now more strict, therefore some tests need to be adapted.


itests/org.openhab.binding.modbus.tests/src/main/java/org/openhab/binding/modbus/tests/ModbusPollerThingHandlerTest.java
```diff
@@ -374,24 +374,6 @@ protected boolean matchesSafely(ModbusReadCallback callback) {
         verifyNoMoreInteractions(mockedModbusManager);
     }
 
-    @Test
```
J-N-K, line 377: This test is no longer working because core now prevents initialization of things without a bridge if the thing requires a bridge.

hmerk (approved):
LGTM

kaikreuzer (approved):
Thanks!

PR #14432, opened 2023-02-18
[plugwise] Remove org.apache.common

lsiepel:

### DIFF
--- a/bundles/org.openhab.binding.plugwise/src/main/java/org/openhab/binding/plugwise/internal/PlugwiseUtils.java
+++ b/bundles/org.openhab.binding.plugwise/src/main/java/org/openhab/binding/plugwise/internal/PlugwiseUtils.java
@@ -20,8 +20,6 @@ import java.time.ZoneId;
 import java.time.format.DateTimeFormatter;
 import java.util.Map;
 
-import org.apache.commons.lang3.StringUtils;
-import org.apache.commons.lang3.text.WordUtils;
 import org.eclipse.jdt.annotation.NonNullByDefault;
 import org.eclipse.jdt.annotation.Nullable;
 import org.openhab.binding.plugwise.internal.protocol.InformationResponseMessage;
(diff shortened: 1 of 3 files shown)


bundles/org.openhab.binding.plugwise/src/main/java/org/openhab/binding/plugwise/internal/PlugwiseUtils.java
```diff
@@ -102,7 +100,18 @@ public static void stopBackgroundThread(@Nullable Thread thread) {
...
-        String upperCamel = StringUtils.remove(WordUtils.capitalizeFully(text, new char[] { '_' }), "_");
+        final String delimiter = "_";
+        String upperCamel = "";
+        for (String str : text.split(delimiter)) {
+            String properlyCapitalized = "";
+            if (str.length() > 0) {
+                properlyCapitalized = str.substring(0, 1).toUpperCase();
+            }
+            if (str.length() > 1) {
+                properlyCapitalized = str.substring(1).toLowerCase();
```
wborn, line 111: I think this introduces a bug:

`HELLO_WORLD` will result in `elloorld` instead of `helloWorld`

```suggestion
                properlyCapitalized += str.substring(1).toLowerCase();
```

wborn (changes requested):
Thanks for having a look! I still had this on my todo list. :slightly_smiling_face:

bundles/org.openhab.binding.plugwise/src/main/java/org/openhab/binding/plugwise/internal/PlugwiseUtils.java
wborn: I don't know if the compiler will optimize it to use a `StringBuilder` automatically... if not, explicitly using it would also be better because the concatenation is done in a for loop.

lsiepel: Fxid + stringbuilder + test

wborn (approved):
Thanks!

PR #14433, opened 2023-02-18
[rme] Remove org.apache.common

lsiepel:


bundles/org.openhab.binding.rme/src/main/java/org/openhab/binding/rme/internal/handler/RMEThingHandler.java
```diff
@@ -78,7 +77,7 @@ public void handleCommand(ChannelUID channelUID, Command command) {
 
     @Override
     public void onDataReceived(String receivedLine) {
-        String line = StringUtils.chomp(receivedLine);
+        String line = receivedLine.replaceAll("\r", "").replaceAll("\n", "");
```
jlaur, line 80: Behavior is not the same as previously, but I can't assess if this change will work correctly or not. See:
https://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/StringUtils.html#chomp-java.lang.String-

lsiepel: More of the PR's about apache have this, will look into it and create a comparison test. I'll come back to this

lsiepel: The apach chomp function handles null but don't see any other difference from this test that i created from the examples.
```
 @Test 
    public void chompTest() {
        String[] testStrings = new String[]{
            "", 
            "abc \\r",
            "abc\\n",
            "abc\\r\\n",
            "abc\\r\\n\\r\\n",
            "abc\\n\\r",
            "abc\\n\\rabc",
            "\\r",
            "\\n",
            "\\r\\n" };
       
        for (String tst : testStrings) {
            assertEquals(StringUtils.chomp(tst),tst.replaceAll("\r", "").replaceAll("\n", ""));
            }
    }
```

jlaur: Hmm, strange. In the link I sent: _"Removes one newline from end of a String if it's there, otherwise leave it alone."_ - so it shouldn't remove all newlines. From the examples:
```java
StringUtils.chomp("abc\r\n\r\n") = "abc\r\n"
```

It looks like your result would be "abc". So `chomp` seems to behave differently than documented, or it's not version 2.0.

lsiepel: I did a straight comparison on with all verisons and imports from the bindings itself.
For this binding i'm pretty sure it won't matter anyway, as the goal is to format the receivedline to fit on one line. and later work on that.

lsiepel: @kgoderis are you able to comment on this? I think it is fine, as per test result, but maybe you want to test the binding?

lsiepel: I don't see an issue with this alternative implementation of chomp as leaving the last \r\n does not have any use. 

The main usage is here:
`Pattern responsePattern = Pattern.compile("(.*);(0|1);(0|1);(0|1);(0|1);(0|1);(0|1);(0|1);(0|1);(0|1)");`

lsiepel: @jlaur @J-N-K  Let me know how to proceed. As it seems like the documentation and the actual implementation shows differences. As long as i didn't make a stupid mistake, my alternative should be the same as the apache implementation.

lsiepel: I think i got it sorted.

jlaur: Can you explain why your test found unexpected/undocumented behavior of `StringUtils.chomp(receivedLine)`?

lsiepel: > ted/undocument

:-) because my test was wrong.

jlaur (approved):
Thanks!

jlaur (changes requested):
A null annotation warning has been introduced.

lsiepel:
> A null annotation warning has been introduced.

+100 for null-coalescing operator in Java :-/

jlaur (approved):
LGTM

PR #14434, opened 2023-02-18
[sensibo] Remove org.apache.common

lsiepel:

Created some tests, not all is tested, minor refactoring from code perspective

### DIFF
--- a/bundles/org.openhab.binding.sensibo/src/main/java/org/openhab/binding/sensibo/internal/util/StringUtils.java
+++ b/bundles/org.openhab.binding.sensibo/src/main/java/org/openhab/binding/sensibo/internal/util/StringUtils.java
@@ -1,0 +1,77 @@
+/**
+ * Copyright (c) 2010-2023 Contributors to the openHAB project
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information.
+ *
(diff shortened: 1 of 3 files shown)


bundles/org.openhab.binding.sensibo/src/main/java/org/openhab/binding/sensibo/internal/util/StringUtils.java
seime, line 1: Assuming several other bindings used the now removed dependency - shouldn't this code instead be moved to core so it can be used by all bindings?

seime (commented):
Except for the comment, otherwise LGTM

bundles/org.openhab.binding.sensibo/src/main/java/org/openhab/binding/sensibo/internal/util/StringUtils.java
lsiepel: These specific methods are hardly used elsewhere, i didn;t find any traction on getting these uritlity classes merged to core, also whenever that may happen, this code can easily be adapted. For now the goal is to get all the forbidden packages out of addons repo.

Please approve, so this can get merged.

jlaur (approved):
Thanks!

PR #7941, opened 2020-06-18
[upnpcontrol] UPnP control binding initial contribution

mherwege:
This binding controls UPnP servers and renderers. It finds UPnP servers and renderers in the network and allows querying the server content directory and sending it to a renderer for playback.
It should be device independent and only implement (a subset of) the UPnP AV Media 1.0 standard. There is no device specific logic, and this should be kept out of this binding. E.g. switching on/off a device is not part of the UPnP standard and should be done using a different mechanism not part of this binding.
It is roughly based on previous work done in the Sonos binding.

All media renderers will also be registered as audio sinks in openHAB.

This binding has been in development for a while and has been discussed on the forum in this [thread](https://community.openhab.org/t/dlna-upnp-binding/1790).

The binding uses DynamicStateDescriptionProvider and DynamicCommandDescriptionProvider extensively. Current UI's have limited support for a dynamic refresh of these option lists. See [openhab-core#1505](https://github.com/openhab/openhab-core/pull/1505) and [openhab/openhab-addons#7396 (comment)](https://github.com/openhab/openhab-addons/pull/7396#issuecomment-630773749).
Also, it would benefit from better support of text input fields in UI's, see [openhab-core#1523](https://github.com/openhab/openhab-core/issues/1523).

Anyway, it has proven useful for a number of users in the current state, therefore submitting this for review.

Signed-off-by: Mark Herwege mark.herwege@telenet.be

PR #14439, opened 2023-02-18
[upnpcontrol] Remove org.apache.common

lsiepel:


mherwege:
Shouldn’t this be turned into a XmlUtils class in core now? See the original PR of this binding, and also a comment in the issue [here](https://github.com/openhab/openhab-addons/pull/7941#discussion_r453940350). The Sonos binding has the same requirement.

mherwege:
And maybe have a look at [this](https://github.com/openhab/openhab-addons/issues/7722#issuecomment-679997251). I do think string escaping/unescaping is a general requirement when working with integrations, and a more general solution than doing it separately for every binding could make sense.

lsiepel:
Agree, but after more than two and a half years these utility classes have not been added yet. So I decided to add these PR’s to atleast get this going. I hope it get’s merged and the discussion gets boosted at the topic you mentioned. 
From here it should not be hard to consolidate these utility classes to core. Hope you can confirm the implementation looks fit.

mherwege:
Yes, LGTM

lolodomo (approved):
LGTM, thank you

PR #14441, opened 2023-02-18
[imperihome] Remove org.apache.common

lsiepel:


lolodomo (approved):
LGTM, thank you

jlaur:
@lolodomo - I don't think we should have "enhancement" label on these PR's since it's purely technical refactoring? With this label in will be mentioned in release notes.

lolodomo:
I think it is good to have any code change in the release notes. It can help users identifying what changes could have broken something.

lsiepel:
Are you sure? As that does not seem to be common practice in the past: https://github.com/openhab/openhab-addons/pulls?page=2&q=is%3Apr+is%3Aclosed+no%3Alabel many technicall changes without tags

jlaur:
> I think it is good to have any code change in the release notes. It can help users identifying what changes could have broken something.

I don't think that has been common until now, but of course there can be gray areas where we need to assess individually whether to mention or not. But impression so far has been not to mention small purely technical refactoring PR's. @openhab/add-ons-maintainers, any opinions to have some common ground, so we don't continuously add/remove labels based on different rules?

PR #14442, opened 2023-02-18
[neeo] Remove org.apache.common

lsiepel:


lsiepel:
@tmrobert8 are you able to review?

mherwege:
> I may be one of the last neeo users at this point.

You are not alone, still using it as well.

morph166955:
Awesome, glad to know I'm not the lone ranger.

morph166955:
@lsiepel What is the expected behavior that we are looking for to pass/fail this?  I haven't looked at the code enough to see where the strings would be impacted with this.  I'm loading the jar now.

lsiepel:
> for to pass/fail this? I haven't looked at the code enough to see where the strings would be impacted with this. I'm loading the jar now.

I would call it succesfull if you managed to discover, add and control things. At that point about all changed lines have been called. Maybe use it for some days to see if anything strange happens.

morph166955:
Just to confirm, I'm only replacing the binding piece.  Not the integration right.

lsiepel:
> Just to confirm, I'm only replacing the binding piece. Not the integration right.

Yes.

morph166955:
I've dropped the jar in, it came up no errors so far.  I'll use it for a few days to make sure it behaves as expected.

morph166955:
The binding looked good last night when using the remote.  I didn't see any issues.

morph166955 (approved):
LGTM

morph166955:
I'm fine with the proposed change assuming the dependencies on the core are obviously resolved.

morph166955:
@lsiepel Why are some of the imports pointing to the neeo StringUtils and some are pointed to the openhab core StringUtils?

bundles/org.openhab.io.neeo/src/main/java/org/openhab/io/neeo/internal/util/StringUtils.java
```diff
@@ -0,0 +1,39 @@
...
+
+/**
+ * The {@link StringUtils} class defines some static string utility methods
+ *
+ * @author Leo Siepel - Initial contribution
+ */
+@NonNullByDefault
+public class StringUtils {
+
+    public static String[] split(@Nullable String input, String delimiter) {
```
jlaur, line 30: Looking at the usages, I don't think "support" for null is needed?

bundles/org.openhab.io.neeo/src/main/java/org/openhab/io/neeo/internal/util/StringUtils.java
```diff
@@ -0,0 +1,39 @@
...
+        if (input == null) {
+            return new String[0];
+        }
+        String[] splitted = input.split(delimiter);
+        List<String> list = new ArrayList<String>(Arrays.asList(splitted));
+        list.removeAll(Arrays.asList("", null));
+        return list.toArray(new String[0]);
```
jlaur, line 37: Hmm, two arrays instantiated, I'm wondering if this would be faster and perhaps also simpler to read and understand quickly?
```suggestion
        return Arrays.stream(input.split(delimiter))
            .filter(str -> !str.isEmpty())
            .toArray(String[]::new);
```

jlaur (changes requested):
Thanks! Two minor suggestions.

bundles/org.openhab.io.neeo/src/main/java/org/openhab/io/neeo/internal/util/StringUtils.java
lsiepel: Yeah that looks much simpler.

bundles/org.openhab.io.neeo/src/main/java/org/openhab/io/neeo/internal/util/StringUtils.java
lsiepel: dropped null support, i just liked it as it is more generic.

jlaur: Well, a method accepting `null` as parameter and returning `null` from that is indeed very generic. 😉 Too generic to be useful, IMHO.

In general, for a public method, if such usage doesn't make sense, it would be preferable to throw [IllegalArgumentException](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/lang/IllegalArgumentException.html) to fail fast, and avoid calling it with a `null` parameter, i.e. not make it the responsibility of external methods to handle such error/corner case condition.

Internally we can often use null annotations to avoid such boiler-plate code and instead get help from static code analysis.

jlaur (approved):
LGTM

bundles/org.openhab.io.neeo/src/main/java/org/openhab/io/neeo/internal/util/StringUtils.java
lsiepel: Actually it was returning an empty array, not null. It is how many utility methods are graceful handling null cases. Your suggestion is also a very good pattern, I’ll try to keep it in mind. But anyway this is better now. Thanks

PR #14443, opened 2023-02-18
[exec] Remove org.apache.common

lsiepel:


bundles/org.openhab.binding.exec/src/main/java/org/openhab/binding/exec/internal/handler/ExecHandler.java
```diff
@@ -289,7 +288,7 @@ public void execute() {
 
             outputBuilder.append(errorBuilder.toString());
 
-            String transformedResponse = StringUtils.chomp(outputBuilder.toString());
+            String transformedResponse = outputBuilder.toString().replaceAll("\r", "").replaceAll("\n", "");
```
J-N-K, line 291: This is not the correct equivalent. According to https://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/StringUtils.html#chomp-java.lang.String- only the last `\n`, `\r`, `\r\n` is removed.

J-N-K (changes requested):

bundles/org.openhab.binding.exec/src/main/java/org/openhab/binding/exec/internal/handler/ExecHandler.java
jlaur: See also https://github.com/openhab/openhab-addons/pull/14433#discussion_r1111050883

lsiepel: The tests i used was wrong, got it fixed now including a util class with tests.

lsiepel: This got replaced by a core util class.

bundles/org.openhab.binding.exec/src/main/java/org/openhab/binding/exec/internal/handler/ExecHandler.java
```diff
@@ -292,7 +292,7 @@ public void execute() {
             String transformedResponse = StringUtils.chomp(outputBuilder.toString());
             String transformation = (String) getConfig().get(TRANSFORM);
 
-            if (transformation != null && transformation.length() > 0) {
+            if (transformation != null && transformation.length() > 0 && transformedResponse != null) {
```
jlaur, line 295: In line 292 you could do this to avoid this check:
```java
String transformedResponse = Objects.requireNonNull(StringUtils.chomp(outputBuilder.toString()));
```

It seems not possible for `chomp` to return null for `outputBuilder.toString()`.

jlaur (changes requested):
Close to finish line. 🙂 Can you fix the wrong import as well?

jlaur (approved):
Thanks!

PR #14447, opened 2023-02-19
[sleepiq] Add new channels

mhilbush:
Add 4 new channels for detailed sleep time breakdown.

### DIFF
--- a/bundles/org.openhab.binding.sleepiq/README.md
+++ b/bundles/org.openhab.binding.sleepiq/README.md
@@ -84,6 +84,10 @@ All channels within this group are read-only, except for the sleepNumber and pri
 | todayAverageRespirationRate       | Number         | The average respiration rate for the current day |
 | todayMessage                      | String         | A description of the sleep quality for the current day |
 | todaySleepDurationSeconds         | Number:Time    | The duration of sleep for the current day |
+| todaySleepInBedSeconds            | Number:Time    | The duration of time in bed for the current day |
+| todaySleepOutOfBedSeconds         | Number:Time    | The duration of time out of bed for the current day |
+| todaySleepRestfulSeconds          | Number:Time    | The duration of restful sleep for the current day |
+| todaySleepRestlessSeconds         | Number:Time    | The duration of restless sleep for the current day |
 | monthlySleepIQ                    | Number         | The average Sleep IQ score for the current month |
 | monthlyAverageHeartRate           | Number         | The average heart rate for the current month |
 | monthlyAverageRespirationRate     | Number         | The average respiration rate for the current month |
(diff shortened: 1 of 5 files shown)


jlaur (changes requested):
Thank you, LGTM! Only thing missing is regeneration of the I18N properties file. 🙂

mhilbush:
Right. 🙄

jlaur (approved):
Thanks!

jlaur:
@J-N-K - FYI, new UoM channels added here (to be included in #14389).

J-N-K:
Thing upgrade instructions are missing

mhilbush:
@J-N-K @jlaur I don't mind being a guinea pig to use the new thing update feature, but as there's no documentation, I could use some advice.

So the obvious part is that I need to create the update directory in OH-INF, then create an instructions xml file.

Is there a naming convention for the instructions file? I see the test uses `singleInstructions.xml` and `multiInstructions.xml`. Are those standard names, or can I pick whatever I want? Or, is there a naming convention you would prefer?

Then I guess I create some `add-channel` tags for the new channels. 
- But what is the purpose of the `<label>` tag? Is that duplicating the `<label>` tag in the `thing-types.xml`? 
- Why do I need to provide the `<type>` tag? The type of that channel id is already in the thing types. Or, if there are multiple thing typoes defines, you need that to know which one(s) to which the change should be applied?

Note I can't easily test it because my things are defined in a `.things` file, and I don't currently have a separate test environment.

J-N-K:
- you can pick any name you want
- `label` (`description`) sets the label (description). If you don't use it, the label (description) used should be the one from the channel-type
- the update instructions don't know what's defined in the thing-type, just put the correct channel-type UID (i.e. including binding-id) in there
- don't forget to add `thingTypeVersion` property to the thing-type (otherwise also newly created things will be updated)

mhilbush:
Thanks.

I'll submit something shortly. If you would please look it over that would be great.

lolodomo:
> * don't forget to add `thingTypeVersion` property to the thing-type (otherwise also newly created things will be updated)

Do you mean with a higher value (l2 for example) than the target value (1) in the upgrade file?

mhilbush:
> Do you mean with a higher value (l2 for example) than the target value (1) in the upgrade file?

Yeah, I just was wondering whether it should be that instead of what I wrote.

mhilbush:
BTW, I don't see the `thingTypeVersion` in the thing description xsd. Is there a new version of the xsd that includes that new property?

mhilbush:
This is what I have for the instructions. So I just need to sort out how to handle the `thingTypeVersion` property.
```
<?xml version="1.0" encoding="UTF-8" standalone="yes" ?>
<update:update-descriptions xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
    xmlns:update="https://openhab.org/schemas/update-description/v1.0.0"
    xsi:schemaLocation="https://openhab.org/schemas/update-description/v1.0.0 https://openhab.org/schemas/update-description-1.0.0.xsd">

    <thing-type uid="sleepiq:dualBed">

        <instruction-set targetVersion="1">
            <add-channel id="todaySleepInBedSeconds">
                <type>sleepiq:todaySleepInBedSecondsType</type>
            </add-channel>
            <add-channel id="todaySleepOutBedSeconds">
                <type>sleepiq:todaySleepOutBedSecondsType</type>
            </add-channel>
            <add-channel id="todaySleepRestfulSeconds">
                <type>sleepiq:todaySleepRestfulSecondsType</type>
            </add-channel>
            <add-channel id="todaySleepRestlessSeconds">
                <type>sleepiq:todaySleepRestlessSecondsType</type>
            </add-channel>
        </instruction-set>

    </thing-type>

</update:update-descriptions>
```

mhilbush:
Duh! I see now. I need a property **tag**...
```
<property name="thingTypeVersion">1</property>
```

PR #14451, opened 2023-02-19
[upnpcontrol] Multiple codestyle fixes

lsiepel:
Follow up to https://github.com/openhab/openhab-addons/pull/14439
- Fogotten null annotation
- Removed not used imports (where allready there)
- Some checkstyle warnigns (where allready there)


lolodomo (approved):

mherwege:
@Isiepel Removing the imports creates a problem for the javadoc of some methods, that contain a link to these classes. While maven complains about them, the Eclipse IDE puts them in automatically when creating a link in the javadoc.

lsiepel:
> @isiepel Removing the imports creates a problem for the javadoc of some methods, that contain a link to these classes. While maven complains about them, the Eclipse IDE puts them in automatically when creating a link in the javadoc.

Would a solution be to use the full package name+type in the javadocs instead of using the imports?  I’m using va code, so I cannot check how eclipse handles this.

mherwege:
> Would a solution be to use the full package name+type in the javadocs instead of using the imports?

Yes, it fixes it. I created [#14471](https://github.com/openhab/openhab-addons/pull/14471) for this.

PR #14461, opened 2023-02-20
[nibeheatpump] nibegw: Fix setting debug verboseLevel

gaaf:
Set the correct field when changing verbose level.


lolodomo (approved):
LGTM, thank you

PR #14468, opened 2023-02-20
[tr064] Enhancements, code improvements and fixes

### LINKED ISSUE #10327: [tr064] Address book unavailable due to "IllegalStateException: Duplicate key" when phone numbers are not unique
Hi guys,

I am using openhab3 and tr046 binding and the phonebook feature for fritzbox. In principle all works fine. However, there is a challenge for my setup that is certainly relevant for others.

In my address book, I have the same phone number in two entries. There are two cases:
- same number in two contacts (e.g. fixed line of my brother and his wife)
- same number in one contact (e.g. because entries coming from an exchange server contain the same number as "mobile" and "work")

These cases, here shown for the second case, lead to a warning:

```
2021-03-14 13:00:15.924 [WARN ] [mmon.WrappedScheduledExecutorService] - Scheduled runnable ended with an exception: 
java.lang.IllegalStateException: Duplicate key +491XXXXXXXXXX (attempted merging values Doe, John and Doe, John)
	at java.util.stream.Collectors.duplicateKeyException(Collectors.java:133) ~[?:?]
	at java.util.stream.Collectors.lambda$uniqKeysMapAccumulator$1(Collectors.java:180) ~[?:?]
	at java.util.stream.ReduceOps$3ReducingSink.accept(ReduceOps.java:169) ~[?:?]
...

```

Reason for this error is that the number +491XXXXXXXXXX is configured as both, mobile and work number of John Doe. Unfortunately I cannot change that because it is coming from the exchange server at my work.

The result is that the address book is not available at all. I think it would be reasonable that the value (name) for the key (phone number) is simply overwritten by the last entry.

### LINKED ISSUE #10919: [tr064] Add current downstream/upstream rates for FritzBox
Please provide the information of current downstream and upstream rates in channels

J-N-K:
Fixes #10327
Fixes #10919 

Includes the changes from

https://github.com/smarthomej/addons/pull/21
https://github.com/smarthomej/addons/pull/55
https://github.com/smarthomej/addons/pull/59
https://github.com/smarthomej/addons/pull/79
https://github.com/smarthomej/addons/pull/110
https://github.com/smarthomej/addons/pull/111
https://github.com/smarthomej/addons/pull/119
https://github.com/smarthomej/addons/pull/126
https://github.com/smarthomej/addons/pull/129
https://github.com/smarthomej/addons/pull/135
https://github.com/smarthomej/addons/pull/139
https://github.com/smarthomej/addons/pull/144
https://github.com/smarthomej/addons/pull/160
https://github.com/smarthomej/addons/pull/179
https://github.com/smarthomej/addons/pull/188
https://github.com/smarthomej/addons/pull/273


bundles/org.openhab.binding.tr064/src/main/resources/OH-INF/thing/thing-types.xml
```diff
@@ -146,7 +169,7 @@
 			</parameter>
 			<parameter name="macOnline" type="text" multiple="true">
 				<label>MAC Online</label>
-				<description>List of MACs for "online" status detection (format: 11:11:11:11:11:11). One MAC-Address per row.</description>
+				<description>List of MACs for "online" status detection (format: 11:11:11:11:11:11).</description>
```
lsiepel, line 172: This was recently fixed, please revert.

bundles/org.openhab.binding.tr064/src/main/resources/channels.xml
```diff
@@ -65,7 +65,7 @@
 			serviceId="urn:X_AVM-DE_HostFilter-com:serviceId:X_AVM-DE_HostFilter1"/>
 		<getAction name="GetWANAccessByIP" argument="NewDisallow">
 			<parameter name="NewIPv4Address" thingParameter="wanBlockIPs"
-				pattern="^((25[0-5]|(2[0-4]|1\d|[1-9]|)\d)(\.(?!$)|$)){4}$"/>
+				pattern="[1-9]\d{0,2}\.(\d{1,3}\.){2}[1-9]\d{0,2}(\s*#.*)*"/>
```
lsiepel, line 68: This was recently fixed, please revert.

bundles/org.openhab.binding.tr064/README.md
```diff
@@ -70,12 +74,17 @@ If the `PHONEBOOK` profile shall be used, it is necessary to retrieve the phoneb
 The `phonebookInterval` is used to set the refresh cycle for phonebooks.
 It defaults to 600 seconds, and it can be set to 0 if phonebooks are not used.
 
-Some parameters (e.g. `macOnline`, `wanBlockIPs`) accept lists. 
-List items are configured one per line in the UI, or are comma separated values when using textual config.
-These parameters that accept list can also contain comments.
+Parameters that accept lists (e.g. `macOnline`, `wanBlockIPs`) can contain comments.
```
lsiepel, line 77: This was recently fixed, please revert.

bundles/org.openhab.binding.tr064/src/main/resources/OH-INF/thing/thing-types.xml
J-N-K: Why is that necessary?

bundles/org.openhab.binding.tr064/src/main/resources/channels.xml
J-N-K: The "fix" prevents adding comments. This is wrong and against the documentation.

bundles/org.openhab.binding.tr064/src/main/java/org/openhab/binding/tr064/internal/phonebook/Tr064PhonebookImpl.java
```diff
@@ -33,37 +33,42 @@
...
+                    number -> normalizeNumber(number.getValue()), number -> contactName, this::mergeSameContactNames));
+        }).collect(HashMap::new, HashMap::putAll, HashMap::putAll);
+        logger.debug("Downloaded phonebook {}: {}", phonebookName, phonebook);
     }
 
     // in case there are multiple phone entries with same number -> name mapping, i.e. in phonebooks exported from
     // mobiles containing multiple accounts like: local, cloudprovider1, messenger1, messenger2,...
     private String mergeSameContactNames(String nameA, String nameB) {
-        if (nameA.equals(nameB)) {
+        if (nameA != null && nameA.equals(nameB)) {
```
lsiepel, line 71: This was recently fixed, please revert. 
`nameA` cannot be null

bundles/org.openhab.binding.tr064/src/main/resources/OH-INF/thing/thing-types.xml
lsiepel: Because the documentation was unclear. Details here https://github.com/openhab/openhab-addons/pull/14291 and https://github.com/openhab/openhab-addons/issues/13798

J-N-K: I don't understand why one would like to add a comma-separated list there, but ok.

bundles/org.openhab.binding.tr064/src/main/resources/channels.xml
lsiepel: Sorry, i stared too much at the ip-address part and forgot about the optional comment.

bundles/org.openhab.binding.tr064/src/main/java/org/openhab/binding/tr064/internal/phonebook/Tr064PhonebookImpl.java
J-N-K: I'm not 100% sure this is correct, but in case the name is `null` the check must be done at another place. Please see my fix.

bundles/org.openhab.binding.tr064/src/main/resources/channels.xml
```diff
@@ -244,6 +196,24 @@
 			serviceId="urn:WANCIfConfig-com:serviceId:WANCommonInterfaceConfig1"/>
 		<getAction name="GetTotalBytesSent" argument="NewTotalBytesSent"/>
 	</channel>
+	<channel name="wanCurrentDownstreamBitrate" label="Current Downstream Rate">
```
jlaur, line 199: Thing upgrade instructions seems missing for these new channels?

bundles/org.openhab.binding.tr064/src/main/resources/channels.xml
```diff
@@ -138,74 +138,26 @@
...
+		advanced="true">
 		<item type="Switch"/>
 		<service deviceType="urn:dslforum-org:device:LANDevice:1" serviceId="urn:LanDeviceHosts-com:serviceId:Hosts1"/>
 		<getAction name="GetSpecificHostEntry" argument="NewActive">
 			<parameter name="NewMACAddress" thingParameter="macOnline"
 				pattern="([0-9A-Fa-f]{2}:){5}[0-9A-Fa-f]{2}(\s*#.*)*"/>
 		</getAction>
 	</channel>
-	<channel name="macIP" label="MAC IP" description="IP of the device with the given MAC">
+	<channel name="macOnlineIpAddress" label="MAC Online IP" description="IP of the device with the given MAC"
```
jlaur, line 150: Is this a channel rename? The README has changed the description to "IP of the MAC (uses same parameter as `macOnline`)"

jlaur (changes requested):
Thanks! I have added a few comments, mostly questions. Also, please regenerate the I18N properties file.

bundles/org.openhab.binding.tr064/src/main/java/org/openhab/binding/tr064/internal/phonebook/Tr064PhonebookImpl.java
jlaur: @lsiepel, maybe you can recall from #13873?

bundles/org.openhab.binding.tr064/src/main/resources/channels.xml
J-N-K: The `channels.xml` only defines the channels. All channels are added/removed dynamically based on what the device reports as supported. This check is done during initialization of the handler (every time, not only the first time).

bundles/org.openhab.binding.tr064/src/main/java/org/openhab/binding/tr064/internal/phonebook/Tr064PhonebookImpl.java
lsiepel: The class is `@NonNullByDefault` so atleast in this method it can't be null. 

It comes from => contact.getPerson().getRealName() (line 61) that calls PersonType.getRealName() that gets it from XML node realName that has attribute requireed = true

The compiler also warns: Redundant null check: comparing '@NonNull String' against null

So i guess this really can;t be null ;-) but, its just a null check, if still unsure, let's keep it. It should certainly not block this mayor update to TR064 binding

bundles/org.openhab.binding.tr064/src/main/resources/channels.xml
J-N-K: Yes. This channel was added to the other version before it was added here. That's why I renamed it.

bundles/org.openhab.binding.tr064/src/main/java/org/openhab/binding/tr064/internal/phonebook/Tr064PhonebookImpl.java
J-N-K: @NonNullByDefault is a compile-time check. That’s why the compiler complains: IF the annotation is correct, THEN the null-check is not needed.

OTOH, there is nothing preventing a null value coming here at runtime. IIRC the I have written the XSD for that by hand, so I would not bet on it being correct in all aspects. Maybe we can just try if it is possible to add a phonebook contact in the FritzBox with a blank name and how that looks like in the XML.

bundles/org.openhab.binding.tr064/src/main/resources/channels.xml
jlaur: OK, please remember to mention this in https://github.com/openhab/openhab-distro/blob/main/distributions/openhab/src/main/resources/bin/update.lst.

Should the description here be updated to be in line with the description in the README? In this PR this has changed from "IP of the device with the given MAC" to "IP of the MAC (uses same parameter as `macOnline`)", probably for a reason?

bundles/org.openhab.binding.tr064/src/main/java/org/openhab/binding/tr064/internal/FritzboxActions.java
```diff
@@ -76,16 +95,67 @@ public class PhonebookActions implements ThingActions {
...
                 Collection<Phonebook> phonebooks = handler.getPhonebooks();
-                return phonebooks.stream().map(p -> p.lookupNumber(phonenumber, matchCountInt))
-                        .filter(Optional::isPresent).map(Optional::get).findAny().orElse(phonenumber);
+                return Objects.requireNonNull(phonebooks.stream().map(p -> p.lookupNumber(phonenumber, matchCountInt))
+                        .filter(Optional::isPresent).map(Optional::get).findAny().orElse(phonenumber));
             }
         }
     }
 
+    @RuleAction(label = "create configuration backup", description = "Creates a configuration backup")
```
jlaur, line 108: @J-N-K, @lolodomo - what is your opinion about I18N action labels/descriptions? Since they are not supported, in my opinion we should have no such requirement, but we could accept I18N since it seems not to be shown anywhere in the UI, and to be future-ready. WDYT?

See https://github.com/openhab/openhab-addons/pull/14134#issuecomment-1374644503

Just looking for establishing some guidelines for the time being in this area.

bundles/org.openhab.binding.tr064/src/main/resources/channels.xml
J-N-K: Yes, Thera have been questions how to configure that channel, that's why I added a clarification.

I wonder if we should also rename the other channels that act on the same list of parameters (`macSignalStrength1` and similar) for consistency.

bundles/org.openhab.binding.tr064/src/main/java/org/openhab/binding/tr064/internal/FritzboxActions.java
J-N-K: Yes, until https://github.com/openhab/openhab-core/issues/1745 is resolved and actions are enabled in UI these annotations are not used.

bundles/org.openhab.binding.tr064/src/main/resources/channels.xml
J-N-K: I guess your regex is more strict, so maybe we should find a solution that uses the more strict checking from your regex and the comments from mine.

bundles/org.openhab.binding.tr064/src/main/java/org/openhab/binding/tr064/internal/FritzboxActions.java
jlaur: Thanks for the reference and by that confirming. Question was about I18N texts for label/description though. In #14134 @lolodomo commented the lack of them, so I added them, but the REST API showed that they are not supported yet either. So I think it doesn't matter much if we are adding them now. Since some bindings have them, which means we can't add actions to the UI without also adding I18N support.

J-N-K: Well my understanding is: If the annotations are non-i18n, then their value is used, otherwise the localized version is used. So in any case having a plain text here should never cause issues (except that i18n is not working). I would consider it a bug if it is REQUIRED to have i18 here.

That aside: it does not matter at all until UI support for rules is implemented. Once this is the case, we can re-visit the add-ons that support actions.

jlaur: Agreed. 😄

bundles/org.openhab.binding.tr064/src/main/resources/channels.xml
J-N-K: @lsiepel 

```
((25[0-5]|(2[0-4]|1\d|[1-9]|)\d)(\.(?!([\s#]|$))|([\s#]|$))){4}(\s*#.*)*
```

fulfills

```
        assertTrue(pattern.matcher("192.168.0.1").matches());
        assertTrue(pattern.matcher("192.168.0.1 # foo").matches());
        assertTrue(pattern.matcher("192.168.0.1    # f00 # bar").matches());
        assertTrue(pattern.matcher("225.168.0.1").matches());
        assertTrue(pattern.matcher("1.0.0.1").matches());
        assertTrue(pattern.matcher("94.168.0.1").matches());
        assertFalse(pattern.matcher("322.168.0.1").matches());
        assertFalse(pattern.matcher("192.168.1").matches());
        assertFalse(pattern.matcher("192.168.0.1 foo").matches());
```

WDYT? The only invalid pattern I found that passes the regex is `0.168.1.2` because the single digit is not limited for the first group.

lsiepel: Agree, don't have an immediate fix for that very specific case and it is an improvement anyway. Thanks!

J-N-K: There is a solution, but that would nearly triple the length of the regex (1-254 for first group, a dot, 0-254 followed by dot (2 times), 1-254 for last group, then the comment regex). IMO that's overkill.

lsiepel: Nah, I would opt your solution

J-N-K:
I think everything is addressed now.

lsiepel (approved):
LGTM

jlaur:
> I think everything is addressed now.

Thanks! Only thing I can see missing is I18N properties file regeneration.

jlaur (approved):
Thanks for backporting these improvements.

PR #14473, opened 2023-02-21
[danfossairunit] Add instruction for channel removal

jlaur:
Follow-up to #14030
Made possible by openhab/openhab-core#3330
Requires also openhab/openhab-core#3391

### DIFF
--- a/bundles/org.openhab.binding.danfossairunit/src/main/resources/OH-INF/thing/thing-types.xml
+++ b/bundles/org.openhab.binding.danfossairunit/src/main/resources/OH-INF/thing/thing-types.xml
@@ -18,6 +18,7 @@
 		</channel-groups>
 		<properties>
 			<property name="vendor">Danfoss</property>
+			<property name="thingTypeVersion">1</property>
 		</properties>
 		<representation-property>host</representation-property>
 		<config-description>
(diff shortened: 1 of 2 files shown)


jlaur:
I will retest this tomorrow when there is a new snapshot build.

jlaur:
I have now tested with build 3358, and got it working. First attempt failed though. Dropping the new JAR with the upgrade instructions caused the `thingTypeVersion` property to be set to 1, but didn't execute the upgrade instructions. Restarting openHAB may have triggered the upgrade, but now version was already 1, so nothing happened (removed channel still there). Finally I went back to old version and recreated thing. Then stopped openHAB, dropped new JAR, cleared cache and started openHAB again. This time the channel was removed.

@J-N-K - can you confirm if this is the intended behavior and/or if my test is invalid?

jlaur:
@J-N-K - I don't know if you saw my question, but let me rephrase: Can you confirm that the upgrade instructions are only executed when openHAB is started? Otherwise I don't understand my test results. This would probably mean that replacing JAR's between releases, or upgrading bindings from Community Marketplace, will prevent the upgrade from being triggered.

J-N-K:
They are checked when the thing is initialized, so: no, this does not depend on openHAB startup. It may be that it fails if you replace a bundle with `bundle:update` because the thing is already initialized when this happens. Uninstalling and dropping a .jar in the addons folder should work, also deleting and later (i.e. after uninstall is done) adding it again will also work. Overwriting may result in confusion, just like using `bundle:update`.

J-N-K (approved):
LGTM, if you have a scenario that triggers the faulty behavior I can try to see if there is a bug in core.

lolodomo (approved):
LGTM

jlaur:
@J-N-K - I did some additional testing, both replacing and deleting/creating JAR as well as with and without removed channel being linked to an item. I cannot reproduce the mentioned problem in a reliable way, so it seems to be working.

PR #14474, opened 2023-02-22
[unifi] Use createHttpClient

lolodomo:

### DIFF
--- a/bundles/org.openhab.binding.unifi/src/main/java/org/openhab/binding/unifi/internal/UniFiThingHandlerFactory.java
+++ b/bundles/org.openhab.binding.unifi/src/main/java/org/openhab/binding/unifi/internal/UniFiThingHandlerFactory.java
@@ -50,9 +50,7 @@ public class UniFiThingHandlerFactory extends BaseThingHandlerFactory {
 
     @Activate
     public UniFiThingHandlerFactory(@Reference final HttpClientFactory httpClientFactory) {
-        // [wip] mgb: disabled due to missing common name attributes with certs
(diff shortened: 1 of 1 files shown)


jlaur (approved):
Thanks!

bundles/org.openhab.binding.unifi/src/main/java/org/openhab/binding/unifi/internal/UniFiThingHandlerFactory.java
```diff
@@ -50,9 +50,7 @@ public class UniFiThingHandlerFactory extends BaseThingHandlerFactory {
 
     @Activate
     public UniFiThingHandlerFactory(@Reference final HttpClientFactory httpClientFactory) {
-        // [wip] mgb: disabled due to missing common name attributes with certs
-        // this.httpClient = httpClientFactory.getCommonHttpClient();
-        httpClient = new HttpClient(new SslContextFactory.Client(true));
+        httpClient = httpClientFactory.createHttpClient(BINDING_ID, new SslContextFactory.Client(true));
```
jlaur, line 53: I'm wondering if we in core even could have a common client with this often used configuration?

lolodomo: At the same time, creating such client (trust all) should be probably avoided. So providing a method for that could be a not so good idea.

jlaur: Good point.

PR #14479, opened 2023-02-22
[dlinksmarthome] Reboot device daily following shutdown of cloud service

MikeJMajor:
Following the shutdown of the mydlink Home service, the devices have a tendency to become unresponsive after a few days.

This change performs a daily reboot of the device to keep it responsive.

Discovery has also been changed to avoid using ipv6 addresses which don't seem to work. The change is based on  a change made to the Chromecast binding. The binding previously had a problem in that it would occasionally switch to using an ipv6 address and go offline, rebooting daily increases the chance it ends up with an ipv6 address.

### DIFF
--- a/bundles/org.openhab.binding.dlinksmarthome/README.md
+++ b/bundles/org.openhab.binding.dlinksmarthome/README.md
@@ -6,7 +6,8 @@ A binding for D-Link Smart Home devices.
 
 ### DCH-S150 (WiFi motion sensor)
 
-The binding has been tested with hardware revisions A1 and A2 running firmware version 1.22. The mydlink Home service is now end of life and the device requires a daily reboot (performed by the binding) to keep it responsive.
+The binding has been tested with hardware revisions A1 and A2 running firmware version 1.22. 
+The mydlink Home service is now end of life and the device requires a daily reboot (performed by the binding) to keep it responsive.
 
 ## Discovery
 
@@ -25,7 +26,7 @@ Once added the configuration must be updated to specify the PIN code located on 
 
 - **ipAddress** - Hostname or IP of the device
 - **pin** - PIN code from the back of the device
-- **rebootHour** - Hour (24h) of the day that the device will be rebooted to ensure that it remains responsive. Default is 3.
+- **rebootHour** - Hour (24h) of the day that the device will be rebooted to ensure that it remains responsive (default is 3).
 
 To manually configure a DCH-S150 Thing you must specify its IP address and PIN code.
 
--- a/bundles/org.openhab.binding.dlinksmarthome/README.md
+++ b/bundles/org.openhab.binding.dlinksmarthome/README.md
@@ -6,11 +6,11 @@ A binding for D-Link Smart Home devices.
 
 ### DCH-S150 (WiFi motion sensor)
 
-The binding has been tested with hardware revisions A1 and A2 running firmware version 1.22.
+The binding has been tested with hardware revisions A1 and A2 running firmware version 1.22. The mydlink Home service is now end of life and the device requires a daily reboot (performed by the binding) to keep it responsive.
 
 ## Discovery
 
-The binding can automatically discover devices that have already been added to the Wifi network. Please refer to your mydlink Home app for instructions on how to add your device to your Wifi network.
+The binding can automatically discover devices that have already been added to the Wifi network.
 
 ## Binding Configuration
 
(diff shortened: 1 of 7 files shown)


bundles/org.openhab.binding.dlinksmarthome/README.md
```diff
@@ -6,11 +6,11 @@ A binding for D-Link Smart Home devices.
 
 ### DCH-S150 (WiFi motion sensor)
 
-The binding has been tested with hardware revisions A1 and A2 running firmware version 1.22.
+The binding has been tested with hardware revisions A1 and A2 running firmware version 1.22. The mydlink Home service is now end of life and the device requires a daily reboot (performed by the binding) to keep it responsive.
```
lolodomo, line 9: Minor: please go to a new line for each sentence.

bundles/org.openhab.binding.dlinksmarthome/src/main/java/org/openhab/binding/dlinksmarthome/internal/DLinkSmartHomeDiscoveryParticipant.java
```diff
@@ -96,7 +97,11 @@ private ThingTypeUID getThingType(final ServiceInfo serviceInfo) {
 
     private DiscoveryResult createMotionSensor(final ThingUID thingUID, final ThingTypeUID thingType,
             final ServiceInfo serviceInfo) {
-        final String host = serviceInfo.getHostAddresses()[0];
+        final Inet4Address[] addresses = serviceInfo.getInet4Addresses();
+        if (addresses.length != 1) {
+            return null;
+        }
+        final String host = addresses[0].getHostAddress();
```
lolodomo, line 104: This is a change in the discovery that you even don't announce ?
I think this should be mentioned in the description of your PR.

bundles/org.openhab.binding.dlinksmarthome/src/main/java/org/openhab/binding/dlinksmarthome/internal/handler/DLinkMotionSensorHandler.java
```diff
@@ -64,6 +64,9 @@ public void sensorStatus(final DeviceStatus status) {
             case ONLINE:
                 updateStatus(ThingStatus.ONLINE);
                 break;
+            case REBOOTING:
+                updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.DUTY_CYCLE, "Device rebooting");
```
lolodomo, line 68: ThingStatusDetail.DUTY_CYCLE is I believe not so much used but could be the right choice there.
Only used in 3 bindings: homematic, lutron and MQTT generic. I will see how it is used bu others.

bundles/org.openhab.binding.dlinksmarthome/src/main/java/org/openhab/binding/dlinksmarthome/internal/motionsensor/DLinkMotionSensorConfig.java
```diff
@@ -34,4 +33,9 @@ public class DLinkMotionSensorConfig {
      * The pin code of the device
      */
     public String pin;
+
+    /**
+     * The hour to reboot the device
+     */
+    public Integer rebootHour;
```
lolodomo, line 40: You can simply use `int` type.

lolodomo (changes requested):
Partial review

bundles/org.openhab.binding.dlinksmarthome/src/main/java/org/openhab/binding/dlinksmarthome/internal/DLinkSmartHomeDiscoveryParticipant.java
MikeJMajor: Description updated to include why this change was made.

bundles/org.openhab.binding.dlinksmarthome/src/main/java/org/openhab/binding/dlinksmarthome/internal/handler/DLinkMotionSensorHandler.java
MikeJMajor: Are you happy with the use of DUTY_CYCLE?

lolodomo (approved):
LGTM, thank you

bundles/org.openhab.binding.dlinksmarthome/src/main/java/org/openhab/binding/dlinksmarthome/internal/handler/DLinkMotionSensorHandler.java
lolodomo: Let's assume it is OK.

PR #14482, opened 2023-02-22
[neohub] Use createWebSocketClient

lolodomo:

### DIFF
--- a/bundles/org.openhab.binding.neohub/src/test/java/org/openhab/binding/neohub/test/NeoHubProtocolTests.java
+++ b/bundles/org.openhab.binding.neohub/src/test/java/org/openhab/binding/neohub/test/NeoHubProtocolTests.java
@@ -13,16 +13,23 @@
 package org.openhab.binding.neohub.test;
 
 import static org.junit.jupiter.api.Assertions.*;
+import static org.mockito.ArgumentMatchers.*;
+import static org.mockito.Mockito.*;
 
 import java.io.IOException;
 
 import org.eclipse.jdt.annotation.NonNullByDefault;
+import org.eclipse.jetty.client.HttpClient;
+import org.eclipse.jetty.util.ssl.SslContextFactory;
+import org.eclipse.jetty.websocket.client.WebSocketClient;
 import org.junit.jupiter.api.Test;
 import org.openhab.binding.neohub.internal.NeoHubBindingConstants;
 import org.openhab.binding.neohub.internal.NeoHubConfiguration;
 import org.openhab.binding.neohub.internal.NeoHubException;
 import org.openhab.binding.neohub.internal.NeoHubSocket;
 import org.openhab.binding.neohub.internal.NeoHubWebSocket;
+import org.openhab.core.io.net.http.WebSocketFactory;
+import org.openhab.core.thing.ThingUID;
 
 /**
  * JUnit for testing WSS and TCP socket protocols.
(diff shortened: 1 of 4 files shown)


bundles/org.openhab.binding.neohub/src/test/java/org/openhab/binding/neohub/test/NeoHubProtocolTests.java
```diff
@@ -70,7 +77,15 @@ void testWssConnection() throws NeoHubException, IOException {
             config.socketTimeout = SOCKET_TIMEOUT;
             config.apiToken = HUB_API_TOKEN;
 
-            NeoHubWebSocket socket = new NeoHubWebSocket(config, "test");
+            SslContextFactory.Client sslContextFactory = new SslContextFactory.Client();
```
andrewfg, line 80: There are three problems here:
1. When RUN_WSS_TEST is true and HUB_IP_ADDRESS is a valid value, the JUnit test does not exit at line 70, but actually proceeds to line 85; however the test then freezes at that line.
2. Not only should the Junit test not freeze, but if the value of HUB_API_TOKEN is valid , then the test should make an actual live online test to an existing online NeoHub; however with a mocked web socket factory, this will probably fail.
3. The code probably also needs `assertNonNull(webSocketFactory)` to eliminate a compiler warning.

```
    private static final String HUB_IP_ADDRESS = "192.168.1.xxx"; <= VALID
    private static final String HUB_API_TOKEN = "12345678-1234-1234-1234-123456789ABC"; <= VALID
    private static final boolean RUN_WSS_TEST = false; <= TRUE
```

andrewfg (changes requested):
@lolodomo unfortunately your code breaks the JUnit tests.

bundles/org.openhab.binding.neohub/src/test/java/org/openhab/binding/neohub/test/NeoHubProtocolTests.java
lolodomo: I agree, these unit tests have no sense. They are even doing nothing with the current code because a constant is not set to true. I have the strange feeling they were implemented by the author for testing in real conditions and then disabled by setting constants.
I don't know what to do with that.
I do not plan to spend more time on that PR, so I can close it if this is easier.

andrewfg: > I have the strange feeling they were implemented by the author for testing in real conditions and then disabled by setting constants.

Correct.

> I do not plan to spend more time on that PR, so I can close it if this is easier.

Your choice. The ideal solution would be to fix the JUnit code so that it can actually do the tests. That would mean creating a real webSocketFactory instead of using `mock`..

lolodomo:
I forgive due to unit tests.

andrewfg:
> I forgive due to unit tests.

@lolodomo 'mea culpa' but you were too fast in deleting this PR; it is my mistake, but when I tested the code, I was doing it within the Eclipse debugger, and that was suffering a delay of over 30 seconds in initialising the mock; which I had wrongly interpreted as a complete application freeze. But in fact, if I run your code outside the debugger, (or wait patiently for the debugger task to complete), then your mocked code DOES actually work as it should do! So in fact, I CAN approve your PR, as it was, after all. Again many apologies for my mistake.

andrewfg (commented):
LGTM

lolodomo:
PR reopened

andrewfg (approved):
LGTM

jlaur (approved):
Thanks!

PR #14484, opened 2023-02-23
[km200] Added missing config descriptions

Markinus:
Added missing config descriptions. In the past it was not generating any errors but now the users will get a lot of errors without this empty config description.
Community: [137906](https://community.openhab.org/t/km200-warning-at-startup/137906)


lolodomo:
Please do not submit PR against 3.4.x branch but rather main branch.

PR #14252, opened 2023-01-21
[netatmo] Extend webhook support for doorbell and presence camera

Novanic:
This pull-request fixes issue #14237, see issue #14237 for more information.

PR #14485, opened 2023-02-23
[netatmo] Add Siren capability to Presence Outdoor Camera

### LINKED ISSUE #14466: [Netatmo] Supporting siren on Netatmo Presence Outdoor Camera with Siren
I would like to suggest the Netatmo binding enhanced for the Netatmo Presence Outdoor camera (with siren) - so that the siren gets an own on/off channel.

According to the api (dev.netatmo.com) the siren can be turned on/off as easy as you can turn the floodlight on/off.

I have tried this - and can verify that it works , as I tested with this command (security values x'd out);
`curl -X POST "https://api.netatmo.com/api/setstate" -H "accept: application/json" -H "Authorization: Bearer xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx" -H  "Content-Type: application/json" -d "{\"home\":{\"id\":\"xxxxxxxxxxxxxxxxxxxxxxxxxxxxx\",\"modules\":[{\"id\":\"xx:xx:xx:xx:xx:xx\",\"siren_status\":\"sound\"}]}}"`

I can confirm that the home id (x'd out above" as well as the module id (x'd out mac address above) is exactly the same as I use in the api to turn the floodlight on/off/auto.

To turn the siren off again, the siren_status have to receive a "no_sound".

I will be very happy to test when implemented - as I have the device at hand.

clinique:
Resolves #14466

### DIFF
--- a/bundles/org.openhab.binding.netatmo/src/main/java/org/openhab/binding/netatmo/internal/handler/capability/Capability.java
+++ b/bundles/org.openhab.binding.netatmo/src/main/java/org/openhab/binding/netatmo/internal/handler/capability/Capability.java
@@ -12,7 +12,7 @@
  */
 package org.openhab.binding.netatmo.internal.handler.capability;
 
-import static org.openhab.binding.netatmo.internal.NetatmoBindingConstants.VENDOR;
+import static org.openhab.binding.netatmo.internal.NetatmoBindingConstants.*;
 import static org.openhab.core.thing.Thing.*;
 
 import java.util.Collection;
@@ -97,10 +97,13 @@ public class Capability {
     protected void beforeNewData() {
         properties = new HashMap<>(thing.getProperties());
         firstLaunch = properties.isEmpty();
-        if (firstLaunch && !moduleType.isLogical()) {
-            String name = moduleType.apiName.isBlank() ? moduleType.name() : moduleType.apiName;
-            properties.put(PROPERTY_MODEL_ID, name);
-            properties.put(PROPERTY_VENDOR, VENDOR);
+        if (firstLaunch) {
+            properties.put(PROPERTY_THING_TYPE_VERSION, Integer.toString(moduleType.thingTypeVersion));
+            if (!moduleType.isLogical()) {
+                String name = moduleType.apiName.isBlank() ? moduleType.name() : moduleType.apiName;
+                properties.put(PROPERTY_MODEL_ID, name);
+                properties.put(PROPERTY_VENDOR, VENDOR);
+            }
         }
         statusReason = null;
     }
--- a/bundles/org.openhab.binding.netatmo/src/main/resources/OH-INF/i18n/netatmo.properties
+++ b/bundles/org.openhab.binding.netatmo/src/main/resources/OH-INF/i18n/netatmo.properties
@@ -309,10 +309,10 @@ channel-type.netatmo.setpoint.label = Setpoint
 channel-type.netatmo.setpoint.description = Thermostat temperature setpoint.
 channel-type.netatmo.siren-monitoring.label = Monitoring
 channel-type.netatmo.siren-monitoring.description = Monitoring state of the equipment
-channel-type.netatmo.siren-status.label = Siren
-channel-type.netatmo.siren-status.description = Status of the siren
 channel-type.netatmo.siren-status-rw.label = Siren
 channel-type.netatmo.siren-status-rw.description = Status of the siren
+channel-type.netatmo.siren-status.label = Siren
+channel-type.netatmo.siren-status.description = Status of the siren
 channel-type.netatmo.tag-status.label = Door Status
 channel-type.netatmo.th-mode.label = Thermostat Mode
 channel-type.netatmo.th-mode.description = Chosen thermostat mode (home, frost guard, manual, max).
@@ -350,13 +350,6 @@ channel-type.config.netatmo.live-stream-url.quality.description = Defines qualit
 channel-type.config.netatmo.live-stream-url.quality.option.low = Low Quality
 channel-type.config.netatmo.live-stream-url.quality.option.poor = Poor Quality
 channel-type.config.netatmo.live-stream-url.quality.option.high = High Quality
-
-# channel types
-
-channel-type.netatmo.siren-status.label = Status
-channel-type.netatmo.siren-status.description = Status of the siren
-channel-type.netatmo.siren-status.state.option.no_sound = Silent
-channel-type.netatmo.siren-status.state.option.sound = Alarm
 
 # channel types config
 
--- a/bundles/org.openhab.binding.netatmo/src/main/resources/OH-INF/i18n/netatmo.properties
+++ b/bundles/org.openhab.binding.netatmo/src/main/resources/OH-INF/i18n/netatmo.properties
@@ -309,10 +309,10 @@ channel-type.netatmo.setpoint.label = Setpoint
 channel-type.netatmo.setpoint.description = Thermostat temperature setpoint.
 channel-type.netatmo.siren-monitoring.label = Monitoring
 channel-type.netatmo.siren-monitoring.description = Monitoring state of the equipment
-channel-type.netatmo.siren-status.label = Status
+channel-type.netatmo.siren-status.label = Siren
 channel-type.netatmo.siren-status.description = Status of the siren
-channel-type.netatmo.siren-status.state.option.no_sound = Silent
-channel-type.netatmo.siren-status.state.option.sound = Alarm
+channel-type.netatmo.siren-status-rw.label = Siren
+channel-type.netatmo.siren-status-rw.description = Status of the siren
 channel-type.netatmo.tag-status.label = Door Status
 channel-type.netatmo.th-mode.label = Thermostat Mode
 channel-type.netatmo.th-mode.description = Chosen thermostat mode (home, frost guard, manual, max).
@@ -350,6 +350,13 @@ channel-type.config.netatmo.live-stream-url.quality.description = Defines qualit
 channel-type.config.netatmo.live-stream-url.quality.option.low = Low Quality
 channel-type.config.netatmo.live-stream-url.quality.option.poor = Poor Quality
 channel-type.config.netatmo.live-stream-url.quality.option.high = High Quality
+
+# channel types
+
+channel-type.netatmo.siren-status.label = Status
+channel-type.netatmo.siren-status.description = Status of the siren
+channel-type.netatmo.siren-status.state.option.no_sound = Silent
+channel-type.netatmo.siren-status.state.option.sound = Alarm
 
 # channel types config
 
--- a/bundles/org.openhab.binding.netatmo/src/main/resources/OH-INF/instructions.xml
+++ b/bundles/org.openhab.binding.netatmo/src/main/resources/OH-INF/instructions.xml
@@ -1,0 +1,16 @@
+<?xml version="1.0" encoding="UTF-8" standalone="yes" ?>
+<update:update-descriptions xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+	xmlns:update="https://openhab.org/schemas/update-description/v1.0.0"
+	xsi:schemaLocation="https://openhab.org/schemas/update-description/v1.0.0 https://openhab.org/schemas/update-description-1.0.0.xsd">
+
+	<thing-type uid="netatmo:presence">
+
+		<instruction-set targetVersion="1">
+			<add-channel id="siren">
+				<type>netatmo:siren-status-rw</type>
+			</add-channel>
+		</instruction-set>
+
+	</thing-type>
+
+</update:update-descriptions>
--- a/bundles/org.openhab.binding.netatmo/src/main/resources/OH-INF/thing/channels.xml
+++ b/bundles/org.openhab.binding.netatmo/src/main/resources/OH-INF/thing/channels.xml
@@ -19,7 +19,7 @@
 
 	<channel-type id="siren-status-ro">
 		<item-type>Switch</item-type>
-		<label>Status</label>
+		<label>Siren</label>
 		<description>Status of the siren</description>
 		<category>Soundvolume</category>
 		<state readOnly="true"/>
@@ -27,7 +27,7 @@
 
 	<channel-type id="siren-status-rw">
 		<item-type>Switch</item-type>
-		<label>Status</label>
+		<label>Siren</label>
 		<description>Status of the siren</description>
 		<category>Soundvolume</category>
 	</channel-type>
--- a/bundles/org.openhab.binding.netatmo/src/main/resources/OH-INF/thing/channels.xml
+++ b/bundles/org.openhab.binding.netatmo/src/main/resources/OH-INF/thing/channels.xml
@@ -17,16 +17,19 @@
 		<state readOnly="true"/>
 	</channel-type>
 
-	<channel-type id="siren-status">
-		<item-type>String</item-type>
+	<channel-type id="siren-status-ro">
+		<item-type>Switch</item-type>
 		<label>Status</label>
 		<description>Status of the siren</description>
-		<state pattern="%s" readOnly="true">
-			<options>
-				<option value="no_sound">Silent</option>
-				<option value="sound">Alarm</option>
-			</options>
-		</state>
+		<category>Soundvolume</category>
+		<state readOnly="true"/>
+	</channel-type>
+
+	<channel-type id="siren-status-rw">
+		<item-type>Switch</item-type>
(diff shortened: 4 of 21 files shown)


lolodomo:
Shouldn't we wait for the merge of #14252 first to avoid conflicts ?

clinique:
Sure, there no urgency.

bundles/org.openhab.binding.netatmo/src/main/resources/OH-INF/thing/channels.xml
```diff
@@ -17,16 +17,19 @@
 		<state readOnly="true"/>
 	</channel-type>
 
-	<channel-type id="siren-status">
-		<item-type>String</item-type>
-		<label>Status</label>
+	<channel-type id="siren-status-ro">
```
lolodomo, line 20: I expect an adjustment in the default translation file (updated key).

bundles/org.openhab.binding.netatmo/src/main/resources/OH-INF/thing/channels.xml
```diff
@@ -17,16 +17,19 @@
...
-			<options>
-				<option value="no_sound">Silent</option>
-				<option value="sound">Alarm</option>
-			</options>
-		</state>
+		<category>Soundvolume</category>
+		<state readOnly="true"/>
+	</channel-type>
+
+	<channel-type id="siren-status-rw">
```
lolodomo, line 28: I expect an adjustment in the default translation file (new entries).

bundles/org.openhab.binding.netatmo/src/main/resources/OH-INF/thing/security.xml
```diff
@@ -197,6 +197,7 @@
 	<channel-group-type id="presence">
 		<label>Presence Camera</label>
 		<channels>
+			<channel id="siren" typeId="siren-status-rw"/>
```
lolodomo, line 200: Can you use the new thing upgrade mechanism to automatically add this new channel in the existing things ?

lolodomo (changes requested):

bundles/org.openhab.binding.netatmo/src/main/resources/OH-INF/thing/security.xml
```diff
@@ -26,7 +26,7 @@
 	<channel-group-type id="siren">
 		<label>Siren Status</label>
 		<channels>
-			<channel id="status" typeId="siren-status"/>
+			<channel id="status" typeId="siren-status-ro"/>
```
lolodomo, line 29: Maybe you should avoid renaming the channel type, I am not sure what could be the impact on existing channels (when thing was created through UI).
@jlaur is more expert than me on that subject I believe. WDYT ?

jlaur: I believe you need to create update instructions with `update-channel` according to https://www.openhab.org/schemas/update-description-1.0.0.xsd. Perhaps @J-N-K can confirm.

bundles/org.openhab.binding.netatmo/src/main/resources/OH-INF/thing/channels.xml
clinique: properties files regenerated

bundles/org.openhab.binding.netatmo/src/main/resources/OH-INF/thing/channels.xml
clinique: properties file regenerated

bundles/org.openhab.binding.netatmo/src/main/resources/OH-INF/thing/security.xml
clinique: Ok, I reverted to siren-status the read-only version, only adding siren-status-rw

clinique: This will be another beast because I do not use xml defined things in this binding.

bundles/org.openhab.binding.netatmo/src/main/resources/OH-INF/thing/security.xml
clinique: This will be another beast because I do not use xml defined things in this binding.

bundles/org.openhab.binding.netatmo/src/main/resources/OH-INF/i18n/netatmo.properties
```diff
@@ -344,6 +344,13 @@ channel-type.config.netatmo.live-stream-url.quality.option.low = Low Quality
 channel-type.config.netatmo.live-stream-url.quality.option.poor = Poor Quality
 channel-type.config.netatmo.live-stream-url.quality.option.high = High Quality
 
+# channel types
+
+channel-type.netatmo.siren-status.label = Status
+channel-type.netatmo.siren-status.description = Status of the siren
+channel-type.netatmo.siren-status.state.option.no_sound = Silent
+channel-type.netatmo.siren-status.state.option.sound = Alarm
```
lolodomo, line 352: This block should be removed ?
You already have label and description added below and the two options are now obsolete.

lolodomo (changes requested):

bundles/org.openhab.binding.netatmo/src/main/java/org/openhab/binding/netatmo/internal/handler/capability/Capability.java
```diff
@@ -97,10 +97,13 @@ public class Capability {
     protected void beforeNewData() {
         properties = new HashMap<>(thing.getProperties());
         firstLaunch = properties.isEmpty();
-        if (firstLaunch && !moduleType.isLogical()) {
-            String name = moduleType.apiName.isBlank() ? moduleType.name() : moduleType.apiName;
-            properties.put(PROPERTY_MODEL_ID, name);
-            properties.put(PROPERTY_VENDOR, VENDOR);
+        if (firstLaunch) {
+            properties.put(PROPERTY_THING_TYPE_VERSION, Integer.toString(moduleType.thingTypeVersion));
```
lolodomo, line 101: If your thing was previously created with UI, the model and vendor properties are already set. That means the thingTypeVersion property will never be set ?

lolodomo (changes requested):

bundles/org.openhab.binding.netatmo/src/main/java/org/openhab/binding/netatmo/internal/handler/capability/Capability.java
J-N-K: Why is `thingTypeVersion` set at all by the binding? It is set on creation of the thing (either each time when using file-based or once when using UI) or set/updated by the `ThingManager`.

lolodomo: @J-N-K : this binding is very particular with no XML files to describe thing types. Everything is built dynamically. So `thingTypeVersion` has to be set by the code ?
Or maybe you have not understood at all how is handled this property and how and when to set it !

J-N-K: Understood. But if the binding handles all this dynamically, it also adds/removes channels dynamically, correct? Why would it need the `ThingManager` to update managed things when it does that all on its own (which is fine, just a little bit complex).

The update code in the `ThingManager` is for modifying the channels of a things whose thing type is XML defined. The changed thing type definition in the XML is only applied to textual things (because they are essentially re-created when the file is loaded) and to new managed things, but not to "old" managed things. The update instructions allow the `ThingManager` to apply the needed steps to bring the "old" managed things in line with the new definition (so that they are essentially like a "new" thing). The `thingTypeVersion` is used to determine which steps need to be taken and to prevent applying the same instruction twice (because there is no need to add channel `foo` if this has already been done before).

So as long as no update instructions are involved, there is no need to set the `thingTypeVersion`. When there is no XML that defines the thing type, I don't see how update instructions would be needed. Maybe I'm missing something here.

clinique: > If your thing was previously created with UI, the model and vendor properties are already set. That means the thingTypeVersion property will never be set ?

No, the discovery process does not add any property. This is only done at first start (identified by the absence of any property) of the handler.

clinique: > @J-N-K : this binding is very particular with no XML files to describe thing types. Everything is built dynamically. So `thingTypeVersion` has to be set by the code ? Or maybe we have not understood at all how is handled this property and how and when to set it !

You're correct, as there is no XML, I had to add this property in module descriptions enum and then add it by binding code.

clinique: > Understood. But if the binding handles all this dynamically, it also adds/removes channels dynamically, correct? Why would it need the `ThingManager` to update managed things when it does that all on its own (which is fine, just a little bit complex).
> 

Mmm, correct, maybe. Have to think of it twice. I would not want to have to redevelop all this magic you added to code in the binding to support this feature.

clinique: > If your thing was previously created with UI, the model and vendor properties are already set. That means the thingTypeVersion property will never be set ?

Correct. And if so, no other mean that asking users to delete and recreate their UI defined things.

J-N-K: I didn't check your code, but I have also bindings that have a generic thing type (which only holds the thing configuration) and all channels are added/removed dynamically.

In these bindings I have a `checkChannels` method that runs during `initialize` and checks if all expected channels are present (and adds them when missing) or removes those channels that should not be there. After that the thing is `updateThing`ed (if needed). Since thing initialization is only done once, efficiency is not that important here and the check can be as complex as needed.

clinique: The only difference with you is that I currently only do it once - the first time the handler is initialized (denoted if the property list is empty).

J-N-K: IMO it doesn't hurt to do it every time and it removes the need to keeping track of different thing type versions. You just need to make sure to only update the thing if it was modified, otherwise you could end up in a loop.

lolodomo:
@clinique : you have to fix a conflict on README

lolodomo:
Still a conflict to solve.

What is the status about tests ? You are still waiting for the feedback from a user ?

clinique:
> Still a conflict to solve.
> 
> What is the status about tests ? You are still waiting for the feedback from a user ?

Yes, my Presence does not have any siren - so I would like a clear "it works"

bundles/org.openhab.binding.netatmo/src/main/resources/OH-INF/instructions.xml
```diff
@@ -0,0 +1,16 @@
+<?xml version="1.0" encoding="UTF-8" standalone="yes" ?>
```
lolodomo, line 1: Please move this file in a folder named "update"

bundles/org.openhab.binding.netatmo/src/main/resources/OH-INF/thing/channels.xml
```diff
@@ -18,15 +18,18 @@
 	</channel-type>
 
 	<channel-type id="siren-status">
-		<item-type>String</item-type>
-		<label>Status</label>
+		<item-type>Switch</item-type>
```
lolodomo, line 21: You need update instructions for the channel using this channel type. You will then instruct to update the item type.

The channel for the thing will then be OK after update but note that if user has linked this channel to a Switch item, he will have to unlink it manually and link it again to a String item.

lolodomo (changes requested):

bundles/org.openhab.binding.netatmo/src/main/java/org/openhab/binding/netatmo/internal/handler/capability/Capability.java
lolodomo: What is the conclusion of this discussion. Is the code OK now ?

bundles/org.openhab.binding.netatmo/src/main/resources/OH-INF/instructions.xml
clinique: Done

bundles/org.openhab.binding.netatmo/src/main/java/org/openhab/binding/netatmo/internal/handler/capability/Capability.java
clinique: Code did not change. @J-N-K prefers doing it at each initialization, I do it only at first initialization. It works, is it worth changing ?

bundles/org.openhab.binding.netatmo/src/main/resources/OH-INF/update/instructions.xml
```diff
@@ -0,0 +1,19 @@
...
+	xsi:schemaLocation="https://openhab.org/schemas/update-description/v1.0.0 https://openhab.org/schemas/update-description-1.0.0.xsd">
+
+	<thing-type uid="netatmo:presence">
+
+		<instruction-set targetVersion="1">
+			<add-channel id="siren">
+				<type>netatmo:siren-status-rw</type>
+			</add-channel>
+			<update-channel id="status" groupIds="siren">
+				<type>netatmo:siren-status</type>
```
lolodomo, line 13: I was rather expecting a tag `itemType` and not `type`.

lolodomo (changes requested):

bundles/org.openhab.binding.netatmo/src/main/java/org/openhab/binding/netatmo/internal/handler/capability/Capability.java
lolodomo: Ok

bundles/org.openhab.binding.netatmo/src/main/resources/OH-INF/update/instructions.xml
```diff
@@ -0,0 +1,19 @@
+<?xml version="1.0" encoding="UTF-8" standalone="yes" ?>
+<update:update-descriptions xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+	xmlns:update="https://openhab.org/schemas/update-description/v1.0.0"
+	xsi:schemaLocation="https://openhab.org/schemas/update-description/v1.0.0 https://openhab.org/schemas/update-description-1.0.0.xsd">
+
+	<thing-type uid="netatmo:presence">
+
+		<instruction-set targetVersion="1">
+			<add-channel id="siren">
```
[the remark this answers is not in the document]
clinique: You're correct. Done.

bundles/org.openhab.binding.netatmo/src/main/resources/OH-INF/update/instructions.xml
clinique: The is no `itemType` tag, only `type`. I looked at some examples in the `openhab-addons` repo, it was done this way also.

lolodomo: You're right, I just checked https://www.openhab.org/schemas/update-description-1.0.0.xsd and only type, label, description and tags are allowed.
By updating the type, the itemType will be updated when PR openhab/openhab-core#3576 will be merged (I tested that with your meteoalerte PR).

lolodomo:
In addition to a successful test of this change by a user, we are also waiting for the merge of PR openhab/openhab-core#3576.

lolodomo:
Core PR has just been merged.

jlaur:
@lolodomo, @clinique - anything left to do here? Do you have any user to test the changes?

clinique:
Let's assume it works...

jlaur:
@lolodomo - did you see @clinique's latest comment?

lolodomo:
No but now I see it :)

PR #14487, opened 2023-02-23
[km200] Added missing config descriptions

Markinus:
Added missing config descriptions. In the past it was not generating any errors but now the users will get a lot of warnings without this empty config description.
Community: [137906](https://community.openhab.org/t/km200-warning-at-startup/137906)
(it was #14484, not rebased to main)


lolodomo:
I do not find these config descriptions used in static XML definitions? 
Where are they used? 
Maybe it would be better to fix the root cause?

Markinus:
The channels with this URIs are created dynamicaly depending on the devices configuration. That's the reason why there is not a static link in the config files. Some changes in openhab itself needs this binding extension. I have no idea why.

Markinus:
Ohh, I think I found something what could be wrong. Wait, I will check something else.

Markinus:
This is not the fix for the root cause. I will close this and prepare a new PR.

PR #14488, opened 2023-02-23
[km200] removed unused configuration URIs

Markinus:
This PR removes unused configuration URI/description. This was prepared but never used. See users comments: [137906](https://community.openhab.org/t/km200-warning-at-startup/137906)


lolodomo (approved):
LGTM, thank you

PR #14489, opened 2023-02-23
[jrubyscripting] Update README, and default configuration installing gem

ccutrer:
README is generated by YARD from USAGE.md in the gem.


ccutrer:
This is ready for review. The 5.0.0 gem isn't published yet, but it will be in the next 48 hours, and there will be no more changes to its USAGE.md.

ccutrer:
5.0.0 of the gem as been released

jimtng:
@openhab/add-ons-maintainers this should also be included into M1 along with #14556

bundles/org.openhab.automation.jrubyscripting/src/main/resources/OH-INF/config/config.xml
```diff
@@ -26,14 +26,16 @@
 		<parameter name="gems" type="text" required="false" groupName="gems">
 			<label>Ruby Gems</label>
 			<description><![CDATA[A comma separated list of Ruby Gems to install. Versions may be constrained by separating with an
-			    <tt>=</tt> and then the standard RubyGems version constraint, such as "<tt>openhab-scripting=~>4.0</tt>".
+			    <tt>=</tt> and then the standard RubyGems version constraint, such as "<tt>openhab-scripting=~>5.0.0</tt>".
```
jimtng, line 29: Since descriptions go into the translation files, every time we change this, it'll require re-translations. How can we avoid that?

If we do want it here, please regenerate the i18n file with `mvn i18n:generate-default-translations -pl :org.openhab.automation.jrubyscripting`

ccutrer: I've updated it to the current iteration, but I don't think we need to update it with every bump of the gem version. it is only a "help text". perhaps only major version bumps?

jimtng (approved):
LGTM

bundles/org.openhab.automation.jrubyscripting/src/main/resources/OH-INF/config/config.xml
jimtng: I wasn't suggesting to change it more often. I was thinking of changing that to a generic "gemfile=~>x.x" but having it show openhab-scripting is actually really helpful, so yeah changing it on every major version is OK considering the benefit.

PR #14490, opened 2023-02-24
[upnpcontrol] fix shuffle switch also switching only play one

mherwege:
Fixes switching on shuffle also switching only playing a single song on.


jlaur (approved):
Thanks for the fix!

Issue #14458, opened 2023-02-20
[Netatmo] Floodlight not working on Presence Camera (with siren)


multimartin:
I am on OH3.4 and have an issue with the Netatmo binding related to the floodlight function on my Netatmo Presence Outdoor camera (with siren).

The floodlight on/off/auto simply does not work according to docs: [Netatmo - Bindings | openHAB](https://www.openhab.org/addons/bindings/netatmo/)

My floodlight works fine using either Homekit, the Netatmo app and through the Netatmo api (https://dev.netatmo.com/) - but not through OH. Everything else through OH seem to work fine (camera/snapshots etc).

The docs describe the channel as a switch - but with three states (on, off or auto). Perhaps that is the issue ?

(EDIT: the enhancement request regarding the siren function has been cut out of the text above and put into a seperate issue)

clinique:
Could you split your issue in two ? One is an enhancement, the other a potential bug. Thanks

clinique:
> The floodlight on/off/auto simply does not work according to docs: [Netatmo - Bindings | openHAB](https://www.openhab.org/addons/bindings/netatmo/)
> 
Does the binding report correctly the current status of the flood light if activated/deactivated by movement or Netatmo App ? (my Presence does not work anymore so I'm unable to test this currently).

clinique:
Ok, thanks. I'm going to try to address this issue that may not have been heavily tested or used since the global binding revamp. I'll probably need your help - get ready to switch the binding in trace mode -, because as said my Presence is out of order.

multimartin:
I am ready ;-)

clinique:
Ok. Would you be able to catch the API answer when a request to update camera values is sent ?
Something that would look like this : 
[...]

PR #14492, opened 2023-02-24
[Netatmo] Floodlight issue on Presence Camera

clinique:
Solves issue #14458 

 Removes the usage of "app.netatmo.net" since api now holds these entry points.
I think it was the origin of malfunction of floodlight (to be controlled by somebody who has one).


lolodomo (approved):
LGTM, thank you

lolodomo:
@clinique : you want to wait for a user feedback before merging ?

clinique:
Yes, I'm unable to ensure that what I did works really, so I would prefer have somebody (@multimartin) validate this.

lolodomo:
Confirmed to work in the issue.

PR #14493, opened 2023-02-24
[denonmarantz] Add representation-property to thing-types xml

mhilbush:
The discovery service calls `withrepresentationProperty` but it's not in thing-types.xml.


jlaur (approved):
Thanks!

PR #14494, opened 2023-02-24
[goecharger] Fix command sending

DeJe63:
There was missing "break" preventing commands to be sent correctly.


schogit:
Hello together,

I just can confirm @DeJe63  Bugfix - I was debugging the binding and found the same issue.
Best regards 

schogit

lolodomo:
Change looks good to me.
Can you please sign properly your PR ?

lolodomo (approved):

lolodomo:
As this is only 2 lines added and before the fix is obvious, I am merging even if the PR is not properly signed off.

PR #14507, opened 2023-02-26
[harmonyhub] Use AbstractStorageBasedTypeProvider

J-N-K:
This is the second of two reference implementations for the use of dynamic type providers.

### DIFF
--- a/bundles/org.openhab.binding.harmonyhub/src/main/java/org/openhab/binding/harmonyhub/internal/HarmonyHubDynamicTypeProvider.java
+++ b/bundles/org.openhab.binding.harmonyhub/src/main/java/org/openhab/binding/harmonyhub/internal/HarmonyHubDynamicTypeProvider.java
@@ -15,7 +15,7 @@ package org.openhab.binding.harmonyhub.internal;
 import org.eclipse.jdt.annotation.NonNullByDefault;
 import org.openhab.core.storage.StorageService;
 import org.openhab.core.thing.ThingUID;
-import org.openhab.core.thing.binding.AbstractDynamicTypeProvider;
+import org.openhab.core.thing.binding.AbstractStorageBasedTypeProvider;
 import org.openhab.core.thing.type.ChannelType;
 import org.openhab.core.thing.type.ChannelTypeProvider;
 import org.osgi.service.component.annotations.Activate;
@@ -23,14 +23,15 @@ import org.osgi.service.component.annotations.Component;
 import org.osgi.service.component.annotations.Reference;
 
 /**
- * The {@link HarmonyHubDynamicTypeProvider} is an instance of a {@link AbstractDynamicTypeProvider} for the HarmonyHub
+ * The {@link HarmonyHubDynamicTypeProvider} is an instance of a {@link AbstractStorageBasedTypeProvider} for the
+ * HarmonyHub
  * binding
  *
  * @author Jan N. Klug - Initial contribution
  */
 @Component(service = { HarmonyHubDynamicTypeProvider.class, ChannelTypeProvider.class })
 @NonNullByDefault
-public class HarmonyHubDynamicTypeProvider extends AbstractDynamicTypeProvider {
+public class HarmonyHubDynamicTypeProvider extends AbstractStorageBasedTypeProvider {
 
     @Activate
     public HarmonyHubDynamicTypeProvider(@Reference StorageService storageService) {
--- a/bundles/org.openhab.binding.harmonyhub/src/main/java/org/openhab/binding/harmonyhub/internal/HarmonyHubDynamicTypeProvider.java
+++ b/bundles/org.openhab.binding.harmonyhub/src/main/java/org/openhab/binding/harmonyhub/internal/HarmonyHubDynamicTypeProvider.java
@@ -39,7 +39,7 @@ public class HarmonyHubDynamicTypeProvider extends AbstractStorageBasedTypeProvi
     }
 
     public void removeChannelTypesForThing(ThingUID uid) {
-        String thingUid = uid.getAsString();
+        String thingUid = uid.getAsString() + ":";
         getChannelTypes(null).stream().map(ChannelType::getUID).filter(c -> c.getAsString().startsWith(thingUid))
                 .forEach(this::removeChannelType);
     }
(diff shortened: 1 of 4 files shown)


bundles/org.openhab.binding.harmonyhub/src/main/java/org/openhab/binding/harmonyhub/internal/HarmonyHubDynamicTypeProvider.java
```diff
@@ -0,0 +1,46 @@
...
+public class HarmonyHubDynamicTypeProvider extends AbstractStorageBasedTypeProvider {
+
+    @Activate
+    public HarmonyHubDynamicTypeProvider(@Reference StorageService storageService) {
+        super(storageService);
+    }
+
+    public void removeChannelTypesForThing(ThingUID uid) {
+        String thingUid = uid.getAsString();
+        getChannelTypes(null).stream().map(ChannelType::getUID).filter(c -> c.getAsString().startsWith(thingUid))
```
ccutrer, line 43: what if you have thingUids `harmonyhub:thing1` and `harmonyhub:thing11`, and the former gets removed. wouldn't this accidentally remove the channel types for thing11? seems like you need to suffix thingUid with `:` to avoid that problem.

ccutrer (approved):

bundles/org.openhab.binding.harmonyhub/src/main/java/org/openhab/binding/harmonyhub/internal/HarmonyHubDynamicTypeProvider.java
```diff
@@ -0,0 +1,46 @@
...
+@Component(service = { HarmonyHubDynamicTypeProvider.class, ChannelTypeProvider.class })
+@NonNullByDefault
+public class HarmonyHubDynamicTypeProvider extends AbstractStorageBasedTypeProvider {
+
+    @Activate
+    public HarmonyHubDynamicTypeProvider(@Reference StorageService storageService) {
+        super(storageService);
+    }
+
+    public void removeChannelTypesForThing(ThingUID uid) {
```
lolodomo, line 41: Isn't it possible and useful to put that method in the abstract class (core framework)) ?

lolodomo (commented):
The changes look easy to apply.
Just one question.

bundles/org.openhab.binding.harmonyhub/src/main/java/org/openhab/binding/harmonyhub/internal/HarmonyHubDynamicTypeProvider.java
J-N-K: No, because the core class also works for thing-types, where this method should to make sense.

lolodomo: What is the problem if the method is properly documented ?

If not, I understand that every binding implementing a channel type provider will have to duplicate this method ?

lolodomo:
@digitaldan ! can you please have a look to this PR and the related PR in core framework to tell us if the change looks good to you ?

bundles/org.openhab.binding.harmonyhub/src/main/java/org/openhab/binding/harmonyhub/internal/HarmonyHubDynamicTypeProvider.java
jlaur: Side comment: Can it be specialized into a subclass in core?

J-N-K: @lolodomo IMO it's bad design to implement methods in a generic super-class that only apply to some specialized cases. It doesn't even apply to all `ChannelTypeProviders` because it is not necessary to have them "per-thing", you could also make them available for all things, especially if you don't have a thing-type/thing but a limited set of dynamic thing-types with dynamic channel-types.

@jlaur That would be possible. But on the other hand: It's a pretty straight-forward two-liner. I don't think its necessary to create a class for that.

bundles/org.openhab.binding.harmonyhub/src/main/java/org/openhab/binding/harmonyhub/internal/HarmonyHubDynamicTypeProvider.java
```diff
@@ -0,0 +1,46 @@
...
+        String thingUid = uid.getAsString() + ":";
+        getChannelTypes(null).stream().map(ChannelType::getUID).filter(c -> c.getAsString().startsWith(thingUid))
```
lolodomo, line 43: I have the strange feeling that this filter makes no sense. How a channel type UID could start by a thing UID ?
So I prefer to ask the question.

J-N-K: Because the channel-types are created per thing-type, so the (functionally) same channel has different channel-types on different things. The channel-type UIDs of these channels contain the thing's UID. 

Anyway, I would suggest to wait merging this (and the system info) PR until we solved the state description issue, so we don't create corrupted storages.

bundles/org.openhab.binding.harmonyhub/src/main/java/org/openhab/binding/harmonyhub/internal/HarmonyHubDynamicTypeProvider.java
lolodomo: By the way, I realized that this method is in fact very specific to this binding and will not be required by all bindings.

bundles/org.openhab.binding.harmonyhub/src/main/java/org/openhab/binding/harmonyhub/internal/HarmonyHubDynamicTypeProvider.java
lolodomo: Ok

lolodomo (approved):
LGTM, thank you

lolodomo:
@digitaldan : would you like to test and review this PR before we merge it ?

digitaldan:
>@digitaldan : would you like to test and review this PR before we merge it ?

Thanks, i'll test this out today and get back to everyone

digitaldan:
I compiled and updated this on my local instance , the binding stays offline for 2 mins as before, then comes backs  online and works normally from there.  I have not had time to run through the core changes, i might be able to look on sunday, is there something else i need to be doing, like removing the thing and adding it back in ?  (i assume its suppose to not have the 2 min delay now)

J-N-K:
The 2min delay will be there for old things on first startup (because the binding can only create the dynamic types after/during first initialization). Ok subsequent starts it should not be there.

digitaldan:
Yep , after restarting the binding it now loads up right away.

digitaldan (approved):
LGTM!

PR #14512, opened 2023-02-26
[miio] Update readme wrt unsupported devices

marcelrv:
* Rename channel names to better reflect purpose
* Update readme


daMihe (approved):
Extending documentation is always appreciated. I understand it and even though i'm not a native speaker it is good readable.

jlaur (approved):
Thanks!

PR #14513, opened 2023-02-27
[kaleidescape] Remove Apache StringEscapeUtils

mlobstein:
Replace StringEscapeUtils.unescapeHtml4() with basic Java methods to address #7722. Also a couple very minor tweaks that I have had pending for some time now.

### DIFF
--- a/bundles/org.openhab.binding.kaleidescape/README.md
+++ b/bundles/org.openhab.binding.kaleidescape/README.md
@@ -229,7 +229,9 @@ String z1_Detail_DiscLocation "Disc Location: [%s]" { channel="kaleidescape:play
 ksecondsformat.js:
 
 ```javascript
-(function(totalSeconds) {
+(function(timestamp) {
+    var totalSeconds = Date.parse(timestamp) / 1000
+
     if (isNaN(totalSeconds)) {
         return '-';
     } else {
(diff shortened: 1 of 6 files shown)


lsiepel:
The code looks good. Only thing is, i can't tell if it handles all required cases.  `StringEscapeUtils.unescapeHtml4` handles ~290 different cases. 
It would be really helpfull if a unit test was provided. Throwing in some examples into the stringformatter method.

lsiepel (commented):
The code looks good. Only thing is, i can't tell if it handles all required cases. StringEscapeUtils.unescapeHtml4 handles ~290 different cases.
It would be really helpfull if a unit test was provided. Throwing in some examples into the stringformatter method.

J-N-K:
IIRC we use unbescape in other add-ons

lsiepel:
> IIRC we use unbescape in other add-ons

And we try to get rid of them. Would be nice to have such a utility function in core. Please see https://github.com/openhab/openhab-addons/issues/7722#issuecomment-1434669804

J-N-K:
If you scroll up in that issue, you can see that this es exactly the reason why we introduced unbescape. It‘s hard to get it right.

lsiepel:
> hat issue, you can see that this es exactly the

Totally agree, unbescape got somewhat burried in that topic. i re-read and allready got a PR for BSBLAN with that dependency. Would be nice if someone could have commented in that topic about my30+ PR's though.

mlobstein:
I added some tests. This implementation only needs to un-escape the Latin-1 characters as described in the protocol reference manual:
image

lsiepel (approved):
lgtm

jlaur (approved):
Thanks!

PR #14515, opened 2023-02-28
[bsblan] Remove org.apache.commons

lsiepel:
- Replaced the org.apache.commons with the unbescape package.

Not tested, just technicall change. Should work as this package has been used in several other bindings as replacement for `unescapeHtml4`


hypetsch (approved):
LGTM

PR #14517, opened 2023-03-01
[ecobee] Increase temperature precision to decimal

darthbevis:
Correcting rounding of setHold values so we have 0.1°F tolerance instead of 1°F.  Makes it easier to use in °C

Signed-off-by: Bevan Watkiss <darthbevis@gmail.com>

### DIFF
--- a/bundles/org.openhab.binding.ecobee/src/main/java/org/openhab/binding/ecobee/internal/handler/EcobeeUtils.java
+++ b/bundles/org.openhab.binding.ecobee/src/main/java/org/openhab/binding/ecobee/internal/handler/EcobeeUtils.java
@@ -110,7 +110,7 @@ public final class EcobeeUtils {
             QuantityType<Temperature> convertedTemp = ((QuantityType<Temperature>) value)
                     .toUnit(ImperialUnits.FAHRENHEIT);
             if (convertedTemp != null) {
-                return Integer.valueOf((int)(convertedTemp.doubleValue() * 10));
+                return Integer.valueOf((int) (convertedTemp.doubleValue() * 10));
             }
         }
         throw new IllegalArgumentException("temperature is not a QuantityType");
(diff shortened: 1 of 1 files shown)


lsiepel (approved):
lgtm

bundles/org.openhab.binding.ecobee/src/main/java/org/openhab/binding/ecobee/internal/handler/EcobeeUtils.java
```diff
@@ -110,7 +110,7 @@ public static Integer convertQuantityTypeToEcobeeTemp(Object value) {
             QuantityType<Temperature> convertedTemp = ((QuantityType<Temperature>) value)
                     .toUnit(ImperialUnits.FAHRENHEIT);
             if (convertedTemp != null) {
-                return Integer.valueOf(convertedTemp.intValue() * 10);
+                return Integer.valueOf((int)(convertedTemp.doubleValue() * 10));
```
mhilbush, line 113: Yes, since the Ecobee API supports tenths, it's a good idea to take advantage of it.

But, this always will truncate (i.e. round down), correct? So, for example, 70.999 will convert to 709. Is that what we want, or would we rather it round up or down (to 710 in my example)? Just wondering why you chose to truncate vs round up/down.

mhilbush (commented):
Just have one question about the conversion truncating vs rounding up/down.

bundles/org.openhab.binding.ecobee/src/main/java/org/openhab/binding/ecobee/internal/handler/EcobeeUtils.java
mhilbush: Yes, clearly it was wrong for the origin code to truncate to whole degrees. 😉 

And for C values of .0 and .5 (which I suspect are the most likely), truncating will work perfectly fine.

mhilbush (approved):
LGTM

lsiepel:
You still have to apply spotless, to fix the build.

jlaur:
> You still have to apply spotless, to fix the failing build.

When spotless is applied and DCO is fixed, we can merge.

lsiepel:
May i suggest a hsorter title for the release notes?:
[ecobee] Increase temperature precision to decimal

jlaur (approved):
Thanks!

PR #14518, opened 2023-03-01
[http] Add thing config for User-Agent header

### LINKED ISSUE #9265: [HTTP] Feature Request - Off Cycle Refresh & Agent String
Requesting two features be added to the HTTP binding:

1) Ability to trigger an off cycle refresh.  Some devices do not work well getting flooded with API requests so dialing the refresh down is not necessarily desirable.  Example usage: Assume we send a power on/off command to a device via the binding moments after a scheduled update.  There will be an obvious want to update the channels more frequently than the standard refresh.  An action for the thing could likely do this as part of a rule.  For example:

rule "TV Power Change"
when
    Item TVPower received command
then
    val httpActions = getActions("http","http:url:mytv")
    httpActions.forceRefresh()
end

2) Some APIs require specific agent strings to be set for functionality.  If this could be set as part of the Thing it would enable this.  Similar in function to "curl -A"

jimtng:
Resolve #9265


J-N-K:
I would prefer not to merge this as it will make re-integration of the SmartHome/J changes more complex and these include a solution for the issue as well.

PR #14520, opened 2023-03-02
[renault] Add Pause / Resume feature

fifipil909:
**Adding pause / resume charge feature**

Description

I add the Pause / Resume API Endpoint.
This can be handy to pause / Resume the charge of the car. Especially like in france when there is some period of the day where the electricity is less expensive.

Test succefully on 3.4


dougculnane:
@fifipil909 Thanks for the PR.  

The it looks like the main branch which runs on openHAB 4 has a lot of code changes that are confused with your PR.

Looks like you got to merge in the main branch and sign off your commits.... and do your PR to main.  

I am very interested to see how you pause charging.  I use the charge mode or the car charger binding to control charging.

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/api/MyRenaultHttpSession.java
```diff
@@ -233,108 +248,110 @@ public void getLocation(Car car)
...
+                + config.vin + "/actions/charge-mode?country=" + getCountry(config);
+        postKamereonRequest(path,
+                "{\"data\":{\"type\":\"ChargeMode\",\"attributes\":{\"action\":\"" + apiMode + "\"}}}");
+    }
+
+    public void actionPause(PauseMode mode)
+            throws RenaultForbiddenException, RenaultNotImplementedException, RenaultActionException {
+        final String apiMode = PauseMode.PAUSE.equals(mode) ? "pause" : "resume";
+        final String path = "/commerce/v1/accounts/" + kamereonaccountId + "/kamereon/kcm/v1/vehicles/" + config.vin
+                + "/charge/pause-resume?country=" + getCountry(config);
```
dougculnane, line 282: I did not know this existed.  It is not documented here: https://renault-api.readthedocs.io/en/latest/endpoints.html#action-endpoints.

dougculnane:
@fifipil909 There is a helpful community here: https://community.openhab.org/t/betatest-renault-ze-services-binding/72226/237 where you might find help testing this and getting an idea of which cars this feature is available and work on.

bundles/org.openhab.binding.renault/src/main/java/org/openhab/binding/renault/internal/api/MyRenaultHttpSession.java
fifipil909: Hey,

you can find it here : https://renault-api.readthedocs.io/en/latest/endpoints/vehicle_kcm_actions.charge-pause-resume.html

The charge mode is not working on all model. For example I do have a Dacia Spring which is using the same Renault API, but there are some limitation. "Charging mode" is not compatible. The only thing I found to pause the charge is to use the above.

I did not migrate on OH 4 that's why I push on the 3.4 branch.

if you prefer I can try to do the same change on the OH 4 main branch.

Let me know what is you thoughs.

Vincent.

dougculnane: @fifipil909  I think it would be easier for me to see your changes if we do the PR to main.  Maybe then a PR from main to 3.4 would be worth doing but it would need a bit of basic testing on 3.4.  I have migrated to 4-SNAPSHOTs so that is where I can test your changes on my wife's Zoe.  

For sure this is a useful feature.  We should document it and the charge_mode in the possible limitations of README.md if they do not work of some cars.  (however it is very hard to test on all car models etc.. but we should give the users the idea that some stuff might not work on their car).

PR #14004, opened 2022-12-18
Adapt addons to core watch service changes

J-N-K:
Depends on https://github.com/openhab/openhab-core/pull/3004

PR #14522, opened 2023-03-02
[exec] Fix whitelist not read

J-N-K:
Reported on the forum.


jlaur:
I assume it's a regression from #14004.

jlaur (approved):
LGTM, thanks.

J-N-K:
Exactly, that’s why I wanted to have these core changes merged early, so that we can find and fix what went wrong.

jlaur:
> Exactly, that’s why I wanted to have these core changes merged early, so that we can find and fix what went wrong.

Yes, just wanted to add this for context as I needed to check the change from that PR myself in order to understand the issue. 🙂

Issue #14502, opened 2023-02-25
[knx] Backport SmartHome/J enhancements and fixes


holgerfriedrich:
knx binding was forked to [SmartHome/J](https://github.com/smarthomej), and in the meantime has evolved in both repositories.

This issue summarizes all improvements done in [openHAB KNX Binding at SmartHome/J](https://github.com/smarthomej/addons/tree/4.0.x/bundles/org.smarthomej.binding.knx)
and the status of backporting those enhancements as discussed in #14129.

- [x] smarthomej/addons#13 - thread safety #14509
- [x] smarthomej/addons#46 - (minor null annotations; I did not import changes to ```Firmware.h``` as it is rewritten in #14050) #14509
- [x] smarthomej/addons#60 - (minor null annotations) #14509
- [x] smarthomej/addons#73 - already implemented, backport doc change, comment on 1.009 #14510
- [x] ⛔ smarthomej/addons#74 - backport doc change, backport dpt22 to sting, maybe new test in KNXCoreTypeMapperTest. Note that Calimero supports more subtypes by now. #14511 (obsoleted by following, 107)
- [x] smarthomej/addons#104 - remove unnecessary workaround in knx (already fixed in calimero 2.4), KNXCoreTypeMapper.java:902 #14503 
- [x] smarthomej/addons#107
  - [x] prepare refactoring of KnxCoreTypeMapper #14534; 
  - [x] documentation of dpts, dpt.lst, #14851
- [x] ⛔ smarthomej/addons#125 - no longer necessary, subtype is not converted to int
- [x] smarthomej/addons#114 - KNXChannel #14534
- [x] smarthomej/addons#162 - fix DPT_Control_Dimming, fix updateState call #14504
- [x] smarthomej/addons#203 - warning for incompatible types, #14534
- [x] smarthomej/addons#206 - UoM #14534
- [x] smarthomej/addons#208
  - [x] Refactor KNXCoreTypeMapper, add RGBW and xyY #14534 ⛔ did not port device readout due to #14050,
  - [x] port reading of manufacturer list from external file #15297  
  - [x] use ```ColorUtil``` from openhab/openhab-core#3434
- [x] smarthomej/addons#219 - Fix RGB conversion #14534
- [x] smarthomej/addons#226 - Remove workarounds obsoleted by Calimero 2.5 #14534
- [x] smarthomej/addons#230 - Add parameter for disabling incoming UoM #14534
[...]

Issue #14124, opened 2022-12-31
[deconz] Backport SmartHome/J enhancements and fixes


jlaur:
The deCONZ binding was forked to [SmartHome/J](https://github.com/smarthomej) and has since lacked support in the openHAB addons repository. This issue is for discussing what to do with this binding and possibly backport changes from the fork.

Issues that would need to be tackled to make this backport possible:
- Some parts of [org.smarthomej.commons](https://github.com/smarthomej/addons/tree/3.2.x/bundles/org.smarthomej.commons) are needed, for example `ColorUtil`.

## ColorUtil

https://github.com/smarthomej/addons/blob/813a17163dcd1c52ba11fe4dc8755f3bfd899b70/bundles/org.smarthomej.commons/src/main/java/org/smarthomej/commons/util/ColorUtil.java#L38-L82

Is this the same?

https://github.com/openhab/openhab-core/blob/27b847f40ca6e81723666faddb455294e39a4f8e/bundles/org.openhab.core/src/main/java/org/openhab/core/library/types/HSBType.java#L367-L394

Now being consolidated by openhab/openhab-core#3434

See also #13108

jlaur:
@MHerbst - FYI.

PR #14534, opened 2023-03-04
[knx] Refactoring of KnxCoreTypeMapper and UOM Support

holgerfriedrich:
This is a merge of smarthome/j code to current code base.
It will include several PRs from smarthome/j related to UOM and refactoring of KnxCoreTypeMapper.

Carryover from 
https://github.com/smarthomej/addons/pull/107
https://github.com/smarthomej/addons/pull/114
https://github.com/smarthomej/addons/pull/203
https://github.com/smarthomej/addons/pull/206
https://github.com/smarthomej/addons/pull/208
https://github.com/smarthomej/addons/pull/219
https://github.com/smarthomej/addons/pull/226
https://github.com/smarthomej/addons/pull/230
https://github.com/smarthomej/addons/pull/279
https://github.com/smarthomej/addons/pull/299
https://github.com/smarthomej/addons/pull/330
https://github.com/smarthomej/addons/pull/349

Also-by: Jan N. Klug <github@klug.nrw>

Refs: #14502


bundles/org.openhab.binding.knx/src/main/java/org/openhab/binding/knx/internal/dpt/KNXCoreTypeMapper.java
```diff
@@ -732,74 +319,72 @@ private String quantityTypeToDPTValue(QuantityType<?> qt, int mainNumber, int su
...
-                                return "r:" + convertPercentToByte(hc.getRed()) + " g:"
-                                        + convertPercentToByte(hc.getGreen()) + " b:"
-                                        + convertPercentToByte(hc.getBlue());
-                        }
-                    default:
-                        HSBType hc = ((HSBType) type);
-                        return "r:" + hc.getRed().intValue() + " g:" + hc.getGreen().intValue() + " b:"
-                                + hc.getBlue().intValue();
+            if (value instanceof HSBType) {
+                HSBType hsb = ((HSBType) value);
```
J-N-K, line 353: Use pattern matching:

```suggestion
            if (value instanceof HSBType hsb) {
```

J-N-K (commented):
I didn't check in detail if everything is ok, just a comment while scrolling though the code. But I did not see anything obviously wrong.

bundles/org.openhab.binding.knx/src/main/java/org/openhab/binding/knx/internal/dpt/KNXCoreTypeMapper.java
holgerfriedrich: thanks @J-N-K. that's a nice java17 feature... I have now used this also for the other occurrences.

holgerfriedrich:
@J-N-K I am just integrating smarthomej/addons/#208 which requires ```ColorUtil```. How should I proceed with it? For now add it to the KNX binding, unchanged or stripped to a minimum?

jlaur:
> @J-N-K I am just integrating smarthomej/addons/#208 which requires ```ColorUtil```. How should I proceed with it? For now add it to the KNX binding, unchanged or stripped to a minimum?

See also comment about `ColorUtil` here: https://github.com/openhab/openhab-addons/issues/14124#issue-1515136901

holgerfriedrich:
Thanks, @jlaur. It looks like ```HSBType``` from core could cover 2 of the 3 occurrences (though the implementation might be different), but it is lacking support for xyY color format used by ```DPT_Colour_xyY``` 242.600.

jlaur:
> It looks like `HSBType` from core could cover 2 of the 3 occurrences (though the implementation might be different), but it is lacking support for xyY color format used by `DPT_Colour_xyY` 242.600.

I didn't analyze it further since I was busy with something else at the time, so just took note. Maybe it would be worth adding the missing format to the core `HSBType` implementation? My initial thought was to compare unit test coverage in the two implementations and possibly detect any differences by new coverage, or have proof that the implementations indeed produce same results.

J-N-K:
They don‘t provide the same results, otherwise I wouldn’t have implemented it in another way :-)

I would suggest to adjust core to the implementation on SmartHome/J (including the tests).

J-N-K:
Regarding color: see https://github.com/openhab/openhab-core/pull/3434

holgerfriedrich:
@J-N-K this PR has become quite huge, as it includes the complete restructuring of ```KNXCoreTypeMapper``` and the introduction of UoM. Functionally, it seems complete. The other carryover tasks can be handled separately.

I have kept the work split into several commits to match the changes in smarthome/j and added additional commits to fix the unit tests for UoM we had added in this repo only.

One open point is HSB color transform with (x,y,Y) - this is currently failing until we have ColorUtil in the core.

I did not have time yet to check it with my KNX installation.

J-N-K:
With all the work you have done here, I would suggest to add yourself as a CODEOWNER.

holgerfriedrich:
@J-N-K I just gave it a test with my production system, where UoM is actively in use.
Seems to work fine. 🎉 

The only problem I see is the different behavior of DPT 5.001. It was mapped 0..1 before, not it is 0..100 (percentage).
Item is ```Number::Dimensionless```, Thing is  type ```number```. Maybe you could give some hint how percentage values should be dealt with. I think I have read a complaint before that KNX binding does not properly care about percentages.

J-N-K:
If the item is `Number:Dimensionless` the state update should be `67 %` which is the same as `0.67` if you chose unit `ONE`.

The "old" issue was that 5.001 can be used with `Number[:Dimensionless]`, `Dimmer` and `Rollershutter` which expect different type. IIRC, in the early days the conversion was always to `PercentType` which extends `DecimalType` and therefore can be used for `Number`, `Dimmer` and `Rollershutter`. Conversion between `DecimalType` and `PercentType` is sometimes confusing and should be avoided.

The code now uses `QuantityType` for `Number` (the unit is stripped for a plain item and therefore ends up with `67` instead of `0.67`) and `PercentType` for `Dimmer` and `Rollershutter`.

The behavior now is IMO consistent with the way other bindings work, although it may be a bit different than before. Unfortunately I don't know any way around this, if we go for 100% backward compatibility (which would probably mean to remove UoM for 5.001) this will result in unexpected behavior for new users.

My suggestions would be to add a warning to the upgrade list and keep it as it is now.

holgerfriedrich:
@J-N-K thanks for the comments above, lets proceed this way.

I have made some progress, but ran into an issue with UoM. I had a few items which did not update, nothing in the log. ```updateState(knxChannel.getChannelUID(), (State) value);``` failed silently. In the end this was caused by state description ```Number Livingroom_Temperature "Temperature [%.1f C]``` (C instead of °C). (this happened on my 3.4.2 system, have to check with 4.0.0 later). 

I have seen some exception in the log as well, which need further handling.

Overall, I think this needs a few more days of testing.

J-N-K:
I don't think this is an issue with the KNX code. The reason it fails is that `C` is also a valid unit, but not for temperature but electric charge. Converting `°C` to `C` is obviously not supported and the `NumberItem` checks if the configured unit (from the state description) is compatible with the provided state update. If it is not, then the state update is rejected:

```
if (state instanceof QuantityType) {
            Unit<?> itemUnit = getUnit(dimension, true);
            Unit<?> stateUnit = ((QuantityType<?>) state).getUnit();
            if (itemUnit != null && (!stateUnit.getSystemUnit().equals(itemUnit.getSystemUnit())
                    || UnitUtils.isDifferentMeasurementSystem(itemUnit, stateUnit))) {
                QuantityType<?> convertedState = ((QuantityType<?>) state).toInvertibleUnit(itemUnit);
                if (convertedState != null) {
                    super.setState(convertedState);
                    return;
                }

                // the state could not be converted to an accepted unit.
                return;
            }
        }
```

holgerfriedrich:
@J-N-K I agree this is not an issue with KNX binding. My configuration was plainly wrong. The point I am trying to make is that is fails *silently*. This is quite annoying, especially as the state description was just related to presentation (before the introduction of UoM). When you start porting old configurations, an error message - visible at normal INFO or WARN level - would be very much appreciated.
This would probably be something to be implemented at core, as the binding does not get any feedback calling ```updateState()``` from ```BaseThingHandler```.

J-N-K:
I agree, this should be solved with the solution for https://github.com/openhab/openhab-core/issues/3282

holgerfriedrich:
just a quick look into the logs, and of course my config had still errors, and I reloaded the plugin a lot....

```
openhab.log.6.gz:2023-03-12 22:59:04.396 [WARN ] [mmon.WrappedScheduledExecutorService] - Scheduled runnable ended with an exception:
openhab.log.6.gz-java.lang.IllegalArgumentException: Value must be between 0 and 100
openhab.log.6.gz-       at org.openhab.core.library.types.PercentType.validateValue(PercentType.java:93) ~[?:?]
openhab.log.6.gz-       at org.openhab.core.library.types.PercentType.<init>(PercentType.java:88) ~[?:?]
openhab.log.6.gz-       at org.openhab.core.library.types.DecimalType.as(DecimalType.java:192) ~[?:?]
openhab.log.6.gz-       at org.openhab.core.internal.items.ItemStateConverterImpl.convertToAcceptedState(ItemStateConverterImpl.java:64) ~[?:?]
openhab.log.6.gz-       at org.openhab.core.thing.internal.profiles.ProfileCallbackImpl.sendUpdate(ProfileCallbackImpl.java:126) ~[?:?]
openhab.log.6.gz-       at org.openhab.core.thing.internal.profiles.SystemDefaultProfile.onStateUpdateFromHandler(SystemDefaultProfile.java:53) ~[?:?]
openhab.log.6.gz-       at org.openhab.core.thing.internal.CommunicationManager.lambda$13(CommunicationManager.java:568) ~[?:?]
openhab.log.6.gz-       at org.openhab.core.thing.internal.CommunicationManager.lambda$15(CommunicationManager.java:588) ~[?:?]
openhab.log.6.gz-       at java.lang.Iterable.forEach(Iterable.java:75) ~[?:?]
openhab.log.6.gz-       at org.openhab.core.thing.internal.CommunicationManager.handleCallFromHandler(CommunicationManager.java:584) ~[?:?]
openhab.log.6.gz-       at org.openhab.core.thing.internal.CommunicationManager.stateUpdated(CommunicationManager.java:566) ~[?:?]
openhab.log.6.gz-       at org.openhab.core.thing.internal.ThingHandlerCallbackImpl.stateUpdated(ThingHandlerCallbackImpl.java:65) ~[?:?]
openhab.log.6.gz-       at org.openhab.core.thing.binding.BaseThingHandler.updateState(BaseThingHandler.java:268) ~[?:?]
openhab.log.6.gz-       at org.openhab.binding.knx.internal.handler.DeviceThingHandler.processDataReceived(DeviceThingHandler.java:360) ~[?:?]
openhab.log.6.gz-       at org.openhab.binding.knx.internal.handler.DeviceThingHandler.onGroupWrite(DeviceThingHandler.java:310) ~[?:?]
openhab.log.6.gz-       at org.openhab.binding.knx.internal.client.AbstractKNXClient$1.lambda$0(AbstractKNXClient.java:123) ~[?:?]
openhab.log.6.gz-       at org.openhab.binding.knx.internal.client.AbstractKNXClient.lambda$8(AbstractKNXClient.java:328) ~[?:?]
```


```
openhab.log.6.gz:2023-03-12 16:38:07.900 [WARN ] [mmon.WrappedScheduledExecutorService] - Scheduled runnable ended with an exception:
openhab.log.6.gz-java.lang.NoSuchMethodError: 'void org.openhab.core.library.types.DecimalType.<init>(double)'
openhab.log.6.gz-       at org.openhab.binding.knx.internal.dpt.ValueDecoder.handleNumericDpt(ValueDecoder.java:370) ~[?:?]
openhab.log.6.gz-       at org.openhab.binding.knx.internal.dpt.ValueDecoder.decode(ValueDecoder.java:149) ~[?:?]
openhab.log.6.gz-       at org.openhab.binding.knx.internal.handler.DeviceThingHandler.processDataReceived(DeviceThingHandler.java:322) ~[?:?]
openhab.log.6.gz-       at org.openhab.binding.knx.internal.handler.DeviceThingHandler.onGroupWrite(DeviceThingHandler.java:310) ~[?:?]
openhab.log.6.gz-       at org.openhab.binding.knx.internal.client.AbstractKNXClient$1.lambda$0(AbstractKNXClient.java:123) ~[?:?]
openhab.log.6.gz-       at org.openhab.binding.knx.internal.client.AbstractKNXClient.lambda$8(AbstractKNXClient.java:328) ~[?:?]
```

So I added a try catch block to give a more userfriendly message.....

bundles/org.openhab.binding.knx/README.md
```diff
@@ -9,14 +9,27 @@ The KNX binding then can communicate directly with this gateway.
...
 Since the protocol is identical, the KNX binding can also communicate with it transparently.
 
+***Attention:*** With the introduction of UoM support (see `number` channel below) some data types have changed:
+
+- the data type for DPT 5.001 (Percent 8bit, 0 -> 100%) has changed from `PercentType` to `QuantityType`for `number` channels (`dimmer`, `color`, `rollershutter` channels stay with `PercentType`)
+- the data type for DPT 5.004 (Percent 8bit, 0 -> 255%) has changed from `PercentType` to `QuantityType`
+- the data type for DPT 6.001 (Percent 8bit -128 -> 127%) has changed from `PercentType` to `QuantityType`
+- the data type for DPT 9.007 (Humidity) has changed from `PercentType` to `QuantityType`
+
+Rules that check for or compare states and transformations that expect a raw value might need adjustments.
```
jlaur, line 19: I'm no expert in English comma rules, but this sentence seems a little complex to read without any. 🙂

bundles/org.openhab.binding.knx/README.md
```diff
@@ -76,45 +89,30 @@ All channels of a device share one configuration parameter defined on device lev
...
-Note: After changing the DPT of already existing Channels, openHAB needs to be restarted for the changes to become effective.
-
-##### Channel Type "switch"
-
-| Parameter | Description                         | Default DPT |
-|-----------|-------------------------------------|-------------|
-| ga        | Group address for the binary switch | 1.001       |
+Control channel types (suffix `-control`) are used for cases where the KNX bus does not own the physical state of a device.
+This could be the case if e.g. a lamp from another binding should be controlled by a KNX wall switch.
+If from the KNX bus a `GroupValueRead` telegram is sent to a *-control Channel, the bridge responds with a `GroupValueResponse` telegram to the KNX bus.
```
jlaur, line 99:
```suggestion
If a `GroupValueRead` telegram from the KNX bus is sent to a *-control Channel, the bridge responds with a `GroupValueResponse` telegram to the KNX bus.
```

bundles/org.openhab.binding.knx/README.md
```diff
@@ -123,26 +121,57 @@ Note: After changing the DPT of already existing Channels, openHAB needs to be r
...
+##### Channel Type `number`, `number-control` 
 
 | Parameter | Description   | Default DPT |
 |-----------|---------------|-------------|
 | ga        | Group address | 9.001       |
 
-Note: Using the Units Of Measurement feature of openHAB (Quantitytype) requires that the DPT value is set correctly.
+Note: The `number` channel has full support for UoM. 
+
+Using the Units Of Measurement feature of openHAB (QuantityType) requires that the DPT value is set correctly.
```
jlaur, line 146: If full name is needed, perhaps this should be introduced in the first of the two sentences?
```suggestion
Note: The `number` channel has full support for Units Of Measurement (UoM).

Using the UoM feature of openHAB (QuantityType) requires that the DPT value is set correctly.
```

bundles/org.openhab.binding.knx/README.md
```diff
@@ -123,26 +121,57 @@ Note: After changing the DPT of already existing Channels, openHAB needs to be r
...
 Automatic type conversion will be applied if required.
 
-##### Channel Type "string"
+Incoming values from the KNX bus are converted to values with units (e.g. `23 °C`).
+If the channel is linked to the correct item-type (`Number:Temperature` in this case) the display unit can be controlled my item metadata (e.g. `%.1f °F` for 1 digit of precision in Fahrenheit).
```
jlaur, line 150:
```suggestion
If the channel is linked to the correct item-type (`Number:Temperature` in this case) the display unit can be controlled by item metadata (e.g. `%.1f °F` for 1 digit of precision in Fahrenheit).
```

bundles/org.openhab.binding.knx/src/main/java/org/openhab/binding/knx/internal/channel/GroupAddressConfiguration.java
```diff
@@ -12,41 +12,106 @@
...
 /**
- * Data structure representing a single group address configuration within a channel configuration parameter.
+ * Data structure representing the content of a channel's group address configuration.
  *
  * @author Simon Kaufmann - initial contribution and API.
  *
  */
 @NonNullByDefault
 public class GroupAddressConfiguration {
+    public static final Logger LOGGER = LoggerFactory.getLogger(GroupAddressConfiguration.class);
```
jlaur, line 36: Loggers should be non-static. For reference: https://www.openhab.org/docs/developer/guidelines.html#f-logging

bundles/org.openhab.binding.knx/src/main/java/org/openhab/binding/knx/internal/channel/KNXChannelFactory.java
```diff
@@ -0,0 +1,61 @@
...
+
+import org.eclipse.jdt.annotation.NonNullByDefault;
+import org.openhab.core.thing.Channel;
+import org.openhab.core.thing.type.ChannelTypeUID;
+
+/**
+ * Helper class to find the matching {@link KNXChannel} for any given {@link ChannelTypeUID}.
+ *
+ * @author Simon Kaufmann - initial contribution and API
+ * @author Jan N. Klug - refactored ro factory class
```
jlaur, line 27:
```suggestion
 * @author Simon Kaufmann - Initial contribution
 * @author Jan N. Klug - refactored to factory class
```

bundles/org.openhab.binding.knx/src/main/java/org/openhab/binding/knx/internal/channel/KNXChannel.java
```diff
@@ -0,0 +1,154 @@
...
+import org.openhab.core.types.Type;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import tuwien.auto.calimero.GroupAddress;
+
+/**
+ * Meta-data abstraction for the KNX channel configurations.
+ *
+ * @author Simon Kaufmann - initial contribution and API
```
jlaur, line 43: Not sure, but I believe this would otherwise trigger a SAT warning?
```suggestion
 * @author Simon Kaufmann - Initial contribution
```

bundles/org.openhab.binding.knx/src/main/java/org/openhab/binding/knx/internal/client/OutboundSpec.java
```diff
@@ -39,13 +38,14 @@ public interface OutboundSpec {
...
 
     /**
      * The command or state to be sent.
      *
      * @return the command/state
      */
-    Type getType();
+    Type getValue();
+
+    boolean matchesDestination(GroupAddress groupAddress);
```
jlaur, line 50: Can we have Javadoc for this method also?

bundles/org.openhab.binding.knx/src/main/java/org/openhab/binding/knx/internal/dpt/DPTUtil.java
```diff
@@ -0,0 +1,127 @@
...
+/**
+ * This class provides support to determine compatibility between KNX DPTs and openHAB data types
+ *
+ * Parts of this code are based on the openHAB KNXCoreTypeMapper by Kai Kreuzer et al.
+ *
+ * @author Jan N. Klug - Initial contribution
+ */
+@NonNullByDefault
+public class DPTUtil {
+    private static final Logger LOGGER = LoggerFactory.getLogger(DPTUtil.class);
```
jlaur, line 51: Loggers should be non-static.

bundles/org.openhab.binding.knx/src/main/java/org/openhab/binding/knx/internal/dpt/ValueDecoder.java
```diff
@@ -0,0 +1,384 @@
...
+/**
+ * This class decodes raw data received from the KNX bus to an openHAB datatype
+ *
+ * Parts of this code are based on the openHAB KNXCoreTypeMapper by Kai Kreuzer et al.
+ *
+ * @author Jan N. Klug - Initial contribution
+ */
+@NonNullByDefault
+public class ValueDecoder {
+    private static final Logger LOGGER = LoggerFactory.getLogger(ValueDecoder.class);
```
jlaur, line 65: Loggers should be non-static.

bundles/org.openhab.binding.knx/src/main/java/org/openhab/binding/knx/internal/dpt/ValueDecoder.java
```diff
@@ -0,0 +1,384 @@
...
+            String id = dptId; // prefer using the user-supplied DPT
+
+            Matcher m = DPTUtil.DPT_PATTERN.matcher(id);
+            if (!m.matches() || m.groupCount() != 2) {
+                LOGGER.trace("User-Supplied DPT '{}' did not match for sub-type, using DPT returned from Translator",
+                        id);
+                id = translator.getType().getID();
+                m = DPTUtil.DPT_PATTERN.matcher(id);
+                if (!m.matches() || m.groupCount() != 2) {
+                    LOGGER.warn("couldn't identify main/sub number in dptID '{}'", id);
```
jlaur, line 103:
```suggestion
                    LOGGER.warn("Couldn't identify main/sub number in dptID '{}'", id);
```

bundles/org.openhab.binding.knx/src/main/java/org/openhab/binding/knx/internal/dpt/ValueDecoder.java
```diff
@@ -0,0 +1,384 @@
...
+             * date of "1970-01-01".
+             * Replace "no-day" with the current day name
+             */
+            StringBuilder stb = new StringBuilder(value);
+            int start = stb.indexOf("no-day");
+            int end = start + "no-day".length();
+            stb.replace(start, end, String.format(Locale.US, "%1$ta", Calendar.getInstance()));
+            value = stb.toString();
+        }
+        Date date = null;
```
jlaur, line 212: Would you consider skipping the usage of the old Java 8 `Date`? See for example https://github.com/openhab/openhab-addons/pull/14170#discussion_r1063545638

Resulting code from that:
https://github.com/openhab/openhab-addons/blob/596324635d89d293ec9d54100dfde43327ab7273/bundles/org.openhab.binding.ecobee/src/main/java/org/openhab/binding/ecobee/internal/dto/thermostat/LocalDateTimeDeserializer.java#L42-L47

and:
https://github.com/openhab/openhab-addons/blob/596324635d89d293ec9d54100dfde43327ab7273/bundles/org.openhab.binding.ecobee/src/main/java/org/openhab/binding/ecobee/internal/dto/thermostat/InstantDeserializer.java#L44-L50

bundles/org.openhab.binding.knx/src/main/java/org/openhab/binding/knx/internal/dpt/ValueDecoder.java
```diff
@@ -0,0 +1,384 @@
...
+                double x = Double.parseDouble(stringx.replace(",", "."));
+                double y = Double.parseDouble(stringy.replace(",", "."));
+                if (stringY == null) {
+                    return HSBType.fromXY((float) x, (float) y);
+                    // return ColorUtil.xyToHsv(new double[] { x, y });
+                } else {
+                    // FIXME core does not support xyY yet
+                    return HSBType.fromXY((float) x, (float) y);
+                    // double Y = Double.parseDouble(stringY.replace(",", "."));
+                    // return ColorUtil.xyToHsv(new double[] { x, y, Y });
```
jlaur, line 316: Comment for remembering about this.

bundles/org.openhab.binding.knx/src/main/java/org/openhab/binding/knx/internal/dpt/ValueEncoder.java
```diff
@@ -0,0 +1,258 @@
...
+/**
+ * This class encodes openHAB data types to strings for sending via Calimero
+ *
+ * Parts of this code are based on the openHAB KNXCoreTypeMapper by Kai Kreuzer et al.
+ *
+ * @author Jan N. Klug - Initial contribution
+ */
+@NonNullByDefault
+public class ValueEncoder {
+    private static final Logger LOGGER = LoggerFactory.getLogger(ValueEncoder.class);
```
jlaur, line 61: Loggers should be non-static.

bundles/org.openhab.binding.knx/src/main/java/org/openhab/binding/knx/internal/dpt/ValueEncoder.java
```diff
@@ -0,0 +1,258 @@
...
+
+    private ValueEncoder() {
+        // prevent instantiation
+    }
+
+    /**
+     * formats the given value as String for outputting via Calimero
```
jlaur, line 68:
```suggestion
     * Formats the given value as String for outputting via Calimero
```

bundles/org.openhab.binding.knx/src/test/java/org/openhab/binding/knx/internal/channel/KNXChannelTest.java
```diff
@@ -0,0 +1,170 @@
...
+import org.openhab.core.thing.Channel;
+import org.openhab.core.thing.type.ChannelTypeUID;
+import org.openhab.core.types.UnDefType;
+
+import tuwien.auto.calimero.GroupAddress;
+import tuwien.auto.calimero.KNXFormatException;
+
+/**
+ *
+ * @author Simon Kaufmann - initial contribution and API.
```
jlaur, line 35:
```suggestion
 * @author Simon Kaufmann - Initial contribution
```

jlaur (changes requested):
Thanks for taking this big backport task upon you. Looks good so far, I have added some very minor comments. Please also remember to regenerate the I18N properties file.

bundles/org.openhab.binding.knx/src/main/java/org/openhab/binding/knx/internal/dpt/ValueDecoder.java
holgerfriedrich: waiting for openhab-core/3434 to be merged....
Does it make sense to catch java.lang.NoSuchMethodError in case someone deploys this to 4.0.0M1?

bundles/org.openhab.binding.knx/src/main/java/org/openhab/binding/knx/internal/channel/GroupAddressConfiguration.java
J-N-K: Not possible because the logger is used in static method `parse`.

bundles/org.openhab.binding.knx/src/main/java/org/openhab/binding/knx/internal/dpt/DPTUtil.java
J-N-K: logger is used in static method

bundles/org.openhab.binding.knx/src/main/java/org/openhab/binding/knx/internal/dpt/ValueDecoder.java
J-N-K: Now don't provide compatibility for such a case and it would be nearly impossible to write workarounds for all of these cases.

bundles/org.openhab.binding.knx/src/main/java/org/openhab/binding/knx/internal/dpt/ValueDecoder.java
holgerfriedrich: Actually I would prefer not to touch this. This is the implementation which is currently in use in the plugin. The test coverage is low, and I am afraid to break something. Maybe this is a cleanup task for later....

holgerfriedrich:
@J-N-K From my perspective, this is finally ready for testing and review.

- I did a cross-check against the version you have in smarthome/j and did not find anything unexpected. A few differences (intended or related to other changes in the master branch) can be found.
- The exception mentioned above is resolved (re-adding % for DPT 5.001 in DPTUnit makes it work in case it is mapped to a Dimmer item).
- Regarding openhab/openhab-core#3434: We could merge without, but then xyY decoding would be broken
- We should do some more testing - I just deployed to my production system and will see if anything goes wrong. But I don't have many -control items in use.....
- update.lst is to be updated to announce the UoM change

florian-h05:
I can help testing this, although I don‘t use much UoM and most of my KNX channels are either DPT 1.xxx or 9.001. Just let me know if you‘d like to have me test this.

J-N-K (approved):
LGTM. I checked especially the DPT conversion and the differences look good to me.

As far as I can see only the refactoring of the manufacturer and device properties is missing now (which is independent from this PR of course).

holgerfriedrich:
@florian-h05 
> I can help testing this, although I don‘t use much UoM and most of my KNX channels are either DPT 1.xxx or 9.001. Just let me know if you‘d like to have me test this.

Thanks! In case you have some -control items, or anything which reads data *from* openHAB, this would be helpful. 
As we are quite close to the smarthome/j implementation, it might also be fine to merge (once the change in core is done).

holgerfriedrich:
@jlaur Thanks for looking into this. I tried to resolve everything besides the Date topic.
@kaikreuzer Any additional comments?

bundles/org.openhab.binding.knx/src/main/resources/OH-INF/config/channelConfig.xml
```diff
@@ -13,7 +13,7 @@
 		</parameter>
 		<parameter name="position" type="text" required="false">
 			<label>Address</label>
-			<description>The group address(es) in Group Address Notation to set the absolute position of the dimmer</description>
+			<description>The group address(es) in Group Address Notation to set the absolute brightness</description>
```
kaikreuzer, line 16: I think the original description was more precise - firstly, because the connected device might be a heater or a speaker (yeah, I am e.g. using those for volume control) and secondly, because "absolute brightness" sounds as if it should be a value in lux or similar.

bundles/org.openhab.binding.knx/src/main/resources/OH-INF/config/channelConfig.xml
```diff
@@ -27,7 +27,7 @@
...
```
kaikreuzer, line 30: see above

kaikreuzer (approved):
Many thanks @holgerfriedrich and @J-N-K for this huge refactoring/contribution,
Code looks very clean to me and I'd suggest I help testing it on my production system as soon as it is merged.
I've only left one very minor wording comment below.

bundles/org.openhab.binding.knx/src/main/java/org/openhab/binding/knx/internal/channel/KNXChannelFactory.java
jlaur: @holgerfriedrich - this is still open. 🙂

jlaur:
> Thanks for looking into this. I tried to resolve everything besides the Date topic.

You're welcome. 🙂 Actually there is still a few open comments which are not resolved, perhaps GitHub has been collapsing and hiding them from you?

bundles/org.openhab.binding.knx/src/main/resources/OH-INF/config/channelConfig.xml
holgerfriedrich: removed from PR, porting smarthomej/addons#257 will override it anyway

bundles/org.openhab.binding.knx/src/main/resources/OH-INF/config/channelConfig.xml
holgerfriedrich: removed from PR, porting https://github.com/smarthomej/addons/pull/257 will override it anyway

bundles/org.openhab.binding.knx/src/main/java/org/openhab/binding/knx/internal/channel/KNXChannel.java
holgerfriedrich: seems not.... It ts this way in almost all classes and the SAT is now nearly empty (just complaints about the static loggers and TODOs)

bundles/org.openhab.binding.knx/src/main/java/org/openhab/binding/knx/internal/dpt/ValueDecoder.java
holgerfriedrich: used in static function....

jlaur:
@holgerfriedrich - LGTM except for https://github.com/openhab/openhab-addons/pull/14534#discussion_r1136177669. FYI, @kaikreuzer has reviewed and merged openhab/openhab-core#3434 less than an hour ago, so can be completed now. 🙂

holgerfriedrich:
@jlaur I just pushed then new version which uses ColorUtil. I think it will only build after the next core snapshot is built tomorrow morning.

jlaur (approved):
LGTM

holgerfriedrich:
Ok, then let's wait for tomorrow. I will rebuild.

Regarding the squash merge of this PR, please take care of the references to the original PRs, please keep @J-N-K in the joint sign-off.
I thought of a rebase commit (as long as we had ~10 commits from the carryover PRs). but with the additional intermediate commits it is too much for the history and probably should be squashed during merge. The single commits can still be found in this ticket.

florian-h05:
@holgerfriedrich & @J-N-K
Huge thanks for your work on the KNX binding, so far works great on my production system. I have checked my (few) -control channels, they still work.

One thing that I noted was that I had a few DPTs wrongly set in my `knx.things` file, I corrected them and the log said `Loading model knx.things` (as usual), but it seems that the thing was not updated. If I check inside MainUI, the old DPT is still displayed. 

I am not sure whether that‘s a core or a KNX binding problem, but when I change my yamaha.things file, the Yamaha devices go OFFLINE and then ONLINE, my KNX devices don‘t (and they did in the past).

J-N-K:
Interesting that this error is back, I'm pretty sure I fixed that one. I'll have a look but I have some other things I'm working on before.

florian-h05:
@J-N-K Any progress regarding this issue?

J-N-K:
Totally forgot about it.

florian-h05:
Unfortunately, this issue persists. The file change was noticed all the time and is still noticed (`[el.core.internal.ModelRepositoryImpl] - Loading model 'knx.things'`), but the KNX devices/Things don't update.

I am not also encountering another weird behaviour: my `-control` channels have stopped working, the `handleCommand` method of `DeviceThingHandler` is not called for control channels.

J-N-K:
I don't normally use things files. I tried and it works for me when I change the channel-name, but not the configuration or the channel label. Can you confirm that this is specific to the KNX binding? I'm asking because when I only change the label, the `updated` method in the thing registry is not called and that should affect all bindings, not only KNX.

florian-h05:
You are right, changing channel name works, but neither channel label nor configuration change work for KNX and MQTT, so it must be unspecific to the KNX binding. And it seems that you already found the cause.

florian-h05:
Stays just the other problem: My `-control` channels have stopped working, they worked with this PR. I have just checked, with SNAPSHOT 3390 from 31.03.2023 5AM they work. So this PR can‘t be the cause, but there were two PRs since then. I will try to find out which of those caused that regression.

J-N-K:
I have no idea what causes this behavior. But it’s a core issue, not an addons-issue. Please open a bug report there.

PR #13570, opened 2022-10-19
[hue] Implement CLIP 2 / API v2

andrewfg:
Resolves #13456 

### Description

This Pull Request adds support for the new Hue CLIP 2 API.

It comprises the following..
1. New CLIP 2 DTO classes
2. New bridge thing handler that supports HTTPS, CLIP 2 end points, and SSE, via HTTP 2.0.
3. New thing handler that can represent any CLIP 2 device (e.g. light, button, sensor, rotary dial, etc.), or grouped light, and has channels dynamically created according to its capabilities.
4. Added support  to the thing handler factory for creating CLIP 2 thing handlers.
5. Added support  to the MDNS discovery service for discovering CLIP 2 bridges.
6. Added thing discovery service to detect CLIP 2 things within a CLIP 2 bridge.
7. Added JUNIT tests, and respective JSON test files, for CLIP 2 DTO classes.

Note: the prior API v1 DTO classes were (sorry to say it) not well structured, and have also accumulated too much 'spaghetti code', so that it was impossible to directly map between API V1 and CLIP 2 things. So this is a 'breaking change' as far as migration of existing API V1 installations to CLIP 2. (This is presumably in line with why Philips Signify also chose to make a breaking change between API V1 and CLIP 2).

Nevertheless it includes all the code from API V1 so it is still backwards compatible for existing API V1 installations that will not be migrated.

Furthermore when new API v2 things are created via the discovery services, then if a legacy API v1 thing exists, the new v2 thing will attempt too clone the attributes of the existing v1 thing. And also, if a legacy API v1 thing exists and has items linked to its channels, then the new v2 thing will replicate the links between those items and the respective new v2 channels.

### JAR Files for OH v4.0

You can install the binding from the Marketplace. The binding requires the UPnP transport feature which is sometimes not automatically installed, so you may need to execute the following command..

```
feature:install openhab-transport-upnp
```

### JAR Files for OH v3.4.x

You can install the binding from the Marketplace. In OH 3.4.x the binding requires the Jetty Http2 features which are not part of openHAB v3.4.x distribution, and the UPnP transport feature which is sometimes not automatically installed. So before installing please i) download the [Jetty-Http2-Feature-Jars](https://github.com/andrewfg/openhab-addons/releases/download/v2/jetty-http2-feature-9.4.46.v20220331.zip), ii) drop them in the `/addons` folder, iii) enter the following commands at the console command prompt, and restart your system.

```
feature:install openhab-transport-upnp
[...]

PR #14538, opened 2023-03-05
Resolve itest runbundles for JUPnP upgrade

wborn:
Related to openhab/openhab-core#3421

### DIFF
--- a/itests/org.openhab.automation.jsscriptingnashorn.tests/itest.bndrun
+++ b/itests/org.openhab.automation.jsscriptingnashorn.tests/itest.bndrun
@@ -59,7 +59,6 @@ Fragment-Host: org.openhab.automation.jsscriptingnashorn
 	org.threeten.extra;version='[1.5.0,1.5.1)',\
 	com.sun.jna;version='[5.12.1,5.12.2)',\
 	io.methvin.directory-watcher;version='[0.17.1,0.17.2)',\
-	org.osgi.service.cm;version='[1.6.0,1.6.1)',\
 	xstream;version='[1.4.19,1.4.20)',\
 	org.apache.felix.configadmin;version='[1.9.26,1.9.27)',\
 	org.apache.felix.http.servlet-api;version='[1.2.0,1.2.1)',\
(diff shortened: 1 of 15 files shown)


hmerk (approved):
LGTM for Wemo binding

kaikreuzer (approved):

andrewfg:
@wborn each time you modify the runbundles you change the `itest.bndrun` for the Hue binding, and this creates a merge conflict with my #13570 PR for Hue Clip 2, which I have to resolve.

The new Hue API v2 requires Jetty HTTP/2 and currently the OH Core only imports Jetty HTTP/1.1 modules. Obviously I would like OH to implement Jetty HTTP/2 directly in the core (see my conversation with @J-N-K in https://github.com/openhab/openhab-core/issues/3317), but this is not yet done. So in the meantime #13570 declares dependencies on the five required Jetty HTTP/2 modules in the Hue binding's respective `pom.xml` and `features.xml` files, and loads them at runtime.

If I resolve the Hue `itest.bndrun` merge conflict by simply accepting your new version of the file, it causes #13570 to fail the integration test part of the CI build process. So I need to merge your modifications with _**some of**_ the new Jetty HTTP/2 dependencies of #13570. But it is not clear to me which of the #13570 HTTP/2 declared dependencies are necessary in the Hue `itest.bndrun` so that the CI build integration tests will succeed. My uncertainty is because the Hue integration tests do not IN FACT REQUIRE the Jetty HTTP/2 modules since the tests currently only validate the Hue API v1 HTTP/1.1 functionality (so far). => So can you please let me know which of the #13570 dependencies REALLY need to be added to `itest.bndrun` so that the CI integration tests will work ??

```
Jetty :: HTTP2 :: Client
Jetty :: HTTP2 :: Common
Jetty :: HTTP2 :: Hpack
Core :: ALPN :: Client
Core :: ALPN :: Java Client
```

wborn:
> REALLY need to be added to itest.bndrun so that the CI integration tests will work

If the tests succeed without adding the bundles you simply don't add them.

If one day you add tests that require the bundles (and they are not automatically resolved by bnd) you add them to `runrequires` so Bnd resolves the right bundle versions in `runbundles`. 

Usually you can also prevent merge conflicts by not adding new `runbundles` at the end of the file.

PR #12726, opened 2022-05-12
[netatmo] Switch to Code Granting process

clinique:
Based on the work done in the spotify binding.
This exposes a second servlet, so I refactor servlet handling in the binding.

Closing issue #12677 

Signed-off-by: clinique <gael@lhopital.org>

PR #14548, opened 2023-03-07
[Netatmo] Modification of the tokenRefresh handling process

### LINKED ISSUE #14546: [Netatmo] oAuth token refresh process evolutions
Authentication update
 
Dear Netatmo developer,

As of today, when you refresh an Access Token using the associated endpoint https://api.netatmo.com/oauth2/token, Netatmo servers respond with a couple of tokens : an Access Token and a Refresh Token.

If the previous Access Token is still valid, the newly returned access token is identical but its expiration time is extended for 3 hours.

In any case, the refresh token is not renewed.

Starting from the 17/04/2023, this behavior will change to to be compliant with the recommendations of the [RFC of the OAuth2 Authorization Framework](https://mailer.netatmo.com/lnk/AUgAAA5uXMAAAcuOqLAAALCkp1AAAAAACjcAnClaAA8COwBkBg6TSZw4ySmDR1K0nWGM_iTkeAAOzw4/1/6uiM2R0W8A6dzVb_hK03tg/aHR0cHM6Ly9ldXIwMS5zYWZlbGlua3MucHJvdGVjdGlvbi5vdXRsb29rLmNvbS8_dXJsPWh0dHBzJTNBJTJGJTJGbWFpbGVyLm5ldGF0bW8uY29tJTJGbG5rJTJGQkFBQUE5WjZ2Q3dBQUFBQUFBQUFBTExqNHBnQUFBQUFDamNBQUFBQUFBOENPd0JqX2NwbFZjUU4wYUE5Ulo2cVNTUm16bmozU2dBT3p3NCUyRjElMkZBSnlRT2Nra1YtdGFVQzNjTDROZUJRJTJGYUhSMGNITTZMeTkzZDNjdWNtWmpMV1ZrYVhSdmNpNXZjbWN2Y21aakwzSm1ZelkzTkRrJmRhdGE9MDUlN0MwMSU3Q3JvbWFpbi5sZS1zY291YXJuZWMlNDBsZWdyYW5kLmNvbSU3Q2NkZjllYTNkN2U5MTRlMTI1NDcxMDhkYjE5NmVkN2VlJTdDMTk5Njg2YjViZWY0NDk2MDg3ODY3YTZiMTg4OGZlZTMlN0MxJTdDMCU3QzYzODEzMTczNjA3Njc0NDc5MyU3Q1Vua25vd24lN0NUV0ZwYkdac2IzZDhleUpXSWpvaU1DNHdMakF3TURBaUxDSlFJam9pVjJsdU16SWlMQ0pCVGlJNklrMWhhV3dpTENKWFZDSTZNbjAlM0QlN0MzMDAwJTdDJTdDJTdDJnNkYXRhPXlGTjdwUUR4d3glMkYwZWlzaU9MSDByN3djMlY3MUs0RG1jeGxudGx0WkhLUSUzRCZyZXNlcnZlZD0w) (section 10.4) and improving the security of the data of our users.

When refreshing tokens, Access Token and Refresh Token will be automatically renewed and former tokens invalidated.

 What does it means for me ?

If you were already updating the tokens provided when refreshing your tokens, this change will not impact you.

If you do not update the refresh token when refreshing your Access Token, your users will be disconnected after 3 hours as the former tokens will become invalidated.

To fix it, you need to update the tokens as soon as you get the newly generated ones.

Sincerely,

Legrand - Netatmo - Bticino

clinique:
Resolves #14546


jlaur:
Thanks, I will give it a test run. Has textual configuration been considered?

clinique:
> Thanks, I will give it a test run. Has textual configuration been considered?

You're correct. Textual configuration could be a real challenge because we can't override the stored configuration - so I think I should move the refreshToken to properties instead of configuration.

jlaur:
> > Thanks, I will give it a test run. Has textual configuration been considered?
> 
> You're correct. Textual configuration could be a real challenge because we can't override the stored configuration - so I think I should move the refreshToken to properties instead of configuration.

Can you always reauthenticate from clientId and clientSecret to get new access and refresh token? I'm wondering if we even need a property. I guess the advantage is that we don't need to get new token when restarting openHAB/binding, but besides that we could just keep the refresh token in memory?

lolodomo:
> You're correct. Textual configuration could be a real challenge because we can't override the stored configuration - so I think I should move the refreshToken to properties instead of configuration.

That would mean you have to re-authenticate each time you restart OH !
If I correctly remember, you are not using the oAuth2 client from OH core for this binding ? Do you remember what and do you think the original reason is still valid ?
Because this would be the solution as OH is then storing the necessary data in file userdata/json/StorageHandler.For.OAuthClientService.json

lolodomo:
The best solution would be really to use the oAuth2 client implementation from the core framework.
In case you really can't, I would suggest that you store your token(s) in your own file somewhere in userdata. The data will then be persisted and this will work whatever the way you define your thing.

clinique:
Yes, there is a valid reason why I do not use the core oAuth mechanism. Netatmo is not complying to oAuth standard.
I opened an issue a while ago that has never been adressed [here](https://forum.netatmo.com/viewtopic.php?t=19293).

I tried to adress it from a core perspective [here](https://github.com/openhab/openhab-core/issues/1888) and we discussed this [here](https://github.com/openhab/openhab-addons/pull/12726)

lolodomo:
If you can't use the OH oAuth2 client, I would then like that you store your token data in a file in folder userdata/netatmo.

clinique:
> If you can't use the OH oAuth2 client, I would then like that you store your token data in a file in folder userdata/netatmo.

Ok but why not in properties ?

lolodomo:
Because thing properties are lost when you restart OH when your thing is defined in a things file. 

If you need to persist something (a token( in a file to be re-used after a OH restart or upgrade, do not consider the thing properties option.

lolodomo (approved):
LGTM, thank you

clinique:
I have resolved the conflict directly online. This conflict is weird, but git may have some mysteries.

clinique:
> > I have resolved the conflict directly online. This conflict is weird, but git may have some mysteries.
> 
> Conflict still mentioned on file ApiBridgeHandler.java.

Understood why in the meantime, it's due to the merge of CO detector, I need to rebate and force push

jlaur:
> Understood why in the meantime, it's due to the merge of CO detector, I need to rebase and force push

If these two PR's conflict, we cannot cherry-pick into the 3.4.x branch, but will instead have to create a new branch tailored for that, i.e. at PR targeting 3.4.x.

clinique:
@jlaur , @lolodomo : rebased and conflict fixed.

lolodomo:
@clinique: tests are OK for me.

lolodomo:
This change also simplified the process for users using config files.

I am asking myself if we should document where are stored the token files ?

lolodomo:
And thank you @clinique for providing that change so quickly.
The delay forced by netatmo is very short !

clinique:
You're welcome. I'll add a note in the README

lolodomo:
Just to understand, does it mean that if you stop your OH server more than 3 hours, you will have to go through the authorization process when restarting OH ?

clinique:
Hopefully not - I wondered the same reading the oAuth documentation. I hope they refresh with the last provided token, or else, it is the end of this binding. It's a bit hard to predict because the new refreshing refresh token is not in place currently.

clinique:
I had this fear and you reactivated it. I opened a question on the Netatmo [forum](https://forum.netatmo.com/viewtopic.php?t=21239)

lolodomo:
I assume you are storing the updated refresh token in the new file.
Re-reading the sentence, I understand the old refresh token will be invalidated only when you try to refresh. So if API is not requested even during several days, there should be no invalidation.

clinique:
You assume correctly, and I read it also the way you do.

jlaur:
@clinique, @lolodomo - I assume we now need a separate PR targeting 3.4.x, since the merged branch seemed to be overlapping with other changes for 4.0?

clinique:
Yes, a distinct PR has to be created for 3.4.x - I will try to find time to produce it.

PR #14549, opened 2023-03-07
[bluetooth.radoneye] Support for RadonEye with v2.x.x firmware

joerg1985:
This PR will add support for the v2.x.x firmware to the RadonEye binding, by using other UUIDs for this version.

An additional change is the removing the `Optional<>` from the configuration.
Without a configuration the binding is missing the needed address to connect to the device.
Therefore the configuration is mandatory and not an optional field.

### DIFF
--- a/bundles/org.openhab.binding.bluetooth.radoneye/README.md
+++ b/bundles/org.openhab.binding.bluetooth.radoneye/README.md
@@ -18,10 +18,11 @@ As any other Bluetooth device, RadonEye devices are discovered automatically by 
 
 Supported configuration parameters for the things:
 
-| Property                        | Type    | Default | Required | Description                                                     |
-|---------------------------------|---------|---------|----------|-----------------------------------------------------------------|
-| address                         | String  |         | Yes      | Bluetooth address of the device (in format "XX:XX:XX:XX:XX:XX") |
-| refreshInterval                 | Integer | 300     | No       | How often a refresh shall occur in seconds                      |
+| Property        | Type    | Default | Required | Description                                                     |
+|-----------------|---------|---------|----------|-----------------------------------------------------------------|
+| address         | String  |         | Yes      | Bluetooth address of the device (in format "XX:XX:XX:XX:XX:XX") |
+| fwVersion       | Integer | 1       | No       | The major version of the firmware on the device                 |
+| refreshInterval | Integer | 300     | No       | How often a refresh shall occur in seconds                      |
 
 ## Channels
 
--- a/bundles/org.openhab.binding.bluetooth.radoneye/src/main/java/org/openhab/binding/bluetooth/radoneye/internal/AbstractRadoneyeHandler.java
+++ b/bundles/org.openhab.binding.bluetooth.radoneye/src/main/java/org/openhab/binding/bluetooth/radoneye/internal/AbstractRadoneyeHandler.java
@@ -12,7 +12,6 @@
  */
 package org.openhab.binding.bluetooth.radoneye.internal;
 
-import java.util.Optional;
 import java.util.UUID;
 import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.TimeUnit;
@@ -45,10 +44,9 @@ abstract public class AbstractRadoneyeHandler extends BeaconBluetoothHandler {
     private final Logger logger = LoggerFactory.getLogger(AbstractRadoneyeHandler.class);
 
     private AtomicInteger sinceLastReadSec = new AtomicInteger();
-    private Optional<RadoneyeConfiguration> configuration = Optional.empty();
+    private RadoneyeConfiguration configuration = new RadoneyeConfiguration();
     private @Nullable ScheduledFuture<?> scheduledTask;
 
-    private volatile int refreshInterval;
     private volatile int errorConnectCounter;
     private volatile int errorReadCounter;
     private volatile int errorWriteCounter;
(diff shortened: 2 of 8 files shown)


bundles/org.openhab.binding.bluetooth.radoneye/src/main/java/org/openhab/binding/bluetooth/radoneye/internal/AbstractRadoneyeHandler.java
```diff
@@ -45,10 +44,9 @@ abstract public class AbstractRadoneyeHandler extends BeaconBluetoothHandler {
     private final Logger logger = LoggerFactory.getLogger(AbstractRadoneyeHandler.class);
 
     private AtomicInteger sinceLastReadSec = new AtomicInteger();
-    private Optional<RadoneyeConfiguration> configuration = Optional.empty();
+    private @Nullable RadoneyeConfiguration configuration;
```
lolodomo, line 47: To avoid null check warnings, please define configuration like that:
```
private RadoneyeConfiguration configuration = new RadoneyeConfiguration();
```

bundles/org.openhab.binding.bluetooth.radoneye/src/main/java/org/openhab/binding/bluetooth/radoneye/internal/AbstractRadoneyeHandler.java
```diff
@@ -293,7 +289,16 @@ public void onConnectionStateChange(BluetoothConnectionStatusNotification connec
...
-        return sinceLastRead >= refreshInterval;
+        return sinceLastRead >= configuration.refreshInterval;
+    }
+
+    /**
+     * Provides the configuration of this instance
+     *
+     * @return the configuration of this instance
+     */
+    protected @Nullable RadoneyeConfiguration getConfiguration() {
```
lolodomo, line 300: With the change I proposed below, you can then remove `@Nullable` here.

Considering how this method is used, I would suggest to replace it by a method getFwVersion.

bundles/org.openhab.binding.bluetooth.radoneye/src/main/java/org/openhab/binding/bluetooth/radoneye/internal/RadoneyeConfiguration.java
```diff
@@ -22,10 +22,11 @@
 @NonNullByDefault
 public class RadoneyeConfiguration {
     public String address = "";
+    public int fwVersion;
     public int refreshInterval;
```
lolodomo, line 26: You could align thje default value to the one from your XML definition:
```
    public int fwVersion = 1;
    public int refreshInterval = 300;
```

bundles/org.openhab.binding.bluetooth.radoneye/src/main/resources/OH-INF/thing/radoneye.xml
```diff
@@ -24,6 +24,11 @@
 				<label>Address</label>
 				<description>Bluetooth address in XX:XX:XX:XX:XX:XX format</description>
 			</parameter>
+			<parameter name="fwVersion" type="integer" min="1" max="2">
```
lolodomo, line 27: Please also update the default translation file to add an entries for this new parameter.

bundles/org.openhab.binding.bluetooth.radoneye/README.md
```diff
@@ -18,10 +18,11 @@ As any other Bluetooth device, RadonEye devices are discovered automatically by
...
 Supported configuration parameters for the things:
 
-| Property                        | Type    | Default | Required | Description                                                     |
-|---------------------------------|---------|---------|----------|-----------------------------------------------------------------|
-| address                         | String  |         | Yes      | Bluetooth address of the device (in format "XX:XX:XX:XX:XX:XX") |
-| refreshInterval                 | Integer | 300     | No       | How often a refresh shall occur in seconds                      |
+| Property        | Type    | Default | Required | Description                                                     |
+|-----------------|---------|---------|----------|-----------------------------------------------------------------|
+| address         | String  |         | Yes      | Bluetooth address of the device (in format "XX:XX:XX:XX:XX:XX") |
+| fwVersion       | Integer | 1       | Yes      | The major version of the firmware on the device                 |
```
lolodomo, line 24: Required can be set to "No" as you provide a default value.

bundles/org.openhab.binding.bluetooth.radoneye/src/main/java/org/openhab/binding/bluetooth/radoneye/internal/RadoneyeHandler.java
```diff
@@ -65,16 +66,40 @@ protected void updateChannels(int[] is) {
...
+            default:
+                throw new UnsupportedOperationException("fwVersion: " + fwVersion + " is not implemented");
+        }
     }
 
     @Override
     protected UUID getTriggerUUID() {
-        return triggerUuid;
+        int fwVersion = getConfiguration().fwVersion;
+        switch (getConfiguration().fwVersion) {
```
lolodomo, line 83: Just use `fwVersion`

lolodomo (changes requested):

bundles/org.openhab.binding.bluetooth.radoneye/src/main/resources/OH-INF/thing/radoneye.xml
joerg1985: There was no translations file, so i added them. I hope this was correct, could you please have a look at the new `bluetooth.properties` and the `bluetooth_de.properties` files?

lolodomo: Please avoid providing translations for different languages, it should be rather done with Crowdin. As the files do not yet exist, if you really want to provide German translations, this is acceptable but a manual action will be required by a Crowdin maintainer to upload them in Crowdin.

You only provided a partial list of entries. Please run "mvn i18n:generate-default-translations" so that your file is generated automatically. This binding is "sub-binding" of bluetooth so maybe the tool will generate a propoerties file with a bad name. Considering the other bluetooth "sub-bindings", I believe you make the good choice with file named bluetooth.properties.

jlaur: > As the files do not yet exist, if you really want to provide German translations, this is acceptable but a manual action will be required by a Crowdin maintainer to upload them in Crowdin.

This would then skip the proof-reading of texts which is also a part of the Crowdin process.

joerg1985: The 'bluetooth.properties' file is updated and the 'bluetooth_de.properties' file is removed, thanks to hint with the `mvn i18n:generate-default-translations`.

I have removed the 'This could have impact to battery lifetime' text, the device has no battery.

lolodomo (approved):
LGTM, thank you

PR #14553, opened 2023-03-08
[androiddebugbridge] Update docs of androiddebugbridge

mueller-ma:
# Title

Update docs of androiddebugbridge

# Description

* Use code fence for intents
* Fix Item type of awake-state
* Pretty print JSON and add Netflix

# Testing

Not tested, as no code was changed.


wborn (approved):
👍

PR #14558, opened 2023-03-09
[enocean] Bugfix - condition in getPositionData function of EEP D2_05_00

### LINKED ISSUE #14528: [enocean] D2-05-00 getPositionData(): condition incomlete?
## Expected Behavior
EnOcean Profile D2-05-00 specifies for answer on position request (“CMD 4 Reply Position and Angle”) for bit with offset 0 (size 1) that it is not used and should be “0”.
Actually I also get 0xFF as value for first byte, i.e. “not used bit” has value “1”.

To be more robust, i.e. the "not used" bit should not be evaluated.

## Current Behavior
In D2_05_00.java I found following implementation for getPositionData():

```
protected State getPositionData() {
   if (getCMD() == CMD_ACTUATOR_POSITION_RESPONE) {
      if (bytes[0] != 127) {
           return new PercentType(bytes[0] & 0x7f);
       }
   }
```
## Possible Solution

`if ((bytes[0] & 0x7f) != 127) {`

lovery:
Fix #14528

Change condition in function getPositionData in EEP D2_05_00 to check data at offset 1 and size 7 instead offset 0 and size 8.

According the enocean documentation visible [here](http://tools.enocean-alliance.org/EEPViewer/profiles/D2/05/00/D2-05-00.pdf), the section `CMD 4 - Reply Position and Angle` the first bit is not used, so the validation should not evaluate it.

### DIFF
--- a/bundles/org.openhab.binding.enocean/src/main/java/org/openhab/binding/enocean/internal/eep/D2_05/D2_05_00.java
+++ b/bundles/org.openhab.binding.enocean/src/main/java/org/openhab/binding/enocean/internal/eep/D2_05/D2_05_00.java
@@ -84,8 +84,9 @@ public class D2_05_00 extends _VLDMessage {
 
     protected State getPositionData() {
         if (getCMD() == CMD_ACTUATOR_POSITION_RESPONE) {
-            if (bytes[0] != 127) {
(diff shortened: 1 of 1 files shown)


jlaur (approved):
Thanks for the fix!

PR #14559, opened 2023-03-09
[transform.rollershutterposition] Add missing addon.xml

jlaur:
Add missing addon.xml from #13259.


jlaur:
@kaikreuzer - please consider this for the M1 build. I'm not sure about consequence of this file missing.

PR #14561, opened 2023-03-09
[voicerss] Add LRU cache

lolodomo:

### DIFF
--- a/bundles/org.openhab.voice.voicerss/README.md
+++ b/bundles/org.openhab.voice.voicerss/README.md
@@ -166,9 +166,10 @@ It supports the following audio formats: MP3, OGG, AAC and WAV.
 
 ## Caching
 
-The VoiceRSS extension does cache audio files from previous requests, to reduce traffic, improve performance, reduce number of requests and provide same time offline capability.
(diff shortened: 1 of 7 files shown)


lolodomo:
@dalgwen : here is a first implementation.
It is working well when the cache is enabled. But when I disable the cache, I see that my method `synthesizeForCache` is called but then I have no rendering (no sound). I am using the command `say`. I suppose there is a bug in the code of the cache when the cache is disabled.

lolodomo:
For the review, `VoiceRSSFileAudioStream` is not a new class but the previous class named `VoiceRSSAudioStream` and I added/defined a different class `VoiceRSSAudioStream`.

J-N-K:
What is the state here?

lolodomo:
This PR is now finished and properly tested. But it is preferable to first wait for an update of all audio sinks. If not, the case where the TTS cache is disabled will probably not work with most of audio sinks.

lolodomo:
To be updated when openhab/openhab-core#3675 is merged.

lolodomo:
This PR is now ready for a review. It has been fully tested with Sonos as audio sink, either with cache enabled or disabled.

kaikreuzer (approved):
lgtm, thanks!

lolodomo:
Excellent, thank you.

PR #14563, opened 2023-03-09
Ignore HTTP errors != 400 for 2-FA authentication

maihacke:
This PR should fix https://github.com/openhab/openhab-addons/issues/14323

It changes error handling for iCloud 2-fa authentication to ignore all http error except 400.

### DIFF
--- a/bundles/org.openhab.binding.icloud/src/main/java/org/openhab/binding/icloud/internal/ICloudService.java
+++ b/bundles/org.openhab.binding.icloud/src/main/java/org/openhab/binding/icloud/internal/ICloudService.java
@@ -246,8 +246,16 @@ public class ICloudService {
             this.session.post(AUTH_ENDPOINT + "/verify/trusteddevice/securitycode", JsonUtils.toJson(requestBody),
                     headers);
         } catch (ICloudApiResponseException ex) {
-            logger.debug("Code verification failed.", ex);
-            return false;
+            logger.debug("Exception on code verification with HTTP status {}. Verification might still be successful.",
+                    ex.getStatusCode());
+            logger.trace("Exception on code verification with HTTP status {}. Verification might still be successful.",
+                    ex.getStatusCode(), ex);
+            // iCloud API returns different 4xx error codes even if validation is successful
+            // currently 400 seems to show that verification "really" failed.
+            if (ex.getStatusCode() == 400 || ex.getStatusCode() >= 500) {
+                this.logger.debug("Verification failed with HTTP status {}.", ex.getStatusCode());
+                return false;
+            }
         }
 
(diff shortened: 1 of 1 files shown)


bundles/org.openhab.binding.icloud/src/main/java/org/openhab/binding/icloud/internal/ICloudService.java
```diff
@@ -246,8 +246,16 @@ public boolean validate2faCode(String code) throws IOException, InterruptedExcep
             this.session.post(AUTH_ENDPOINT + "/verify/trusteddevice/securitycode", JsonUtils.toJson(requestBody),
                     headers);
         } catch (ICloudApiResponseException ex) {
-            logger.debug("Code verification failed.", ex);
-            return false;
+            logger.debug("Exception on code verification with HTTP status {}. Verification might still be successful.",
+                    ex.getStatusCode());
+            logger.trace("Exception on code verification with HTTP status {}. Verification might still be successful.",
+                    ex.getStatusCode(), ex);
```
lolodomo, line 252: I am sorry to request so many changes for something as simple, but your code will log twice when TRACE level is set.
Rather something like:
```
        if (logger.isTraceEnabled()) {
            logger.trace("Exception on code verification with HTTP status {}. Verification might still be successful.",
                    ex.getStatusCode(), ex);
        } else {
            logger.debug("Exception on code verification with HTTP status {}. Verification might still be successful.",
                    ex.getStatusCode());
        }

lolodomo (changes requested):

bundles/org.openhab.binding.icloud/src/main/java/org/openhab/binding/icloud/internal/ICloudService.java
maihacke: No offense, but as already said I would go with a simple debug log which also logs the stack trace. In my opinion this here is overkill.

maihacke: If you really like the complexity, I could do it. Please decide.

lolodomo: Ok, keep only the trace version, please.

maihacke: 👍

lolodomo (approved):
LGTM now

lolodomo:
@jlaur : this is a candidate for branch 3.4.x.

PR #14564, opened 2023-03-09
[mimictts] Add LRU cache

dalgwen:
Hello,

This pull request add the common TTS cache capability to the mimic TTS Service. (MimicTTSService.java)

And also : [mimictts] Fix playing with audio servlet
It also fixes an issue with a sound not playing with servlet based sink. (AutoDeleteFileAudioStream.java)
(The system responsible for deleting a file that is not used anymore was too greedy. The sound file was deleted, detecting a closing stream, before even being read for "real")


lolodomo:
Maybe your class AutoDeleteFileAudioStream could be moved into core framework to be used by different add-ons?

dalgwen:
Hum yes, but is it really a good implementation for core ?
I designed it as a workaround (temporary), and it is not very bug proof. If the stream is not read by the sink, for whatever reason, then it won't even be deleted.

The more I think about it, the more I think it should be handled in the `AudioHTTPServer` implementation.
Instead of :
    `String serve(FixedLengthAudioStream stream, int seconds);`
The method could be :
    `String serve(AudioStream stream, int seconds);`

And then, in the `AudioServlet` implementation, the `serve (AudioStream stream, int second)` method implementation could
- if stream is FixedAudioStream, then as proceed usual (putting the stream in a multi time enabled map for further reference)
- if not, then using a temporary cache file, a wrapper, or whatever magic, and still putting it in the multi time enabled map.

Then we won't have to do workarounds in all TTS services for the audio sinks that only support this multi time `serve (FixedLengthAudioStream stream, int second)` method.

WDYT ?

lolodomo:
Looks like a great idea.

lolodomo:
@dalgwen : do you propose to implement your idea (`AudioHTTPServer`) ?

lolodomo:
@dalgwen : I have another idea. `synthesizeForCache` is called by the cache handler in two cases:

1. When the cache is enabled and a new TTS has to be added in the cache
2. When the cache is disabled

Our problem is only with the second case. Why not distinguishing the two cases and add a new method `synthesizeWithoutCache` in `AbstractCachedTTSService`. For this method, the TTS service would know that it has to create a file and return a FileAudioStream.
mimic and voiceRRS would then build a temporary file when this method is called.

WDYT ?

dalgwen:
> do you propose to implement your idea (`AudioHTTPServer`) ?

I made a try : 
https://github.com/dalgwen/openhab-core/commit/9b95051bd4d3ed549ad1b58f4a910872590e63ee

With this, the servlet and the audiosinks using it do not need to restrain to FixedLengthAudioStream.
The servlet take a copy in memory if it is small (< 1MB, to avoid disk write operation), or in a temporary file (should be very rare because 1MB is, I think, larger than the majority of use cases). It then serves the copy. And the copy can be cloned and used several times.

I still have one weird behaviour : the code to delete the file when expired is called, but the file remains on disk.

lolodomo:
Looks like a good solution? I hope you can fix your weird behaviour with file not deleted.

dalgwen:
I did a PR(https://github.com/openhab/openhab-core/pull/3461) for what we discuss here.

In the meantime, I put this PR on "draft" mode

lolodomo:
@dalgwen : is it ready for a review ?

dalgwen:
> @dalgwen : is it ready for a review ?

I just force pushed after a rebase.

As this PR removes the workaround for the HTTP audio servlet, it will break compatibility with audio sink using audio servlet and still testing against FixedLengthAudioStream.
Apart of this, it is ready.

lolodomo (changes requested):

lolodomo:
> As this PR removes the workaround for the HTTP audio servlet, it will break compatibility with audio sink using audio servlet and still testing against FixedLengthAudioStream.
> Apart of this, it is ready.

Please unmark the PR as draft in this case.
I will merge it as soon as most of audio sinks are changed.

bundles/org.openhab.voice.mimictts/src/main/java/org/openhab/voice/mimic/internal/MimicTTSService.java
```diff
@@ -325,16 +302,16 @@ public AudioStream synthesize(String text, Voice voice, AudioFormat requestedFor
         }
     }
 
-    private AudioStream createTemporaryFile(InputStream inputStream, AudioFormat audioFormat) throws TTSException {
-        File mimicDirectory = new File(OpenHAB.getUserDataFolder(), "mimic");
-        mimicDirectory.mkdir();
+    @Override
+    public String getCacheKey(String text, Voice voice, AudioFormat requestedFormat) {
```
lolodomo, line 306: Interesting. I realize that I have to do something similar in VoiceRSS.

bundles/org.openhab.voice.mimictts/src/main/resources/OH-INF/config/config.xml
```diff
@@ -11,12 +11,6 @@
 			<description>Mimic 3 URL.</description>
 			<default>http://localhost:59125</default>
 		</parameter>
-		<parameter name="workaroundServletSink" type="boolean" required="false">
-			<label>Workaround For Servlet-Based Audiosink</label>
-			<description>Enable this workaround to store temporarily the file on disk. Needed if you play on audiosink based on
-				the openHAB audio servlet.</description>
-			<default>false</default>
-		</parameter>
```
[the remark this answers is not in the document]
dalgwen: Good catch. thanks. Done

bundles/org.openhab.voice.mimictts/src/main/java/org/openhab/voice/mimic/internal/MimicTTSService.java
dalgwen: The default cache key generator is sufficient for most of the case. But for some configuration particularities (pitch, speed, etc.) it could indeed be extended.

bundles/org.openhab.voice.mimictts/src/main/resources/OH-INF/config/config.xml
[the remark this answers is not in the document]
lolodomo: You did not push any new commit

bundles/org.openhab.voice.mimictts/src/main/java/org/openhab/voice/mimic/internal/MimicTTSService.java
lolodomo: Yes, you're right, the default key generation is fine for VoiceRSS.

bundles/org.openhab.voice.mimictts/src/main/java/org/openhab/voice/mimic/internal/MimicTTSService.java
```diff
@@ -325,16 +302,16 @@ public AudioStream synthesize(String text, Voice voice, AudioFormat requestedFor
...
         try {
-            File tempFile = File.createTempFile(UUID.randomUUID().toString(), ".wav", mimicDirectory);
-            tempFile.deleteOnExit();
-            Files.copy(inputStream, tempFile.toPath(), StandardCopyOption.REPLACE_EXISTING);
-            return new AutoDeleteFileAudioStream(tempFile, audioFormat);
-        } catch (AudioException | IOException e) {
-            throw new TTSException("Cannot create temporary audio file", e);
+            md = MessageDigest.getInstance("MD5");
+        } catch (NoSuchAlgorithmException e) {
+            return "nomd5algorithm";
```
lolodomo, line 310: Maybe you could rather call the super class in case of exception ?

bundles/org.openhab.voice.mimictts/src/main/java/org/openhab/voice/mimic/internal/MimicTTSService.java
```diff
@@ -325,16 +302,16 @@ public AudioStream synthesize(String text, Voice voice, AudioFormat requestedFor
...
         }
+        byte[] binaryKey = ((text + voice.getUID() + requestedFormat.toString() + config.speakingRate
+                + config.audioVolatility + config.phonemeVolatility).getBytes());
+        return String.format("%032x", new BigInteger(1, md.digest(binaryKey)));
```
lolodomo, line 314: In GoogleTTS, I saw that you excluded the voice from the hash, probably to have file names not containing only a hash ?
Would you like to do the same here ?

bundles/org.openhab.voice.mimictts/src/main/resources/OH-INF/config/config.xml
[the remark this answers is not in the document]
dalgwen: Third time... Sorry. I hope it will "sink" in 😉

bundles/org.openhab.voice.mimictts/src/main/java/org/openhab/voice/mimic/internal/MimicTTSService.java
dalgwen: The super class will fail in the same way (it also uses md5)

bundles/org.openhab.voice.mimictts/src/main/java/org/openhab/voice/mimic/internal/MimicTTSService.java
dalgwen: I did this in google tts this because the old custom google tts cache made the same thing.
Here, I'm not this confident, as I don't know if the voice name will always be "filename friendly".

bundles/org.openhab.voice.mimictts/src/main/java/org/openhab/voice/mimic/internal/MimicTTSService.java
lolodomo: Ok

bundles/org.openhab.voice.mimictts/src/main/java/org/openhab/voice/mimic/internal/MimicTTSService.java
lolodomo: Ok

lolodomo (approved):
LGTM, thank you

lolodomo:
I will wait a little (maybe end of weekend), I hope someone will merge my changes in audio sinks.
But this will be included in Milestone 5.

PR #14565, opened 2023-03-10
[picotts] Add LRU cache

dalgwen:
This pull request make use of the [new TTS LRU cache](https://github.com/openhab/openhab-core/pull/3057) by applying it to the Pico TTS service.

### DIFF
--- a/bundles/org.openhab.voice.picotts/src/main/java/org/openhab/voice/picotts/internal/PicoTTSService.java
+++ b/bundles/org.openhab.voice.picotts/src/main/java/org/openhab/voice/picotts/internal/PicoTTSService.java
@@ -35,9 +35,9 @@ import org.osgi.service.component.annotations.Reference;
 /**
  * @author Florian Schmidt - Initial Contribution
  */
-@Component
+@Component(service = TTSService.class)
 @NonNullByDefault
-public class PicoTTSService extends AbstractCachedTTSService implements TTSService {
+public class PicoTTSService extends AbstractCachedTTSService {
 
     @Activate
     public PicoTTSService(@Reference TTSCache ttsCache) {
--- a/bundles/org.openhab.voice.picotts/src/main/java/org/openhab/voice/picotts/internal/PicoTTSService.java
+++ b/bundles/org.openhab.voice.picotts/src/main/java/org/openhab/voice/picotts/internal/PicoTTSService.java
@@ -23,17 +23,27 @@ import org.eclipse.jdt.annotation.Nullable;
 import org.openhab.core.audio.AudioException;
 import org.openhab.core.audio.AudioFormat;
 import org.openhab.core.audio.AudioStream;
+import org.openhab.core.voice.AbstractCachedTTSService;
+import org.openhab.core.voice.TTSCache;
 import org.openhab.core.voice.TTSException;
 import org.openhab.core.voice.TTSService;
 import org.openhab.core.voice.Voice;
+import org.osgi.service.component.annotations.Activate;
 import org.osgi.service.component.annotations.Component;
+import org.osgi.service.component.annotations.Reference;
 
 /**
  * @author Florian Schmidt - Initial Contribution
  */
 @Component
 @NonNullByDefault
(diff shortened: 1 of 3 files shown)


bundles/org.openhab.voice.picotts/src/main/java/org/openhab/voice/picotts/internal/PicoTTSService.java
```diff
@@ -23,17 +23,27 @@
...
 import org.osgi.service.component.annotations.Component;
+import org.osgi.service.component.annotations.Reference;
 
 /**
  * @author Florian Schmidt - Initial Contribution
  */
 @Component
 @NonNullByDefault
-public class PicoTTSService implements TTSService {
+public class PicoTTSService extends AbstractCachedTTSService implements TTSService {
```
lolodomo, line 40: `AbstractCachedTTSService` already implements `TTSService `.
Another option is to rather use 
```
@Component(service = TTSService.class)
```

lolodomo (commented):
Just one comment.

lolodomo:
No risk of problem with the audio servlet for this one ?

lolodomo:
> No risk of problem with the audio servlet for this one ?

I answer to myself: this TTS service always create a (temporary) file, so no risk.

lolodomo:
picotts (like mactts) is creating a temporary file.
I think we could add a method in `AbstractCachedTTSService` that would be called by the cache mechanism when the cache  has created its own file from the returned AudioStream. That method will inform the TTS service that it can release/delete its file. This will avoid having audio files duplicated, one as a temporary file by the TTS service and one in the TTS cache.

WDYT ?

dalgwen:
Yes, indeed, it is not clean to keep several files.
If I had to do this, I would create an interface Releasable (or something like this), with a method release() inside.
The FileAudioStream, PicoTTSAudioStream, etc, could then implements it.

And inside LRUMediaCacheEntry, in the read method (line 281, which is the only place where we could say with certainty that the original InputStream is no longer needed), we could check if the InputStream implements the Releasable interface and in this case call the release method (which, in our case, is a deletion).

lolodomo:
We could even make it simpler. Just add a method `release` to `FileAudioStream`. That would also require to fix the picotts and mactts so that their audio stream class extend FileAudioStream instead of FixedLengthAudioStream. Would it be not enough "clean" to do that ?

lolodomo:
In the addons database, the only bindings really using FileAudioStream (other than for its public constants) are 3 TTS services: mimic, voicerss and polytts.
So if we change `FileAudioStream`, it will be easy to check the possible impact on the existing addons suing it.

lolodomo:
Regarding the file deletion, I believe we should let each TTS service decide what to do with the files they produced.
So I would really prefer a new method in `AbstractCachedTTSService` that will be called by the TTS cxache handler.

lolodomo:
I now realized that the place where it should be called is in LRUMediaCacheEntry and we don't have at this place an access to the `AbstractCachedTTSService`.

> If I had to do this, I would create an interface Releasable (or something like this), with a method release() inside.
> The FileAudioStream, PicoTTSAudioStream, etc, could then implements it.

Yes, I think now that that you're right.

lolodomo:
@dalgwen : I believe you just have to add the `dispose` method like I did in #15109

lolodomo:
> @dalgwen : I believe you just have to add the `dispose` method like I did in #15109

Can you add that and look at my comment ?
Then I will approve and merge your PR.

lolodomo:
@dalgwen : no chance to finish? 
If you prefer, I can create an alternative PR.

dalgwen:
@lolodomo 
Excuse me I forgot this, was thinking about finishing sinks first. I will check this this today or tomorrow.

dalgwen:
dispose and readme done

lolodomo (approved):
LGTM, thank you

lolodomo:
I merged this PR immediately because a `FixedLengthAudioStream` is used by this TTS service, so we do not depend on any enhancement of audio sinks.

PR #14571, opened 2023-03-10
[shelly] Remove unused headings

felixschndr:
Remove the unused headings for generation 2. There is probably a manual missing for gen 2, but while this is the case the headlines can be removed

Signed-off-by: Felix Schneider <fs@felix-schneider.org>


jlaur (approved):
LGTM

PR #14573, opened 2023-03-11
[tibber] Query for subscription url every reconnect attempt

kjoglum:
Enhancement to allow binding to check for potential change in subscription URL at every reconnect attempt, not just at initial startup.

### DIFF
--- a/bundles/org.openhab.binding.tibber/src/main/java/org/openhab/binding/tibber/internal/handler/TibberHandler.java
+++ b/bundles/org.openhab.binding.tibber/src/main/java/org/openhab/binding/tibber/internal/handler/TibberHandler.java
@@ -146,28 +146,14 @@ public class TibberHandler extends BaseThingHandler {
 
                 if ("true".equals(rtEnabled)) {
                     logger.debug("Pulse associated with HomeId: Live stream will be started");
+                    getSubscriptionUrl();
 
-                    InputStream wsURL = tibberQuery.getWebsocketUrl();
-                    String wsResponse = HttpUtil.executeUrl("POST", BASE_URL, httpHeader, wsURL, null, REQUEST_TIMEOUT);
-
-                    JsonObject wsobject = (JsonObject) JsonParser.parseString(wsResponse);
-                    JsonObject dataObject = wsobject.getAsJsonObject("data");
-                    if (dataObject != null) {
-                        JsonObject viewerObject = dataObject.getAsJsonObject("viewer");
-                        if (viewerObject != null) {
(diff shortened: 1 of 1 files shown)


bundles/org.openhab.binding.tibber/src/main/java/org/openhab/binding/tibber/internal/handler/TibberHandler.java
```diff
@@ -297,7 +297,31 @@ public void updateRequest() throws IOException {
...
+            if (dataObject != null) {
+                JsonObject viewerObject = dataObject.getAsJsonObject("viewer");
+                if (viewerObject != null) {
+                    JsonElement subscriptionElement = viewerObject.get("websocketSubscriptionUrl");
+                    if (subscriptionElement != null) {
+                        subscriptionURL = subscriptionElement.toString().replaceAll("^\"|\"$", "");
+                    }
+                }
+            }
+            String url = subscriptionURL;
```
jlaur, line 316: This code is duplicated from line 150-164. Could it be extracted to a private method?

jlaur (changes requested):
Thanks for the enhancement!

bundles/org.openhab.binding.tibber/src/main/java/org/openhab/binding/tibber/internal/handler/TibberHandler.java
kjoglum: The common part of requesting subscription url is now combined into a common method.

bundles/org.openhab.binding.tibber/src/main/java/org/openhab/binding/tibber/internal/handler/TibberHandler.java
```diff
@@ -297,7 +283,34 @@ public void updateRequest() throws IOException {
...
+                updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.COMMUNICATION_ERROR,
+                        "Unexpected subscription result from the server");
+            } else {
+                logger.debug("Reconnecting Subscription to: {}", subscriptionURL);
+                open();
+            }
+        }
+    }
+
+    public void getSubscriptionUrl() throws IOException {
```
jlaur, line 299: It looks like this method could be private?

kjoglum: Method made private.

jlaur (approved):
LGTM

PR #14577, opened 2023-03-11
[danfossairunit] Unify thing labels

jlaur:
Discovery result uses old (hardcoded) label "Danfoss HRV" rather than "Danfoss Air Unit":
image

Now discovery result and manually added thing are unified as "Danfoss Air Unit" and localization support for discovery result has been added.

Also categorized binding as "local".


lsiepel (approved):
LGTM

wborn (approved):
Thanks!

PR #14583, opened 2023-03-11
[tibber] Add channel for lastMeterProduction

kjoglum:
Adding requested channel for lastMeterProduction, part of live measurements.


bundles/org.openhab.binding.tibber/README.md
```diff
@@ -244,9 +247,8 @@ Number:ElectricCurrent     TibberAPILiveCurrent1                 "Live Current P
 Number:ElectricCurrent     TibberAPILiveCurrent2                 "Live Current Phase 2 [%.1 A]"              {channel="tibber:tibberapi:7cfae492:live_current2"}
 Number:ElectricCurrent     TibberAPILiveCurrent3                 "Live Current Phase 3 [%.1 A]"              {channel="tibber:tibberapi:7cfae492:live_current3"}
 Number:Power               TibberAPILivePowerProduction          "Live Power Production [%.0f W]"            {channel="tibber:tibberapi:7cfae492:live_powerProduction"}
+Number:Energy              TibberAPILiveAccumulatedProduction    "Accumulated Production [%.2f kWh]"         {channel="tibber:tibberapi:7cfae492:live_accumulatedProduction"}
+Number:Energy              TibberAPILiveLastMeterProduction      "Last Meter Production [%.2f kWh]"         {channel="tibber:tibberapi:7cfae492:live_lastMeterProduction"}
```
lsiepel, line 251: Very minor. As all lines have a perfect layout, except this line lacks a space :-)

lsiepel (approved):
If you can fix the space, otherwise LGTM

jlaur (changes requested):
Thanks! LGTM. Besides @lsiepel's comment, please also add an update instruction for the new channel. See for example #14572. This will ensure that the channel is automatically added for managed things when upgrading to the new version without a need to recreate thing.

bundles/org.openhab.binding.tibber/README.md
kjoglum: Corrected.

bundles/org.openhab.binding.tibber/src/main/resources/OH-INF/update/instructions.xml
```diff
@@ -0,0 +1,20 @@
+<?xml version="1.0" encoding="UTF-8" standalone="yes" ?>
+<update:update-descriptions xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+	xmlns:update="https://openhab.org/schemas/update-description/v1.0.0"
+	xsi:schemaLocation="https://openhab.org/schemas/update-description/v1.0.0 https://openhab.org/schemas/update-description-1.0.0.xsd">
+
+	<thing-type uid="tibber:tibberapi">
+		<instruction-set targetVersion="1">
+			<add-channel id="tomorrow_prices">
+				<type>tibber:tomorrow_prices</type>
+			</add-channel>
```
jlaur, line 10: This channel was added already in 3.4: #13416. I'm not sure what will happen when adding channels that already exists. 

If nothing bad will happen, in principle one could just add all channels here to "play it safe". If a user would upgrade straight from 3.3 to 4.0, the instruction above would actually help.

But in this case we wouldn't need the update instructions for "add-channel" at all (conceptually), since they could be added/updated from the thing definitions, and check/upgrade simply triggered when `thingTypeVersion` is bumped.

@J-N-K - I could use some input here. The next question that will come up will be whether to bump `thingTypeVersion` per release, milestone or PR.

kjoglum: Being unaware of the update feature, I added the latest channels added believing to achieve what you describes wrt different version upgrades. Will update accordingly based on @J-N-K’s answer.

kjoglum: Going away for some days, so made the update instructions applicable only for the lastMeterProduction channel.

J-N-K: Don't add channels that could already be present, calling `.withChannel` on a `ThingBuilder` will result in an `IllegalArgumentException` when the channel is already present. You can use update, because removing a non-existing channel does not result in an error.

jlaur (approved):
LGTM

PR #14586, opened 2023-03-12
[renault] Improve handling of HTTP 502 responses from Renault API

dougculnane:
The Renault API returns HTTP 502 if the lock status is not supported.   In which case this feature is disabled so that the binding does not repeat unnecessary log warnings and requests.

However there have been occasions where the battery status API returned HTTP 502 temporarily.  This disabled the battery status feature which is supported.  The battery level is then not updated until the RenaultCar "thing" is re-initialized.

This PR disables lock status on HTTP 502 but not other features.


jlaur:
> However there have been occasions where the battery status API returned HTTP 502 temporarily. This disabled the battery status feature which is supported. The battery level is then not updated until the RenaultCar "thing" is re-initialized.

Are you sure 502 should be interpreted as "not supported" for **any** calls, even the lock status call? After such occasions of temporarily returning 502, will it then go back to returning 501 if not supported? Or will it **always** return 502 when lock status is not supported?

dougculnane:
Hi @jlaur,

I have limited testing opportunity but the testing I have done (and that reported in the forum) suggests that (although inconsistent) the HTTP 502 on the lock status is a permanent error on not supported cars.

Maybe Renault will improve their implementation but this PR seams to be the best interpretation of what is there at the moment.

jlaur (approved):
Thanks for the fix!

PR #14590, opened 2023-03-12
[jrubyscripting] Filter files that are added to the watch list

jimtng:
~Depends on https://github.com/openhab/openhab-core/pull/3451~
The original NPE bug is fixed entirely within https://github.com/openhab/openhab-core/pull/3451 so this PR no longer depends on it. 

Somewhat related though is to filter the file types that are added to the watch list, so that the filewatcher won't try to load files with other extensions within the `automation/ruby` (sub)directory.


jlaur (approved):
Thanks!

Issue #12764, opened 2022-05-19
[Netatmo] Problem during update of 'room' thing


RickyTR:
## Current Behavior
At every refresh the room thing goes offline and then online again. Channels are updated correctly.

`2022-05-19 10:27:58.038 [hingStatusInfoChangedEvent] - Thing 'netatmo:room:xxxxxx:xxxxxx:xxxxxx' changed from ONLINE to OFFLINE: Thing is not reachable

2022-05-19 10:27:58.269 [hingStatusInfoChangedEvent] - Thing 'netatmo:room:xxxxxx:xxxxxx:xxxxxx' changed from OFFLINE: Thing is not reachable to ONLINE`

Every other thing of the binding updates correctly without problems

## Your Environment
Version used: Openhab 3.3 M5
Default Netatmo binding for 3.3 M5 distribuition (3.3.0.M5)

jlaur:
@clinique, @lolodomo - FYI.

lolodomo:
I believe TRACE logs around the moment when the two events are produced could help to understand what is happening. It will show the different requests and what is returned by Netatmo.
So please enable temporarily the TRACE logs and just wait for the next poll: log:set TRACE org.openhab.binding.netatmo
PS: Before publishing these logs here, replace the few ids identifying your products with "xxx".

RickyTR:
I enabled TRACE logs and I can see a lot of calls from the ApiBridgeHandler but it seems that every call is correctly completed:

executeUri returned : code [200 OK]

I cannot see any error before the switch OFFLINE/ONLINE of the room thing...

lolodomo:
This is not a surprise that requests are succeeding. 
The question is more what data they return and in particular the reachable field.

RickyTR:
Is there something in particular to look for?

clinique:
@RickyTR : don't bother, I'll have a look at it soon. As @lolodomo said, we've got a good idea at what to look for.

clinique:
Well, the weird thing is that I do not reproduce at all. My rooms are remarquably stuck ONLINE without flapping :-/

RickyTR:
May be the problem is related to channels? I use Indoor temp, Setpoint and Thermostat mode…

clinique:
Non, the status is disconnected from channels.

robnielsen:
>Non, the status is disconnected from channels. Maybe we can ask @robnielsen if he observes this also ? He uses the energy side of Netatmo devices.

[...]

PR #14595, opened 2023-03-13
[Netatmo] Add the capability to have a single home

clinique:
This enhancement let the user configure a single home instead of two (security, energy).

While doing it, I think I solved issue #12764 - and understood the origin of this uncommon bug.

Backward compatibility is preserved.

### DIFF
--- a/bundles/org.openhab.binding.netatmo/src/main/java/org/openhab/binding/netatmo/internal/config/HomeConfiguration.java
+++ b/bundles/org.openhab.binding.netatmo/src/main/java/org/openhab/binding/netatmo/internal/config/HomeConfiguration.java
@@ -28,10 +28,10 @@ public class HomeConfiguration extends NAThingConfiguration {
 
     @Override
     public String getId() {
-        return getId(energyId.isBlank() ? FeatureArea.SECURITY : FeatureArea.ENERGY);
+        return getIdForArea(energyId.isBlank() ? FeatureArea.SECURITY : FeatureArea.ENERGY);
     }
 
-    public String getId(FeatureArea feature) {
+    public String getIdForArea(FeatureArea feature) {
         return FeatureArea.ENERGY.equals(feature) ? energyId.isBlank() ? id : energyId
                 : FeatureArea.SECURITY.equals(feature) ? securityId.isBlank() ? id : securityId : id;
     }
--- a/bundles/org.openhab.binding.netatmo/src/main/java/org/openhab/binding/netatmo/internal/config/HomeConfiguration.java
+++ b/bundles/org.openhab.binding.netatmo/src/main/java/org/openhab/binding/netatmo/internal/config/HomeConfiguration.java
@@ -28,7 +28,7 @@ public class HomeConfiguration extends NAThingConfiguration {
 
     @Override
     public String getId() {
-        return energyId.isBlank() ? securityId : energyId;
+        return getId(energyId.isBlank() ? FeatureArea.SECURITY : FeatureArea.ENERGY);
     }
 
     public String getId(FeatureArea feature) {
--- a/bundles/org.openhab.binding.netatmo/src/main/java/org/openhab/binding/netatmo/internal/config/HomeConfiguration.java
+++ b/bundles/org.openhab.binding.netatmo/src/main/java/org/openhab/binding/netatmo/internal/config/HomeConfiguration.java
@@ -1,0 +1,38 @@
+/**
+ * Copyright (c) 2010-2023 Contributors to the openHAB project
+ *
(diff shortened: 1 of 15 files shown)


bundles/org.openhab.binding.netatmo/src/main/java/org/openhab/binding/netatmo/internal/handler/capability/HomeCapability.java
```diff
@@ -44,77 +42,67 @@
...
-            }
-            HomeStatus homeStatus = api.getHomeStatus(handler.getId());
-            if (homeStatus != null) {
-                result.add(homeStatus);
-            }
-        } catch (NetatmoException e) {
-            logger.warn("Error getting Home informations : {}", e.getMessage());
-        }
+        handler.getConfiguration().values().stream().filter(String.class::isInstance) //
+                .map(String.class::cast).collect(Collectors.toSet()).forEach(id -> {
```
lolodomo, line 91: Isn't it a little dangerous and a source of error in the future ? The day we add a new String member to the configuration class, it will be handled as an "id".
I would prefer a loop on securityId, energyId (and id ?)

bundles/org.openhab.binding.netatmo/src/main/java/org/openhab/binding/netatmo/internal/handler/capability/SecurityCapability.java
```diff
@@ -58,77 +64,60 @@ class SecurityCapability extends RestCapability<SecurityApi> {
...
+        if (objectId == null) {
+            return;
         }
-        String cameraId = homeEvent.getCameraId();
-        handler.getActiveChildren().stream().filter(handler -> cameraId.equals(handler.getId())).findFirst()
-                .ifPresent(handler -> {
-                    homeEvent.setIgnoredForThingUpdate(true);
-                    handler.setNewData(homeEvent);
-                });
+        handler.getActiveChildren().stream().filter(child -> child.getId().equals(objectId))
```
lolodomo, line 112: No need to pass FeatureArea.SECURITY here ?

bundles/org.openhab.binding.netatmo/src/main/java/org/openhab/binding/netatmo/internal/config/HomeConfiguration.java
```diff
@@ -0,0 +1,38 @@
...
+ */
+@NonNullByDefault
+public class HomeConfiguration extends NAThingConfiguration {
+    public String securityId = "";
+    public String energyId = "";
+
+    @Override
+    public String getId() {
+        return getId(energyId.isBlank() ? FeatureArea.SECURITY : FeatureArea.ENERGY);
+    }
```
lolodomo, line 32: Doi you reaaly need to override this method ?
It is a little weird to return something that is absolutely not an ID.

bundles/org.openhab.binding.netatmo/src/main/java/org/openhab/binding/netatmo/internal/handler/capability/HomeCapability.java
clinique: You're correct, this is not the most foolproof thing I ever wrote :)

bundles/org.openhab.binding.netatmo/src/main/java/org/openhab/binding/netatmo/internal/handler/capability/SecurityCapability.java
clinique: Yes, even better.

bundles/org.openhab.binding.netatmo/src/main/java/org/openhab/binding/netatmo/internal/config/HomeConfiguration.java
clinique: Gosh, this is a real mistake.

lolodomo (approved):
LGTM, thank you

lolodomo:
@clinique : are you not enough confident with your own tests ?

clinique:
@lolodomo : yes sir.

lolodomo:
If I am not wrong, you own energy and security stuff. What is your difficulty to test your change ?
What additional tests would you expect that you can't run yourself ?

clinique:
> If I am not wrong, you own energy and security stuff. What is your difficulty to test your change ? What additional tests would you expect that you can't run yourself ?

Sorry, I misread your question : are you **not** enough confident with your own tests ? => I answered to "are you comfident enough ? " :-D

lolodomo:
So please remove the tag "additional testing preferred" if this is not necessary.

Issue #14601, opened 2023-03-14
[Air Quality Binding] don't work properly with stantionid contains more than 4 symbols


Alex-Smart:
##Expected Behavior
[Air Quality Binding] don't work properly with stantionid contains more than 4 symbols.
With default stationId (automatically assigned) works as desired.

##Current Behavior
When I add a custom stationId with 4 or more symbols - thing is failed to initialize:
Status: UNINITIALIZED HANDLER_INITIALIZING_ERROR
`
2023-03-14 08:50:08.153 [ERROR] [nal.common.AbstractInvocationHandler] - An error occurred while calling method 'ThingHandler.initialize()' on 'org.openhab.binding.airquality.internal.handler.AirQualityStationHandler@49eaae55': null
java.lang.NullPointerException: null
	at org.openhab.binding.airquality.internal.handler.AirQualityStationHandler.lambda$1(AirQualityStationHandler.java:117) ~[?:?]
	at java.util.Optional.ifPresent(Optional.java:183) ~[?:?]
	at org.openhab.binding.airquality.internal.handler.AirQualityStationHandler.discoverAttributes(AirQualityStationHandler.java:111) ~[?:?]
	at org.openhab.binding.airquality.internal.handler.AirQualityStationHandler.initialize(AirQualityStationHandler.java:96) ~[?:?]
	at jdk.internal.reflect.GeneratedMethodAccessor109.invoke(Unknown Source) ~[?:?]
	at jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43) ~[?:?]
	at java.lang.reflect.Method.invoke(Method.java:566) ~[?:?]
	at org.openhab.core.internal.common.AbstractInvocationHandler.invokeDirect(AbstractInvocationHandler.java:154) [bundleFile:?]
	at org.openhab.core.internal.common.Invocation.call(Invocation.java:52) [bundleFile:?]
	at java.util.concurrent.FutureTask.run(FutureTask.java:264) [?:?]
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128) [?:?]
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628) [?:?]
	at java.lang.Thread.run(Thread.java:834) [?:?]
[...]

PR #14602, opened 2023-03-14
[AirQuality] Enhance API error handling

clinique:
Raised by #14601 
Also removed some dead code.

### DIFF
--- a/bundles/org.openhab.binding.airquality/src/main/java/org/openhab/binding/airquality/internal/api/ApiBridge.java
+++ b/bundles/org.openhab.binding.airquality/src/main/java/org/openhab/binding/airquality/internal/api/ApiBridge.java
@@ -72,18 +72,18 @@ public class ApiBridge {
 
         try {
             String response = HttpUtil.executeUrl("GET", urlStr, null, null, null, REQUEST_TIMEOUT_MS);
-            logger.debug("aqiResponse = {}", response);
-            AirQualityResponse result = GSON.fromJson(response, AirQualityResponse.class);
-            if (result != null) {
-            String error = result.getErrorMessage();
-            AirQualityData data = result.getData();
-            if (error.isEmpty()) {
-                Objects.requireNonNull(data);
-                return data;
+            if (response != null) {
+                logger.debug("aqiResponse = {}", response);
+                AirQualityResponse result = GSON.fromJson(response, AirQualityResponse.class);
+                if (result != null) {
+                    String error = result.getErrorMessage();
+                    if (error.isEmpty()) {
+                        return Objects.requireNonNull(result.getData());
+                    }
+                    throw new AirQualityException("Error raised : %s", error);
+                }
             }
-            throw new AirQualityException("Error raised : %s", error);
-            }
-            throw new JsonSyntaxException("Error deserializing API response".formatted(response));
+            throw new JsonSyntaxException("API response is null");
         } catch (IOException | JsonSyntaxException e) {
             throw new AirQualityException("Communication error", e);
         }
--- a/bundles/org.openhab.binding.airquality/src/main/java/org/openhab/binding/airquality/internal/api/ApiBridge.java
+++ b/bundles/org.openhab.binding.airquality/src/main/java/org/openhab/binding/airquality/internal/api/ApiBridge.java
@@ -74,16 +74,13 @@ public class ApiBridge {
             String response = HttpUtil.executeUrl("GET", urlStr, null, null, null, REQUEST_TIMEOUT_MS);
             logger.debug("aqiResponse = {}", response);
             AirQualityResponse result = GSON.fromJson(response, AirQualityResponse.class);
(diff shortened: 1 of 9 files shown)


bundles/org.openhab.binding.airquality/src/main/java/org/openhab/binding/airquality/internal/api/ApiBridge.java
```diff
@@ -66,21 +66,21 @@ private String buildRequestURL(String key, int stationId, String location) {
...
         try {
             String response = HttpUtil.executeUrl("GET", urlStr, null, null, null, REQUEST_TIMEOUT_MS);
             logger.debug("aqiResponse = {}", response);
             AirQualityResponse result = GSON.fromJson(response, AirQualityResponse.class);
-            if (result != null && result.getStatus() == ResponseStatus.OK) {
-                return result.getData();
-            } else if (retryCounter == 0) {
-                logger.debug("Error in aqicn.org, retrying once");
-                return getData(stationId, location, retryCounter + 1);
+            String error = result.getErrorMessage();
```
lolodomo, line 77: Caution: result could be null

bundles/org.openhab.binding.airquality/src/main/java/org/openhab/binding/airquality/internal/discovery/AirQualityDiscoveryService.java
```diff
@@ -84,14 +84,14 @@ protected void startScan() {
             PointType location = provider.getLocation();
             AirQualityBridgeHandler bridge = this.bridgeHandler;
             if (location == null || bridge == null) {
-                logger.debug("LocationProvider.getLocation() is not set -> Will not provide any discovery results");
+                logger.info("openHab server location is not defined, will not provide any discovery results");
```
lolodomo, line 87: openHAB

bundles/org.openhab.binding.airquality/src/main/java/org/openhab/binding/airquality/internal/handler/AirQualityStationHandler.java
```diff
@@ -110,13 +110,13 @@ public void initialize() {
     private void discoverAttributes() {
         getAirQualityData().ifPresent(data -> {
             // Update thing properties
-            Map<String, String> properties = new HashMap<>();
-            properties.put(ATTRIBUTIONS, data.getAttributions());
+            Map<String, String> properties = new HashMap<>(Map.of(ATTRIBUTIONS, data.getAttributions()));
```
lolodomo, line 113: Isn't it ok to do only
```
Map<String, String> properties = Map.of(ATTRIBUTIONS, data.getAttributions());
```
?

lolodomo (changes requested):

bundles/org.openhab.binding.airquality/src/main/java/org/openhab/binding/airquality/internal/api/ApiBridge.java
clinique: In theory, yes. Practically, no because there is a JsonSyntaxException catch. Lets satisfy the IDE and handle it.

clinique: Done

bundles/org.openhab.binding.airquality/src/main/java/org/openhab/binding/airquality/internal/discovery/AirQualityDiscoveryService.java
clinique: Corrected

bundles/org.openhab.binding.airquality/src/main/java/org/openhab/binding/airquality/internal/handler/AirQualityStationHandler.java
clinique: No because properties would then be immutable and I need the ability to add some other entries.

bundles/org.openhab.binding.airquality/src/main/java/org/openhab/binding/airquality/internal/api/ApiBridge.java
lolodomo: Just to be clear, the case that would have triggered a NPE with your previous code is the case if response is null or an empty string. Look at description of `HttpUtil.executeUrl`:

> return the response body or <code>NULL</code> when the request went wrong

Note that you need to apply spotless to make your code building.

clinique: Yes, you're correct - I woke up to early :)

lolodomo (approved):
LGTM, thank you

PR #14199, opened 2023-01-09
[miele] Improve multicast implementation

jlaur:
The main intention here was to fix some [MulticastSocket](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/net/MulticastSocket.html) deprecation warnings after upgrading to Java 17 - see description in #14198. However, while testing the refactoring needed for this, I came across some other issues that I have fixed as well:

- The multicast listener thread did not allow to be properly interrupted and thus would keep running after stopping or uninstalling the binding. I noticed this when replacing the JAR a few times and ended up with multiple threads in different versions running simultaneously.
- The conversion from byte buffer to `String` was not correct and the resulting string was filled up with trailing null-bytes. This could be seen in logs with log level **trace**, but there was a work-around in place for correctly handling the received data.
- Minor additional refactoring.

Fixes #14198

JARs for testing:
- [org.openhab.binding.miele-3.4.2-SNAPSHOT.jar](https://drive.google.com/file/d/1UX22yFdORvEn8zzud_bGn5C_he1tp6TE/view?usp=sharing)
- [org.openhab.binding.miele-4.0.0-SNAPSHOT.jar](https://drive.google.com/file/d/1A_QI_z7hTKwrnvCev68uNc4BobnoUrbY/view?usp=sharing)

PR #14603, opened 2023-03-14
Fix NullPointerException when getting network interface name

mhilbush:
While it occurs very rarely, it's possible for `NetworkInterface.getByInetAddress` to return null. This PR prevents the NPE in all bindings where this call is used.

Reported on forum [here](https://community.openhab.org/t/globalcache-binding-error-upgrade-from-openhab-3-2-to-3-4-2/145214).

### DIFF
--- a/bundles/org.openhab.binding.benqprojector/src/main/java/org/openhab/binding/benqprojector/internal/discovery/MulticastListener.java
+++ b/bundles/org.openhab.binding.benqprojector/src/main/java/org/openhab/binding/benqprojector/internal/discovery/MulticastListener.java
@@ -12,9 +12,7 @@
  */
 package org.openhab.binding.benqprojector.internal.discovery;
 
-import static org.openhab.binding.benqprojector.internal.BenqProjectorBindingConstants.DEFAULT_PORT;
-import static org.openhab.binding.benqprojector.internal.BenqProjectorBindingConstants.THING_PROPERTY_HOST;
-import static org.openhab.binding.benqprojector.internal.BenqProjectorBindingConstants.THING_PROPERTY_PORT;
+import static org.openhab.binding.benqprojector.internal.BenqProjectorBindingConstants.*;
 
 import java.io.IOException;
 import java.net.DatagramPacket;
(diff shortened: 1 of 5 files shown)


bundles/org.openhab.binding.benqprojector/src/main/java/org/openhab/binding/benqprojector/internal/discovery/MulticastListener.java
```diff
@@ -59,8 +57,9 @@ public class MulticastListener {
      */
     public MulticastListener(String ipv4Address) throws IOException, SocketException {
         InetAddress ifAddress = InetAddress.getByName(ipv4Address);
+        NetworkInterface netIF = NetworkInterface.getByInetAddress(ifAddress);
         logger.debug("Discovery job using address {} on network interface {}", ifAddress.getHostAddress(),
-                NetworkInterface.getByInetAddress(ifAddress).getName());
+                netIF != null ? netIF.getName() : "UNKNOWN");
```
jlaur, line 62: If you're interested in going one step further, two lines below:
`[WARNING] /home/jenkins/jenkins-agent1/workspace/PR-openHAB-Addons/bundles/org.openhab.binding.benqprojector/src/main/java/org/openhab/binding/benqprojector/internal/discovery/MulticastListener.java:[64,16] The method setInterface(java.net.InetAddress) from the type java.net.MulticastSocket is deprecated since version 14`

For reference: #14199

jlaur (approved):
LGTM - thanks.

mlobstein (approved):
LGTM

bundles/org.openhab.binding.benqprojector/src/main/java/org/openhab/binding/benqprojector/internal/discovery/MulticastListener.java
mhilbush: Yeah, I saw those were deprecated. But I didn't immediately know the correct way to resolve it. Thanks for the reference. I'll have a look.

mhilbush: But the fix should be in a separate PR, correct? Just in case you decide to back port this fix to 3.4.x.

jlaur: Didn't consider backporting, but good point. Let's keep the PRs small and tidy and merge this one as-is.

PR #14611, opened 2023-03-15
Replace deprecated MulticastSocket methods

mhilbush:
Replace deprecated `MulticastSocket` methods `setInterface` and `joinGroup`.

I believe these are the bindings that need to be changed: globalcache, benqprojector, epsonprojector, lcn.

Note I did not change lcn because the model is different than gc, benq, and epson. I was afraid I would break something, and I have no way to test my changes.

### DIFF
--- a/bundles/org.openhab.binding.benqprojector/src/main/java/org/openhab/binding/benqprojector/internal/discovery/MulticastListener.java
+++ b/bundles/org.openhab.binding.benqprojector/src/main/java/org/openhab/binding/benqprojector/internal/discovery/MulticastListener.java
@@ -17,11 +17,14 @@ import static org.openhab.binding.benqprojector.internal.BenqProjectorBindingCon
 import java.io.IOException;
 import java.net.DatagramPacket;
 import java.net.InetAddress;
+import java.net.InetSocketAddress;
+import java.net.InterfaceAddress;
 import java.net.MulticastSocket;
 import java.net.NetworkInterface;
 import java.net.SocketException;
 import java.net.SocketTimeoutException;
 import java.nio.charset.StandardCharsets;
+import java.util.Enumeration;
 import java.util.HashMap;
 import java.util.Locale;
 import java.util.Map;
@@ -44,6 +47,7 @@ public class MulticastListener {
     private final Logger logger = LoggerFactory.getLogger(MulticastListener.class);
 
     private MulticastSocket socket;
+    private InetSocketAddress inetSocketAddress;
 
     // BenQ projector devices announce themselves on the AMX DDD multicast port
     private static final String AMX_MULTICAST_GROUP = "239.255.250.250";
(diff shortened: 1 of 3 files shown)


mhilbush:
@jlaur Can you please look this over and let me know if this is on the right track? I've tested it and it's functioning correctly.

I'm throwing `SocketException` if I can't find a network interface (unlikely, but possible). Not sure of that's the most appropriate one to throw in this situation.

Thanks for the help/feedback.

jlaur:
> Can you please look this over and let me know if this is on the right track? I've tested it and it's functioning correctly.

Sure, I'll have a look.

For others - this PR was triggered from https://github.com/openhab/openhab-addons/pull/14603#discussion_r1136192158 and #14199 could possibly provide some inspiration, although based on reading documentation, not expert knowledge about multicast implementations. 🙂

bundles/org.openhab.binding.globalcache/src/main/java/org/openhab/binding/globalcache/internal/discovery/MulticastListener.java
```diff
@@ -59,18 +63,17 @@ public class MulticastListener {
...
         logger.debug("Discovery job using address {} on network interface {}", ifAddress.getHostAddress(),
-                netIF != null ? netIF.getName() : "UNKNOWN");
+                netIF.getName());
         socket = new MulticastSocket(GC_MULTICAST_PORT);
-        socket.setInterface(ifAddress);
+        socket.setNetworkInterface(netIF);
         socket.setSoTimeout(DEFAULT_SOCKET_TIMEOUT);
-        InetAddress mcastAddress = InetAddress.getByName(GC_MULTICAST_GROUP);
-        socket.joinGroup(mcastAddress);
+        socket.joinGroup(new InetSocketAddress(InetAddress.getByName(GC_MULTICAST_GROUP), GC_MULTICAST_PORT), null);
```
jlaur, line 76: No `leaveGroup` needed, for example in method below, `shutdown`?

mhilbush: Yes, I think that would make sense.

mhilbush:
@wborn I was going to change the methods in the other bindings where those deprecated methods are used, which is why I didn't tag the title with just the globalcache binding. I was going to add those bindings in another commit once there was agreement on the correct approach.

jlaur:
Sorry, it seems I forgot to respond to this one.

> Can you please look this over and let me know if this is on the right track? I've tested it and it's functioning correctly.

The approach looks good to me!

> I'm throwing `SocketException` if I can't find a network interface (unlikely, but possible). Not sure of that's the most appropriate one to throw in this situation.

It seems sensible.

mhilbush:
No worries. I'll do a commit for the other bindings.

bundles/org.openhab.binding.benqprojector/src/main/java/org/openhab/binding/benqprojector/internal/discovery/MulticastListener.java
```diff
@@ -57,20 +61,25 @@ public class MulticastListener {
...
     }
 
     public void shutdown() {
         logger.debug("Multicast listener closing down multicast socket");
-        socket.close();
+        try {
+            socket.leaveGroup(inetSocketAddress, null);
+            socket.close();
+        } catch (IOException e) {
+            logger.debug("Exception shutting down multicast socket");
```
jlaur, line 81: For this to be a bit more useful, perhaps add the exception message?
```suggestion
            logger.debug("Exception shutting down multicast socket: {}", e.getMessage());
```

(applies to the other two similar places in the PR as well)

jlaur (changes requested):
Thanks! Only one minor comment.

bundles/org.openhab.binding.benqprojector/src/main/java/org/openhab/binding/benqprojector/internal/discovery/MulticastListener.java
mhilbush: Good catch. Fixed.

jlaur (approved):
LGTM

Issue #14037, opened 2022-12-22
[easee] Missing values from https://api.easee.cloud/api/chargers/.../state


Diapolo:
I recently upgraded to openHAB 3.4 on my Raspi 4 and was happy to see an Easee-Addon. Before I was parsing the results from https://api.easee.cloud/api/chargers/[SERIALNUMBER]/state after requesting an access token. After setting up Easee Binding I'm now missing quite a few values that I used in my sitemap, these include (not a complete list):

energyPerHour
cellQuality
wifiQuality
inVoltageXYZ (XYZ replaced by phase etc. like inVoltageT2T3)
inCurrentXYZ (XYZ replaced by phase etc.like inCurrentT4)
cableRating

Did Easee change their API, seems I can't find documentation for https://api.easee.cloud/api/chargers/.../state anymore... well it seems that call is deprecated, but how can we get these values now?

@alexf2015 Thank you for the initial contribution!

Diapolo:
Did no other person using this addon make these observations?

alexf2015:
It seems that the endpoint was moved to the site which now returns overall site state + all charger states.

Diapolo:
Could you also consider adding the missing values I mentioned in the first posting? I'm willing to discover which ones are missing, but I can't help in implementing the changes for this addon.

Diapolo:
@alexf2015 It seems they changed their JSON-response so that `latestFirmware` isn't promoted anymore (was often lagging behind the installed version anyway) and the added a boolean value `connectedToCloud`. Just for reference if you plan to update your addon.

alexf2015:
The endpoint was readded to the docs. But as I already assumed it is now deprecated: https://developer.easee.cloud/reference/get_api-chargers-id-state

I will rework that when I find some time that means I will switchover to the new endpoint and also add the missing values.

alexf2015:
@Diapolo 
Do you have an idea what is the difference between
1. isOnline and connectedToCloud
2. totalPower and energyPerHour (the unit is different but the value should be the same)

Diapolo:
@alexf2015 AFAIK totalPower is the current charging power in kW (for me this sits around 10.X in the top, which is reasonable for a 11 kW charger), energyPerHour is a value I've not seen yet? Where are you getting that from?

[...]

PR #14613, opened 2023-03-15
[easee] migrated charger state API endpoint, added channels

alexf2015:
This PR implements some points mentioned in issue #14037

1. the charger state API endpoint is now migrated to the new API
2. some missing channels were added
3. changed behaviour in error case: instead of taking the bridge offline only the dedicated thing will be put offline. Thus a faulty charger config will not take down the whole bridge including all things.


alexf2015:
sorry messed this up. wrong base branch.

PR #14557, opened 2023-03-09
[paradoxalarm] Implement zone bypass command and additional zone states

theater:
It's a request from a member of the community to implement a zone bypass functionality from within the OpenHAB. Has been tested by me and the customer. Additionally I enhanced a bit the unit tests.

- Zone bypass command
- 8 new zone special read-only states
- Refactoring of some classes and extraction in interfaces of logic common for partition and zone commands
- Changed RAM pages enumeration so the interfaces of the related methods always use enumeration from 0(as in all Java APIs) and not from 1(as in Paradox system)
- Fixed potential NPEs in several places from the static code scan maven warnings (more to come in separate PR if this one gets merged)
- Minor fixes and code clean up

PR #14618, opened 2023-03-17
[paradoxalarm] Implement detailed partition state

### LINKED ISSUE #14608: [paradoxalarm] Add access to the different Alarm States (Armed, Armed in Stay, Was Is in Alarm)
The current version of the Binding seems to groups the different Armed states into the general Armed state, so it is not possible to see what type of Arm in the partition is in.

Armed - is usually used to arm all zones and exit the partition/area
Armed Stay - allows arming selected Zones so that you can "Stay" in the area, but certain selected Zones are "bypassed/ignored". This allows movement in select areas.

It would be really nice to have access to the additional states, either as distinct "switches" such as:

Armed Exit - On/Off
Armed Stay - On/Off
Armed Instant - On/Off
Was Is in Alarm - On/Off

The other option would be to Just Have a String in the Current State String to denote the current state, but as you mentioned that would require:

```
The other issue is that in any rules that we have we need to have "if (armed || armed in stay || armed in other way) do ...."
```
I think the separate Channel option would be easier to work with, though both have there pros and cons.

The Caddx binding which I used to use had separate channels for each Armed state

theater:
Hi,

The idea in this PR is when a partition is in "special" state (armed in stay or armed in no entry), to calculate the value and show it as a separate, detailed channel. Some users want this more detailed state for certain calculations/rules.

The armed in stay is still an "armed" state. For backwards compatibility we will keep the regular calculated, more simple state channel and we will introduce a new channel, called detailed state where a calculated effort will return the particular state.
In the backend these values are represented by a certain bits set and usually if you trigger Arm in stay, the bit for Armed and for Armed in stay are set to 1.
If no particular special state is set on the system, the detailed state will act as our old, overall state.

Depends on #14557 
Closes #14608

Cheers,
Konstantin


theater:
Hi @fwolter, Thanks for reviewing and merging my other PR.
I wonder as you already reviewed the other PR and merged it if you could have a look at this one too?

I hoped that when the one is merged into main most commits will disappear but they stayed in the PR. Basically this PR is on top of PR #14557 and the changes start with commit [2249249](https://github.com/openhab/openhab-addons/pull/14618/commits/2249249c8e85ec6bad3dbad5f08cbc866ea1228d)
Is it better to squash the commits and try to rebase on top of main after the merge? Will that reduce the changed files in this PR?

Best regards,
Konstantin

lolodomo:
Of course, you have to rebase on current main branch, so that only new code is reviewed.

theater:
OK. I rebased interactively and skipped the commits that were in the other PR. I guess that's the right thing to do...

lolodomo:
I probably missed your rebase. Your PR is now very small., I am going to review it.

theater:
Great. Looking forward it :)

bundles/org.openhab.binding.paradoxalarm/src/main/resources/OH-INF/thing/partition.xml
```diff
@@ -15,6 +15,7 @@
 		<channels>
 			<channel id="label" typeId="partitionLabel"/>
 			<channel id="state" typeId="state"/>
+			<channel id="detailedState" typeId="detailedState"/>
```
lolodomo, line 18: Please update the i18n file named paradoxalarm.properties. 2 entries to be added, one for label and one for description.

Please also provide an upgrade instuction XML file so that your new channel can be added automatically to existing things created with UI.

lolodomo (changes requested):

bundles/org.openhab.binding.paradoxalarm/src/main/resources/OH-INF/thing/partition.xml
theater: > Please also provide an upgrade instuction XML file so that your new channel can be added automatically to existing things created with UI.

Hm.... I'm not sure how it's done. Can you point me to where to see it? I never use the UI. Use the text files only...

theater: Disregard the last one. I think I figured out what I need to do (got advised from another PR). Working on it...

theater: OK I think I did the requested... Please review the XML as it's the first time I do it. 
There is a thing I don't understand in the documentation... if I change anything in the same channel in a subsequent PR, does that mean that i need to keep version 1 <add channel> and I need to have version 2 with <update channel> and keep a full track for every channel type which gets changed? basically for any change for a specific channel ID I need to have 1, 2, 3, etc... in the XML. Is my understanding correct?

bundles/org.openhab.binding.paradoxalarm/src/main/resources/OH-INF/update/partition_type_update.xml
```diff
@@ -0,0 +1,16 @@
...
+<update:update-descriptions xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+	xmlns:update="https://openhab.org/schemas/update-description/v1.0.0"
+	xsi:schemaLocation="https://openhab.org/schemas/update-description/v1.0.0 https://openhab.org/schemas/update-description-1.0.0.xsd">
+
+	<thing-type uid="paradoxalarm:partition">
+		<instruction-set targetVersion="1">
+			<add-channel id="detailedState">
+				<type>paradoxalarm:detailedState</type>
+				<label>Detailed Partition State</label>
+				<description>The detailed state of partition (contains sub-states like Stay in Armed, Armed no entry, etc)</description>
```
lolodomo, line 11: Your file looks good.
You can remove label/description, they will be inherited.
What is missing is the "thingTypeVersion" property you need to add in the thing type (XML). Here is an example:
https://github.com/openhab/openhab-addons/blob/main/bundles/org.openhab.binding.mielecloud/src/main/resources/OH-INF/thing/washingMachine.xml#L47

lolodomo (changes requested):

bundles/org.openhab.binding.paradoxalarm/src/main/resources/OH-INF/thing/partition.xml
lolodomo: Your thing type description (XML) will contain only the last version of your thing type with the higher `thingTypeVersion` property.
But you will keep all the upgrade instructions, allowing the different upgrades even for a user having an old version of the thing type.

theater:
Hi, I did the requested changes by adding the property to the partition.xml.
I wonder if I should expect anything in particular in the logs, regarding this change? 
Checked that overall the binding works as expected but not sure if it should log something regarding this update of the channel. 
From what I understand the idea is the runtime to know that something got changed, so it forcefully updates the channels. 

Had this issue before that some channels were not available in the UI when hot-replacing the addon but after a full OH restart, everything was OK. Not sure if these instructions to the framework address these issues...

lolodomo (approved):
Code LGTM, thank you

lolodomo:
> Had this issue before that some channels were not available in the UI when hot-replacing the addon but after a full OH restart, everything was OK. Not sure if these instructions to the framework address these issues...

The new thing upgrade feature only concerns things added using Main UI, that is stored in JSON DB. These things were never updated, even after a server restart. With this new OH 4 feature, the thing is automatically upgraded when the new binding version starts.
For things defined in file, all this is useless, a simple restart was enough to get the last version of the thing.

> I wonder if I should expect anything in particular in the logs, regarding this change?
> Checked that overall the binding works as expected but not sure if it should log something regarding this update of the channel.
> From what I understand the idea is the runtime to know that something got changed, so it forcefully updates the channels.

Yes. I am not sure something special is logged.

Your upgrade instructions look good to me but ideally it would be better that you first test that it works before I merge. Can you do that ?

Here is the scenario:

1. With 4.1M2 or a recent 4.1 snapshot, create with Main UI a partition thing. Of course, you will not see in Main UI the new channel.
2. Uninstall the paradoxalarm binding and install your new version of the binding.
3. After the start of the binding, the new channel should then be present in Main UI.

theater:
> Your upgrade instructions look good to me but ideally it would be better that you first test that it works before I merge. Can you do that ?

Yep. Thanks for the instructions. I did that. It works 😄

lolodomo:
> Yep. Thanks for the instructions. I did that. It works 😄

Excellent

PR #14623, opened 2023-03-17
[knx] Add hint for using *-controls to trigger rules

mgeramb:
Hint for using *-control channels added. Sorry, sign-off is missing, because the chance was done online in the github repo where I found no way for signoff.


mgeramb:
I have to stop the PR, I found that the OH sends out in some situation the broadcast 0/0/0 message.

PR #14630, opened 2023-03-18
[tradfri] Fix color tests after HSBType changes

jlaur:
Fix tests after openhab/openhab-core#3434

See https://ci.openhab.org/job/openHAB-Addons/1013/


jlaur:
@J-N-K - I have just adapted the tests according to actual values, but don't have the knowledge to assess if the actual values are correct. Some of them seem very different from the previous actual results, so maybe you can better assess this?

J-N-K:
What I usually do is using a tool like https://colorizer.org and draw both colors to the palette and compare if they are very similar. This is especially useful for xyY values, because they are very abstract. For HSB 0 and 356 seem to be very different, but considering they are angles and 360 is the same as 0, it's clear they are close. 

This PR LGTM.

wborn (approved):
LGTM!

PR #14632, opened 2023-03-18
[goecharger] Add check for correct temperature data

schogit:
Added check for correct length of the temperature array returnde by the Go eCharger Web API.

For details see the following thread: https://community.openhab.org/t/go-echarger-binding-stops-working-after-a-few-days-perhaps-issue-with-firmware-version-0-55/145070

Signed-off-by:  schogit <schoby1@gmx.de>

### DIFF
--- a/bundles/org.openhab.binding.goecharger/src/main/java/org/openhab/binding/goecharger/internal/handler/GoEChargerV2Handler.java
+++ b/bundles/org.openhab.binding.goecharger/src/main/java/org/openhab/binding/goecharger/internal/handler/GoEChargerV2Handler.java
@@ -145,7 +145,7 @@ public class GoEChargerV2Handler extends GoEChargerBaseHandler {
             case ALLOW_CHARGING:
                 return goeResponse.allowCharging == true ? OnOffType.ON : OnOffType.OFF;
             case TEMPERATURE_TYPE2_PORT:
-                if (goeResponse.temperatures == null) {
+                if (goeResponse.temperatures == null || goeResponse.temperatures.length < 2) {
                     return UnDefType.UNDEF;
                 }
                 return new QuantityType<>(goeResponse.temperatures[0], SIUnits.CELSIUS);


holgerfriedrich:
Your change LGTM. External input is unexpected und now handled as undef as all other missing data.

Power consumption uses the same unchecked array access just a few lines below - maybe time to apply the check there as well.

Your signoff is not properly formatted, maybe you can amend and force push.

schogit:
Hello,

I tried to adjust the commit message with properly formatted signed off line. Hope this helps. Regarding the other checks - should I also add them to this PR?

lolodomo:
@SamuelBrucksch : are you ok with this change ?

lolodomo:
Check also case TEMPERATURE_CIRCUIT_BOARD: length should be at least 2 to avoid a NPE when getting goeResponse.temperatures[1]

bundles/org.openhab.binding.goecharger/src/main/java/org/openhab/binding/goecharger/internal/handler/GoEChargerV2Handler.java
```diff
@@ -145,7 +145,7 @@ protected State getValue(String channelId, GoEStatusResponseBaseDTO goeResponseB
             case ALLOW_CHARGING:
                 return goeResponse.allowCharging == true ? OnOffType.ON : OnOffType.OFF;
             case TEMPERATURE_TYPE2_PORT:
-                if (goeResponse.temperatures == null) {
+                if (goeResponse.temperatures == null || goeResponse.temperatures.length < 2) {
```
lolodomo, line 148: Why not 1 instead of 2 ?
The idea is to be sure to have at least one value to then get it with `goeResponse.temperatures[0]` ?

schogit: Hello,

in my case it was like this. If the Array Length if goeResponse.temperatures was 2 there were realistic values for the temperatures. If it was only one  value in the array the value was far away from possible values (e.g. -16843.01). That was the reason why I would ignore such values.

lolodomo: Ok, could you please add a comment in the code to explain that ?

lolodomo:
> Check also case TEMPERATURE_CIRCUIT_BOARD: length should be at least 2 to avoid a NPE when getting `goeResponse.temperatures[1]`

I also see potential NPE when `goeResponse.energy` is used in the same method.

@schogit : as you are not able to sign-off properly your PR, keep it with the minimum changes (what you did but with an additional comment required) and I will create another PR fixing the other switch cases.

lolodomo:
I created another PR that includes your change.

lolodomo:
Closed by #14933

PR #14647, opened 2023-03-21
[actiontemplatehli] Simplify usage and add UI

GiviMAD:
This is a rework of this add-on.

Mayor difference is that it stops requiring file system configurations as it includes a small UI to configure the interpreter actions. 

This UI is developed with a minimum amount of dependencies thinking on avoiding as much maintenance work as posible in the future. Still I used typescript and vite for the building process, axios and pkce-challenge (for supporting the authentication workflow), and also picocss to have some predefined styles.

Also there are some minor structure changes, now actions can target multiple item types and be restricted by its semantic and/or tags.


GiviMAD:
Some ui recoding 
[actiontemplatehliUI.webm](https://user-images.githubusercontent.com/9007708/226726112-728bff23-a8b9-4d19-8f0e-4033bc5795fd.webm)

GiviMAD:
@lolodomo probably still a lot of room for improvement but I think now its a little simple to use.

GiviMAD:
Another option, which will avoid the having a separate UI, can be create things that represent the actions that way the configuration can be done using the main UI.

Also after reading @dalgwen idea of adding an intent catalog, maybe is easier to remove all the target configuration from the actions and link them to the items using a metadata tag.

If someone have some feedback about it, please let me know.

PR #14649, opened 2023-03-21
[googlestt] Fix drop bytes

GiviMAD:
There was a malfunction some times with this bundle and realize we were dropping the data chunk when we were not able to write the desired size. 

This is a problem because as those are bytes not samples so we were probably breaking the audio encoding when an odd number of bytes were dropped.
I copied the same solution used in the rustpotterks bundle to read until fill the buffer.

### DIFF
--- a/bundles/org.openhab.voice.googlestt/src/main/java/org/openhab/voice/googlestt/internal/GoogleSTTService.java
+++ b/bundles/org.openhab.voice.googlestt/src/main/java/org/openhab/voice/googlestt/internal/GoogleSTTService.java
@@ -255,10 +255,12 @@ public class GoogleSTTService implements STTService {
         long startTime = System.currentTimeMillis();
         long maxTranscriptionMillis = (config.maxTranscriptionSeconds * 1000L);
         long maxSilenceMillis = (config.maxSilenceSeconds * 1000L);
-        int readBytes = 6400;
+        int bufferSize = 6400;
+        int numBytesRead;
+        int remaining = bufferSize;
+        byte[] audioBuffer = new byte[bufferSize];
         while (!aborted.get()) {
-            byte[] data = new byte[readBytes];
-            int dataN = audioStream.read(data);
+            numBytesRead = audioStream.read(audioBuffer, bufferSize - remaining, remaining);
             if (aborted.get()) {
                 logger.debug("Stops listening, aborted");
                 break;
@@ -272,16 +274,18 @@ public class GoogleSTTService implements STTService {
                 logger.debug("Stops listening, max silence time reached");
                 break;
             }
-            if (dataN != readBytes) {
+            if (numBytesRead != remaining) {
+                remaining = remaining - numBytesRead;
                 try {
                     Thread.sleep(100);
                 } catch (InterruptedException e) {
                 }
                 continue;
             }
+            remaining = bufferSize;
             StreamingRecognizeRequest dataRequest = StreamingRecognizeRequest.newBuilder()
-                    .setAudioContent(ByteString.copyFrom(data)).build();
-            logger.debug("Sending audio data {}", dataN);
+                    .setAudioContent(ByteString.copyFrom(audioBuffer)).build();
+            logger.debug("Sending audio data {}", bufferSize);
             clientStream.send(dataRequest);
         }
     }


bundles/org.openhab.voice.googlestt/src/main/java/org/openhab/voice/googlestt/internal/GoogleSTTService.java
```diff
@@ -255,10 +255,12 @@ private void streamAudio(ClientStream<StreamingRecognizeRequest> clientStream, A
         long startTime = System.currentTimeMillis();
         long maxTranscriptionMillis = (config.maxTranscriptionSeconds * 1000L);
         long maxSilenceMillis = (config.maxSilenceSeconds * 1000L);
-        int readBytes = 6400;
+        int bufferSize = 6400;
```
jlaur, line 258: It seems this could be `final`?

bundles/org.openhab.voice.googlestt/src/main/java/org/openhab/voice/googlestt/internal/GoogleSTTService.java
```diff
@@ -272,16 +274,18 @@ && isExpiredInterval(maxSilenceMillis, responseObserver.getLastInputTime())) {
                 logger.debug("Stops listening, max silence time reached");
                 break;
             }
-            if (dataN != readBytes) {
+            if (numBytesRead != remaining) {
+                remaining = remaining - numBytesRead;
```
jlaur, line 278: It believe this will be wrong when `numBytesRead` is -1 because we reached the end of the stream?

bundles/org.openhab.voice.googlestt/src/main/java/org/openhab/voice/googlestt/internal/GoogleSTTService.java
```diff
@@ -272,16 +274,18 @@ && isExpiredInterval(maxSilenceMillis, responseObserver.getLastInputTime())) {
...
                 try {
                     Thread.sleep(100);
```
jlaur, line 280: Is this actually needed?

jlaur (changes requested):
Thanks for the fix! A few minor questions from my side.

bundles/org.openhab.voice.googlestt/src/main/java/org/openhab/voice/googlestt/internal/GoogleSTTService.java
Miguel-MAD: Yes, we need to update the remaining so next read just add the missing bytes to the buffer in the correct offset.

GiviMAD: Yes, we need to update the remaining so next read just add the missing bytes to the correct buffer offset.

bundles/org.openhab.voice.googlestt/src/main/java/org/openhab/voice/googlestt/internal/GoogleSTTService.java
GiviMAD: I didn't think about this as those services are only used with real time audio right now, but it's a good catch, better have the Stt services ready to support a finite stream.

bundles/org.openhab.voice.googlestt/src/main/java/org/openhab/voice/googlestt/internal/GoogleSTTService.java
jlaur: I was referring to the thread sleep. 🙂 The `read()` itself is blocking.

GiviMAD: Ups,sorry. Probably not, it was a sugestión on the previous review to prevent damaging the cpu in case the stream start returning 0 bytes forever which I think can not happen but also I think is not breaking anything as it's so small. Also after see the logs the other day I saw we stop there each 4 or 5 reads, so maybe it's already reducing the cpu usage a little.
But really I'm not sure whether or not is a good a idea to have it.

jlaur: If this is the `InputStream` implementation, it should always read at least 1 byte, unless it is requested to read 0 (`len` = 0), but this would be a bug:
https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/io/InputStream.html#read(byte%5B%5D,int,int)

GiviMAD: Thank you for the reference.
Yes seems quite clear. I didn't read this one before.

So probably better remove it as you said, in case it causes some unwanted behaviour. 

I should remove it in the rustpotter one also.

bundles/org.openhab.voice.googlestt/pom.xml
```diff
@@ -14,146 +14,155 @@
...
+    <dependency>
+      <groupId>com.google.api</groupId>
+      <artifactId>gax-httpjson</artifactId>
+      <version>0.108.3</version>
+    </dependency>
+    <dependency>
+      <groupId>com.google.code.gson</groupId>
+      <artifactId>gson</artifactId>
+      <version>2.10.1</version>
+    </dependency>
```
jlaur, line 83: Do you need this for the dependencies? Just checking, since it will add more than 200 kB to the bundle until this is upgraded in core, which is unfortunately not easy: openhab/openhab-core#3321.

GiviMAD: Seems to be required in that range. 
```
filter:="(&(osgi.wiring.package=com.google.gson)(version>=2.10.0)(!(version>=3.0.0)))"]]
```
I think I already tried to bundle without it and it failed but I'm not totally sure it was not the mentioned error and I mixed things.

Let me do a check.

jlaur:
> Do you think is better to split some part in to a different PR or is it ok like this?

It's okay, but perhaps you can rephrase the PR title to mention something user-relatable, i.e. what is fixed for the user? If you think the fix should be cherry-picked into 3.4.x, it would need to be split.

bundles/org.openhab.voice.googlestt/pom.xml
GiviMAD: Seems to work ok without it!

GiviMAD:
Perfect, I'm going move the dependency update to a separate PR.

I think the part about the onComplete method can stay in this one as I think is the correct way to use the library, is that ok?

About the title I'm not totally sure, because could be something like "Fix audio streaming implementation"? because it was really broken, I think it should be something that tells you "If you already tried and it performs badly, give it another try, as it was broken".

jlaur:
> I think the part about the onComplete method can stay in this one as I think is the correct way to use the library, is that ok?

It seems to be a part of the fix, so yes. 🙂

> About the title I'm not totally sure, because could be something like "Fix audio streaming implementation"? because it was really broken, I think it should be something that tells you "If you already tried and it performs badly, give it another try, as it was broken".

If I understand this correctly, this add-on is a "Speech-to-Text" service, i.e. you will most likely use it to trigger something when saying something? If the audio streaming implementation had a severe bug, does it mean that it didn't work at all, or perhaps that only the first part of the sentence would be translated correctly, because the buffer would then not be filled up correctly? Or would it drop some of the audio stream randomly, but then again be able to stream and translate? I guess it all depends on the details, i.e. how this could be perceived by the user. I can help formulating it when I understand how this could possibly influence the users.

GiviMAD:
Yes, I think it was randomly injecting noise into the words, because I don't think that loosing that audio part was causing so much trouble, maybe some wrong word in the middle of a sentence or a short connector, but as the audio used was encoded using two bytes per sample I think each time we dropped an odd number of bytes, we where sending mostly noise, utils dropping another odd number of bytes, so causing a very random behavior when some times it works almost perfect another times missing multiple consecutive words in phrase.

jlaur (approved):
LGTM

PR #14651, opened 2023-03-22
[tapocontrol] Fix for L920 color values with decimals

seime:
Fix for L920 lightstrip when sending colour values with decimals (server reported error -1008)
Also updated the doc with actual channels listed for L920 in README.


jlaur:
@seime - thanks for the fix! Can you please create a new PR targeting the **main** branch and consisting only of the fix itself (as I believe there will be conflicts for the README)? We can then cherry-pick that fix into the **3.4.x** branch.

PR #14659, opened 2023-03-24
[MQTT] Add example usage for incoming value transformations using a map

markfickett:
# Add example usage for incoming value transformations using a map.

# Description

This is a quick improvement to the docs for the Map Transformation. It has an example map file, but didn't have an example of how to use that file as an input transform. This will hopefully help beginners who want to use maps.

I'm just getting started with OH, so I welcome any fixes to my terminology, or caveats around which particular versions / contexts this example works in.

More context at [this community post](https://community.openhab.org/t/oh3-transformation-service-config-for-js-and-map-in-a-thing-channel-what-worked-qs-about-what-didnt/145507)

Signed-off-by: Mark Fickett <mfseas@gmail.com>

### DIFF
--- a/bundles/org.openhab.transform.map/README.md
+++ b/bundles/org.openhab.transform.map/README.md
@@ -33,6 +33,9 @@ white\ space=using escape
 | `white space` | `using escape` |
 | `anything`    | `default`      |
 
+## Usage as a Thing Channel Transformation
+
+To use the above map on a Channel value coming into your Thing, set the Incoming Value Transformations value to `MAP:binary.map`.
 
 ## Usage as a Profile
 


bundles/org.openhab.transform.map/README.md
```diff
@@ -33,6 +33,9 @@ white\ space=using escape
 | `white space` | `using escape` |
 | `anything`    | `default`      |
 
+## Usage as a Thing Channel Transformation
```
lolodomo, line 36: That is not clear for me what we are talking about !
Maybe you should provide an example of thing (things file and if possible UI thing).

markfickett: Adding that context sounds useful, but unfortunatley is way beyond me. I've been using OpenHAB for about a week, and have only configured Things via the UI so far -- I'm not sure what "UI Thing" means, or how to helpfully incorporate a Thing example into this documentation. 

This is one of a bunch of things I got confused about trying to set up my first couple things, so I'm trying to just contribute back some helpful notes. I found these docs while trying to get a `MAP` input transform working, and was quite confused about whether to use `MAP(mymap.map)`, `MAP:mymap.map:%s`, etc. So I could see this simple example with some relevant keywords ("incoming value transformation" showed up in the UI I was trying to use) helping someone else out

If you want to just take this as a request to add an example like that, that would be awesome, no need to specifically accept my PR. But if you don't have time, then maybe this would be a helpful stepping stone.

jimtng: @markfickett can you please take a screenshot and show us what/where you're referring to?

markfickett: For sure!

On the Thing > Channel > Show advanced page, the input box at the bottom of this screenshot. Under the "Transform Values " heading, the input box labeled "Incoming Value Transformations". The URL to this page on my server is `http://localhost:8084/settings/things/mqtt:topic:d432b36bc2:b1f8c9243a/channels/openevse_state/edit`. I currently have it as `JS:openevse_parse_state.js`, but I had also tested out using a `MAP` for this with `MAP:openevse_state.map` before settling on javascript so I could customize the behavior a little bit more.

openhab_thing_channel_incoming_value_transform

lolodomo: Ok, this a MQTT stuff, PR title should be updated and this is the MQTT binding documentation that should be updated.

J-N-K: I think it‘s already there. Maybe something like „For usage of transformations in add-ons/bindings please see the corresponding add-on/binding documentation“. Maybe it‘s not clear that there is additional documentation there.

jimtng: Thanks for the screenshot. Could you explain what was confusing? 
Would [this document](https://www.openhab.org/addons/bindings/mqtt.generic/#incoming-value-transformation) help with clarifying that confusion?

lolodomo: An example with Map transformation could then be added in MQTT documentation. 
As this MQTT channel configuration accepts apparently different kind of transformations, this is better to example the general syntax in MQTT, even more when this syntax is specific to MQTT., rather than adding a MQTT chapter in all transformation services documentation.

jimtng: Also an idea is to provide a link from the config description to the relevant doc page, in this case to the mqtt binding section for the incoming/outgoing transformation. Saves people from hunting around the docs esp when they're unfamiliar with the doc structure / where to look.

markfickett: Yes, that was exactly my confusion -- I didn't realize that part of the UI was drawing from the add-on/binding, so I was trying to apply the [main openHAB Transformations docs](https://www.openhab.org/docs/configuration/transformations.html) where I should have been looking up the MQTT docs ([I think it's these ones](https://www.openhab.org/addons/bindings/mqtt.generic/#incoming-value-transformation)?). Some kind of heading or link saying "Add-on settings, see these docs" for the relevant section/page would be awesome.

I have only set up MQTT things so far, aside from a SolarEdge API binding which just worked without adding much extra detail, so I don't have an expectation yet for what's common to all Things and what comes from a binding. Thanks all for bearing with my confusion to get to the root of it.

lolodomo:
I close this PR as there is no more activity and as suggested the doc has to be enhanced but not at the proposed place.

Issue #14660, opened 2023-03-24
[iCloud] Changes in device Id in iCloud API


maihacke:
Lately users discover problems with changes in the device ID in the iCloud API.
See discussion here https://community.openhab.org/t/icloud-binding-communication-error/122131/597

I saved the device data json before and after this change happened.
Each device has three IDs in the json:

id, baUUID, deviceDiscoveryId

Currently we use `id` to identify devices. This id has changed lately (back and forth and now again, see discussion). That's why openhab identfies these devices as new and put them into the inbox. The other IDs did not change - at least between my two files. So we have some chances that changing the binding to use one of the other IDs this problem could be solved.

Since we have no docs for the API, I propose to go with `deviceDiscoveryId`.

PR #14661, opened 2023-03-24
[iCloud] Fix things reappearing in inbox

maihacke:
# Change handling for device IDs in icloud binding

Lately users discover problems with changes in the device ID in the iCloud API.
See discussion here https://community.openhab.org/t/icloud-binding-communication-error/122131/597

I saved the device data json before and after this change happened.
Each device has three IDs in the json:

id, baUUID, deviceDiscoveryId

Currently we use id to identify devices. This id has changed lately (back and forth and now again, see discussion). That's why openhab identfies these devices as new and put them into the inbox. The other IDs did not change - at least between my two files. So we have some chances that changing the binding to use one of the other IDs this problem could be solved.

This PR changes the binding to use deviceDiscoveryId instead of id. As fallback it uses the device name, since old devices do not have a deviceDiscoveryId.

Should fix https://github.com/openhab/openhab-addons/issues/14660

### DIFF
--- a/bundles/org.openhab.binding.icloud/src/main/java/org/openhab/binding/icloud/internal/discovery/ICloudDeviceDiscovery.java
+++ b/bundles/org.openhab.binding.icloud/src/main/java/org/openhab/binding/icloud/internal/discovery/ICloudDeviceDiscovery.java
@@ -64,7 +64,7 @@ public class ICloudDeviceDiscovery extends AbstractDiscoveryService implements I
 
             String thingLabel = deviceOwnerName + " (" + deviceTypeName + ")";
             String deviceDiscoveryId = deviceInformationRecord.getDeviceDiscoveryId();
-            if (deviceDiscoveryId == null || "".equals(deviceDiscoveryId)) {
+            if (deviceDiscoveryId == null || deviceDiscoveryId.isBlank()) {
                 logger.debug("deviceDiscoveryId is empty, using device name for identification.");
                 deviceDiscoveryId = deviceOwnerName;
             }
--- a/bundles/org.openhab.binding.icloud/src/main/java/org/openhab/binding/icloud/internal/discovery/ICloudDeviceDiscovery.java
+++ b/bundles/org.openhab.binding.icloud/src/main/java/org/openhab/binding/icloud/internal/discovery/ICloudDeviceDiscovery.java
@@ -64,6 +64,10 @@ public class ICloudDeviceDiscovery extends AbstractDiscoveryService implements I
 
             String thingLabel = deviceOwnerName + " (" + deviceTypeName + ")";
             String deviceDiscoveryId = deviceInformationRecord.getDeviceDiscoveryId();
+            if (deviceDiscoveryId == null || "".equals(deviceDiscoveryId)) {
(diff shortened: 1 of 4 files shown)


maihacke:
@jlaur Without this change the icloud binding stops working after a few days. Maybe we could consider it for 4.0.0.

bundles/org.openhab.binding.icloud/src/main/java/org/openhab/binding/icloud/internal/discovery/ICloudDeviceDiscovery.java
```diff
@@ -63,19 +63,23 @@ public void deviceInformationUpdate(List<ICloudDeviceInformation> deviceInformat
             String deviceOwnerName = deviceInformationRecord.getName();
 
             String thingLabel = deviceOwnerName + " (" + deviceTypeName + ")";
-            String deviceId = deviceInformationRecord.getId();
-            String deviceIdHash = Integer.toHexString(deviceId.hashCode());
+            String deviceDiscoveryId = deviceInformationRecord.getDeviceDiscoveryId();
+            if (deviceDiscoveryId == null || "".equals(deviceDiscoveryId)) {
```
jlaur, line 67:
```suggestion
            if (deviceDiscoveryId == null || deviceDiscoveryId.isBlank()) {
```

bundles/org.openhab.binding.icloud/src/main/java/org/openhab/binding/icloud/internal/handler/ICloudDeviceHandler.java
```diff
@@ -185,8 +185,11 @@ private void updateLocationRelatedStates(ICloudDeviceInformation deviceInformati
         this.logger.debug("Device: [{}]", this.deviceId);
 
         for (ICloudDeviceInformation deviceInformationRecord : deviceInformationList) {
-            String currentId = deviceInformationRecord.getId();
-
+            String currentId = deviceInformationRecord.getDeviceDiscoveryId();
+            if (currentId == null || "".equals(currentId)) {
```
jlaur, line 189:
```suggestion
            if (currentId == null || currentId.isBlank()) {
```

jlaur (changes requested):
Thanks, only two minor suggestions added.

Do you think "Fix things reappearing in inbox" would be an accurate title for the PR (to go into the release notes)?

maihacke:
Thanks for the review.

jlaur (approved):
LGTM

jlaur:
@maihacke - for my understanding, why did the binding stop working after a few days? It seems that the "Fix things reappearing in inbox" does not fully cover the actual fix.

maihacke:
> @maihacke - for my understanding, why did the binding stop working after a few days? It seems that the "Fix things reappearing in inbox" does not fully cover the actual fix.

The reason is that the IDs which where used in the binding before that fix where no longer constant. So they changed from time to time in the API results. This lead to
A: New things appeared in the in inbox for the new IDs.
B: Existing things stopped working because there IDs where no longer valid in the iCloud API.

Maybe we could use

"Existing things stopped working and re-appeared as new in the inbox".

jlaur:
@maihacke - thanks.

PR #14663, opened 2023-03-24
[androiddebugbridge] Fix screen state channel on Android 12

GiviMAD:
Fixes for the screen state channel, tested on chrome cast with android 12.


jlaur (approved):
Thanks!

PR #14664, opened 2023-03-24
[rustpotterks] Remove unnecesary thread sleep call

GiviMAD:
As discussed in #14649 seems unnecessary, removed in case it causes problems.


jlaur (approved):
Thanks!

jlaur:
@GiviMAD - perhaps you should also add the `-1` check here to potentially avoid an infinite loop?

GiviMAD:
Done, thank you for pointing it

PR #14667, opened 2023-03-24
[boschshc] Code Enhancements

david-pace:
This commit fixes several compiler warnings as well as several Checkstyle, FindBugs, PMD and Sonar issues.

* instantiate loggers consistently
* add missing logger calls
* add chained exceptions to logger calls
* provide central Gson instance that does not serialize or deserialize loggers
* avoid catching NullPointerExceptions
* extract methods where cyclomatic complexity was too high
* remove unnecessary null check in combination with instanceof
* rename local variables that shadow fields
* add missing `@NonNull` and `@Nullable` annotations

### DIFF
--- a/bundles/org.openhab.binding.boschshc/src/main/java/org/openhab/binding/boschshc/internal/devices/bridge/BoschHttpClient.java
+++ b/bundles/org.openhab.binding.boschshc/src/main/java/org/openhab/binding/boschshc/internal/devices/bridge/BoschHttpClient.java
@@ -37,10 +37,10 @@ import org.eclipse.jetty.http.HttpMethod;
 import org.eclipse.jetty.http.HttpStatus;
 import org.eclipse.jetty.util.ssl.SslContextFactory;
 import org.openhab.binding.boschshc.internal.exceptions.BoschSHCException;
+import org.openhab.binding.boschshc.internal.serialization.GsonUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import com.google.gson.Gson;
 import com.google.gson.JsonSyntaxException;
 
 /**
@@ -51,9 +51,8 @@ import com.google.gson.JsonSyntaxException;
  */
 @NonNullByDefault
 public class BoschHttpClient extends HttpClient {
-    private static final Gson GSON = new Gson();
 
-    private final Logger logger = LoggerFactory.getLogger(BoschHttpClient.class);
+    private final Logger logger = LoggerFactory.getLogger(getClass());
 
     private final String ipAddress;
     private final String systemPassword;
@@ -177,8 +176,10 @@ public class BoschHttpClient extends HttpClient {
                 logger.debug("Online check failed with status code: {}", contentResponse.getStatus());
                 return false;
             }
-        } catch (TimeoutException | ExecutionException | NullPointerException e) {
-            logger.debug("Online check failed because of {}!", e.getMessage());
+        } catch (InterruptedException e) {
+            throw e;
+        } catch (Exception e) {
+            logger.debug("Online check failed because of {}!", e.getMessage(), e);
             return false;
         }
     }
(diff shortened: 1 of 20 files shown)


bundles/org.openhab.binding.boschshc/src/main/java/org/openhab/binding/boschshc/internal/devices/bridge/BoschHttpClient.java
```diff
@@ -249,13 +252,13 @@ public boolean doPairing() throws InterruptedException {
...
-        } catch (TimeoutException | CertificateEncodingException | KeyStoreException | NullPointerException e) {
-            logger.warn("Pairing failed with exception {}", e.getMessage());
+        } catch (TimeoutException | CertificateEncodingException | KeyStoreException | RuntimeException e) {
+            logger.warn("Pairing failed with exception {}", e.getMessage(), e);
             return false;
         } catch (ExecutionException e) {
             // javax.net.ssl.SSLHandshakeException: General SSLEngine problem
             // => usually the pairing failed, because hardware button was not pressed.
-            logger.trace("Pairing failed - Details: {}", e.getMessage());
+            logger.trace("Pairing failed - Details: {}", e.getMessage(), e);
```
GerdZanker, line 261: As far as I'm able remember one of first contribution to the binding, this ExecutionException can and will happen.
Compare also with comment. 
Therefore this is not a bug which can be fixed and should not log the stack trace from my point of view. Compare with [CodeGuideline](https://www.openhab.org/docs/developer/guidelines.html#f-logging) says "Configuration errors by users should only print log messages about what's wrong."

bundles/org.openhab.binding.boschshc/src/main/java/org/openhab/binding/boschshc/internal/services/dto/BoschSHCServiceState.java
```diff
@@ -27,15 +27,7 @@
...
-    /**
-     * gson instance to convert a class to json string and back.
-     */
-    private static final Gson GSON = new Gson();
-
-    /**
-     * Logger marked as transient to exclude the logger from JSON serialization.
-     */
-    private final transient Logger logger = LoggerFactory.getLogger(BoschSHCServiceState.class);
+    private final Logger logger = LoggerFactory.getLogger(getClass());
```
GerdZanker, line 30: How or why is the logger now  excluded from JSON serialization?

bundles/org.openhab.binding.boschshc/src/main/java/org/openhab/binding/boschshc/internal/devices/bridge/BridgeHandler.java
```diff
@@ -473,8 +473,7 @@ private void forwardStateToHandlers(DeviceServiceData deviceServiceData, JsonEle
             // All children of this should implement BoschSHCHandler
             @Nullable
             ThingHandler baseHandler = childThing.getHandler();
-            if (baseHandler != null && baseHandler instanceof BoschSHCHandler) {
```
GerdZanker, line 476: I think you will the same line in my new ThingDiscoveryService.
Can you please replace my c&p lines with the nice instance check with cast?

GerdZanker (changes requested):
Some question and one additional change please.

bundles/org.openhab.binding.boschshc/src/main/java/org/openhab/binding/boschshc/internal/services/dto/BoschSHCServiceState.java
david-pace: The logger was excluded from serialization before using the `transient` keyword. However, this triggers the Sonar issue `Fields in non-serializable classes should not be "transient" (java:S2065)`.

The new solution solves this issue by using a central Gson instance retrieved via `GsonUtils.DEFAULT_GSON_INSTANCE`. This instance is configured to skip all fields named `logger` using an `ExclusionStrategy`. This does not only solve the Sonar issue, but also makes the code more efficient since we only need one Gson instance. Before every class that needed JSON serialization created an individual Gson instance.

GerdZanker: Thank you for explanation

Issue #14635, opened 2023-03-18
[yeelight] Device going offline


goopilot:
Facing the issue with device status going offline. After some troubleshooting, found that following method

https://github.com/openhab/openhab-addons/blob/ee1de1186470f9602110f9e9d2eb2d2d3b037487/bundles/org.openhab.binding.yeelight/src/main/java/org/openhab/binding/yeelight/internal/lib/device/connection/WifiConnection.java#L54

kills the socket that was created under

https://github.com/openhab/openhab-addons/blob/ee1de1186470f9602110f9e9d2eb2d2d3b037487/bundles/org.openhab.binding.yeelight/src/main/java/org/openhab/binding/yeelight/internal/lib/device/connection/WifiConnection.java#L96

so any new call of this method fails due to non-existent connection. 

Unfortunately I don't have enough knowledge to fix this, so I was wondering if somebody else can have a look?

There is a workaround for this is to increase Duration (for example to 5000) https://community.openhab.org/t/yeelight-color-led-bulb-offline-using-yeelight-binding/96583/4 but it's not clear how this is related.

PR #14670, opened 2023-03-25
[Yeelight] Fix disconnect method

goopilot:
Signed-off-by: Alexandr Salamatov [goopilot@gmail.com](mailto:goopilot@gmail.com)

Fixing an issue with having connection active after disposing thing. This will fix https://github.com/openhab/openhab-addons/issues/14635

### DIFF
--- a/bundles/org.openhab.binding.yeelight/src/main/java/org/openhab/binding/yeelight/internal/lib/device/connection/WifiConnection.java
+++ b/bundles/org.openhab.binding.yeelight/src/main/java/org/openhab/binding/yeelight/internal/lib/device/connection/WifiConnection.java
@@ -130,8 +130,12 @@ public class WifiConnection implements ConnectionBase {
         mDevice.setAutoConnect(false);
         mCmdRun = false;
         try {
-            mConnectThread.interrupt();
-            mSocket.close();
+            if (mConnectThread != null) {
+                mConnectThread.interrupt();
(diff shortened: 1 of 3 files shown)


bundles/org.openhab.binding.yeelight/src/main/java/org/openhab/binding/yeelight/internal/lib/device/connection/WifiConnection.java
```diff
@@ -128,6 +128,16 @@ public boolean connect() {
     @Override
     public boolean disconnect() {
         mDevice.setAutoConnect(false);
-        return false;
+        mCmdRun = false;
+        try {
+            mConnectThread.interrupt();
+            mSocket.close();
```
lolodomo, line 134: You could check that mConnectThread / mSocket are not null.

lolodomo (changes requested):

bundles/org.openhab.binding.yeelight/src/main/java/org/openhab/binding/yeelight/internal/lib/device/connection/WifiConnection.java
goopilot: Thanks, added

lolodomo (approved):
LGTM

Issue #14674, opened 2023-03-25
[innogysmarthome] Remove binding (replaced by livisismarthome)


Novanic:
The innogysmarthome binding should get removed, because there is a new binding "livisismarthome" since OpenHAB 3.3.0.

The innogysmarthome binding will stop working soon (at least in early 2024) because the cloud API will get shutdown. The new livisismarthome binding isn't affected by this issue, because it uses a new local API (which is also faster and more reliable).

See https://github.com/openhab/openhab-addons/pull/12440 for more information.

PR #14675, opened 2023-03-25
[innogysmarthome] Remove binding (replaced by livisismarthome)

Novanic:
Remove binding (replaced by livisismarthome)

See issue #14674 for more information.


ollie-dev (approved):

Novanic:
Hello @wborn , is it normal, that the build fails (because the building binding was removed) or is there something which has to get changed before the pull request can get merged? Thanks

wborn:
It's an issue in the build scripts. They should do a full build instead of a partial build in case an add-on gets removed. So we should make them a little bit smarter. 😉 Haven't found the time yet to look into this edge case.

wborn:
> Haven't found the time yet to look into this edge case.

It was an easy fix, so I was able to scrap from my todo list with https://github.com/openhab/openhab-addons/pull/14680. :slightly_smiling_face:

wborn (approved):
The changes look good to me.

The only issue I see is that it may be a bit too early to already remove this when it is only retired in about a year.

See: https://community.livisi.de/paq-en/

> On this page we want to answer some of your questions regarding the retirement of the LIVISI SmartHome Services on March 1st 2024.

jlaur:
> The only issue I see is that it may be a bit too early to already remove this when it is only retired in about a year.

I agree.

Novanic:
> The changes look good to me.
> 
> The only issue I see is that it may be a bit too early to already remove this when it is only retired in about a year.
> 
> See: https://community.livisi.de/paq-en/
> 
> > On this page we want to answer some of your questions regarding the retirement of the LIVISI SmartHome Services on March 1st 2024.

@wborn and @jlaur 

Some arguments to remove the binding with OpenHAB 4.0 (probably summer 2023):
- The official apps are supported until March 2024, the binding isn't part of that.
- The vendor wishes that more users are switching to the local API to reduce the traffic and costs and it's also the vendors wish to remove the old binding.
- Originally it was planned to remove the binding with OpenHAB 3.4.0.
- The binding is documented as deprecated for 1 year (from OpenHAB 3.3 / summer 2022 till OpenHAB 4.0 / summer 2023). The users have at least 1 year time to switch to the new binding.
- I think the major release of OpenHAB 4.0.0 is a good time. Why should a deprecated binding move to a new major version and why should users upgrade to the next major release 4.0.0 but don't want to switch to the new binding?
- Users which currently don't want to switch to the new binding till March 2024, they could stay on OpenHAB 3.x.
- As I know the vendor has already started to block some client IDs...

Novanic:
I got now the info that the vendor has already shutdown the cloud API for OpenHAB this month. The binding is already not working anymore. Therefore the binding can now get removed.

jlaur:
Those are good and convincing arguments. Let's merge then!

PR #14680, opened 2023-03-26
Trigger full build when add-on is removed

wborn:
When an add-on is removed a partial build fails and instead a full build should be done.

See: https://github.com/openhab/openhab-addons/pull/14675#issuecomment-1484020311

Tested in my fork using https://github.com/wborn/openhab-addons/pull/9.


J-N-K (approved):

jlaur (approved):
Thanks!

PR #14684, opened 2023-03-26
[hdpowerview] Declare connection

jlaur:


jlaur:
@wborn - it seems this one triggered a full Jenkins build, do you have any idea why?

andrewfg:
@jlaur to be honest I don't really know what the 'connection' xml element even means. ?? 

.. does it indicate if the binding uses a LAN bases API resp. a Cloud based one ??

Anyway LGTM :)

andrewfg (approved):
LGTM

wborn:
The Jenkins job was updated similar to #14680. But apparently it was not looking in the correct workspace dir for the existence of the add-on bundle which has now been corrected!

wborn (approved):

jlaur:
> to be honest I don't really know what the 'connection' xml element even means. ??

See openhab/openhab-core#2058.

> .. does it indicate if the binding uses a LAN based API resp. a Cloud based one ??

Exactly.

andrewfg:
@jlaur ok thanks. I guess I need to add this to a few more of the bindings where I am involved. For the avoidance of doubt is the ‘category’ the same as the category for channels? Or is it just a more generic text tag?

jlaur:
> For the avoidance of doubt is the ‘category’ the same as the category for channels? Or is it just a more generic text tag?

I don't think I understand the question. 🙂 If you mean if declared connection is similar to category for channels, then probably no. It more or less declares per binding if there is a dependency to external services. I don't know yet how it will be used, but could imagine filtering or even grouping. I guess you would need to ask creator of the issue if you would like to know more about the intentions and perhaps plans. In any case, to unlock the possible use cases, first we need to declare this in as many bindings as possible.

andrewfg:
@jlaur I was referring to the following.. (keywords / categories)..

> keywords - optional but recommended - Keywords (categories) out of an open list to filter against, e.g. "heating, lighting, graphing"

PR #14688, opened 2023-03-26
[zway] Code improvements

lsiepel:
No functional changes. 
 - removed a dependency on org.apache.commons
 - some null annotations
 - some other SAT warnings
 - regenerated the i18n file (ordering and aligning)

I would also like to fix the exception handling as it now (on several places) catches Throwable and it then does an instanceof with both exception and Error. Not sure how/why this was setup this way, so i left it behind. If the reviewer knows, please advice how to fix those too.


clinique (approved):
LGTM

bundles/org.openhab.binding.zway/src/main/java/org/openhab/binding/zway/internal/converter/ZWayDeviceStateConverter.java
```diff
@@ -109,13 +112,9 @@ private static State getPercentState(String multilevelValue) {
...
      */
-    private static State getBinaryState(String binarySwitchState) {
+    private static State getBinaryState(@Nullable String binarySwitchState) {
         if (binarySwitchState != null) {
-            if (binarySwitchState.equals("on")) {
-                return OnOffType.ON;
-            } else if (binarySwitchState.equals("off")) {
-                return OnOffType.OFF;
-            }
+            return OnOffType.from(binarySwitchState);
```
jlaur, line 117: This will return `OFF` when neither "on" or "off", so it breaks the previous behavior where the method would ultimately return `UNDEF`.

jlaur (changes requested):
Thanks! One comment to be checked.

bundles/org.openhab.binding.zway/src/main/java/org/openhab/binding/zway/internal/converter/ZWayDeviceStateConverter.java
lsiepel: Fixed

jlaur (approved):
LGTM

jlaur:
@pathec - are you still involved? I can see you are code owner for this binding, but not in the contributor group, thus you were not automatically added as reviewer.

PR #14714, opened 2023-03-27
[sleepiq] Add functionality to control the bed foundation

mhilbush:
Adds new channels for controlling the bed foundation. Supports 
- adjusting the head and foot by selecting a preset
- adjusting the head and foot independently using a specific adjustment setting (between 0 and 100)
- controlling the night light outlet
- controlling the under-bed lighting

Channels are updated based on status returned form the SleepIQ API so it should stay in sync with operations performed using the smartphone app.

### DIFF
--- a/bundles/org.openhab.binding.sleepiq/src/main/java/org/openhab/binding/sleepiq/internal/handler/SleepIQCloudHandler.java
+++ b/bundles/org.openhab.binding.sleepiq/src/main/java/org/openhab/binding/sleepiq/internal/handler/SleepIQCloudHandler.java
@@ -200,11 +200,8 @@ public class SleepIQCloudHandler extends ConfigStatusBridgeHandler {
      * @param bedId the bed identifier
      * @return the identified {@link Bed} or <code>null</code> if no such bed exists
      */
-    public @Nullable Bed getBed(final @Nullable String bedId) {
+    public @Nullable Bed getBed(final String bedId) {
         logger.debug("CloudHandler: Get bed object for bedId={}", bedId);
-        if (bedId == null) {
-            return null;
-        }
         List<Bed> beds = getBeds();
         if (beds != null) {
             for (Bed bed : beds) {
@@ -223,11 +220,8 @@ public class SleepIQCloudHandler extends ConfigStatusBridgeHandler {
      * @param side the side of the bed
      * @return the sleeper or null if sleeper not found
      */
-    public @Nullable Sleeper getSleeper(@Nullable String bedId, Side side) {
+    public @Nullable Sleeper getSleeper(String bedId, Side side) {
         logger.debug("CloudHandler: Get sleeper object for bedId={}, side={}", bedId, side);
-        if (bedId == null) {
-            return null;
-        }
         List<Sleeper> localSleepers = sleepers;
         if (localSleepers != null) {
             for (Sleeper sleeper : localSleepers) {
@@ -246,10 +240,7 @@ public class SleepIQCloudHandler extends ConfigStatusBridgeHandler {
      * @param sleepNumber the sleep number multiple of 5 between 5 and 100
      * @param side the chamber to set
      */
-    public void setSleepNumber(@Nullable String bedId, Side side, int sleepNumber) {
-        if (bedId == null) {
-            return;
-        }
+    public void setSleepNumber(String bedId, Side side, int sleepNumber) {
         try {
             cloud.setSleepNumber(bedId, side, sleepNumber);
         } catch (SleepIQException e) {
@@ -263,10 +254,7 @@ public class SleepIQCloudHandler extends ConfigStatusBridgeHandler {
      * @param bedId the bed identifier
      * @param mode turn pause mode on or off
      */
-    public void setPauseMode(@Nullable String bedId, boolean mode) {
-        if (bedId == null) {
-            return;
-        }
+    public void setPauseMode(String bedId, boolean mode) {
         try {
             cloud.setPauseMode(bedId, mode);
         } catch (SleepIQException e) {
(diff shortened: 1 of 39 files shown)


bundles/org.openhab.binding.sleepiq/src/main/java/org/openhab/binding/sleepiq/internal/handler/SleepIQCloudHandler.java
```diff
@@ -262,6 +274,116 @@ public void setPauseMode(@Nullable String bedId, boolean mode) {
...
+     *
+     * @param bedId the bed identifier
+     * @param side the side of the bed
+     * @param actuator the head or foot of the bed
+     * @param position the new position of the actuator
+     * @param speed the speed with which to make the adjustment
+     */
+    public void setFoundationPosition(@Nullable String bedId, Side side, FoundationActuator actuator, int position,
+            FoundationActuatorSpeed speed) {
+        if (bedId == null) {
```
jlaur, line 350: It doesn't make sense to call this method with `bedId == null`. Consider making it non-nullable and avoid calling it in this case.

bundles/org.openhab.binding.sleepiq/src/main/java/org/openhab/binding/sleepiq/internal/handler/SleepIQCloudHandler.java
```diff
@@ -262,6 +274,116 @@ public void setPauseMode(@Nullable String bedId, boolean mode) {
...
+     * Set a foundation adjustment preset.
+     *
+     * @param bedId the bed identifier
+     * @param side the side of the bed
+     * @param preset the preset to be applied
+     * @param speed the speed with which to make the adjustment
+     */
+    public void setFoundationPreset(@Nullable String bedId, Side side, FoundationPreset preset,
+            FoundationActuatorSpeed speed) {
+        if (bedId == null) {
```
jlaur, line 325: It doesn't make sense to call this method with `bedId == null`. Consider making it non-nullable and avoid calling it in this case.

bundles/org.openhab.binding.sleepiq/src/main/java/org/openhab/binding/sleepiq/internal/handler/SleepIQCloudHandler.java
```diff
@@ -262,6 +274,116 @@ public void setPauseMode(@Nullable String bedId, boolean mode) {
...
+        return null;
+    }
+
+    /**
+     * Get the foundation status of the specified bed
+     *
+     * @param bedId the bed identifier
+     */
+    public @Nullable FoundationStatusResponse getFoundationStatus(@Nullable String bedId) {
+        if (bedId == null) {
```
jlaur, line 302: It doesn't make sense to call this method with `bedId == null`. Consider making it non-nullable and avoid calling it in this case.

bundles/org.openhab.binding.sleepiq/src/main/java/org/openhab/binding/sleepiq/internal/handler/SleepIQCloudHandler.java
```diff
@@ -262,6 +274,116 @@ public void setPauseMode(@Nullable String bedId, boolean mode) {
         }
     }
 
+    /**
+     * Get the foundation features of the specified bed
+     *
+     * @param bedId the bed identifier
+     */
+    public @Nullable FoundationFeaturesResponse getFoundationFeatures(@Nullable String bedId) {
+        if (bedId == null) {
```
jlaur, line 283: It doesn't make sense to call this method with `bedId == null`. Consider making it non-nullable and avoid calling it in this case.

jlaur (changes requested):
The code looks good and clean to me. I have added only a minor comment. Also, consider dynamically removing the new channels if a foundation is not present - see for example #13275.

bundles/org.openhab.binding.sleepiq/src/main/java/org/openhab/binding/sleepiq/internal/handler/SleepIQCloudHandler.java
mhilbush: Fixed

bundles/org.openhab.binding.sleepiq/src/main/java/org/openhab/binding/sleepiq/internal/handler/SleepIQCloudHandler.java
mhilbush: Fixed

bundles/org.openhab.binding.sleepiq/src/main/java/org/openhab/binding/sleepiq/internal/handler/SleepIQCloudHandler.java
mhilbush: Fixed

jlaur (approved):
LGTM

jlaur:
> I did think about making the foundation channels dynamic. I don't want to just remove them dynamically because they could be removed incorrectly if there's a transient error on the call to the SleepIQ API. Then you're kinda screwed until you delete and readd the thing, right?

Yes, it would need to make that decision only when there is a reliable response with the correct capabilities. There is nothing wrong with providing channels not supported by all devices (beds), so this would only be sugar on the top. 🙂

jlaur:
> Yes, it would need to make that decision only when there is a reliable response with the correct capabilities. There is nothing wrong with providing channels not supported by all devices (beds), so this would only be sugar on the top. 🙂

Sorry, was leaving the computer, so missed to ask the implicit question: Do you want to do more in this PR, or prefer to merge as-is?

mhilbush:
If you're good with it, merge as is please.

PR #14719, opened 2023-03-28
[knx] Code cleanup

holgerfriedrich:
cleanup.
- replace replaceAll by replace (as no regex needed)
- compare enum values with ```==``` instead ```.equals```


wborn (approved):
Thanks for the cleanup!

PR #14724, opened 2023-03-30
Adapt to ColorUtil changes in core

holgerfriedrich:
see openhab/openhab-core#3479

may be merged as separate commits without squashing

### DIFF
--- a/bundles/org.openhab.binding.deconz/src/main/java/org/openhab/binding/deconz/internal/handler/LightThingHandler.java
+++ b/bundles/org.openhab.binding.deconz/src/main/java/org/openhab/binding/deconz/internal/handler/LightThingHandler.java
@@ -446,7 +446,7 @@ public class LightThingHandler extends DeconzBaseThingHandler {
                 xyY[0] = xy[0];
                 xyY[1] = xy[1];
                 xyY[2] = ((double) bri) / BRIGHTNESS_MAX;
-                updateState(channelUID, ColorUtil.xyToHsv(xyY));
+                updateState(channelUID, ColorUtil.xyToHsb(xyY));
             }
         } else if (bri != null && hue != null && sat != null) {
             updateState(channelUID,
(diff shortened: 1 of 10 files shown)


antroids:
Hi [@ccutrer](https://github.com/ccutrer), could you take a look

andrewfg:
^
I wonder if rather than using `assertEquals(expected, actual)` (or similar assertions) you should instead use the `assertEquals(expected, actual, delta)` variant? The `delta` variant allows you to set a comparison precision analogous to `assertTrue(Math.abs(actual - expected) < delta)`

holgerfriedrich:
@andrewfg Thanks for mentioning. I agree that double/float comparisons should always use a delta - even if the current check works.
In this case it is not my intent to change existing code in the plugins, except the value changes to adapt to the next rounding behaviour of core functions.
Do you refer to a specific file, TradfriColorTest.java?

andrewfg:
> Do you refer to a specific file, TradfriColorTest.java?

Yes. And also, I think the ones using assertThat..is..

J-N-K:
The correct hamcrest matcher is `closeTo`

J-N-K:
What is the state here? I believe it needs to be merged immediately after the PR in core to not break the build.

holgerfriedrich:
Correct. I renamed one function and the rounding behavior will require tests to be slightly adapted.

I did not check if it is still building after the latest changes, but it should be fine.

J-N-K (approved):
LGTM

holgerfriedrich:
I can have a look later tonight....

J-N-K:
The issue is a missing new core snapshot. It builds fine locally.

jlaur:
See also https://github.com/openhab/openhab-addons/actions/runs/4644251102

holgerfriedrich:
@jlaur yes, could your pls merge the PR?

PR #14748, opened 2023-04-05
[fsinternetradio] Add device "Technisat VIOLA 2 C IR" as an supported radio

schmidmuc:
Added device "Technisat VIOLA 2 C IR" as an supported radio


jlaur (approved):
Thanks!

Issue #13938, opened 2022-12-13
[mielecloud] Integration tests failure


jlaur:
Today it failed in this **Ubuntu 22.04** CI build: https://github.com/openhab/openhab-addons/actions/runs/3679100300/jobs/6223149500

```
TEST org.openhab.binding.mielecloud.internal.handler.DryerDeviceThingHandlerTest#testActionsChannelUpdatesForValidValues() <<< ERROR: expected: <ON> but was: <NULL>
org.opentest4j.AssertionFailedError: expected: <ON> but was: <NULL>
	at org.junit.jupiter.api.AssertionUtils.fail(AssertionUtils.java:55)
	at org.junit.jupiter.api.AssertionUtils.failNotEqual(AssertionUtils.java:62)
	at org.junit.jupiter.api.AssertEquals.assertEquals(AssertEquals.java:182)
	at org.junit.jupiter.api.AssertEquals.assertEquals(AssertEquals.java:177)
	at org.junit.jupiter.api.Assertions.assertEquals(Assertions.java:1141)
	at org.openhab.binding.mielecloud.internal.handler.DryerDeviceThingHandlerTest.lambda$5(DryerDeviceThingHandlerTest.java:246)
	at org.openhab.core.test.java.JavaTest.waitForAssert(JavaTest.java:245)
	at org.openhab.core.test.java.JavaTest.waitForAssert(JavaTest.java:213)
	at org.openhab.core.test.java.JavaTest.waitForAssert(JavaTest.java:177)
	at org.openhab.binding.mielecloud.internal.handler.AbstractMieleThingHandlerTest.waitForAssert(AbstractMieleThingHandlerTest.java:137)
	at org.openhab.binding.mielecloud.internal.handler.DryerDeviceThingHandlerTest.testActionsChannelUpdatesForValidValues(DryerDeviceThingHandlerTest.java:245)
```

_Originally posted by @wborn in https://github.com/openhab/openhab-addons/issues/12627#issuecomment-1347312919_

BjoernLange:
I will have another look at this. However, I was not able to reproduce the issue in a local build, so any help doing this is appreciated. What I can tell from a quick look at the logs is that some channel state updates sometimes don't seem to get through in the integration tests.

jlaur:
One more: https://github.com/openhab/openhab-addons/actions/runs/3719345202/jobs/6308147859

BjoernLange:
[...]

PR #14762, opened 2023-04-08
Disable mielecloud tests

J-N-K:
See #13938 

As they always fail and therefore CI builds fail.


jlaur (approved):

jlaur:
It seems we still have an issue: https://github.com/openhab/openhab-addons/actions/runs/4644543063/jobs/8219876804#step:9:1

PR #14771, opened 2023-04-09
[feican] Use new core class ColorUtil for RGB conversion

holgerfriedrich:
Use new class ```ColorUtil``` which has recently been added to core in openhab/openhab-core#3479.

```convertColorPercentToByte``` is not longer needed. It has been using a statement ```...divide(BigDecimal.valueOf(100), 2, RoundingMode.HALF_UP).byteValue();``` which can be found in several bindings and but probably does not exactly behave as intended. It rounds the BigDecimal with behavior HALF_UP, keeping *2* scales behind the decimal point and then ignores the scales by using byteValue. If HALF_UP is intended, scales should be set to 0.

### DIFF
--- a/bundles/org.openhab.binding.feican/src/main/java/org/openhab/binding/feican/internal/Commands.java
+++ b/bundles/org.openhab.binding.feican/src/main/java/org/openhab/binding/feican/internal/Commands.java
@@ -12,13 +12,11 @@
  */
 package org.openhab.binding.feican.internal;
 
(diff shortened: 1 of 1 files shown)


Hilbrand (approved):
LGTM

jlaur:
@Hilbrand - just checking, any reason why you have not merged this PR yet after approving it?

lolodomo (approved):
LGTM, thank you

Issue #12643, opened 2022-04-23
[KNX] DPT 232.600 converted from RGB to HSB without decimal places


Gehetzter:
**Describe the bug**
Openhab works with HSB color values. To use the DPT 232.600 the Knx RGB values are converted to HSB.
In this conversion decimal places are lost. When converting back to RGB the result is not the same that was sent over Knx.

**Expected behavior**
I expect the RGB values being converted to HSB with decimal places.

**Actual result**
The HSB is truncated to integers.

**Additional context**
Guess the problem is here, somewere between Line 600 and 620:
[KNXCoreTypeMapper.java](https://github.com/openhab/openhab-addons/blob/main/bundles/org.openhab.binding.knx/src/main/java/org/openhab/binding/knx/internal/dpt/KNXCoreTypeMapper.java)

holgerfriedrich:
We have improved the rounding behavior during the color conversions. This decreases the difference between RGB values being received and sent out again.

Our primary color representation is HSB, and the current implementation rounds to integer values during all conversions.
This is the expected behavior. This implies that converting from any representation to HSB and back is not a lossless transformation.

@J-N-K @andrewfg @kaikreuzer For me, the handling of color conversions should be kept as it is (i.e. rounding to integer values). Could you give your opinion on this please?

So for me, this issue should be closed as "won't fix".

Gehetzter:
Thanks for reviewing my issue, but I´m a bit shocked of your suggestion.
Why would you restrict it like that?
Don't know about the "current" implementation, I´m on 3.4.2 but there I can easily have decimals in HSB.
e.g. when I do 
![grafik](https://user-images.githubusercontent.com/104268512/230883306-23a81608-f1ff-48a5-ad27-1bbad8d95df4.png)
It seems to me that only the knx binding is the limit.

If anyone wants to have it rounded to Integer it can be easily done e.g. in the state Description with %.0f
On the other hand, if you do this in the knx binding already there is no chance to get the exact value on application side...

J-N-K:
I'm not sure. What speaks against using decimals internally? Of course the 0-255 part should be `int`, that is because they represent the values of one byte and therefore decimals don't make sense. `DecimalType` and `PercentType` both support decimals.

andrewfg:
> handling of color conversions should be kept as it is

I am not sure how the following relates to this particular issue, but..

[...]

PR #14772, opened 2023-04-09
[knx] Improve rounding when encoding DPT 232.60000

holgerfriedrich:
Correct the rounding behavior.
See also openhab/openhab-core#3479.

Refs. #12643

Some additional fixes for compilation and SAT warnings that were recently introduced.

### DIFF
--- a/bundles/org.openhab.binding.knx/src/main/java/org/openhab/binding/knx/internal/handler/IPBridgeThingHandler.java
+++ b/bundles/org.openhab.binding.knx/src/main/java/org/openhab/binding/knx/internal/handler/IPBridgeThingHandler.java
@@ -171,12 +171,13 @@ public class IPBridgeThingHandler extends KNXBridgeBaseThingHandler {
         if (!config.getLocalIp().isEmpty()) {
             localEndPoint = new InetSocketAddress(config.getLocalIp(), 0);
         } else {
-            if (networkAddressService == null) {
+            final var localNetworkAddressService = networkAddressService;
+            if (localNetworkAddressService == null) {
                 logger.debug("NetworkAddressService not available, cannot create bridge {}", thing.getUID());
                 updateStatus(ThingStatus.OFFLINE);
                 return;
             } else {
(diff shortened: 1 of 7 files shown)


bundles/org.openhab.binding.knx/src/main/java/org/openhab/binding/knx/internal/handler/IPBridgeThingHandler.java
```diff
@@ -171,12 +171,13 @@ public void initializeLater() {
         if (!config.getLocalIp().isEmpty()) {
             localEndPoint = new InetSocketAddress(config.getLocalIp(), 0);
         } else {
-            if (networkAddressService == null) {
+            final var localNetworkAddressService = networkAddressService;
```
J-N-K, line 174:
```suggestion
            NetworkAddressService localNetworkAddressService = networkAddressService;
```

Local variables don't need to be final because they can't be accessed by other threads anyway. We try to avoid using `var` because it hides the original type and is harder to read.

J-N-K (changes requested):
Otherwise LGTM

bundles/org.openhab.binding.knx/src/main/java/org/openhab/binding/knx/internal/handler/IPBridgeThingHandler.java
holgerfriedrich: There were more occurrences of "final var" in the binding, I removed all of them now....

J-N-K:
Can you fix the DCO?

```
git commit --amend -s
```

and then force-push.

holgerfriedrich:
> Can you fix the DCO?

Sure. Sorry.

J-N-K (approved):
Thanks

jlaur:
@holgerfriedrich - does it resolve #12643?

holgerfriedrich:
> @holgerfriedrich - does it resolve #12643?

@jlaur not fully. For me #12643 is a "won't fix". Let me write a comment there.

holgerfriedrich:
@J-N-K, @jlaur could you please merge this one?

J-N-K (approved):
LGTM

PR #14773, opened 2023-04-09
[nanoleaf] Use new core class ColorUtil for RGB conversion

holgerfriedrich:
Use new class ```ColorUtil``` which has recently been added to core in https://github.com/openhab/openhab-core/pull/3479.

```HsbType::getRGB()``` has been replaced since it has been deprecated.


stefan-hoehn:
Thanks Holger, that makes sense to me (I also looked at the original PR).
Please bare with me as I need quite some time to make sure it works as expected because I need to build and deploy the addon and then test it on the various devices in my home. Even worse, I don't have line-devices, so we would need to ask @austvik to check that as well.

austvik (approved):
Sorry for the delay, I have been on vacation.

I have tested this patch out locally with emphesize on the changes in the PanelHandler. I have not tested all shape types, but the code is the same for all the shape types so I think it is sufficient.

lolodomo (approved):
LGTM, thank you

PR #14774, opened 2023-04-09
[imperihome] Use new core class ColorUtil for RGB conversion

holgerfriedrich:
Use new class ```ColorUtil``` which has recently been added to core in https://github.com/openhab/openhab-core/pull/3479.


lolodomo:
No need to enhance this binding, it has rather to be removed from the distribution. 
Imperihome app died several years ago.

holgerfriedrich:
Thanks @lolodomo!

How do we start the removal process? We seem to have a lot of deprecated or unmaintained bindings which cause a lot of useless work when it comes to updates, SAT, warnings, etc....

lolodomo:
When the API is no more available and no new API is even available because the company went bankrupt, we just have to remove the binding from the distribution. Hopefully, it does not happen too much often.

PR #14776, opened 2023-04-10

lolodomo:
Done for bindings I am maintaining or contributing.

### DIFF
--- a/bundles/org.openhab.binding.sonos/src/main/resources/OH-INF/addon/addon.xml
+++ b/bundles/org.openhab.binding.sonos/src/main/resources/OH-INF/addon/addon.xml
@@ -6,6 +6,7 @@
 	<type>binding</type>
 	<name>Sonos Binding</name>
 	<description>This is the binding for the Sonos multi-room audio system.</description>
+	<connection>cloud</connection>
 
 	<config-description>
 		<parameter name="opmlUrl" type="text">
--- a/bundles/org.openhab.binding.sonos/src/main/resources/OH-INF/addon/addon.xml
+++ b/bundles/org.openhab.binding.sonos/src/main/resources/OH-INF/addon/addon.xml
@@ -6,7 +6,7 @@
 	<type>binding</type>
 	<name>Sonos Binding</name>
 	<description>This is the binding for the Sonos multi-room audio system.</description>
-	<connection>cloud</connection>
+	<connection>local</connection>
 
 	<config-description>
 		<parameter name="opmlUrl" type="text">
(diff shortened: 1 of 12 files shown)


lolodomo:
Note: the Sonos binding uses mainly a local API but needs Internet for downloading album art.

bundles/org.openhab.binding.sonos/src/main/resources/OH-INF/addon/addon.xml
```diff
@@ -6,6 +6,7 @@
 	<type>binding</type>
 	<name>Sonos Binding</name>
 	<description>This is the binding for the Sonos multi-room audio system.</description>
+	<connection>cloud</connection>
```
jlaur, line 9: Shouldn't this be:
```suggestion
	<connection>local</connection>
```

?

jlaur (changes requested):
👍 - only thing spotted.

bundles/org.openhab.binding.sonos/src/main/resources/OH-INF/addon/addon.xml
jlaur: And only now read the note you already put. 😄 I think we can declare this as local, even if some limited functionality extends into the cloud. If you remove the internet connection, this binding will still work.

lolodomo: Look at my second message.

jlaur: The description puts us between two choices since it neither requires cloud to run (thus local) or only interacts locally (thus not local):
https://next.openhab.org/docs/developer/addons/addon.html#xml-structure-for-add-on-definitions

I would still opt for local here, but @kaikreuzer / @J-N-K - maybe we already need to further refine these options, or adapt documentation slightly? Hybrid bindings seem not fully covered.

lolodomo: The definition is in https://github.com/openhab/openhab-core/issues/2058

> highlight if a binding/automation/... interacts locally or if it connects to an online service (seldom or continuously)

lolodomo: Maybe we should create an option that would say that the binding can work without Internet but then with certain degraded features ?

lolodomo: > I think we can declare this as local, even if some limited functionality extends into the cloud.

Done

jlaur (approved):
LGTM

bundles/org.openhab.binding.sonos/src/main/resources/OH-INF/addon/addon.xml
jlaur: Leaving this comment open. We can merge this PR, but we still need to at least clarify the documentation, since Sonos currently doesn't match any of the documented options:
- `local` for add-ons that only interact locally
- `cloud` for add-ons that require a cloud connection
- `cloudDiscovery` for add-ons that require a cloud connection for set-up

We are closest to `local`, but the word "only" disqualifies when reading this strictly.

PR #14777, opened 2023-04-10

holgerfriedrich:
Inspired by #14776.


jlaur (approved):
👍

holgerfriedrich:
@lolodomo can you comment how the attribute connection should be interpreted? For example: does it apply to the exec binding which does not connect at all (->"local"?)....

lolodomo:
The definition is in https://github.com/openhab/openhab-core/issues/2058

> highlight if a binding/automation/... interacts locally or if it connects to an online service (seldom or continuously)

PR #14778, opened 2023-04-10
[dsmr] Use lower-case country codes in addon.xml

wborn:
When using upper-case country codes the XSD pattern does not match and IDEs will show errors for this.

See: https://github.com/openhab/website/blob/257652aafc8a4ff56a2d75198f910abba7445a3d/.vuepress/public/schemas/addon-1.0.0.xsd#L60-L64

error

Caused by: #14380

### DIFF
--- a/bundles/org.openhab.binding.dsmr/src/main/resources/OH-INF/addon/addon.xml
+++ b/bundles/org.openhab.binding.dsmr/src/main/resources/OH-INF/addon/addon.xml
@@ -7,5 +7,5 @@
 	<name>DSMR Binding</name>
 	<description>This binding integrates Dutch, Belgium, Luxembourg and Austrian Smart Meters</description>
 	<connection>local</connection>
-	<countries>NL,BE,LU,AT</countries>
+	<countries>nl,be,lu,at</countries>
 </addon:addon>


jlaur:
@wborn - whoops. But shouldn't we rather fix the XSD? The two-letter ISO country codes are uppercase.

jlaur:
> @wborn - whoops. But shouldn't we rather fix the XSD? The two-letter ISO country codes are uppercase.

Or actually, it seems that they might be case insensitive, but usually written in uppercase?

wborn:
Perhaps @kaikreuzer can comment on why the lower-case pattern was added in https://github.com/openhab/website/pull/387?

jlaur:
> Perhaps @kaikreuzer can comment on why the lower-case pattern was added in [openhab/website#387](https://github.com/openhab/website/pull/387)?

For consistency with #14786 I'll merge this now, and we can come back to this when the issue is settled.

jlaur (approved):

J-N-K:
Sorry for being late to the party. Since I authored the `addon.xsd` I can say: there was no particular reason. I just used the same convention (two lower-case letters) that we use for language in i18n.

jlaur:
@J-N-K - okay, let's keep it simple and use the same case always. I have created openhab/openhab-docs#2066.

PR #14784, opened 2023-04-11
[Netatmo] Fix remaining warnings at Room startup

clinique:
Delayed initialization of the HomeEnergyCapability


jlaur:
Regression of #14595?

clinique:
No, RoomCapability was not involved in this PR - and these warning may be masked depending upon thing initialization speed/order I guess. I never faced them in my dev environment - they appeared when I switched production on OH4.

jlaur:
@clinique - so probably still a regression in 4.0, or do you think the problem also existed in 3.4?

clinique:
I think it exists since a while but has gone unnoticed. Maybe the new thing initialization mechanism of OH4 made him visible.

lolodomo (approved):
LGTM

PR #14787, opened 2023-04-11

hmerk:
Done for bindings I am maintaining.


jlaur (approved):
Thanks!

jlaur:
@hmerk - I don't know why DCO succeeded, but it seems like your commits are not signed off?

hmerk:
> @hmerk - I don't know why the DCO passed, but it seems like your commits are not signed off? I see this: image

Cause I manually set it to pass after it did not when I created the PR via web editing.

jlaur:
> Cause I manually set it to pass after it did not when I created the PR via web editing.

I'm not sure if there is a GitHub option for configuring a sign-off for the web editor, but otherwise you can manually add the sign-off message when preparing your commit. Manually approving the DCO without sign-off doesn't really solve the problem. 🙂

hmerk:
> Manually approving the DCO without sign-off doesn't really solve the problem.

It does, as you can see, the DCO check is green.
I normally don't do web edits, but prefer proper PR's...
But will check how to set the options right.

PR #14790, opened 2023-04-12
[tapocontrol] Moved error messages to i18n

wildcs:
before this pr error messages were declared as constant.
moved them to i18n


bundles/org.openhab.binding.tapocontrol/src/main/java/org/openhab/binding/tapocontrol/internal/helpers/TapoErrorHandler.java
```diff
@@ -90,27 +90,25 @@ public TapoErrorHandler(Exception ex, String infoMessage) {
...
-                    Field constantName = TapoErrorConstants.class.getDeclaredField(constName + "_MSG");
-                    String msg = getValueOrDefault(constantName.get(null), "").toString();
-                    if (msg.length() > 2) {
-                        return msg;
-                    } else {
-                        return infoMessage + " (" + constName + ")";
-                    }
+                    /* convert constatn name to i18n key */
+                    i18nKey = i18nKey.replaceFirst("ERR_", "error-");
+                    i18nKey = constName.replace("_", "-").toLowerCase();
```
lolodomo, line 107: Looks like these two lines are in the wrong order.
With your current code, `i18nKey` is first a simple empty string so the first line will do nothing.

bundles/org.openhab.binding.tapocontrol/src/main/resources/OH-INF/i18n/tapocontrol.properties
```diff
@@ -163,3 +163,64 @@ channel-type.tapocontrol.effectName.label = Effect Name
...
+
+# error messages
+
+error-api-account = received error code (-2101);
+error-api-aes-decode-fail = aes decode failed (-1005)
+error-api-antitheft-conflict = received error code (-2002);
+error-api-antitheft = received error code (-2001);
+error-api-antitheft-save = received error code (-2003);
+error-api-cloud-failed = cloud failed (-1007)
+error-api-common-failed = ;
```
lolodomo, line 175: Strange to have no value. Maybe this error is not required ?

bundles/org.openhab.binding.tapocontrol/src/main/resources/OH-INF/i18n/tapocontrol.properties
```diff
@@ -163,3 +163,64 @@ channel-type.tapocontrol.effectName.label = Effect Name
 channel-type.tapocontrol.effectName.description = Name of LightningEffect
 channel-type.tapocontrol.effectOn.label = Lightning Effect Enable
 channel-type.tapocontrol.effectOn.description = Switches the lightning effect on/off
+
+# error messages
+
+error-api-account = received error code (-2101);
+error-api-aes-decode-fail = aes decode failed (-1005)
+error-api-antitheft-conflict = received error code (-2002);
```
lolodomo, line 171: Why defining error messages if finally the message is always the same with only difference a number ?
In that case, you just need one entry with a parameter.

bundles/org.openhab.binding.tapocontrol/src/main/resources/OH-INF/i18n/tapocontrol.properties
```diff
@@ -163,3 +163,64 @@ channel-type.tapocontrol.effectName.label = Effect Name
...
+error-api-countdown-conflict = received error code (-1902);
+error-api-countdown = received error code (-1901);
+error-api-countdown-save = received error code (-1903);
+error-api-device = received error code (-1301);
+error-api-device-next-event = received error code (-1302);
+error-api-dst = received error code (-2301);
+error-api-dst-save = received error code (-2302);
+error-api-firmware = received error code (-1401);
+error-api-firmware-ver-error = received error code (-1402);
+error-api-hand-shake-failed = can't create handshake (1100)
```
lolodomo, line 185: Why some messages ends you a ";" ?

bundles/org.openhab.binding.tapocontrol/src/main/resources/OH-INF/i18n/tapocontrol.properties
```diff
@@ -163,3 +163,64 @@ channel-type.tapocontrol.effectName.label = Effect Name
...
+error-api-http-transport-failed = http transport failed (1112)
+error-api-json-decode-fail = json decode failed (-1003)
+error-api-json-encode-fail = json encode failed (-1004)
+error-api-login-failed = login failed (1111)
+error-api-login = invalid request or credentials (-1501);
+error-api-multi-request-failed = multi request failed (1200)
+error-api-null-transport = -1
```
lolodomo, line 192: This message looks strange. No need to translate a simple number.

lolodomo (changes requested):

lolodomo:
Your approach looks strange. If you want to provide generic messages with a number as parameter, you do not need to duplicate entries to be translated. Create one with a parameter, something like `received error code ({0})`

lolodomo:
And your way to retrieve the constant name to get the i18n entry is not very good.
Why not simply create an enum type containing all your errors ? Then you just have to retrieve the enum name from an int value.

bundles/org.openhab.binding.tapocontrol/src/main/java/org/openhab/binding/tapocontrol/internal/helpers/TapoErrorHandler.java
lolodomo: Ok, I understood what you try to achieve.
I would then suggest only one line:
```
                     i18nKey = constName.replaceFirst("ERR_", "error-").replace("_", "-").toLowerCase();
```

wildcs:
Thanks for your hint. Changed it and hope it's okay now.

bundles/org.openhab.binding.tapocontrol/src/main/java/org/openhab/binding/tapocontrol/internal/helpers/TapoErrorHandler.java
```diff
@@ -90,30 +106,13 @@ public TapoErrorHandler(Exception ex, String infoMessage) {
...
-                    }
-                }
-            } catch (Exception e) {
-                // next loop
-            }
-        }
-        return infoMessage + " (" + errCode.toString() + ")";
+        String key = TapoErrorCode.fromCode(errCode).name().replace("ERR_", "error-").replace("_", "-").toLowerCase();
+        String message = "@text/" + key;
+        return message.replace("%CODE", errCode.toString());
```
lolodomo, line 115: I do not understand your replace.
Just before message will have for example this value: `@text/error-config-no-bridge`
I can't understand how it could contain `%CODE`

lolodomo (changes requested):
Looks clearly better.

bundles/org.openhab.binding.tapocontrol/src/main/resources/OH-INF/i18n/tapocontrol.properties
```diff
@@ -163,3 +163,64 @@ channel-type.tapocontrol.effectName.label = Effect Name
...
+error-api-request-len-error = request length error (-1006)
+error-api-request = invalid request or command (1002)
+error-api-rsa-key-length = invalid public key length (-1010)
+error-api-schedule-conflict = schedule conflict error (-1803)
+error-api-schedule-full = schedule full error (-1802)
+error-api-schedule-index = schedule index error (-1805)
+error-api-schedule = schedule error (-1801)
+error-api-schedule-save = saving schedule rule failed (-1804)
+error-api-session-param = invalid session key (-1101)
+error-api-session-timeout = "device session expired (9999)"
```
lolodomo, line 203: Quotes to be removed

bundles/org.openhab.binding.tapocontrol/src/main/resources/OH-INF/i18n/tapocontrol.properties
```diff
@@ -163,3 +163,64 @@ channel-type.tapocontrol.effectName.label = Effect Name
...
+error-api-stat = received error code (-2201)
+error-api-stat-save = received error code (-2202)
```
lolodomo, line 205: Could you differentiate the errors ?

bundles/org.openhab.binding.tapocontrol/src/main/resources/OH-INF/i18n/tapocontrol.properties
```diff
@@ -163,3 +163,64 @@ channel-type.tapocontrol.effectName.label = Effect Name
...
+error-api-time = device time error (-1601)
+error-api-time-save = saving time failed (-1603)
+error-api-time-sys = time system error (-1602)
+error-api-cloud-token-expired = cloud token expired or invalid
```
lolodomo, line 209: It is the only one without the code.

lolodomo (changes requested):

bundles/org.openhab.binding.tapocontrol/src/main/resources/OH-INF/i18n/tapocontrol.properties
lolodomo: You changed nothing. Please provide a message that makes a sense for a user.

bundles/org.openhab.binding.tapocontrol/src/main/java/org/openhab/binding/tapocontrol/internal/helpers/TapoErrorHandler.java
lolodomo: I understand what you try to do, it is for `error-unknown`.
You should apply the replace not on the key but on the value returned by i18n request.

lolodomo: Instead of `%CODE` you should use `{0]` in the properties file.
And then you have a special syntax to use to provide the parameter. If I correctly remember (to be verified), it should be something like : `@text/error-unknown [ "xxx" ]` where "xxx" is your number.

lolodomo: Something like that should work:
```
String message = "@text/" + key;
if (message.contains("{0}") {
    message += " [ \"" + errCode.toString() + "\" ]";
}
```

lolodomo: In fact, you could even use {0} in all your i18n entries, rather than having the code hardcoded in the i18n entries.
I mean for example for error-config-no-bridge: `no bridge configured ({0})`
To be tested with at least one error to be sure that the proposed syntax is correct.

wildcs: It's working exactyl as you said. Testet it. Many Thanks again.
And of course in fact i can use `{0}` in any line, but if i there is it necesarry to assign a code to an enum (for debugging, etc)  it would be easyer if it's hardcoded.

lolodomo (approved):
LGTM, thank you

PR #14807, opened 2023-04-14
[windcentrale] Discover windmills when account goes online

wborn:
This makes the binding automatically discover windmills when the account goes online so users do not have to manually trigger a scan during initial setup.


jlaur (approved):
Thanks - LGTM!

PR #14371, opened 2023-02-09
[Documentation] Markdown improvements t to z

Confectrician:
This is a bulk pr for getting some consistency into the doc markdown files.

Reference for a general linting approach here:  #13858
Discussion Reference for codeblock styles i have applied: openhab/openhab-docs#999

(Sorry again for the many review requests, there is plenty much to improve in the repo. 🤷‍♂️)

Issue #13405, opened 2022-09-18
Script to create new binding skeleton: problem with file bom/pom.xml


lolodomo:
The script to create a new binding skeleton is generating a lot of changes in the file bom/pom.xml. Only one line should be added (for the new binding) but in fact it generates a lot of more changes, certainly for formatting reason.

Tested with the Windows bat command file.

wborn:
Do you really mean `bom/pom.xml` or those new lines that are added to `bundles/pom.xml`?

lolodomo:
You,'re right, it is bundles/pom.xml. I correct issue title and description,

lolodomo:
Look at PR #13400 for an example.

wborn:
This will be fixed when the following issues are fixed in the Maven archetype plugin:

* https://issues.apache.org/jira/browse/ARCHETYPE-584
* https://issues.apache.org/jira/browse/ARCHETYPE-587

wborn:
Is  this still an issue for you @lolodomo now that [ARCHETYPE-637](https://issues.apache.org/jira/browse/ARCHETYPE-637) got fixed in 3.3.0 of the maven-archetype-plugin?
The scripts don't specify a version so they use the latest version.

lolodomo:
I can try during the coming weekend.

lsiepel:
I just created a new binding with the skeleton. It seems to have been improved, but i see some issues remaining:
1. Many newlines are no longer added, but now there are more removed, for instance line 714 between end tag profiles and end tag project.
2. The first three lines are combined onto one line (xml and project tags) and not broken apart
3. The module tag is not added: `<module>org.openhab.binding.testbinding</module>`

lolodomo:
I just tried again.
To build the new generated binding, I have first to run "mvn spotless:apply" in the new binding folder andthen "mvn clean install" is working.

Regarding the file bundles/pom.xml, I can see that the new binding is now no more added to bundles/pom.xml.
So the initial problem is solved but a new is introduced.

I also noticed that CODEOWNERS is not updated, I don't know if it was in the past.

wborn:
The Spotless issue is fixed with https://github.com/openhab/openhab-core/pull/4851.

PR #14809, opened 2023-04-14
[chatgpt] Initial contribution of the ChatGPT binding

kaikreuzer:
Everyone is talking about ChatGPT and thus openHAB also needs a binding for it!

I see an opportunity to personalize the interaction with the home by replacing the "static" announcements that many people use in the home for different purposes. ChatGPT allows bringing in emotions and moods and ensures that it isn't always the very same text, but that everything becomes much more personal and individual.

The examples in the README show how it can be used to have a personal advice for the day based on the weather or to have a morning message announcing the time (e.g. for an alarm clock use case).

There are certainly many more things one could think of and I am looking forward to what the community makes out of it!


bundles/org.openhab.binding.chatgpt/src/main/java/org/openhab/binding/chatgpt/internal/ChatGPTHandler.java
```diff
@@ -0,0 +1,200 @@
...
+        }
+    }
+
+    @Override
+    public void initialize() {
+        ChatGPTConfiguration config = getConfigAs(ChatGPTConfiguration.class);
+
+        String apiKey = config.apiKey;
+
+        if (apiKey == null || apiKey.isEmpty()) {
```
jlaur, line 153: See comment in `ChatGPTConfiguration`.
```suggestion
        if (apiKey.isBlank()) {
```

bundles/org.openhab.binding.chatgpt/src/main/java/org/openhab/binding/chatgpt/internal/ChatGPTHandler.java
```diff
@@ -0,0 +1,200 @@
...
+            updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.CONFIGURATION_ERROR, "No API key configured");
```
jlaur, line 154: Perhaps this text could be provided as I18N string for localization?

bundles/org.openhab.binding.chatgpt/src/main/java/org/openhab/binding/chatgpt/internal/ChatGPTHandler.java
```diff
@@ -0,0 +1,200 @@
...
+                            String id = model.get("id").getAsString();
+                            modelIds.add(id);
+                        }
+                        this.models = Collections.unmodifiableList(modelIds);
+                    } else {
+                        logger.warn("Did not receive a valid JSON response from the models endpoint.");
+                    }
+                } else {
+                    updateStatus(ThingStatus.OFFLINE, ThingStatusDetail.COMMUNICATION_ERROR,
+                            "Could not connect to OpenAI API");
```
jlaur, line 184: Consider I18N.

bundles/org.openhab.binding.chatgpt/src/main/resources/OH-INF/addon/addon.xml
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<addon:addon id="chatgpt" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+	xmlns:addon="https://openhab.org/schemas/addon/v1.0.0"
+	xsi:schemaLocation="https://openhab.org/schemas/addon/v1.0.0 https://openhab.org/schemas/addon-1.0.0.xsd">
+
+	<type>binding</type>
+	<name>ChatGPT Binding</name>
+	<description>This is the binding for ChatGPT integration.</description>
```
jlaur, line 8: Can we add `connection` here?
```suggestion
	<description>This is the binding for ChatGPT integration.</description>
	<connection>cloud</connection>
```

See https://next.openhab.org/docs/developer/addons/addon.html#xml-structure-for-add-on-definitions

bundles/org.openhab.binding.chatgpt/src/main/resources/OH-INF/i18n/chatgpt.properties
jlaur, line 1: Please generate this file, see https://www.openhab.org/docs/developer/utils/i18n.html#generating-i18n-properties-file.

bundles/org.openhab.binding.chatgpt/src/main/resources/OH-INF/thing/thing-types.xml
jlaur, line 1: It seems there is a mix of spaces and tabs in this file, can you replace each tab by four spaces?

bundles/pom.xml
jlaur, line 1: Something went wrong with this file.

bundles/org.openhab.binding.chatgpt/README.md
```diff
@@ -0,0 +1,104 @@
...
+| chat    | String | RW         | This channel takes prompts as commands and delivers the response as a state update |
+
+Each channel of type `chat` takes the following configuration parameters:
+
+| Name            | Type    | Description                             | Default | Required | Advanced |
+|-----------------|---------|-----------------------------------------|---------|----------|----------|
+| model           | text    | The model to be used for the responses. | N/A     | yes      | no       |
+| temperature     | decimal | A value between 0 and 2. Higher values like 0.8 will make the output more random, while lower values like 0.2 will make it more focused and deterministic. | 0.5   | no      | no       |
+| systemMessage   | text    | The system message helps set the behavior of the assistant. | N/A     | no      | no       |
+| maxTokens       | decimal | The maximum number of tokens to generate in the completion. | 500     | no      | yes      |
```
jlaur, line 37: If this table would be properly aligned, it would be easier to review in source form.

bundles/org.openhab.binding.chatgpt/src/main/java/org/openhab/binding/chatgpt/internal/ChatGPTChannelConfiguration.java
```diff
@@ -0,0 +1,37 @@
...
+    public String model;
+
+    @Nullable
+    public Float temperature;
+
+    @Nullable
+    public String systemMessage;
+
+    @Nullable
+    Integer maxTokens;
```
jlaur, line 36: Perhaps this could be changed to primitive type to avoid boxing/null.

jlaur (changes requested):
Incredible - thanks! I have provided some quick feedback, but might need another look tomorrow. 🙂

digitaldan:
@kaikreuzer you beat me too it !!! 

FYI I was also hoping to eventually support local LLMs like the LLaMa variations ( Alpaca, GTP4All, Vicuna) .  I have an obsession with running them locally right now ;-)  

Can't wait to try this out.

digitaldan:
Quick comment after trying it once. Should there be two channels, one for sending a prompt, one for receiving a prompt?  Think of it as a "user" channel and a "assistant" channel.  Right now the channels state changes from "null/previous answer -> question -> answer"  over the course of a second, which will make it hard to monitor in rules for reacting to the response which is what matters.

The other possibility is to add `<autoUpdatePolicy>veto</autoUpdatePolicy>` to the channel configuration, so only the response from openAI can update that channel .  This would work nicely as well, and keep the single channel design

bundles/org.openhab.binding.chatgpt/src/main/java/org/openhab/binding/chatgpt/internal/ChatGPTConfiguration.java
```diff
@@ -0,0 +1,28 @@
...
+/**
+ * The {@link ChatGPTConfiguration} class contains fields mapping thing configuration parameters.
+ *
+ * @author Kai Kreuzer - Initial contribution
+ */
+@NonNullByDefault
+public class ChatGPTConfiguration {
+
+    @Nullable
+    public String apiKey = "";
```
wborn, line 27:
```suggestion
    public String apiKey = "";
```

bundles/org.openhab.binding.chatgpt/src/main/java/org/openhab/binding/chatgpt/internal/ChatGPTHandler.java
```diff
@@ -0,0 +1,200 @@
...
+
+    private final Logger logger = LoggerFactory.getLogger(ChatGPTHandler.class);
+
+    private HttpClient httpClient;
+    private Gson gson = new Gson();
+
+    private String apiKey = "";
+    private String lastPrompt = "";
+
+    private List<String> models = Collections.emptyList();
```
wborn, line 69:
```suggestion
    private List<String> models = List.of();
```

bundles/org.openhab.binding.chatgpt/src/main/java/org/openhab/binding/chatgpt/internal/ChatGPTModelOptionProvider.java
```diff
@@ -0,0 +1,67 @@
...
+ * The {@link ChatGPTModelOptionProvider} provides the available models from OpenAI as options for the channel
+ * configuration.
+ *
+ * @author Kai Kreuzer - Initial contribution
+ */
+@NonNullByDefault
+public class ChatGPTModelOptionProvider implements ThingHandlerService, ConfigOptionProvider {
+
+    @Nullable
+    private ThingHandler thingHandler;
```
wborn, line 38:
```suggestion
    private @Nullable ThingHandler thingHandler;
```

bundles/org.openhab.binding.chatgpt/src/main/java/org/openhab/binding/chatgpt/internal/ChatGPTHandler.java
```diff
@@ -0,0 +1,200 @@
...
+                    if (jsonObject != null) {
+                        JsonArray data = jsonObject.getAsJsonArray("data");
+
+                        List<String> modelIds = new ArrayList<>();
+                        for (JsonElement element : data) {
+                            JsonObject model = element.getAsJsonObject();
+                            String id = model.get("id").getAsString();
+                            modelIds.add(id);
+                        }
+                        this.models = Collections.unmodifiableList(modelIds);
```
wborn, line 178:
```suggestion
                        this.models = List.of(modelIds);
```

wborn (changes requested):
Looks like a lot of fun without not that much code. :slightly_smiling_face: 
I have added few minor comments below.

bundles/org.openhab.binding.chatgpt/src/main/java/org/openhab/binding/chatgpt/internal/ChatGPTChannelConfiguration.java
```diff
@@ -0,0 +1,37 @@
...
+/**
+ * The {@link ChatGPTChannelConfiguration} class contains fields mapping chat channel configuration parameters.
+ *
+ * @author Kai Kreuzer - Initial contribution
+ */
+@NonNullByDefault
+public class ChatGPTChannelConfiguration {
+
+    @Nullable
+    public String model;
```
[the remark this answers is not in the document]
kaikreuzer: But neither `model` nor `systemMessage` are required parameter...

bundles/org.openhab.binding.chatgpt/src/main/java/org/openhab/binding/chatgpt/internal/ChatGPTHandler.java
kaikreuzer: I guess you meant `List.copyOf`.

bundles/pom.xml
kaikreuzer: Which is weird, since this was generated by our binding creation script...
But I have fixed it now manually.

wborn: See https://github.com/openhab/openhab-addons/issues/13405

kaikreuzer:
Thanks for your immensely quick reviews @jlaur and @wborn! I have addressed your comments now.

> The other possibility is to add <autoUpdatePolicy>veto</autoUpdatePolicy> to the channel configuration, so only the response from openAI can update that channel .

@digitaldan This is exactly what I did, see [here](https://github.com/openhab/openhab-addons/pull/14809/files#diff-4a0c22cc3d6dcc4c484fc42659fd338b1bf1bd9f67e2df51d7aa935a43c77663R29).
This is imho much better than two separate channels - note that the thing is extensible, so you can add additional "chat" channels with different configurations (see the example in the README). This wouldn't be easily possible if there were separate channels for prompt and response.

> Right now the channel state changes from "previous answer -> question -> new answer" over the course of a second

If this really happens for you right now, then something is buggy; it isn't the case for me.

> FYI I was also hoping to eventually support local LLMs like the LLaMa variations ( Alpaca, GTP4All, Vicuna)

I tried Alpaca myself already as well, but decided that those local LLMs are probably rather something for a very few users, so I didn't try to generify this binding to support all kinds of LLMs.

bundles/org.openhab.binding.chatgpt/src/main/java/org/openhab/binding/chatgpt/internal/dto/ChatResponse.java
```diff
@@ -0,0 +1,81 @@
...
+        return object;
+    }
+
+    public String getModel() {
+        return model;
+    }
+
+    public static class Choice {
+        private Message message;
+        private String finish_reason;
```
wborn, line 53: You can fix this SAT naming issue by using `@SerializedName(finish_reason")` or creating a `Gson` instance using:

```java
new GsonBuilder().setFieldNamingPolicy(FieldNamingPolicy.LOWER_CASE_WITH_UNDERSCORES).create()
```

bundles/org.openhab.binding.chatgpt/README.md
```diff
@@ -0,0 +1,104 @@
...
+end
+```
+
+Assuming that `Temperature_Forecast_Low` and `Temperature_Forecast_High` have meaningful states, these rules result e.g. in:
+
+```
+23:31:05.766 [INFO ] [openhab.event.ItemCommandEvent      ] - Item 'Morning_Message' received command Current time is 7am
+23:31:07.718 [INFO ] [openhab.event.ItemStateChangedEvent ] - Item 'Morning_Message' changed from NULL to Good morning. It's 7am, but what's the point of time when everything is meaningless and we are all doomed to a slow and painful demise?
+```
+and
```
wborn, line 95:
```suggestion

and
```

wborn (commented):
These SAT issues should also be easy to fix:

kaikreuzer:
@wborn I've addressed your comments.

bundles/org.openhab.binding.chatgpt/src/main/java/org/openhab/binding/chatgpt/internal/ChatGPTChannelConfiguration.java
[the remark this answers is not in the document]
wborn: `model` has a default value so it will probably never be `null`.

The `systemMessage` will be `null` if is not set, but then you have to check both `null` and the empty string which can also be provided by users?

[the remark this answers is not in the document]
kaikreuzer: Ok, I removed all Nullables.

wborn (approved):
Looks nice and clean now! :+1:

bundles/org.openhab.binding.chatgpt/README.md
```diff
@@ -0,0 +1,105 @@
...
+With this binding, you can use ChatGPT to formulate proper sentences for any kind of information that you would like to output.
+
+## Supported Things
+
+The binding supports a single thing type `account`, which corresponds to the OpenAI account that is to be used for the integration.
+
+## Thing Configuration
+
+The `account` thing requires a single configuration parameter, which is the API key that allows accessing the account.
+API keys can be created and managed under https://platform.openai.com/account/api-keys.
```
J-N-K, line 15: Should this be a link?

bundles/org.openhab.binding.chatgpt/README.md
```diff
@@ -0,0 +1,105 @@
...
+
+| Name            | Type    | Description                             | Default | Required | Advanced |
+|-----------------|---------|-----------------------------------------|---------|----------|----------|
+| apiKey          | text    | The API key to be used for the requests | N/A     | yes      | no       |
```
J-N-K, line 19:
```suggestion
| `apiKey`          | text    | The API key to be used for the requests | N/A     | yes      | no       |
```

In a lot of bindings (and I believe this is some sort of unwritten convention) we use code fences for channel/parameter names.

J-N-K (commented):
Just two very minor comments

bundles/org.openhab.binding.chatgpt/README.md
kaikreuzer: I agree in situations where the parameters are mentioned in text, but not within the table - this column is clearly meant for the parameter name, so I do not think it needs to be made explicit by code fences.

Note that our archetype is not suggesting code fences in these tables, see https://github.com/openhab/openhab-core/blob/main/tools/archetype/binding/src/main/resources/archetype-resources/README.md?plain=1#L52-L56.

bundles/org.openhab.binding.chatgpt/README.md
kaikreuzer: Won't this be automatically rendered as a link?

digitaldan:
>@digitaldan This is exactly what I did, see [here](https://github.com/openhab/openhab-addons/pull/14809/files#diff-4a0c22cc3d6dcc4c484fc42659fd338b1bf1bd9f67e2df51d7aa935a43c77663R29).

Ah, sorry did not check the code, i was quickly looking at the logs which look like this

```
11:30:28.872 [INFO ] [openhab.event.ItemCommandEvent       ] - Item 'OpenAIChat' received command FooBar
11:30:28.874 [INFO ] [openhab.event.ItemStatePredictedEvent] - Item 'OpenAIChat' predicted to become FooBar
11:30:28.878 [INFO ] [openhab.event.ItemStateUpdatedEvent  ] - Item 'OpenAIChat' updated to FooBar
11:30:28.879 [INFO ] [openhab.event.ItemStateChangedEvent  ] - Item 'OpenAIChat' changed from I'm sorry, but there is no input to summarize. Please provide a valid input to summarize. to FooBar
11:30:30.074 [INFO ] [openhab.event.ItemStateUpdatedEvent  ] - Item 'OpenAIChat' updated to I'm sorry, but I cannot summarize the input "FooBar" as it does not contain any meaningful information.
11:30:30.075 [INFO ] [openhab.event.ItemStateChangedEvent  ] - Item 'OpenAIChat' changed from FooBar to I'm sorry, but I cannot summarize the input "FooBar" as it does not contain any meaningful information.
```

You can see that the channel has both a change event and an update event to "FooBar" which is the command i posted.   Then the channel has another change event and update event with the openAI response.    I have not looked into this, I guess it could be something i did when setting the item up (although i used the UI to do so) .

bundles/org.openhab.binding.chatgpt/README.md
jlaur: I believe it should be surrounded by `<` and `>` now to avoid linting, but I don't know the details. At least @Confectrician made a great effort changing this for all README's recently, see for example #14371.
```suggestion
API keys can be created and managed under <https://platform.openai.com/account/api-keys>.
```

kaikreuzer:
@jlaur I've fixed the link syntax.

kaikreuzer:
@digitaldan That's interesting, I do not have the ItemStateUpdatedEvent/ItemStateChangedEvent.
What I see as well, though, is the ItemStatePredictedEvent, which I also wanted to investigate, why this happens although there is a "veto" by the binding - I guess there could be a bug in the core framework due to some recent refactorings.

bundles/org.openhab.binding.chatgpt/README.md
kaikreuzer: Fixed.

jlaur (approved):
LGTM

digitaldan:
> I guess there could be a bug in the core framework due to some recent refactorings.

So i just checked another binding that uses the veto policy, its correctly not triggering a Predicted/Updated event when a command is posted.  Not sure why this binding would be different, or why you see a Predicted event, but not Updated/Changed like i do.   I went ahead and recreated the item in a file, since this is the only obvious difference to me  with the rest of my system ( i use files for all my items still), but that did not change the behavior (not surprising) .

Dustinhoefer:
Really looking forward to this! Thank you for your work!

jlaur:
@kaikreuzer, @digitaldan - I did not merge yet since I'm not sure if you are looking into event issues? Let me know when you think it's fine to be merged.

kaikreuzer:
@jlaur I just tested it once more and it all behaves as expected for me:
```
11:40:28.338 [INFO ] [openhab.event.ItemCommandEvent      ] - Item 'OpenAI_Account_Chat' received command Test
11:40:29.950 [INFO ] [openhab.event.ItemStateChangedEvent ] - Item 'OpenAI_Account_Chat' changed from NULL to This is a test response from the AI language model.
11:40:49.094 [INFO ] [openhab.event.ItemCommandEvent      ] - Item 'OpenAI_Account_Chat' received command Test
11:40:50.352 [INFO ] [openhab.event.ItemStateChangedEvent ] - Item 'OpenAI_Account_Chat' changed from This is a test response from the AI language model. to Hello! How can I assist you with your test?
```
I would therefore say that everything is fine and the PR can be merged.
If there are still issues for @digitaldan, I would actually consider those to be rather a bug in core, which we should then handle there.

kaikreuzer:
> If there are still issues for @digitaldan, I would actually consider those to be rather a bug in core, which we should then handle there.

@digitaldan I actually now had the same effect (veto not being considered) on my production system. I tracked it down to some issues in the core indeed - when the Main UI creates things through the REST API, it does not set autoUpdatePolicy for the channels, which makes the core use "RECOMMEND" as a policy without looking at a potential information at the channel type definition. I have created https://github.com/openhab/openhab-core/pull/3575, which hopefully fixes this (but faulty channels will need to be recreated).

kaikreuzer:
Hi @Artur-Fedjukevits,

This looks like some encoding issue. I've checked the code and I couldn't find any obvious error in it - we process strings as UTF-8 and also consider content-type headers, if ChatGPT returns something different. I'd need to deeper debug this with your example.

Could you maybe open a dedicated issue about this, so that it can be tracked and isn't lost here on the merged PR?

mudler:
@kaikreuzer that's great addition! - sorry a bit late in the game here. Is there any chance to make this binding work with LocalAI? I can help along as I'm the author and an avid openhab user willing to integrate it in my domotic system :)

I'm reading from the docs that it is possible to just configure an `apiKey` - https://www.openhab.org/addons/bindings/chatgpt/#thing-configuration. However, for https://github.com/go-skynet/LocalAI to work would be enough just exposing the openai base url to the Thing configuration.

Also - any plans maybe we can use LocalAI as a TTS engine, or Audio-to-text?

kaikreuzer:
Hey @mudler, thanks for your great work on LocalAI. I didn't yet come across it, but since you say it is API-compatible to OpenAI, it should indeed be very easy to add support for it by making the server url configurable.

> Also - any plans maybe we can use LocalAI as a TTS engine, or Audio-to-text? 

I'm not familiar with the possibilities of LocalAI here. On openHAB side, @GiviMAD is the person who is most active in TTS & STT topics, so he might be interested in such new possibilities.

kaikreuzer:
I've created https://github.com/openhab/openhab-addons/pull/15385 as a starting point.

GiviMAD:
Hey @mudler, first of all great project.

I have an open PR for using whisper.cpp for speech-to-text https://github.com/openhab/openhab-addons/pull/15166, maybe we can add there an option to switch between the library and the LocalAI api.

But in case you prefer to create a separate add-on to expose in a single place all the capabilities than LocalAI can provide to openHAB, don't hesitate to steal any code in the PR that can help you.

I'm planning to create an add-on for using llama.cpp within openHAB but I have no idea how long it could take me, it's great to have this available.

BR

mudler:
> I've created #15385 as a starting point. Since I'm on the road right now, I unfortunately cannot test with LocalAI. @mudler Would be great if you could help on that PR to make it work!

fantastic, had a look already, seems good here! It's enough to allow changing the urls - as it is 1:1 compatible with OpenAI nothing else should be needed and it just "works"


> Hey @mudler, first of all great project.
> 
> I have an open PR for using whisper.cpp for speech-to-text #15166, maybe we can add there an option to switch between the library and the LocalAI api.

That's looking great, amazing job! 

> 
> But in case you prefer to create a separate add-on to expose in a single place all the capabilities than LocalAI can provide to openHAB, don't hesitate to steal any code in the PR that can help you.

I'd be happy to help but my java-fu it's very limited and my bandwidth for it as well, however there are TTS capabilities (besides voice-to-text) and OpenAI functions as well that can be a nice addition that might be a good fit for OpenHAB.

PR #14810, opened 2023-04-14
[mqtt][tradfri][webthing] Adapt to ColorUtil changes in core

holgerfriedrich:
Adapt tests due to https://github.com/openhab/openhab-core/pull/3542.
To be merged after core change.

This PR fixes the broken tests after core change for now, but especially the tests checking for double valued numbers should be changed to allow small differences between colors (e.g. a deviation of 0.5% due to color conversions),

### DIFF
--- a/bundles/org.openhab.binding.mqtt.generic/src/main/java/org/openhab/binding/mqtt/generic/values/ColorValue.java
+++ b/bundles/org.openhab.binding.mqtt.generic/src/main/java/org/openhab/binding/mqtt/generic/values/ColorValue.java
@@ -28,6 +28,7 @@ import org.openhab.core.library.types.PercentType;
 import org.openhab.core.library.types.StringType;
 import org.openhab.core.types.Command;
 import org.openhab.core.types.UnDefType;
+import org.openhab.core.util.ColorUtil;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -144,14 +145,12 @@ public class ColorValue extends Value {
                 return String.format(formatPattern, hsbState.getHue().intValue(), hsbState.getSaturation().intValue(),
                         hsbState.getBrightness().intValue());
             case RGB:
-                PercentType[] rgb = hsbState.toRGB();
-                return String.format(formatPattern, rgb[0].toBigDecimal().multiply(factor).intValue(),
-                        rgb[1].toBigDecimal().multiply(factor).intValue(),
-                        rgb[2].toBigDecimal().multiply(factor).intValue());
+                int[] rgb = ColorUtil.hsbToRgb(hsbState);
+                return String.format(formatPattern, rgb[0], rgb[1], rgb[2]);
             case XYY:
-                PercentType[] xyY = hsbState.toXY();
-                return String.format(Locale.ROOT, formatPattern, xyY[0].floatValue() / 100.0f,
-                        xyY[1].floatValue() / 100.0f, hsbState.getBrightness().floatValue());
+                double[] xyY = ColorUtil.hsbToXY(hsbState);
+                return String.format(Locale.ROOT, formatPattern, xyY[0], xyY[1],
+                        hsbState.getBrightness().doubleValue());
             default:
                 throw new NotSupportedException(String.format("Non supported color mode: {}", this.colorMode));
         }
--- a/bundles/org.openhab.binding.mqtt.homeassistant/src/test/java/org/openhab/binding/mqtt/homeassistant/internal/component/JSONSchemaLightTests.java
+++ b/bundles/org.openhab.binding.mqtt.homeassistant/src/test/java/org/openhab/binding/mqtt/homeassistant/internal/component/JSONSchemaLightTests.java
@@ -76,7 +76,7 @@ public class JSONSchemaLightTests extends AbstractComponentTests {
         publishMessage("zigbee2mqtt/light/state", "{ \"color\": {\"r\": 10, \"g\": 20, \"b\": 30 } }");
         assertState(component, Light.COLOR_CHANNEL_ID, HSBType.fromRGB(10, 20, 30));
         publishMessage("zigbee2mqtt/light/state", "{ \"brightness\": 255 }");
-        assertState(component, Light.COLOR_CHANNEL_ID, new HSBType("209.99999999999999,66.6666666666666700,100"));
+        assertState(component, Light.COLOR_CHANNEL_ID, new HSBType("210,67,100"));
 
         sendCommand(component, Light.COLOR_CHANNEL_ID, HSBType.BLUE);
         assertPublished("zigbee2mqtt/light/set/state",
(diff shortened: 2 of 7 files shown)


holgerfriedrich:
@J-N-K can you please have a look?

bundles/org.openhab.binding.mqtt.homeassistant/src/test/java/org/openhab/binding/mqtt/homeassistant/internal/component/JSONSchemaLightTests.java
```diff
@@ -76,7 +76,7 @@ public void testRgb() throws InterruptedException {
         publishMessage("zigbee2mqtt/light/state", "{ \"color\": {\"r\": 10, \"g\": 20, \"b\": 30 } }");
         assertState(component, Light.COLOR_CHANNEL_ID, HSBType.fromRGB(10, 20, 30));
         publishMessage("zigbee2mqtt/light/state", "{ \"brightness\": 255 }");
-        assertState(component, Light.COLOR_CHANNEL_ID, new HSBType("210,67,100"));
+        assertState(component, Light.COLOR_CHANNEL_ID, new HSBType("209.99999999999999,66.6666666666666700,100"));
```
J-N-K, line 79: I'm not convinced that this is a good solution. We would have to adapt the test again when we decide to limit the precision to scale 10 or so. It would be better to adjust `assertState` to something like


```
   State actualState = component.getChannel(channelId).getState().getCache().getChannelState();
   if (actualState instanceof HSBType hsb) {
       assertThat(hsb.closeTo((HSBType) state, 0.01), is(true));
   } else { 
       assertThat(actual, is(state));
   }
```

bundles/org.openhab.binding.webthing/src/test/java/org/openhab/binding/webthing/internal/link/WebthingChannelLinkTest.java
```diff
@@ -140,8 +140,8 @@ public void testDataTypeMapping() throws Exception {
         performDataTypeMappingTest("targettemp_prop", 18.6, new DecimalType(18.6), 23.2, new DecimalType(23.2));
         performDataTypeMappingTest("open_prop", true, OpenClosedType.OPEN, false, OpenClosedType.CLOSED);
         performDataTypeMappingTest("colortemp_prop", 10, new PercentType(10), 60, new PercentType(60));
-        performDataTypeMappingTest("color_prop", "#f2fe00", new HSBType("63,100,100"), "#ff0000",
-                new HSBType("0.0,100.0,100.0"));
+        performDataTypeMappingTest("color_prop", "#f2fe00", new HSBType("62.834645669291325,100,99.60784313725490"),
+                "#ff0000", new HSBType("0.0,100.0,100.0"));
```
J-N-K, line 144: See above, l.188 needs changing in this test

J-N-K (changes requested):

holgerfriedrich:
@J-N-K can you pls have a look?
Changes by @andrewfg allow for keeping the original target values.

For string based comparison, I have no idea how to proceed....
https://github.com/openhab/openhab-addons/blob/efed2156e1781697fd7a4051546057f8d0d94c75/bundles/org.openhab.binding.mqtt.generic/src/test/java/org/openhab/binding/mqtt/generic/ChannelStateTests.java#L315-L316

But I would take the risk that this needs to be modified again in case we change the accuracy....

andrewfg:
> For string based comparison, I have no idea how to proceed....
> risk that this needs to be modified again in case we change the accuracy

@holgerfriedrich indeed. The issue is that the authors of that binding had created a `ColorValue` class in the binding containing color manipulation utilities, because at that time the core `ColorUtil` class did not yet exist.

In fact the `ColorValue` class does a kind of `HSBType.closeTo()` comparison, not via comparing numeric values, but by casting the numeric values to a string via a `String.Format` command that specifies the precision of the rounding. Now it is not even clear if `String.Format` uses the same rounding algorithms and `BigDecimal` does, so in any case I think this form of comparison is quite precarious.

I suppose that ultimately a possible solution may be to persuade the binding owners to either a) use parts of `ColorUtil` in `ColorValue`, or b) to migrate code from `ColorValue` into `ColorUtil` -- maybe something as simple as adding a `toString()` method. ??

So I am pinging code owners of the binding @ccutrer @davidgraeff in case you have thoughts on this?

https://github.com/openhab/openhab-addons/blob/efed2156e1781697fd7a4051546057f8d0d94c75/bundles/org.openhab.binding.mqtt.generic/src/main/java/org/openhab/binding/mqtt/generic/values/ColorValue.java

andrewfg:
^
In the mean-time the 'closeTo' fix for the Junit test is as follows..

```
   @Test
    public void receiveXYYColorTest() {
        ColorValue value = new ColorValue(ColorMode.XYY, "FON", "FOFF", 10);
        ChannelState c = spy(new ChannelState(config, channelUIDMock, value, channelStateUpdateListenerMock));
        c.start(connectionMock, mock(ScheduledExecutorService.class), 100);

        c.processMessage("state", "ON".getBytes()); // Normal on state
        assertThat(value.getChannelState().toString(), is("0,0,10"));
        assertThat(value.getMQTTpublishValue((Command) value.getChannelState(), null), is("0.322700,0.329000,10.00"));

        c.processMessage("state", "FOFF".getBytes()); // Custom off state
        assertThat(value.getChannelState().toString(), is("0,0,0"));
        assertThat(value.getMQTTpublishValue((Command) value.getChannelState(), null), is("0.000000,0.000000,0.00"));

        c.processMessage("state", "10".getBytes()); // Brightness only
        assertThat(value.getChannelState().toString(), is("0,0,10"));
        assertThat(value.getMQTTpublishValue((Command) value.getChannelState(), null), is("0.322700,0.329000,10.00"));

        HSBType t = ColorUtil.xyToHsb(new double[] { 0.3f, 0.6f });

        c.processMessage("state", "0.3,0.6,100".getBytes());
        assertTrue(((HSBType) value.getChannelState()).closeTo(t, 0.001)); // HSB

        assertCloseTo(value.getMQTTpublishValue((Command) value.getChannelState(), null), "0.300000,0.599900,100.00");
        assertCloseTo(value.getMQTTpublishValue((Command) value.getChannelState(), "%3$.1f,%2$.4f,%1$.4f"),
                "100.0,0.5999,0.3000");
    }

    private void assertCloseTo(String aString, String bString) {
        double[] a = Arrays.stream(aString.split(",")).mapToDouble(Double::parseDouble).toArray();
        double[] b = Arrays.stream(bString.split(",")).mapToDouble(Double::parseDouble).toArray();
        assertEquals(a.length, b.length);
        for (int i = 0; i < a.length; i++) {
            assertEquals(a[i], b[i], 0.02);
        }
    }
```

ccutrer:
I've updated ColorValue to use ColorUtil as an additional commit on this PR. This both reduced the code, and improved the accuracy, since there's less truncating instead of rounding going on in ColorValue. I also adjusted the tests to avoid checking the actual color value when the test is actually testing something else (the formatting of messages, for example).

bundles/org.openhab.binding.mqtt.generic/src/main/java/org/openhab/binding/mqtt/generic/values/ColorValue.java
```diff
@@ -144,14 +145,12 @@ public String getMQTTpublishValue(Command command, @Nullable String pattern) {
...
-                        rgb[2].toBigDecimal().multiply(factor).intValue());
+                int[] rgb = ColorUtil.hsbToRgb(hsbState);
+                return String.format(formatPattern, rgb[0], rgb[1], rgb[2]);
             case XYY:
-                PercentType[] xyY = hsbState.toXY();
-                return String.format(Locale.ROOT, formatPattern, xyY[0].floatValue() / 100.0f,
-                        xyY[1].floatValue() / 100.0f, hsbState.getBrightness().floatValue());
+                double[] xyY = ColorUtil.hsbToXY(hsbState);
+                return String.format(Locale.ROOT, formatPattern, xyY[0], xyY[1],
+                        hsbState.getBrightness().doubleValue());
```
J-N-K, line 153: Why not use `xyY[2]`?

ccutrer: Because Y is Y, not brightness, and the MQTT binding expects brightness. Like I said, I don't fully understand the different color spaces (especially xyY), but I know Y != B.

J-N-K (approved):
Thanks!

PR #14798, opened 2023-04-14
[jrubyscripting] Update README.md

jimtng:
The current readme has comments after the headings which got included into the main documentation site navigation link

image

Note this should be merged to 3.4.x as well.

PR #14814, opened 2023-04-15
[jrubyscripting] Update README.md

jimtng:
The equivalent of #14798 but for 3.4.x


jimtng:
@jlaur

jlaur:
@jimtng - just to be sure:
- The README is a straight copy from main?
- Nothing in the README is only relevant for 9.4.x (since 3.4 and 4.0 are different versions)?

jimtng:
> * The README is a straight copy from main?

Yes, it's exactly the same.

> * Nothing in the README is only relevant for 9.4.x (since 3.4 and 4.0 are on different versions)?

Correct. It was written to be applicable to both versions, with notes when differences exist. It was the same text as this too: https://openhab.github.io/openhab-jruby/main/index.html

jlaur (approved):
LGTM

PR #14817, opened 2023-04-15
[velux] Softer initialization after power failure

andrewfg:
After a power failure when both OH and the Velux bridge are restarting, the binding now waits quietly for a few minutes during which it 'pings' the bridge to check that it has come online again, before actually attempting to log-in.


lolodomo (approved):
LGTM, thank you

lolodomo:
No feedback, so let's merge.

PR #14821, opened 2023-04-15
[mimictts] Fix addon.xml info

wborn:
It seems to have been copied over from macOS TTS without properly updating all details.


lolodomo (approved):
LGTM

PR #14822, opened 2023-04-15
[openwebnet] Fix for dimmer blocked after two OFF are received

### LINKED ISSUE #14785: [openwebnet] switch command OFF when already OFF blocks dimmer
## Current Behavior
sending command OFF to a dimmed light that is already OFF blocks the light and it cannot switch it ON again with a ON-command, only with a command that sends a level it works (although this works from a physical switch).

- first i sent command ON - light switches ON with the known behaviour that it changes always to 100 an then immediately back to the last set dimmer level
- then i sent command OFF - light switches OFF (changes to 0)
- then again i sent command OFF although light is already OFF
- after that when i try to switch ON again it changes to 100 and immediately back to 0 AND now i cannot switch it on with a switch from sitemap in basic-ui, also not with a slider from basic-ui on my windows-computer, strange but with the same slider from basic-ui on my mobile i can switch it ON again and then it works again

## Your Environment
raspberry 4b with openhabian installed with oh 4.0.0.m1

mvalla:
This fix is needed to avoid "block" dimmers when two consecutive OFF commands are received: when a ON command is then received, the brightness remains 0 anyway.
The cause being that `brightnessBeforeOff` is wrongly set to 0 when the second OFF is received.
See: https://community.openhab.org/t/openwebnet-bticino-legrand-binding-beta-4-0-0-0-4-1-0-0/129775/72?u=massi
Fixes #14785.

### DIFF
--- a/bundles/org.openhab.binding.openwebnet/src/main/java/org/openhab/binding/openwebnet/internal/handler/OpenWebNetLightingHandler.java
+++ b/bundles/org.openhab.binding.openwebnet/src/main/java/org/openhab/binding/openwebnet/internal/handler/OpenWebNetLightingHandler.java
@@ -194,13 +194,16 @@ public class OpenWebNetLightingHandler extends OpenWebNetThingHandler {
     private void dimLightTo(int percent, Command command) {
         logger.debug("   DIM dimLightTo({}) bri={} briBeforeOff={}", percent, brightness, brightnessBeforeOff);
         int newBrightness = percent;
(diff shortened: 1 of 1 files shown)


lolodomo (approved):
LGTM

mvalla:
Thanks @lolodomo . This one should also be backported to OH3 in my opinion

lolodomo:
Pinging @jlaur for backporting

PR #14825, opened 2023-04-16
[map][scale] Fix localization

### LINKED ISSUE #14823: [map] Transformation language not working
## Expected Behavior

> `language` is used to select transformations based on locales. You could have a transformation "switch" which depending on the locale translated ON to different values. This is also available in file-based transformations by adding a `_de` suffix to the filename (before the extension).

_Originally posted by @J-N-K in https://github.com/openhab/openhab-webui/pull/1448#issuecomment-1484022478_

> So if you have foo_en.script and a foo_de.script (or the corresponding managed UIDs), base on the locale either the first or the second is used when you configure SCRIPT(foo.script):%s.

_Originally posted by @J-N-K in https://github.com/openhab/openhab-webui/pull/1448#issuecomment-1484023389_

## Current Behavior
I have these transformations:

- files
  - `switch_de.map`
  - `switch_en.map`
- UI
  - `config:map:switch:de`
  - `config:map:switch:en`

Using `MAP(switch.map)` or `MAP(config:map:switch)` for Item states does not work and leads to:

```
Failed transforming the state 'OFF' on item 'test_switch' with pattern 'MAP(switch.map):%s': Could not find configuration 'switch.map' or failed to parse it.
```

See above.

## Context
I have discovered this while working on the UI transformation menu & editor.
If this is still supported, the UI needs a minor adjustment for the tip how to use that transformation.

## Your Environment
* Version used: `SNAPSHOT openhab/openhab-core#3415`
* Environment name and version: Chrome 112
* Operating System and version: Fedora 36

J-N-K:
Fixes #14823 

The wrong key was used in the `cachedTransformations` map. Instead of the correct localized UID from the registry the provided transformation function was used.

### DIFF
--- a/bundles/org.openhab.transform.map/src/main/java/org/openhab/transform/map/internal/MapTransformationService.java
+++ b/bundles/org.openhab.transform.map/src/main/java/org/openhab/transform/map/internal/MapTransformationService.java
@@ -81,7 +81,7 @@ public class MapTransformationService
             if (!cachedTransformations.containsKey(transformation.getUID())) {
                 importConfiguration(transformation);
             }
-            Properties properties = cachedTransformations.get(function);
+            Properties properties = cachedTransformations.get(transformation.getUID());
             if (properties != null) {
                 String target = properties.getProperty(source);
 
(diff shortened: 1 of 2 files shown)


clinique (approved):

jlaur (approved):
Thanks!

PR #14827, opened 2023-04-16
[deconz] Update the roller shutter status

### LINKED ISSUE #14826: [deconz] roller shutter state is not updating
## Expected Behavior

When opening or closing a roller shutter, the state should change  (0 - 100%). 

## Current Behavior

The REST interface in the newest deCONZ (2.21.02) is not transmitting the brightness value (0-255).
Instate a lift value is transmitted, but it is currently not used. Therefore the state is ether 0% or 100%.

## Possible Solution

Use the lift value from the REST interface.

1. include a roller shutter with the deconz binding
2. add it to the sitemap 
3. open or close the roller shutter
4. state is not changing in the sitemap

## Context

I want to see, if the roller shutters are open or closed. 

## Your Environment

* Version used: openHAB 4.0, deconz 2.21.02
* Operating System and version (desktop or mobile, Windows 10, Raspbian Buster, ...):Raspbian
* Hardware: Conbee II

dza1:
Fixes #14826

if the REST interface from deconz provides a 'lift' value to show the current position of the roller shutter, it will be used. Otherwise, the 'bri' value will be used.

### DIFF
--- a/bundles/org.openhab.binding.deconz/src/main/java/org/openhab/binding/deconz/internal/handler/LightThingHandler.java
+++ b/bundles/org.openhab.binding.deconz/src/main/java/org/openhab/binding/deconz/internal/handler/LightThingHandler.java
@@ -397,10 +397,8 @@ public class LightThingHandler extends DeconzBaseThingHandler {
             }
             case CHANNEL_POSITION -> {
                 Integer lift = newState.lift;
-                if (lift != null && lift >= PercentType.ZERO.intValue() && lift <= PercentType.HUNDRED.intValue()) {
+                if (lift != null) {
                     updateState(channelUID, new PercentType(lift));
-                } else {
-                    updatePercentTypeChannel(channelUID, newState.bri, true); // always post value
                 }
             }
             case CHANNEL_EFFECT -> updateStringChannel(channelUID, newState.effect);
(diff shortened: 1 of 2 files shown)


bundles/org.openhab.binding.deconz/src/main/java/org/openhab/binding/deconz/internal/handler/LightThingHandler.java
```diff
@@ -395,7 +395,14 @@ private void valueUpdated(ChannelUID channelUID, LightState newState) {
...
                 }
             }
-            case CHANNEL_POSITION -> updatePercentTypeChannel(channelUID, newState.bri, true); // always post value
+            case CHANNEL_POSITION -> {
+                Integer lift = newState.lift;
+                if (lift != null && lift >= PercentType.ZERO.intValue() && lift <= PercentType.HUNDRED.intValue()) {
+                    updateState(channelUID, new PercentType(lift));
+                } else {
+                    updatePercentTypeChannel(channelUID, newState.bri, true); // always post value
+                }
```
J-N-K, line 404: I wonder if this is necessary. Is there really any deconz firmware that accepts `lift` as command but does not report the status in `lift`? Since we only set `lift` for the command, I believe it's enough to consider `lift`. I think you can also omit the range check.

dza1: I tried a one year old deconz firmware and `lift` was already included.
Therefore I modified my change according your suggestion.

J-N-K (approved):
LGTM

PR #14829, opened 2023-04-16
[boschindego] Add device properties

### LINKED ISSUE #14828: [boschindego] Obtain device information and set as properties
While working on #14745 I noticed that this request:
`GET https://api.indego-cloud.iot.bosch-si.com/api/v1/alms/xxx/`

returns some useful information like:
```json
{
    "alm_sn": "xxx",
    "service_counter": 47931,
    "needs_service": false,
    "alm_mode": "manual",
    "bareToolnumber": "3600HB0303",
    "alm_firmware_version": "30036.01251",
    "renew_date": "2027-05-29T18:29:41.928Z"
}
```

This would be nice to add to the thing as properties, so the user would be able to see this.

I remember when I had to fill out a form before sending my device to a service center, that I struggled to find the correct bare tool number from the model.

jlaur:
Adds some useful properties like Bosch bare tool number, model name and firmware version.

Use the model name as thing label from discovery.

Resolves #14828


lolodomo (approved):
LGTM, thank you

PR #14842, opened 2023-04-18
Upgrade maven-surefire-plugin to 3.0.0

wborn:
This fixes deprecation warnings when runnings tests with Maven 3.9.x:

`[WARNING] Parameter 'localRepository' is deprecated core expression; Avoid use of ArtifactRepository type. If you need access to local repository, switch to '${repositorySystemSession}' expression and get LRM from it instead.`

See: https://issues.apache.org/jira/browse/SUREFIRE-2154

Related to openhab/openhab-core#3512


jlaur (approved):
Thanks for the due diligence and upgrade!

PR #14846, opened 2023-04-19
[renault] update default Kamereon API Key

dougculnane:
# Update default Kamereon API Key in Renault binding

The API key used in the Renault binding to access the MyRenault services changes sometimes...  This can be updated in the bindings configuration to repair communication of existing installations.

This PR updates the default key to the new one for new installations.


jlaur (approved):
Thanks!

PR #14848, opened 2023-04-19
[smartmeter] Remove copy-paste-error, as this binding doesn't play sounds

nWRqenkF:
Seems to be a copy-paste-error for parameter port, because this bindung doesn't play any sounds.


lolodomo (approved):
LGTM

PR #14858, opened 2023-04-20
[mqtt.homeassistant] don't set range on current temp channel

ccutrer:
the min/max from the config are specifically only for the set point. i.e. for a heater in winter when the heater is off, it may have a set point range of 65-104°F, but if it's off for an unoccopied room the ambient temperature might drop far below 65


lolodomo (approved):
LGTM

lolodomo:
Let's wait for @antroids approval.

antroids:
Good catch! These values are for command topics only.
LGTM

PR #14865, opened 2023-04-21
[openwebnet] Make initialization value of Auxiliary to NONE

### LINKED ISSUE #14744: [openwebnet] Make initialization value of Auxiliary to NONE
Currently an Auxiliary item's initialization value is set to Null which seems an erratic status  to the user. So it  would be better to initialize the item  to NONE as AlarmHandler class  does. 

Minor fix: correct the Auxiliary class' Javadoc to cut the sentence "NOTICE: Support for handling messages from the bus regarding alarm control has to be implemented" as  the feature  has been implemented

fabgio:
**Description:**

Currently an Auxiliary item's initialization value is set to Null which seems an erratic status to the user. So it  initializes the item to NONE as AlarmHandler class does.

Minor fix: correct the Auxiliary class' Javadoc to cut the sentence "NOTICE: Support for handling messages from the bus regarding alarm control has to be implemented" as the feature has been implemented. This PR closes #14744

**Test:**
Tested on BTicino/Legrand bulgrar-alarm unit 3486

### DIFF
--- a/bundles/org.openhab.binding.openwebnet/src/main/java/org/openhab/binding/openwebnet/internal/handler/OpenWebNetAuxiliaryHandler.java
+++ b/bundles/org.openhab.binding.openwebnet/src/main/java/org/openhab/binding/openwebnet/internal/handler/OpenWebNetAuxiliaryHandler.java
@@ -14,6 +14,7 @@ package org.openhab.binding.openwebnet.internal.handler;
 
 import static org.openhab.binding.openwebnet.internal.OpenWebNetBindingConstants.CHANNEL_AUX;
 
+import java.util.HashSet;
 import java.util.Set;
 
 import org.eclipse.jdt.annotation.NonNullByDefault;
@@ -41,8 +42,6 @@ import org.slf4j.LoggerFactory;
  * The {@link OpenWebNetAuxiliaryHandler} is responsible for sending Auxiliary (AUX) commands/messages to the bus
  * It extends the abstract {@link OpenWebNetThingHandler}.
  *
- * NOTICE: Support for handling messages from the bus regarding alarm control has to be implemented
- *
  * @author Giovanni Fabiani - Initial contribution
  *
  */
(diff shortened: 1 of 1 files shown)


fabgio:
From Javadoc 17  on HashSet class "...note that this implementation is not synchronized. If multiple threads access a hash set concurrently, and at least one of the threads modifies the set, it must be synchronized externally". it looks it is not thread-safe  in case of concurrent access by multiple threads. Should we consider thread-safe alternatives: such as: "" Set s = Collections.synchronizedSet(new HashSet(...));"?. i need your opinion...

bundles/org.openhab.binding.openwebnet/src/main/java/org/openhab/binding/openwebnet/internal/handler/OpenWebNetAuxiliaryHandler.java
```diff
@@ -52,10 +51,22 @@ public class OpenWebNetAuxiliaryHandler extends OpenWebNetThingHandler {
     public static final Set<ThingTypeUID> SUPPORTED_THING_TYPES = OpenWebNetBindingConstants.AUX_SUPPORTED_THING_TYPES;
     private final Logger logger = LoggerFactory.getLogger(OpenWebNetAuxiliaryHandler.class);
 
+    private static final Set<OpenWebNetAuxiliaryHandler> auxHandlers = new HashSet<>();
```
lolodomo, line 54: Except addind to and removing from this set, it looks like it is not used ?
And why static ?

lolodomo (changes requested):

fabgio:
Yes exactly, Except adding to and removing from this set, it looks like it is not used ?  Yes  exactly it is  used solely for adding and removing from the set. I believe it is a no use for an instance variable to be declared as static... so I  would  remove the static keyword

lolodomo:
The question is why creating and managing a new set that you never use ? What is its purpose ?

mvalla:
I gave a look to this PR. These are my comments:
- the proposed implementation is not complete, as a set of other AUX Things is created, but never used anywhere. The proposer say it has been tested but I do not understand how it was tested since the code does not do anything useful other than adding/removing Things from the set
- more importantly I think this PR is invalid as possible WHAT values defined for AUX, as documented in the Specs (Who=5, page 7) are: ON, OFF, TOGGLE, STOP, UP, DOWN, ENABLED, DISABLED, RESET_GEN, RESET_BI, RESET_TRI and there is no NONE value defined. So at initialization is perfectly fine that the channle value is Null as it happens for every other channel in OH which has not been updated yet
- finally regarding the fix to Javadoc: since support for handling AUX messages **from** the BUS is in fact not supported yet, I think the comment should stay there.

So for me this PR (and related Issue) should be closed as invalid.

As a finel comment I suggest discussing these imporvements and their possible implemnetations in the forum first.

lolodomo:
> So for me this PR (and related Issue) should be closed as invalid.
> 
> As a final comment I suggest discussing these improvements and their possible implementations in the forum first.

I close the PR.
+1 for the suggestion.

PR #14868, opened 2023-04-21
[robonect] Improve job type configuration

### LINKED ISSUE #6142: [robonect] Remove time limit for the duration parameter of a mower job
I would like to trigger a smaller change for the Robonect binding.

Currently, there is a channel to trigger a mower “job” and setup a duration.
Unfortunately, this seems to be limited to 60 minutes. In the channel configuration on PaperUI I will get automatically an error, when I try to input a duration higher than 60.
When I try to parse a higher value to the channel using a file, it does not work.
But robonect is able to accept a duration higher than 60 minutes.
Is it possible to change this limitation?

Michael

joni1993:
* Add allowed pattern to description
* Increase allowed duration to 12h - thats what the web gui allows

fixes #6142

### DIFF
--- a/bundles/org.openhab.binding.robonect/src/main/resources/OH-INF/config/config.xml
+++ b/bundles/org.openhab.binding.robonect/src/main/resources/OH-INF/config/config.xml
@@ -6,19 +6,32 @@
 	https://openhab.org/schemas/config-description-1.0.0.xsd">
 
 	<config-description uri="channel-type:robonect:jobType">
-		<parameter name="remoteStart" type="text" pattern="^STANDARD|REMOTE_1|REMOTE_2$">
+		<parameter name="remoteStart" type="text">
 			<label>Remote Start</label>
 			<description>The location to start the mowing job from.</description>
+			<options>
+				<option value="STANDARD">Standard</option>
+				<option value="REMOTE_1">Remote 1</option>
+				<option value="REMOTE_2">Remote 2</option>
+			</options>
+
 		</parameter>
-		<parameter name="duration" type="integer" min="0" max="60" unit="m">
+		<parameter name="duration" type="integer" min="0" max="720" unit="m">
 			<context>time</context>
 			<label>Job Duration</label>
 			<description>The duration of the job.</description>
 			<default>0</default>
+
 		</parameter>
-		<parameter name="afterMode" type="text" pattern="^AUTO|HOME|EOD$">
+		<parameter name="afterMode" type="text">
 			<label>After Job Mode</label>
 			<description>The Mode to put the mower into after the job is done.</description>
+			<options>
+				<option value="AUTO">Auto</option>
+				<option value="HOME">Home</option>
+				<option value="EOD">End of day</option>
+			</options>
+
 		</parameter>
 	</config-description>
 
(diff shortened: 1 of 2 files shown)


bundles/org.openhab.binding.robonect/src/main/resources/OH-INF/config/config.xml
```diff
@@ -8,17 +8,17 @@
 	<config-description uri="channel-type:robonect:jobType">
 		<parameter name="remoteStart" type="text" pattern="^STANDARD|REMOTE_1|REMOTE_2$">
 			<label>Remote Start</label>
-			<description>The location to start the mowing job from.</description>
+			<description>The location to start the mowing job from. [STANDARD|REMOTE_1|REMOTE_2]</description>
```
jlaur, line 11: Perhaps something like this would be even better?
```suggestion
			<description>The location to start the mowing job from.</description>
			<options>
				<option value="STANDARD">Standard</option>
				<option value="REMOTE_1">Remote 1</option>
				<option value="REMOTE_2">Remote 2</option>
			</options>
```

bundles/org.openhab.binding.robonect/src/main/resources/OH-INF/config/config.xml
```diff
@@ -8,17 +8,17 @@
...
+		<parameter name="duration" type="integer" min="0" max="720" unit="m">
 			<context>time</context>
 			<label>Job Duration</label>
 			<description>The duration of the job.</description>
 			<default>0</default>
 		</parameter>
 		<parameter name="afterMode" type="text" pattern="^AUTO|HOME|EOD$">
 			<label>After Job Mode</label>
-			<description>The Mode to put the mower into after the job is done.</description>
+			<description>The Mode to put the mower into after the job is done. [AUTO|HOME|EOD]</description>
```
jlaur, line 21: And also here.

jlaur (changes requested):
Thanks for the enhancement! I have provided a suggestion that might make it even more user-friendly.

Please also fix the DCO.

After finishing changes in config.xml, please regenerate I18N properties - see https://www.openhab.org/docs/developer/utils/i18n.html#generating-i18n-properties-file

bundles/org.openhab.binding.robonect/src/main/resources/OH-INF/config/config.xml
joni1993: Thank you for your inputs - just got in touch with the code after using OH for some years now ;)

jlaur: Welcome as contributor. 🙂 Only one question left for the code: Is it safe to limit to those options or should it be possible to freely input other options not defined here?

jlaur: I now see the pattern already prevented other options, so the pattern can now be removed as restricted by options.

joni1993: I will do that in a minute.

Do you maybe know if it's intended behavour that the parameter.unit is not reflected in the UI?
```unit="m"``` indicates its minutes, but it's nowhere visible in the MainUI. Is it recommended to have a ```unitLabel``` despite having a unit defined?
The same applies for e.g. https://github.com/openhab/openhab-addons/blob/f08de2642973692b4f22cad007ff4eb27be076f1/bundles/org.openhab.binding.robonect/src/main/resources/OH-INF/thing/thing-types.xml#L72  - i needed to check the source to verify if its seconds or minutes because its not reflected in the UI.

jlaur: I'm not sure, but it might simply never have been implemented in MainUI. See also https://github.com/openhab/openhab-addons/pull/12461#discussion_r841053479.

jlaur (approved):
Thanks - LGTM!

PR #14872, opened 2023-04-22
[ipcamera] Fix servlet exceptions due to non unique names

apella12:
[IP camera] on OH4 has the same servlet issue that was fixed by PR #14554.  This PR uses that PR as a template to fix the issue in the IPcamera binding.  (That issue being that only one camera was able to stream.) This was tested locally on OH4M2 and it works.  Also the referenced PR worked with [neeo] and this used that PR as a template.

Signed-off-by: Bob Eckhoff  <katmandodo@yahoo.com>

### DIFF
--- a/bundles/org.openhab.binding.ipcamera/src/main/java/org/openhab/binding/ipcamera/internal/servlet/IpCameraServlet.java
+++ b/bundles/org.openhab.binding.ipcamera/src/main/java/org/openhab/binding/ipcamera/internal/servlet/IpCameraServlet.java
@@ -18,6 +18,7 @@ import java.io.File;
 import java.io.FileInputStream;
 import java.io.IOException;
 import java.util.Dictionary;
+import java.util.Hashtable;
 import java.util.Properties;
 
 import javax.servlet.ServletOutputStream;
@@ -59,6 +60,8 @@ public abstract class IpCameraServlet extends HttpServlet {
 
     public void startListening() {
         try {
+            Hashtable<Object, Object> initParameters = new Hashtable<>();
+            initParameters.put("servlet-name", "/ipcamera/" + handler.getThing().getUID().getId());
             httpService.registerServlet("/ipcamera/" + handler.getThing().getUID().getId(), this, initParameters,
                     httpService.createDefaultHttpContext());
         } catch (Exception e) {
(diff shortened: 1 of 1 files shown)


bundles/org.openhab.binding.ipcamera/src/main/java/org/openhab/binding/ipcamera/internal/servlet/IpCameraServlet.java
```diff
@@ -59,6 +60,8 @@ public IpCameraServlet(ThingHandler handler, HttpService httpService, Dictionary
 
     public void startListening() {
         try {
+            Hashtable<Object, Object> initParameters = new Hashtable<>();
```
jlaur, line 63: You are completely overriding the `initParameters` provided by the second constructor in lines 54-59. This constructor is used from this subclass:

https://github.com/openhab/openhab-addons/blob/275329d4858acf5067c74bf6a1a5d88b91f88f93/bundles/org.openhab.binding.ipcamera/src/main/java/org/openhab/binding/ipcamera/internal/servlet/CameraServlet.java#L47-L56

so it will break `async-supported` being set to `true`.

You could instead clone `this.initParameters` and add the additional entry.

jlaur (changes requested):
Thanks for the fix! It seems there might be an issue, please see my comment.

bundles/org.openhab.binding.ipcamera/src/main/java/org/openhab/binding/ipcamera/internal/servlet/IpCameraServlet.java
Skinah: Perhaps my java terminology is poor, but you don't need to clone it, you can just delete the new import and the following line to directly use initParameters.

remove this line...
```
Hashtable<Object, Object> initParameters = new Hashtable<>();
```

You can use the initParameters.put directly leaving it as a single line change. I have tested this tonight and it fixes the issue here. Thanks for the PR.

apella12: Added commit to address comment

jlaur: @Skinah - you are right, thanks for the proposal which I see is adopted by @apella12.

In general I'm hesitant to manipulate collections provided by external callers in case they would not expect this when not clearly documented, for example by a method signature like `void addAdditionalParameters(Dictionary<Object, Object> parameters)`.

jlaur (approved):
LGTM

PR #14875, opened 2023-04-25
[knx] Fix decoding of DPT 242.600 and add tests

holgerfriedrich:
Correct handling of Y parameter for DPT 242.600.

Add back to back tests for DPT 232.600 and 242.600, testing color conversion from raw bytes to HSB and back to Calimero color representation as String.

### DIFF
--- a/bundles/org.openhab.binding.knx/src/main/java/org/openhab/binding/knx/internal/client/SerialTransportAdapter.java
+++ b/bundles/org.openhab.binding.knx/src/main/java/org/openhab/binding/knx/internal/client/SerialTransportAdapter.java
@@ -39,12 +39,12 @@ import tuwien.auto.calimero.serial.spi.SerialCom;
  * The {@link SerialTransportAdapter} provides org.openhab.core.io.transport.serial
  * services to the Calimero library.
  * 
- * @ServiceProvider annotation (biz.aQute.bnd.annotation) automatically creates the file
- *                  /META-INF/services/tuwien.auto.calimero.serial.spi.SerialCom
- *                  to register SerialTransportAdapter to the service loader.
- *                  Additional attributes for SerialTansportAdapter can be specified as well, e.g.
- *                  attribute = { "position=1" }
- *                  and will be part of MANIFEST.MF
+ * {@literal @}ServiceProvider annotation (biz.aQute.bnd.annotation) automatically creates the file
+ * /META-INF/services/tuwien.auto.calimero.serial.spi.SerialCom
+ * to register SerialTransportAdapter to the service loader.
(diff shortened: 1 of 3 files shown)


bundles/org.openhab.binding.knx/src/test/java/org/openhab/binding/knx/internal/dpt/DPTTest.java
```diff
@@ -402,13 +404,81 @@ public void unitsValid(String unit) {
...
+
+        // encoding will return a String in notation defined by Calimero: "(x,xxxx y,yyyy) YY,Y %"
+        String result = ValueEncoder.encode(hsb, dpt);
+
+        // for back to back test, compare numerical values to allow tolerances
+        double dx = (((value[0] & 0xff) << 8) | (value[1] & 0xff)) / 65535.0;
+        double dy = (((value[2] & 0xff) << 8) | (value[3] & 0xff)) / 65535.0;
+        double dY = ((double) (value[4] & 0xff)) * 100.0 / 255.0;
+
+        final Pattern XYY_PATTERN = Pattern.compile(
```
J-N-K, line 463: Is this the exact same pattern as above in `ValueDecoder`? If so, just remove the `private` from there and re-use it. This ensures changes are not missed in the future.

J-N-K (commented):
Otherwise LGTM.

bundles/org.openhab.binding.knx/src/test/java/org/openhab/binding/knx/internal/dpt/DPTTest.java
holgerfriedrich: @J-N-K done as requested

bundles/org.openhab.binding.knx/src/main/java/org/openhab/binding/knx/internal/client/SerialTransportAdapter.java
```diff
@@ -39,12 +39,12 @@
  * The {@link SerialTransportAdapter} provides org.openhab.core.io.transport.serial
  * services to the Calimero library.
  * 
- * @ServiceProvider annotation (biz.aQute.bnd.annotation) automatically creates the file
```
J-N-K, line 42: Is this change intended?

holgerfriedrich: Yes, it fixes the javadoc build. @ character in a javadoc was interpreted as javadoc tag.
This is only the documentation, the ServiceLoader tag is just below.

J-N-K (approved):
Thanks

PR #14884, opened 2023-04-26
[transform.exec] Fix service whitelisting

J-N-K:
Reported on the german openHAB forum.


jlaur:
Déjà vu comment for others: Same fix as in #14522. It seems we have duplicated code in exec **binding** and exec **transform**. So also a regression of #14004.

jlaur (approved):
Thanks for the fix!

J-N-K:
The code-duplication is unfortunately necessary because exec transformation and binding are independent.

jlaur:
> The code-duplication is unfortunately necessary because exec transformation and binding are independent.

I understand, I just wanted to share this information as I was not previously aware. I guess the only current way to share that code would be to bundle them, but that would probably be quite counter-intuitive/undesirable. I didn't check how much code is duplicated.

PR #14888, opened 2023-04-27
[influxdb] Write asynchronously to database and improve performance

### LINKED ISSUE #14887: [influxdb] Persistence service blocks item notifications
According to the documentation of the `PersistenceService` interface storing of values should return fast. This is not the case for InfluxDB which directly writes to the database. Since this is time-consuming the thread pool that is responsible for notifying listeners about event state changes might exhaust. Found while investigating https://github.com/openhab/openhab-core/issues/3578.

J-N-K:
Fixes #14887 

@florian-h05 FYI.


florian-h05 (commented):
Works fine with InfluxDB 2 and solves the problem described in https://github.com/openhab/openhab-core/issues/3578 when using InfluxDB 👍 
Thanks for the fix!

Code LGTM, just two minor comments.

florian-h05 (approved):
Still works fine 👍
Code LGTM.

florian-h05:
@J-N-K Can we remove the additional testing preferred tag since it works fine for me?

J-N-K:
It also works fine for me.

bundles/org.openhab.persistence.influxdb/src/main/java/org/openhab/persistence/influxdb/InfluxDBPersistenceService.java
```diff
@@ -157,26 +189,26 @@ public void store(Item item) {
...
-            InfluxPoint point = itemToStorePointCreator.convert(item, alias);
-            if (point != null) {
-                try {
-                    influxDBRepository.write(point);
-                    logger.trace("Stored item {} in InfluxDB point {}", item, point);
-                } catch (UnexpectedConditionException e) {
-                    logger.warn("Failed to store item {} in InfluxDB point {}", point, item);
-                }
+        if (!serviceActivated) {
+            logger.warn("InfluxDB service not ready. Storing {} rejected.", item);
```
lolodomo, line 193: Rather item.getName()

bundles/org.openhab.persistence.influxdb/src/main/java/org/openhab/persistence/influxdb/InfluxDBPersistenceService.java
```diff
@@ -157,26 +189,26 @@ public void store(Item item) {
...
+        convert(item, alias).thenAccept(point -> {
+            if (point == null) {
+                logger.trace("Ignoring item {}, conversion to an InfluxDB point failed.", item.getName());
+                return;
+            }
+            if (pointsQueue.offer(point)) {
+                logger.trace("Queued {} for item {}", point, item);
             } else {
-                logger.trace("Ignoring item {}, conversion to an InfluxDB point failed.", item);
+                logger.warn("Failed to queue {} for item {}", point, item);
```
lolodomo, line 204: Rather item.getName()

bundles/org.openhab.persistence.influxdb/src/main/java/org/openhab/persistence/influxdb/InfluxDBPersistenceService.java
```diff
@@ -101,8 +124,9 @@ public InfluxDBPersistenceService(final @Reference ItemRegistry itemRegistry,
         if (configuration.isValid()) {
             this.influxDBRepository = createInfluxDBRepository();
             this.influxDBRepository.connect();
-            this.itemToStorePointCreator = new ItemToStorePointCreator(configuration, influxDBMetadataService);
-            tryReconnection = true;
+            this.storeJob = ThreadPoolManager.getScheduledPool("org.openhab.influxdb")
+                    .scheduleWithFixedDelay(this::doStore, 1, 1, TimeUnit.SECONDS);
```
lolodomo, line 128: I am asking myself if 1 second is the good frequency ?
We have to find the good compromise between the size of the queue and how much often the data is stored in DB.

lolodomo (changes requested):
Just minor comments.
Tests not yet reviewed.

bundles/org.openhab.persistence.influxdb/src/main/java/org/openhab/persistence/influxdb/InfluxDBPersistenceService.java
J-N-K: I have thought about that, too. But logging `item` also logs the state, not only the item name. That's why I kept `item`.

bundles/org.openhab.persistence.influxdb/src/main/java/org/openhab/persistence/influxdb/InfluxDBPersistenceService.java
J-N-K: What do you suggest? The longer the time, the greater the risk of losing data. IMO 10s is far too long, but as you suggest, maybe 1s is too small.

lolodomo: Is there really a risk to loose data ?? Why ?

I agree that 10s is too high.

bundles/org.openhab.persistence.influxdb/src/main/java/org/openhab/persistence/influxdb/InfluxDBPersistenceService.java
lolodomo: Ok

bundles/org.openhab.persistence.influxdb/src/main/java/org/openhab/persistence/influxdb/InfluxDBPersistenceService.java
J-N-K: Power loss e.g.. I agree that it's not very likely, but the risk is there. What do you think, shall we try 3s?

lolodomo: Yes, I was thinking about 3s.

lolodomo (approved):
LGTM, thank you

Issue #14885, opened 2023-04-26
Declare connection for addons where missing


jlaur:
In order for the new connection declaration to be really useful, for example for filtering when installing addons or on the documentation website, we should declare the remaining addons.

So far many contributors already declared it for addons known to them, but still a lot are missing. This issue is created for the purpose of tracking this progress.

The documentation is here, although not yet regenerated after merging openhab/openhab-docs#2069: https://next.openhab.org/docs/developer/addons/addon.html#xml-structure-for-add-on-definitions

## Current status
_Updated 30.04.2023_

```shell
[openhab-addons]$ grep -R "<connection>" --include addon.xml | wc -l
368
[openhab-addons]$ grep -RL "<connection>" --include addon.xml | wc -l
27
```

### The list

```shell
grep -RL "<connection>" --include addon.xml | sed 's/^bundles\/org\.openhab\.\(.*\)\/src\(.*\)/- \1/' | sort
```

- io.homekit
- io.hueemulation
- io.imperihome
- io.metrics
- io.neeo
- io.openhabcloud
- persistence.dynamodb
- persistence.influxdb
- persistence.jdbc
- persistence.jpa
- persistence.mapdb
- persistence.mongodb
- persistence.rrd4j
- transform.bin2json
- transform.exec
- transform.jinja
- transform.jsonpath
- transform.map
- transform.regex
- transform.rollershutterposition
- transform.scale
- transform.xpath
- transform.xslt
- voice.actiontemplatehli
- voice.porcupineks
- voice.rustpotterks
- voice.voskstt

jlaur:
shelly is back on the list:
https://github.com/openhab/openhab-addons/pull/15031#discussion_r1253655353

markus7017:
fixed the Shelly binding

jlaur:
Fixed again by #15205.

PR #14890, opened 2023-04-27
Fix automation name/description and declare connection

jlaur:
Fix regression of #13289
Related to #14885


florian-h05 (approved):
LGTM for the JS Scripting add-ons.

PR #14897, opened 2023-04-28
[sleepiq] Restore type adapter for LocalDate

mhilbush:
This type adapter was mistakenly deleted when I migrated the 3rd party code to openHAB in PR #14180

Also fix name of json object in `SleepDataSession` dto.


jlaur (approved):
Thanks!

PR #14690, opened 2023-03-27
Add `connection` element to `addon.xml`

andrewfg:
This PR adds a `connection` element to the `addon.xml` file of some bindings where I have been involved.

Signed-off-by: Andrew Fiddian-Green <software@whitebear.ch>

PR #14902, opened 2023-04-29
Declare connection for bindings G thru M

mlobstein:
References #14885

Declared connection type in addon.xml based on review of documentation/README.

### DIFF
--- a/bundles/org.openhab.binding.hccrubbishcollection/src/main/resources/OH-INF/addon/addon.xml
+++ b/bundles/org.openhab.binding.hccrubbishcollection/src/main/resources/OH-INF/addon/addon.xml
@@ -7,5 +7,6 @@
 	<name>HCC Rubbish Collection Binding</name>
 	<description>Get the rubbish collection dates for Hamilton City Council (New Zealand).</description>
 	<connection>cloud</connection>
+	<countries>nz</countries>
 
 </addon:addon>
--- a/bundles/org.openhab.binding.hccrubbishcollection/src/main/resources/OH-INF/addon/addon.xml
+++ b/bundles/org.openhab.binding.hccrubbishcollection/src/main/resources/OH-INF/addon/addon.xml
@@ -6,5 +6,6 @@
 	<type>binding</type>
 	<name>HCC Rubbish Collection Binding</name>
 	<description>Get the rubbish collection dates for Hamilton City Council (New Zealand).</description>
+	<connection>cloud</connection>
 
 </addon:addon>
--- a/bundles/org.openhab.binding.icalendar/src/main/resources/OH-INF/addon/addon.xml
+++ b/bundles/org.openhab.binding.icalendar/src/main/resources/OH-INF/addon/addon.xml
@@ -6,5 +6,6 @@
 	<type>binding</type>
 	<name>iCalendar Binding</name>
 	<description>Binding for using iCal calendars</description>
+	<connection>cloud</connection>
 
 </addon:addon>
--- a/bundles/org.openhab.binding.icalendar/src/main/resources/OH-INF/addon/addon.xml
+++ b/bundles/org.openhab.binding.icalendar/src/main/resources/OH-INF/addon/addon.xml
@@ -6,6 +6,6 @@
 	<type>binding</type>
 	<name>iCalendar Binding</name>
 	<description>Binding for using iCal calendars</description>
-	<connection>cloud</connection>
+	<connection>local</connection>
 
 </addon:addon>
--- a/bundles/org.openhab.binding.mail/src/main/resources/OH-INF/addon/addon.xml
+++ b/bundles/org.openhab.binding.mail/src/main/resources/OH-INF/addon/addon.xml
@@ -6,6 +6,6 @@
 	<type>binding</type>
 	<name>Mail Binding</name>
 	<description>This binding is used to access POP3, IMAP and SMTP servers.</description>
-	<connection>cloud</connection>
+	<connection>local</connection>
 
 </addon:addon>
--- a/bundles/org.openhab.binding.mail/src/main/resources/OH-INF/addon/addon.xml
+++ b/bundles/org.openhab.binding.mail/src/main/resources/OH-INF/addon/addon.xml
@@ -6,5 +6,6 @@
 	<type>binding</type>
 	<name>Mail Binding</name>
 	<description>This binding is used to access POP3, IMAP and SMTP servers.</description>
+	<connection>cloud</connection>
 
 </addon:addon>
--- a/bundles/org.openhab.binding.goecharger/src/main/resources/OH-INF/addon/addon.xml
+++ b/bundles/org.openhab.binding.goecharger/src/main/resources/OH-INF/addon/addon.xml
@@ -6,5 +6,6 @@
 	<type>binding</type>
 	<name>Go-eCharger Binding</name>
 	<description>This is the binding for Go-eCharger.</description>
+	<connection>local</connection>
 
 </addon:addon>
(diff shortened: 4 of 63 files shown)


ssalonen:
Lgtm for modbus. Validated mqtt from point of view of mqtt.ruuvigateway

ssalonen (approved):

MikeTheTux:
Ok for Keba and Guntamatic

bundles/org.openhab.binding.mail/src/main/resources/OH-INF/addon/addon.xml
```diff
@@ -6,5 +6,6 @@
 	<type>binding</type>
 	<name>Mail Binding</name>
 	<description>This binding is used to access POP3, IMAP and SMTP servers.</description>
+	<connection>cloud</connection>
```
jlaur, line 9: I would argue about this - see also https://github.com/openhab/openhab-addons/pull/14813#discussion_r1167470570

The clear "cloud" case is when using an API provided by some specific vendor, i.e. "someone else's computer". Unfortunately that leaves some ambiguity for the cases where a standard or open protocol is implemented. For remoteopenhab we decided to use the criteria "the most common use-case" and defined it as `local`.

I run my own IMAP and SMTP servers, so for me it would be `local`. It's certainly possible to use it for remote servers, but so is anything running locally, really.

So difficult to say what is the most common use-case here, but the protocols are defined in RFC's and anyone can run such servers locally - and there is no vendor lock-in in play here like for most cloud bindings. So I'm leaning towards `local`.

jlaur (changes requested):
Thanks again! One minor suggestion to slightly increase scope, and one comment up for discussion. 🙂

bundles/org.openhab.binding.mail/src/main/resources/OH-INF/addon/addon.xml
daMihe: With that in mind `icalendar` (also in this PR) is also `local`. Is there a connection `network`?

bundles/org.openhab.binding.icalendar/src/main/resources/OH-INF/addon/addon.xml
```diff
@@ -6,5 +6,6 @@
 	<type>binding</type>
 	<name>iCalendar Binding</name>
 	<description>Binding for using iCal calendars</description>
+	<connection>cloud</connection>
```
daMihe, line 9: See [discussion about mail](https://github.com/openhab/openhab-addons/pull/14902/files/57c3cd4846f187f93a18e81532e9a7db22a29aa7#r1181026937)

bundles/org.openhab.binding.mail/src/main/resources/OH-INF/addon/addon.xml
daMihe: The descriptions [XSD](https://www.openhab.org/schemas/addon-1.0.0.xsd) defines also `hybrid` which may also match to this use case. For Mail: in the usual case there is an internet connection required to interact with the world, but the connection itself to the mail server may be also internal. This is similar to use case of `icalendar`, there it's just dependent where the calendar is located (defined by the user) but if it's in the internet, `cloud` may also match. Different from mail is, that the calendar usually gets any interaction by other users than the owner (which may be an external person).

jlaur: No, in this case it's a choice between `local` and `cloud`:
https://github.com/jlaur/openhab-docs/blob/6cf713c1cae012e0d417c46bf2582bf6c51c7a5c/developers/addons/addon.md#xml-structure-for-add-on-definitions

I realize there is still a gap, but I would opt for interpreting `local` as "non-cloud" even if it may not always be "local".

daMihe: I'm confused, your link also describes `hybrid`.

jlaur: Yes, and `none`, but I don't think any of them would apply here. The intention of `hybrid` is not to describe ambiguity, but rather a binding that works perfectly well within your own infrastructure, but can optionally use external cloud services for extended functionality. It will still work if the cloud services are shut down.

bundles/org.openhab.binding.hccrubbishcollection/src/main/resources/OH-INF/addon/addon.xml
```diff
@@ -6,5 +6,6 @@
 	<type>binding</type>
 	<name>HCC Rubbish Collection Binding</name>
 	<description>Get the rubbish collection dates for Hamilton City Council (New Zealand).</description>
+	<connection>cloud</connection>
```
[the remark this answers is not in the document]
mlobstein: done

bundles/org.openhab.binding.mail/src/main/resources/OH-INF/addon/addon.xml
mlobstein: > I would argue about this - see also [#14813 (comment)](https://github.com/openhab/openhab-addons/pull/14813#discussion_r1167470570)
> 
> The clear "cloud" case is when using an API provided by some specific vendor, i.e. "someone else's computer". Unfortunately that leaves some ambiguity for the cases where a standard or open protocol is implemented. For remoteopenhab we decided to use the criteria "the most common use-case" and defined it as `local`.

I was thinking of this comment: https://github.com/openhab/openhab-addons/pull/14690#issuecomment-1485551835

For NTP it is also a an open/RFC defined service.... But maybe the lack of a common use case of running an NTP server locally pushed that one to the 'cloud' category?

bundles/org.openhab.binding.icalendar/src/main/resources/OH-INF/addon/addon.xml
mlobstein: done

bundles/org.openhab.binding.mail/src/main/resources/OH-INF/addon/addon.xml
daMihe: Why? You could also set your home router as NTP peer as these devices usually offer one, which is quite good and has stable latencies. At least i would do. Another use case would be to monitor time differences of multiple devices. No internet needed at all, if the devices implement not only sntp. Same issue as for mail or icalendar.

jlaur: @mlobstein - ouch, you beat me with my own argument. 😉 I admit than even now the documentation for `local` is unclear. I think we are mixing things:
- Location: "local" (e.g. LAN) vs. "remote" (internet).
- Ownership: my server or someone else's server.
- Protocol: open protocol defined by RFC or vendor-specific protocol.
- Maybe even more?

For me - now - "cloud" is someone else's computer not using an open protocol. You cannot run it locally, and if the service shuts down, you cannot just move to another company implementing the same service. Essentially this means that cloud bindings can die overnight. By this definition, "local" is the opposite, so either something running entirely locally, or possibly outside your LAN for whichever reason. That could include remoteopenhab in the summerhouse, a friend's FTP server or your current ISP's SMTP server.

This means that I would now back paddle on the NTP comment and redefine it as "local".

This is just my opinion, and I still think the definition/documentation is not very clear. But I think the intention was to differentiate between traditional cloud (outside your control) and something can at least could potentially run locally.

jlaur (approved):

bundles/org.openhab.binding.mail/src/main/resources/OH-INF/addon/addon.xml
jlaur: I just merged the PR as it looks good to me, and I see no reason to block it. We can still continue this discussion and change declarations where appropriate. Perhaps it would be better to move the discussion to #14885, WDYT?

Markinus (commented):
LGTM

bundles/org.openhab.binding.mail/src/main/resources/OH-INF/addon/addon.xml
mlobstein: I agree, for the remainder of the bindings, I will judge cloud to mean connecting to a proprietary remote api (either via login credentials or a read-only scraper).

PR #14903, opened 2023-04-29
[easee] Fix start/stop due to changed API

### LINKED ISSUE #14898: [easee] recent changes in Easee-API break Start/Stop Channel
Easee ha added new status values that represent the charger operation status. This breaks existing logic in the binding as the new values are not considered.
see https://developer.easee.cloud/docs/enumerations 
opMode now has new values 7 and 8

alexf2015:
this fixes #14898


jlaur (approved):
Thanks for the fix. I know this was not a bug in the binding, but for the user it was something broken. What would you think about using the "bug" label to have it categorized in the release notes as something "fixed"?

alexf2015:
@jlaur I do not care if it is flagged as bug or enhancement. As you mentioned this fixes something which was broken. For the user it is not always obvious if this is a bug in the binding or a change. From a user perspective I also don't want to care about such details I just want it to work as expected :-)

Issue #13456, opened 2022-09-28
[hue] Implement Hue CLIP 2


andrewfg:
### Introduction

This issue is a place holder for the planned implementation of API v2 in the Hue binding.

Specifically it requires the following..
- Implement new CLIP 2 endpoints
- Implement new CLIP 2 DTOs
- Implement SSE

### Dependency

Any PR arising from this issue will depend on #11842 being merged first.

### Approach

I am tending to follow the same approach as I took with the HDPowerView API generation change as follows..

- Refactor the OH facing parts of the binding to use abstract base DTO classes (such as Scene, Light, etc.) and access the Bridge via an abstract base communication class HueBridge
- Relocate the current API V1 implementation code from Scene, Light, etc., and HueBridge into new SceneV1, LightV1, etc., and HueBridgeV1 implementation classes which implement/extend the abovementioned base abstract classes
- Write new SceneV2, LightV2, etc., and HueBridgeV2 implementation classes that implement/extend the base abstract classes too.
- Use a Config Parameter in the the thing handler to select whether the OH facing code instantiates/consumes the xxxV1 or xxxV2 implementation classes.

This allows the the OH facing parts of the binding to remain almost unchanged. And the V1 code would remain functionally unchanged too; although the actual code has to be split between the xxx base/abstract class, and the xxxV1 implementation class.

In the case of the HDPowerView migration, we have the good luck that the manufacturer chose to keep the V1 and V2 DTOs fairly similar, with many fields being common, and only a subset of fields having changed names and meanings. However in the case of the Hue migration, I can already see that the DTOs have changed MASSIVELY between V1 and V2, so I can already say that unlike the HDPowerView migration which I was able to write in a week or so, this Hue migration will certainly take a lot longer.

In any case, before I start writing code, I will first write a Class Structure diagram similar to the one I wrote for https://github.com/openhab/openhab-addons/issues/12678#issuecomment-1239297302 -- this should help to clarify if my above mentioned approach would work (or not)..

### Constraints

Two requests from @lolodomo ..
- Use the OH standards for the SSE implementation (JAX-RS +Jetty}.
- Keep the compatibility with the old bridge in case it does not support SSE.

---
Signed-off-by: Andrew Fiddian-Green <software@whitebear.ch>

andrewfg:
**Dead Code in API V1**

[...]

PR #14914, opened 2023-05-01
[hue] Restore UPnP discovery for old bridges

lolodomo:

### DIFF
--- a/bundles/org.openhab.binding.hue/src/main/java/org/openhab/binding/hue/internal/discovery/HueBridgeUPNPDiscoveryParticipant.java
+++ b/bundles/org.openhab.binding.hue/src/main/java/org/openhab/binding/hue/internal/discovery/HueBridgeUPNPDiscoveryParticipant.java
@@ -1,0 +1,162 @@
+/**
+ * Copyright (c) 2010-2023 Contributors to the openHAB project
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ */
+package org.openhab.binding.hue.internal.discovery;
+
+import static org.openhab.binding.hue.internal.HueBindingConstants.*;
+
+import java.net.URL;
+import java.util.Dictionary;
+import java.util.Map;
+import java.util.Set;
+
+import org.eclipse.jdt.annotation.NonNullByDefault;
+import org.eclipse.jdt.annotation.Nullable;
+import org.jupnp.model.meta.DeviceDetails;
+import org.jupnp.model.meta.ModelDetails;
+import org.jupnp.model.meta.RemoteDevice;
+import org.openhab.binding.hue.internal.handler.HueBridgeHandler;
+import org.openhab.core.config.discovery.DiscoveryResult;
+import org.openhab.core.config.discovery.DiscoveryResultBuilder;
+import org.openhab.core.config.discovery.DiscoveryService;
+import org.openhab.core.config.discovery.upnp.UpnpDiscoveryParticipant;
+import org.openhab.core.config.discovery.upnp.internal.UpnpDiscoveryService;
+import org.openhab.core.thing.Thing;
+import org.openhab.core.thing.ThingTypeUID;
+import org.openhab.core.thing.ThingUID;
+import org.osgi.service.component.ComponentContext;
+import org.osgi.service.component.annotations.Activate;
+import org.osgi.service.component.annotations.Component;
+import org.osgi.service.component.annotations.Modified;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * The {@link HueBridgeUPNPDiscoveryParticipant} is responsible for discovering new and removed Hue Bridges. It uses the
+ * central {@link UpnpDiscoveryService}.
+ *
+ * The discovery through UPnP was replaced by mDNS discovery for recent bridges (V2).
+ * For old bridges (V1), the UPnP discovery is still required (as mDNS is not implemented).
+ * This class allows discovering only old bridges using UPnP.
+ *
+ * @author Laurent Garnier - Initial contribution
+ */
+@Component(configurationPid = "discovery.hue")
+@NonNullByDefault
+public class HueBridgeUPNPDiscoveryParticipant implements UpnpDiscoveryParticipant {
+
+    private static final String EXPECTED_MODEL_NAME_PREFIX = "Philips hue bridge";
+
+    private final Logger logger = LoggerFactory.getLogger(HueBridgeUPNPDiscoveryParticipant.class);
+
+    private long removalGracePeriod = 0L;
+
+    private boolean isAutoDiscoveryEnabled = true;
+
+    @Activate
+    protected void activate(ComponentContext componentContext) {
+        activateOrModifyService(componentContext);
+    }
+
+    @Modified
+    protected void modified(ComponentContext componentContext) {
+        activateOrModifyService(componentContext);
+    }
+
+    private void activateOrModifyService(ComponentContext componentContext) {
+        Dictionary<String, @Nullable Object> properties = componentContext.getProperties();
+        String autoDiscoveryPropertyValue = (String) properties
+                .get(DiscoveryService.CONFIG_PROPERTY_BACKGROUND_DISCOVERY);
+        if (autoDiscoveryPropertyValue != null && !autoDiscoveryPropertyValue.isBlank()) {
+            isAutoDiscoveryEnabled = Boolean.valueOf(autoDiscoveryPropertyValue);
+        }
+        String removalGracePeriodPropertyValue = (String) properties.get(REMOVAL_GRACE_PERIOD);
+        if (removalGracePeriodPropertyValue != null && !removalGracePeriodPropertyValue.isBlank()) {
+            try {
+                removalGracePeriod = Long.parseLong(removalGracePeriodPropertyValue);
+            } catch (NumberFormatException e) {
+                logger.warn("Configuration property '{}' has invalid value: {}", REMOVAL_GRACE_PERIOD,
+                        removalGracePeriodPropertyValue);
+            }
+        }
+    }
+
+    @Override
+    public Set<ThingTypeUID> getSupportedThingTypeUIDs() {
+        return HueBridgeHandler.SUPPORTED_THING_TYPES;
+    }
+
+    @Override
+    public @Nullable DiscoveryResult createResult(RemoteDevice device) {
+        if (isAutoDiscoveryEnabled) {
+            DeviceDetails details = device.getDetails();
+            ThingUID uid = getThingUID(device);
+            if (details != null && uid != null) {
+                URL baseUrl = details.getBaseURL();
+                String serialNumber = details.getSerialNumber();
+                String label = details.getFriendlyName();
+                String modelName = EXPECTED_MODEL_NAME_PREFIX;
+                ModelDetails modelDetails = details.getModelDetails();
+                if (modelDetails != null && modelDetails.getModelName() != null
+                        && modelDetails.getModelNumber() != null) {
+                    modelName = String.format("%s (%s)", modelDetails.getModelName(), modelDetails.getModelNumber());
+                }
+                if (baseUrl != null && serialNumber != null && !serialNumber.isBlank() && label != null) {
+                    return DiscoveryResultBuilder.create(uid) //
+                            .withProperties(Map.of( //
+                                    HOST, baseUrl.getHost(), //
+                                    PORT, baseUrl.getPort(), //
+                                    PROTOCOL, baseUrl.getProtocol(), //
+                                    Thing.PROPERTY_MODEL_ID, modelName, //
+                                    Thing.PROPERTY_SERIAL_NUMBER, serialNumber.toLowerCase())) //
+                            .withLabel(label) //
+                            .withRepresentationProperty(Thing.PROPERTY_SERIAL_NUMBER) //
+                            .build();
+                }
+            }
+        }
+        return null;
+    }
+
+    @Override
+    public @Nullable ThingUID getThingUID(RemoteDevice device) {
+        DeviceDetails details = device.getDetails();
+        if (details != null) {
+            String serialNumber = details.getSerialNumber();
+            ModelDetails modelDetails = details.getModelDetails();
+            if (serialNumber != null && !serialNumber.isBlank() && modelDetails != null) {
+                String modelName = modelDetails.getModelName();
+                // Model name has the format "Philips hue bridge <year>" with <year> being 2012
+                // for a hue bridge V1 or 2015 for a hue bridge V2.
+                if (modelName != null && modelName.startsWith(EXPECTED_MODEL_NAME_PREFIX)) {
+                    boolean ignored = false;
+                    try {
+                        int year = Integer.parseInt(modelName.substring(EXPECTED_MODEL_NAME_PREFIX.length()));
+                        // The bridge is ignored if year is lower than 2015
+                        ignored = year < 2015;
+                    } catch (NumberFormatException e) {
+                        // No int value found, we consider the bridge
+                    }
+                    if (!ignored) {
+                        return new ThingUID(THING_TYPE_BRIDGE, serialNumber.toLowerCase());
+                    }
+                }
+            }
+        }
+        return null;
+    }
+
+    @Override
+    public long getRemovalGracePeriodSeconds(RemoteDevice device) {
+        return removalGracePeriod;
+    }
+}
--- a/bundles/org.openhab.binding.hue/src/main/java/org/openhab/binding/hue/internal/discovery/HueBridgeUPNPDiscoveryParticipant.java
+++ b/bundles/org.openhab.binding.hue/src/main/java/org/openhab/binding/hue/internal/discovery/HueBridgeUPNPDiscoveryParticipant.java
@@ -110,14 +110,14 @@ public class HueBridgeUPNPDiscoveryParticipant implements UpnpDiscoveryParticipa
         }
         URL baseUrl = details.getBaseURL();
         String serialNumber = details.getSerialNumber();
+        if (baseUrl == null || serialNumber == null || serialNumber.isBlank()) {
+            return null;
+        }
         String label = String.format(DISCOVERY_LABEL_PATTERN, baseUrl.getHost());
         String modelName = EXPECTED_MODEL_NAME_PREFIX;
         ModelDetails modelDetails = details.getModelDetails();
         if (modelDetails != null && modelDetails.getModelName() != null && modelDetails.getModelNumber() != null) {
             modelName = String.format("%s (%s)", modelDetails.getModelName(), modelDetails.getModelNumber());
-        }
-        if (baseUrl == null || serialNumber == null || serialNumber.isBlank()) {
-            return null;
         }
         return DiscoveryResultBuilder.create(uid) //
                 .withProperties(Map.of( //
--- a/bundles/org.openhab.binding.hue/src/main/java/org/openhab/binding/hue/internal/discovery/HueBridgeUPNPDiscoveryParticipant.java
+++ b/bundles/org.openhab.binding.hue/src/main/java/org/openhab/binding/hue/internal/discovery/HueBridgeUPNPDiscoveryParticipant.java
@@ -140,9 +140,9 @@ public class HueBridgeUPNPDiscoveryParticipant implements UpnpDiscoveryParticipa
                 if (modelName != null && modelName.startsWith(EXPECTED_MODEL_NAME_PREFIX)) {
                     boolean ignored = false;
                     try {
-                        int year = Integer.parseInt(modelName.substring(EXPECTED_MODEL_NAME_PREFIX.length()));
+                        int year = Integer.parseInt(modelName.substring(EXPECTED_MODEL_NAME_PREFIX.length()).trim());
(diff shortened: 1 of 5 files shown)


lolodomo:
Related to discussion in PR #14871

lolodomo:
@jlaur @andrewfg : to be discussed if the way I chose to ignore the recent bridges is OK and the best.

lolodomo:
> @jlaur @andrewfg : to be discussed if the way I chose to ignore the recent bridges is OK and the best.

Adjusted to what @andrewfg proposed in #14871

bundles/org.openhab.binding.hue/src/main/java/org/openhab/binding/hue/internal/discovery/HueBridgeUPNPDiscoveryParticipant.java
```diff
@@ -0,0 +1,162 @@
...
+            if (serialNumber != null && !serialNumber.isBlank() && modelDetails != null) {
+                String modelName = modelDetails.getModelName();
+                // Model name has the format "Philips hue bridge <year>" with <year> being 2012
+                // for a hue bridge V1 or 2015 for a hue bridge V2.
+                if (modelName != null && modelName.startsWith(EXPECTED_MODEL_NAME_PREFIX)) {
+                    boolean ignored = false;
+                    try {
+                        int year = Integer.parseInt(modelName.substring(EXPECTED_MODEL_NAME_PREFIX.length()));
+                        // The bridge is ignored if year is lower than 2015
+                        ignored = year < 2015;
```
jlaur, line 145: Shouldn't it be the opposite, i.e. newer bridges should be ignored?

I'm wondering if any of this logic is necessary since bridge V2 with new firmware can't be discovered by UPnP since Q2 2022, that seems like effective way of ignoring? Reference: https://developers.meethue.com/new-hue-api/

bundles/org.openhab.binding.hue/src/main/java/org/openhab/binding/hue/internal/discovery/HueBridgeUPNPDiscoveryParticipant.java
```diff
@@ -0,0 +1,162 @@
...
+ */
+@Component(configurationPid = "discovery.hue")
+@NonNullByDefault
+public class HueBridgeUPNPDiscoveryParticipant implements UpnpDiscoveryParticipant {
+
+    private static final String EXPECTED_MODEL_NAME_PREFIX = "Philips hue bridge";
+
+    private final Logger logger = LoggerFactory.getLogger(HueBridgeUPNPDiscoveryParticipant.class);
+
+    private long removalGracePeriod = 0L;
```
jlaur, line 61: Sorry for being lazy and asking: Is the property required, i.e. can we trust that it is always set? In the old version of the discovery class removed in #11842 the value was 50:
https://github.com/openhab/openhab-addons/pull/11842/files#diff-5cbad99227639ffbe9e3137aeb63db178c2062763f9b711aa7de8bf1ce5ef2a6L58-L59

bundles/org.openhab.binding.hue/src/main/java/org/openhab/binding/hue/internal/discovery/HueBridgeUPNPDiscoveryParticipant.java
lolodomo: Oups, you're right and my test should not have worked so I will have to test it again seriously.

@andrewfg shows this morning using Device Spy that UPnP is still available for bridge V2.

jlaur: > @andrewfg shows this morning using Device Spy that UPnP is still available for bridge V2.

Oh, I missed that, I need to catch up. Will test with my bridge also.

bundles/org.openhab.binding.hue/src/main/java/org/openhab/binding/hue/internal/discovery/HueBridgeUPNPDiscoveryParticipant.java
andrewfg: The removal grace period was introduced because sometimes the bridge is too late in sending its SSDP:ALIVE refresh notifications. This has the consequence that the bridge is frequently repeatedly added to and removed from the Inbox. The value that we had set at that time was sufficient to prevent that from occurring.

lolodomo: Yes, I remember. Strangely, I was never concerned by this problem.

bundles/org.openhab.binding.hue/src/main/java/org/openhab/binding/hue/internal/discovery/HueBridgeUPNPDiscoveryParticipant.java
jlaur: I tested with this line added after line 140:

```java
                    logger.info("Discovered Hue bridge: {}, {}, {}", modelDetails.getModelName(),
                            modelDetails.getModelNumber(), modelDetails.getModelDescription());
```

Result:
```
2023-05-01 21:30:38.087 [INFO ] [ry.HueBridgeUPNPDiscoveryParticipant] - Discovered Hue bridge: Philips hue bridge 2015, BSB002, Philips hue Personal Wireless Lighting
2023-05-01 21:30:38.088 [INFO ] [ry.HueBridgeUPNPDiscoveryParticipant] - Discovered Hue bridge: Philips hue bridge 2015, BSB002, Philips hue compatible Personal Wireless Lighting
```

So I can confirm that UPnP is still enabled with firmware version 1957113050. And I can also confirm that no discovery result is created by this code.

bundles/org.openhab.binding.hue/src/main/java/org/openhab/binding/hue/internal/discovery/HueBridgeUPNPDiscoveryParticipant.java
```diff
@@ -0,0 +1,163 @@
...
+                                    PROTOCOL, baseUrl.getProtocol(), //
+                                    Thing.PROPERTY_MODEL_ID, modelName, //
+                                    Thing.PROPERTY_SERIAL_NUMBER, serialNumber.toLowerCase())) //
+                            .withLabel(label) //
+                            .withRepresentationProperty(Thing.PROPERTY_SERIAL_NUMBER) //
+                            .build();
+                }
+            }
+        }
+        return null;
```
jlaur, line 127: Nesting could be reduced:
```suggestion
        if (!isAutoDiscoveryEnabled) {
            return null;
        }
        DeviceDetails details = device.getDetails();
        ThingUID uid = getThingUID(device);
        if (details == null || uid == null) {
            return null;
        }
        URL baseUrl = details.getBaseURL();
        String serialNumber = details.getSerialNumber();
        String label = details.getFriendlyName();
        String modelName = EXPECTED_MODEL_NAME_PREFIX;
        ModelDetails modelDetails = details.getModelDetails();
        if (modelDetails != null && modelDetails.getModelName() != null
                && modelDetails.getModelNumber() != null) {
            modelName = String.format("%s (%s)", modelDetails.getModelName(), modelDetails.getModelNumber());
        }
        if (baseUrl == null || serialNumber == null || serialNumber.isBlank() || label == null) {
            return null;
        }
        return DiscoveryResultBuilder.create(uid) //
                .withProperties(Map.of( //
                        HOST, baseUrl.getHost(), //
                        PORT, baseUrl.getPort(), //
                        PROTOCOL, baseUrl.getProtocol(), //
                        Thing.PROPERTY_MODEL_ID, modelName, //
                        Thing.PROPERTY_SERIAL_NUMBER, serialNumber.toLowerCase())) //
                .withLabel(label) //
                .withRepresentationProperty(Thing.PROPERTY_SERIAL_NUMBER) //
                .build();
```

bundles/org.openhab.binding.hue/src/main/java/org/openhab/binding/hue/internal/discovery/HueBridgeUPNPDiscoveryParticipant.java
```diff
@@ -0,0 +1,163 @@
...
+                        ignored = true;
+                    }
+                    if (!ignored) {
+                        return new ThingUID(THING_TYPE_BRIDGE, serialNumber.toLowerCase());
+                    }
+                }
+            }
+        }
+        return null;
+    }
```
jlaur, line 157: Nesting could be reduced:
```suggestion
    @Override
    public @Nullable ThingUID getThingUID(RemoteDevice device) {
        DeviceDetails details = device.getDetails();
        if (details == null) {
            return null;
        }
        String serialNumber = details.getSerialNumber();
        ModelDetails modelDetails = details.getModelDetails();
        if (serialNumber == null || serialNumber.isBlank() || modelDetails == null) {
            return null;
        }
        String modelName = modelDetails.getModelName();
        // Model name has the format "Philips hue bridge <year>" with <year> being 2012
        // for a hue bridge V1 or 2015 for a hue bridge V2.
        if (modelName == null || !modelName.startsWith(EXPECTED_MODEL_NAME_PREFIX)) {
            return null;
        }
        boolean ignored = false;
        try {
            int year = Integer.parseInt(modelName.substring(EXPECTED_MODEL_NAME_PREFIX.length()).trim());
            // The bridge is ignored if year is greater or equal to 2015
            ignored = year >= 2015;
        } catch (NumberFormatException e) {
            // No int value found, this bridge is ignored
            ignored = true;
        }
        if (!ignored) {
            return new ThingUID(THING_TYPE_BRIDGE, serialNumber.toLowerCase());
        }
        return null;
    }
```

jlaur (changes requested):
Minor comments about possibility to reduce nesting, otherwise LGTM.

bundles/org.openhab.binding.hue/src/main/java/org/openhab/binding/hue/internal/discovery/HueBridgeUPNPDiscoveryParticipant.java
lolodomo: Done

bundles/org.openhab.binding.hue/src/main/java/org/openhab/binding/hue/internal/discovery/HueBridgeUPNPDiscoveryParticipant.java
lolodomo: Done

bundles/org.openhab.binding.hue/src/main/java/org/openhab/binding/hue/internal/discovery/HueBridgeUPNPDiscoveryParticipant.java
```diff
@@ -0,0 +1,164 @@
...
+        boolean ignored = false;
+        try {
+            int year = Integer.parseInt(modelName.substring(EXPECTED_MODEL_NAME_PREFIX.length()).trim());
+            // The bridge is ignored if year is greater or equal to 2015
+            ignored = year >= 2015;
+        } catch (NumberFormatException e) {
+            // No int value found, this bridge is ignored
+            ignored = true;
+        }
+        return ignored ? null : new ThingUID(THING_TYPE_BRIDGE, serialNumber.toLowerCase());
```
jlaur, line 157: Sorry, with nesting reduced this now became more visible to me - could be simplified further to:
```suggestion
            if (year >= 2015) {
                return null;
            }
        } catch (NumberFormatException e) {
            // No int value found, this bridge is ignored
            return null;
        }
        return new ThingUID(THING_TYPE_BRIDGE, serialNumber.toLowerCase());
```

lolodomo:
Fighting again with the resolution in the integration tests !

bundles/org.openhab.binding.hue/src/main/java/org/openhab/binding/hue/internal/discovery/HueBridgeUPNPDiscoveryParticipant.java
```diff
@@ -0,0 +1,164 @@
...
+        }
+        String modelName = modelDetails.getModelName();
+        // Model name has the format "Philips hue bridge <year>" with <year> being 2012
+        // for a hue bridge V1 or 2015 for a hue bridge V2.
+        if (modelName == null || !modelName.startsWith(EXPECTED_MODEL_NAME_PREFIX)) {
+            return null;
+        }
+        boolean ignored = false;
+        try {
+            int year = Integer.parseInt(modelName.substring(EXPECTED_MODEL_NAME_PREFIX.length()).trim());
```
andrewfg, line 150: Suggested change 
```
        Pattern pattern = Pattern.compile("\\d{4}");
        Matcher matcher = pattern.matcher(modelName);
        int year = Integer.parseInt(matcher.find() ? matcher.group() : "0");
```

andrewfg (changes requested):

lolodomo:
> Fighting again with the resolution in the integration tests !

By miracle, I find a way to make the integration tests running again by adding stuff in itest.bndrun.

lolodomo:
Build finally succeeded. It took me more time to fix itest.bndrun than everything else in this PR !

bundles/org.openhab.binding.hue/src/main/java/org/openhab/binding/hue/internal/discovery/HueBridgeUPNPDiscoveryParticipant.java
lolodomo: Ok, done

bundles/org.openhab.binding.hue/src/main/java/org/openhab/binding/hue/internal/discovery/HueBridgeUPNPDiscoveryParticipant.java
lolodomo: Ok, done.

bundles/org.openhab.binding.hue/src/main/java/org/openhab/binding/hue/internal/discovery/HueBridgeUPNPDiscoveryParticipant.java
```diff
@@ -0,0 +1,170 @@
...
+            return null;
+        }
+        DeviceDetails details = device.getDetails();
+        ThingUID uid = getThingUID(device);
+        if (details == null || uid == null) {
+            return null;
+        }
+        URL baseUrl = details.getBaseURL();
+        String serialNumber = details.getSerialNumber();
+        String label = details.getFriendlyName();
```
jlaur, line 113: Perhaps this should be moved down and changed to:

```java
String label = String.format(DISCOVERY_LABEL_PATTERN, baseUrl.getHost());
```

for being consistent with #14919, WDYT? What will be the current value of this label?

lolodomo: "Philips hue (IP)"
So the difference is only the h of hue that is in lowercase. 
I am ok to rather use DISCOVERY_LABEL_PATTERN.

jlaur: It would only be a problem if NUPnP would also work for the old bridge, otherwise there is only one discovery method, so no collisions like we had for mDNS and NUPnP. However, I guess consistency still wouldn't hurt and it would be safer - for example if NUPnP would start to work again suddenly.

lolodomo: Yes

lolodomo: Done (and tested).

bundles/org.openhab.binding.hue/src/main/java/org/openhab/binding/hue/internal/discovery/HueBridgeUPNPDiscoveryParticipant.java
```diff
@@ -110,13 +110,13 @@ public Set<ThingTypeUID> getSupportedThingTypeUIDs() {
         }
         URL baseUrl = details.getBaseURL();
         String serialNumber = details.getSerialNumber();
-        String label = details.getFriendlyName();
+        String label = String.format(DISCOVERY_LABEL_PATTERN, baseUrl.getHost());
```
jlaur, line 113: Sorry, but the `baseUrl` null check is in line 119, so potential NPE here.

lolodomo: Arghhhh
I am going to fix that.

lolodomo: Done

jlaur (approved):

andrewfg:
@lolodomo I synched your code into #13456 but when testing my Jar on OH 4.0 SNAPSHOT build 3401 I got the following. It looks as if UPNP was removed from the OH core. Or ??

```
2023-05-05 18:03:33.478 [WARN ] [org.apache.felix.fileinstall        ] - Error while starting bundle: file:/G:/OH4/addons/org.openhab.binding.hue-4.0.0-SNAPSHOT.jar
org.osgi.framework.BundleException: Could not resolve module: org.openhab.binding.hue [239]
  Unresolved requirement: Import-Package: org.openhab.core.config.discovery.upnp

	at org.eclipse.osgi.container.Module.start(Module.java:463) ~[org.eclipse.osgi-3.18.0.jar:?]
	at org.eclipse.osgi.internal.framework.EquinoxBundle.start(EquinoxBundle.java:445) ~[org.eclipse.osgi-3.18.0.jar:?]
	at org.apache.felix.fileinstall.internal.DirectoryWatcher.startBundle(DirectoryWatcher.java:1260) ~[?:?]
	at org.apache.felix.fileinstall.internal.DirectoryWatcher.startBundles(DirectoryWatcher.java:1233) ~[?:?]
	at org.apache.felix.fileinstall.internal.DirectoryWatcher.startAllBundles(DirectoryWatcher.java:1221) ~[?:?]
	at org.apache.felix.fileinstall.internal.DirectoryWatcher.doProcess(DirectoryWatcher.java:515) ~[?:?]
	at org.apache.felix.fileinstall.internal.DirectoryWatcher.process(DirectoryWatcher.java:365) ~[?:?]
	at org.apache.felix.fileinstall.internal.DirectoryWatcher.run(DirectoryWatcher.java:316) ~[?:?]
```

jlaur:
> 2023-05-05 18:03:33.478 [WARN ] [org.apache.felix.fileinstall        ] - Error while starting bundle: file:/G:/OH4/addons/org.openhab.binding.hue-4.0.0-SNAPSHOT.jar
> org.osgi.framework.BundleException: Could not resolve module: org.openhab.binding.hue [239]
>   Unresolved requirement: Import-Package: org.openhab.core.config.discovery.upnp


See https://community.openhab.org/t/creating-a-binding-with-upnp-discovery/37839

andrewfg:
The mDNS feature is already installed automatically, but the UPNP is not. So there is still something fishy here. I suspect it is because UPNP was removed from the standard OH core distro.

lolodomo:
> So there is still something fishy here. I suspect it is because UPNP was removed from the standard OH core distro.

Of course not ! I am using several bindings relying on UPnP like sonos, lgwebos, ...

Please read my message, I think it is clear, the problem is that you just drop a jar in addons folder. In that case, some dependencies are not installed.
```
feature:list | grep upnp
openhab-core-io-transport-upnp                    x 4.0.0.SNAPSHOT   x          x Started     x distro-4.0.0-SNAPSHOT         x
openhab-transport-upnp                            x 4.0.0.SNAPSHOT   x          x Started     x distro-4.0.0-SNAPSHOT         x UPnP Transport
openhab.tp-jupnp                                  x 4.0.0.SNAPSHOT   x          x Started     x distro-4.0.0-SNAPSHOT         x UPnP/DLNA library for Java
openhab-binding-upnpcontrol                       x 4.0.0.SNAPSHOT   x          x Uninstalled x openhab-addons-4.0.0-SNAPSHOT x UPnP Control Binding
```
If not started, install the feature with `feature:install`.
Run `feature:list | grep upnp` to check if the feature is started. If not,  I do not remember if you must install openhab-core-io-transport-upnp or openhab-transport-upnp.

andrewfg:
EDIT: for the avoidance of doubt: I am testing on a bare OH 4 snapshot installation, with just one binding (hue) in addons. The mDNS disco loads and runs automatically. The UPNP disco does not load.

lolodomo:
Check also that in your branch, the feature.xml file contains the last changes. (dependency to UPnP).

andrewfg:
^
Both mDNS and UPNP are in the feature xml file. The former loads fine, the latter does not. There are no other Jars in addons and no other bindings. So the auto load of mDNS is presumably triggered by feature xml, and there is no reason why the UPNP is not found or loaded.

lolodomo:
What is defined in feature.xml is not considered when dropping the jar file in addons folder.
Did you try at least the command: `feature:install openhab-transport-upnp` ?

lolodomo:
Maybe the mDNS stuff is installed by default with the core framework (because OH needs to expose itself through mDNS). I have not checked but it looks probable.
UPnP stuff is not installed by default.

andrewfg:
> install OH4 without any binding and check what features are installed.

I confirm that mDNS (and not UPNP) is installed in bare system.

